Issue #1647, opened 2015-05-08
@JSExport on an `apply` method is confusing


sjrd:
Exporting a method named `apply`, like this:
```scala
@JSExport
def apply(...) = ...
```
is confusing. It exports a method named `apply` to JS, but the expected behavior would be to make the enclosing object callable (with the `()` syntax). The latter is not possible to implement, given the current state of JS. But we should emit a deprecation warning in this case. If exporting an `apply` method is really what's wanted, the following syntax should be used instead:
```scala
@JSExport("apply")
def apply(...) = ...
```
which would reconcile how `@JSName` and `@JSExport` view `apply` methods.

gzm0:
What about this?

```scala
@JSExportAll
class A {
  def apply(...) = ...
}
```

sjrd:
Oh! That's a nasty one. What about this: we emit a warning if there is an `@JSExportAll`, and an `apply` method without any `@JSExport` annotation. The proper way to avoid all warnings would therefore be:
```scala
@JSExportAll
class A {
  @JSExport("apply")
  def apply(...) = ...
}
```

gzm0:
Yeah, I was thinking of that as well. It sounds reasonable.

PR #1726, opened 2015-06-16
Fix #1717: Handle JS properties named apply

### LINKED ISSUE #1717: Val named apply in raw js type causes application
Consider this:

```scala
trait A extends js.Object {
  val apply: Int = js.native
}
```

reading `apply` on a value like this will actually result in an application:

Scala:
```scala
val a: A = // ...
a.apply
```

JavaScript:
```javascript
var a = // ...
a()
```

This is definitely not reasonable. We should select the property named `apply` or force the user to use `@JSName("apply")`. Further, `@JSExport` should mirror this behavior (it currently exports these as property, see #1647).

I am slightly in favor of forcing `@JSName`, to avoid accidental misunderstanding. (And properties named `apply` are not so common in JS).

gzm0:
A property in a facade type named apply needs to be explicitly made a
property (as opposed to a function application) by adding @JSName.

The changes in #1647 (@JSExport on method named apply) will align the
export behavior with this change.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -408,6 +408,13 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       if (sym.isAccessor)
         sym.accessed.getAnnotation(JSNameAnnotation).foreach(sym.addAnnotation)
 
+      if (jsInterop.isJSGetter(sym) && sym.name == nme.apply &&
+          !sym.hasAnnotation(JSNameAnnotation)) {
+        reporter.error(sym.pos, s"A member named apply represents function " +
+            "application in JavaScript. A parameterless member should be " +
+            "exported as a property. You must add @JSName(\"apply\")")
+      }
+
       if (jsInterop.isJSSetter(sym))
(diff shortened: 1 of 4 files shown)


gzm0:
Review by @sjrd

sjrd:
There should be a positive test for the case where the annotation `@JSName` is present.

That's all.

gzm0:
Updated

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -401,6 +401,20 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
 
+      /* If this is an accessor, copy a potential @JSName annotation from
+       * the field since otherwise it will get lost for traits (since they
+       * have no fields).
+       */
+      if (sym.isAccessor)
+        sym.accessed.getAnnotation(JSNameAnnotation).foreach(sym.addAnnotation)
+
+      if (jsInterop.isJSGetter(sym) && sym.name == nme.apply &&
+          !trgSym.hasAnnotation(JSNameAnnotation)) {
```
sjrd: This line doesn't compile.

gzm0:
Updated. Sorry for the compile error. Was a rebase mistake.

sjrd:
LGTM

PR #1728, opened 2015-06-16
Fix #1722: Disable optimizations in Rhino all the time.

### LINKED ISSUE #1722: Compiling code with scalatest leads to a compiler crash
Here is a minimalist project to reproduce.
https://github.com/ramnivas/scalajs-scalatest-large-method-bug

When I run "test" in sbt, I get the following crash:
```
> test

...

[info] Compiling 1 Scala source to /Users/ramnivas/test-dev/scalajs-scalatest-large-method-bug/target/scala-2.11/test-classes...
[trace] Stack trace suppressed: run last test:test for the full output.
[error] Could not run test paya.ScalaJSExampleTest: InternalError: Encountered code generation error while compiling script: generated bytecode for method exceeds 64K limit. (Lorg_scalatest_Resources$.sjsir#1)
[trace] Stack trace suppressed: run last test:executeTests for the full output.
[error] (test:executeTests) java.util.concurrent.TimeoutException: Futures timed out after [4625269705 nanoseconds]
[error] Total time: 22 s, completed Jun 15, 2015 2:18:58 PM
```

I tried switching to scalatest 3.0.0-SNAP5, but got the same crash.

sjrd:
We already disabled optimizations when including env.js. This commit generalizes this to any codebase.

This enables Rhino to correctly execute codebases that use ScalaTest. Since Rhino is slow anyway, and its only benefit is the out-of-the-box experience, it's better to make it work out-of-the-box with ScalaTest as well.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/rhino/RhinoJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/rhino/RhinoJSEnv.scala
@@ -131,6 +131,9 @@ final class RhinoJSEnv private (
     try {
       val scope = context.initStandardObjects()
 
+      // Rhino has trouble optimizing some big things, e.g., env.js or ScalaTest
+      context.setOptimizationLevel(-1)
+
       if (withDOM)
         setupDOM(context, scope)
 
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #1730, opened 2015-06-16
Miscellaneous fixes in version numbers.

sjrd:


sjrd:
Review by @gzm0

ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
```diff
@@ -2,19 +2,26 @@ package org.scalajs.core.ir
...
-  final val current = "0.6.4-SNAPSHOT"
+  /* DO NOT MAKE THESE 'final val's!
+   * When referring to these "constants" from separate libraries, if it is a
+   * 'final val', the value will be copied in the binaries of those libraries.
+   * If they are then linked to a different version of the IR artifact, their
+   * copy of these constants will not be updated.
+   */
+
+  /** Scala.js version. */
+  val current = "0.6.4-SNAPSHOT"
```
gzm0: Missing return type.

ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
```diff
@@ -2,19 +2,26 @@ package org.scalajs.core.ir
...
 
-  /** true iff the Scala.js version of this build is a snapshot version. */
-  final val currentIsSnapshot = current endsWith "-SNAPSHOT"
+  /** true iff the Scala.js version is a snapshot version. */
+  val currentIsSnapshot = current endsWith "-SNAPSHOT"
```
gzm0: Same

ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
```diff
@@ -2,19 +2,26 @@ package org.scalajs.core.ir
...
 
-  /** Version of binary IR this Scala.js version emits
+  /** Version of binary IR emitted by this version of Scala.js.
    *
    *  This should be either of:
    *  - a prior release version (i.e. "0.5.0", *not* "0.5.0-SNAPSHOT")
    *  - `current`
    */
-  final val binaryEmitted = current
+  val binaryEmitted = current
```
gzm0: Same

gzm0:
LGTM otherwise.

gzm0:
(It feels like we had the `MinorSnapshotVersion` thingy before... strange we removed it)

gzm0:
I assume you tested the regex?

sjrd:
> I assume you tested the regex?

Yes I did ^^

> (It feels like we had the MinorSnapshotVersion thingy before... strange we removed it)

We removed in commit 216f2faabd7a1eba42aa7e400ed5b6096ed43ece, probably because we thought that due to using `binaryEmitted`, we would not have this problem anymore. But that was without taking into account that, at times, even `binaryEmitted` is a minor snapshot.

sjrd:
Updated.

gzm0:
LGTM

PR #1735, opened 2015-06-17
Fix #1734: Fix {ASCII,latin1} decoding for inputs with !hasArray.

### LINKED ISSUE #1734: Charset decoding fails with read-only byte buffers
The scala.js version of `Charset#decode` fails to decode with a `MalformedInputException` when using charsets other than UTF-8 and when decoding a read-only byte buffer.

```scala
import java.nio.ByteBuffer
import java.nio.charset.Charset

def decode(charset: Charset, data: Array[Byte]) = {
  val decoder = charset.newDecoder
  //decoder.decode(ByteBuffer.wrap(data)).toString <-- works
  decoder.decode(ByteBuffer.wrap(data).asReadOnlyBuffer).toString // <-- doesn't work
}

val arr = Array[Byte](0x74, 0x65, 0x73, 0x74)
assert(decode(Charset.forName("UTF-8"), arr) == "test") // <-- passes for both RO and RW
assert(decode(Charset.forName("US-ASCII"), arr) == "test") // <-- passes for RW, throws for RW
assert(decode(Charset.forName("ISO-8859-1"), arr) == "test")  // <-- passes for RW, throws for RW
```

sjrd:


sjrd:
Another fix whose edit-distance = 1 :P (not taking into account the removal of the obsolete comment)

Review by @nicolasstucki

library/src/main/scala/scala/scalajs/niocharset/ISO_8859_1_And_US_ASCII_Common.scala
```diff
@@ -66,13 +66,10 @@ private[niocharset] abstract class ISO_8859_1_And_US_ASCII_Common protected (
...
           out.position(outPos - outOffset)
         } else {
-          /* Here, it's fine to read all the remaining bytes from the input,
-           * because we will *always* use all of them.
-           */
           var i = 0
           while (i != rem) {
             // Apparently ignoring the bit 7 in US_ASCII is the expected behavior
-            out.put((in.get(i).toInt & maxValue).toChar)
+            out.put((in.get().toInt & maxValue).toChar)
```
nicolasstucki: cool fix

nicolasstucki:
LGTM

PR #1720, opened 2015-06-15
Implement java.util.Collections

nicolasstucki:
Implemented `java.util.Collections` object.
Implemented tests for each function in the object and each wrapped
collection.
Modified the testing infrastructure for `Collection` and `Map` to
allow code reuse in `CollectionsTest`.

PR #1738, opened 2015-06-18
Fix #1737: Add configuration to include tests from require-jdk[7|8] folders.

### LINKED ISSUE #1737: Add JDK version specific tests infrastructure.
Add `testSuite/test/require-jdk7` and `testSuite/test/require-jdk8` source folders that only get compiled and executed on that version or a more recent one.

nicolasstucki:
Configuration loads tests in `testSuite/src/test/require-jdk7` when in JDK7 or JDK8.
Configuration loads tests in `testSuite/src/test/require-jdk8` when in JDK8.
Add placeholders for JDK version test.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -932,12 +932,14 @@ object Build extends sbt.Build {
           })),
 
           sources in Test ++= {
-            if (!scalaVersion.value.startsWith("2.10") &&
-                scalacOptions.value.contains("-Xexperimental")) {
-              (((sourceDirectory in Test).value / "require-sam") ** "*.scala").get
-            } else {
-              Nil
-            }
+            def listAllScalaFilesIf(testDir: String, condition: Boolean): Seq[File] =
+              if (condition) (((sourceDirectory in Test).value / testDir) ** "*.scala").get
+              else Nil
+            val javaVersion = System.getProperty("java.version")
+            listAllScalaFilesIf("require-jdk7", javaVersion.matches("1\\.[78].*")) ++
+            listAllScalaFilesIf("require-jdk8", javaVersion.startsWith("1.8")) ++
+            listAllScalaFilesIf("require-sam", !scalaVersion.value.startsWith("2.10") &&
+              scalacOptions.value.contains("-Xexperimental"))
           },
 
           /* Generate a scala source file that throws exceptions in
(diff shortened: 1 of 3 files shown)


nicolasstucki:
Review by @sjrd

project/Build.scala
```diff
@@ -932,12 +932,14 @@ object Build extends sbt.Build {
...
-              Nil
-            }
+            def listAllScalaFilesIf(testDir: String, condition: Boolean): Seq[File] =
+              if (condition) (((sourceDirectory in Test).value / testDir) ** "*.scala").get
+              else Nil
+            val javaVersion = System.getProperty("java.version")
+            listAllScalaFilesIf("require-jdk7", javaVersion.matches("1\\.[78].*")) ++
+            listAllScalaFilesIf("require-jdk8", javaVersion.startsWith("1.8")) ++
+            listAllScalaFilesIf("require-sam", !scalaVersion.value.startsWith("2.10") &&
+              scalacOptions.value.contains("-Xexperimental"))
```
sjrd: Indentation issue (should be 4).
Also, I would recommend breaking the line already before `!scalaVersion`, because as it stands, I read `&&` as being on the same level as `++`, which is weird.

test-suite/src/test/require-jdk7/org/scalajs/testsuite/OnJDK7OrHigher.scala
```diff
@@ -0,0 +1,22 @@
...
+
+object OnJDK7OrHigher extends JasmineTest {
+
+  describe("require-jdk7 tests") {
+    it("should only be running JDK7 or JDK8") {
+      // This method did not exist id JDK6 and hence this class will not link
+      // java.util.Collections.emptyIterator[Int]()
+      // TODO remove comment one the implementation is added
+    }
+  }
```
sjrd: This doesn't really make any sense. I don't think we should introduce a dummy test like this.

Come to think of it, I'm not sure this commit is useful on its own, and hence doesn't qualify as a "useful unit of change". I think this should simply be squashed with the first commit that needs to put something specific to JDK 7.

Or, you could test something useful that already exists, e.g., that `java.nio.charset.StandardCharsets` constants are identical to those in `scala.scalajs.niocharset.StandardCharsets`. For JDK 8, we don't anything yet, though, I think.

nicolasstucki:
Then I'll add it to #1720. And test the methods that are only there rather than this placeholders.

PR #1740, opened 2015-06-18
Fix #1736: Add a scalajsp input task in the sbt plugin.

### LINKED ISSUE #1736: A scalajsp command/task in the sbt plugin
I find myself wanting this all the time when I work on the compiler/optimizer. It's only the only feature that the CLI provides and the sbt plugin doesn't.

sjrd:
It accepts relative paths to .sjsir files on the classpath.

### DIFF
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -185,6 +185,11 @@ object ScalaJSPlugin extends AutoPlugin {
         "Private class loader to load jetty8 without polluting classpath. Only use this " +
         "as the `jettyClassLoader` argument of the PhantomJSEnv",
         KeyRanks.Invisible)
+
+    /** Prints the content of a .sjsir file in human readable form. */
+    val scalajsp = InputKey[Unit]("scalajsp",
+        "Prints the content of a .sjsir file in human readable form.",
+        CTask)
   }
 
   import autoImport._
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -3,8 +3,9 @@ package org.scalajs.sbtplugin
 import sbt._
 import sbt.inc.{IncOptions, ClassfileManager}
 import Keys._
-import sbinary.DefaultProtocol.StringFormat
+import sbinary.DefaultProtocol._
 import Cache.seqFormat
+import complete.Parser
 import complete.DefaultParsers._
 
 import Implicits._
@@ -23,8 +24,10 @@ import org.scalajs.jsenv.rhino.RhinoJSEnv
 import org.scalajs.jsenv.nodejs.NodeJSEnv
 import org.scalajs.jsenv.phantomjs.{PhantomJSEnv, PhantomJettyClassLoader}
 
+import org.scalajs.core.ir
 import org.scalajs.core.ir.Utils.escapeJS
 import org.scalajs.core.ir.ScalaJSVersions
+import org.scalajs.core.ir.Printers.{InfoPrinter, IRTreePrinter}
 
 import org.scalajs.testadapter.ScalaJSFramework
 
@@ -69,6 +72,11 @@ object ScalaJSPluginInternal {
   /** Lookup key for CompleteClasspath in attribute maps */
   val scalaJSCompleteClasspath =
       AttributeKey[CompleteClasspath]("scalaJSCompleteClasspath")
+
+  /** All .sjsir files on the fullClasspath, used by scalajsp. */
+  val sjsirFilesOnClasspath = TaskKey[List[String]]("sjsirFilesOnClasspath",
+      "All .sjsir files on the fullClasspath, used by scalajsp",
+      KeyRanks.Invisible)
 
   /** Patches the IncOptions so that .sjsir files are pruned as needed.
    *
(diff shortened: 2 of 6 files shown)


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/classpath/builder/JarTraverser.scala
```diff
@@ -61,9 +61,10 @@ trait JarTraverser extends ClasspathContentHandler with FileSystem {
     def entryBinaryContent: Array[Byte] = IO.readInputStreamToByteArray(in)
     def entryVersion: Option[String] = Some(entry.getTime.toString)
 
-    if (longName == JSDependencyManifest.ManifestFileName)
-      handleDepManifest(JSDependencyManifest.read(entryReader))
-    else if (longName.endsWith(".js")) {
+    if (longName == JSDependencyManifest.ManifestFileName) {
+      lazy val vf = JSDependencyManifest.read(entryReader)
+      handleDepManifest(vf)
```
gzm0: Why this change? Do you just think the line is too long? I think in any case you should make this a def. We don't need the lazy behavior.

tools/shared/src/main/scala/org/scalajs/core/tools/classpath/builder/JarTraverser.scala
```diff
@@ -74,7 +75,7 @@ trait JarTraverser extends ClasspathContentHandler with FileSystem {
       getOrCreateJSFile(relPath, fullPath, shortName)
         .withSourceMap(Some(entryContent))
     } else if (longName.endsWith(".sjsir")) {
-      val vf = new MemVirtualSerializedScalaJSIRFile(fullPath) {
+      lazy val vf = new MemVirtualSerializedScalaJSIRFile(fullPath) {
```
gzm0: Same here. Make this a `def`

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -133,8 +142,77 @@ object ScalaJSPluginInternal {
...
+  private def filterOutReflProxies(tree: ir.Trees.ClassDef): ir.Trees.ClassDef = {
+    import ir.Trees._
+    import ir.Definitions.isReflProxyName
+    val newDefs = tree.defs.filter {
+      case MethodDef(_, Ident(name, _), _, _, _) => !isReflProxyName(name)
+      case _ => true
+    }
+    tree.copy(defs = newDefs)(tree.optimizerHints)(tree.pos)
+  }
+
```
gzm0: Have you considered moving this to `tools` or even `ir`? It seems it could be generally useful...

tools/shared/src/main/scala/org/scalajs/core/tools/classpath/builder/JarTraverser.scala
sjrd: It's precisely to give lazy behavior, so that several accesses to the virtual file don't reload it every time.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: I did consider it, but I don't think it's generally useful. In theory one should never remove those. It's only useful for display. Besides, if they're in tools or ir, it becomes a public API, and I don't think this should be a public API, for the reasons outlined in the first sentence.

tools/shared/src/main/scala/org/scalajs/core/tools/classpath/builder/JarTraverser.scala
gzm0: I see... but shouldn't the callee handle that? It seems strange to assume that a callee unnecessarily evaluates a by name parameter multiple times...

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Ok.

tools/shared/src/main/scala/org/scalajs/core/tools/classpath/builder/JarTraverser.scala
sjrd: But it's true that the lazy behavior is not really needed. And a client who needs it can build it on its side.

sjrd: So OK, I'll revert this line, and make the other one a def.

sjrd:
Updated.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -185,6 +185,11 @@ object ScalaJSPlugin extends AutoPlugin {
         "Private class loader to load jetty8 without polluting classpath. Only use this " +
         "as the `jettyClassLoader` argument of the PhantomJSEnv",
         KeyRanks.Invisible)
+
+    /** Prints the content of a .sjsir file in human readable form. */
+    val scalajsp = InputKey[Unit]("scalajsp",
+        "Prints the content of a .sjsir file in human readable form.",
+        DSetting)
```
gzm0: This should probably be a `DTask`.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -133,8 +142,77 @@ object ScalaJSPluginInternal {
...
+        infos: Boolean = false,
+        showReflProxy: Boolean = false
+    )
+
+    val optionsParser: Parser[Options] = {
+      token(OptSpace ~> (
+          (literal("-i") | "--infos") ^^^ ((_: Options).copy(infos = true))
+        | (literal("-p") | "--reflProxies") ^^^ ((_: Options).copy(showReflProxy = true))
+      )).* map {
+        fns => fns.foldLeft(Options())((prev, fn) => fn(prev))
```
gzm0: Consider:

```scala
map { fns => Function.chain(fns)(Options()) }
```

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -3,8 +3,10 @@ package org.scalajs.sbtplugin
 import sbt._
 import sbt.inc.{IncOptions, ClassfileManager}
 import Keys._
-import sbinary.DefaultProtocol.StringFormat
+import sbinary.DefaultProtocol._
 import Cache.seqFormat
+import complete.Parser
+import complete.DefaultParsers
```
gzm0: This import is unused.

gzm0:
That's all. Very nice!

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: Oh neat. I didn't know that one!

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: Yep. Made it a CTask actually (sbt says C is "advanced tasks", which I think fits).

sjrd:
Updated.

gzm0:
LGTM

sjrd:
Updated with just a Scalastyle comment to let me use `return`.

Issue #997, opened 2014-09-02
scala.util.matching.Regex - case-insensitive Regex


erikhansendk:
I think this is a bug:

```
//  ScalaJVM
"(?i)abc".r.findAllIn("abcABC").mkString(",")   // "abc,ABC"

//  ScalaJS
"(?i)abc".r.findAllIn("abcABC").mkString(",")   // Uncaught SyntaxError: Invalid regular expression: /(?i)abc/: Invalid group 
```

Links:
http://daily-scala.blogspot.dk/2010/01/regular-expression-2-rest-regex-class.html
http://stackoverflow.com/questions/23115023/case-insensitive-pattern-matching-for-strings
http://www.scala-lang.org/old/node/415.html

PR #1741, opened 2015-06-19
Fix #1718: Validate the regex directly in Pattern.compile().

### LINKED ISSUE #1718: Pattern.compile doesn't validate regex
In Java/Scala, `Pattern.compile` fails immediately on invalid regex.
In ScalaJS, it isn't until the pattern is used that it fails.
```scala
// Doesn't throw in ScalaJS
val p = Pattern.compile("*")
```

To mimic JVM behaviour here (because I need to validate user-entered regex immediately), I just do this as a workaround:
```scala
// Throws an exception
p.matcher("").matches()
```

sjrd:
Since this requires the creation of a `js.RegExp` at the time `Pattern.compile()` is called anyway, we also store that native `RegExp` in the `Pattern` instance. `RegExp`'s for the `Matcher`s are created from that model `RegExp`, instead of from pattern and flags every time. This *might* be optimized in JS VMs to reuse the internally compiled representation for the regex.

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Pattern.scala
+++ b/javalib/src/main/scala/java/util/regex/Pattern.scala
@@ -4,33 +4,36 @@ import scala.annotation.switch
 
 import scala.scalajs.js
 
-final class Pattern private (pattern0: String, flags0: Int)
+final class Pattern private (jsRegExp: js.RegExp, _pattern: String, _flags: Int)
     extends Serializable {
 
   import Pattern._
 
-  def pattern(): String = pattern0
-  def flags(): Int = flags1
+  def pattern(): String = _pattern
+  def flags(): Int = _flags
 
-  private[regex] val (jspattern, flags1) = {
-    if ((flags0 & LITERAL) != 0) (quote(pattern0), flags0)
-    else {
-      trySplitHack(pattern0, flags0) orElse
-      tryFlagHack(pattern0, flags0) getOrElse
-      (pattern0, flags0)
+  override def toString(): String = pattern
+
+  private[regex] def newJSRegExp(): js.RegExp = {
+    val r = new js.RegExp(jsRegExp)
+    if (r ne jsRegExp) {
+      r
+    } else {
+      /* Workaround for the PhantomJS 1.x bug
+       * https://github.com/ariya/phantomjs/issues/11494
+       * which causes new js.RegExp(jsRegExp) to return the same object,
+       * rather than a new one.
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

gzm0:
This changes the behavior of `flags()`. IIUC it should also return the in-pattern flags (see #997).

gzm0:
Oh... I see.. No it's actually OK... But why on earth does `Pattern` have a `compile` method in the non-static context???

sjrd:
Er ... `compile()` *is* in the companion object, so in a static context. 

And there's a test asserting that `flags` includes the ones mentioned in the pattern, so I could not have broken that. (I had checked.)

gzm0:
> Er ... compile() is in the companion object, so in a static context.

I must have been drunk...

> And there's a test asserting that flags includes the ones mentioned in the pattern, so I could not have broken that. (I had checked.)

Yeah, I found the test later. It would have amazed me if I hadn't written a test for that when I introduced the flag hack :P

sjrd:
Updated. I had to work around a bug in PhantomJS 1.x, see https://github.com/scala-js/scala-js/pull/1741/files#diff-180072e950a6b3db11ffd0d53d383c7fR22

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -82,11 +85,33 @@ object Pattern {
...
+      }
+    }
+
+    val jsFlags = {
+      var f = "g"
+      if ((flags1 & CASE_INSENSITIVE) != 0)
+        f += "i"
+      if ((flags1 & MULTILINE) != 0)
+        f += "m"
+      f
```
nicolasstucki: Should write it in the same format as above (without the `var`):
```scala
val jsFlags = {
  "g" +
  (if ((flags1 & CASE_INSENSITIVE) != 0) "i" else "") +
  (if ((flags1 & MULTILINE) != 0) "m" else "")
}
```

nicolasstucki: Could also create a function that tests if the flag is on. Something that would be used like `containsFlag(flags1, CASE_INSENSITIVE)`. I see at least 3 places where it would be used.

nicolasstucki:
Thats all. Otherwise LGTM

sjrd:
Updated.

nicolasstucki:
LGTM

PR #1757, opened 2015-06-24
Fix #1756: Use the linking info in ScalaJSClassEmitter.

### LINKED ISSUE #1756: run/t4171.scala is broken on Rhino
It still works with `--fastOpt` and `--fullOpt`.

sjrd:
The issue was caused by the following pattern:

```scala
class A
class B extends A
println(classOf[B])
```

In this case, `B` is retained by the linker because its runtime type info are required. But `A` is discarded because it is not used in any way, despite it being a superclass of `B`. This is normal.

The problem is that the Rhino runner, when lazy loading `B`, did not use the fact that `linkedClassB.hasInstances` is false, and therefore generated the (empty) class. This empty class *did*
reference its parent class `A`, so when the script was evaluated, it would look for `A`. However it could not find `A` at all because it has been completely eliminated by the linker.

This commit fixes this by making `ScalaJSClassEmitter` leverage the linking info available in the `LinkedClass` it prints to emit only the necessary parts, basically with the same tests as those used
by `Emitter`.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
@@ -28,6 +28,7 @@ import org.scalajs.core.tools.optimizer.{LinkedClass, LinkingUnit}
 final class ScalaJSClassEmitter(semantics: Semantics, outputMode: OutputMode,
     globalInfo: LinkingUnit.GlobalInfo) {
 
+  import ScalaJSClassEmitter._
   import JSDesugaring._
 
   @deprecated("Use the constructor with an explicit output mode", "0.6.2")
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0, I guess.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
```diff
@@ -386,6 +390,13 @@ final class ScalaJSClassEmitter(semantics: Semantics, outputMode: OutputMode,
     case StringLiteral(value) => js.StringLiteral(value)(name.pos)
   }
 
+  private[tools] def needInstanceTests(tree: LinkedClass): Boolean = {
+    tree.hasInstanceTests || {
+      tree.hasRuntimeTypeInfo &&
+      ClassesWhoseDataReferToTheirInstanceTests.contains(tree.encodedName)
+    }
+  }
+
```
gzm0: Shouldn't this be on the companion? It doesn't need the instance IIUC.

gzm0: Hmmm... It might depend on the semantics in the future, doesn't it :-/

gzm0:
I'm wondering whether we should not move the checks for the individual `genX` methods inside the actual generators. This would allow us to remove the duplication of the checking logic. On the `Emitter` side we could just cache an `EmptyTree` if nothing is to be emitted...

WDYT?

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
sjrd: Yes, precisely ;-)

sjrd: Or, also likely, the `OutputMode`.

gzm0: Yep... or the global info :P

sjrd:
> I'm wondering whether we should not move the checks for the individual genX methods inside the actual generators.

It would make sense. But there are two complications:

* The [test for whether a class should be emitted](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala#L154) guards quite a lot of code, which does not correspond to a single method in `ScalaJSClassEmitter`. (Not a big deal, we could repeat the test.)
* The [emission for Strong Mode](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala#L218) is dramatically different, and doesn't have exactly the same tests. For example, static methods in a class without instances are emitted in a special way.

This requires more thought than a simple "let's put the tests inside the methods". I'll try and come up with something as a follow up cleanup commit.

gzm0:
Ok. LGTM

PR #1758, opened 2015-06-24
Fix #1646: Implement Character.is{Upper,Lower}Case as explicit regex.

### LINKED ISSUE #1646: Char#isUpper behavior diverges between Scala.js/Scala-JVM
```scala
scala> ')'.isUpper
false

scala-js> ')'.isUpper
true
```

`Char#isLower` also shares this behavior, discovered porting https://github.com/lihaoyi/fastparse to Scala.js

nicolasstucki:


nicolasstucki:
Review by @sjrd.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/CharacterTest.scala
```diff
@@ -679,5 +679,438 @@ object CharacterTest extends JasmineTest {
       expect(Character.isUnicodeIdentifierPart('\u2d2c')).toBeFalsy
     }
 
+    it("should provide isUpper") {
```
sjrd: isUpper<ins>Case</ins>

test-suite/src/test/scala/org/scalajs/testsuite/javalib/CharacterTest.scala
```diff
@@ -679,5 +679,438 @@ object CharacterTest extends JasmineTest {
...
+      expect(Character.isUpperCase('\u8473')).toBeFalsy
+      expect(Character.isUpperCase('\u478B')).toBeFalsy
+      expect(Character.isUpperCase('\uB916')).toBeFalsy
+      expect(Character.isUpperCase('\uBBFA')).toBeFalsy
+      expect(Character.isUpperCase('\uDC65')).toBeFalsy
+      expect(Character.isUpperCase('\u65E9')).toBeFalsy
+      expect(Character.isUpperCase('\u97FF')).toBeFalsy
+    }
+
+    it("should provide isLower") {
```
sjrd: isLower<ins>Case</ins>

sjrd:
The commit title is a bit too long (you can see it wrap in GitHub). You can shorten it by writing `Character.is{Upper,Lower}Case`.

sjrd:
That's all.

nicolasstucki:
Updated.

sjrd:
LGTM

sjrd:
retest this please

sjrd:
It seems this PR breaks checksizes.sh.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -303,6 +308,13 @@ object Character {
...
+
+
```
sjrd: And just noticed this now: Double blank line.

lihaoyi:
lol, and here I was generating the upper/lower cases using a [macro pre-computed bitset](https://github.com/lihaoyi/fastparse/blob/master/utils/shared/src/main/scala/fastparse/Utils.scala#L32-L37)

Will be so glad to start ripping those out when this lands and I can shrink the size of scalaparse by dozens of KB. Hopefully some day you'll be able to implement [the rest of the character predicates...](https://github.com/lihaoyi/fastparse/blob/master/fastparse/js/src/main/scala/fastparse/CharPredicates.scala)

FWIW pre-computing the bitset in JVM-land and looping over all 65536 characters could be a good way to verify the "this should behave the same as the JVM" invariant. Otherwise, even looking at the test suite, I have no idea whether it's verifying the right things or not...

Including the macro pre-computer and for-looping over everything may even be fewer lines of code than the manual asserts

nicolasstucki:
@lihaoyi, I grant you that the regex might not be the most compact representation. Our first goal is to solve the bug, in the we could worry about performance.

To generate this regex and the tests I actually looped through all the characters in the JVM and checked the values returned by the functions.

nicolasstucki:
Updated `checksizes.sh`

sjrd:
No change for 2.12.0-M1?

nicolasstucki:
Maybe there was a change, but I did not see a `checksizes.sh` failures on that version.

sjrd:
LGTM

sjrd:
Here's your failing build for 2.12.0-M1: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/46987/console ^^

sjrd:
And actually here 2.11.6 seems to have a worse version: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/46984/console (can sometimes depend on the JDK)

nicolasstucki:
Updated

PR #1762, opened 2015-06-25
Fix #1759: Use `new` when copying a TypedArray in Array.clone().

### LINKED ISSUE #1759: new Int8Array(n).toArray throws TypeError
```
new Int8Array(10).toArray
```
throws TypeError when testing using phantom.js
```
caused by: TypeError: Int8ArrayConstructor is not a constructor (evaluating 'this.u.constructor(this.u)')
```

in a browser, it throws
```
Uncaught TypeError: Constructor Int8Array requires 'new'
```

it looks like in scalajsenv.js, "this.u.constructor" is a constructor for Int8Array and is not called with "new".

sjrd:
The ES6 spec says that it is illegal to call a `TypedArray` constructor without `new`.

This does not come with any new test because we already have the tests that would trigger this error. But this we don't have any testing runtime that triggers it: the latest versions of Node.js and io.js are too lenient about this, and are fine with the absence of `new`.

### DIFF
--- a/tools/scalajsenv.js
+++ b/tools/scalajsenv.js
@@ -793,7 +793,7 @@ initArray(
         return new ArrayClass(this.u["slice"](0));
       else
         // The underlying Array is a TypedArray
-        return new ArrayClass(this.u.constructor(this.u));
+        return new ArrayClass(new this.u.constructor(this.u));
     };
   };
 //!endif
--- a/tools/strongmodeenv.js
+++ b/tools/strongmodeenv.js
@@ -769,7 +769,7 @@ class $TypeData {
           return new ArrayClass(this.u["slice"](0));
         else
           // The underlying Array is a TypedArray
-          return new ArrayClass(this.u.constructor(this.u));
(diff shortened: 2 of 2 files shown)


sjrd:
Review by @gzm0 
Note that, although this definitely fixes a misuse of the TypedArray spec, I don't have any proof that this actually fixes #1759 as seen from PhantomJS. I would have to test with PhantomJS 2 at home, on Windows. At the moment, there does not seem to be any binaries available for PhantomJS 2 on Linux, and I don't really want to spend time building it from source ...

sjrd:
The failure is spurious. I'll wait for a review before retesting.

sjrd:
retest this please

gzm0:
LGTM

gzm0:
There is a typo in the commit message: "But ~~this~~..."

sjrd:
Fixed the commit message.

weihsiu:
ok, i repeated the comment at https://github.com/scala-js/scala-js/issues/1759

PR #1770, opened 2015-06-29
Fix #1693: Implement java.util.concurrent.CopyOnWriteArrayList.

### LINKED ISSUE #1693: Implement java.util.concurrent.CopyOnWriteArrayList
Implementation for `java.util.concurrent.CopyOnWriteArrayList` as a `js.Array` wrapper.

nicolasstucki:
* Implement `java.util.concurrent.CopyOnWriteArrayList` without copy on each write.
  Only make a copy if there might be an iterator using the array (simplification due to single-threading).
* Implement tests for `CopyOnWriteArrayList`.

### DIFF
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -5,7 +5,7 @@ import scala.collection.JavaConversions._
 import scala.annotation.tailrec
 
 abstract class AbstractList[E] protected () extends AbstractCollection[E]
-                                               with List[E] {
+    with List[E] {
   self =>
 
   override def add(element: E): Boolean = {
@@ -172,18 +172,6 @@ private abstract class AbstractListView[E](protected val list: List[E],
     toIndex += delta
 }
 
-private trait SizeChangeEvent {
-  protected var end: Int
-
-  @inline
-  protected final def changeSize(delta: Int): Unit = {
-    end += delta
-    onSizeChanged(delta)
-  }
-
-  protected def onSizeChanged(delta: Int): Unit = () // override if needed
-}
-
 /* BackedUpListIterator implementation assumes that the underling list is not
  * necessarily on a RandomAccess list. Hence it wraps the underling list
  * iterator and assumes that this one is more efficient than accessing
@@ -228,56 +216,18 @@ private class BackedUpListIterator[E](innerIterator: ListIterator[E], fromIndex:
 /* RandomAccessListIterator implementation assumes that the has an efficient
  * .get(index) implementation.
  */
-private class RandomAccessListIterator[E](list: List[E], private var i: Int,
-    start: Int, override protected var end: Int) extends ListIterator[E] with SizeChangeEvent {
+private class RandomAccessListIterator[E](list: List[E], i: Int, start: Int, end: Int)
+    extends AbstractRandomAccessListIterator[E](i, start, end) {
 
-  private var last = -1
+  protected def get(index: Int): E =
+    list.get(index)
 
-  def hasNext(): Boolean =
-    i < end
+  protected def set(index: Int, e: E): Unit =
+    list.set(index, e)
 
-  def next(): E = {
-    last = i
-    i += 1
-    list.get(last)
-  }
+  protected def remove(index: Int): Unit =
+    list.remove(index)
 
-  def hasPrevious(): Boolean =
-    start < i
-
-  def previous(): E = {
-    i -= 1
-    last = i
-    list.get(last)
-  }
-
-  def nextIndex(): Int = i
-
-  def previousIndex(): Int = i - 1
-
-  def remove(): Unit = {
-    checkThatHasLast()
-    list.remove(last)
-    if (last < i)
-      i -= 1
-    last = -1
-    changeSize(-1)
-  }
-
-  def set(e: E): Unit = {
-    checkThatHasLast()
-    list.set(last, e)
-  }
-
-  def add(e: E): Unit = {
-    list.add(i, e)
-    changeSize(1)
-    last = -1
-    i += 1
-  }
-
-  private def checkThatHasLast(): Unit = {
-    if (last == -1)
-      throw new IllegalStateException
-  }
+  protected def add(index: Int, e: E): Unit =
+    list.add(index, e)
 }
--- a/javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
+++ b/javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
@@ -1,0 +1,338 @@
+package java.util.concurrent
+
+import java.lang.{reflect => jlr}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+
+import scala.scalajs._
+
+import java.util._
+
+class CopyOnWriteArrayList[E] private (private var inner: js.Array[E])
+    extends List[E] with RandomAccess with Cloneable with Serializable {
+  self =>
+
+  // requiresCopyOnWrite is false if and only if no other object
+  // (like the iterator) may have a reference to inner
+  private var requiresCopyOnWrite = false
+
+  def this() = {
+    this(new js.Array[E])
+  }
+
+  def this(c: Collection[_ <: E]) = {
+    this()
+    addAll(c)
+  }
+
+  def this(toCopyIn: Array[E]) = {
+    this()
+    toCopyIn.foreach(add)
+  }
+
+  def size(): Int =
+    inner.size
+
+  def isEmpty: Boolean =
+    size == 0
+
+  def contains(o: scala.Any): Boolean =
+    iterator.exists(o === _)
+
+  def indexOf(o: scala.Any): Int =
+    indexOf(o.asInstanceOf[E], 0)
+
+  def indexOf(e: E, index: Int): Int = {
+    checkIndexInBounds(index)
+    index + listIterator(index).indexWhere(_ === e)
+  }
+
+  def lastIndexOf(o: scala.Any): Int =
+    lastIndexOf(o.asInstanceOf[E], 0)
+
+  def lastIndexOf(e: E, index: Int): Int = {
+    @tailrec
+    def findIndex(iter: ListIterator[E]): Int = {
+      if (!iter.hasPrevious) -1
+      else if (iter.previous() === e) iter.nextIndex
+      else findIndex(iter)
+    }
+    findIndex(listIterator(size))
+  }
+
+  override def clone(): AnyRef =
+    new CopyOnWriteArrayList[E](this)
+
+  def toArray(): Array[AnyRef] =
+    toArray(new Array[AnyRef](size))
+
+  def toArray[T](a: Array[T]): Array[T] = {
+    val componentType = a.getClass.getComponentType
+    val toFill: Array[T] =
+      if (a.size >= size) a
+      else jlr.Array.newInstance(componentType, size).asInstanceOf[Array[T]]
+
+    val iter = iterator
+    for (i <- 0 until size)
+      toFill(i) = iter.next().asInstanceOf[T]
+    if (toFill.size > size)
+      toFill(size) = null.asInstanceOf[T]
+    toFill
+  }
+
+  def get(index: Int): E = {
+    checkIndexInBounds(index)
+    innerGet(index)
+  }
+
+  def set(index: Int, element: E): E = {
+    checkIndexInBounds(index)
+    copyIfNeeded()
+    val oldValue = innerGet(index)
+    innerSet(index, element)
+    oldValue
+  }
+
+  def add(e: E): Boolean = {
+    copyIfNeeded()
+    innerPush(e)
+    true
+  }
+
+  def add(index: Int, element: E): Unit = {
+    checkIndexOnBounds(index)
+    copyIfNeeded()
+    innerSplice(index, 0, element)
+  }
+
+  def remove(index: Int): E = {
+    checkIndexInBounds(index)
+    copyIfNeeded()
+    innerSplice(index, 1)(0)
+  }
+
+  def remove(o: scala.Any): Boolean = {
+    val index = indexOf(o)
+    if (index == -1) false else {
+      remove(index)
+      true
+    }
+  }
+
+  def addIfAbsent(e: E): Boolean = {
+    if (contains(e)) false else {
+      copyIfNeeded()
+      innerPush(e)
+      true
+    }
+  }
+
+  def containsAll(c: Collection[_]): Boolean =
+    c.iterator.forall(this.contains(_))
+
+  def removeAll(c: Collection[_]): Boolean = {
+    copyIfNeeded()
+    c.foldLeft(false)((prev, elem) => remove(elem) || prev)
+  }
+
+  def retainAll(c: Collection[_]): Boolean = {
+    val iter = iterator()
+    clear()
+    var modified = false
+    for (elem <- iter) {
+      if (c.contains(elem))
+        innerPush(elem)
+      else
+        modified = true
+    }
+    modified
+  }
+
+  def addAllAbsent(c: Collection[_ <: E]): Int = {
+    val elemsToAdd = c.iterator().filter(!contains(_)).toList
+    copyIfNeeded()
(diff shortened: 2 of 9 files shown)


nicolasstucki:
Review by @sjrd

andreaTP:
it was on my list! thanks a lot @nicolasstucki !

nicolasstucki:
Cheers @andreaTP

javalib/src/main/scala/java/util/AbstractList.scala
```diff
@@ -228,7 +228,23 @@ private class BackedUpListIterator[E](innerIterator: ListIterator[E], fromIndex:
 /* RandomAccessListIterator implementation assumes that the has an efficient
  * .get(index) implementation.
  */
-private class RandomAccessListIterator[E](list: List[E], private var i: Int,
+private class RandomAccessListIterator[E](list: List[E], i: Int, start: Int, end: Int)
+  extends AbstractRandomAccessListIterator[E](i, start, end) {
```
sjrd: 4 spaces

javalib/src/main/scala/java/util/AbstractList.scala
```diff
@@ -228,7 +228,23 @@ private class BackedUpListIterator[E](innerIterator: ListIterator[E], fromIndex:
...
+    list.set(index, e)
+
+  protected def remove(index: Int): Unit =
+    list.remove(index)
+
+  protected def add(index: Int, e: E): Unit =
+    list.add(index, e)
+}
+
+abstract private[util] class AbstractRandomAccessListIterator[E](private var i: Int,
```
sjrd: This class should be in its own file.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+import java.lang.{reflect => jlr}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+
+import scala.scalajs._
+
+class CopyOnWriteArrayList[E](private var inner: js.Array[E])
+    extends AbstractCopyOnWriteArrayList[E] with List[E] with RandomAccess
+    with Cloneable with Serializable { self =>
```
sjrd: `self =>` should be on its own line, indented 2 spaces, as soon as the entire class declaration does not fit on one line.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+
+  // requiresCopyOnWrite is false if and only if no other object (like the iterator)
+  // may have a reference to inner
+  private var requiresCopyOnWrite = false
+
+  def this() = {
+    this(new js.Array[E])
```
sjrd: Any reason this is not the primary constructor? It seems all the other constructors call this one anyway.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+  def this(toCopyIn: Array[E]) = {
+    this()
+    toCopyIn.foreach(add)
+  }
+
+  def size(): Int =
+    inner.size
+
+  override def clone(): AnyRef =
+    new CopyOnWriteArrayList[E](this)
```
sjrd: Could `clone()` return a new instance backed by the same `inner` with `requiresCopyOnWrite = true`?

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+    requiresCopyOnWrite = false
+  }
+
+  def listIterator(index: Int): ListIterator[E] = {
+    checkIndexOnBounds(index)
+    new CopyOnWriteArrayListIterator[E](innerSnapshot(), index, 0, size)
+  }
+
+  def subList(fromIndex: Int, toIndex: Int): List[E] = {
+    if (fromIndex < 0 || fromIndex > toIndex || toIndex > size )
```
sjrd: Spurious space before `)`

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+      throw new  IndexOutOfBoundsException
```
sjrd: Double space

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
sjrd: Ah, no, only iterators have the copy-on-write-behaviors. OK.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+
+  protected def innerSnapshot(): js.Array[E] = {
+    requiresCopyOnWrite = true
+    inner
+  }
+
+  class CopyOnWriteArrayListView(fromIndex: Int, private var toIndex: Int)
+      extends AbstractCopyOnWriteArrayList[E] { viewSelf =>
```
sjrd: `viewSelf =>` on its own line.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+    def listIterator(index: Int): ListIterator[E] = {
+      checkIndexOnBounds(index)
+      new CopyOnWriteArrayListIterator[E](innerSnapshot(), fromIndex + index, fromIndex, toIndex) {
+        override protected def onSizeChanged(delta: Int): Unit = changeSize(delta)
+      }
+    }
+
+    def subList(fromIndex: Int, toIndex: Int): List[E] = {
+      if (fromIndex < 0 || fromIndex > toIndex || toIndex > size )
+        throw new  IndexOutOfBoundsException
```
sjrd: Space before `)` + double space.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+
+import java.lang.{reflect => jlr}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+
+import scala.scalajs._
+
+class CopyOnWriteArrayList[E](private var inner: js.Array[E])
+    extends AbstractCopyOnWriteArrayList[E] with List[E] with RandomAccess
```
sjrd: I'm afraid you can't introduce an `AbstractCopyOnWriteArrayList` here, because it's not part of the declaration of the Java class.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+      toFill(size) = null.asInstanceOf[T]
+    toFill
+  }
+
+  def get(index: Int): E = {
+    checkIndexInBounds(index)
+    innerGet(index)
+  }
+
+
```
sjrd: Double blank line.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,348 @@
...
+    if (fromIndex < 0 || fromIndex > toIndex || toIndex > size )
+      throw new  IndexOutOfBoundsException
+    new CopyOnWriteArrayListView(fromIndex, toIndex)
+  }
+
+  protected def innerGet(index: Int): E =
+    inner(index)
+
+  protected def innerSet(index: Int, elem: E): Unit =
+    inner(index) = elem
```
sjrd: Why doesn't this call `copyIfNeeded()`? And why doesn't it set `requiresCopyOnWrite = true` ?

test-suite/src/test/scala/org/scalajs/testsuite/javalib/CopyOnWriteArrayListTest.scala
```diff
@@ -0,0 +1,101 @@
...
+        expect(list.get(i)).toEqual(42)
+    }
+
+    it ("should implement a snapshot iterator") {
+      val list = listFactory.empty[Int]
+      list.addAll(0 to 10)
+
+      val iter =  list.iterator()
+      list.clear()
+      val iter2 =  list.iterator()
```
sjrd: Double space, twice.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
nicolasstucki: No reason. I'll change it.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
nicolasstucki: I could have, but then both objects would have to copy the array on the next addition.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
nicolasstucki: Because it assumes that `requiresCopyOnWrite == false` and it is called only by
```scala
def set(index: Int, element: E): E = {
  checkIndexInBounds(index)
  copyIfNeeded()
  val oldValue = innerGet(index)
  innerSet(index, element)
  oldValue
}
```

sjrd: Wouldn't it make more sense to move `copyIfNeeded()` from `set(index, elem)` to `innerSet()`? `copyIfNeeded()` looks more of an implementation detail related to `innerSet` than `set`, because `copyIfNeeded()` is manipulating `inner`, at the same level of abstraction as `innerSet`. The same reasoning applies to other `innerMutate` methods.

nicolasstucki: The `innerSet` meant as a bridge to the `inner.set` and it's supposed to be overriden by the view. I left the copy is needed in the `set` method to have that check in only one place.

sjrd: Hum OK.

javalib/src/main/scala/java/util/AbstractList.scala
```diff
@@ -228,56 +216,18 @@ private class BackedUpListIterator[E](innerIterator: ListIterator[E], fromIndex:
...
-  def next(): E = {
-    last = i
-    i += 1
-    list.get(last)
-  }
-
-  def hasPrevious(): Boolean =
-    start < i
+private class RandomAccessListIterator[E](list: List[E], i: Int, start: Int, end: Int)
+  extends AbstractRandomAccessListIterator[E](i, start, end) {
```
sjrd: 4 spaces

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
...
+package concurrent
+
+import java.lang.{reflect => jlr}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+
+import scala.scalajs._
+
+class CopyOnWriteArrayList[E] private(private var inner: js.Array[E])
```
sjrd: Missing space before `(`

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
...
+    extends List[E] with RandomAccess with Cloneable with Serializable {
+  self =>
+
+  // requiresCopyOnWrite is false if and only if no other object
+  // (like the iterator) may have a reference to inner
+  private var requiresCopyOnWrite = false
+
+  def this() = {
+    this(new js.Array[E])
```
sjrd: Reminder about this constructor as primary constructor, since the previous comment disappeared.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
...
+    new CopyOnWriteArrayList[E](this)
+
+  def toArray(): Array[AnyRef] =
+    toArray(new Array[AnyRef](size))
+
+  def toArray[T](a: Array[T]): Array[T] = {
+    val toFill: Array[T] =
+      if (a.size >= size) a
+      else jlr.Array.newInstance(
+          a.getClass.getComponentType, size).asInstanceOf[Array[T]]
```
sjrd: The `else` branch must be on its line if it's too long. If it has more than one line, it needs `{}`. The then branch must do the same. And then the entire if/else needs `{}`.

Or, you decide to force it one line despite the long line, on the grounds that all these additional `{}` will make it much uglier.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
...
+      requiresCopyOnWrite = false
+    }
+  }
+
+  protected def innerSnapshot(): js.Array[E] = {
+    requiresCopyOnWrite = true
+    inner
+  }
+
+  class CopyOnWriteArrayListView(fromIndex: Int, private var toIndex: Int)
```
sjrd: This should probably be `private`.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
+package java.util
+package concurrent
```
sjrd: Don't use multiple `package` declarations. Prefer this:
```scala
package java.util.concurrent

import java.util._
```

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
```diff
@@ -0,0 +1,337 @@
...
+      throw new IndexOutOfBoundsException(index.toString)
+  }
+
+  protected def checkIndexOnBounds(index: Int): Unit = {
+    if (index < 0 || index > size)
+      throw new IndexOutOfBoundsException(index.toString)
+  }
+}
+
+class CopyOnWriteArrayListIterator[E](arraySnapshot: js.Array[E], i: Int, start: Int, end: Int)
```
sjrd: This should probably be `private`.

sjrd:
That's all.

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
nicolasstucki: If I use this one as primary constructor I end up initialising a new `js.Array` when creating a new `View`. This is only the case since I removed the `AbstractCopyOnWriteArrayList`. Before this change it should have been the primary constructor.

sjrd: Ah. Hum the `CopyOnWriteArrayListView` extending `CopyOnWriteArrayList` is suspicious, especially since the `inner` is set to `null` :-s
How much code duplication does this really avoid?

nicolasstucki:
Updated

javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala
nicolasstucki: A lot because `CopyOnWriteArrayList ` inherits directly the `List` interface and hence has no implementations whatsoever.

sjrd: OK.

sjrd:
LGTM

PR #1772, opened 2015-06-29
Fix #1764: Fix BigInteger modInverseMontgomery method.

### LINKED ISSUE #1764: BigInt / BigInteger modInverse always throw an Exception
A call to method modInverse in BigInt systematically throws an exception.
Scala.js version : 0.6.3

Code sample to reproduce the bug:

     // test Scala BigInt available since scala.js 0.6.3
     import scala.math.BigInt
     val size = 32  // small size here as prime generation is very long 
     val n = BigInt(size, 50, rand)  // our modulus is a prime number

     val g = BigInt(size, rand)
     val e = BigInt(size, rand)
     println("n = " + n.toString(16))
     println("g = " + g.toString(16))
     println("e = " + e.toString(16))
     
     val res = g.modPow(e, n)  // modPow is OK
     println("g^e mod n = " + res.toString(16))
     
     // compute inverse of g mod n
     val inv = g.modInverse(n)  // pb -> exception
     println("g^-n mod n = " + inv.toString(16))  
    

Javascript console output:

Uncaught java.lang.ArithmeticException: BigInteger not invertible.
$c_Ljava_math_BigInteger.modInverse__Ljava_math_BigInteger__Ljava_math_BigInteger @ BigInteger.scala:532
$c_s_math_BigInt.modInverse__s_math_BigInt__s_math_BigInt @ BigInt.scala:275
$c_Lorangelabs_TestApp$.main__V @ TestApp.scala:140

The problem seems to occur in Division.modInverseMontgomery(...)

nicolasstucki:


nicolasstucki:
Review by @sjrd

test-suite/src/test/scala/org/scalajs/testsuite/javalib/math/BigIntegerModPowTest.scala
```diff
@@ -196,6 +196,16 @@ object BigIntegerModPowTest extends JasmineTest {
       expect(result.signum()).toEqual(1)
     }
 
+    it("testmodInverse: #1764") {
```
sjrd: We typically use the notation `title - #1764` for test cases with an issue.

sjrd:
That's all.

nicolasstucki:
Updated

sjrd:
LGTM

PR #1776, opened 2015-06-30
Port java.util.concurrent.ConcurrentLinkedQueue.

andreaTP:
ConcurrentLinkedQueue backed by a java.util.LinkedList,
there is also a fix of peek and poll methods of LinkedList that were bugged

### DIFF
--- a/javalib/src/main/scala/java/util/LinkedList.scala
+++ b/javalib/src/main/scala/java/util/LinkedList.scala
@@ -211,16 +211,20 @@ class LinkedList[E]() extends AbstractSequentialList[E]
   }
 
   def peekFirst(): E =
-    head.value
+    if (head eq null) null.asInstanceOf[E]
+    else head.value
 
   def peekLast(): E =
-    peekFirst()
+    if (last eq null) null.asInstanceOf[E]
+    else last.value
 
   def pollFirst(): E =
-    removeFirst()
+    if (isEmpty()) null.asInstanceOf[E]
+    else removeFirst()
 
   def pollLast(): E =
-    removeLast()
+    if (isEmpty) null.asInstanceOf[E]
+    else removeLast()
 
   def push(e: E): Unit =
     addFirst(e)
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
@@ -1,0 +1,153 @@
+package java.util.concurrent
+
+import java.util._
+import scala.collection.JavaConversions._
+
+class ConcurrentLinkedQueue[E]()
+    extends AbstractQueue[E] with Queue[E] with Serializable {
+
+  def this(c: Collection[_ <: E]) = {
+    this()
+    addAll(c)
+  }
+
+  import ConcurrentLinkedQueue._
+
+  private var head: Node[E] = null
+  private var last: Node[E] = null
+
+  private var _size: Double = 0
+
+  override def add(e: E): Boolean = {
+    if (e == null) {
+      throw new NullPointerException()
+    } else {
+      val oldLast = last
+
+      last = new Node(e)
+
+      _size += 1
+
+      if (oldLast ne null)
+        oldLast.next = last
+      else
+        head = last
+
+      true
+    }
+  }
+
+  override def offer(e: E): Boolean =
+    add(e)
+
+  override def poll(): E = {
+    if (isEmpty()) null.asInstanceOf[E]
+    else {
+      val oldHead = head
+      head = oldHead.next
+
+      if (head eq null)
+        last = null
+
+      _size -= 1
+      oldHead.value
+    }
+  }
+
+  override def peek(): E =
+    if (isEmpty()) null.asInstanceOf[E]
+    else head.value
+
+  override def isEmpty(): Boolean =
+    _size == 0
+
+  override def size(): Int =
+    _size.toInt
+
+  private def getNodeAt(index: Int): Node[E] = {
+    var current: Node[E] = head
+    for (_ <- 0 until index)
+      current = current.next
+    current
+  }
+
+  private def removeNode(node: Node[E]): Unit = {
+    if (node eq head) {
+      poll()
+    } else if (head ne null) {
+      var prev = head
+      var current: Node[E] = head.next
+
+      while ((current ne null) && (current ne node)) {
+        prev = current
+        current = current.next
+      }
+
+      if (current eq null) {
+        null.asInstanceOf[E]
+      } else {
+        _size -= 1
+
+        prev.next = current.next
+        if (current eq last)
+          last = prev
+      }
+    }
+  }
+
+  override def iterator(): Iterator[E] = {
+    new Iterator[E] {
+
+      private var nextNode: Node[Node[E]] = {
+        val originalHead: Node[Node[E]] =
+          if (head ne null) new Node(head)
+          else null
+
+        var current = originalHead
+        while (current ne null) {
+          val newNode: Node[Node[E]] =
+            if (current.value.next ne null) new Node(current.value.next)
+            else null
+
+          current.next = newNode
+          current = newNode
+        }
+
+        originalHead
+      }
+
+      private var lastNode: Node[Node[E]] = null
+
+      def hasNext(): Boolean =
+        nextNode ne null
+
+      def next(): E = {
+        if (nextNode eq null)
+          throw new NoSuchElementException()
+
+        lastNode = nextNode
+        nextNode = nextNode.next
+
+        lastNode.value.value
+      }
+
+      def remove(): Unit = {
+        if (lastNode eq null)
+          throw new IllegalStateException()
+
+        removeNode(lastNode.value)
+
+        lastNode = null
+      }
+    }
+  }
+
+}
+
+object ConcurrentLinkedQueue {
+
+  private final class Node[T](
+      var value: T,
+      var next: Node[T] = null)
+
+}
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
@@ -1,0 +1,184 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package  org.scalajs.testsuite.javalib
+
+import scala.language.implicitConversions
+
+import scala.collection.JavaConversions._
+
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.util.concurrent.ConcurrentLinkedQueue
+
+object ConcurrentLinkedQueueTest extends ConcurrentLinkedQueueTest(new ConcurrentLinkedQueueFactory)
+
+abstract class ConcurrentLinkedQueueTest[F <: ConcurrentLinkedQueueFactory](
+    listFactory: F) extends AbstractCollectionTest(listFactory) {
+
+  override def testApi(): Unit = {
+
+    super.testApi()
+
+    it("should store and remove ordered integers") {
+      val pq = new ConcurrentLinkedQueue[Int]()
+
+      expect(pq.size()).toEqual(0)
+      expect(pq.add(111)).toBeTruthy
+      expect(pq.size()).toEqual(1)
+      expect(pq.add(222)).toBeTruthy
+      expect(pq.size()).toEqual(2)
+      expect(pq.poll()).toEqual(111)
+      expect(pq.size()).toEqual(1)
+      expect(pq.poll()).toEqual(222)
+      expect(pq.add(222)).toBeTruthy
+      expect(pq.add(222)).toBeTruthy
+      expect(pq.remove(222)).toBeTruthy
+      expect(pq.remove(222)).toBeTruthy
+      expect(pq.remove(222)).toBeFalsy
+    }
+
+    it("should store and remove strings") {
+      val pq = new ConcurrentLinkedQueue[String]()
+
+      expect(pq.size()).toEqual(0)
+      expect(pq.add("aaa")).toBeTruthy
+      expect(pq.size()).toEqual(1)
+      expect(pq.add("bbb")).toBeTruthy
+      expect(pq.size()).toEqual(2)
+      expect(pq.poll()).toEqual("aaa")
+      expect(pq.size()).toEqual(1)
+      expect(pq.poll()).toEqual("bbb")
+      expect(pq.add("bbb")).toBeTruthy
+      expect(pq.add("bbb")).toBeTruthy
+      expect(pq.remove("bbb")).toBeTruthy
+      expect(pq.remove("bbb")).toBeTruthy
+      expect(pq.remove("bbb")).toBeFalsy
+      expect(pq.poll()).toBeNull
+    }
+
+    it("should store Double even in corner cases") {
+      val pq = new ConcurrentLinkedQueue[Double]()
+
+      expect(pq.add(1.0)).toBeTruthy
+      expect(pq.add(+0.0)).toBeTruthy
+      expect(pq.add(-0.0)).toBeTruthy
+      expect(pq.add(Double.NaN)).toBeTruthy
+
+      expect(pq.poll.equals(1.0)).toBeTruthy
+
+      expect(pq.poll.equals(+0.0)).toBeTruthy
+
+      expect(pq.poll.equals(-0.0)).toBeTruthy
+
+      expect(pq.peek.isNaN).toBeTruthy
+
+      expect(pq.remove(Double.NaN)).toBeTruthy
+
+      expect(pq.isEmpty).toBeTruthy
+    }
+
+    it("could be instantiated with a prepopulated Collection") {
+      val l = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val pq = new ConcurrentLinkedQueue[Int](l)
+
+      expect(pq.size()).toEqual(5)
+      for (i <- l) {
+        expect(pq.poll()).toEqual(i)
+      }
+      expect(pq.isEmpty).toBeTruthy
+    }
+
+    it("should be cleared in a single operation") {
+      val l = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val pq = new ConcurrentLinkedQueue[Int](l)
+
+      expect(pq.size()).toEqual(5)
+      pq.clear()
+      expect(pq.size()).toEqual(0)
+    }
+
+    it("should add multiple elemnt in one operation") {
+      val l = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val pq = new ConcurrentLinkedQueue[Int]()
+
+      expect(pq.size()).toEqual(0)
+      pq.addAll(l)
+      expect(pq.size()).toEqual(5)
+      pq.add(6)
+      expect(pq.size()).toEqual(6)
+    }
+
+    it("should check contained values even in double corner cases") {
+      val pq = new ConcurrentLinkedQueue[Double]()
+
+      expect(pq.add(11111.0)).toBeTruthy
+      expect(pq.size()).toEqual(1)
+      expect(pq.contains(11111.0)).toBeTruthy
+      expect(pq.iterator.next()).toEqual(11111.0)
+
+      expect(pq.add(Double.NaN)).toBeTruthy
+      expect(pq.size()).toEqual(2)
+      expect(pq.contains(Double.NaN)).toBeTruthy
+      expect(pq.contains(+0.0)).toBeFalsy
+      expect(pq.contains(-0.0)).toBeFalsy
+
+      expect(pq.remove(Double.NaN)).toBeTruthy
+      expect(pq.add(+0.0)).toBeTruthy
+      expect(pq.size()).toEqual(2)
+      expect(pq.contains(Double.NaN)).toBeFalsy
+      expect(pq.contains(+0.0)).toBeTruthy
+      expect(pq.contains(-0.0)).toBeFalsy
+
+      expect(pq.remove(+0.0)).toBeTruthy
+      expect(pq.add(-0.0)).toBeTruthy
+      expect(pq.size()).toEqual(2)
+      expect(pq.contains(Double.NaN)).toBeFalsy
+      expect(pq.contains(+0.0)).toBeFalsy
+      expect(pq.contains(-0.0)).toBeTruthy
+
+      expect(pq.add(+0.0)).toBeTruthy
+      expect(pq.add(Double.NaN)).toBeTruthy
+      expect(pq.contains(Double.NaN)).toBeTruthy
+      expect(pq.contains(+0.0)).toBeTruthy
+      expect(pq.contains(-0.0)).toBeTruthy
+    }
+
+    it("should provide a weakly consistent iterator") {
+      val queue = new ConcurrentLinkedQueue[Int]()
+      queue.add(1)
+      queue.add(2)
+      val iter1 = queue.iterator()
+      expect(iter1.next()).toEqual(1)
+      expect(iter1.hasNext).toBeTruthy
+      queue.remove(2)
+      expect(iter1.hasNext).toBeTruthy
+      expect(iter1.next()).toEqual(2)
+      expect(iter1.hasNext).toBeFalsy
+
+      val queue2 = new ConcurrentLinkedQueue[Int]()
+      queue2.add(1)
+      queue2.add(2)
+      queue2.add(3)
+      val iter2 = queue2.iterator()
+      expect(iter2.next()).toEqual(1)
+      iter2.remove()
+      expect(iter2.next()).toEqual(2)
+      expect(iter2.next()).toEqual(3)
+    }
+  }
+}
+
+class ConcurrentLinkedQueueFactory extends AbstractCollectionFactory {
+  override def implementationName: String =
+    "java.util.concurrent.ConcurrentLinkedQueue"
+
+  override def empty[E]: ConcurrentLinkedQueue[E] =
+    new ConcurrentLinkedQueue[E]()
+}
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/LinkedListTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/LinkedListTest.scala
@@ -35,6 +35,8 @@ abstract class LinkedListTest[F <: LinkedListFactory](listFactory: F) extends Ab
       ll.addLast(2)
       expect(ll.peekFirst()).toEqual(2)
 
+      ll.clear()
+
       ll.addFirst(1)
       ll.removeLast()
       ll.addFirst(2)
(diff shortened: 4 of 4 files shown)


javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,45 @@
...
+
+class ConcurrentLinkedQueue[E]()
+    extends AbstractQueue[E] with Queue[E] with Serializable {
+
+  def this(c: Collection[_ <: E]) = {
+    this()
+    addAll(c)
+  }
+
+  private var inner: LinkedList[E] = new LinkedList()
```
nicolasstucki: Should be a `val`

andreaTP: sure!

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,45 @@
...
+    inner.poll()
+
+  override def peek(): E =
+    inner.peek()
+
+  override def isEmpty(): Boolean =
+    inner.isEmpty()
+
+  override def size(): Int =
+    inner.size()
```
nicolasstucki: The contract of `ConcurrentLinkedQueue` says that if the queue has more than `Int.MaxValue` it has to return `Int.MaxValue`. The implementation in `LinkedList` is not currently doing that, but it should be because the `List` interface also defines it as such. 

Look at the case where your linked list has `Int.MAX_VALUE` and you add an element. You will do at at some point a `_size += 1` which will make the size equal to `Int.MIN_VALUE` :(

andreaTP: I will double check but the method signature returns an ```int ``` in java, so probably also the size in  java will overflow, I will check that this evening

nicolasstucki: I've added it as a bug on `LinkedList` in #1777.

nicolasstucki: I've added a proposed way to fix this in #1777.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
```diff
@@ -0,0 +1,148 @@
...
+import scala.collection.JavaConversions._
+
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.util.concurrent.ConcurrentLinkedQueue
+
+object ConcurrentLinkedQueueTest extends JasmineTest {
+
```
nicolasstucki: You should extend the `AbstractCollectionTest` class or even better if you create an `AbstractQueueTest` that extends `AbstractCollectionTest` and extend this test to that one. 

If you do that you may need to define a `def allowsNullElements: Boolean` in `CollectionFactory` and special case some of the tests.

nicolasstucki:
That is all.

javalib/src/main/scala/java/util/LinkedList.scala
```diff
@@ -214,13 +214,15 @@ class LinkedList[E]() extends AbstractSequentialList[E]
...
 
   def pollFirst(): E =
-    removeFirst()
+    if (isEmpty()) null.asInstanceOf[E]
+    else removeFirst()
 
   def pollLast(): E =
-    removeLast()
+    if (isEmpty) null.asInstanceOf[E]
+    else removeLast()
```
sjrd: These fixes--and their tests--should be in a separate first commit.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,45 @@
...
+    inner.contains(o)
+
+  override def remove(o: Any): Boolean =
+    inner.remove(o)
+
+  override def addAll(c: Collection[_ <: E]): Boolean =
+    inner.addAll(c)
+
+  override def iterator(): Iterator[E] =
+    inner.iterator()
```
sjrd: Why does this satisfy the weakly consistent property of `ConcurrentLinkedQueue`?

> Iterators are weakly consistent, returning elements reflecting the state of the queue at some point at or since the creation of the iterator. They do not throw ConcurrentModificationException, and may proceed concurrently with other operations. Elements contained in the queue since the creation of the iterator will be returned exactly once.

For example, how does it address the following test:
```scala
val queue = new ConcurrentLinkedQueue[Int]
queue.add(1)
queue.add(2)
val iter = queue.iterator()
expect(iter.next()).toEqual(1)
expect(iter.hasNext).toBeTruthy
queue.remove(2)
expect(iter.hasNext).toBeTruthy // yes, must still be truthy
expect(iter.next()).toEqual(2)
expect(iter.hasNext).toBeFalsy
```
Another one:
```scala
val queue = new ConcurrentLinkedQueue[Int]
queue.add(1)
queue.add(2)
queue.add(3)
val iter = queue.iterator()
expect(iter.next()).toEqual(1)
iter.remove(1)
iter.remove(3)
/* Since, we observed 1 before, and all the states of the queue since the creation of the
 * iterator that had 1 also had 2 and 3. So the iterator must also show 2 and 3.
 */
expect(iter.next()).toEqual(2)
expect(iter.next()).toEqual(3)
```
It is also unclear to me what should happen about calls to `remove()` on the iterator if they happen "concurrently" with modifications of the underlying queue. The Java spec doesn't seem to address this case, yet says that `remove()` is a supported operation.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: used Double as proposed

javalib/src/main/scala/java/util/LinkedList.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok will work on this tomorrow

andreaTP: I just figure out that the Iterator on ConcurrentLinkedQueue does not guarantee as said:
```guarantees to traverse elements as they existed upon construction of the iterator```
in fact:
```
scala> val queue= new ConcurrentLinkedQueue[Int]()
queue: java.util.concurrent.ConcurrentLinkedQueue[Int] = []

scala> queue.add(1)
res6: Boolean = true

scala> queue.add(2)
res7: Boolean = true

scala> queue.add(3)
res8: Boolean = true

scala> val iter = queue.iterator
iter: java.util.Iterator[Int] = java.util.concurrent.ConcurrentLinkedQueue$Itr@c10d272

scala> queue.clear

scala> iter.next
res10: Int = 1

scala> iter.remove

scala> iter.next
java.util.NoSuchElementException
  at java.util.concurrent.ConcurrentLinkedQueue$Itr.next(ConcurrentLinkedQueue.java:732)
  ... 43 elided
```

looks like that a "next element caching" can exactly reproduce the behaviour (and also satisfy your tests...)

**Comment by andreaTP** [comment]
this satisfy requirements, or it looks like, but is not nice at all...

**Comment by sjrd** [comment]
It looks more complicated than I thought. By experimentation with `q.remove()` and `iter.remove()`, I infer the following spec for `iter.remove()` in the presence of concurrent modifications:

* `iterator()` returns an iterator that will return all elements as they were in the Queue at *some* point since the creation of the iterator and now. Note that a stronger spec (and hence correct implem) is to always return all elements that were present in the Queue exactly when the iterator was created.
* `iter.remove()` will remove the *node* in the original Queue, if it is still present. If the node for that element has been removed, but there is another node with the same element, that other node is *not* removed. That means that calling `enclosingQueue.remove(lastElemReturnedByNext)` is not a correct implementation.

Some test cases:
```scala
val q = new ConcurrentLinkedQueue[Int]
q.add(1)
q.add(2)
q.add(3)
q.add(2)
val i = q.iterator()
expect(i.next()).toEqual(1)
expect(i.next()).toEqual(2) // observe the first 2
q.remove(2) // remove the first 2 through the queue
i.remove() // remove the same first 2, no-op (the second 2 is still there)
expect(q.size).toEqual(3) // [1, 3, 2] at this point
expect(i.next()).toEqual(3)
expect(i.next()).toEqual(2)
expect(i.hasNext).toBeFalsy
```

If we invert the two `remove` operations hereabove, then the two 2's are removed:
```scala
val q = new ConcurrentLinkedQueue[Int]
q.add(1)
q.add(2)
q.add(3)
q.add(2)
val i = q.iterator()
expect(i.next()).toEqual(1)
expect(i.next()).toEqual(2) // observe the first 2
i.remove() // remove the first 2 which we just iterated
q.remove(2) // remove now the *second* 2 through the queue
expect(q.size).toEqual(2) // [1, 3] at this point
expect(i.next()).toEqual(3)
/* Since we iterated 2 before, we must also iterate 2 now,
 * because there was no state of the queue where the first 2
 * was there but not the second.
 */
expect(i.next()).toEqual(2)
expect(i.hasNext).toBeFalsy
```

I think that the simplest way to implement this specification is to wrap all elements in "identity boxes":
```scala
private class ID[+E](val elem: E)
```
(*not* a case class, we want identity equality) in the underlying `LinkedList` so that we can distinguish "different 2's" from each other. With this, the iterator is pretty trivial to implement in terms of the iterator of a clone of the underlying linked list. The `q.remove(e)` operation, however, needs to be reimplemented as a loop on top of the underlying list's iterator, because calling `innerList.remove(new ID(elem))` will not work (since it creates a new `ID` which is therefore different than all the ones in the list).

The "proper" way to implement all of this would be to have a full-fledged implementation based on a singly-linked list of `Node`s, obviously. But this requires a lot of work, again, so the `ID` approach seems better to me at this point.

sjrd:
Ah ah, I think I found a bug in the JDK. My second "test case" does not work on the JDK, in a way that breaks the weakly consistent property of the iterator:
```scala
scala> val q = new ConcurrentLinkedQueue[Int]
q: java.util.concurrent.ConcurrentLinkedQueue[Int] = []

scala> q.add(1)
res62: Boolean = true

scala> q.add(2)
res63: Boolean = true

scala> q.add(3)
res64: Boolean = true

scala> q.add(2)
res65: Boolean = true

scala> val i = q.iterator()
i: java.util.Iterator[Int] = java.util.concurrent.ConcurrentLinkedQueue$Itr@478bc78e

scala> expect(i.next()).toEqual(1)
warning: there were 1 feature warnings; re-run with -feature for details

scala> expect(i.next()).toEqual(2)
warning: there were 1 feature warnings; re-run with -feature for details

scala> q
res68: java.util.concurrent.ConcurrentLinkedQueue[Int] = [1, 2, 3, 2]

scala> i.remove()

scala> q
res70: java.util.concurrent.ConcurrentLinkedQueue[Int] = [1, 3, 2]

scala> q.remove(2)
res71: Boolean = true

scala> q
res72: java.util.concurrent.ConcurrentLinkedQueue[Int] = [1, 3]

scala> i.next()
res73: Int = 3

scala> i.next()
java.util.NoSuchElementException
        at java.util.concurrent.ConcurrentLinkedQueue$Itr.next(ConcurrentLinkedQueue.java:614)
        at .<init>(<console>:11)
        at .<clinit>(<console>)
        at .<init>(<console>:7)
        at .<clinit>(<console>)
        at $print(<console>)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
        at java.lang.reflect.Method.invoke(Method.java:597)
        at scala.tools.nsc.interpreter.IMain$ReadEvalPrint.call(IMain.scala:731)
        at scala.tools.nsc.interpreter.IMain$Request.loadAndRun(IMain.scala:980)
        at scala.tools.nsc.interpreter.IMain.loadAndRunReq$1(IMain.scala:570)
        at scala.tools.nsc.interpreter.IMain.interpret(IMain.scala:601)
        at scala.tools.nsc.interpreter.IMain.interpret(IMain.scala:565)
        at scala.tools.nsc.interpreter.ILoop.reallyInterpret$1(ILoop.scala:745)
        at scala.tools.nsc.interpreter.ILoop.interpretStartingWith(ILoop.scala:790)
        at scala.tools.nsc.interpreter.ILoop.command(ILoop.scala:702)
        at scala.tools.nsc.interpreter.ILoop.processLine$1(ILoop.scala:566)
        at scala.tools.nsc.interpreter.ILoop.innerLoop$1(ILoop.scala:573)
        at scala.tools.nsc.interpreter.ILoop.loop(ILoop.scala:576)
        at scala.tools.nsc.interpreter.ILoop$$anonfun$process$1.apply$mcZ$sp(ILoop.scala:867)
        at scala.tools.nsc.interpreter.ILoop$$anonfun$process$1.apply(ILoop.scala:822)
        at scala.tools.nsc.interpreter.ILoop$$anonfun$process$1.apply(ILoop.scala:822)
        at scala.tools.nsc.util.ScalaClassLoader$.savingContextLoader(ScalaClassLoader.scala:135)
        at scala.tools.nsc.interpreter.ILoop.process(ILoop.scala:822)
        at scala.tools.nsc.MainGenericRunner.runTarget$1(MainGenericRunner.scala:83)
        at scala.tools.nsc.MainGenericRunner.process(MainGenericRunner.scala:96)
        at scala.tools.nsc.MainGenericRunner$.main(MainGenericRunner.scala:105)
        at scala.tools.nsc.MainGenericRunner.main(MainGenericRunner.scala)
```
The iterator has enumerated `1 2 3`, but that list was never a state of the queue ^^

We *should not* follow this bug.

andreaTP:
Could we try to be lazy on this and pedantically follow the description:
```
 guarantees to traverse elements as they existed upon construction of the iterator
```
so an iterator on a copy of the original list will satisfy the sentence.
```
and may (but is not guaranteed to) reflect any modifications subsequent to construction.
```
we can avoid reflecting modifications and so respect the fact that this behaviour is not guaranteed.

About removals I cannot find in any place of documentation that this specific iterator MUST have the remove operation, so we can throw ```UnsupportedOperationException``` and stay on the safe side.

I know that this is not so fair, but, at least I prefer to provide a limited implementation still correct instead of a full implementation that is going to be bugged or overcomplicated.

sjrd:
Could we try to be lazy on this and pedantically follow the description:

> guarantees to traverse elements as they existed upon construction of the iterator

Yes, as I said above, this a perfectly valid implementation. Since the "may (but is not guaranteed to)" is correctly implemented as the more specific "will never".
My suggestion based on a `LinkedList[ID[E]]` would do that.

> About removals I cannot find in any place of documentation that this specific iterator MUST have the remove operation

I wish this were true, but it is not. The top documentation of `ConcurrentLinkedQueue` says this:

> This class and its iterator implement all of the *optional* methods of the `Queue` and `Iterator` interfaces.

In particular, `Iterator.remove()`. So throwing `UnsupportedOperationException` in the iterator's `remove()` method is *not* a correct implementation.

andreaTP:
at the end it is not so much work a proper single linked list plain implementation.
hope you will appreciate this!

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+  import ConcurrentLinkedQueue._
+
+  private var head: Node[E] = null
+  private var last: Node[E] = null
+
+  private var _size: Double = 0
+
+  override def add(e: E): Boolean =
+    if (e == null) throw new NullPointerException()
+    else {
```
sjrd: According to https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#ifelse, the then and else parts should have the same "look", hence:
```scala
if (e == null) {
  throw new NullPointerException()
} else {
```
Also, according to https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#braces, The entire body of `add()` should have `{}`.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  override def isEmpty(): Boolean =
+    _size == 0
+
+  override def size(): Int =
+    _size.toInt
+
+  private def getNodeAt(index: Int): Node[E] = {
+    if (index == 0) head
+    else {
```
sjrd: The if branch is redundant. This is covered by the code in the else branch.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+      var current: Node[E] = null
+      current = head
```
sjrd: These two lines are
```scala
var current: Node[E] = head
```

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+      for (_ <- 0 until index)
+        current = current.next
+      current
+    }
+  }
+
+  private def removeNode(node: Node[E]): E = {
```
sjrd: This method should return `Unit`. Its result is never used.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+    if (node eq head) poll()
+    else {
```
sjrd: Same:
```scala
if (node eq head) {
  poll()
} else {
```
Also, the else branch will crash if the list is empty. It can happen with the following test case:
```scala
val q = new ConcurrentLinkedQueue
q.add(1)
val iter = q.iterator()
iter.next()
q.poll()
iter.remove() // must not throw
```
So this else should be:
```scala
} else if (head ne null) {
```
(if `head eq null`, there is nothing to do)

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+      var prev = head
+      var current: Node[E] = head.next
+
+      while ((current ne null) && (current ne node)) {
+        prev = current
+        current = current.next
+      }
+
+      if (current eq null) null.asInstanceOf[E]
+      else {
```
sjrd: Same if formatting.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+        _size -= 1
+
+        prev.next = current.next
+        if (current eq last)
+          last = prev
+
+        current.value
```
sjrd: This last line should be removed once the method returns `Unit`.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+          if (current.value.next ne null) {
+            new Node(current.value.next)
+          } else null
+
+        current.next = newOne
+        current = newOne
+      }
+
+      private var currentNode: Node[Node[E]] =
+        originalHead
```
sjrd: This way of writing things will create *fields* `originalHead` and `current` in the iterator object. We don't want that. Write it like this to avoid this issue:
```scala
private var currentNode: Node[Node[E]] = {
  val originalHead = ...
  ...
  originalHead
}
```

sjrd: Unless there is precedent in `LinkedList` (I don't remember), it might be better to call this variable `nextNode` instead of `currentNode`, since it is the node whose value will be returned by a call to `next()`.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+
+      private var lastNode: Node[Node[E]] = null
+
+      def hasNext(): Boolean =
+        (currentNode ne null)
```
sjrd: `()` are not needed.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+
+        lastNode.value.value
+      }
+
+      def remove(): Unit = {
+        if (lastNode eq null)
+          throw new IllegalStateException()
+
+        removeNode(lastNode.value)
+      }
```
sjrd: You forgot
```scala
lastNode = null
```
so that a second call to `remove()` throws the `IllegalStateException`.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+
+}
+
+object ConcurrentLinkedQueue {
+
+  protected[ConcurrentLinkedQueue] final class Node[T](
```
sjrd: Simply `private` will do the trick.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  override def iterator(): Iterator[E] = {
+    new Iterator[E] {
+
+      private var originalHead: Node[Node[E]] =
+        if (head ne null) new Node(head)
+        else null
+
+      var current = originalHead
+      while(current ne null) {
```
sjrd: Missing space after `while`

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+        val newOne: Node[Node[E]] =
+          if (current.value.next ne null) {
+            new Node(current.value.next)
```
sjrd: Put this on the same line as `if` to make this if/else a two-liner if/else.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
```diff
@@ -0,0 +1,153 @@
...
+  override def iterator(): Iterator[E] = {
+    new Iterator[E] {
+
+      private var originalHead: Node[Node[E]] =
+        if (head ne null) new Node(head)
+        else null
+
+      var current = originalHead
+      while(current ne null) {
+        val newOne: Node[Node[E]] =
```
sjrd: `newNode`

test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
```diff
@@ -0,0 +1,174 @@
...
+      queue.add(2)
+      val iter1 = queue.iterator()
+      expect(iter1.next()).toEqual(1)
+      expect(iter1.hasNext).toBeTruthy
+      queue.remove(2)
+      expect(iter1.hasNext).toBeTruthy // yes, must still be truthy
+      expect(iter1.next()).toEqual(2)
+      expect(iter1.hasNext).toBeFalsy
+
+      queue.clear()
```
sjrd: Create a new `val queue2` here, to make absolutely certain the two tests are independent.

sjrd:
That's all.

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: agree

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok and ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: I like them, but ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: correct

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
andreaTP: ok

andreaTP:
rebased, still need an ok on the other PR, try to include all these in 0.6.4

test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
sjrd: You forgot this comment.

andreaTP:
ok this is directly mergeable just after the other pending PR

sjrd:
If you want this to have a chance to make it into 0.6.4:

1. Address @nicolasstucki's comment.
2. Rebase *on top of 80ee9f4*, so that this PR actually contains exactly 2 commits. First 80ee9f4 then the commit "Port java.util.concurrent.ConcurrentLinkedQueue."

If this works (tests pass), then this PR will supersede #1784 and I'll merge this one. Otherwise I'll merge #1784 but I won't have the time to include a PR *after* that. (and if I delay, there won't be a release at all for two weeks).

andreaTP:
I'm doing it

andreaTP:
hope to be on time

andreaTP:
will be ok now!

andreaTP:
p.s.
sorry for stressing you @sjrd 

p.p.s.
good holidays after that ^^

test-suite/src/test/scala/org/scalajs/testsuite/javalib/ConcurrentLinkedQueueTest.scala
```diff
@@ -0,0 +1,183 @@
...
+
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.util.concurrent.ConcurrentLinkedQueue
+
+object ConcurrentLinkedQueueTest extends ConcurrentLinkedQueueTest(new ConcurrentLinkedQueueFactory)
+
+abstract class ConcurrentLinkedQueueTest[F <: ConcurrentLinkedQueueFactory](listFactory: F) extends AbstractCollectionTest(listFactory) {
```
sjrd: This line is too long. It won't pass scalastyle. Break the line before `listFactory`, with 4 spaces.

sjrd:
That's all. If the line were not > 120 chars, I would still LGTM and run the tests. But here I know the tests won't pass, so you'll have to fix that.

sjrd:
https://github.com/scala-js/scala-js/pull/1784#issuecomment-117839781 also applies here, of course (need to fix the first commit for that, not the second one)

sjrd:
The easiest way to fix the first commit is, after you've fixed the other PR, re-rebase this on top of the other branch.

andreaTP:
will be ok now

sjrd:
Needs to be rebased on top of 949fdbf

sjrd:
(as it stands, the first commit of the two here does not pass the tests)

andreaTP:
rebased

sjrd:

andreaTP:
ok I go to sleep now,
will se the results tomorrow

sjrd:
One spurious failure. It'll take the chance and merge this anyway now.

PR #1782, opened 2015-07-01
Fix internal compiler bugs

sjrd:


sjrd:
Review by @nicolasstucki (possibly also @gzm0).

ir/src/main/scala/org/scalajs/core/ir/Tags.scala
```diff
@@ -67,7 +67,7 @@ private[ir] object Tags {
   final val TagJSEnvInfo = TagJSObjectConstr + 1
 
   final val TagUndefined = TagJSEnvInfo + 1
-  final val TagUndefinedParam = TagUndefined + 1
+  final val TagUndefinedParam = TagUndefined + 1 // TODO Move this
```
nicolasstucki: Is this TODO intentional?

sjrd: Yes. It can be fixed only when we can break binary compatibility, like the other TODOs in this file.

gzm0:
LGTM

PR #1789, opened 2015-07-05
Fix Scala version in referencedCrossProject.

sjrd:


sjrd:
Review by @nicolasstucki 
This PR should be merged before anything else gets in, since it fixes the build, which currently does not pass.

nicolasstucki:
LGTM

PR #1793, opened 2015-07-13
Fix #1791: Provide conversions from Java boxed types to js.Any.

### LINKED ISSUE #1791: Nullable primitives from java.lang should be convertible to js.Any
Since types from `java.lang` can be used in JS interop and have the same representation as their corresponding primitives, I think there should be implicit conversions from them to `js.Any`

i.e.
```
import java.{lang => jl}

@inline implicit def fromJBoolean(value: jl.Boolean): js.Any =
  value.asInstanceOf[js.Any]
@inline implicit def fromJByte(value: jl.Byte): js.Any =
  value.asInstanceOf[js.Any]
@inline implicit def fromJShort(value: jl.Short): js.Any =
  value.asInstanceOf[js.Any]
@inline implicit def fromJInteger(value: jl.Integer): js.Any =
  value.asInstanceOf[js.Any]
@inline implicit def fromJLong(value: jl.Long): js.Any =
  value.toDouble.asInstanceOf[js.Any]
@inline implicit def fromJFloat(value: jl.Float): js.Any =
  value.asInstanceOf[js.Any]
@inline implicit def fromJDouble(value: jl.Double): js.Any =
  value.asInstanceOf[js.Any]
```

sjrd:
Java boxed types can be used as an accurate representation of `null | PrimType`. These conversions allow these types to be used where a js.Any is expected.


sjrd:
Review by @nicolasstucki 
/cc @ghik

library/src/main/scala/scala/scalajs/js/Any.scala
```diff
@@ -126,6 +126,24 @@ object Any extends LowPrioAnyImplicits {
...
+
+  @inline implicit def fromJLong(value: java.lang.Long): Any =
+    if (value eq null) null
+    else value.doubleValue.asInstanceOf[Any]
+
+  @inline implicit def fromJFloat(value: java.lang.Float): Any =
+    value.asInstanceOf[Any]
+  @inline implicit def fromJDouble(value: java.lang.Double): Any =
+    value.asInstanceOf[Any]
+
```
nicolasstucki: Doesn't this also apply to `java.lang.Character`?

sjrd: No, we don't have the conversion `Char` to `js.Any`, so we shouldn't have `java.lang.Character` to `js.Any` either.

nicolasstucki:
Thats all. Otherwise LGTM

sjrd:
Updated to fix the binary compatibility problems: moved the methods from `LowPrioAnyImplicits` to `Any` itself.

nicolasstucki:
LGTM.
Is this fix something that must be changed back on next major release. If so we should include this in #1796.

sjrd:
Not necessarily. I initially thought I needed to put that in LowPrio, but it turns out they're fine in `js.Any`, so they should stay there.

nicolasstucki:
Ok

PR #1794, opened 2015-07-13
Fix #1790: Don't crash for a lambda param containing a dash.

### LINKED ISSUE #1790: Compiler crash with a dash in a parameter of a lambda
I have run into a curious bug, which I have attempted to minimize in [this github project](https://github.com/mrerrormessage/dashy-ident-bug). It occurs when using macros with a dashy identifier. The macros work fine when compiled with stock scala, but they error when compiled with scala-js. The exception begins:
```
java.lang.IllegalArgumentException: requirement failed: a-b is not a valid identifier
        at scala.Predef$.require(Predef.scala:219)
        at org.scalajs.core.ir.Trees$.requireValidIdent(Trees.scala:61)
        at org.scalajs.core.ir.Trees$Ident.<init>(Trees.scala:45)
        at org.scalajs.core.compiler.JSEncoding$class.freshLocalIdent(JSEncoding.scala:71)
        at org.scalajs.core.compiler.GenJSCode.freshLocalIdent(GenJSCode.scala:31)
        at org.scalajs.core.compiler.GenJSCode$JSCodePhase$$anonfun$78.apply(GenJSCode.scala:3864)
        at org.scalajs.core.compiler.GenJSCode$JSCodePhase$$anonfun$78.apply(GenJSCode.scala:3858)
        at scala.collection.TraversableLike$WithFilter$$anonfun$map$2.apply(TraversableLike.scala:728)
        at scala.collection.immutable.List.foreach(List.scala:381)
        at scala.collection.TraversableLike$WithFilter.map(TraversableLike.scala:727)
```

I would be happy to contribute if there's anything I can do to help fix the issue, but I'm not sure where to start.

sjrd:
This boils down to not using the original name of a parameter to derive the identifier for its boxed parameter. Use its identifier instead.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -3861,7 +3861,7 @@ abstract class GenJSCode extends plugins.PluginComponent
         val paramName = param.name
         val js.Ident(name, origName) = paramName
         val newOrigName = origName.getOrElse(name)
-        val newNameIdent = freshLocalIdent(newOrigName)(paramName.pos)
+        val newNameIdent = freshLocalIdent(name)(paramName.pos)
         val patchedParam = js.ParamDef(newNameIdent, jstpe.AnyType,
             mutable = false, rest = param.rest)(param.pos)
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #1800, opened 2015-07-15
Fix #1596: Safe casts from Scala class to JS trait

### LINKED ISSUE #1596: Macro to safely cast from Scala class to JS trait
Write a macro to do the following with static checking:

```scala
trait Foo extends js.Object {
  def foo(x: Int): Int = js.Native
}

class Bar {
  @JSExport
  def foo(x: Int): Int = x * 2
}

val x: Bar = ???

// Macro should check this:
val y: Foo = x.asInstanceOf[Foo]
```

gzm0:


gzm0:
Review by @sjrd

sjrd:
Oh ow ^^ Lots of lines to review ...

library/src/main/scala/scala/scalajs/js/package.scala
```diff
@@ -105,4 +109,75 @@ package object js {
       "type of a Scala.js library has been called. This is most likely " +
       "because you tried to run Scala.js binaries on the JVM. Make sure you " +
       "are using the JVM version of the libraries.")
+
+  /** Allows to type safely cast an instance of a Scala class to facade trait
```
sjrd: Missing `.` at the end of the line.

"type safely cast" is a bit weird. I suggest "to cast an instance ... in a type-safe way."

library/src/main/scala/scala/scalajs/js/package.scala
```diff
@@ -105,4 +109,75 @@ package object js {
...
+   *  js.use(x1).as[MyFacade] // OK
+   *  }}}
+   *
+   *  {{{
+   *  class MyClass2 {
+   *    @JSExport
+   *    def foo(x: Int): String = x.toString
+   *
+   *    @JSExport
+   *    val bar: Int = 1
```
sjrd: Worth noting that the `def`/`val` is the difference between the two examples. I had to search for the difference.

library/src/main/scala/scala/scalajs/js/package.scala
```diff
@@ -105,4 +109,75 @@ package object js {
...
+   *  {{{
+   *  class MyClass3 {
+   *    @JSExport
+   *    def foo(x: String): String = x.toString
+   *  }
+   *
+   *  val x3 = new MyClass3
+   *  js.use(x2).as[MyFacade] // Fails: bar is missing and foo has wrong type
+   *  }}}
+   *
```
sjrd: Also worth writing an example that fails because (one of) the methods is not exported.

library/src/main/scala/scala/scalajs/js/package.scala
```diff
@@ -105,4 +109,75 @@ package object js {
...
+   *  == Restrictions ==
+   *  - Facade types may only be traits and not have any class ancestors
```
sjrd: Why?

library/src/main/scala/scala/scalajs/macroImpls/Compat210.scala
```diff
@@ -0,0 +1,51 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.macroImpls
```
sjrd: Avoid a package name with an uppercase. Consider `macroimpls` instead.

library/src/main/scala/scala/scalajs/macroImpls/Compat210.scala
```diff
@@ -0,0 +1,51 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
```
sjrd: `2013-2015`
(there are probably other occurrences of this)

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Consider:

```scala
class A extends js.Object
trait B extends A

val a: A = use(x).as[B]
a.isInstanceOf[A] // --> false --> bad
```

sjrd: `x.isInstanceOf[A]` returns false, indeed. But `x.asInstanceOf[A]` does work nevertheless.

library/src/main/scala/scala/scalajs/js/package.scala
```diff
@@ -105,4 +109,75 @@ package object js {
...
+   *  - Polymorphic methods are currently not supported
+   *  - Facade types defining an apply method cannot used (this is a JavaScript
+   *    restriction).
+   */
+  def use[A](x: A): Using[A] = new Using[A](x)
+
+  /** Helper for syntactic sugar of [[use]]. Do not use directly */
+  final class Using[A](val x: A) extends AnyVal {
```
sjrd: Can we hide this better? e.g., can the constructor be `private`?

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+ *  that allows to distinguish setters and getters from methods.
+ *
+ *  It also allows to check method conformance based on Scala.js' JavaScript
+ *  calling conventions.
+ *
+ *  Currently does not support polymorphic method types.
+ *
+ *  @author Tobias Schlatter
+ */
+trait JSMembers {
```
sjrd: This should be `private[macroImpls]`.

library/src/main/scala/scala/scalajs/js/package.scala
sjrd: Irrespective of that, `Using` should be in its own file. Classes inside package objects should be avoided (there are still bugs about that in scalac, AFAIK). Having it its own file would also make it easier to override that file and in the entire `macroImpls` package, should we need a completely different implementation (say scala.meta) in a different compiler (say, dotty ^^).

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSMethod(ps, ret) =>
+        val (used, unused) = params.splitAt(ps.size)
+
+        params.size >= ps.size &&
+        returnType <:< ret &&
+        unused.forall(_.isDefault) &&
+        (used zip ps).forall { case (x,y) => x.conformsTo(y) }
+      case _ =>
```
sjrd: Since the first case contains a blank line, the cases should be separated from each other by a blank line.

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+      returnType: Type) extends JSMember {
+
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSMethod(ps, ret) =>
+        val (used, unused) = params.splitAt(ps.size)
+
+        params.size >= ps.size &&
+        returnType <:< ret &&
+        unused.forall(_.isDefault) &&
+        (used zip ps).forall { case (x,y) => x.conformsTo(y) }
```
sjrd: Missing space after `,`

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    def displayStr(name: String): String =
+      s"method $name(${params.mkString(", ")}): $returnType"
+  }
+
+  case class JSGetter(tpe: Type) extends JSMember {
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSGetter(t) => tpe <:< t
+      case _ => false
```
sjrd: Align the arrows.

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    def displayStr(name: String): String = s"getter $name: $tpe"
+  }
+
+  case class JSSetter(tpe: Type) extends JSMember {
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSSetter(t) => t <:< tpe
+      case _ => false
```
sjrd: Align the arrows.

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+    override def toString(): String =
+      if (isDefault) s"$info = ???"
+      else info.toString
+  }
+
+  case class JSMethod(params: List[JSMethodParam],
+      returnType: Type) extends JSMember {
+
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSMethod(ps, ret) =>
```
sjrd: Consider renaming `ps` to `thatParams` and `ret` to `thatReturnType`.
Also, consider using `resultType` instead of `returnType` (Scala methods have *results*, not *returns*).

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+        false
+    }
+
+    def displayStr(name: String): String =
+      s"method $name(${params.mkString(", ")}): $returnType"
+  }
+
+  case class JSGetter(tpe: Type) extends JSMember {
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSGetter(t) => tpe <:< t
```
sjrd: Consider `thatTpe` instead of `t`.

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
```diff
@@ -0,0 +1,87 @@
...
+      case _ => false
+    }
+
+    def displayStr(name: String): String = s"getter $name: $tpe"
+  }
+
+  case class JSSetter(tpe: Type) extends JSMember {
+    def conformsTo(that: JSMember): Boolean = that match {
+      case JSSetter(t) => t <:< tpe
```
sjrd: Same.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+
+
+package scala.scalajs.macroImpls
+
+import scala.annotation.tailrec
+import scala.scalajs.js
+
+import Compat210._
+
+/** Macros for js.use(x).as[T]
```
sjrd: `.`

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+ *
+ *  This implements a structural typechecker conforming to the JavaScript calling
+ *  convention in Scala.js' export and facade system.
+ *
+ *  @author Tobias Schlatter
+ */
+object UseAsMacros {
```
sjrd: This should be `private[js]`.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+    /** Perform the actual structural typechecking.
+     *
+     *  Checks if [[srcTpe]] conforms to [[trgTpe]]. Reports errors otherwise.
+     */
+    private def check(srcTpe: Type, trgTpe: Type): Unit = {
+      val rawMembers = rawJSMembers(trgTpe)
+
+      /* Nothing exports everything, but we still need to calculate the members
+       * of the raw JS class, since they might be malformed
+       * (i.e. apply or bad refinements)
```
sjrd: If you support `Nothing`, you should also allow `Null`.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+    private val JSExportAllAnnotation = typeOf[js.annotation.JSExportAll].typeSymbol
+
+    /** Base classes that are allowed in a target type.
+     *  These are also the classes whose methods do not need to be provided.
+     */
+    private val JSObjectAncestors = typeOf[js.Object].baseClasses.toSet
+
+    def as[A : WeakTypeTag, B <: js.Any : WeakTypeTag]: Expr[B] = {
+      val trgTpe = verifyTargetType(weakTypeOf[B])
+      val srcTpe = weakTypeOf[A]
```
sjrd: Shouldn't you check that `A` is not a subtype of `js.Any`?

sjrd: Or else, use `exportedMembers(srcTpe)` if `srcTpe` is not a subtype of `js.Any`, but use `rawJSMembers(srcTpe)` if it is. And then you suddenly have a macro that can also cast between a pair of raw JS types with structural type checking (which can be useful in some cases to retroactively "implement" a super-trait, e.g., sometimes core ES classes implement DOM interfaces, because those DOM interfaces are structurally supertypes of those, but that subtyping relationship cannot be expressed with `extends`).

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+          }
+        }
+      }
+    }
+
+    /** Whether the [[candidates]] together conform to [[jsMem]] */
+    private def hasConformingMembers(jsMem: JSMember,
+        candidates: List[JSMember]): Boolean = {
+
+      def find(mem: JSMember) = candidates.exists(_.conformsTo(mem))
```
sjrd: I keep reading `mem` as `memory`, not `member`. ^^

sjrd: Public method should have an explicit result type ... or be private.

sjrd: Ah, never mind, it's local to a method ^^

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+
+      def hasConformingOverloads = jsMem match {
+        case jsMem @ JSMethod(params, _) if params.last.isDefault =>
+          /* If the last parameter is a default parameter, the method can also
+           * be implemented by two methods, one with and one without the
+           * parameter. Note that this does not generalize to more parameters
+           * due to named parameters.
```
sjrd: If it doesn't generalize, maybe we shouldn't support this case at all. More regularity is good.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+        case _ => false
+      }
+
+      find(jsMem) || hasConformingOverloads
+    }
+
+    /** Members that a facade type defines */
+    private def rawJSMembers(tpe: Type): Iterable[(String, JSMember)] = {
+
+      def isValidMember(mem: Symbol) = (
```
sjrd: According to the style guide, `{}` should be used instead of `()` here (and then the body indented 2 spaces, of course).

sjrd: The name `isValidMember` suggests that there are *invalid* members, in the sense that it would be an error if they appeared. Consider something like `isAPIMember` or `isPartOfAPI`.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      for (mem <- tpe.members if isValidMember(mem)) yield {
+        val (name, explicit) = jsName(mem.asMethod)
+
+        // apply is a special case, since when it is not explicitly specified,
+        // it will result in an actual function application in JavaScript
+        // Scala defined types cannot support this --> fail
+        if (name == "apply" && !explicit) {
+          c.abort(c.enclosingPosition, s"$tpe defines an apply method. This " +
+            "cannot be implemented by any Scala exported type, since it " +
+            "would need to chain Function's prototype.")
```
sjrd: 4 spaces.

sjrd: You should also check that there are no `@JSBracketSelect` or `@JSBracketCall` members in the type. Those can similarly not be implemented by any export.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      }
+    }
+
+    /** Returns the JavaScript name of a method and whether it is specified
+     *  explicitly or not.
+     */
+    private def jsName(sym: MethodSymbol): (String, Boolean) = {
+      val annot = memberAnnotations(sym).find(annotIs(_, JSNameAnnotation))
+      val name = annot.flatMap(annotStringArg).getOrElse(defaultName(sym))
+      (name, annot.isDefined)
```
sjrd: I think this is a bad API, even for a private method. The explicitness should be irrelevant when returning from this method. The `apply` case (and the `JSBracketSelect` and `JSBracketCall` cases) should either fail early in here and not leak, or return a value of the following ADT (or something similar):
```scala
sealed abstract class JSMemberSelection
case class JSMemberSelectByName(name: String) extends JSMemberSelection
case object JSMemberCall extends JSMemberSelection
case object JSMemberBracketSelect extends JSMemberSelection
case object JSMemberBracketCall extends JSMemberSelection
```
or its useful contraction as `Option[String]` ^^ where `None` is used for the three last cases.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      val exports = tpe.baseClasses.flatMap(exportedDecls(tpe, _))
+
+      // Group exports by name
+      val lists = for {
+        (name, elems) <- exports.groupBy(_._1)
+      } yield {
+        (name, elems.map(_._2).toList)
+      }
+
+      lists.toMap
```
sjrd: Isn't `lists` already a `Map`?

sjrd: Also consider `exports.groupBy(_._2).mapValues(_.toList).toMap` (in which case `toMap` is necessary to make it eager).

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+        (name, jsMemberFor(origTpe, mem.asMethod))
+      }
+    }
+
+    /** Get the JS member for a method in [[origTpe]] */
+    private def jsMemberFor(origTpe: Type, sym: MethodSymbol): JSMember = {
+      sym.info.asSeenFrom(origTpe, sym.owner) match {
+        case MethodType(List(param), returnType)
+          if returnType.typeSymbol == definitions.UnitClass &&
+             sym.name.decodedName.toString.endsWith("_=") =>
```
sjrd: I guess `sym.name.endsWith(nme.SETTER_SUFFIX)` does not exist in the macro reflect API?

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+              case tpe =>
+                JSMethod(acc.reverse, tpe)
+            }
+          }
+
+          flatParams(info, Nil)
+
+        case PolyType(_, _) =>
+          c.abort(c.enclosingPosition, "Polymorphic methods are currently " +
+            s"not supported. Offending method: ${sym.fullName}")
```
sjrd: 4 spaces.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      } yield {
+        (name, jsMemberFor(origTpe, mem.asMethod))
+      }
+    }
+
+    /** Get the JS member for a method in [[origTpe]] */
+    private def jsMemberFor(origTpe: Type, sym: MethodSymbol): JSMember = {
+      sym.info.asSeenFrom(origTpe, sym.owner) match {
+        case MethodType(List(param), returnType)
+          if returnType.typeSymbol == definitions.UnitClass &&
```
sjrd: `if` should be indented 4 spaces wrt. `case`, since it's a continuation line.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+        case NullaryMethodType(returnType) =>
+          JSGetter(returnType)
+
+        case info: MethodType =>
+          @tailrec
+          def flatParams(tpe: Type, acc: List[JSMethodParam]): JSMethod = {
+            tpe match {
+              case MethodType(params, returnTpe) =>
+                val ps = params map { p =>
+                  JSMethodParam(p.info, p.asTerm.isParamWithDefault)
```
sjrd: What happens to repeated parameters (the `*`s)

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      lazy val default = defaultName(sym)
+
+      val explicitNames = for {
+        annot <- memberAnnotations(sym)
+        if annotIs(annot, JSExportAnnotation)
+      } yield {
+        annotStringArg(annot).getOrElse(default)
+      }
+
+      if (exportAll && sym.isPublic) default :: explicitNames
```
sjrd: What happens to a method named `apply` with a custom `@JSExport("blahblah")` in the presence of `@JSExportAll`?

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+
+      trgSym.annotations
+    }
+
+    /** Retrieve first argument to the annotation as literal string */
+    private def annotStringArg(annot: Annotation): Option[String] = {
+      val args = annot.tree.children.tail
+      args match {
+        case List(Literal(Constant(s: String))) => Some(s)
+        case _ => None
```
sjrd: Align the arrows.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
```diff
@@ -0,0 +1,304 @@
...
+      }
+    }
+
+    /** Checks if [[annot]] is of class [[clsSym]] */
+    private def annotIs(annot: Annotation, clsSym: Symbol) =
+      annot.tree.tpe.typeSymbol == clsSym
+
+  }
+
+
```
sjrd: Double blank line.

project/Build.scala
```diff
@@ -691,7 +691,9 @@ object Build extends sbt.Build {
           scalaJSSourceMapSettings,
           scalacOptions in (Compile, doc) ++= Seq("-implicits", "-groups"),
           exportJars := true,
-          previousArtifactSetting
+          previousArtifactSetting,
+          libraryDependencies +=
+            "org.scala-lang" % "scala-reflect" % scalaVersion.value
```
sjrd: `% "provided"`

test-suite/src/main/scala/org/scalajs/testsuite/TypecheckingMacros.scala
```diff
@@ -0,0 +1,52 @@
...
+        c.typecheck(c.parse(s"{ $code }"))
+        c.abort(c.enclosingPosition,
+            "Expected type error, type checked successfully.")
+      } catch {
+        case e: TypecheckException =>
+          val errMsg = e.getMessage
+          for (msg <- expectedMsg) {
+            if (errMsg != msg) {
+              c.abort(c.enclosingPosition,
+                s"Type error messages mismatch.\nExpected: $msg\nFound: $errMsg")
```
sjrd: 4 spaces.

sjrd:
That's all.

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Yes it does. But shouldn't we include in the definition of "cast in a type-safe way" that we cannot cause `(_: A).isInstanceOf[A] == false`?

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Ok.

library/src/main/scala/scala/scalajs/macroImpls/Compat210.scala
gzm0: Ok.

library/src/main/scala/scala/scalajs/macroImpls/Compat210.scala
gzm0: Ok

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Ok. I improved this section.

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Ok

library/src/main/scala/scala/scalajs/macroImpls/JSMembers.scala
gzm0: Ah, indeed... I thought `UseAsMacros` extends this so it would need to be public... But it doesnt :)

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: I don't think you are allowed to make macro implementations non-public. But I'll try.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: > Shouldn't you check that A is not a subtype of js.Any?

No. If `A` is a subtype of `js.Any`, it will just not export anything. So either we don't need any methods, in which case we are fine, or we need some methods and the macro fails.

I like your other proposal though. I'll implement it as a follow up (will require quite some tests, want to get this in first).

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: Ok.

gzm0: Although it might be a common occurrence... I don't know...

gzm0: Should we drop it and add again if necessary?

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: Yes! Thanks. I hated this name, too.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: Hmmm... I might actually go with the ADT, otherwise the error messages will be very bad (and in the future we'll need to not always fail, since two `js.Any`s can actually conform in their apply method).

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: `toMap` does not make it eager:

```scala
scala> val a = Map(1 -> "a", 2 -> "b")
a: scala.collection.immutable.Map[Int,String] = Map(1 -> a, 2 -> b)

scala> val b = a.mapValues(println).toMap
a
b
b: scala.collection.immutable.Map[Int,Unit] = Map(1 -> (), 2 -> ())

scala> b(1)
a
```

And yes, `list` already a `Map`.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: Nope...

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: They are of a special type (the repeated type). Macros currently do not give you any ability to query for such a type, but the comparisons work properly.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: I don't know :P I think the class definition wouldn't compile.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: I changed it

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
gzm0: Ok

gzm0:
First update coming up TODOs:

- Bracket select/call
- method named apply with a custom `@JSExport("blahblah")` in the presence of `@JSExportAll`
- `jsName` API
- Traits only?
- Also test `js.Any`s (follow up ticket)
- Overloading/default interop

library/src/main/scala/scala/scalajs/js/package.scala
sjrd: Maybe. I don't think so. But in the absence of consensus, your definition is the most conservative one, so it should win. Widening the possibilities is possible later; constraining them is not.

library/src/main/scala/scala/scalajs/js/package.scala
sjrd: You did. But I still don't see a case where the method is present, has the right type and everything, but is not exported. And I think this is pretty central to the point of this macro.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
sjrd: OK

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
sjrd: Yes, I think we should drop it. Keep it safe in a commit somewhere, in case we need it later.

library/src/main/scala/scala/scalajs/macroImpls/UseAsMacros.scala
sjrd: OK

library/src/main/scala/scala/scalajs/js/package.scala
gzm0: Oh... I read the comment wrong...

gzm0:
> method named apply with a custom @JSExport("blahblah") in the presence of @JSExportAll

There is a warning (since `@JSExport("apply")` is not present). The method is exported under both names, typechecking works as expected. Do you want me to add a test for that?

library/src/main/scala/scala/scalajs/macroimpls/UseAsMacros.scala
```diff
@@ -0,0 +1,303 @@
...
+
+          if (!sym.isTrait)
+            c.abort(c.enclosingPosition, "Only traits can be used with as")
+
+          def allowedParent(sym: Symbol) =
+            sym.asClass.isTrait || JSObjectAncestors(sym)
+
+          for (base <- sym.baseClasses if !allowedParent(base)) {
+            c.abort(c.enclosingPosition, s"Supertype ${base.fullName} of $sym " +
+                "is a class. Cannot be used with as")
```
gzm0: I will also add a `.` here.

gzm0:
Updated. @sjrd PTAL

sjrd:
> Do you want me to add a test for that?

Not necessarily. But it can't hurt.

sjrd:
LGTM :D

PR #1802, opened 2015-07-16
Fix #964: Add missing Character methods getType and isXXX.

### LINKED ISSUE #964: Implement `isXXX` methods on Character
https://github.com/gzm0/mothereff.in/commits/scala-js-character shows how this can be done automatically. f35027725f9e4007c413ed11478a25638d21867f is an example that uses the generated regexes.

The code in the fork should be integrated in an utilitary script directory and all (or more) methods on `java.lang.Character` should be generated.

### LINKED ISSUE #1801: Character.isLetter is missing
and [is used by `scala.concurrent.duration.Duration.apply(s: String)`](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/concurrent/duration/Duration.scala#L60).

nicolasstucki:
* Add `getType`
* Add `isDefined`, `isMirrored`, `isTitleCase`, `isSpaceChar`, `isLetterOrDigit`, `isLetter`,
`isAlphabetic`, `isJavaLetter`, `isJavaLetterOrDigit`, `isJavaIdentifierStart`, `isJavaIdentifierPart`.
* Reimplement methods based on `getType` rather than regex.
* Fix #1801
* Fix `isDigit`


nicolasstucki:
Review by @sjrd

nicolasstucki:
Fail because `isAlphabetic` is not defined in JDK6.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -194,8 +215,25 @@ object Character {
   final val MIN_SUPPLEMENTARY_CODE_POINT = 0x10000
 
   // Not implemented:
-  //def getType(ch: scala.Char): scala.Int
-  //def getType(codePoint: scala.Int): scala.Int
+  def getType(ch: scala.Char): scala.Int = getType(ch.toInt)
+
+  def getType(codePoint: scala.Int): scala.Int = {
+    if (codePoint < 256) getTypeTL256(codePoint)
+    else getTypeGTE256(codePoint)
```
sjrd: `LT` and `GE` are more customary acronyms for lower-than and greater-or-equal in this codebase.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -194,8 +215,25 @@ object Character {
   final val MIN_SUPPLEMENTARY_CODE_POINT = 0x10000
 
   // Not implemented:
-  //def getType(ch: scala.Char): scala.Int
-  //def getType(codePoint: scala.Int): scala.Int
+  def getType(ch: scala.Char): scala.Int = getType(ch.toInt)
+
+  def getType(codePoint: scala.Int): scala.Int = {
+    if (codePoint < 256) getTypeTL256(codePoint)
```
sjrd: Need to take into account the case `codePoint < 0` (which should return 0).

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -227,15 +265,22 @@ object Character {
...
+  def isSpace(c: scala.Char): scala.Boolean =
+    c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' '
+  def isWhitespace(c: scala.Char): scala.Boolean =
+    js.RegExp("^\\s$").test(c.toString)
+
+  def isSpaceChar(ch: scala.Char): scala.Boolean = isSpaceChar(ch.toInt)
+  def isSpaceChar(codePoint: scala.Int): scala.Boolean =
+    isSpaceCharImpl(getType(codePoint))
+  @inline private[this] def isSpaceCharImpl(tpe: scala.Int): scala.Boolean =
+    tpe == SPACE_SEPARATOR || tpe == LINE_SEPARATOR || tpe == PARAGRAPH_SEPARATOR
```
sjrd: Add a blank line between each method.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -227,15 +265,22 @@ object Character {
...
-  def isWhitespace(c: scala.Char): scala.Boolean = js.RegExp("^\\s$").test(c.toString)
-  //def isSpaceChar(c: scala.Char): scala.Boolean
+  @deprecated("Replaced by isWhitespace(char)", "")
+  def isSpace(c: scala.Char): scala.Boolean =
+    c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' '
+  def isWhitespace(c: scala.Char): scala.Boolean =
+    js.RegExp("^\\s$").test(c.toString)
+
+  def isSpaceChar(ch: scala.Char): scala.Boolean = isSpaceChar(ch.toInt)
+  def isSpaceChar(codePoint: scala.Int): scala.Boolean =
```
sjrd: `scala.Int` -> `Int`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -258,23 +303,31 @@ object Character {
...
+      '\u00AA' == c || '\u00BA' == c || getTypeTL256(c) == LOWERCASE_LETTER
+    } else {
+      ('\u02B0' <= c && c <= '\u02B8') || ('\u02C0' <= c && c <= '\u02C1') ||
+          ('\u02E0' <= c && c <= '\u02E4') || '\u0345' == c || '\u037A' == c ||
+          ('\u1D2C' <= c && c <= '\u1D6A') || '\u1D78' == c ||
+          ('\u1D9B' <= c && c <= '\u1DBF') || '\u2071' == c || '\u207F' == c ||
+          ('\u2090' <= c && c <= '\u209C') || ('\u2170' <= c && c <= '\u217F') ||
+          ('\u24D0' <= c && c <= '\u24E9') || ('\u2C7C' <= c && c <= '\u2C7D') ||
+          '\uA770' == c || ('\uA7F8' <= c && c <= '\uA7F9') ||
+          getTypeGTE256(c) == LOWERCASE_LETTER
```
sjrd: Extract the `else` branch in a separate method, so that the fast-path method is smaller.
Also, there should be no additional indentation for the second line and following wrt. the first line of the big `||`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -258,23 +303,31 @@ object Character {
...
+    }
+ }
 
   def isUpperCase(c: scala.Char): scala.Boolean =
-    reIsUpper.test(c.toString)
+    isUpperCase(c.toInt)
+
+  def isUpperCase(c: scala.Int): scala.Boolean = {
+    ('\u2160' <= c && c <= '\u216F') || ('\u24B6' <= c && c <= '\u24CF') ||
+        getType(c) == UPPERCASE_LETTER
```
sjrd: No additional indentation.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -282,8 +335,151 @@ object Character {
...
+    if (c < 888) true
+    else getTypeGTE256(c) != UNASSIGNED
+
+  def isLetter(c: scala.Char): scala.Boolean = isLetter(c.toInt)
+
+  def isLetter(cp: scala.Int): scala.Boolean = isLetterImpl(getType(cp))
+
+  @inline private[this] def isLetterImpl(tpe: scala.Int): scala.Boolean = {
+    tpe == UPPERCASE_LETTER || tpe == LOWERCASE_LETTER ||
+        tpe == TITLECASE_LETTER || tpe == MODIFIER_LETTER || tpe == OTHER_LETTER
```
sjrd: No additional indentation.

sjrd: (and a few more like that ...)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -282,8 +335,151 @@ object Character {
...
+      ('\u007F' <= codePoint && codePoint <= '\u009F') ||
+      tpe == FORMAT
+  }
+
+  def isMirrored(c: scala.Char): scala.Boolean =
+    isMirrored(c.toInt)
+
+  def isMirrored(codePoint: scala.Int): scala.Boolean = {
+    val idx = Arrays.binarySearch(isMirrorIndices, codePoint) + 1
+    (Math.abs(idx) & 1) != 0
```
sjrd: Hum, this should simply be `Arrays.binarySearch(isMirrorIndices, codePoint) >= 0`, shouldn't it?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -316,25 +512,455 @@ object Character {
...
+      25, 2, 25, 2, 1, 25, 2, 25, 2, 1, 2, 0, 9, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5,
+      0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5,
+      0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5, 0, 5,
+      0, 5, 0, 25, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 11, 0, 28, 0, 28, 0,
+      28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0,
+      28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0, 28, 0,
+      28, 0, 5, 0, 5, 0, 5, 0, 5, 0, 16, 0, 16, 0, 6, 0, 18, 0, 18, 0)
+
+  // Types of characters from 0 to 255
+  private[this] lazy val charTypesFirst256 = Array[scala.Byte](15, 15, 15, 15,
```
sjrd: If feel like this table should come before the two other tables ;)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -316,25 +512,455 @@ object Character {
...
+      15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 12, 24, 26, 26, 26,
+      26, 28, 24, 27, 28, 5, 29, 25, 16, 28, 27, 28, 25, 11, 11, 27, 2, 24, 24,
+      27, 11, 5, 30, 11, 11, 11, 24, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+      1, 1, 1, 1, 1, 1, 1, 1, 1, 25, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2,
+      2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 25, 2, 2, 2, 2, 2, 2,
+      2, 2)
+
+  // Indices representing the start of ranges of codePoint that have the same
+  // `isMirrored` result. It is true for the first range
+  // (i.e. isMirrored(40)==true, isMirrored(41)==true, isMirrored(42)==false)
```
sjrd: Wasn't it more compact to simply have all the `i`s such that `isMirrored(i)` is true?

sjrd:
That's all.

javalanglib/src/main/scala/java/lang/Character.scala
nicolasstucki: No, the representation of `isMirrorIndices` is the same one as the `charTypeIndices` (by indices ranges). The version with an explicit list of mirrored character ended up being 3 times larger.

javalanglib/src/main/scala/java/lang/Character.scala
nicolasstucki: Clearly it should go first

javalanglib/src/main/scala/java/lang/Character.scala
nicolasstucki: 3 times larger. There are a few 'large' ranges that make it more compact this way.

nicolasstucki: 3 times larger in character count.

gzm0:
Could you list all the methods you are adding in the commit body, please? Also, if these are *all* `isXXX` methods on `Character`, I think we have a ticket for that.

sjrd:
Yes, we have #964. We also have #1801. These tickets should be mentioned.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -28,30 +30,30 @@ class Character(private val value: scala.Char) extends Comparable[Character] {
   protected def toByte: scala.Byte     = value.toByte
   protected def toShort: scala.Short   = value.toShort
   protected def toChar: scala.Char     = value.toChar
-  protected def toInt: scala.Int       = value
+  protected def toInt: Int       = value
```
sjrd: Hum, don't change places in the code that have nothing to do with the commit. (I guess this was a massing search-and-replace.)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -194,10 +215,28 @@ object Character {
   final val MIN_SUPPLEMENTARY_CODE_POINT = 0x10000
 
   // Not implemented:
```
sjrd: The `// Not implemented:` should disappear too :)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -194,10 +215,28 @@ object Character {
...
+
+  private[this] def getTypeGE256(codePoint: Int): scala.Byte = {
+    // the idx is increased by 1 due to the differences in indexing
+    // between charTypeIndices and charType
+    val idx = Arrays.binarySearch(charTypeIndices, codePoint) + 1
+    // in the case where idx is negative (-insertionPoint - 1)
+    charTypes(Math.abs(idx))
+  }
+
+
```
sjrd: Double blank line.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -227,15 +266,26 @@ object Character {
...
-  //def isLetter(c: scala.Char): scala.Boolean
-  //def isLetterOrDigit(c: scala.Char): scala.Boolean
-  def isWhitespace(c: scala.Char): scala.Boolean = js.RegExp("^\\s$").test(c.toString)
-  //def isSpaceChar(c: scala.Char): scala.Boolean
+  @deprecated("Replaced by isWhitespace(char)", "")
+  def isSpace(c: scala.Char): scala.Boolean =
+    c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' '
+
+  def isWhitespace(c: scala.Char): scala.Boolean =
+    js.RegExp("^\\s$").test(c.toString)
```
sjrd: The overload taking an `Int` is missing. It would also be a good opportunity to check that this regex-based implementation is really accurate. Maybe we should just use the same infrastructure as everything else anyway, at this point?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -253,28 +303,39 @@ object Character {
...
-
-  def isIdentifierIgnorable(c: scala.Char): scala.Boolean =
-    reIdentIgnorable.test(c.toString)
+  def isLowerCase(c: scala.Char): scala.Boolean =
+    isLowerCase(c.toInt)
 
-  //def isMirrored(c: scala.Char): scala.Boolean
+  def isLowerCase(c: Int): scala.Boolean = {
+    if (c < 256)
+      '\u00AA' == c || '\u00BA' == c || getTypeLT256(c) == LOWERCASE_LETTER
```
sjrd: Prefer `c == '\u00AA'`. In general, we put constants at the right-hand-side of comparisons. (with the exception of the range check, as in `'\u02B0' <= c && c <= '\u02B8'` like you do below)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -253,28 +303,39 @@ object Character {
...
-    reIsLower.test(c.toString)
+  private[this] def isLowerCaseGE256(c: Int): scala.Boolean = {
+    ('\u02B0' <= c && c <= '\u02B8') || ('\u02C0' <= c && c <= '\u02C1') ||
+      ('\u02E0' <= c && c <= '\u02E4') || '\u0345' == c || '\u037A' == c ||
+      ('\u1D2C' <= c && c <= '\u1D6A') || '\u1D78' == c ||
+      ('\u1D9B' <= c && c <= '\u1DBF') || '\u2071' == c || '\u207F' == c ||
+      ('\u2090' <= c && c <= '\u209C') || ('\u2170' <= c && c <= '\u217F') ||
+      ('\u24D0' <= c && c <= '\u24E9') || ('\u2C7C' <= c && c <= '\u2C7D') ||
+      '\uA770' == c || ('\uA7F8' <= c && c <= '\uA7F9') ||
+      getTypeGE256(c) == LOWERCASE_LETTER
```
sjrd: I really meant *no* indentation wrt to the first line, so:
```scala
    ('\u02B0' <= c && c <= '\u02B8') || ('\u02C0' <= c && c <= '\u02C1') ||
    ('\u02E0' <= c && c <= '\u02E4') || '\u0345' == c || '\u037A' == c ||
    ...
    getTypeGE256(c) == LOWERCASE_LETTER
```
See https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#long-expressions-with-binary-operators
Same everywhere else with this pattern, of course.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -282,8 +343,151 @@ object Character {
...
+  def isDigit(cp: Int): scala.Boolean =
+    if (cp < 256) '0' <= cp && cp <= '9'
+    else isDigitImpl(getTypeGE256(cp))
+
+  @inline private[this] def isDigitImpl(tpe: Int): scala.Boolean =
+    tpe == DECIMAL_DIGIT_NUMBER
+
+  def isDefined(c: scala.Char): scala.Boolean =
+    if (c < 888) true
+    else getTypeGE256(c) != UNASSIGNED
```
sjrd: The overload with `Int` is missing (and for that one, beware the `cp < 0` case)

sjrd:
[`isBmpCodePoint`](http://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isBmpCodePoint%28int%29) is missing. It's a trivial one, you might as well include it.

sjrd:
I will trust the test suite ^^
That's all.

ci/checksizes.sh
```diff
@@ -34,22 +34,22 @@ REVERSI_OPT_GZ_SIZE=$(stat '-c%s' "$REVERSI_OPT.gz")
...
+    REVERSI_PREOPT_GZ_EXPECTEDSIZE=99000
+    REVERSI_OPT_GZ_EXPECTEDSIZE=49000
     ;;
   2.11.7)
-    REVERSI_PREOPT_EXPECTEDSIZE=608000
-    REVERSI_OPT_EXPECTEDSIZE=134000
-    REVERSI_PREOPT_GZ_EXPECTEDSIZE=84000
-    REVERSI_OPT_GZ_EXPECTEDSIZE=38000
+    REVERSI_PREOPT_EXPECTEDSIZE=639000
+    REVERSI_OPT_EXPECTEDSIZE=155000
```
sjrd: Should be 156000, [says CI](https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/50476/console)

nicolasstucki:
Updated

ci/checksizes.sh
nicolasstucki: Did not increase because the character count of  `charTypeIndices` was reduced by half by representing it with deltas.

nicolasstucki:
Updated

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -34,24 +36,24 @@ class Character(private val value: scala.Char) extends Comparable[Character] {
   protected def toDouble: scala.Double = value.toDouble
 
   // scalastyle:off disallow.space.before.token
-  protected def unary_~ : scala.Int = ~value
-  protected def unary_+ : scala.Int = value
-  protected def unary_- : scala.Int = -value
+  protected def unary_~ : Int = ~value
+  protected def unary_+ : Int = value
+  protected def unary_- : Int = -value
```
sjrd: There's still a bunch of unrelated changes. Also below.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -253,37 +315,198 @@ object Character {
...
+
+  def isJavaIdentifierPart(ch: scala.Char): scala.Boolean =
+    isJavaIdentifierPart(ch.toInt)
+
+  def isJavaIdentifierPart(codePoint: Int): scala.Boolean =
+    isJavaIdentifierPartImpl(codePoint, getType(codePoint))
+
+  @inline private[this] def isJavaIdentifierPartImpl(codePoint: Int,
+      tpe: Int): scala.Boolean = {
+    isLetterImpl(codePoint) || tpe == CURRENCY_SYMBOL ||
```
sjrd: This is suspicious. Shouldn't it be `isLetterImpl(tpe)`?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -313,28 +536,353 @@ object Character {
...
+      5, 4, 5, 3, 3, 14, 403, 22, 2, 21, 8, 1, 7, 6, 3, 1, 4, 5, 1, 2, 2, 5,
+      4, 1, 1, 3, 2, 2, 10, 6, 2, 2, 12, 19, 1, 4, 2, 1, 1, 1, 2, 1, 1, 4,
+      5, 2, 6, 3, 24, 2, 11, 2, 4, 4, 1, 2, 2, 2, 4, 43, 2, 8, 1, 40, 5, 1,
+      1, 1, 3, 5, 5, 3, 4, 1, 3, 5, 1, 1, 772, 4, 3, 2, 1, 2, 14, 2, 2, 10,
+      478, 10, 2, 8, 52797, 6, 5, 2, 162, 2, 18, 1, 1, 1, 28, 1, 1, 1, 29,
+      1, 1, 1, 1, 2, 1, 2, 55159, 1, 57, 1, 57, 1, 57, 1, 57, 1)
+    uncompressedDeltas(deltas)
+  }
 
+  private[this] def uncompressedDeltas(deltas: Array[Int]): Array[Int] = {
```
sjrd: Ouch. If it's side-effecting, it should be called `uncompressDeltas`, and return `Unit`.

nicolasstucki:
retest this please

nicolasstucki:
retest this please

nicolasstucki:
Updated

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -253,37 +315,198 @@ object Character {
...
+  @inline private[this] def isDigitImpl(tpe: Int): scala.Boolean =
+    tpe == DECIMAL_DIGIT_NUMBER
+
+  def isDefined(c: scala.Char): scala.Boolean =
+    isDefined(c.toInt)
+
+  def isDefined(c: scala.Int): scala.Boolean =
+    if (c < 0) false
+    else if (c < 888) true
+    else getTypeGE256(c) != UNASSIGNED
```
sjrd: To be completely precise, this method needs `{}`. Since it has 3 lines, it's not a two-liner if/else.

sjrd:
And finally the commit message should mention the issues.

Title: "Fix #964: ..."
And in the body, when mentioning `isLetter`, also say "Fix #1801" somehow.

That's all.

javalanglib/src/main/scala/java/lang/Character.scala
gzm0: I think we should allow this...

sjrd: I disagree. If you allow 3, you have to allow 10. There's no limit anymore. And I'm pretty sure I don't want to see 10 lines of else if's without braces around them.

gzm0: Ok :)

gzm0:
How did you generate the indices array? The reason I am asking, is it might be a good idea to put that script under version control, too. So we can use it again, in case we want to upgrade to a newer unicode version. (Which one are you based on ?)

nicolasstucki:
@gzm0 They where patly genreated by a script a then to commpress them i just used the REPL. I'll write a consice version that will generate the indices and types on any JVM. I did not follow some Unicode version, I took all the outputs directly from the JVM.

gzm0:
Ok. Sounds good. In this case IMO it is the right thing to **not** make a source generator  (since the result might depend on the JDK version), but we should have a way to reproduce the results.

nicolasstucki:
retest this please

sjrd:
LGTM

sjrd:
retest this please

sjrd:
Rebuild failed job: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/51181/

sjrd:
retest this please

PR #1807, opened 2015-07-17
Remove JDK 6 from CI, essentially dropping support of JDK 6.

sjrd:
We have been dragging support for JDK 6 for a very long time already. Even though some parts of our architecture did not even work correctly on that platform (e.g., partests).

This commit removes all traces of JDK 6 from our CI matrix, essentially dropping support of JDK 6.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -266,19 +266,11 @@
     <!-- Main test tasks -->
     <run task="main">
       <v n="scala">2.10.2</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.10.2</v>
       <v n="java">1.7</v>
     </run>
     <run task="main">
       <v n="scala">2.10.2</v>
       <v n="java">1.8</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.11.7</v>
-      <v n="java">1.6</v>
     </run>
     <run task="main">
       <v n="scala">2.11.7</v>
@@ -296,17 +288,12 @@
     <!-- Bootstrap test tasks -->
     <run task="bootstrap">
       <v n="scala">2.10.2</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="bootstrap">
-      <v n="scala">2.10.2</v>
       <v n="java">1.7</v>
     </run>
     <run task="bootstrap">
       <v n="scala">2.10.2</v>
       <v n="java">1.8</v>
     </run>
-    <!-- Tools do not compile on JDK6, Scala 2.11.x (see #1235) -->
     <run task="bootstrap">
       <v n="scala">2.11.7</v>
       <v n="java">1.7</v>
@@ -323,17 +310,12 @@
     <!-- Tools / CLI / Stubs / sbtPlugin test tasks -->
     <run task="tools-cli-stubs-sbtplugin">
       <v n="scala">2.10.5</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="tools-cli-stubs-sbtplugin">
-      <v n="scala">2.10.5</v>
       <v n="java">1.7</v>
     </run>
     <run task="tools-cli-stubs-sbtplugin">
       <v n="scala">2.10.5</v>
       <v n="java">1.8</v>
     </run>
-    <!-- Tools do not compile on JDK6, Scala 2.11.x (see #1235) -->
     <run task="tools-cli-stubs">
       <v n="scala">2.11.7</v>
       <v n="java">1.7</v>
@@ -370,10 +352,6 @@
     </run>
     <run task="main">
       <v n="scala">2.10.5</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.10.5</v>
       <v n="java">1.7</v>
     </run>
     <run task="main">
@@ -390,10 +368,6 @@
     </run>
     <run task="main">
       <v n="scala">2.11.2</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.11.2</v>
       <v n="java">1.7</v>
     </run>
     <run task="main">
@@ -402,10 +376,6 @@
     </run>
     <run task="main">
       <v n="scala">2.11.4</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.11.4</v>
       <v n="java">1.7</v>
     </run>
     <run task="main">
@@ -414,19 +384,11 @@
     </run>
     <run task="main">
       <v n="scala">2.11.5</v>
-      <v n="java">1.6</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.11.5</v>
       <v n="java">1.7</v>
     </run>
     <run task="main">
       <v n="scala">2.11.5</v>
       <v n="java">1.8</v>
-    </run>
-    <run task="main">
-      <v n="scala">2.11.6</v>
-      <v n="java">1.6</v>
     </run>
     <run task="main">
       <v n="scala">2.11.6</v>
@@ -445,10 +407,6 @@
     <run task="bootstrap">
       <v n="scala">2.10.4</v>
       <v n="java">1.7</v>
-    </run>
-    <run task="bootstrap">
-      <v n="scala">2.10.5</v>
-      <v n="java">1.6</v>
     </run>
     <run task="bootstrap">
       <v n="scala">2.10.5</v>
@@ -614,10 +572,6 @@
       <v n="scala">2.11.7</v>
       <v n="java">1.8</v>
     </run>
-    <!--
-        Partest does sometimes not compile on JDK6 (see #1227) we
-        therefore do not run any JDK6 partests.
-      -->
   </matrix>
 
 </ci>


sjrd:
Review by @gzm0, @nicolasstucki

gzm0:
Yeah... I guess considering that JDK 6 has reached public EOL over 2 years ago (!), this is safe to do. However, I would like to wait at least 2-3 weeks before merging this, since it has a very severe impact if someone still relies on this.

Is there anything in our pipeline that needs us to drop JDK 6 immediately?

Otherwise, how about the following: We officially announce that we will drop support for JDK 6 on Sep 1 unless asked to not do so. This will allow me to announce this at Scala By The Bay on Aug 15 and still give people ample time to react.

Another, less hardcore option would be to announce for a earlier point in time and write a blog post, cross post to scala-lang (scala-announce) and try to make as large of a twitter buzz as we can.

I just think this is not something to be taken lightly now that we are really trying to drive adoption as strongly as we can. Our users need to have the confidence that they have ample time to react if we consider taking something away from them. Especially if it is something that costs us relatively little (although not having a newer closure version is kind of sad).

sjrd:
> Especially if it is something that costs us relatively little (although not having a newer closure version is kind of sad).

Actually, at this point, it costs us a lot in terms of CI wall clock time. Each PR spawns more jobs than we have machines running Jenkins, so even a single PR needs to wait for some of its jobs to complete before starting new ones. This has come to a point where it slows down development in a noticeable way.

> Is there anything in our pipeline that needs us to drop JDK 6 immediately?

No, nothing.

We can definitely wait for a couple weeks. Anyway from July 29 to August 16 I'll be on vacation. So we can wait until mid August "for free". We can then add another two weeks until Sep 1 if you think that's relevant.

Announcement on scala-lang is a bit of a stretch. But we should advertise through all the Scala.js channels. I've done Gitter and the mailing list already. I'll also post a tweet.

gzm0:
Ok. That sounds reasonable.

antonkulaga:
Yes, it will be great to drop it!

bvenners:
Um, does this mean that for ScalaTest/Scalactic builds we'd need to build for the JVM on Java 6, then switch and build for Scala.js on Java 8? We do support Java 6 as a JVM target because Scala supports it.

That aside I wonder why you wouldn't just wait until Scala 2.12, when the whole Scala ecosystem will need to do a build on Java 6 for 2.10, 2.11, and another build on Java 7 for 2.12. I'm not sure what the latest roadmap for Scala 2.12 is, but surely it isn't that far off (for some value of "far off").

bvenners:
+1 to what Bill said. By which I mean Java 7 should be Java 8.

gzm0:
@bvenners could you please clarify or adapt your old comment? I don't understand anymore where you mean Java 6, Java 7 or Java 8 and where you want what to be replaced...

sjrd:
@bvenners We will still support java 7, and builds that target the JDK 6 classfile format. The Scala 2.11 jvm compiler is perfectly able to produce artifacts for jvm 6 while itself and sbt running on a jvm 7 or 8.

You will need a JDK 7 to *build* (by which I mean using the sbt plugin). Right now you are probably already using a JDK 8 to build even your JDK 6-compatible jvm artifacts.

Does that clarify anything?

bvenners:
I had problems in the past trying to build for earlier JDK versions using later JDK versions. The class file format can be produced by the later JDK, yes, but you can inadvertently link to a JDK library method that only exists in the later JDK. I'm not sure if that matters at all for Scala.js, since the ultimate target isn't the JDK. But to clarify, the problem was not the classfile format. It was linking to libraries. I searched and found one example:

https://groups.google.com/forum/#!topic/scalatest-users/9PJy1FKGlUo

bvenners:
Hi @sjrd,

I wanted to reply to your comments specifically:

> We will still support java 7, and builds that target the JDK 6 classfile format. The Scala 2.11 jvm
> compiler is perfectly able to produce artifacts for jvm 6 while itself and sbt running on a jvm 7 or 8.

I believe that's true for class file format, but is it true for standard libraries? It wasn't true in the past.

> You will need a JDK 7 to build (by which I mean using the sbt plugin). Right now you are probably
> already using a JDK 8 to build even your JDK 6-compatible jvm artifacts.

No we always use whatever minimum version of the JDK we say we support, which right now is JDK 6 (because we always just support whatever minimum JDK version is supported by the Scala versions we support). 
 
> Does that clarify anything? 

I'm not sure yet whether "dropping support for JDK 6" will mean we can't build for Scala.js on JDK 6, but I'm guessing that's what it means. Worst comes to worst, what I'll have to do is build for the JVM on JDK 6, then switch to JDK 7 or 8 and build for Scala.js. I'd rather just stick to JDK 6, so I can continue to use our one-button build, which works on any platform.

The other thing I could try to do is make people point to a JDK 6 library, as Seth suggested, perhaps with an environment variable. Perhaps that would work, but as I said before, I don't trust it enough. ScalaTest is too widely used and it needs to just work. The simplest and lowest risk way to make sure we really support JDK 6 is if I can continue to just build ScalaTest and Scalactic (for the JVM) on JDK 6 for Scala 2.10 and 2.11.

JDK 6 has been end-of-lifed by Oracle already, so at some point we should all stop supporting it and move on. The way I decide when to do that for ScalaTest & Scalactic, and what I would suggest you consider doing for Scala.js, is follow the lead of Scala itself. Scala already announced it will jump to JDK 8 for 2.12, so that's what we should do. Continue to support JDK 6 for 2.10 and 2.11, and jump to JDK 8 for 2.12.

sjrd:
> I'm not sure yet whether "dropping support for JDK 6" will mean we can't build for Scala.js on JDK 6, but I'm guessing that's what it means.

Yes, that's what it means.

> Worst comes to worst, what I'll have to do is build for the JVM on JDK 6, then switch to JDK 7 or 8 and build for Scala.js. I'd rather just stick to JDK 6, so I can continue to use our one-button build, which works on any platform.

Hum, those are valid concerns, indeed ...

> The way I decide when to do that for ScalaTest & Scalactic, and what I would suggest you consider doing for Scala.js, is follow the lead of Scala itself. Scala already announced it will jump to JDK 8 for 2.12, so that's what we should do. Continue to support JDK 6 for 2.10 and 2.11, and jump to JDK 8 for 2.12.

This is what I intended to do, but I realized that's wishful thinking. Since Scala.js supports 2.10.2+, it doesn't matter whether 2.12 drops supports for JDK 8. With your reasoning, we'll have to support JDK 6 until we drop support for 2.10 *and* 2.11. That means we'll have to continue to support JDK 6 for several *years*. That's an immense drag for an EOL'ed VM :(

lihaoyi:
One question w.r.t. Scala, ScalaTest and Scala.js is the profile of its users.

If we think Scala.js has a profile similar to Scala & ScalaTest in terms of legacy inertia, then it makes sense to provide a similar level of legacy support. If we think they're generally more progressive, it makes sense of provide less.

I'd say that the community is generally more progressive than that which uses ScalaTest or Scala as a whole, and so supporting Java 6 for 7 years (2 past, + maybe 5 more in future!) past EOL simply because that's what Scala/ScalaTest does doesn't really make sense.

It makes sense if we bill ourselves as an "enterprise-stable legacy-supporting solution", the same way Oracle or Typesafe have billed themselves, but we're not. And even Oracle has EOLed Java 6, and Typesafe is winding down work on 2.11 to prepare to work on 2.12! Should the Scala.js community really consider itself more enterprise-stable legacy-supporting than either Typesafe or Oracle?

bvenners:
Yes, that's true. You'd need to support JDK 6 until you drop support for 2.10 and 2.11. But I'm not 100% sure I understand what that means exactly, since Scala.js doesn't run on the JVM. Is what it means that we won't be able to *build* for Scala.js on JDK 6? That means people running the Scala.js compiler will need to use JDK 7 or 8? If so it sounds like everybody building libraries for both Scala and Scala.js for 2.10 and 2.11 will need to build on multiple versions of the JDK. Maybe other libraries don't bother building on JDK 6, but I would ask around. Perhaps it is a solved problem now to build on JDK 7 or 8 for JDK 6. If not, though, then dropping for Scala.js support for JDK 6 would be "a drag" for library developers. I.e., it would make it more painful to port and support Scala libraries on Scala.js. (I may be the odd man out though. We should ask around to find out how many people build Scala libraries on JDK 6.)

bvenners:
@lihaoyi I would look at it more from the perspective of what do your users need. What I've observed in the past is most ScalaTest users are on more recent JDKs, but a few aren't.

The way I found this out was that back in the old days when Apple released the JDK for the Mac (before Oracle started doing it), Apple would upgrade my Java version automatically without me realizing it (and delete the older version I wanted to use for ScalaTest builds). So I thought I was building on JDK 5, but I was really building on JDK 6. Tests passed. Release went smoothly, but a few users couldn't upgrade because they were still on JDK 5. So I had to do *another* release. Apple kept doing that, so at one point we added something to our ant build to test the JDK version to make sure it was actually JDK 5! It finally stopped happening when Oracle took over making JDK releases for the Mac.

I would recommend *not* waiting until you drop support for JDK 6 to find out how many users are on it. The ScalaTest/Scalactic project is one user on JDK 6, but if the only one, then I can deal with it. I'd ask around to find out how many people actually build on JDK 6. If not many, then you can decide to make life more of a pain for those users. If most, then I'd think it would be in the interest of Scala.js project to swallow some of the pain and stick with JDK 6. It would make things easier for your users.

That said, we'll continue to support Scala.js regardless.

antonkulaga:
I agree with @lihaoyi Supporting JDK6 that even Oracle does not support does no make sence. Most of scalajs community is progressive and usually uses scala 2.11.x and JDK8  And actually the community is got used to migration as even minor releases of scalajs can contain serious changes, so all people who use outdated JDK6 with scalajs will simply migrate to JDK7/8

bvenners:
Hi @antonkulaga,

Maybe I'm beginning to see where you guys are coming from. Most of the Scala.s commmunity is likely progressive and using Scala 2.11 and at least JDK 7, so why bother supporting JDK 6 that Oracle does not support? Now that I think of it, Scala 2.10 and 2.11 support JDK 6! Those guys are clearly old fuddy duddies, not as progressive as us Scala.js people. Why should we bother supporting Scala 2.10 and 2.11 at all in fact? We should just drop support for those non-progressive Scala versions!

Since it seems to be non-obvious to some here, the reason Scala.js would support JDK 6 is simply because Scala 2.10 and 2.11 support it. It doesn't matter what Oracle may have end-of-lifed. Yes it is a bit of a pain to support Scala or JDK releases that a minority of users are using, but it isn't that painful and it serves the users. It makes *their* lives easier. And if you want to be a tool that a lot of people use, you need to adopt the attitude that you serve the users, not the other way around. 

Scala.js is a compiler, which puts it at the very center of the Scala ecosystem. I'd think you would probably want to adopt as a requirement that users can do seamless cross-builds for the JVM and JS. That will encourage libraries to be cross-built for Scala.js. Not doing that may encourage more of a two worlds ecosystem, where there are different libraries on the Scala JVM and Scala JS.

That said, if it really is possible to build on JDK 7 and support JDK 6, then it is a moot point. We should investigate that, and also find out how many other people build for Scala JVM on JDK 6. As I said, it may just be me, and if it is, then I can deal.

bvenners:
I posted a question on scala-internals here:

https://groups.google.com/forum/#!topic/scala-internals/Idyk9yem-ho

I just reread my previous post. Sorry @antonkulag if I came across too strong. I should never post to github while driving. I think the key idea is that some people, though they may be progressive at heart, are stuck on older JDK versions and can't upgrade even if they want to. People like the Scala folks doing 2.10 and 2.11, us doing ScalaTest/Scalactic, etc., and a minority of our users out there. I've heard of other libraries moving on ahead of Scala, like Play 2.4 requires JDK 8. So now I have to switch to JDK 8 to do a build of ScalaTest+Play. That's a pain but I do it. So I can do that for Scala.js, but Scala.js being a compiler, it seems it is more central so it may affect more users.

nicolasstucki:
LGTM

sjrd:
@nicolasstucki Don't post an "LGTM" on a PR which has so much controversy, even the code itself is OK ;) It's confusing ^^

bvenners:
Hi All,

So far I've heard back from only Miles Sabin about Shapeless and Sam Haliday about ENSIME. Sam does build in JDK 6 currently but plans to drop JDK 6 support for a future 1.0 release. Not sure when that is. Miles builds shapeless on JDK 7, but targets JDK 6 and includes JDK 6 in his test matrix to catch if he accidentally uses a method added in JDK 7. 

Bill

sjrd:
OK, so I guess that settles it, then. We're stuck with JDK 6 forever :-(

antonkulaga:
What about just going "typesafa way": I see that next akka will be free and opensource but people who want to use it with JDK6 will pay for its support.
@bvenners if JDK6 support is so crucial for you, maybe you can simply pay  @sjrd and @gzm0 some small ammount of money to compensate their inconveniences with JDK6 support?

sjrd:
Er ... Let's not go that way ^^

lihaoyi:
If we do that we'll have to rename the project ReactiveJS

antonkulaga:
@sjrd what is wrong with that? I think it is good that some money will go to fund your amazing work. Or, for scalajs tshirts.

sjrd:
@antonkulaga What's wrong is that @bvenners and others are open source contributors just as much as we are. I don't wanna talk about involving money, at all.

antonkulaga:
Sorry, I did not want to hurt anybody's feeling. I was just thinking about the way how the work of scalajs core devs can be funded, so everybody will feel safer 

BTW, @sjrd  what about adding "DONATE" button to the website, are you also against it?

gzm0:
@antonkulaga I think one of the more practical issues is that we completely lack a structure to even make use of this money. It would not be enough to pay anyone of us, our development infrastructure is on GitHub / at EPFL. So what should we do with the money except for buying beer? We prefer seeing people use what we do than getting material rewards (at least that is the case for me).

Scala.js has had a history of doing things the way "they should be"* rather than "how the company/entity that pays the money likes it hacked to make their specific use case easier but make everyone else's life harder". I am personally very proud and fond of that and its one of the things that keeps me working on this project even though I do so solely in my free time.

I support the concern that involving money may bring this out of balance and damage the community noticeably. (Would you like to be told: Sorry, I can't fix <generally relevant and severe bug>, since XYZ pays me to work on <loosely related and not generally useful feature).

In any case, I strongly agree that charging other open source projects for support they need is not the right way to go :)

*of course modulo the subjective perception of every individual of what the right thing is and human mistake

antonkulaga:
As you are afraid that money may spoil the community then it is better to leave it as it is. My main fear about money was that there are only several core devs of scalajs compiler and without any donations if they will suddently loose their jobs and will be unable to find alternative jobs with enough free time for scalajs development then scalajs compiler development will get stuck.

gzm0:
Yes, this is indeed a concern and is a relatively high risk for the Scala.js project IMO. However, we are not at a point where donations would actually pay a salary (I don't know how many OSS developers actually get payed that way). Also, I firmly believe that these donations should not come in on a feature basis but on a more global basis... (not that I have a solution for this... just thoughts...)

bvenners:
Hi All,

I'd like to apologize again too as I think my tone kicked off the feeling of "controversy." I have been ridiculously busy lately and posted that comment about the progressiveness of the Scala team without proof reading it. I surely would have edited it had I done that.

As for the question of money, ScalaTest/Scalactic actually comes from Artima, Inc., which does pay a small number developers to work on it. We have funded that over the years through a combination of consulting, training, and books. Some of the consulting contracts involved funds specifically for certain features to be added to ScalaTest then given away open source, but such contracts are quite hard to come by and certainly hasn't paid for the whole development cost as yet at least. But that model does seem to work (given that I'm willing to just give away free software that costs a lot of money to create), and it means that big things can get done. We have had very many community contributions as well, but they tend to be much smaller in scope. We very much appreciate every pull request we get no matter the size, but my experience so far has been it has been unlikely to expect a user to contribute anything very large.

I think that is due in part to how I've organized the projects. ScalaTest & Scalactic doesn't have a nice do list, like Cats, for example. (Maybe I should do that.) Other open source projects in our community are organized, or have self-organized, to have many active contributors doing relatively big things--at least that's my observation. For a project like that (say Scalaz, Cats), it would not be easy to figure out how to mix in a trait Money later down the road. 

The other thing I've noticed about open source projects with an associated business entity, is that they have sometimes been picked up by an established company that wants to adopt them. Maybe that could be a future for Scala.js or maybe not. Whatever the future, I think Scala.js is already a very important project for Scala, and will continue to grow in importance. That's why I just spent maybe five person-months time on porting ScalaTest and Scalactic to run on Scalajs, which unfortunately is ongoing because of a yak shave kicked off by one github issue to ScalaTest submitted by one of our first Scala.js users. I had to slow down because we needed to take a short-term contract, but as soon as possible I hope to release 3.0.0 final, which will have Scala.js support.

My observation is that it is hard to get users to contribute money for the free software they love, and can be hard to figure out what to do with it even if you get it. Even if you had a business entity or some way to charge for support, Artima would be quite hard pressed to pay it. What can work from our perspective, I think, is we'd be happy to help Scala.js in areas that help the ScalaTest/Scalactic project. So instead of money, we can donate is time (which is money), in areas that help our users. Two examples from recent past is we created a plugin for testing for the Eclipse IDE, which was a big investment on our part. (We wrote that to run any test framework, not just ScalaTest, but it hasn't been used that way much as far as I can tell.) We also spent a lot of time doing a 2nd version of the sbt Framework API, that sbt uses to talk to all test frameworks (not just ScalaTest). Both of those Artima just paid for. We never got any funding for that from anyone. I'm not sure there's any way we could help you with JDK 6 support, but I'm actually curious to know what issues you had other than the long CI cycle. 

Anyway, the Scala.js contributors have really done something quite impressive here, and I think the fact that people start complaining when you do things like mention you're thinking about dropping support for a JDK is a sign that you've created a successful project, one with a growing user base that cares about it.

(OK, hopefully that's a better tone. I did at least proof read it while riding the train to the airport.)

gzm0:
@bvenners Thanks a lot for your thoughts.

Just from the top of my head, reasons why we'd like to drop JDK 6:

- CI time
- Special handling of tests/forwarders for JDK 7 implementations (e.g. `scala.scalajs.niocharset.StandardCharsets`)
- Inability to use JDK 7 methods in `tools` (notably atomic move and `Files` methods).
- Lock-in to old Google Closure compiler version

So it is not tragic, but it is definitely an annoyance.

antonkulaga:
>Yes, this is indeed a concern and is a relatively high risk for the Scala.js project IMO. 

I think the bottleneck is the knowledge of scalajs compiler. If popular library developer will be "hitten by a bus" the ecosystem will recover very soon as there are many developers willing to contribute. Most time will be needed for sbt-tools and for scalajs compiler plugin it will be extremely dangerous. Probably a tutorial about sbt part of scalajs and scalajs-compiler may help somehow, but anyway the threshold seems to be very hight for compiler developers.


>Also, I firmly believe that these donations should not come in on a feature basis but on a more global basis... 

Global donations can be used to compensate travel expences to conferences and for T-shirts => Accomodation is often provided for free by local scala communities but plane tickets often are not.

>Just from the top of my head, reasons why we'd like to drop JDK 6:

It is possible just to make a code that will switch of scalajs projects in sbt configuration? As I undestand the problem of artima is that they provide tools/libs that crossbuild for JDK6 and if scalajs will only support JDK7-8 then they will have to make a branch for JDK6 that will not contains scalajs stuff

lihaoyi:
As another alternate POV, I've never published any of my libraries (which probably make up the majority of library downloads for Scala.js overall, no joke) for JDK 6, and I think the only complain I ever got was form @bvenners; he wasn't even going to use the library, just trying it out after I badmouthed Scalatest as bloated =P IIRC a number of people found things that didn't work but just upgraded to JDK8 when I suggested they try that. Zero people actually said that JDK6 support was a problem...

lihaoyi:
>  the fact that people start complaining when you do things like mention you're thinking about dropping support for a JDK is a sign that you've created a successful project

So far I've heard zero users complain, and only one library writer (i.e. @bvenners =P) I don't know what @milessabin said, but @fommil doesn't even publish ENSIME for Scala.js, isn't likely to in the foreseeable future, and a response like "oh yeah I publish for JDK 6" isn't the same as "I don't want you to drop support for JDK6", and even explicitly said he's gonna drop JDK6 in the near future!

I honestly don't see why this issue was closed so quickly, on the basis of basically zero relevant data. I'm not the one paying the price for JDK6 support since I don't contribute to scala-js/scala-js much anymore, and have no horse in this race, but it seems completely mysterious to me. One person thinks we should maintain support, 4 people are happy dropping it, and of the two randomly surveyed third parties, 1 is of unknown preference (did we even ask @milessabin whether he'd care if we dropped support? The fact that he supports it now isn't sufficient to make a decision), and the other is gonna drop it in the near future anyway.

That hardly seems like it should settle the matter in favor of keeping JDK6 support forever O.o

fommil:
It would impact us in ENSIME if somehow Scala.js forces shapeless and ScalaTest to use Java 7. But so long as those projects aren't dropping Java 6 runtime support, it's OK with me. We us she JDK6 in our CI because it lets us support the official Scala target platform without accidental inclusions of JDK7+ APIs.

milessabin:
Published shapeless binaries have been built with JDK 7 for as long as I can remember. Quite possibly forever.

I'm in the process of switching to JDK 8 for Scala 2.12.x compatibility and I'm removing JDK 6 from my main Travis build matrix. I'll be keeping a JDK 6 branch and Travis build around indefinitely as a canary, but TBH I'm really not sure that it's providing any benefit any more. The shapeless build is multi-branch, so juggling the various permutations is very straightforward for me.

IIRC the main issues that JDK 6 builds are supposed to catch are uses of JDK 7 Swing API and a method on `java.lang.String` that was introduced in JDK 7 (I can't remember which now) and collides with a Scala enrichment,  producing JDK 6 incompatible binaries if used. shapeless doesn't touch either of those so there won't be (and demonstrably hasn't been) a problem.

sjrd:
Hum, maybe we could approach this from a different angle, and make sure and test that the sbt-plugin of Scala.js is able to *load* on JDK 6 (but would possibly crash if any of its tasks were actually executed) and doesn't prevent JVM projects in the build to run?

That would guarantee to maintainers of cross-compiling libraries such as @bvenners that they can easily run the JVM tests on a JDK 6, to ensure API compatibility. To test the Scala.js artifacts, and to package/publish them, a JDK 7 would be necessary. But that just means two different Travis configurations, which is trivial to achieve.

Would that make sense?

SethTisue:
re "it must be a solved problem to build on different JDKs there"... sort of, depending on what you mean. The 2.11.x and 2.12.x community builds are basically separate. They are configured in separate branches of the community-builds repo and have separate Jenkins jobs. The 2.11.x build uses Java 6 only; the 2.12.x build uses Java 8 only. (We have discussed making a new Jenkins job that runs the 2.11.x community build on Java 8, but it's not absolutely clear it's worth the effort, since this hypothetical 2.11.x/JDK8 build would have to be another, separately maintained branch. In an ideal world the community build would test some crazy nineteen-dimensional matrix of every possible combinations of things, but in practice, this stuff is finicky and keeping two builds going is already a bunch of work. I intend to improve the documentation so it's easier for the community to help, but there's a limit to how much time I can spend on that, too.)

But I don't want to hijack this thread with a lot of detail about the community build. If you have questions about the role of ScalaTest in the community build, w/r/t different JDK and JRE versions, maybe open a ticket at https://github.com/scala/community-builds/issues and let's discuss there?

sjrd:
> I think basically you are suggesting we always build on JDK 7 and target JDK 6, and only test JDK 6 on Jenkins (not by building on JDK 6, but by running the tests of the JDK 7 build that targets JDK 6, on JDK 6). Is that correct?

Assuming you don't want to build selectively on 6 vs 7 for JVM vs JS, then yes.

> Also I'm not 100% sure that it will work if I build with JDK 7 and target JDK 6 just because our tests pass on JDK 6. Tests are not proof. [...] I want it to be guaranteed to work, and it is quite easy to do, just build on JDK 6 for Scala versions that support JDK 6.

I don't get your point, here. Your build is no more proof than your tests, so building locally on JDK 6 does not give you more guarantee than having one CI config running on JDK 6 to test for you. Or I'm missing something, here.

sjrd:
> The 2.11.x build uses Java 6 only

Ah, ouch! So, if we kick even the tiniest amount of support for JDK 6, even in Scala.js' specific tasks (with the possible exception of fullOptJS), we'll ruin the presence of Scala.js in the community build. That's not good at all. The same applies to any Scala project in the community build, IIUC. So basically we (at large) have no choice: we all have to keep supporting JDK 6 if we want to be in the community build (which, I believe, every core library wants).

SethTisue:
Well, I'll consider that a vote for getting the potential 2.11.x-on-Java-8 build I mentioned going, then!

SethTisue:
Sigh, yeah, you're right... I was re-thinking this even before you posted. Yeah, this (2.11.x community build in Java 8) is pretty important, so I'll make it a priority.

bvenners:
> I don't get your point, here. Your build is no more proof than your tests, so building locally on JDK 6
> does not give you more guarantee than having one CI config running on JDK 6 to test for you. Or
> I'm missing something, here.
>
To clarify, what I mean is that building on JDK 6 itself gives me great confidence that the resulting binaries will work with JDK 6, i.e., won't accidentally link to API artifacts introduced in JDK 7 or JDK 8. You're right it also isn't a proof, but it have quite high confidence that will be the case. 

I would have less confidence if I build on JDK 7 or JDK 8 targeting JDK 6 (which generates JDK 6 compatible bytecodes, but still links to the JDK 7/8 libraries), then run our tests using those binaries on JDK 6. The reason is that I will only catch an accidental invocation of a JDK 7/8-only method, for example, only if our test suite happens to call exercise the ScalaTest code that calls that method.

Does that make sense?

What might work, as I mentioned this earlier, is the suggestion Seth posted to my old Scala compiler issue. He said we could target earlier JDKs that we could tell the Scala compiler running on JDK 7 not only to target JDK 6, but also to link to this old JDK 6 library, not the JDK 7 one. I don't remember ever trying that or even noticing Seth's comment at the time. To do that repeatedly would make building ScalaTest a pain for everybody, because they'd need to set an environment variable pointing to the JDK 6 libraries. A lot of people would not know where their JDK 6 install is, and many would not even have one, and really that's just too painful for our users. So what I'd be forced to do is either drop JDK 1.6 support, drop Scala.js support, or start doing builds in two different steps, a JDK 6 step and a JDK 7 step. I'd probably do the latter I guess, because I can't really drop JDK 6 support and can't drop Scala.js support. Eventually I'll have to do that for JDK 6 and JDK 8, so I'll need to figure it out (and so will everyone else). But since the annoyance you have with JDK 6 seems manageable, it would be greatly preferred if you keep supporting it for 2.10 and 2.11.

Seth is the list of projects being built by the community build on JDK 6?

[ensime] --== End Building ensime ==--
[info] --== Console Notifications ==--
[info] ---==  Execution Report ==---
[info] Report from the dbuild run for dbuild: 
[info] Project akka--------------------: SUCCESS (unchanged, not rebuilt)
[info] Project async-------------------: SUCCESS (unchanged, not rebuilt)
[info] Project browse------------------: SUCCESS (unchanged, not rebuilt)
[info] Project discipline--------------: SUCCESS (project rebuilt ok)
[info] Project ensime------------------: SUCCESS (project rebuilt ok)
[info] Project genjavadoc--------------: SUCCESS (unchanged, not rebuilt)
[info] Project genjavadoc-plugin-------: SUCCESS (unchanged, not rebuilt)
[info] Project json4s------------------: SUCCESS (unchanged, not rebuilt)
[info] Project lift-json---------------: SUCCESS (project rebuilt ok)
[info] Project parboiled---------------: SUCCESS (unchanged, not rebuilt)
[info] Project pimpathon---------------: SUCCESS (unchanged, not rebuilt)
[info] Project play-twirl--------------: SUCCESS (project rebuilt ok)
[info] Project Play2-core--------------: SUCCESS (project rebuilt ok)
[info] Project sbinary-----------------: SUCCESS (unchanged, not rebuilt)
[info] Project sbt---------------------: SUCCESS (unchanged, not rebuilt)
[info] Project sbt-republish-----------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-arm---------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-continuations-----: SUCCESS (unchanged, not rebuilt)
[info] Project scala-io----------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-js----------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-logging-api-------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-logging-slf4j-----: SUCCESS (unchanged, not rebuilt)
[info] Project scala-partest-----------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-partest-interface-: SUCCESS (unchanged, not rebuilt)
[info] Project scala-records-----------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-refactoring-------: SUCCESS (project rebuilt ok)
[info] Project scala-stm---------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-swing-------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalacheck--------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalamock---------------: SUCCESS (project rebuilt ok)
[info] Project scalariform-------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalatest---------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalaz------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scoverage---------------: SUCCESS (unchanged, not rebuilt)
[info] Project shapeless---------------: SUCCESS (unchanged, not rebuilt)
[info] Project slick-------------------: SUCCESS (unchanged, not rebuilt)
[info] Project specs2------------------: SUCCESS (unchanged, not rebuilt)
[info] Project spire-------------------: SUCCESS (project rebuilt ok)
[info] Project spray-------------------: SUCCESS (project rebuilt ok)
[info] Project spray-json--------------: SUCCESS (project rebuilt ok)
[info] Project spray-twirl-------------: SUCCESS (project rebuilt ok)
[info] Project zeromq-scala-binding----: SUCCESS (unchanged, not rebuilt)
[info] Project zinc--------------------: SUCCESS (unchanged, not rebuilt)
[info] >>> The dbuild result is------------: SUCCESS (project rebuilt ok)
 
I got it from the bottom of: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/lastStableBuild/consoleText

SethTisue:
@bvenners yes. gory version details at https://github.com/scala/community-builds/blob/2.11.x/common.conf. very gory, since many projects/builds had to be customized or forked. (fsvo "had to" anyway)

bvenners:
@SethTisue, you mentioned javabootclasspath here:

https://issues.scala-lang.org/browse/SUGGEST-5

Have you ever used that in practice to build on a later JDK and target an earlier JDK?

SethTisue:
yes, did it in practice for years on the NetLogo project. we were still supporting Java 5 but Mac OS X only supported Java 6 anymore. it saved us several times. 5/6 is the only version combo I have firsthand experience with this with, but I can't see why it wouldn't work just as well with 6, 7, 8.

iirc correctly we initially were passing `-javabootclasspath` to the compiler to get this safety, but it got annoying to have to have the JDK 5 stdlib around during ordinary development work, so we changed it to do the checking after the fact with ProGuard during release-building only. so, that works too. (we needed ProGuard anyway for other reasons, we didn't bring it on board solely for this purpose, so it was an easy tweak to our ProGuard config.)

bvenners:
Ah, actually I was thinking there should be some way to do this with a test. What you're saying is you use ProGuard after the fact to catch whether you accidentally linked to a method added to a later JDK than what you are targetting? If so I would feel a lot more comfortable about actually doing release builds fully on JDK 8, but still probably will prefer to build on the actual JDK when deploying to sonatype.

I have been thinking about how we'll all have to deal with Scala 2.12. (Sorry, I will get my 2.12.0-M2 build out tonight, hopefully. Have not had any extra time this week.) Once 2.12 becomes a regular part of the ScalaTest build I won't want to drop 2.10 or 2.11 support. Someday I'll drop 2.10 support, but a lot of people are still on it, so it seems premature today. To do development I think we (people who hep with ScalaTest/Scalactic) should just always develop on JDK 8 at that point.

It is indeed a pain to keep old JDK standard libs around, but I think I myself would probably do that on the laptop I personally use for development, and set environment variables that are picked up by the build to point to those libs. But then if those environment variables aren't set, the build will just link to the JDK 8 libs. That way anybody out there who clones the repo can just build on JDK 8. They may add some code that links to JDK-7 or JDK-8-only libraries and submit that in a PR, but maybe we can add the ProGuard thing to Travis and it would be caught when that's build in CI. If not, I'd catch it when I test out the PR.

For deployment of releases right now I have a one-button build that runs on JDK 6. What I was thinking I will probably do is make a script that checks out 2 repos (or 3 repos if Scala.js drops JDK 6 support), and do independent builds in those 2 or 3 repos on different JDKs, deploying a subset of the artifacts from each repo. Everyone will need to do something either like that, or just build on JDK 8 and target back, either pointing to old standard libraries and/or with a ProGuard sanity check.

I won't have time to try that sort of thing for many more weeks, but that is the future I see on the horizon. I'd definitely like to test out the -javabootclasspath trick and ProGuard trick sooner rather than later to make sure they actually work.

SethTisue:
> What you're saying is you use ProGuard after the fact to catch whether you accidentally linked to a method added to a later JDK than what you are targetting

exactly!

sjrd:
Closing this. We're keeping JDK 6 ...

PR #1821, opened 2015-07-24
Fix #1819: Accept any number in isFloat when non-strict floats.

### LINKED ISSUE #1819: Double doesn't match Float even with non-strict Float
```scala

    def test(x: Any) = x match {
       case f: Float => "ok"
    }
    
    test(0.2)
```

This fails with a runtime match error. According to [this documentation](http://www.scala-js.org/doc/semantics.html), with strict-floats disabled (default), a `Double` should match as a `Float`.

sjrd:
When strict-floats are disabled, any number, i.e., any Double, now also matches as a Float, as per the specification.


sjrd:
Review by @nicolasstucki

sjrd:
retest this please

sjrd:
retest this please

nicolasstucki:
LGTM

PR #1825, opened 2015-07-27
Implement java.util.TreeSet.

andreaTP:
implementation of ju.TreeSet

### DIFF
--- a/javalib/src/main/scala/java/util/TreeSet.scala
+++ b/javalib/src/main/scala/java/util/TreeSet.scala
@@ -1,0 +1,245 @@
+package java.util
+
+import java.lang.Comparable
+import scala.collection.mutable
+import scala.math.Ordering
+
+import scala.collection.JavaConversions._
+
+class TreeSet[E] (_comparator: Comparator[_ >: E])
+    extends AbstractSet[E]
+    with NavigableSet[E]
+    with Cloneable
+    with Serializable { self =>
+
+  def this() =
+    this(null.asInstanceOf[Comparator[_ >: E]])
+
+  def this(collection: Collection[_ <: E]) = {
+    this(null.asInstanceOf[Comparator[E]])
+    addAll(collection)
+  }
+
+  def this(sortedSet: SortedSet[E]) = {
+    this(sortedSet.comparator())
+    addAll(sortedSet)
+  }
+
+  private implicit object BoxOrdering extends Ordering[Box[E]] {
+
+    val cmp = {
+      if (_comparator ne null) _comparator
+      else defaultOrdering[E]
+    }
+
+    def compare(a: Box[E], b: Box[E]): Int = cmp.compare(a.inner, b.inner)
+
+  }
+
+  private val inner: mutable.TreeSet[Box[E]] = new mutable.TreeSet[Box[E]]()
+
+  def iterator(): Iterator[E] = {
+    new Iterator[E] {
+      private val iter = inner.clone.iterator
+
+      private var last: Option[E] = None
+
+      def hasNext(): Boolean = iter.hasNext
+
+      def next(): E = {
+        last = Some(iter.next().inner)
+        last.get
+      }
+
+      def remove(): Unit = {
+        if (last.isEmpty) {
+          throw new IllegalStateException()
+        } else {
+          last.foreach(self.remove(_))
+          last = None
+        }
+      }
+    }
+  }
+
+  def descendingIterator(): Iterator[E] = {
+    new Iterator[E] {
+      private val iter = inner.iterator.toList.reverse.iterator
+
+      private var last: Option[E] = None
+
+      def hasNext(): Boolean = iter.hasNext
+
+      def next(): E = {
+        val nxt = iter.next().inner
+        last = Some(nxt)
+        nxt
+      }
+
+      def remove(): Unit = {
+        if (last.isEmpty) {
+          throw new IllegalStateException()
+        } else {
+          last.foreach(self.remove(_))
+          last = None
+        }
+      }
+    }
+  }
+
+  def descendingSet(): NavigableSet[E] = {
+    val descSetFun = { () =>
+      val retSet = new mutable.TreeSet[Box[E]]()(BoxOrdering.reverse)
+      retSet.addAll(inner)
+      retSet
+    }
+    new NavigableView(this, descSetFun, None, true, None, true)
+  }
+
+  def size(): Int =
+    inner.size
+
+  override def isEmpty(): Boolean =
+    inner.headOption.isEmpty
+
+  override def contains(o: Any): Boolean =
+    inner.contains(Box(o.asInstanceOf[E]))
+
+  override def add(e: E): Boolean = {
+    val boxed = Box(e)
+
+    if (isEmpty)
+      BoxOrdering.compare(boxed, boxed)
+
+    inner.add(boxed)
+  }
+
+  override def remove(o: Any): Boolean =
+    inner.remove(Box(o.asInstanceOf[E]))
+
+  override def clear(): Unit =
+    inner.clear()
+
+  override def addAll(c: Collection[_ <: E]): Boolean = {
+    val iter = c.iterator()
+    var changed = false
+    while (iter.hasNext)
+      changed = add(iter.next()) || changed
+    changed
+  }
+
+  override def removeAll(c: Collection[_]): Boolean = {
+    val iter = c.iterator()
+    var changed = false
+    while (iter.hasNext)
+      changed = inner.remove(Box(iter.next).asInstanceOf[Box[E]]) || changed
+    changed
+  }
+
+  def subSet(fromElement: E, fromInclusive: Boolean, toElement: E,
+      toInclusive: Boolean): NavigableSet[E] = {
+    val boxedFrom = Box(fromElement)
+    val boxedTo = Box(toElement)
+    val subSetFun = { () =>
+      // the creation of a new TreeSet is to avoid a mysterious bug with scala 2.10
+      var base = new mutable.TreeSet[Box[E]]
+      base ++= inner.range(boxedFrom, boxedTo)
+      if (!fromInclusive)
+        base = base - boxedFrom
+
+      if (toInclusive && inner.contains(boxedTo))
+        base = base + boxedTo
+
+      base
+    }
+
+    new NavigableView(this, subSetFun,
+        Some(fromElement), fromInclusive,
+        Some(toElement), toInclusive)
+  }
+
+  def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] = {
+    val boxed = Box(toElement)
+    val headSetFun = { () =>
+      // the creation of a new TreeSet is to avoid a mysterious bug with scala 2.10
+      var base = new mutable.TreeSet[Box[E]]
+      if (inclusive)
+        base ++= inner.to(boxed)
+      else
+        base ++= inner.until(boxed)
+
+      base
+    }
+
+    new NavigableView(this, headSetFun,
+        None, true,
+        Some(toElement), inclusive)
+  }
+
+  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] = {
+    val boxed = Box(fromElement)
+    val tailSetFun = { () =>
+      // the creation of a new TreeSet is to avoid a mysterious bug with scala 2.10
+      var base = new mutable.TreeSet[Box[E]]
+      base ++= inner.from(boxed)
+      if (!inclusive)
+        base -= boxed
+
+      base
+    }
+
+    new NavigableView(this, tailSetFun,
+        Some(fromElement), inclusive,
+        None, true)
+  }
+
+  def subSet(fromElement: E, toElement: E): SortedSet[E] =
+    subSet(fromElement, true, toElement, false)
+
+  def headSet(toElement: E): SortedSet[E] =
+    headSet(toElement, false)
+
+  def tailSet(fromElement: E): SortedSet[E] =
+    tailSet(fromElement, true)
+
+  def comparator(): Comparator[_ >: E] = _comparator
+
+  def first(): E =
+    inner.head.inner
+
+  def last(): E =
+    inner.last.inner
+
+  def lower(e: E): E =
+    headSet(e, false).lastOption.getOrElse(null.asInstanceOf[E])
+
+  def floor(e: E): E =
+    headSet(e, true).lastOption.getOrElse(null.asInstanceOf[E])
+
+  def ceiling(e: E): E =
+    tailSet(e, true).headOption.getOrElse(null.asInstanceOf[E])
+
+  def higher(e: E): E =
+    tailSet(e, false).headOption.getOrElse(null.asInstanceOf[E])
+
+  def pollFirst(): E = {
+    val polled = inner.headOption
+    if (polled.isDefined) {
+      val elem = polled.get.inner
+      remove(elem)
+      elem
+    } else null.asInstanceOf[E]
+  }
+
+  def pollLast(): E = {
+    val polled = inner.lastOption
+    if (polled.isDefined) {
+      val elem = polled.get.inner
+      remove(elem)
+      elem
+    } else null.asInstanceOf[E]
+  }
+
+  override def clone(): TreeSet[E] =
+    new TreeSet(this)
+}
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
@@ -2,39 +2,27 @@ package java.util.concurrent
 
 import java.util._
 import java.lang.Comparable
-import scala.collection.mutable
 import scala.math.Ordering
 
 import scala.collection.JavaConversions._
 
-class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparator: Comparator[_ >: E])
+class ConcurrentSkipListSet[E] private (inner: TreeSet[E])
     extends AbstractSet[E]
     with NavigableSet[E]
     with Cloneable
-    with Serializable { self =>
+    with Serializable {
 
   def this() =
-    this(defaultOrdering[E], null.asInstanceOf[Comparator[_ >: E]])
+    this(new TreeSet[E]())
 
   def this(comparator: Comparator[_ >: E]) =
-    this(Ordering.comparatorToOrdering(comparator), null.asInstanceOf[Comparator[E]])
+    this(new TreeSet[E](comparator))
 
-  def this(collection: Collection[_ <: E]) = {
-    this(defaultOrdering[E], null.asInstanceOf[Comparator[E]])
-    addAll(collection)
-  }
+  def this(collection: Collection[_ <: E]) =
+    this(new TreeSet[E](collection))
 
-  def this(sortedSet: SortedSet[E]) = {
-    this(Ordering.comparatorToOrdering(sortedSet.comparator()),
-        sortedSet.comparator())
-    addAll(sortedSet)
-  }
-
-  private implicit object BoxOrdering extends Ordering[Box[E]] {
-    def compare(a: Box[E], b:Box[E]): Int = ordering.compare(a.inner, b.inner)
-  }
-
-  private val inner: mutable.SortedSet[Box[E]] = new mutable.TreeSet[Box[E]]()
+  def this(sortedSet: SortedSet[E]) =
+    this(new TreeSet[E](sortedSet))
 
   override def clone(): ConcurrentSkipListSet[E] =
     new ConcurrentSkipListSet(this)
@@ -47,349 +35,74 @@ class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparato
 
   override def contains(o: Any): Boolean =
     if (o == null) false
-    else inner.contains(Box(o.asInstanceOf[E]))
+    else inner.contains(o)
 
   override def add(e: E): Boolean =
     if (e == null) throw new NullPointerException()
-    else inner.add(Box(e))
+    else inner.add(e)
 
   override def remove(o: Any): Boolean =
     if (o == null) throw new NullPointerException()
-    else inner.remove(Box(o.asInstanceOf[E]))
+    else inner.remove(o)
 
   override def clear(): Unit =
     inner.clear()
 
-  def iterator(): Iterator[E] = {
-    new Iterator[E] {
-      private val iter = inner.clone.iterator
+  def iterator(): Iterator[E] =
+    inner.iterator()
 
-      private var last: Option[E] = None
+  def descendingIterator(): Iterator[E] =
+    inner.descendingIterator()
 
-      def hasNext(): Boolean = iter.hasNext
-
-      def next(): E = {
-        last = Some(iter.next().inner)
-        last.get
-      }
-
-      def remove(): Unit = {
-        if (last.isEmpty) {
-          throw new IllegalStateException()
-        } else {
-          last.foreach(self.remove(_))
-          last = None
-        }
-      }
-    }
-  }
-
-  def descendingIterator(): Iterator[E] = {
-    new Iterator[E] {
-      private val iter = inner.iterator.toList.reverse.iterator
-
-      private var last: Option[E] = None
-
-      def hasNext(): Boolean = iter.hasNext
-
-      def next(): E = {
-        last = Some(iter.next().inner)
-        last.get
-      }
-
-      def remove(): Unit = {
-        if (last.isEmpty) {
-          throw new IllegalStateException()
-        } else {
-          last.foreach(self.remove(_))
-          last = None
-        }
-      }
-    }
-  }
-
-  override def removeAll(c: Collection[_]): Boolean = {
-    val iter = c.iterator()
-    var changed = false
-    while (iter.hasNext)
-      changed = inner.remove(Box(iter.next).asInstanceOf[Box[E]]) || changed
-    changed
-  }
-
-  override def addAll(c: Collection[_ <: E]): Boolean = {
-    val iter = c.iterator()
-    var changed = false
-    while (iter.hasNext)
-      changed = add(iter.next()) || changed
-    changed
-  }
+  override def removeAll(c: Collection[_]): Boolean =
+    inner.removeAll(c)
 
   def lower(e: E): E =
-    headSet(e, false).lastOption.getOrElse(null.asInstanceOf[E])
+    inner.lower(e)
 
   def floor(e: E): E =
-    headSet(e, true).lastOption.getOrElse(null.asInstanceOf[E])
+    inner.floor(e)
 
   def ceiling(e: E): E =
-    tailSet(e, true).headOption.getOrElse(null.asInstanceOf[E])
+    inner.ceiling(e)
 
   def higher(e: E): E =
-    tailSet(e, false).headOption.getOrElse(null.asInstanceOf[E])
+    inner.higher(e)
 
-  def pollFirst(): E = {
-    val polled = inner.headOption
-    if (polled.isDefined) {
-      val elem = polled.get.inner
-      remove(elem)
-      elem
-    } else null.asInstanceOf[E]
-  }
+  def pollFirst(): E =
+    inner.pollFirst()
 
-  def pollLast(): E = {
-    val polled = inner.lastOption
-    if (polled.isDefined) {
-      val elem = polled.get.inner
-      remove(elem)
-      elem
-    } else null.asInstanceOf[E]
-  }
+  def pollLast(): E =
+    inner.pollLast()
 
-  def comparator(): Comparator[_ >: E] = _comparator
+  def comparator(): Comparator[_ >: E] =
+    inner.comparator()
 
   def first(): E =
-    if (inner.isEmpty) throw new NoSuchElementException()
-    else inner.head.inner
+    inner.first
 
   def last(): E =
-    if (inner.isEmpty) throw new NoSuchElementException()
-    else inner.last.inner
+    inner.last
 
-  def subSet(fromElement: E, fromInclusive: Boolean, toElement: E, toInclusive: Boolean): NavigableSet[E] = {
-    val subSetFun = { () =>
-      val boxedFrom = Box(fromElement)
-      val boxedTo = Box(toElement)
-      val ss = new mutable.TreeSet[Box[E]]()
-      ss ++= inner.range(boxedFrom, boxedTo)
-      if (!fromInclusive && inner.contains(boxedFrom))
-        ss.remove(boxedFrom)
-      if (toInclusive && inner.contains(boxedTo))
-        ss += boxedTo
-      ss
-    }
-    new ConcurrentSkipListSet.NavigableView(this, subSetFun)
-  }
+  def subSet(fromElement: E, fromInclusive: Boolean, toElement: E,
+      toInclusive: Boolean): NavigableSet[E] =
+    inner.subSet(fromElement, fromInclusive, toElement, toInclusive)
 
-  def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] = {
-    val headSetFun = { () =>
-      val boxed = Box(toElement)
-      val hs = new mutable.TreeSet[Box[E]]()
-      hs ++= inner.until(boxed)
-      if (inclusive && inner.contains(boxed))
-        hs += boxed
-      hs
-    }
-    new ConcurrentSkipListSet.NavigableView(this, headSetFun)
-  }
+  def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] =
+    inner.headSet(toElement, inclusive)
 
-  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =  {
-    val tailSetFun = { () =>
-      val boxed = Box(fromElement)
-      val ts = new mutable.TreeSet[Box[E]]()
-      ts ++= inner.from(boxed)
-      if (!inclusive && inner.contains(boxed))
-        ts.remove(boxed)
-      ts
-    }
-    new ConcurrentSkipListSet.NavigableView(this, tailSetFun)
-  }
+  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =
+    inner.tailSet(fromElement, inclusive)
 
   def subSet(fromElement: E, toElement: E): NavigableSet[E] =
-    subSet(fromElement, true, toElement, false)
+    inner.subSet(fromElement, true, toElement, false)
 
   def headSet(toElement: E): NavigableSet[E] =
-    headSet(toElement, false)
+    inner.headSet(toElement, false)
 
   def tailSet(fromElement: E): NavigableSet[E] =
-    tailSet(fromElement, true)
+    inner.tailSet(fromElement, true)
 
-  def descendingSet(): NavigableSet[E] = {
-    val descSetFun = { () =>
-      val retSet = new mutable.TreeSet[Box[E]]()(BoxOrdering.reverse)
-      retSet.addAll(inner)
-      retSet
-    }
-    new ConcurrentSkipListSet.NavigableView(this, descSetFun)
-  }
-}
-
-object ConcurrentSkipListSet {
-
-  private class NavigableView[E](original: NavigableSet[E], inner: () => mutable.SortedSet[Box[E]])
-      extends AbstractCollection[E] with NavigableSet[E] with SortedSet[E] {
-
-    def size(): Int =
-      iterator.size
-
-    override def contains(o: Any): Boolean =
-      inner().contains(Box(o.asInstanceOf[E]))
-
-    override def add(e: E): Boolean =
-      original.add(e)
-
-    override def remove(o: Any): Boolean = {
-      if (contains(o))
-        original.remove(o)
-      else false
-    }
-
-    def iterator(): Iterator[E] =
-      inner().iterator.map(_.inner)
-
-    def descendingIterator(): Iterator[E] = {
-      new Iterator[E] {
-        private val iter = iterator.toList.reverse.iterator
-
-        private var last: Option[E] = None
-
-        def hasNext(): Boolean = iter.hasNext
-
-        def next(): E = {
-          last = Some(iter.next())
-          last.get
-        }
-
-        def remove(): Unit = {
-          if (last.isEmpty) {
-            throw new IllegalStateException()
-          } else {
-            last.foreach(original.remove(_))
-            last = None
-          }
-        }
-      }
-    }
-
-    override def removeAll(c: Collection[_]): Boolean = {
-      val iter = c.iterator()
-      var changed = false
-      while (iter.hasNext)
-          changed = remove(iter.next) || changed
-      changed
-    }
-
-    override def addAll(c: java.util.Collection[_ <: E]): Boolean =
-      original.addAll(c)
-
-    def lower(e: E): E =
-      headSet(e, false).lastOption.getOrElse(null.asInstanceOf[E])
-
-    def floor(e: E): E =
-      headSet(e, true).lastOption.getOrElse(null.asInstanceOf[E])
-
-    def ceiling(e: E): E =
-      tailSet(e, true).headOption.getOrElse(null.asInstanceOf[E])
-
-    def higher(e: E): E =
-      tailSet(e, false).headOption.getOrElse(null.asInstanceOf[E])
-
-    def pollFirst(): E = {
-      val polled = inner().headOption
-      if (polled.isDefined) {
-        val elem = polled.get.inner
-        remove(elem)
-        elem
-      } else null.asInstanceOf[E]
-    }
-
-    def pollLast(): E = {
-      val polled = inner().lastOption
-      if (polled.isDefined) {
-        val elem = polled.get.inner
-        remove(elem)
-        elem
-      } else null.asInstanceOf[E]
-    }
-
-    def comparator(): java.util.Comparator[E] = {
-      new Comparator[E] {
-        val ordering = inner().ordering
-        def compare(a: E, b: E): Int =
-          ordering.compare(Box(a), Box(b))
-      }
-    }
-
-    def first(): E = {
-      val iter = iterator()
-      if (iter.hasNext) iter.next
-      else null.asInstanceOf[E]
-    }
-
-    def last(): E = {
-      val iter = iterator()
-      if (iter.hasNext) iter.toTraversable.last
-      else null.asInstanceOf[E]
-    }
-
-    def subSet(fromElement: E, fromInclusive: Boolean, toElement: E, toInclusive: Boolean): NavigableSet[E] = {
-      val innerNow = inner()
-      val subSetFun = { () =>
-        val boxedFrom = Box(fromElement)
-        val boxedTo = Box(toElement)
-        val ss = new mutable.TreeSet[Box[E]]()(innerNow.ordering)
-        ss ++= innerNow.range(boxedFrom, boxedTo)
-        if (!fromInclusive && innerNow.contains(boxedFrom))
-          ss.remove(boxedFrom)
-        if (toInclusive && innerNow.contains(boxedTo))
-          ss += boxedTo
-        ss
-      }
-      new ConcurrentSkipListSet.NavigableView(this, subSetFun)
-    }
-
-    def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] = {
-      val innerNow = inner()
-      val headSetFun = { () =>
-        val boxed = Box(toElement)
-        val hs = new mutable.TreeSet[Box[E]]()(innerNow.ordering)
-        hs ++= innerNow.until(boxed)
-        if (inclusive && innerNow.contains(boxed))
-          hs += boxed
-        hs
-      }
-      new ConcurrentSkipListSet.NavigableView(this, headSetFun)
-    }
-
-    def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =  {
-      val innerNow = inner()
-      val tailSetFun = { () =>
-        val boxed = Box(fromElement)
-        val ts = new mutable.TreeSet[Box[E]]()(innerNow.ordering)
-        ts ++= innerNow.from(boxed)
-        if (!inclusive && innerNow.contains(boxed))
-          ts.remove(boxed)
-        ts
-      }
-      new ConcurrentSkipListSet.NavigableView(this, tailSetFun)
-    }
-
-    def subSet(fromElement: E, toElement: E): NavigableSet[E] =
-      subSet(fromElement, true, toElement, false)
-
-    def headSet(toElement: E): NavigableSet[E] =
-      headSet(toElement, false)
-
-    def tailSet(fromElement: E): NavigableSet[E] =
-      tailSet(fromElement, true)
-
-    def descendingSet(): NavigableSet[E] = {
-      val descSetFun = { () =>
-        val innerNow = inner()
-        val retSet = new mutable.TreeSet[Box[E]]()(innerNow.ordering.reverse)
-        retSet.addAll(innerNow)
-        retSet
-      }
-      new NavigableView(this, descSetFun)
-    }
-  }
-
+  def descendingSet(): NavigableSet[E] =
+    inner.descendingSet()
 }
--- a/javalib/src/main/scala/java/util/package.scala
+++ b/javalib/src/main/scala/java/util/package.scala
@@ -1,7 +1,5 @@
 package java
 
-import scala.math.Ordering
-
 package object util {
 
   implicit private[util] class CompareNullablesOps(val self: Any) extends AnyVal {
@@ -32,5 +30,4 @@ package object util {
         a.asInstanceOf[Comparable[E]].compareTo(b)
     }
   }
-
 }
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/CollectionTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/CollectionTest.scala
@@ -165,7 +165,10 @@ trait CollectionTest extends JasmineTest with ExpectExceptions {
       coll.add("one")
       expect(coll.contains("one")).toBeTruthy
       expect(coll.contains("two")).toBeFalsy
-      expect(coll.contains(null)).toBeFalsy
+      if (factory.allowsNullElementQuery)
+        expect(coll.contains(null)).toBeFalsy
+      else
+        expectThrows[Exception](coll.contains(null))
     }
 
     it("should check contained presence for double corner cases") {
@@ -205,4 +208,5 @@ trait CollectionFactory {
   def implementationName: String
   def empty[E]: ju.Collection[E]
   def allowsMutationThroughIterator: Boolean = true
+  def allowsNullElementQuery: Boolean = true
 }
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/NavigableSetTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/NavigableSetTest.scala
@@ -2,6 +2,121 @@ package org.scalajs.testsuite.javalib
 
 import java.{util => ju}
 
+import scala.collection.JavaConversions._
+
+trait NavigableSetTest extends SetTest {
+
+  def testNavigableSetApi(setFactory: NavigableSetFactory): Unit = {
+  it("should retrieve ceiling(ordered) elements") {
+      val lInt = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val nsInt = setFactory.empty[Int]
+
+      nsInt.addAll(lInt)
+
+      expect(nsInt.ceiling(-10)).toEqual(1)
+      expect(nsInt.ceiling(0)).toEqual(1)
+      expect(nsInt.ceiling(1)).toEqual(1)
+      expect(nsInt.ceiling(5)).toEqual(5)
+
+      val lString = asJavaCollection(Set("a", "e", "b", "c", "d"))
+      val nsString = setFactory.empty[String]
+
+      nsString.addAll(lString)
+
+      expect(nsString.ceiling("00000")).toEqual("a")
+      expect(nsString.ceiling("0")).toEqual("a")
+      expect(nsString.ceiling("a")).toEqual("a")
+      expect(nsString.ceiling("d")).toEqual("d")
+      expect(nsString.ceiling("z")).toEqual(null)
+    }
+
+    it("should retrieve floor(ordered) elements") {
+      val lInt = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val nsInt = setFactory.empty[Int]
+
+      nsInt.addAll(lInt)
+
+      expect(nsInt.floor(10)).toEqual(5)
+      expect(nsInt.floor(5)).toEqual(5)
+      expect(nsInt.floor(3)).toEqual(3)
+      expect(nsInt.floor(1)).toEqual(1)
+
+      val lString = asJavaCollection(Set("a", "e", "b", "c", "d"))
+      val nsString = setFactory.empty[String]
+
+      nsString.addAll(lString)
+
+      expect(nsString.floor("zzzzz")).toEqual("e")
+      expect(nsString.floor("d")).toEqual("d")
+      expect(nsString.floor("b")).toEqual("b")
+      expect(nsString.floor("a")).toEqual("a")
+      expect(nsString.floor("0")).toEqual(null)
+    }
+
+    it("should retrieve higher(ordered) elements") {
+      val lInt = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val nsInt = setFactory.empty[Int]
+
+      nsInt.addAll(lInt)
+
+      expect(nsInt.higher(4)).toEqual(5)
+      expect(nsInt.higher(3)).toEqual(4)
+      expect(nsInt.higher(1)).toEqual(2)
+      expect(nsInt.higher(-10)).toEqual(1)
+
+      val lString = asJavaCollection(Set("a", "e", "b", "c", "d"))
+      val nsString = setFactory.empty[String]
+
+      nsString.addAll(lString)
+
+      expect(nsString.higher("zzzzz")).toEqual(null)
+      expect(nsString.higher("d")).toEqual("e")
+      expect(nsString.higher("b")).toEqual("c")
+      expect(nsString.higher("a")).toEqual("b")
+      expect(nsString.higher("0")).toEqual("a")
+    }
+
+    it("should retrieve lower(ordered) elements") {
+      val lInt = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val nsInt = setFactory.empty[Int]
+
+      nsInt.addAll(lInt)
+
+      expect(nsInt.lower(5)).toEqual(4)
+      expect(nsInt.lower(4)).toEqual(3)
+      expect(nsInt.lower(3)).toEqual(2)
+      expect(nsInt.lower(10)).toEqual(5)
+
+      val lString = asJavaCollection(Set("a", "e", "b", "c", "d"))
+      val nsString = setFactory.empty[String]
+
+      nsString.addAll(lString)
+
+      expect(nsString.lower("zzzzz")).toEqual("e")
+      expect(nsString.lower("d")).toEqual("c")
+      expect(nsString.lower("b")).toEqual("a")
+      expect(nsString.lower("a")).toEqual(null)
+      expect(nsString.lower("0")).toEqual(null)
+    }
+
+    it("should poll first and last elements") {
+      val lInt = asJavaCollection(Set(1, 5, 2, 3, 4))
+      val ns = setFactory.empty[Int]
+
+      ns.addAll(lInt)
+
+      expect(ns.contains(1)).toBeTruthy
+      expect(ns.pollFirst()).toEqual(1)
+      expect(ns.contains(1)).toBeFalsy
+      expect(ns.pollLast()).toEqual(5)
+      expect(ns.pollFirst()).toEqual(2)
+      expect(ns.pollLast()).toEqual(4)
+      expect(ns.pollFirst()).toEqual(3)
+      expect(ns.isEmpty()).toBeTruthy
+    }
+  }
+}
+
 object NavigableSetFactory {
   def allFactories: Iterator[NavigableSetFactory] =
     ConcurrentSkipListSetFactory.allFactories
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
@@ -1,0 +1,397 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.javalib
+
+import java.util
+
+import scala.language.implicitConversions
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable
+
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.{util => ju}
+import ju.TreeSet
+import ju.Comparator
+
+
+object TreeSetTest extends TreeSetTest(new TreeSetFactory) {
+
+  override def testApi(): Unit = {
+    super.testApi()
+
+    it("should check that comparator is always null") {
+      val ts1 = factory.empty[Int]
+
+      expect(ts1.comparator() == null).toBeTruthy
+
+      val ts2 = factory.empty[String]
+
+      expect(ts2.comparator() == null).toBeTruthy
+    }
+  }
+}
+object TreeSetWithNullTest extends TreeSetTest(new TreeSetWithNullFactory) {
+
+  override def testApi(): Unit = {
+    super.testApi()
+
+    it("should check that comparator is never null") {
+      val ts1 = factory.empty[Int]
+
+      expect(ts1.comparator() == null).toBeFalsy
+
+      val ts2 = factory.empty[String]
+
+      expect(ts2.comparator() == null).toBeFalsy
+    }
+  }
+}
+
+class TreeSetTest[F <: TreeSetFactory](treeSetTestFactory: F)
+    extends AbstractSetTest[F](treeSetTestFactory)
+    with SortedSetTest
+    with NavigableSetTest {
+
+  override def testApi(): Unit = {
+    super.testApi()
+    testSortedSetApi(treeSetTestFactory)
+    testNavigableSetApi(treeSetTestFactory)
+
+    it("should store and remove ordered integers") {
+      val ts = treeSetTestFactory.empty[Int]
+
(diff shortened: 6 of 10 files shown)


javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
```diff
@@ -8,10 +8,7 @@ import scala.math.Ordering
 import scala.collection.JavaConversions._
 
 class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparator: Comparator[_ >: E])
-    extends AbstractSet[E]
-    with NavigableSet[E]
-    with Cloneable
-    with Serializable { self =>
+    extends TreeSet[E](ordering, _comparator) {
```
nicolasstucki: This will not work properly. You Can't change the class dependencies of classes that already exist in java. One problem that might occure is:
```scala
val set = new ConcurrentSkipListSet(...)
if (set.isInstanceOf[TreeSet]) {
  // will execute this in on JS
} else {
  // will execute this on JVM
}

File: javalib/src/main/scala/java/util/TreeSet.scala [hunk 2]
```diff
@@ -0,0 +1,222 @@
+package java.util
+
+import java.lang.Comparable
+import scala.collection.mutable
+import scala.math.Ordering
+
+import scala.collection.JavaConversions._
+
+class TreeSet[E] protected (ordering: Ordering[_ >: E], _comparator: Comparator[_ >: E])
```
nicolasstucki: You should not add a new constructor unless it is `private`. Just like with the methods.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    extends AbstractSet[E]
+    with NavigableSet[E]
+    with Cloneable
+    with Serializable { self =>
+
+  def this() =
+    this(defaultOrdering[E], null.asInstanceOf[Comparator[_ >: E]])
+
+  def this(comparator: Comparator[_ >: E]) =
+    this(Ordering.comparatorToOrdering(comparator), null.asInstanceOf[Comparator[E]])
```
nicolasstucki: Shouldn't the second parameter of `this(...)` be `comparator` or `comparator.asInstanceOf[Comparator[E]]`.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  def this(sortedSet: SortedSet[E]) = {
+    this(Ordering.comparatorToOrdering(sortedSet.comparator()),
+        sortedSet.comparator())
+    addAll(sortedSet)
+  }
+
+  private implicit object BoxOrdering extends Ordering[Box[E]] {
+    def compare(a: Box[E], b:Box[E]): Int = ordering.compare(a.inner, b.inner)
```
nicolasstucki: You are using `ordering` only once in this class, you could write `_comparator.compare(a.inner, b.inner)` and don't use `ordering` at all.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  private val inner: mutable.TreeSet[Box[E]] = new mutable.TreeSet[Box[E]]()
+
+  def iterator(): Iterator[E] = {
+    new Iterator[E] {
+      private val iter = inner.clone.iterator
```
nicolasstucki: You should not clone a `mutable.TreeSet`. You could reset the iterator after a `remove` with `inner.iteratorFrom`.

nicolasstucki: Forget that last comment. Just saw the implementation of `mutable.TreeSet` and it is fine like this.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+      retSet.addAll(inner)
+      retSet
+    }
+    new NavigableView(this, descSetFun)
+  }
+
+  def size(): Int =
+    inner.size
+
+  override def isEmpty(): Boolean =
```
nicolasstucki: This is already defined in `AbstractCollection`

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    inner.isEmpty
+
+  override def contains(o: Any): Boolean =
+    if (o == null) false
```
nicolasstucki: `TreeSet` should be able to contain the value `null`.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    else inner.contains(Box(o.asInstanceOf[E]))
+
+  override def add(e: E): Boolean =
+    if (e == null) throw new NullPointerException()
```
nicolasstucki: Same

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    else inner.add(Box(e))
+
+  override def remove(o: Any): Boolean =
+    if (o == null) throw new NullPointerException()
```
nicolasstucki: Same

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+  def headSet(toElement: E): NavigableSet[E] =
+    headSet(toElement, false)
+
+  def tailSet(fromElement: E): NavigableSet[E] =
+    tailSet(fromElement, true)
+
+  def comparator(): Comparator[_ >: E] = _comparator
+
+  def first(): E =
+    if (inner.isEmpty) throw new NoSuchElementException()
```
nicolasstucki: You don't need to check and throw the exception, `inner.head.inner` will do it for you.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    else inner.head.inner
+
+  def last(): E =
+    if (inner.isEmpty) throw new NoSuchElementException()
```
nicolasstucki: Same

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,222 @@
...
+    else inner.last.inner
+
+  def lower(e: E): E =
+    headSet(e, false).lastOption.getOrElse(null.asInstanceOf[E])
```
nicolasstucki: You could use `headSet(e, false).lastOption.orNull`.

javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
andreaTP: ok, I will use a ```ju.TreeSet``` as inner implementation

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: sure!!

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: _comparator is ```null``` when is not provided, so I must have to explicit ordering, one for the user API and one internal...

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: In this case I want to avoid to call ```size``` which can be expensive operation on tree, instead the check of direct leaf of the root node could be cheaper

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok, also fixed the test

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: yep ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: Doesn't compile:
```
reeSet.scala:186: error: Cannot prove that Null <:< E.
[error]     headSet(e, false).lastOption.orNull//.getOrElse(null.asInstanceOf[E])
[error]                                  ^
[error] one error found
```

javalib/src/main/scala/java/util/TreeSet.scala
nicolasstucki: You could assign the natural ordering comparator singleton instance instead of `null`. 

```scala
new Comparator[T] {
  def compare(o1: T, o2: T): Int = o1.asInstanceOf[Comparable[T]].compareTo(o2)
}
```

javalib/src/main/scala/java/util/TreeSet.scala
nicolasstucki: good :)

javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
```diff
@@ -47,349 +35,76 @@ class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparato
...
-      val hs = new mutable.TreeSet[Box[E]]()
-      hs ++= inner.until(boxed)
-      if (inclusive && inner.contains(boxed))
-        hs += boxed
-      hs
-    }
-    new ConcurrentSkipListSet.NavigableView(this, headSetFun)
-  }
-
-  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =  {
```
nicolasstucki: 2 spaces in `=  {`

javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
```diff
@@ -47,349 +35,76 @@ class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparato
...
-        val hs = new mutable.TreeSet[Box[E]]()(innerNow.ordering)
-        hs ++= innerNow.until(boxed)
-        if (inclusive && innerNow.contains(boxed))
-          hs += boxed
-        hs
-      }
-      new ConcurrentSkipListSet.NavigableView(this, headSetFun)
-    }
-
-    def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =  {
```
nicolasstucki: 2 spaces in `=  {`

javalib/src/main/scala/java/util/package.scala
```diff
@@ -33,4 +35,172 @@ package object util {
...
+        val hs = new mutable.TreeSet[Box[E]]()(innerNow.ordering)
+        hs ++= innerNow.until(boxed)
+        if (inclusive && innerNow.contains(boxed))
+          hs += boxed
+        hs
+      }
+      new NavigableView(this, headSetFun)
+    }
+
+    def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] =  {
```
nicolasstucki: 2 spaces in `=  {`

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+    new Iterator[E] {
+      private val iter = inner.iterator.toList.reverse.iterator
+
+      private var last: Option[E] = None
+
+      def hasNext(): Boolean = iter.hasNext
+
+      def next(): E = {
+        last = Some(iter.next().inner)
+        last.get
```
nicolasstucki: It would be best to fetch the next element only once. You are getting it from `iter.next().inner` and then again from `last.get`. Would be a little bit better with:
```scala
val nxt = iter.next().inner
last = Some(nxt)
nxt
```

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: I need to return ```null``` as comparator when natural ordering is used  as it is the expected behaviour in jvm

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+          throw new IllegalStateException()
+        } else {
+          last.foreach(self.remove(_))
+          last = None
+        }
+      }
+    }
+  }
+
+  def descendingIterator(): Iterator[E] = {
```
nicolasstucki: You could simply use `descendingSet.iterator`.

javalib/src/main/scala/java/util/TreeSet.scala
sjrd: No `null` is actually required to be able to return `null` from the `comparator()` method. But what *is* definitely possible is to have a test in here, testing for `null`:
```scala
def compare(a: Box[E], b: Box[E]): Int =
  if (_comparator != null) _comparator.compare(a, b)
  else a.asInstanceOf[Comparable[Box[E]].compareTo(b)
```

javalib/src/main/scala/java/util/package.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  def descendingSet(): NavigableSet[E] = {
+    val descSetFun = { () =>
+      val retSet = new mutable.TreeSet[Box[E]]()(BoxOrdering.reverse)
+      retSet.addAll(inner)
+      retSet
+    }
+    new NavigableView(this, descSetFun)
```
nicolasstucki: Giving this function to the `NavigableView` is extremely bad performance wise. For example each call to `contains(...)` will create a new `retSet` each time which will take `O(n)`.

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ```NavigableSet``` is always backed by original collection, instead ```Iterator``` is a snapshot of the collection at the time of creation

javalib/src/main/scala/java/util/TreeSet.scala
nicolasstucki: It might be easier for this method to return a `new NavigableSet` and implement the methods directly on inner.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,436 @@
...
+import scala.collection.mutable
+
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.{util => ju}
+import ju.TreeSet
+
+object TreeSetTest extends JasmineTest {
```
nicolasstucki: To get a better test coverage do this.
```scala
object TreeSetTest extends TreeSetTest(new TreeSetTestFactory)
class HashSetTest[F <: TreeSetTestFactory](treeSetTestFactory: F) extends AbstractSetTest[F](treeSetTestFactory) with SortedSetTest with NavigableSetTest {
  override def testApi(): Unit = {
    super.testApi()
    testSortedSetApi(treeSetTestFactory)
    testNavigableSetApi(treeSetTestFactory)
   
    it("should ... ") {
      // Tests specific to TreeSet
    }
    ...
  }
}
```
You will need to fill the tests in `SortedSetTest` and `NavigableSetTest` (this last one does not exist yet). You will have to move some of the tests from this class to one of those if it is testing that API.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,436 @@
...
+  }
+
+}
+
+object TreeSetFactory extends TreeSetFactory {
+  def allFactories: Iterator[TreeSetFactory] =
+    Iterator(new TreeSetFactory)
+}
+
+class TreeSetFactory extends NavigableSetFactory {
```
nicolasstucki:
```scala
class TreeSetFactory extends AbstractSetFactory with NavigableSetFactory with SortedSetFactory { ...
```

javalib/src/main/scala/java/util/TreeSet.scala
nicolasstucki: I can't see anywhere in the documentation that the `Iterator` must be on a snapshot. Usually it assumes that there are no modifications directly on the collection while iterating on it except for the ones done by the iterator (like `remove`). There are a few exceptions to this but they are usually quite explicitly stated in the documentation (for example http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html)

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok, I like this, thanks guys

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: do not care, ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html#iterator()

```
The returned iterator provides a snapshot of the state of the list when the iterator was constructed. No synchronization is needed while traversing the iterator. The iterator does NOT support the remove method.
```

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: I had the very same discussion with @sjrd 
We found that this is terrible in terms of performances but the only kind of "reliable" implementation to full-fill the specification.

Hope in future we could have a better implementation

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+      hs
+    }
+    new NavigableView(this, headSetFun)
+  }
+
+  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] = {
+    val tailSetFun = { () =>
+      val boxed = Box(fromElement)
+      val ts = new mutable.TreeSet[Box[E]]()
+      ts ++= inner.from(boxed)
```
nicolasstucki: No need to create an additional intermediate set: `ts ++= inner.iteratorFrom(boxed)`.

javalib/src/main/scala/java/util/TreeSet.scala
nicolasstucki: In fact `scala.collection.mutable.TreeSet.isEmpty` is defined in `SeqLike` as `size == 0`, so it would do exactly the same. You could do `inner.firstOption.isEmpty` to get it in `log(n)`.

andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: it needs to be a copy, so, I need a new TreeSet

andreaTP:
please, review this now

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+      val boxedFrom = Box(fromElement)
+      val boxedTo = Box(toElement)
+      val ss = new mutable.TreeSet[Box[E]]()
+      ss ++= inner.range(boxedFrom, boxedTo)
+      if (!fromInclusive && inner.contains(boxedFrom))
+        ss.remove(boxedFrom)
+      if (toInclusive && inner.contains(boxedTo))
+        ss += boxedTo
+      ss
+    }
```
nicolasstucki: I found a solution that with help with the complexity problems of `subSet`, `headSet `, `tailSet`, ...
The the idea is to use `var inner: immutable.TreeSet`, update it like `inner = inner + Box(e)` rather than `inner.add(Box(e))`, which is similar to what the `mutable.TreeSet` does. 

With this we can create the functions for the `NavigableView` using `from`, `to`, `until`,  `head` and `tail` (`from` and `tail` to have the exclusive from) you can create all these view snapshot in `O(log(n))` instead of `O(n)`.

nicolasstucki:

nicolasstucki:

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: I think I don't catch the point sorry,
what about removals done from a a listIterator on the original collection?
and what about propagating modifications to all linked views?

The fact that is in testing means that we can do that later on?

nicolasstucki: Actually the fact that it is testing was to test that I could lunch tests, because last time I wasn't able.

nicolasstucki: The point is to reduce the complexity of every access to `inner` in `NavigableView` which requires a call to the function `NavigableView.inner`. Changes are propagated because on each access the collection is still recomputed but in a faster way. For the iterator and list iterator you should just reflect the changes to the `this` in `new NavigableView(this, ...)`. 


The code would be like:

```scala
def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] = {
  val headSetFun = 
    if (inclusive) () => inner.to(toElement)
    else () => inner.until(toElement)
  new NavigableView(this, headSetFun)
}
```

javalib/src/main/scala/java/util/package.scala
```diff
@@ -29,7 +31,180 @@ package object util {
...
+    override def contains(o: Any): Boolean =
+      inner().contains(Box(o.asInstanceOf[E]))
+
+    override def add(e: E): Boolean =
+      original.add(e)
+
+    override def remove(o: Any): Boolean = {
+      if (contains(o))
+        original.remove(o)
+      else false
```
nicolasstucki: Isn't this equivalent to:
```scala 
override def remove(o: Any): Boolean = original.remove(o)
```

javalib/src/main/scala/java/util/package.scala
```diff
@@ -29,7 +31,180 @@ package object util {
...
+    }
+
+    def iterator(): Iterator[E] =
+      inner().iterator.map(_.inner)
```
nicolasstucki: You need to propagate the calls from remove to `original` just like you did with the `descendingIterator`.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  override def removeAll(c: Collection[_]): Boolean = {
+    val iter = c.iterator()
+    var changed = false
+    while (iter.hasNext)
+      changed = inner.remove(Box(iter.next).asInstanceOf[Box[E]]) || changed
+    changed
+  }
+
+  def subSet(fromElement: E, fromInclusive: Boolean, toElement: E, toInclusive: Boolean): NavigableSet[E] = {
```
nicolasstucki: Line is too long

javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
```diff
@@ -47,349 +35,76 @@ class ConcurrentSkipListSet[E] protected (ordering: Ordering[_ >: E], _comparato
...
-      ts ++= inner.from(boxed)
-      if (!inclusive && inner.contains(boxed))
-        ts.remove(boxed)
-      ts
-    }
-    new ConcurrentSkipListSet.NavigableView(this, tailSetFun)
-  }
+    inner.last
+
+  def subSet(fromElement: E, fromInclusive: Boolean, toElement: E, toInclusive: Boolean): NavigableSet[E] =
```
nicolasstucki: Line is too long

javalib/src/main/scala/java/util/package.scala
```diff
@@ -29,7 +31,180 @@ package object util {
...
+        if (a === null) {
+          if (b === null) 0
+          else -1
+        } else {
+          a.asInstanceOf[Comparable[E]].compareTo(b)
+        }
+    }
+  }
+
+  private[util] class NavigableView[E](original: NavigableSet[E], inner: () => mutable.SortedSet[Box[E]])
```
nicolasstucki: Line is too long

javalib/src/main/scala/java/util/package.scala
```diff
@@ -29,7 +31,180 @@ package object util {
...
+      else null.asInstanceOf[E]
+    }
+
+    def last(): E = {
+      val iter = iterator()
+      if (iter.hasNext) iter.toTraversable.last
+      else null.asInstanceOf[E]
+    }
+
+    def subSet(fromElement: E, fromInclusive: Boolean, toElement: E, toInclusive: Boolean): NavigableSet[E] = {
```
nicolasstucki: Line is too long

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }
+}
+
+object TreeSetFactory extends TreeSetFactory {
+  def allFactories: Iterator[TreeSetFactory] =
+    Iterator(new TreeSetFactory)
+}
+
+class TreeSetFactory extends AbstractSetFactory with NavigableSetFactory with SortedSetFactory {
```
nicolasstucki: Line is too long

nicolasstucki: That is all

nicolasstucki:
Some tests from the testSuite are failing. You should run `> testSuite/test`, see which ones and try to fix those.

andreaTP:
ok, I'll finish the work this evening

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: the "view" set must be a copy...

javalib/src/main/scala/java/util/package.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: checking your proposal

javalib/src/main/scala/java/util/package.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: found something that could work, please review on my next commit

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ConcurrentSkipListSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/package.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

andreaTP:
@nicolasstucki  please review here

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+    val subSetFun = { () =>
+      val toTs = 
+        if (toInclusive) inner.to(boxedTo)
+        else inner.until(boxedTo)
+
+      if (fromInclusive) toTs.from(boxedFrom)
+      else toTs.from(boxedFrom) - boxedFrom
+    }
+
+    new NavigableView(this, subSetFun)
```
nicolasstucki: This `NavigableSet` should throw an `IllegalArgumentException` when there is an access outside of the bounds. It might just be a matter of overriding the `add` method.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  def headSet(toElement: E, inclusive: Boolean): NavigableSet[E] = {
+    val boxed = Box(toElement)
+    val headSetFun =
+      if (inclusive) () => inner.to(boxed)
+      else () => inner.until(boxed)
+
+    new NavigableView(this, headSetFun)
```
nicolasstucki: Same

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  def tailSet(fromElement: E, inclusive: Boolean): NavigableSet[E] = {
+    val boxed = Box(fromElement)
+    val tailSetFun = 
+      if (inclusive) () => inner.from(boxed)
+      else () => inner.from(boxed) - boxed
+
+    new NavigableView(this, tailSetFun)
```
nicolasstucki: Same

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  def subSet(fromElement: E, toElement: E): NavigableSet[E] =
```
nicolasstucki: `def subSet(fromElement: E, toElement: E): SortedSet[E] =`

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+    subSet(fromElement, true, toElement, false)
+
+  def headSet(toElement: E): NavigableSet[E] =
```
nicolasstucki: Same, must return a `SortedSet`

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,212 @@
...
+    headSet(toElement, false)
+
+  def tailSet(fromElement: E): NavigableSet[E] =
```
nicolasstucki: Same, must return a `SortedSet`

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

andreaTP:
here you are

nicolasstucki:
@andreaTP it LGTM except for the missing bound checks in `subSet` and related methods. Please also add some tests for this.

andreaTP:
I added the test at 
https://github.com/andreaTP/scala-js/blob/treeSetImpl/javalib/src/main/scala/java/util/package.scala#L53
is this that you mean?

In case I have just to add related tests

nicolasstucki:
It is almost that. You need to define the bounds without using the elements currently in the set. You have:
```scala
override def add(e: E): Boolean = {
  val comp = comparator()
  if (comp.compare(e, first) >= 0 && comp.compare(e, last) <= 0)
    original.add(e)
  else
}
```
This would not work in the case:
```scala
val s = new TreeSet[Int](Seq(2, 3, 6))
val s2 = s.headSet(5, false) // s2 == TreeSet(2, 3) and bounded by 5
s2.add(4) // would throw the exception in your case but should work
s2.add(5) // throws IllegalArgumentException 
```

A solution would be to create a new class for bounded views that would look like:
```scala
private[util] class BoundedNavigableView[E](original: NavigableSet[E],
      inner: () => mutable.SortedSet[Box[E]], 
      lowerBound: Option[E], lowerInclusive: Boolean,
      upperBound: Option[E], upperInclusive: Boolean)
      extends NavigableView[E](original, inner) {

    override def add(e: E): Boolean = {
      val comp = comparator()
      lowerBound.foreach { bound =>
        val cmp = comp.compare(e, bound)
        if (cmp < 0 || (!lowerInclusive && cmp==0))
      }
      upperBound.foreach { bound =>
        val cmp = comp.compare(e, bound)
        if (cmp > 0 || (!upperInclusive && cmp==0))
      }
      original.add(e)
    }
}
```
The `subSet`, `headSet` and `tailSet` also require a bound check on the new bounds. You shouldn't be able to lower the lower bound or increase the upper bound.

andreaTP:
Inserted in NavigableSet directly, just because can be disabled using ```None```s properly.

andreaTP:
added also tests for that

andreaTP:
?

sjrd:
We are mostly unavailable these days. Expect delays ;-)

nicolasstucki:
Jenkins, test this please

nicolasstucki:
@andreaTP some of the tests for `TreeSet` are failing.
```
...
[info] java.util.TreeSet
[info]   + should store strings
[info]   + should store integers
[info]   + should store doubles
[info]   + should store custom objects
[info]   + should remove stored elements
[info]   + should remove stored elements on double corner cases
[info]   + should be cleared with one operation
[info]   + should check contained presence
[info]   + should check contained presence for double corner cases
[info]   + should give proper iterator over elements
[info]   + should check Set size
[info]   + should store integers
[info]   + should store objects with same hashCode but different types
[info]   + should store doubles also in corner cases
[info]   + should store custom objects
[info]   + should remove stored elements
[info]   + should be cleared with one operation
[info]   + should check contained elems presence
[info]   + should put a whole Collection into
[info]   + should iterate over elements
[info]   + should retrieve the first(ordered) element
[info]   + should retrieve the last(ordered) element
[info]   + should return a proper headSet
[info]   + should return a proper tailSet
[error]  x should return a proper subSet
[error]     Expected false to be truthy.
[error]     Expected false to be truthy.
[error]     Expected true to be falsy.
[info]   + should retrieve ceiling(ordered) elements
[info]   + should retrieve floor(ordered) elements
[error]  x should retrieve higher(ordered) elements
[error]     Expected 5 to equal 4.
[error]     Expected 2 to equal 1.
[error]     Expected 'c' to equal 'b'.
[error]     Expected 'c' to equal 'a'.
[info]   + should retrieve lower(ordered) elements
[info]   + should poll first and last elements
[info]   + should store and remove ordered integers
[info]   + should store and remove ordered strings
[info]   + should store objects with custom comparables
[info]   + should store ordered Double even in corner cases
[info]   + could be instantiated with a prepopulated Collection
[info]   + should be cleared in a single operation
[info]   + should add multiple element in one operation
[info]   + should check contained values even in double corner cases
[info] 
[error] Total for suite java.util.TreeSet
[error] 339 specs, 7 failure
...
```

andreaTP:
WOW, it retrieves different results with scala 2.10.2 !!!
... I have to figure out what happens here...

nicolasstucki:
I hate those kinds of bugs :)
You should also try use 2.10.5 rather than 2.10.2

andreaTP:
good point!

andreaTP:
it starts working from Scala 2.11.0
will check it later...

nicolasstucki:
You will also have to join a all these commits into one.

andreaTP:
yes... as soon as I can fix scala 2.10 ... :-S

nicolasstucki:
Yes, no pressure on that one ;). But it would be good if you push it with the fix, so if then it passes all tests I can merge it at one.

andreaTP:
good!

andreaTP:
This is a very lazy fix, hope that is good for you!

nicolasstucki:
What did you do to fix it?

test-suite/src/test/scala/org/scalajs/testsuite/javalib/NavigableSetTest.scala
```diff
@@ -2,9 +2,124 @@ package org.scalajs.testsuite.javalib
...
+      expect(ns.pollFirst()).toEqual(3)
+      expect(ns.isEmpty()).toBeTruthy
+    }
+  }
+}
+
 object NavigableSetFactory {
   def allFactories: Iterator[NavigableSetFactory] =
-    ConcurrentSkipListSetFactory.allFactories
+    Iterator.empty
```
nicolasstucki: Why did you remove the `ConcurrentSkipListSet` from the tests?

andreaTP:
placed the implementation of TreeSet in scala 2.11 under scalalib/overrides

test-suite/src/test/scala/org/scalajs/testsuite/javalib/NavigableSetTest.scala
andreaTP: Not sure, but I think it was because it is just tested in other ways...

nicolasstucki: You should add it back, this will what happens with the `ConcurrentSkipListSet` when passed to the methods of `java.util.Collections`. This will cover more complex test cases and potentially show bug that happen only with specific combinations of operations.

nicolasstucki:
That is right, the implementation of red black trees changed from 2.10 to 2.11. Could you just add the overrides in a separate commit before the rest with a message like "Add overrides for Scala 2.10"

andreaTP:
argh... trying to figure out how to do it before other commits... :-S

nicolasstucki:
You can remove it from the current commit, the add it to a new commit at the end and finally reorder the commits with `git rebase -i` ;)

test-suite/src/test/scala/org/scalajs/testsuite/javalib/NavigableSetTest.scala
andreaTP: good :-) I had an error in addAll reimplementation, catch and fixed :-)

andreaTP:
ok gitmaster ! :-)

doing it

andreaTP:
how does it looks?

nicolasstucki:
The rest LGTM

nicolasstucki:
Jenkins, test this please

andreaTP:
spurious?
one single server said:
```Build timed out (after 30 minutes)```

**Comment by nicolasstucki** [comment]
Jenkins, test this please

**Comment by sjrd** [comment]
I don't understand the overrides thing. You're not supposed to take something from 2.11 and patch it for 2.10. You would be breaking the meaning of 2.10.

Also, how come you even need this? I find that override very suspicious.

File: javalib/src/main/scala/java/util/package.scala [hunk 37]
```diff
@@ -29,7 +31,194 @@ package object util {
   private[util] def defaultOrdering[E]: Ordering[E] = {
     new Ordering[E] {
       def compare(a: E, b: E): Int =
-        a.asInstanceOf[Comparable[E]].compareTo(b)
+        if (a === null) {
+          if (b === null) 0
+          else -1
+        } else {
+          a.asInstanceOf[Comparable[E]].compareTo(b)
+        }
```
sjrd: This is suspicious too. This ordering is not symmetric. `null < foo` returns -1 but `foo < null` throws.

Why is this needed in the first place? JDK collections typically specify that the natural ordering does not support `null` altogether.

andreaTP: ordering itself is not symmetric, do you mean that I have to check also the other side?
TreeSet allow ```null``` values.

andreaTP:
@sjrd the scala 2.10 override is because I have an hidden problem with the scala TreeSet, probably a bug in the 2.10 implementation and the behaviour throws errors in 2.10 tests but not in 2.11.
This was a quick fix, if really needed, I will dig deeper.

javalib/src/main/scala/java/util/package.scala
sjrd: It allows `null` but *not* with the *natural ordering*. I am pretty sure, at least. It only allows `null` if you give it a comparator that supports nulls.

sjrd:
Hum yes I would very much prefer a fix that does not result in a bastard scala library that will affect every other code.

javalib/src/main/scala/java/util/package.scala
andreaTP: something like 
```
    if (a === null) {
         if (b === null) 0
          else -1
        } else if (b === null) {
          1
        } else {
         a.asInstanceOf[Comparable[E]].compareTo(b)
        }
```
sounds better?

andreaTP: with a proper indentation of course XD

andreaTP:
...arg.... ok, hope this will not take too much

andreaTP:
could I have a jenkins test please on this, I'm no more able to reproduce the problem in my env.

javalib/src/main/scala/java/util/package.scala
nicolasstucki: @andreaTP, I tested it on the JVM and @sjrd is right, the default comparator doesn't null values.

```scala
scala> val set = new java.util.TreeSet[String]
set: java.util.TreeSet[String] = []

scala> set.add("a")
res0: Boolean = true

scala> set.add(null)
java.lang.NullPointerException
	at java.util.TreeMap.put(TreeMap.java:563)
	at java.util.TreeSet.add(TreeSet.java:255)
....
```

```scala
scala> val set = new java.util.TreeSet[String](new java.util.Comparator[String] { def compare(a: String, b: String) = 0 })
set: java.util.TreeSet[String] = []

scala> set.add(null)
res0: Boolean = true

scala> set
res8: java.util.TreeSet[String] = [null]
```

Therefore you can leave it as:
```scala
def compare(a: E, b: E): Int =
  a.asInstanceOf[Comparable[E]].compareTo(b)
```

But then the tests will probably fail because `TreeSetFactory.allowsNullElement` is set to `true` and this should only be the case if the comparator allows `null`. Therefore leave `allowsNullElement` as false. 
You will also need to test if it works with custom `Comparator` that allows `null` elements (like the last one you posted).  You can just create a another version of the `TreeSetFactory` that tests this (or add some parameters to `TreeSetFactory`).

andreaTP: :-S ok...

andreaTP: because code is better than words,
check the last commit changes,
I think this could be acceptable...

andreaTP:
review the last commit please

nicolasstucki:
Now you are not testing the default ordering, only the new one.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+    }
+  }
+}
+
+object TreeSetFactory extends TreeSetFactory {
+  def allFactories: Iterator[TreeSetFactory] =
+    Iterator(new TreeSetFactory)
+}
+
+class TreeSetFactory extends AbstractSetFactory with NavigableSetFactory
```
nicolasstucki:
```scala
object TreeSetFactory extends TreeSetFactory {
  def allFactories: Iterator[TreeSetFactory] =
    Iterator(new TreeSetFactory, new TreeSetWithNullFactory)
}

class TreeSetFactory extends AbstractSetFactory with NavigableSetFactory {
  def implementationName: String = "java.util.TreeSet"
  def empty[E]: ju.TreeSet[E] =TreeSet[E]()
  override def allowsNullElement: Boolean = true
}

class TreeSetWithNullFactory extends TreeSetFactory {
  override def empty[E]: ju.TreeSet[E] =
    TreeSet[E](EvenNullComp[E]())
  override def allowsNullElement: Boolean = false
}
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.runtime.UndefinedBehaviorError
+
+import java.{util => ju}
+import ju.TreeSet
+import ju.Comparator
+
+
+object TreeSetTest extends TreeSetTest(new TreeSetFactory)
```
nicolasstucki:
```scala
object TreeSetTest extends TreeSetTest(new TreeSetFactory)
object TreeSetWithNullTest extends TreeSetTest(new TreeSetWithNullFactory)
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+    with SortedSetTest
+    with NavigableSetTest {
+
+  override def testApi(): Unit = {
+    super.testApi()
+    testSortedSetApi(treeSetTestFactory)
+    testNavigableSetApi(treeSetTestFactory)
+
+    it("should store and remove ordered integers") {
+      val ts = new TreeSet[Int]()
```
nicolasstucki: Use the factory to create the instances.
```scala
  val ts = treeSetTestFactory.empty[Int]
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+      expect {
+        try {
+          ts.first
+          false
+        } catch {
+          case _: NoSuchElementException => true
+          case _: Throwable => false
+        }
+      }.toBeTruthy
+    }
```
nicolasstucki: Should test the insertion and removal of `null` if they are allowed

```scala
if (treeSetTestFactory.allowsNullElement) {
  expect( ...
}
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+      expectThrows[IllegalArgumentException](ss2.add(5))
+    }
+
+    it("should throws exception in case of null elements and default ordering") {
+      val hs = new TreeSet[String]()
+
+      expect(hs.add("ONE")).toBeTruthy
+      expect(hs.contains("ONE")).toBeTruthy
+      expect(hs.contains("TWO")).toBeFalsy
+      expectThrows[NullPointerException](hs.add(null))
```
nicolasstucki: This will be pass if `treeSetTestFactory.allowsNullElement` is `false`, test also the other case.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    it("should not put a whole Collection with null elements into") {
+      val l = List[String]("ONE", "TWO", (null: String))
+      val ts1 = new TreeSet[String]()
+
+      expectThrows[NullPointerException] {
```
nicolasstucki: Same.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,219 @@
...
+
+  def this(sortedSet: SortedSet[E]) = {
+    this(sortedSet.comparator())
+    addAll(sortedSet)
+  }
+
+  private implicit object BoxOrdering extends Ordering[Box[E]] {
+    def compare(a: Box[E], b:Box[E]): Int = {
+      if (_comparator ne null) _comparator.compare(a.inner, b.inner)
+      else defaultOrdering[E].compare(a.inner, b.inner)
```
nicolasstucki: Here you are creating a new `defaultOrdering` instance each time you compare this might be a performance killer. Additionally you are checking on each test `if (_comparator ne null)` which could be done once on the creation of the object. You could solve this with: 

```scala
private implicit object BoxOrdering extends Ordering[Box[E]] {
  val cmp = {
    if (_comparator ne null) _comparator
    else defaultOrdering[E]
  }
  def compare(a: Box[E], b:Box[E]): Int = cmp.compare(a.inner, b.inner)
}
```

nicolasstucki:
Thats all.

javalib/src/main/scala/java/util/package.scala
andreaTP: ok, your suggestion breaks tests at line 168 of CollectionTest, obviously...

andreaTP: I try to go back to my previous proposal

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

andreaTP:
just a note before forgetting it in TreeSetTest.scala lines around 102
```
    when("compliant-asinstanceofs").
    it("should throw exception on non comparable objects") {
      case class TestObj(num: Int)
      
      val ts = treeSetTestFactory.empty[TestObj]
```
[error] file:/Users/andrea/Documents/workspace/temp2/scala-js/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala(106:16:ApplyStatically): ClassType(Lorg_scalajs_testsuite_javalib_AbstractSetTest) expected but ClassType(Lorg_scalajs_testsuite_javalib_TreeSetTest$$anonfun$testApi$3) found for tree of type org.scalajs.core.ir.Trees$This
[debug] Linker: cache stats: reused: 5574 -- invalidated: 20 -- trees read: 20
java.lang.RuntimeException: There were 1 IR checking errors.
	at scala.sys.package$.error(package.scala:27)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$link$2.apply$mcV$sp(Linker.scala:119)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$link$2.apply(Linker.scala:116)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$link$2.apply(Linker.scala:116)
	at org.scalajs.core.tools.optimizer.package$.logTime(package.scala:19)
	at org.scalajs.core.tools.optimizer.Linker.link(Linker.scala:115)
	at org.scalajs.core.tools.optimizer.Linker.link(Linker.scala:74)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:243)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:234)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)


with the case class outside as you will see in the next commit it is ok

javalib/src/main/scala/java/util/TreeSet.scala
andreaTP: ok

andreaTP:
ok, review this

sjrd:
Keeping a reference to a discussion on a commit so it doesn't get lost when that commit is rewritten:
https://github.com/andreaTP/scala-js/commit/717eb8dd3acc2e02e50b6ffc35230a5c5932fe0a#commitcomment-12669025

javalib/src/main/scala/java/util/package.scala
```diff
@@ -29,7 +31,196 @@ package object util {
...
       def compare(a: E, b: E): Int =
-        a.asInstanceOf[Comparable[E]].compareTo(b)
+        if (a === null) {
+          if (b === null) 0
+          else -1
+        } else if (b === null) {
+          1
+        } else {
+          a.asInstanceOf[Comparable[E]].compareTo(b)
+        }
```
sjrd: So this change should be reverted to just `a.asInstanceOf[Comparable[E]].compareTo(b)` as it was before, with the change to `add()` that I suggested.

nicolasstucki: @andreaTP, you did revert this in commit 717eb8d but then in 52af608 (the last one) you undid the change.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  private implicit object BoxOrdering extends Ordering[Box[E]] {
+
+    val cmp = {
+      if (_comparator ne null) _comparator
+      else defaultOrdering[E]
+    }
+
+    def compare(a: Box[E], b:Box[E]): Int = cmp.compare(a.inner, b.inner)
```
nicolasstucki: Missing space. `b:Box[E]` should be `b: Box[E]`.

andreaTP: ok

javalib/src/main/scala/java/util/package.scala
andreaTP: as discussed in the other comment with @sjrd

andreaTP:
Hope we can agree on this

javalib/src/main/scala/java/util/package.scala
```diff
@@ -33,4 +35,186 @@ package object util {
     }
   }
 
+  private[util] class NavigableView[E](original: NavigableSet[E],
```
sjrd: This class should be in its own file. There is no reason to put it in the package object.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/CollectionTest.scala
```diff
@@ -165,7 +165,7 @@ trait CollectionTest extends JasmineTest with ExpectExceptions {
       coll.add("one")
       expect(coll.contains("one")).toBeTruthy
       expect(coll.contains("two")).toBeFalsy
-      expect(coll.contains(null)).toBeFalsy
+      //expect(coll.contains(null)).toBeFalsy
```
sjrd: Do not comment out code.

Either remove it, or execute this line conditionally depending on whether or not the CollectionFactory supports `null` values (better).

sjrd: The latter probably requires adding a new `def acceptsNullValues: Boolean` in `CollectionFactory`.

nicolasstucki: I would rather add a new condition `def allowsNullElementQuery: Boolean` because until now the ported collections that did not support the insertion of `null` elements did return `false` on `.contains(null)`. We should not stop testing and hence a new condition is required.

nicolasstucki: It would also be a good idea to add the `def allowsNullElement: Boolean` to `CollectionFactory` and test that we can add a `null` element. If so and if `allowsNullElementQuery == true` we can also test that `.contains(null) == ture` in this corner case.

javalib/src/main/scala/java/util/package.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/CollectionTest.scala
andreaTP: ok for ```def allowsNullElementQuery: Boolean```
but for ```def allowsNullElement: Boolean``` clash with the one in SetFactory, hopefully it will be another pr topic

andreaTP:
if this looks good I'll go to squash commits

nicolasstucki:
Your are missing some spaces between the methods in `NavigableView`. Otherwise LGTM.

andreaTP:
done

nicolasstucki:
Jenkins, retest this please

andreaTP:
sorry, rechecked style

sjrd:
Jenkins, retest this please

andreaTP:
due to the recent modifications on the comparable behaviour one test fail in "compliant-asinstanceof" mode.
I have reviewed the test to satisfy previously discussed fail fast behaviour.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    case class TestObj(num: Int)
+
+    when("compliant-asinstanceofs").
+    it("should throw exception on non comparable objects") {
+      val ts = treeSetTestFactory.empty[TestObj]
+
+      expect(ts.size()).toEqual(0)
+      expectThrows[ClassCastException](ts.add(TestObj(111)))
```
nicolasstucki: It seams that the `expectsThrows` is letting the exception pass through. 

```
[error]  x should throw exception on non comparable objects
org.scalajs.jasminetest.JasmineTestException: java.lang.ClassCastException: TestObj(222) is not an instance of java.lang.Comparable
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/library/src/main/scala/scala/scalajs/runtime/StackTrace.scala:39)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/library/src/main/scala/scala/scalajs/runtime/StackTrace.scala:33)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalanglib/src/main/scala/java/lang/Throwables.scala:24)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalanglib/src/main/scala/java/lang/Throwables.scala:12)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalanglib/src/main/scala/java/lang/Throwables.scala:343)
	at <jscode>.(scalajsenv.js:171)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalanglib/src/main/scala/java/lang/Comparable.scala:3)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala:362)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalib/src/main/scala/java/util/TreeSet.scala:35)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/javalib/src/main/scala/java/util/TreeSet.scala:112)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala:109)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala:105)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala:105)
	at <jscode>.(file:/localhome/jenkins/a/workspace/scalajs-task-worker/jasmine-test-framework/src/main/scala/org/scalajs/jasminetest/JasmineTest.scala:21)
```
The issue seams to be with the `excpectThrows` and not with the `TreeSet` implementation. Maybe @sjrd has some insight of why in this case the `excpectThrows` is not catching the exception.

Maybe try to de-alias `ClassCastException` manually by replacing it with `java.lang.ClassCastException`.

andreaTP: could you retest with the last modifications pleaz?

sjrd:
Jenkins, retest this please

andreaTP:
ok, need help on this

andreaTP:
couldn't understand...

nicolasstucki:
I cloned your branch and will try to figure out what is happening there.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,365 @@
...
+      expect(ts.comparator eq null).toBeTruthy
```
nicolasstucki: The problem is that now `ts` could be a `TreeSet` with or without a custom comparator. I'll add some comments with a solution.

nicolasstucki:
```scala
expect((ts.comparator eq null) ^ treeSetTestFactory.hasExplicitCoparator).toBeTruthy
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,365 @@
...
+    with SortedSetFactory {
+  def implementationName: String =
+    "java.util.TreeSet"
+
+  def empty[E]: ju.TreeSet[E] =
+    new TreeSet[E]
+
+  override def allowsNullElement: Boolean = false
+
+  override def allowsNullElementQuery: Boolean = false
```
nicolasstucki:
```scala
def hasExplicitCoparator: Boolean = false
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,365 @@
...
+        case (None, None) => 0
+      }
+    }
+
+  override def empty[E]: ju.TreeSet[E] =
+    new TreeSet[E](EvenNullComp[E]())
+
+  override def allowsNullElement: Boolean = true
+
+  override def allowsNullElementQuery: Boolean = true
```
nicolasstucki:
```scala
override def hasExplicitCoparator: Boolean = false
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
```diff
@@ -0,0 +1,365 @@
...
+  def empty[E]: ju.TreeSet[E] =
+    new TreeSet[E]
+
+  override def allowsNullElement: Boolean = false
+
+  override def allowsNullElementQuery: Boolean = false
+}
+
+class TreeSetWithNullFactory extends TreeSetFactory {
+
```
nicolasstucki:
```scala
override def implementationName: String =
  super.implementationName + " {allows null}"
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
nicolasstucki: Actually this should be tested even if it is not in `compliant-asinstanceofs` mode. 
Put it in it's own `it("should ...") {...}`.

nicolasstucki:
Thats all

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: You get the point @nicolasstucki  the problem here is that the behaviour of this is not well fixed all around now, can we change the implementation and expect the exception at least other the second insert?
It is a looser check but is what happens on the JVM

andreaTP: ```hasExplicitCoparator``` doesn't sounds to me as a solution, I will think at another proposal...

nicolasstucki: Well, the behaviour of `TreeSet.comparator` is quite well defined for each of the two test cases. Therefore telling the test that it `hasExplicitCoparator` is a simple way to test all cases.

nicolasstucki: Maybe it is better if you do something like:
```scala
if (treeSetTestFactory.hasExplicitCoparator) 
  expect(ts.comparator ne null).toBeTruthy
else 
  expect(ts.comparator eq null).toBeTruthy
```

andreaTP: good catch the check will be done on the comparator null check result
```
if (ts.comparator eq null) {
...
} else {
...
}
```
if the condition is satisfied the inner tests must pass without the needs of the next boolean flag

test-suite/src/test/scala/org/scalajs/testsuite/javalib/TreeSetTest.scala
andreaTP: ok

andreaTP:
@nicolasstucki  check this solution

nicolasstucki:
@andreaTP that looks good. But it would be good to also test that `TreeSet.comparator` returns `null` if and only if the set was instantiated using a constructor that does not take a `Comparator`.

andreaTP:
@nicolasstucki checked in last commit

nicolasstucki:
Jenkins, retest this please

andreaTP:
ok, yes the new implementation have this.
please have you got a set of sbt parameters to reproduce locally pleaz? so my next commit will be good...

andreaTP:
fixed the test and squashed commits.
cross fingers...

nicolasstucki:
Jenkins retest this please

sjrd:
retest this please

nicolasstucki:
retest this please

sjrd:
retest this please

andreaTP:
great guys!

PR #1826, opened 2015-07-28
Fix #1824: Add a union type `js.|[A, B]`, aka `A | B`.

### LINKED ISSUE #1824: js.| union type
Edit: the original proposal was to name it `\/`.

It would be worth thinking about a union type constructor for facade types. Similar to `js.UndefOr[+A]` in its construction, `js.|[+A, +B]`, or whatever it's called, would represent a value of type `A` or `B`. It would have zero API, though. The basic idea is the following:
```scala
@RawJSType // don't do this at home
sealed trait |[+A, +B]

object \/ {
  implicit def fromA[A](a: A): A | Nothing = a.asInstanceOf[A | Nothing]
  implicit def fromB[B](b: B): Nothing | B = b.asInstanceOf[Nothing | B]
}
```
Then of course we need ways to transitively convert an `A`, a `B`, or a `C` to an `A | B | C`. This probably needs a bit of recursive implicit magic. Even trickier: convert `A | B`, `A | C` and `B | C` to `A | B | C` ^^

sjrd:
This is an unboxed union type, which is represent a value of type either A or B.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Union.scala
+++ b/library/src/main/scala/scala/scalajs/js/Union.scala
@@ -1,0 +1,76 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.js
+
+import scala.language.implicitConversions
+
+/** Value of type A or B (union type).
+ *
+ *  Scala does not have union types, but they are important to many
+ *  interoperability scenarios. This type provides a (partial) encoding of
+ *  union types using implicit evidences.
+ */
+@scala.scalajs.js.annotation.RawJSType // Don't do this at home!
+sealed trait |[A, B] // scalastyle:ignore
+
+object | { // scalastyle:ignore
+  /** Evidence that `A <: B`, taking top-level `|`-types into account. */
+  sealed trait Evidence[-A, +B]
+
+  /** A unique (and typically dead-code-eliminated away) instance of
+   *  `Evidence`.
+   */
+  private object ReusableEvidence extends Evidence[scala.Any, scala.Any]
+
+  abstract sealed class EvidenceLowestPrioImplicits {
+    /** If `A <: B2`, then `A <: B1 | B2`. */
+    implicit def right[A, B1, B2](implicit ev: Evidence[A, B2]): Evidence[A, B1 | B2] =
+      ReusableEvidence.asInstanceOf[Evidence[A, B1 | B2]]
+  }
+
+  abstract sealed class EvidenceLowPrioImplicits extends EvidenceLowestPrioImplicits {
+    /** `Int <: Double`, because that's true in Scala.js. */
+    implicit def intDouble: Evidence[Int, Double] =
+      ReusableEvidence.asInstanceOf[Evidence[Int, Double]]
+
+    /** If `A <: B1`, then `A <: B1 | B2`. */
+    implicit def left[A, B1, B2](implicit ev: Evidence[A, B1]): Evidence[A, B1 | B2] =
+      ReusableEvidence.asInstanceOf[Evidence[A, B1 | B2]]
+  }
+
+  object Evidence extends EvidenceLowPrioImplicits {
+    /** `A <: A`. */
+    implicit def base[A]: Evidence[A, A] =
+      ReusableEvidence.asInstanceOf[Evidence[A, A]]
+
+    /** If `A1 <: B` and `A2 <: B`, then `A1 | A2 <: B`. */
+    implicit def allSubtypes[A1, A2, B](
+        implicit ev1: Evidence[A1, B], ev2: Evidence[A2, B]): Evidence[A1 | A2, B] =
+      ReusableEvidence.asInstanceOf[Evidence[A1 | A2, B]]
+  }
+
+  /** Upcast `A` to `B1 | B2`.
+   *
+   *  This needs evidence that `A <: B1 | B2`.
+   */
+  implicit def from[A, B1, B2](a: A)(implicit ev: Evidence[A, B1 | B2]): B1 | B2 =
+    a.asInstanceOf[B1 | B2]
+
+  /** Operations on union types. */
+  implicit class UnionOps[A <: _ | _](val self: A) extends AnyVal {
+    /** Explicitly merge a union type to a supertype (which might not be a
+     *  union type itself).
+     *
+     *  This needs evidence that `A <: B`.
+     */
+    def merge[B](implicit ev: |.Evidence[A, B]): B =
+      self.asInstanceOf[B]
+  }
+}
--- a/test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
@@ -1,0 +1,178 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.library
+
+import scala.language.implicitConversions
+
+import scala.collection.mutable
+
+import scala.scalajs.js
+import js.|
+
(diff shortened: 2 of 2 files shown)


sjrd:
We can still debate on the name, but otherwise I believe it's good. Debates on the name should still happen on the issue #1824.

Review by @gzm0.

library/src/main/scala/scala/scalajs/js/Union.scala
```diff
@@ -0,0 +1,65 @@
...
+  abstract sealed class EvidenceLowestPrioImplicits {
+    /** If `A <: B2`, then `A <: B1 | B2`. */
+    implicit def right[A, B1, B2](implicit ev: Evidence[A, B2]): Evidence[A, B1 | B2] =
+      ReusableEvidence.asInstanceOf[Evidence[A, B1 | B2]]
+  }
+
+  abstract sealed class EvidenceLowPrioImplicits extends EvidenceLowestPrioImplicits {
+    /** `Int <: Double`, because that's true in Scala.js. */
+    implicit def intDouble: Evidence[Int, Double] =
+      ReusableEvidence.asInstanceOf[Evidence[Int, Double]]
```
gzm0: `Byte <: Short <: Int` as well? Also `Float <: Double`?

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
```diff
@@ -0,0 +1,153 @@
...
+import js.|
+
+import org.scalajs.jasminetest.JasmineTest
+import org.scalajs.testsuite.Typechecking._
+
+import js.JSConverters._
+
+object UnionTypeTest extends JasmineTest {
+
+  private implicit def unionAsJSAny(x: _ | _): js.Any = x.asInstanceOf[js.Any]
```
gzm0: Maybe we should have this just like for `UndefOr`: If `A <% js.Any` and `B <% js.Any`, allow `A | B => js.Any`.

library/src/main/scala/scala/scalajs/js/Union.scala
sjrd: We could add them, of course, but they're *way* less useful. `intDouble` is important because if your JS library expects a number or a string, it will take `Double | String`. But then if you write `3` it doesn't typecheck ^^ You're unlikely to write `3.toByte` or `3.1f` in that situation, though.

If I want them all I need `Byte -> Short`, `Byte -> Int`, `Byte -> Float`, `Byte -> Double`, `Short -> Int`, `Short -> Float`, `Short -> Double`, `Float -> Double` (because implicits would not magically make the transitive closure in that case). That's *a lot* of implicits to add in the implicit resolution scope (=  longer compile times) for virtually nothing.

Does that sound like a reasonable rationale, or do you want them all?

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
sjrd: This is not implementable, unfortunately. Because when you have a `A | B`, you don't know whether `x` is an `A` or a `B`, and you have no means of testing which one it is, in general. So you cannot know whether you should apply the conversion from `A` to `js.Any` or the conversion from `B` to `js.Any`.

This is also why the implicit evidences are all "fake", instead of having an `apply` method that "convert" the values (even to just cast in most cases). Because the `allSubtypes` evidence wouldn't know how to construct its `apply` method. This, in turn, is why `Evidence` is sealed. Because we cannot allow "custom" conversions that are not known to boil down to a hard cast.

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
```diff
@@ -0,0 +1,153 @@
...
+
+      val x1: AnyVal | String = x
+      val x2: String | AnyVal = x
+
+      expect(x1).toBe("hello")
+      expect(x2).toBe("hello")
+
+      /* Note: the *total* upper bound does not work, because the expected
+       * type is not a | type.
+       */
```
gzm0: Can't we add:

```scala
implicit def toTopBound[A1, A2, B](x: A1 | A2)(implicit e: Evidence[A1 | A2, B]) = x.asInstanceOf[B]
```

to `|`'s companion?

This should work given the following works:

```scala
case class A(x: String)
object A {
  implicit def str(x: A) = x.x
}

val a = A("foo")
val b: String = a // compiles
```

sjrd: Unfortunately, that goes into diverging implicit search (unless you *explicitly* write `toTopBound(x)` instead of just `x`). So it's a bit useless.

library/src/main/scala/scala/scalajs/js/Union.scala
gzm0: I think if you leave `Evidence` sealed, you need to provide them, since there is no (reasonable) escape hatch for people (or am I missing something)?

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
gzm0: Ah, I see... that makes sense. But can we add "converters" for `String`, `Double` and `Int` at least? Again, having `Evidence` sealed makes it very unpractical to work with these. And I think interpreting the primitive values as `js.Any` is pretty common (or is that just our code?).

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
gzm0: Ah... I see... Could we add a `merge` method just like on `Either`?

gzm0:
That's all by the way.

library/src/main/scala/scala/scalajs/js/Union.scala
sjrd: If you come from a `x: Byte` that you want to assign to an `Int | String`, for example, you can always force its conversion a priori with `x.toInt`, which will typecheck as an `Int | String`.

However, if you already have a `Byte | String`, then you cannot assign it to an `Int | String`. You have to cast. But then again, there are still plenty of cases that are not supported by this encoding, and you have to cast anyway. Basically every time that a cast of one of your base types is needed, you have to cast the whole thing instead. Or for example if you have to use a `js.use(x).as[T]` for one of your elements, you have the same problem. But all of that only happens when you already have an `A | B` and you want to upcast it to `A2 | B` for some conversion of `A` to `A2`. And as I explained in another comment, it is not possible to support that in general.

I initially wanted to *not* have `intDouble` either. But it was just too impractical a limitation for that particular case.

We could also have an additional namespace for "extra" evidences in `|.ExtraEvidences`. One could `import |.ExtraEvidences._` when needed. But well ... not super nice either, and it would still have limitations.

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
sjrd: Interpreting "things" as `js.Any` is mostly useful if you work a lot with `js.Dynamic`. Otherwise very few things should be typed as `js.Any`. In this case it's mostly our code because we made Jasmine take `js.Any` to force conversions in typical cases, because most of our test suite is concerned with interop and implementation details of the language.

In more typical code, you have completely typed APIs where you don't have `js.Any`s (you have `Any` instead). If at all, people were using `js.Any` so far where now they would use a union type ^^

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
sjrd: You mean something like that:
```scala
implicit class UnionOps[A <: _ | _](val self: A) extends AnyVal {
  def merge[B](implicit ev: Evidence[A, B]): B =
    self.asInstanceOf[B]
}
```
?
I would have to try whether `B` can be reasonably inferred in those cases.

library/src/main/scala/scala/scalajs/js/Union.scala
sjrd: I mean I call `|` a "union type", but it does have a lot of limitations compared to a *true* union type as you would find in dotty, obviously.

I would suggest we keep it minimal first. As people start using it, they'll complain that this or that use case is not supported, and we can add more implicits, until most/all common cases are supported. WDYT?

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
gzm0: Yes

library/src/main/scala/scala/scalajs/js/Union.scala
gzm0: Yeah. That sounds good. Let's keep it this way and we'll add stuff as we go if needed.

gzm0:
LGTM (unless there is stuff you still want to change).

test-suite/src/test/scala/org/scalajs/testsuite/library/UnionTypeTest.scala
sjrd: OK I'll try that.

sjrd:
Updated with the addition of `merge`, including one pos test and one neg test. The rest has not been touched.

gzm0:
LGTM

PR #1833, opened 2015-07-30
Fix #1760: Add IRVersionNotSupportedException on IR deserialization.

### LINKED ISSUE #1760: More comprehensive error message if serialization fails due to binary compat
If a library that uses a newer version of Scala.js is on the classpath, the user is likely to get an error message of this kind:

```
java.io.IOException: Failed to deserialize info of /home/nick/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.3.jar#scala/math/Ordering$Byte$.sjsir
```

The real reason of this failure is hidden in the cause. The sbt plugin should catch this specific exception and display a more comprehensive error message.

nicolasstucki:


nicolasstucki:
In this case I explicitly removed the compatibility for `0.6.4` and this is how the new message would look like:
```
[error] (comp/test:scalaJSLinkingUnitClasspath) java.io.IOException: Requires Scala.js 0.6.4 (or compatible version).
[error] Caused by: Failed to deserialize info of /Users/nastucki/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.4.jar#scala/math/Ordering$Byte$.sjsir
[error] Total time: 3 s, completed Jul 30, 2015 4:15:08 PM
```
In `debug` mode or with `last`:
```
[debug] 	
java.io.IOException: Requires Scala.js 0.6.4 (or compatible version).
Failed to deserialize info of /Users/nastucki/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.4.jar#scala/math/Ordering$Byte$.sjsir
	at org.scalajs.core.tools.io.VirtualSerializedScalaJSIRFile$class.info(VirtualFiles.scala:143)
	at org.scalajs.core.tools.io.MemVirtualSerializedScalaJSIRFile.info(MemFiles.scala:104)
	at org.scalajs.core.tools.optimizer.Linker$PersistentIRFile.updateFile(Linker.scala:339)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$17.apply(Linker.scala:301)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$17.apply(Linker.scala:298)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.immutable.List.foreach(List.scala:318)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
	at scala.collection.AbstractTraversable.map(Traversable.scala:105)
	at org.scalajs.core.tools.optimizer.Linker.org$scalajs$core$tools$optimizer$Linker$$updateFiles(Linker.scala:298)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$1.apply(Linker.scala:64)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$1.apply(Linker.scala:64)
	at org.scalajs.core.tools.optimizer.package$.logTime(package.scala:19)
	at org.scalajs.core.tools.optimizer.Linker.link(Linker.scala:64)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:243)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:234)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.scalajs.core.ir.InfoSerializers$IRVersionNotSupportedIOException: This version (0.6.4) of Scala.js IR is not supported. Supported versions are: 0.6.0, 0.6.3
	at org.scalajs.core.ir.InfoSerializers$Deserializer.readHeader(InfoSerializers.scala:149)
	at org.scalajs.core.ir.InfoSerializers$Deserializer.deserialize(InfoSerializers.scala:100)
	at org.scalajs.core.ir.InfoSerializers$.deserializeWithVersion(InfoSerializers.scala:35)
	at org.scalajs.core.ir.InfoSerializers$.deserialize(InfoSerializers.scala:31)
	at org.scalajs.core.tools.io.VirtualSerializedScalaJSIRFile$class.info(VirtualFiles.scala:133)
	at org.scalajs.core.tools.io.MemVirtualSerializedScalaJSIRFile.info(MemFiles.scala:104)
	at org.scalajs.core.tools.optimizer.Linker$PersistentIRFile.updateFile(Linker.scala:339)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$17.apply(Linker.scala:301)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$17.apply(Linker.scala:298)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.immutable.List.foreach(List.scala:318)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
	at scala.collection.AbstractTraversable.map(Traversable.scala:105)
	at org.scalajs.core.tools.optimizer.Linker.org$scalajs$core$tools$optimizer$Linker$$updateFiles(Linker.scala:298)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$1.apply(Linker.scala:64)
	at org.scalajs.core.tools.optimizer.Linker$$anonfun$1.apply(Linker.scala:64)
	at org.scalajs.core.tools.optimizer.package$.logTime(package.scala:19)
	at org.scalajs.core.tools.optimizer.Linker.link(Linker.scala:64)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:243)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$18.apply(ScalaJSPluginInternal.scala:234)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
[error] (comp/test:scalaJSLinkingUnitClasspath) java.io.IOException: Requires Scala.js 0.6.4 (or compatible version).
[error] Failed to deserialize info of /Users/nastucki/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.4.jar#scala/math/Ordering$Byte$.sjsir
[error] Total time: 3 s, completed Jul 30, 2015 4:15:08 PM
```

ir/src/main/scala/org/scalajs/core/ir/InfoSerializers.scala
```diff
@@ -146,12 +146,15 @@ object InfoSerializers {
...
             s"This version ($version) of Scala.js IR is not supported. " +
-            s"Supported versions are: ${supported.mkString(", ")}")
+            s"Supported versions are: ${supported.mkString(", ")}", version)
       }
 
       version
     }
   }
+
+  class IRVersionNotSupportedIOException(msg: String, val version: String)
```
gzm0: I don't think you need the `IO` in here. Subclasses of `IOException` usually don't contain `IO`.

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -132,7 +133,14 @@ trait VirtualSerializedScalaJSIRFile extends VirtualBinaryFile with VirtualScala
...
       case e: IOException =>
-        throw new IOException(s"Failed to deserialize info of $path", e)
+        val pre = e match {
+          case e: IRVersionNotSupportedIOException =>
+            s"Requires Scala.js ${e.version} (or compatible version).\nCaused by: "
+
+          case _ => ""
+        }
+
+        throw new IOException(s"${pre}Failed to deserialize info of $path", e)
```
gzm0: Why not just let this bubble up?

gzm0: Oh I see... I think it is a better idea to format the error message the way you like in the sbt plugin. (Most sbt operations that fail, don't fail with a stacktrace...

nicolasstucki: I did not let it bubble up because the exception would not contain any reference to the file or library that was involved. For the format do you mean that this should throw again a `IRVersionNotSupportedIOException`.

nicolasstucki:
The message will look like this with this new version:
```
[error] (comp/compile:scalaJSLinkingUnitClasspath) org.scalajs.core.ir.InfoSerializers$IRVersionNotSupportedException: Failed to deserialize info of /Users/nastucki/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.4.jar#scala/math/Ordering$Byte$.sjsir
[error] Caused by: This version (0.6.4) of Scala.js IR is not supported. Supported versions are: 0.6.0, 0.6.3
```
Also the exception will be the `IRVersionNotSupportedException`.

nicolasstucki:
retest this please

gzm0:
The commit message is outdated.

ir/src/main/scala/org/scalajs/core/ir/InfoSerializers.scala
```diff
@@ -145,13 +145,31 @@ object InfoSerializers {
...
+      this(version,
+          supportedVersions,
+          s"This version ($version) of Scala.js IR is not supported. " +
+          s"Supported versions are: ${supportedVersions.mkString(", ")}",
+          null)
+    }
+
+    def this(msg: String, ex: IRVersionNotSupportedException) = {
+      this(ex.version, ex.supportedVersions,
+          s"$msg\nCaused by: ${ex.getMessage}", ex)
```
gzm0: "Caused by" is inserted by the stack trace printer.

gzm0:
I think we should not rely on exception messages for display to the user. The exception messages are purely for debugging and should not be shown to a user if the program is bug free.

I believe the right approach to this is to catch the exception in the sbt plugin and log an error message. @sjrd what is your take on this?

gzm0:
I'd propose the following:

```scala
case class IRVersionNotSupportedException(version: String, supported: Set[String]) extends IOException(IRVersionNotSupportedException.mkMessage(version, supported))
```

And throw it in `InfoSerializers`.

Leave `VirtualFiles` unchanged.

[In the linker][1], catch `IOExceptions` and log to the `logger` whenever it fails (and then rethrow).

Catch block could look something like this:

```scala
} catch {
  case ex: IOException =>
    val causes = Stream.iterate(ex)(_.getCause).takeWhile(_ != null).reverse
    val msg = causes.collectFirst {
      case IRVersionNotSupportedException(version, supported) =>
        s"This version ... "
    } getOrElse {
      s"Unknown exception: ${ex.getMessage}"
    }
    throw ex
}
```

[1]: https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala#L76

gzm0:
I think a question we have to answer is: Is it an implementation detail, that `VirtualSerializedScalaJSIRFile` uses `InfoSerializers` under the hood? If yes, we cannot propagate its exceptions. If no, it is ok to do so, but it's unclear to me how we properly chain them. Looking through the cause chain also seems like a hack...

gzm0:
[This guy thinks][1] that in error handling we should ignore the decoupling of the architecture. That is probably reasonable, but in this case we do exception handling: I.e. if it occurs, it doesn't mean the program has a bug, just that an exceptional condition occurred (Scala.js doesn't have a bug if a user wants us to load bad files).
Therefore, I think the abstraction should probably be preserved (meaning my proposal is bad). But that means we need an additional exception... :S

[1]: https://lairdnelson.wordpress.com/2012/02/16/error-handling-done-right-part-1/

sjrd:
Note that, at the end of the day, the only way you can fail an sbt task (and hence prevent dependent tasks to execute) is to let an exception bubble up.

gzm0:
Yes. But we shouldn't tailor the exceptions in the `tools` library so that they work well with sbt but don't make sense otherwise.

ir/src/main/scala/org/scalajs/core/ir/IRVersionNotSupportedException.scala
```diff
@@ -0,0 +1,14 @@
...
+object IRVersionNotSupportedException {
+  def mkMessage(version: String, supported: Set[String]): String = {
+    s"This version ($version) of Scala.js IR is not supported. " +
+    s"Supported versions are: ${supported.mkString(", ")}"
+  }
+}
+
+case class IRVersionNotSupportedException(version: String, supported: Set[String])
+    extends IOException(
+      IRVersionNotSupportedException.mkMessage(version, supported))
```
nicolasstucki: Missed the last return :(

nicolasstucki:
I decoupled a bit the dependencies on the exception, now tools knows about the `IRVersionNotSupportedException` and then the tools just throw an `IOException` with a new message which would look like:
```
[error] (comp/compile:scalaJSLinkingUnitClasspath) IOException: Failed to deserialize info of file compiled with Scala.js 0.6.4 (supported: 0.6.0, 0.6.3): /Users/nastucki/.ivy2/cache/org.scala-js/scalajs-library_2.11/jars/scalajs-library_2.11-0.6.4.jar#scala/math/Ordering$Byte$.sjsir
```
This seams to decouple the exception dependencies while also giving the user a meaningful message.

ir/src/main/scala/org/scalajs/core/ir/IRVersionNotSupportedException.scala
sjrd: Also the indentation is not right. It should be 4 spaces from the beginning of the line. Better would be not to break the `extends` line.

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -131,6 +131,11 @@ trait VirtualSerializedScalaJSIRFile extends VirtualBinaryFile with VirtualScala
     try {
       ir.InfoSerializers.deserialize(stream)
     } catch {
+      case e: ir.IRVersionNotSupportedException =>
+        throw new IOException(
+          s"Failed to deserialize info of file compiled with Scala.js ${e.version}" +
+          s" (supported: ${e.supported.mkString(", ")}): $path", e)
```
sjrd: 4 spaces.

Also, why not rethrow an `IRVersionNotSupportedException` (with a custom message)?

nicolasstucki: The idea was to not let the user see the exception class from the `ir` package as it exposes the concept of IR where the user cares more about the plugin version. Maybe I could throw an `UnsupportedVersionException` that has the current message.

gzm0: I think in this case it is clearly not an implementation detail, that we deserialize the underlying IR. Therefore, throwing an `IRVersionNotSupportedException` is probably OK.

nicolasstucki:
updated

ir/src/main/scala/org/scalajs/core/ir/IRVersionNotSupportedException.scala
```diff
@@ -0,0 +1,13 @@
+package org.scalajs.core.ir
+
+import java.io.IOException
+
+class IRVersionNotSupportedException(val version: String,
+    val supported: Set[String], message: String) extends IOException(message) {
```
gzm0: Consider making this a case class

nicolasstucki: I removed the case to avoid exposing the field `message` as this one should be accessed with the standard `getMessage` method.

gzm0: Makes sense.

sjrd:
LGTM

PR #1835, opened 2015-07-31
Fix #1818: Improve performance of Hashers.hashAsVersion

### LINKED ISSUE #1818: assembleLinkedClasses has a huge performance issue
See timings at #1817. This part of the linker is several times slower than the reachability analysis! That's not normal.

nicolasstucki:
Before the change the performance of `testSuite/test:fastOpt` where:
```
[debug] Linker: Read info: 19896 us
[debug] Linker: Check Infos: 374083 us
[debug] Linker: Compute reachability: 193762 us
[debug] Linker: Assemble LinkedClasses: 1023952 us 
[debug] Linker: Check IR: 2641108 us
[debug] Linker: cache stats: reused: 5164 -- invalidated: 0 -- trees read: 0
```
After the change it has a 13X improvement:
```
[debug] Linker: Read info: 26999 us
[debug] Linker: Check Infos: 418228 us
[debug] Linker: Compute reachability: 1427003 us
[debug] Linker: Assemble LinkedClasses: 78869 us
[debug] Linker: Check IR: 2693044 us
[debug] Linker: cache stats: reused: 5164 -- invalidated: 0 -- trees read: 0
```

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
@@ -53,9 +53,12 @@ object Hashers {
     val size = 2 * (if (considerPos) 2 else 1) * 20
     val builder = new StringBuilder(size)
 
-    def append(hash: Array[Byte]) =
-      for (b <- hash) builder.append(f"$b%02X")
+    def hexDigit(digit: Int): Char = Character.forDigit(digit, 16)
 
+    def append(hash: Array[Byte]): Unit = {
+      for (b <- hash)
+        builder.append(hexDigit(b >> 4)).append(hexDigit(b & 0xF))
+    }
     append(hash.treeHash)
 
     if (considerPos)


sjrd:
Waw! That's amazing! 

Maybe use an indexed array for `hexDigit`?

ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
```diff
@@ -53,8 +53,29 @@ object Hashers {
...
+        case 11 => "B"
+        case 12 => "C"
+        case 13 => "D"
+        case 14 => "E"
+        case _  => "F"
+      }
+    }
+
+    def append(hash: Array[Byte]): Unit =
+      for (b <- hash) builder.append(hexDigit(b >> 4)).append(hexDigit(b & 0xF))
```
sjrd: The body of the for should be on its dedicated line. Which also means that braces will be needed for the method.

sjrd:
That's all.

ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
gzm0: Also, consider having hex digit return a `Char`. That's probably faster than returning a single char string.

ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
```diff
@@ -53,8 +53,29 @@ object Hashers {
...
+        case 8  => "8"
+        case 9  => "9"
+        case 10 => "A"
+        case 11 => "B"
+        case 12 => "C"
+        case 13 => "D"
+        case 14 => "E"
+        case _  => "F"
+      }
+    }
```
gzm0: Consider replacing this by `Character.forDigit(digit, 16)`

nicolasstucki: That is the best solution, the time got down to:
```
[debug] Linker: Assemble LinkedClasses: 40811 us
```
I tried returning a `Char`, it also improved it, but not as much as this. I also tried the indexed array, but it was slower than the `switch`.

nicolasstucki:
Updated

nicolasstucki:
retest this please

nicolasstucki:
Of all the options that where proposed, `Character.forDigit(digit, 16)` was the fastest wit 25X over the original and the most compact.

sjrd:
LGTM

PR #1855, opened 2015-08-18
Implement NotSerializableException.

andreaTP:
added also the relevant tests of the throwables under java.io._

### DIFF
--- a/javalib/src/main/scala/java/io/Throwables.scala
+++ b/javalib/src/main/scala/java/io/Throwables.scala
@@ -21,3 +21,7 @@ class UnsupportedEncodingException(s: String) extends IOException(s) {
 abstract class ObjectStreamException protected (s: String) extends IOException(s) {
   protected def this() = this(null)
 }
+
+class NotSerializableException(s: String) extends ObjectStreamException(s) {
+  def this() = this(null)
+}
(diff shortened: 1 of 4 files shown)


javalib/src/main/scala/java/io/Throwables.scala
```diff
@@ -21,3 +21,7 @@ class UnsupportedEncodingException(s: String) extends IOException(s) {
 abstract class ObjectStreamException protected (s: String) extends IOException(s) {
   protected def this() = this(null)
 }
+
+class NotSerializableException(s: String) extends ObjectStreamException {
```
nicolasstucki: You should pass `s` to the super constructor.

nicolasstucki: It would also be better if you add the `NotSerializableException` to a new file with its name.

nicolasstucki:
Thats all

javalib/src/main/scala/java/io/Throwables.scala
sjrd: > It would also be better if you add the NotSerializableException to a new file with its name.

No, actually, we make an exception for, well, exceptions ^^ We put them all in a file `Throwables.scala`, because there are so many of them and they are so tiny that they fill up the list of files in a package for nothing, effectively harming readability.

nicolasstucki: Ok

andreaTP: ok

andreaTP:
fixed and rebased ^^

nicolasstucki:
LGTM
Jenkins, test this please

sjrd:
Jekins, test this please

sjrd:
Needs a rebase.

andreaTP:
I've properly separated Throwables test, hope this is appreciated.

sjrd:

nicolasstucki:
Jekins, test this please

PR #1861, opened 2015-08-20
Fix #1859: Implement java.util.concurrent.ThreadLocalRandom.

### LINKED ISSUE #1859: java.util.concurrent.ThreadLocalRandom
In scala there is a java implementation of ThreadLocalRandom ( https://github.com/scala/scala/blob/0b47dc2f28c997aed86d6f615da00f48913dd46c/src/forkjoin/scala/concurrent/forkjoin/ThreadLocalRandom.java ) if this implementation of Random (that basically uses Date.now()) satisfy you ( https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/java/util/Random.java ))
I will go for an implementation in pure Scala.

andreaTP:


javalib/src/main/scala/java/util/Random.scala
```diff
@@ -13,10 +13,12 @@ class Random(seed_in: Long) extends AnyRef with java.io.Serializable {
   private var nextNextGaussian: Double = _
   private var haveNextNextGaussian: Boolean = false
 
-  setSeed(seed_in)
+  _setSeed(seed_in)
```
sjrd: You cannot change this. This alters the behavior of `java.util.Random`.

Instead, you have to use a flag `initialized` that is set to `true` in the constructor, and let `ThreadLocalRandom.setSeed()` be called if `!initialized` (and call `super`). And throw `UnsupportedOperationException` only if `initialized`.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+
+class ThreadLocalRandom extends Random {
+
+  override protected def _setSeed(seed_in: Long): Unit = ()
+
+  override def setSeed(seed: Long): Unit =
+    throw new UnsupportedOperationException()
+
+  /*
+  * this code has to be removed!!!
```
sjrd: TODO?

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+  */
+  import scala.scalajs.js
+  /** Generate a random long from JS RNG to seed a new Random */
+  private def randomSeed(): Long = {
```
sjrd: This is unused.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+  /*
+  * end of code to be removed
+  */
+
+
+  def nextInt(least: Int, bound: Int): Int = {
+    if (least >= bound) throw new IllegalArgumentException()
+
+    //nextInt(bound - least) + least
+    (nextLong(bound.toLong - least.toLong) + least.toLong).toInt
```
sjrd: Why not the commented out line? Using Longs is *much* slower.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  def nextLong(_n: Long): Long = {
+    if (_n <= 0)
+      throw new IllegalArgumentException("n must be positive")
+    // Divide n by two until small enough for nextInt. On each
+    // iteration (at most 31 of them but usually much less),
+    // randomly choose both whether to include high bit in result
+    // (offset) and whether to continue with the lower vs upper
+    // half (which makes a difference only if odd).
```
sjrd: Use `/* ... */` for multi-line comments, with [the documented style](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#non-scaladoc-comments)

Also add a blank line between the `throw` line and the comment.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: because of overflows, with Int it doesn't behave as java

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: argh...

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+    var offset: Long = 0L
+    var n: Long = _n
+
+    while (n >= Integer.MAX_VALUE) {
+      val bits: Int = next(2)
+      val half: Long = n >>> 1
```
sjrd: Don't use type annotations for local vals.
Also, `halfn` would be clearer than `half` (yes, I know you took it from the impl in Scala)

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: er...

javalib/src/main/scala/java/util/Random.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
sjrd: That is really *not* normal. Overflows in Scala.js work exactly the same as overflows in Scala/JVM and Java. The Java impl in scala/scala uses ints there. You're saying using Ints here does not behave as scala/scala's impl?

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+      n = nextn
+    }
+    offset + nextInt(n.toInt)
+  }
+
+  def nextLong(least: Long, bound: Long): Long = {
+    if (least >= bound)
+      throw new IllegalArgumentException()
+
+    (nextDouble(bound.toDouble - least.toDouble) + least.toDouble).toLong
```
sjrd: You can't do that. There are Longs that cannot be represented in a Double. You have to use Long operations and `nextLong(Long)`

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: I'm saying that the Scala implementation doesn't strictly follows the Java requested behaviour

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,98 @@
...
+      throw new IllegalArgumentException()
+
+    nextDouble() * (bound - least) + least
+  }
+}
+
+object ThreadLocalRandom {
+
+  val _current = {
+    println("created current")
```
sjrd: println

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: this will not work in corner cases, let say when you are close to the upper bound of the type

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: argh...

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
sjrd: Ah. Then I would suggest to optimize a fast path for cases where no overflow occurs, because Longs are really really slow:
```scala
val diff = bound - least
if (diff > 0) nextInt(diff) + least
else (nextLong(bound.toLong - least.toLong) + least.toLong).toInt
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
sjrd: Then you need to figure something out, because your solution does not work in corner cases either. Also when you are close to the upper bound of the type.

andreaTP:
WOW just realised what is the problem...

```
scala> 1L - (-100000000L)
res0: Long = 100000001

scala> 1L - (Long.MinValue)
res1: Long = -9223372036854775807
```
cannot understand this...

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
...
+
+class ThreadLocalRandom extends Random {
+
+  override def setSeed(seed: Long): Unit =
+    if (initialized) throw new UnsupportedOperationException()
+    else super.setSeed(seed)
+
+
+  def nextInt(least: Int, bound: Int): Int = {
+    if (least >= bound) throw new IllegalArgumentException()
```
sjrd: Keep the then part of an if in statement position on its own line:
```scala
if (least >= bound)
  throw new ...
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
sjrd: You can probably do it with a common implementation of `nextLong(Long)` and `nextLong(Long, Long)` which works as if `_n` was an *unsigned* Long:
```scala
private def nextUnsignedLong(_n: Long): Long = {
  // no bounds check

  ...
  while (n >= Int.MaxValue || n < 0) {
    ... // same (note that n >>> 1 is an unsigned / 2
  }
  offset + nextInt(n.toInt)
}
```
and then:
```scala
  def nextLong(_n: Long): Long = {
    if (_n <= 0)
      throw new IllegalArgumentException("n must be positive")
    nextUnsignedLong(_n) // a positive signed is also an unsigned
  }

  def nextLong(least: Long, bound: Long): Long = {
    if (least >= bound)
    nextUnsignedLong(bound - least) + least // - and + behave as unsigned
  }
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
...
+      throw new IllegalArgumentException()
+
+    nextDouble() * (bound - least) + least
+  }
+}
+
+object ThreadLocalRandom {
+
+  val _current =
+    new ThreadLocalRandom()
```
sjrd: Should be `private`

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
+/*
+ * closely translated from:
+ * https://github.com/scala/scala
+ * scala.concurrent.fokjoin.ThreadLocalRandom
```
sjrd: This reference will eventually become broken. I suggest to copy the license header from that file + say it was translated to Scala.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Random
+
+class ThreadLocalRandom extends Random {
+
+  override def setSeed(seed: Long): Unit =
+    if (initialized) throw new UnsupportedOperationException()
+    else super.setSeed(seed)
+
+
```
sjrd: Double blank line.

javalib/src/main/scala/java/util/Random.scala
```diff
@@ -13,6 +13,8 @@ class Random(seed_in: Long) extends AnyRef with java.io.Serializable {
   private var nextNextGaussian: Double = _
   private var haveNextNextGaussian: Boolean = false
 
+  protected var initialized = false
```
sjrd: This must be in `ThreadLocalRandom`, not in `Random`.

sjrd: (`Random` should not be touched at all.)

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
...
+package java.util.concurrent
+
+import java.util.Random
+
+class ThreadLocalRandom extends Random {
+
+  override def setSeed(seed: Long): Unit =
+    if (initialized) throw new UnsupportedOperationException()
+    else super.setSeed(seed)
```
sjrd: This if is in statement position. It should be expanded with its side-effects on dedicated lines:
```scala
... = {
  if (initialized)
    throw new ...
  else
    super.setSeed(seed)
}
```
or without the else:
```scala
... = {
  if (initialized)
    throw new ...
  super.setSeed(seed)
}
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,81 @@
...
+
+
+  def nextInt(least: Int, bound: Int): Int = {
+    if (least >= bound) throw new IllegalArgumentException()
+
+    (nextLong(bound.toLong - least.toLong) + least.toLong).toInt
```
sjrd: Should have a fast path when there is no overflow as per my older comment (which does not show up in the diff anymore).

sjrd:
> cannot understand this...

Arithmetic modulo 2^64, offset by `Long.MinValue`. Classic in 2's complement representation of integers.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: sure

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/Random.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

andreaTP:
will go on late afternoon

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      expectThrows[UnsupportedOperationException](tlr.setSeed(1))
+    }
+
+    it("should return nextInt that fits bounds") {
+      val tlr = ThreadLocalRandom.current()
+
+      for (_ <- 1 to 100) {
+        val one = Random.nextInt()
+        //to have at least one positive for bound
+        val two = abs(Random.nextInt())
```
sjrd: It is not necessary that `bound` be positive. `abs` is not needed.
However, you need to avoid the situation where you end up with `one == two`. You should skip the turn (or retry until you get something different, or use `two + 1` instead of `two`).

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,119 @@
...
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.javalib.util.concurrent
+
+import org.scalajs.jasminetest.JasmineTest
+import java.util.concurrent.ThreadLocalRandom
+import scala.util.Random
+import scala.math.abs
+import scala.math.max
+import scala.math.min
```
sjrd: Use the multi-import syntax:
```scala
import scala.math.{abs, max, min}
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,119 @@
...
+        val least = min(one, two)
+        val bound = max(one, two)
+
+        val nextInt = tlr.nextInt(least, bound)
+
+        expect((nextInt >= least) && (nextInt < bound))
+      }
+
+      expectThrows[IllegalArgumentException](tlr.nextInt(2,1))
+      expectThrows[IllegalArgumentException](tlr.nextInt(1,1))
```
sjrd: space after `,`

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    it("should return nextLong under a bound") {
+      val tlr = ThreadLocalRandom.current()
+
+      for (_ <- 1 to 100) {
+        val bound = abs(Random.nextLong())
```
sjrd: Careful with `0` and `Long.MinValue` (`abs(Long.MinValue) == Long.MinValue`)

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
sjrd: In general, it is not a very good idea to have tests depend on random input data. I would prefer if you generated once, offline, 100 pairs of `(least, bound)`, and you hard-coded that pair in the test.

Use this opportunity to include in the pairs some known corner cases (preferably visually separated from the random ones), like `least == Int.MinValue && bound == Int.MaxValue` for example.

Same for the other tests.

sjrd:
The commit message should start with "Fix #1859: "

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: found a clean solution to that

sjrd: What do you consider not clean in my suggestion?

andreaTP:
Please review implementation for ```Int```s.
if you consider ok the management of bounds (in ThreadLocalRandom.scala) and the tests as they are done I will complete the same way also long and double.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def nextInt(least: Int, bound: Int): Int = {
+    if (least >= bound)
+      throw new IllegalArgumentException()
+
+    val difference = bound - least
+    if (difference > 0)
+      nextInt(difference) + least
+    else
+      nextInt(-difference) - least
```
sjrd: This does not work.
`-difference == least - bound`
`nextInt(n) ∈ [0, n)`
so
`nextInt(-difference) ∈ [0, least - bound)`
so
`nextInt(-difference) - least ∈ [-least, -bound)`
and not
`[least, bound)`
as required.

If you invent some clever formula, please prove mathematically that it is correct.

sjrd: It might not be obvious from my previous comment, but all the arithmetic there is in signed 2's complement. So `[-least, -bound]` is
`{ -least, -least+1, -least+2, ..., Int.MaxValue, Int.MinValue, Int.MinValue+1, ..., -bound-1, -bound }`

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,96 @@
...
+ * http://creativecommons.org/publicdomain/zero/1.0/
+ *
+ * and translated to Scala
+ */
+
+package java.util.concurrent
+
+import java.util.Random
+
+class ThreadLocalRandom private (var initialized: Boolean = false)
```
sjrd: Don't put `initialized` there. Put it as a `private var` in the body. However, do *not* assign it, i.e.:
```scala
private var initialized: Boolean = _
```
If you assign it (to `false`), it will be reassigned to `false` after the super constructor has been called. But it is that super constructor which calls `setSeed` which is supposed to set it to true, so the reset to `false` destroys the work.

Alternatively, remove `initialized = true` from `setSeed` and set it to `true` in the constructor (this is cleaner, IMO):
```scala
class ThreadLocalRandom ... {
  private var initialized: Boolean = _
  initialized = true // executed after the super constructor is done
```

andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok, review this lines with a math guy, the next implemented solution is the best we can propose now.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+      throw new IllegalArgumentException()
+
+    val difference = bound - least
+    if (difference > 0)
+      nextInt(difference) + least
+    else {
+      /*
+       * Interval size here is at least Int.MaxValue
+       * so loop wsill exit with probability of at least 1/2
+       */
```
sjrd: Typos + comment style:
```scala
/* The interval size here is greater than Int.MaxValue,
 * so the loop will exit with a probability of at least 1/2.
 */
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+      throw new IllegalArgumentException()
+
+    val difference = bound - least
+    if (difference >= 0)
+      nextLong(difference) + least
+    else {
+      /*
+       * Interval size here is at least Long.MaxValue
+       * so loop wsill exit with probability of at least 1/2
+       */
```
sjrd: Same:
```scala
/* The interval size here is greater than Int.MaxValue,
 * so the loop will exit with a probability of at least 1/2.
 */
```

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  def nextInt(least: Int, bound: Int): Int = {
+    if (least >= bound)
+      throw new IllegalArgumentException()
+
+    val difference = bound - least
+    if (difference > 0)
+      nextInt(difference) + least
+    else {
```
sjrd: The then part needs `{}` since the else part has `{}`.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  def nextLong(least: Long, bound: Long): Long = {
+    if (least >= bound)
+      throw new IllegalArgumentException()
+
+    val difference = bound - least
+    if (difference >= 0)
+      nextLong(difference) + least
+    else {
```
sjrd: Same.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,217 @@
...
+      checkBounds(-1507203912, -507923122)
+
+      expectThrows[IllegalArgumentException](tlr.nextInt(2,1))
+      expectThrows[IllegalArgumentException](tlr.nextInt(1,1))
+    }
+
+    it("should return nextLong under a bound") {
+      val tlr = ThreadLocalRandom.current()
+
+      for (_ <- 1 to 100) {
```
sjrd: Same stuff about a hard-coded list of input data. Also the same for the other tests.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
```diff
@@ -0,0 +1,217 @@
...
+      checkBounds(-174356501, 2506025)
+      checkBounds(1205481279, -1674427890)
+      checkBounds(-217617201, -833593065)
+      checkBounds(20848991, -1440699601)
+      checkBounds(2010553201, 797241229)
+      checkBounds(658643437, 315920491)
+      checkBounds(-1507203912, -507923122)
+
+      expectThrows[IllegalArgumentException](tlr.nextInt(2,1))
+      expectThrows[IllegalArgumentException](tlr.nextInt(1,1))
```
sjrd: Space after `,`

sjrd:
That's all.

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ThreadLocalRandomTest.scala
andreaTP: sure

andreaTP:
hope that's all also for me! :-)

sjrd:

nicolasstucki:
@andreaTP It looks like it fails on JDK 6.

sjrd:
Makes sense. `ThreadLocalRandom` is "Since 1.7".
You'll have to move the test file to the [require-jdk7](https://github.com/scala-js/scala-js/tree/master/test-suite/src/test/require-jdk7) subdirectory.

andreaTP:
moved

nicolasstucki:
You should also rebase

sjrd:
@nicolasstucki Why rebase?

sjrd:
Jenkins, retest this please

nicolasstucki:
Rebase to get 1997b122e8e38354c1a11bdb20147963e72d6653 and reduce the chances spurious failures.

sjrd:
No need: Jenkins builds on a provisional merge of the PR branch against master. You can look at the logs for the currently building job: they include the new `sbtretry` stuff.

andreaTP:
? Am I supposed to fix something?

sjrd:
Er no, it's still running.

PR #1862, opened 2015-08-20
Fine-grained timeouts and retries in the main tests in CI.

sjrd:
The idea is to retry at a smaller granularity than the full task-worker: each invocation of sbt.
Assuming there are `n` invocations of sbt, and each has a probability `p` of success, the old scheme would have a probability `p^n` of success. With this change, the probability of success of each invocation with its retry is changed to `p + (1-p)*p = p * (2-p)`, giving a total probability of success of `(p * (2-p))^n = p^n * (2-p)^n`

Since we have `n = 25` sbt invocations which can fail (in our experience), we have the following graphs:
http://fooplot.com/plot/bhcgp6fpft
which depict the probability of total success (y) in terms of the probability of success of one invocation (x).

That doesn't give us any meaningful hindsight, though, as we don't know what is the probability of success of one job. However, we can also plot the probability of success with the change given the probability of success after the change. To do this, we substitute `x` by `x^(1/25)` (which is the probability of success of one invocation given the probability of success now) in the formula for success after the change. This gives
http://fooplot.com/plot/68k2xrmarg

With a rough estimation that we succeed 1/3 of the time in the current situation, we would now succeed about 19/20 of the time. A worthy 2.5x improvement :-)

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -14,48 +14,69 @@
 <ci>
 
   <task id="main"><![CDATA[
+    sbtretry() {
+      local TIMEOUT=25m
+      echo "RUNNING timeout -k 5 $TIMEOUT sbt" "$@"
+      timeout -k 5 $TIMEOUT sbt $SBT_OPTS "$@"
+      local CODE=$?
+      if [ "$CODE" -eq 124 ]; then
+        echo "TIMEOUT after" $TIMEOUT
+      fi
+      if [ "$CODE" -ne 0 ]; then
+        echo "RETRYING timeout -k 5 $TIMEOUT sbt" "$@"
+        timeout -k 5 $TIMEOUT sbt $SBT_OPTS "$@"
+        CODE=$?
+        if [ "$CODE" -eq 124 ]; then
+          echo "TIMEOUT after" $TIMEOUT
+        fi
+        if [ "$CODE" -ne 0 ]; then
+          echo "FAILED TWICE"
+          return $CODE
+        fi
+      fi
+    }
     export JAVA_HOME=$HOME/apps/java-$java
-    sbt ++$scala package packageDoc &&
-    sbt ++$scala helloworld/run \
+    sbtretry ++$scala package packageDoc &&
+    sbtretry ++$scala helloworld/run \
         'set scalaJSStage in Global := FastOptStage' \
         ++$scala helloworld/run \
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala helloworld/run \
         helloworld/clean &&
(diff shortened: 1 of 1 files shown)


sjrd:
Let's see whether that works ...

sjrd:
Let's stress-test it a little bit. Jenkins, retest this please

sjrd:
retest this please

sjrd:
retest this please

sjrd:
There's one successful retry after spurious exception here: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/56039/consoleFull

sjrd:
Review by @gzm0 
Note that this accompanied by a change in the task script for scalajs-task-worker: it now exposes under the environment variable `$SBT_OPTS` the options that are to be passed to sbt. There was an alias from `sbt` to `sbt lots-of-options`. The lots-of-options are now in `$SBT_OPTS` and the alias says `sbt $SBT_OPTS`. This is necessary to launch `sbt` with all its options through the `timeout` command, since as an argument to `timeout` `sbt` it not recognized as an alias.

sjrd:
Looks like it works well :)

nicolasstucki:
LGTM

nicolasstucki:
Do we require more stress testing on this?

sjrd:
No, I don't think so. If it's OK for you, you can merge.

gzm0:
Seems I'm late to the party :) Just a comment: I would try and separate things in matrix and in the Jenkins build as follows:

- `matrix.xml`: *what* we test (important to track in VC)
- Jenkins: How we test (exact dispatch, etc.). Not necessary to track in VC

Rather than adding the command to the `matrix.xml`, I would have redefined `sbt` in Jenkins. Rationale mostly being that now Jenkins won't be able to test the old commits anymore (since the `sbt` alias does not exist anymore IIUC).

sjrd:
Jenkins is still able to test old commits. The alias still exists, and does `alias sbt='sbt $SBT_OPTS`. The change in Jenkins is perfectly backwards compatible: I have only added `SBT_OPTS`.

I considered doing a pure Jenkins change, but I didn't want *all* sbt invocations to have this behavior.

gzm0:
> I considered doing a pure Jenkins change, but I didn't want all sbt invocations to have this behavior.

Understandable. But IMO `sbtretry` should still be part of the Jenkins script, not the CI matrix. Having two commands `sbt` and `sbtretry` that are available to the matrix is completely reasonable IMO.

Now we are exposing details we really shouldn't. `export JAVA_HOME` is another example of that by the way, but doing it differently is harder there (actually no...).

gzm0:
I'll propose a PR with a change, so we can discuss.

gzm0:
See #1871

PR #1867, opened 2015-08-23
Finish fix for #1759: Use `new` in *all* versions of Array.clone().

sjrd:
The original fix in b153f86e883267efba51bd9320b8d6d616dc168f was missing one change, in the variant of Array.clone() used when targeting ES 5.


sjrd:
Review by @nicolasstucki. This time I actually tested this locally on PhantomJS 2.0.0 on Windows.

nicolasstucki:
LGTM

PR #1872, opened 2015-08-24
Fix #1869: Allow Scala.js-defined JS objects.

### LINKED ISSUE #1869: Allow Scala.js-defined JS objects
There is no reason whatsoever not to allow them, so they should be allowed. The semantics are obvious: it's a lazy val whose value is an instance of a Scala.js-defined JS module class.

sjrd:


sjrd:
Review by @gzm0 if you have time tonight or tomorrow; otherwise @nicolasstucki. We're trying to schedule 0.6.5 with the rest of Scala.js-defined JS stuff for this Friday/week-end, and since this PR touches many parts of the code, I'll have to base the other ones on top of this one, I fear; so it would be good for this to get in relatively quickly.

The other PRs will be much simpler.

sjrd:
Actually even this PR is *mostly* tests and one-line changes.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -338,6 +339,16 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
         }
+
+        if (sym.isModule && (enclosingOwner is OwnerKind.JSNonNative)) {
+          def shouldBeExposed: Boolean = {
+            !sym.isSynthetic &&
+            !isPrivateMaybeWithin(sym)
+          }
+          if (shouldBeExposed)
+            sym.addAnnotation(ExposedJSMemberAnnot)
+        }
```
sjrd: This part is to expose `objects` *inside* Scala.js-defined JS classes/objects.

gzm0:
Reviewing this now

compiler/src/test/scala/org/scalajs/core/compiler/test/ScalaJSDefinedTest.scala
```diff
@@ -400,6 +472,26 @@ class ScalaJSDefinedTest extends DirectTest with TestHelpers {
...
+      def foo: Int = js.native
+      @JSName(A.b)
+      def bar: Int = js.native
+    }
+    """ hasErrors
+    """
+      |newSource1.scala:12: error: The argument to JSName must be a literal string
+      |      @JSName(A.a)
+      |       ^
+    """
```
gzm0: These test cases are potentially unstable, since there are other errors in the class... (I don't know how the one for the class slipped my review)

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -589,6 +589,12 @@ object Trees {
     val tpe = AnyType
   }
 
+  /** Like [[LoadModule]] but for a JS module class. */
+  case class JSLoadModule(cls: ClassType)(
+      implicit val pos: Position) extends Tree {
+    val tpe = AnyType
+  }
+
```
gzm0: Consider calling this `LoadJSModule`. It loads a JS module. It is not a JavaScript way of loading a module (such a thing does not really exist, does it...)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Refiner.scala
```diff
@@ -69,10 +69,12 @@ final class Refiner(semantics: Semantics, outputMode: OutputMode) {
...
     val kind = {
-      if (classDef.kind == ClassKind.ModuleClass && !info.isModuleAccessed)
-        ClassKind.Class
-      else
+      if (classDef.kind.hasModuleAccessor && !info.isModuleAccessed) {
+        if (classDef.kind == ClassKind.ModuleClass) ClassKind.Class
+        else ClassKind.JSClass
+      } else {
         classDef.kind
+      }
```
gzm0: Have you considered unifying this code with the one in the linker? It seems like a good candidate for a helper.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
gzm0: Isn't this change unrelated to Scala.js defined objects? This will also expose normal objects (which is correct). I think this should be in a separate commit.

gzm0:
Please add tests for exposing of objects inside Scala.js defined classes. (Preferably in a separate commit).

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -458,12 +458,24 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       gen(cd.impl)
 
+      // Generate exported module accessors
+      val exports = {
+        // Generate exported constructors or accessors
+        val exportedAccessors =
+          if (isStaticModule(sym)) genModuleAccessorExports(sym)
+          else Nil
```
gzm0: I'm assuming this prepares for exporting of Scala.js defined *classes*.

gzm0:
That's all. Neat how little change this needs :)

compiler/src/test/scala/org/scalajs/core/compiler/test/ScalaJSDefinedTest.scala
sjrd: Ouch, indeed. Copy paste error in the first place, I guess.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
sjrd: Should I then also rename `JSLoadConstructor` -> `LoadJSConstructor`? The same reasoning applies to both.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: Yes, you are right. I will make a separate commit with its own tests.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Refiner.scala
sjrd: If you have a suggestion about *where* to put such a helper? `ClassKind`'s companion object?

Eventually (probably in the next major version), I want to entirely refactor `ClassKind`s anyway. They now conflate several axes that have become orthogonal. For example, I would reify `hasModuleAccessor` as a separate flag.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Yes.

sjrd:
Addressed the 2 commits stuff & the "unstable" test.
Waiting for confirmation on the two other items before changing stuff (`LoadJSConstructor`/`LoadJSModule` and the `removeModuleAccessor` helper).

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Indeed. It didn't bother me that much, since "loading" a constructor is nonsensical in non-js land anyways... But maybe it is more consistent indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Refiner.scala
gzm0: > ClassKind's companion object?

Yeah... I guess that's best. It is a step in the right direction probably. You probably want to keep it private to `optimizer` though...

Btw. Is it really a good idea to have a `hasModuleAccessor` flag? This will lead to combinations which are not allowed... (anyways, not subject to discussion right now).

sjrd: > You probably want to keep it private to optimizer though...

`ClassKind` is in `ir/` ... Can't do.

sjrd:
Updated. All concerns should be addressed, now, unless I missed something.

sjrd:
retest this please

gzm0:
LGTM

PR #1877, opened 2015-08-26
Fix #1820: Allow nested native JS stuff in native JS objects.

### LINKED ISSUE #1820: Allow nesting objects, traits, and classes inside objects that extend js.Object
Currently ScalaJS prevents nesting objects, traits, and classes inside objects, traits, and classes that extend js.Object.

It would be nice to relax this restriction abit for modeling certain javascript APIs inside facade types in a way that more naturally maps to their own style.

In code form, it is requested that this:
```
object A extends js.Object {
  object B extends js.Object
  class B extends js.Object
}
class A extends js.Object
```
compiles

Consideration should be given to how the mapping of names occurs but I think generally the names should map for the class B / object B as though they are nested under whatever `@JSName` is assigned to object A

--Or--

a `@JSName` could be required for them as well

sjrd:
Native JS objects can now contain nested native JS traits, classes and objects. For classes and objects, their full JS name is qualified by the full JS name of the enclosing native JS object, since they appear as members of this object. In this example:

```scala
object Foo extends js.Object {
  @JSName("Babar")
  object/class Bar extends js.Object
}
```

the full JS name of Bar is "Foo.Babar", not just "Babar".

This prompted to revisit native JS objects and classes in Scala objects. For those, an explicit `@JSName` is required, and is their full JS name (it is not qualified by their enclosing Scala object). For classes, this is a warning, since we previously erroneously allowed that.

Native JS stuff is not allowed in Scala.js-defined JS objects.

Altogether, these changes to the rules prompted a major refactoring of the checks performed by PrepJSInterop.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/annotation/JSFullName.scala
+++ b/library/src/main/scala/scala/scalajs/js/annotation/JSFullName.scala
@@ -1,0 +1,16 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.js.annotation
+
+/** IMPLEMENTATION DETAIL: Saves the fully qualified JS name of a symbol.
+ *
+ *  Do not use this annotation yourself.
+ */
+class JSFullName(fullName: String) extends scala.annotation.StaticAnnotation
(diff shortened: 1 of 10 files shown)


sjrd:
Review by @gzm0 or @nicolasstucki. Whoever gets here first/has enough time.

gzm0:
Nice! I like the distinction between stuff in native objects and stuff in non-native objects. Didn't have time for thorough review so far though.

sjrd:
Argh, 2.10.2 doesn't have `originalOwner`, and what seems to be the equivalent old API does not work ... I'll need an additional implementation-detail annotation to store the full name.

sjrd:
Updated to work on 2.10.

nicolasstucki:
LGTM

library/src/main/scala/scala/scalajs/js/annotation/JSFullName.scala
```diff
@@ -0,0 +1,16 @@
...
+\*                                                                      */
+
+
+package scala.scalajs.js.annotation
+
+/** IMPLEMENTATION DETAIL: Saves the fully qualified JS name of a symbol.
+ *
+ *  Do not use this annotation yourself.
+ */
+class JSFullName(fullName: String) extends scala.annotation.StaticAnnotation
```
gzm0: Scalac: FAIL

PR #1879, opened 2015-08-26
Fix #1845: Add scalaJSOutputWrapper sbt setting.

### LINKED ISSUE #1845: Setting/task to prepend and append stuff in the emitted .js
So that we can do stuff like http://stackoverflow.com/questions/26637600/how-to-convert-scala-js-application-to-commonjs-module trivially.

We could also follow up with a shortcut `nodeModuleSettings` that automatically adds the appropriate prepend/append to turn the code emitted by Scala.js into a Node.js module.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

project/Build.scala
```diff
@@ -836,7 +836,10 @@ object Build extends sbt.Build {
       base = file("examples") / "reversi",
       settings = exampleSettings ++ Seq(
           name := "Reversi - Scala.js example",
-          moduleName := "reversi"
+          moduleName := "reversi",
+          scalaJSOutputWrapper := (
+              "// Scala.js - Reversi\n//\n// Compiled with Scala.js\n",
+              "// End of Scala.js generated script")
```
sjrd: This would be a better fit for an actual test in one of the projects of `sbt-plugin-test`, with an automated test that opens the output of fastOptJS and fullOptJS, and verifies that the starts and ends with the appropriate content.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -85,28 +85,37 @@ class ScalaJSClosureOptimizer {
...
 
     val sourceMap = Option(compiler.getSourceMap())
 
+    val (header, footer) = cfg.customOutputWrapper
+
     // Write optimized code
     val w = output.contentWriter
     try {
+      if (header != "")
+        w.write(header + "\n")
```
sjrd: Consider first computing
```scala
val (header0, footer0) = ...
val header = if (header0 == "") "" else header0 + "\n"
val footer = if (footer0 == "") "" else footer0 + "\n"
```
and then this will be just `w.write(header)`, but most importantly the computation of `offsetLine` down below will be simplified. I think that should be more readable.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -85,28 +85,37 @@ class ScalaJSClosureOptimizer {
...
       w.write(outputContent)
+      if (footer != "")
+        w.write(footer + "\n")
       if (sourceMap.isDefined)
         w.write("//# sourceMappingURL=" + output.name + ".map\n")
     } finally w.close()
 
     // Write source map (if available)
     sourceMap.foreach { sm =>
+      val offsetLine = (if (header != "") 0 else header.count(_ == '\n') + 1) + 1
```
sjrd: It would be good to comment why the final `+ 1`.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
```diff
@@ -47,6 +47,11 @@ final class Emitter(semantics: Semantics, outputMode: OutputMode) {
     emitPostlude(builder, logger)
   }
 
+  def emitCustomHeader(customHeader: String, builder: JSFileBuilder): Unit = {
+    if(customHeader != "")
+      builder.addLine(customHeader)
```
sjrd: Space after `if`.
Also, `builder.addLine` should be called several times if there are `\n`'s in `customHeader`. `addLine` is not equipped to deal with new lines in its parameter, especially if the builder creates a source map.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
```diff
@@ -108,6 +113,11 @@ final class Emitter(semantics: Semantics, outputMode: OutputMode) {
     }
   }
 
+  def emitCustomFooter(customFooter: String, builder: JSFileBuilder): Unit = {
+    if (customFooter != "")
+      builder.addLine(customFooter)
```
sjrd: Same here.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -177,7 +186,10 @@ object ScalaJSClosureOptimizer {
       /** Base path to relativize paths in the source map */
       relativizeSourceMapBase: Option[URI] = None,
       /** Elements we won't warn even if they don't exist */
-      noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = Nil
+      noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = Nil,
+      /** Custom js code that wraps the output */
+      customOutputWrapper: (String, String) = ("", "")
+
   ) extends OptimizerConfig with ScalaJSOptimizer.OptimizerConfig
```
sjrd: That will cause binary compatibility issues. And since it's in a surface API of the tools, and we haven't broken that part of the API since 0.6.0, we should try very hard not to break binary compat.

It was of course a mistake to make `Config` a case class, but well ... We'll have to fix that in 1.0.0. In the meantime, I suggest moving the a scheme similar to what we do for [`Semantics`](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala), i.e.:

* `final class` with private constructor (without default values)
* add the old constructor as a deprecated `def this()`, with all its default values
* same for a deprecated `apply` method in the companion
* a non-deprecated `apply` method in the companion with only the required parameters, i.e., `output`, which calls the private constructor and explicitly sets every optional parameter to its default value.
* explicitly written `equals`, `hashCode`
* `withXYZ` methods to incrementally set all the parameters

The same applies to the other Config object in `ScalaJSOptimizer`.

Locally, you can test binary compatibility with

    > tools/mimaReportBinaryIssues

sjrd:
That's all.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -85,28 +85,45 @@ class ScalaJSClosureOptimizer {
...
+      else withNewLine(header0) + "(function(){'use strict';\n"
+    val footer =
+      if (hasErrors) ""
+      else "}).call(this);\n" + withNewLine(footer0)
+
+    val outputContent = if (hasErrors) "" else withNewLine(compiler.toSource)
 
     val sourceMap = Option(compiler.getSourceMap())
 
+
```
nicolasstucki: Need to remove this line

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -275,15 +275,15 @@ object ScalaJSPluginInternal {
...
-                output = AtomicWritableFileVirtualJSFile(output),
-                cache = Some(taskCache),
-                wantSourceMap = (emitSourceMaps in fastOptJS).value,
-                relativizeSourceMapBase = relSourceMapBase,
-                bypassLinkingErrors = opts.bypassLinkingErrors,
-                checkIR = opts.checkScalaJSIR,
-                disableOptimizer = opts.disableOptimizer,
-                batchMode = opts.batchMode),
+            Config(AtomicWritableFileVirtualJSFile(output)).
+              withCache(Some(taskCache)).
```
sjrd: Our (undocumented) style is to put the `.` at the beginning of the next line rather than at the end of the previous line (with 2 spaces of indentation, as you did).

(this does not apply to .sbt files, where we follow sbt's style, which is to put the `.` at the end of the previous line)

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
       /** Elements we won't warn even if they don't exist */
-      noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = Nil
+      val noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing],
+      /** Custom js code that wraps the output */
+      val customOutputWrapper: (String, String)
   ) extends OptimizerConfig with ScalaJSOptimizer.OptimizerConfig
+    /* for binary compatibility */ with Product with Serializable with Equals {
+
+    // NOTE: This class was previously a case class and hence many useless
+    // methods where implemented for binary compatibility :(
```
sjrd: typo: where -> were

+ comment style: use `/* */` as described at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#non-scaladoc-comments

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
       /** Base path to relativize paths in the source map */
-      relativizeSourceMapBase: Option[URI] = None,
+      val relativizeSourceMapBase: Option[URI],
       /** Elements we won't warn even if they don't exist */
-      noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = Nil
+      val noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing],
+      /** Custom js code that wraps the output */
+      val customOutputWrapper: (String, String)
   ) extends OptimizerConfig with ScalaJSOptimizer.OptimizerConfig
+    /* for binary compatibility */ with Product with Serializable with Equals {
```
sjrd: 4 spaces

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+
+    // NOTE: This class was previously a case class and hence many useless
+    // methods where implemented for binary compatibility :(
+
+    // For binary compatibility
+    @deprecated("Use Config(output) and .withXYZ(.) methods", "0.6.5")
```
sjrd: The `.` in `(.)` looks weird. I suggest simply writing `.withXYZ()`.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+        unCache: Boolean = true,
+        disableOptimizer: Boolean = false,
+        batchMode: Boolean = false,
+        wantSourceMap: Boolean = false,
+        prettyPrint: Boolean = false,
+        relativizeSourceMapBase: Option[URI] = None,
+        noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = Nil) = {
+
+      this(
+        output = output,
```
sjrd: 4 spaces

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+        unCache: Boolean = this.unCache,
+        disableOptimizer: Boolean = this.disableOptimizer,
+        batchMode: Boolean = this.batchMode,
+        wantSourceMap: Boolean = this.wantSourceMap,
+        prettyPrint: Boolean = this.prettyPrint,
+        relativizeSourceMapBase: Option[URI] = this.relativizeSourceMapBase,
+        noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = this.noWarnMissing
+    ): Config = {
+      copyWith(output, cache, bypassLinkingErrors, checkIR, unCache,
+        disableOptimizer, batchMode, wantSourceMap, prettyPrint,
```
sjrd: 4 spaces

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+        bypassLinkingErrors: Boolean = this.bypassLinkingErrors,
+        checkIR: Boolean = this.checkIR,
+        unCache: Boolean = this.unCache,
+        disableOptimizer: Boolean = this.disableOptimizer,
+        batchMode: Boolean = this.batchMode,
+        wantSourceMap: Boolean = this.wantSourceMap,
+        prettyPrint: Boolean = this.prettyPrint,
+        relativizeSourceMapBase: Option[URI] = this.relativizeSourceMapBase,
+        noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = this.noWarnMissing
+    ): Config = {
```
sjrd: Leave `): Config = {` at the end of the previous line, even it goes over 80 chars, as recommended in https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#method-definition

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+        disableOptimizer: Boolean = this.disableOptimizer,
+        batchMode: Boolean = this.batchMode,
+        wantSourceMap: Boolean = this.wantSourceMap,
+        prettyPrint: Boolean = this.prettyPrint,
+        relativizeSourceMapBase: Option[URI] = this.relativizeSourceMapBase,
+        noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = this.noWarnMissing,
+        customOutputWrapper: (String, String) = this.customOutputWrapper
+    ): Config = {
+      new Config(output, cache, bypassLinkingErrors, checkIR, unCache,
+        disableOptimizer, batchMode, wantSourceMap, prettyPrint,
```
sjrd: 4 spaces (etc.)

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+        checkIR: Boolean = this.checkIR,
+        unCache: Boolean = this.unCache,
+        disableOptimizer: Boolean = this.disableOptimizer,
+        batchMode: Boolean = this.batchMode,
+        wantSourceMap: Boolean = this.wantSourceMap,
+        prettyPrint: Boolean = this.prettyPrint,
+        relativizeSourceMapBase: Option[URI] = this.relativizeSourceMapBase,
+        noWarnMissing: Seq[ScalaJSOptimizer.NoWarnMissing] = this.noWarnMissing,
+        customOutputWrapper: (String, String) = this.customOutputWrapper
+    ): Config = {
```
sjrd: Same: end of the previous line

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,223 @@ object ScalaJSClosureOptimizer {
...
+      new Config(output, cache, bypassLinkingErrors, checkIR, unCache,
+        disableOptimizer, batchMode, wantSourceMap, prettyPrint,
+        relativizeSourceMapBase, noWarnMissing, customOutputWrapper = ("", ""))
+    }
+
+    // For binary compatibility
+    @deprecated("Not a case class anymore", "0.6.5")
+    def unapply(config: Config): Option[(WritableVirtualJSFile,
+      Option[WritableVirtualTextFile], Boolean, Boolean, Boolean,
```
sjrd: 4 spaces

sjrd:
Comments on `ScalaJSClosureOptimizer` also apply to `ScalaJSOptimizer`.

That's all.

sjrd:
OK. Apparently the failures are genuine: they are related to source maps, which are probably off-by-one in either fastOpt or fullOpt

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -85,28 +85,44 @@ class ScalaJSClosureOptimizer {
...
+
+    val (header0, footer0) = cfg.customOutputWrapper
+    val header =
+      if (hasErrors) ""
+      else withNewLine(header0) + "(function(){'use strict';\n"
+    val footer =
+      if (hasErrors) ""
+      else "}).call(this);\n" + withNewLine(footer0)
+
+    val outputContent = if (hasErrors) "" else withNewLine(compiler.toSource)
```
gzm0: I think protecting each individual block here with `hasErrors` adds a lot of clutter. From the code, I believe this was only introduced since otherwise `toSource` throws. Consider dropping all of it. We can write anything if an error happens.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -151,34 +167,224 @@ object ScalaJSClosureOptimizer {
...
+    }
+
+    // For binary compatibility
+    override def equals(other: Any): Boolean = super.equals(other)
+
+    // For binary compatibility
+    override def hashCode(): Int = super.hashCode()
+
+    // For binary compatibility
+    override def toString(): String = super.toString()
```
gzm0: Did you test that these still behave the same way? (I.e. value by value equality rather than object equality?)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
```diff
@@ -108,6 +111,9 @@ final class Emitter(semantics: Semantics, outputMode: OutputMode) {
     }
   }
 
+  def emitCustomFooter(customFooter: String, builder: JSFileBuilder): Unit =
+    emitLines(customFooter, builder)
+
```
gzm0: Do we really need these two methods?

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -154,6 +154,9 @@ object ScalaJSPlugin extends AutoPlugin {
     val emitSourceMaps = SettingKey[Boolean]("emitSourceMaps",
         "Whether package and optimize stages should emit source maps at all", BPlusSetting)
 
+    val scalaJSOutputWrapper = TaskKey[(String, String)]("scalaJSOutputWrapper",
+      "Custom wrapper for the generated .js files", BPlusTask)
```
gzm0: I think it would be helpful to add a note that this is a header and a footer. Wrapper might not be clear enough.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
nicolasstucki: Not really, but @sjrd advised on it because at some point we will have to factor out some common `ScalaJSOptimizer` and `ScalaJSClosureOptimizer`.

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
nicolasstucki: It does not behave the same way, but we do not use those methods anyway.

sjrd: Might be good to keep a reasonable `toString()`, though.

For `equals` and `hashCode`, the behavior changes, but the new behavior is better than the old one.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
nicolasstucki: Ok

sjrd:
retest this please

sjrd:
LGTM

sbt-plugin-test/build.sbt
```diff
@@ -32,7 +35,10 @@ lazy val withDOM = project.settings(baseSettings: _*).
     name := "Scala.js sbt test w/ DOM",
     jsDependencies ++= Seq(
         RuntimeDOM,
-        "org.webjars" % "jquery" % "1.10.2" / "jquery.js"
+        "org.webjars" % "jquery" % "1.10.2" / "jquery.js",
```
sjrd: Erf, missing `)` before the `,` at the end of the line.

sbt-plugin-test/build.sbt
```diff
@@ -23,7 +23,10 @@ lazy val root = project.in(file(".")).
 lazy val noDOM = project.settings(baseSettings: _*).
   enablePlugins(ScalaJSPlugin).
   settings(
-    name := "Scala.js sbt test w/o DOM"
+    name := "Scala.js sbt test w/o DOM",
+    scalaJSOutputWrapper := (
+      "// Scala.js - noDOM sbt test\n//\n// Compiled with Scala.js\n",
+      "// End of Scala.js generated script")
```
sjrd: And while you're here any: 4 spaces.

sbt-plugin-test/build.sbt
antonkulaga: I am just curious, why so old version of jquery is used here?

sjrd: Because we know it works. It was brand new when it put it there. The point here is more to test the `jsDependencies` mechanism than to test jQuery.

nicolasstucki:
Jenkins, retest this please

sjrd:
retest this please

PR #1883, opened 2015-08-27
Fix #1870: Add js.constructorOf[JSClass] to retrieve its constructor.

### LINKED ISSUE #1870: js.constructorOf[SomeJSClass] to retrieve the JS constructor
Add the following primitive method in the package object `js`:
```scala
def constructorOf[C <: js.Any]: js.Dynamic = sys.error("stub")
```
The compiler will check at compile-time that `C` refers to a `class` (not a trait nor module class, nor any other complex type), similarly to `classOf[C]`. `C` can be native or not.

The method returns the JS constructor of the class `C`. In IR terms, this is `JSLoadConstructor(ClassType(C))`.

This is particularly important for Scala.js-defined JS classes, as it allows to obtain the JS constructor of a JS class without exporting it (which allows dce) nor instantiating it.

*Implementation note:* since `C` will be erased by erasure, `PrepJSInterop` must rewrite calls to this method to some runtime method `runtime.constructorOf(classOf[C])`. The backend will then see the `ClassConstant(C)` as a parameter of `runtime.constructorOf`; it is of course invalid to call `runtime.constructorOf(cls)` with a `cls` which is not a `ClassConstant`.

sjrd:
The method `js.constructorOf[JSClass]` returns the JavaScript constructor function for the class `JSClass`. The latter must be a class (not trait nor object) extending js.Any. It may be a native JS class or a Scala.js-defined JS class. The restrictions on what types can be used are enforced at compile-time.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -3147,7 +3147,7 @@ abstract class GenJSCode extends plugins.PluginComponent
       implicit val pos = tree.pos
 
       def receiver = genExpr(receiver0)
-      val genArgs = genPrimitiveJSArgs(tree.symbol, args)
+      def genArgs = genPrimitiveJSArgs(tree.symbol, args)
 
       if (code == DYNNEW) {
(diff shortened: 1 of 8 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #1895, opened 2015-09-02
Adds 'commonLibraries' method to CrossProject.

matthughes:
Proposing to add the method 'commonLibraries' to CrossProject.  This allows you to define libraries that should be on both the JVM and JS classpath.

### DIFF
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
@@ -272,6 +272,18 @@ final class CrossProject private (
   def settings(ss: Def.Setting[_]*): CrossProject =
     copy(jvm.settings(ss: _*), js.settings(ss: _*))
 
+  /** Define common libraries for both JS/JVM projects.  
+   *  JS projects will automatically be converted into the SJS format. 
+   */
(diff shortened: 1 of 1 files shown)


sjrd:
The following existing feature already does what you're trying to achieve:
```scala
.settings(
  libraryDependencies += "org.example" %%% "awesome-lib" % "1.0.0"
)
```

matthughes:
So that works when you are inlining the dependency.  My projects, however, share an SBT plugin where I define all the common dependencies.  So I'll have an object like this:

```
object DependencyManagement {
  val scalatest = "org.scalatest" %% "scalatest" % "3.0.0-M7"
  val scalacheck = "org.scalacheck" %% "scalacheck" % "1.12.4"

  private val scalazVersion = "7.1.0"
  val scalazCore = "org.scalaz" %% "scalaz-core" % scalazVersion
}
```

And then in any downstream project, I can just use the named value `scalatest` to ensure all downstream projects are running the same versions.  This works great for JVM projects but not for SJS projects where the dependency is common among both JS and JVM projects.

I can't use the `%%%` method in `DependencyManagement`.

sjrd:
You can also use `%%%` in your `DependencyManagement`. You just have to surround your definitions with `Def.setting { ... }`. You'll find concrete examples on the mailing list of you search for that. I'm on mobile right now so it's a bit tedious for me to look it up.

PR #1896, opened 2015-09-03
WIP Refactor tools

### LINKED ISSUE #1635: Touching a file in src/test/resources triggers test:fastOptJS
If I touch a file in `src/test/resources`, it causes a rebuild of my test JS and taking about 5 seconds.

```
webapp-client> last test:fastOptJS
[info] Fast optimizing .../target/scala-2.11/webapp-client-test-fastopt.js
[debug] Linker: Read info: 25703 us
[debug] Linker: Compute reachability: 475989 us
[debug] Linker: Assemble LinkedClasses: 859315 us
[debug] Linker: cache stats: reused: 15325 -- invalidated: 0 -- trees read: 0
[debug] Linker: 1367511 us
[debug] Inc. optimizer: Batch mode: false
[debug] Inc. optimizer: Incremental part: 18493 us
[debug] Inc. optimizer: Optimizing 0 methods.
[debug] Inc. optimizer: Optimizer part: 50 us
[debug] Inc. optimizer: 32347 us
[debug] Refiner: Compute reachability: 364890 us
[debug] Refiner: Assemble LinkedClasses: 19782 us
[debug] Refiner: 384803 us
[debug] Emitter: Class tree cache stats: reused: 4193 -- invalidated: 0
[debug] Emitter: Method tree cache stats: resued: 20894 -- invalidated: 0
[debug] Emitter (write output): 278975 us
```

I think this is separate from the incremental linker issue in that regardless of how fast linking can be, it shouldn't be triggered if the underlying compilation units haven't changed.

gzm0:


gzm0:
Review by @sjrd

gzm0:
Jenkins, test this please

gzm0:
I tested this locally.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -38,50 +38,51 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
...
-    files.clear()
-    statsReused = 0
-    statsInvalidated = 0
-    statsTreesRead = 0
+  /** Obsolete method. Does not do anything. */
+  @deprecated("Does not do anything. Don't use", "0.6.5")
+  def clean(): Unit = ()
+
+  // Binary compat
+  def link(irInput: immutable.Traversable[VirtualScalaJSIRFile], logger: Logger,
```
sjrd: Please add "TODO" in the comment, so that this can be found when we grep for "TODO"s and "deprecated"s one day.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -38,50 +38,51 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
...
 
-    val infos = encodedNameToPersistentFile.values.map(_.info).toList
+    for (irFile <- irInput) {
+      val info = irFile.info
+      infosBuilder += info
+      encodedNameToFileBuilder += info.encodedName -> irFile
+    }
+
+    val infos = infosBuilder.result
+    val encodedNameToFile = encodedNameToFileBuilder.result
```
sjrd: `result()` with `()` (twice). Calling `result()` has side effects, in theory, since it invalidates the state of the builder.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRFileCache.scala
```diff
@@ -0,0 +1,155 @@
...
+    statsInvalidated = 0
+    statsTreesRead = 0
+
+    resultStats
+  }
+
+  /** Returns the current set of files for use
+   *
+   *  Can only be invoked after an invocation of update (but an arbitrary
+   *  number of times).
```
sjrd: and before `cleanAfterUse()`?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRFileCache.scala
```diff
@@ -0,0 +1,155 @@
...
+   */
+  def cleanAfterUse(): IRFileCache.Stats = {
+    requireReadyState(true, "Cache already clean or uninitialized. Call update")
+    readyToUse = false
+
+    val resultStats =
+      new IRFileCache.Stats(statsReused, statsInvalidated, statsTreesRead)
+
+    // "Garbage-collect" persisted versions of files that have disappeared
+    irCache.retain((_, f) => f.cleanAfterUse())
```
sjrd: Can't we do this garbage collection in `update()`? That would avoid the `.filter(_.exists)` in `def files`.

sjrd: And, btw, would completely eliminate the need for `cleanAfterUse()`. You could instead clear the stats at the beginning of `update`, and provide a side-effect-free accessor to the stats.

gzm0: Yes. We could do that. I first thought it restricts us from changing the caching strategy, but it doesn't seem to actually do so.
The only downside is that you cannot clean up the cache without updating it. So we might use more memory during the late phase of the pipeline (since we retain everything until the next run). Basically this means this patch potentially *increases* memory usage. Your call.

sjrd: Er ... I meant to `retain` at the *end* of `update`. Not at the beginning.
```scala
val updated = irFiles.foldLeft(...)

irCache.retain(...)

updated
```
How would that increase memory usage, compared to doing exactly the same retain later in the algorithm, but in the same run? On the contrary, potentially this reduces memory retention earlier.

Unless I am completely missing something.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSOptimizer.scala
```diff
@@ -81,29 +84,50 @@ class ScalaJSOptimizer(val semantics: Semantics, val outputMode: OutputMode,
...
+   *  If there is no change, [[hasOldResult]] is called. If it returns true,
+   *  no further action is taken and [[None]] is returned.
+   *  Otherwise, the whole pipeline is run. At the end, [[buildJS]] is called
+   *  with a function that emits the trees to the specified builder. The result
+   *  is wrapped into a [[Some]] and returned.
+   */
+  def optimizeIR[T](buildJS: (JSTreeBuilder => Unit) => T)(
+      hasOldResult: () => Boolean)(
+      irFiles: Traversable[VirtualScalaJSIRFile],
+      cfg: OptimizerConfig, logger: Logger): Option[T] = {
```
sjrd: Consider reordering the parameter lists as follows:
```scala
(irFiles, cfg, logger)(hasOldResult)(buildJS)
```
The rationale is 1. put lambdas at the end, as is more common in Scala APIs and 2. put `hasOldResult` first because it's called first.

sjrd:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSOptimizer.scala
gzm0: I can put `hasOldResult` earlier, but I cannot put `(irFiles, cfg, logger)` first, since it then would collide with another overload...

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRFileCache.scala
gzm0: Ah... Yes we could do this. Two points:

- This will not allow us to change the caching strategy (for example right now, we *never* release trees to GC, even if they are not used in a run).
- We would need to be smarter about when to release the original IR file to GC. The `PersistentIRFile` can definitely be smarter about it (don't keep a reference if you have everything you need). However, references to files whose tree is never read cannot seem to be released (which is the majority IIUC).

This is not an argument against moving the retain call into the `update` method, but an argument against removing the `cleanAfterUse` method. It is probably a good idea to put the `retain` call there, since the `filter` call in the `files` method likely creates a new strict collection (but who knows... Scala collection library FTW)...

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSOptimizer.scala
sjrd: Erf... maybe that's a sign we should use a different name?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRFileCache.scala
sjrd: OK let's keep `cleanAfterUse`, just in case. It's easier to deprecate it than to impose new calls.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSOptimizer.scala
gzm0: Hahaha, touché. Suggestions? `optimizeIRInternal` comes to mind, but I dislike it, since it is not internal...

sjrd: Good question. Back in my Delphi days the answer would be `OptimizeIREx` ^^
Maybe `optimizeIRIfChanged` or `optimizeIRCond`?

gzm0:
There is one thing I do not like about this solution at all btw.

Consider the following:

```scala
val opt = new ScalaJSOptimizer

val output1 = // ...
val output2 = // ...

def write(cp: IRClasspath) = {
  opt.optimizeIR(cp, output1)
  opt.optimizeIR(cp. output2)
}

write(cp1)
write(cp2) // updated classpath
```

The second time `write` is invoked, `output2` will not be written anymore. We could restrict the usage of an optimizer to a single output file, but there are two things I don't like about this:

1. The output file should then be an argument to the optimizer and not to the optimize method
2. In other cases where an option should be an optimizer argument but this is impractical to the user, we have added reset logic to the optimizer (n.b. `wantSourceMap`).

Maybe we should add a similar guard and only short circuit if the following three conditions are met:

1. The IR didn't change since the last run
2. The result file exists
3. (new) the result file is the same as last run

gzm0:
Updated. Please don't merge yet, I'd like to have your opinion on my last comment first.

sjrd:
> Maybe we should add a similar guard and only short circuit if the following three conditions are met:

Yes that sounds good to me.

gzm0:
@sjrd I added an example commit with a guard. This yields to another problem: Now there is state in the optimizer which is not always updated depending on the method. So having a `ScalaJSClosureOptimizer` that uses a `ScalaJSOptimizer` that is actually used will now yield bad behavior:

```scala
val opt = new ScalaJSOptimizer
val fullOpt = new ScalaJSClosureOptimizer(opt)
opt.optimize()
// change in code
fullOpt.optimize()
opt.optimize() // boom, not updated
```

I'm starting to have the feeling we need to look at this from another angle... :S

gzm0:
This is stale. I am working on this, but not on this branch and it will look very different. Closing.

PR #1897, opened 2015-09-03
Fix #1891: Ignore synthetic companions inside native JS objects

### LINKED ISSUE #1891: Class with constructors with optional parameters can't be embedded in Objects that extend js.Object
```scala
@js.native object A extends js.Object {
  @js.native class B(x : Int = ???) extends js.Object
}
```

produces

> Native JS objects cannot contain inner Scala traits, classes or objects (i.e., not extending js.Any)

gzm0:


gzm0:
Review by @sjrd

sjrd:
OK. Might be good to also have a run test (in testSuite), to verify that those default params are indeed handled correctly in the back-end, while we're at it. There shouldn't be any reason for them not to work, but since they now go through the barrier of doom that is `flatten`, it might be good to check.

gzm0:
Yep... Though about that... Was too lazy to define a JS class... I don't like JS :P

gzm0:
Updated.

sjrd:
LGTM

Issue #1618, opened 2015-04-28
java.time.* from Java 1.8 is missing


hrj:
Java 8 has number of convenient Date & Time classes modelled on Joda Time. Would be great to be supported in Scala-JS.

Examples:
```
java.time.LocalDate
java.time.Period
```

sjrd:
This requires a *huge* amount of work to port a BSD- or MIT-licensed implementation of those to Scala. The core team doesn't have the manpower to work on this, but PRs are welcome!

hrj:
Thanks for looking and keeping it open; I didn't expect an overnight solution to this. But I have an idea / doubt:

Is it possible to provide parts of the Java standard library as a separate Scala JS library? Will that help with the license restriction? (I am assuming there is some automated way to translate existing Java code / Java-byte-code to a Scala-JS compatible library).

sjrd:
Oh yes, you can certainly provide parts of the JDK as separate Scala.js libraries. You just have to configure the build not to bundle the `.class` files into the `.jar`s, but only the `.sjsir` files.

There is no automated way to translate Java to Scala.js per se. I know there are a few Java-to-Scala converters out there, but they are known to be imperfect at best. If the Scala code does not depend on other non-supported parts of the library, then they should work with Scala.js out-of-the-box.

rparree:
@hrj you might be better of writing a scalajs interface to something like http://momentjs.com/ (or using it adhoc as `js.Dynamic`)

hrj:
@rparree Thanks for the suggestion, but no, I want to share this code on server and client sides, so that I can be sure they are computing the same results, among other reasons.

rparree:
Gotcha..i am in the same situation. At this point i am using JSR 310 on the server and momentjs (via some partial traits) on the client (so had to make different case classes for client and server for this feature)

refried:
Dumb question: What does the license have to do with it?

refried:
I was about to start using scala.js but would specifically need this.

refried:
Joda-Time has an Apache license.  Since Apache Harmony has a compatible license, does Joda-Time?

gzm0:
Yes. Joda-Time would have a compatible license. But since it would probably be best to port it in a separate project anyway, this would not really matter.

refried:
Maybe I misunderstood, but a separate project wouldn't help people using java.time — right?

gzm0:
[...]

PR #1901, opened 2015-09-13
Add first batch of java.time classes

ummels:
This PR addresses #1618 and adds implementations for an initial batch of java.time classes, namely Duration and LocalTime. Since these classes depend on other classes and interfaces, mostly from java.time.temporal, the relevant classes are also implemented.

This PR also adds an implementation of java.lang.Enum in order to not violate the DRY principle (from the initial batch both java.time.temporal.ChronoUnit and java.time.temporal.ChronoField descend from java.lang.Enum).


gzm0:
Jenkins, ok to test

gzm0:
@ummels the tests will fail, since they require JDK8 when they are being compiled (otherwise `java.time` is not even defined). You will have to move them to `test-suite/src/test/require-jdk8`.

gzm0:
Oh, and thank you very much for the contribution, of course :)

ummels:
@gzm0 Good point. I have moved the tests to `test-suite/src/test/require-jdk8`.

sjrd:
@ummels I must ask: did you look at any GPL'ed code when implementing this, such as OpenJDK or OracleJDK's source code?

ummels:
@sjrd No, this is a clean-room implementation as required by the guidelines. I only looked at the API docs.

sjrd:
OK good. I had to check.

sjrd:
The test suite for `Duration` fails in *fullOpt*.
This is probably not your fault since it passes in fastOpt. It's more likely a bug in either a) Closure itself or b) our AST translator when targeting Closure.

I'll need time to debug that one. It probably won't be before September 24th that I can start looking into this.

ummels:
@sjrd Okay, thanks!

javalanglib/src/main/scala/java/lang/Enum.scala
```diff
@@ -0,0 +1,22 @@
+package java.lang
+
+abstract class Enum[E <: Enum[E]] protected(val name: String, val ordinal: Int)
```
sjrd: The parameters should be non-val `_name` and `_ordinal`, and you should have separate definitions of `final def name()` and `final def ordinal()`. This is because https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#parameterless-methods

Also, there should be a space between `protected` and `(`

javalanglib/src/main/scala/java/lang/Enum.scala
```diff
@@ -0,0 +1,22 @@
...
+    case that: Enum[E] if that.getClass == getClass =>
+      ordinal == that.ordinal && name == that.name
+
+    case _ => false
+  }
+
+  override final def hashCode(): Int =
+    31 * name.hashCode() + ordinal.hashCode()
+
+  final def compareTo(o: E): Int = ordinal.compareTo(o.ordinal)
```
sjrd: `protected final def clone()` is missing, between `hashCode` and `compareTo`. It should always throw `CloneNotSupportedException`.
http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#clone%28%29

javalib/src/main/scala/java/time/DateTimeException.scala
```diff
@@ -0,0 +1,4 @@
+package java.time
+
+class DateTimeException(message: String, cause: Throwable = null)
+  extends RuntimeException(message, cause)
```
sjrd: 4 spaces before `extends` instead of 2.

javalib/src/main/scala/java/time/DateTimeException.scala
```diff
@@ -0,0 +1,4 @@
+package java.time
+
+class DateTimeException(message: String, cause: Throwable = null)
```
sjrd: Parameters cannot have default values in JDK classes. You need to declare separate overloads. (default params and overloads are not binary compatible)

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,234 @@
+package java.time
+
+import scala.collection.JavaConverters._
+
+import java.time.temporal._
+
+final class Duration private(seconds: Long, nanos: Int)
```
sjrd: Space between `private` and `(` (there might be occurrences of this in the PR)

sjrd:
The implementation of `java.lang.Enum` should be in its own commit, and you should then also update [`TimeUnit`](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/TimeUnit.scala) and [`RoundingMode`](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/math/RoundingMode.scala) to extend `Enum` in that commit.

@gzm0 Should we also make [`sbt.testing.Status`](https://github.com/scala-js/scala-js/blob/master/test-interface/src/main/scala/sbt/testing/Status.scala) extend `Enum`?

javalanglib/src/main/scala/java/lang/Enum.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  // Not implemented:
+  // final def getDeclaringClass(): Class[E]
+}
```
sjrd: The static method [`valueOf`](http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf%28java.lang.Class,%20java.lang.String%29) is also missing. There should be a comment to mention that.

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,379 @@
+package java.time
+
+import scala.scalajs.js
+
+import java.time.temporal._
+
+final class LocalTime private(hour: Int,
+                              minute: Int,
+                              second: Int,
+                              nano: Int)
```
sjrd: Wrap parameters with 4 spaces. General style here: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#class-declaration

javalib/src/main/scala/java/time/temporal/ChronoField.scala
```diff
@@ -0,0 +1,151 @@
+package java.time.temporal
+
+final class ChronoField private (name: String,
+                                 ordinal: Int,
+                                 val range: ValueRange,
```
sjrd: Same remark as before: there should be a separate `def range()`. There might be other occurrences.

javalib/src/main/scala/java/time/temporal/ChronoField.scala
```diff
@@ -0,0 +1,151 @@
...
+  def adjustInto[R <: Temporal](temporal: R, newValue: Long): R =
+    temporal.`with`(this, newValue).asInstanceOf[R]
+}
+
+object ChronoField {
+
+  import ChronoUnit._
+
+  val NANO_OF_SECOND = new ChronoField("NanoOfSecond", 0,
+    ValueRange.of(0, 999999999), false, NANOS, SECONDS)
```
sjrd: Inside `( )`, indent with 4 spaces: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#indentation

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,379 @@
...
+    extends Temporal with TemporalAdjuster with Comparable[LocalTime]
+    with java.io.Serializable with Preconditions {
+
+  import LocalTime._
+  import ChronoField._
+  import ChronoUnit._
+
+  private val totalNanos = nano.toLong + (1000000000L * second) +
+      (60000000000L * minute) + (3600000000000L * hour)
```
sjrd: Add parentheses around the rhs of the `=`, because it's a long formula that wraps on the next line.

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,379 @@
...
+  require(hour >= 0 && hour <= 23, s"Invalid value for hour: $hour")
+  require(minute >= 0 && minute <= 59, s"Invalid value for minute: $minute")
+  require(second >= 0 && second <= 59, s"Invalid value for second: $second")
+  require(nano >= 0 && nano <= 999999999, s"Invalid value for nanoOfSecond $nano")
+
+  def isSupported(field: TemporalField): Boolean = field match {
+    case NANO_OF_SECOND | NANO_OF_DAY | MICRO_OF_SECOND | MICRO_OF_DAY |
+         MILLI_OF_SECOND | MILLI_OF_DAY | SECOND_OF_MINUTE | SECOND_OF_DAY |
+         MINUTE_OF_HOUR | MINUTE_OF_DAY | HOUR_OF_AMPM | CLOCK_HOUR_OF_AMPM |
+         HOUR_OF_DAY | CLOCK_HOUR_OF_DAY | AMPM_OF_DAY => true
```
sjrd: As weird as it might look, the second to fourth line should be indented 4 spaces, not 5.
Also I'd recommend putting `true` on the next line, indented 2 spaces wrt. `case`

javalib/src/main/scala/java/time/temporal/ChronoField.scala
```diff
@@ -0,0 +1,151 @@
...
+    AMPM_OF_DAY, DAY_OF_WEEK, ALIGNED_DAY_OF_WEEK_IN_MONTH,
+    ALIGNED_DAY_OF_WEEK_IN_YEAR, DAY_OF_MONTH, DAY_OF_YEAR, EPOCH_DAY,
+    ALIGNED_WEEK_OF_MONTH, ALIGNED_WEEK_OF_YEAR, MONTH_OF_YEAR, PROLEPTIC_MONTH,
+    YEAR_OF_ERA, YEAR, ERA, INSTANT_SECONDS, OFFSET_SECONDS)
+
+  def values(): Array[ChronoField] = fields.clone()
+
+  def valueOf(name: String): ChronoField = {
+    fields.find(name == _.name.toScreamingSnakeCase) match {
+      case Some(field) => field
```
sjrd: Don't use pattern matching on `Option`. Use `fold` instead.
https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#option-and-jsundefor

gzm0:
Yes, we probably should make `sbt.testing.Status` extend `Enum`?

javalib/src/main/scala/java/time/temporal/package.scala
```diff
@@ -0,0 +1,11 @@
+package java.time
+
+package object temporal {
+  implicit class RichString(val s: String) extends AnyVal {
```
sjrd: This should be `private[time]`. Also, is it really worth having an implicit class here? I would recommend having a `private[time] def toScreamingSnakeCase(s: String): String` instead.

javalib/src/main/scala/java/time/temporal/TemporalAmount.scala
```diff
@@ -0,0 +1,11 @@
+package java.time.temporal
+
+trait TemporalAmount {
+  def get(unit: TemporalUnit): Long
+
+  def getUnits: java.util.List[TemporalUnit]
```
sjrd: Methods in JDK classes should always have `()`: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#parameterless-methods

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,234 @@
...
+    extends TemporalAmount with Comparable[Duration]
+    with java.io.Serializable with Preconditions {
+
+  require(nanos >= 0 && nanos <= 999999999, "nanos must be >= 0 and <= 999999999")
```
sjrd: Use `assert` instead of `require`, since this is a private method. Callers might have require's, but this constructor shouldn't.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,234 @@
...
+      seconds == that.getSeconds && nanos == that.getNano
+
+    case _ => false
+  }
+
+  override def hashCode(): Int = 31 * seconds.hashCode + nanos
+}
+
+object Duration {
+  private def apply(nanos: BigInt): Duration = {
```
sjrd: You should try and go to greater lengths to avoid using big numbers, especially if they're not part of the public API. You could have an `apply` method taking two `Long`s to multiply together, and have it detect overflow other than through big numbers.

Depends on how hard to implement that is, but if you manage, it would be a huge performance gain, not to mention that it would avoid to pull in the big numbers just for `java.time`.

sjrd:
That's all for the general comments. I'll let @nicolasstucki review the finer implementation.

javalib/src/main/scala/java/time/Duration.scala
ummels: Could you be more specific about this? Note that the package-private trait Preconditions overrides require here to throw a DateTimeException here. Would it be okay to define a custom method `requireDT` in the package object instead?

javalib/src/main/scala/java/time/Duration.scala
ummels: Okay, I'll see what I can do.

javalib/src/main/scala/java/time/Duration.scala
ummels: I should say that other methods call the constructor, so depending on the arguments passed to these methods the constructor will throw a `DateTimeException`. In particular, `withNanos(1000000000)` will throw because the parameter is too large as required by the JavaDoc. Are you saying I should move these checks into the relevant methods?

ummels:
@sjrd Thanks for your comments. I've just updated my PR to address all of them except the ones about `require` and `BigInt`.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    val (quot, rem) = nanos /% 1000000000
+    val seconds = if (rem < 0) quot - 1 else quot
+    val nanosOfSeconds = if (rem < 0) rem + 1000000000 else rem
+    if (seconds.isValidLong) new Duration(seconds.toLong, nanosOfSeconds.toInt)
+    else throw new ArithmeticException("Long overflow")
+  }
+
+  val ZERO = new Duration(0, 0)
+
+  def ofDays(days: Long): Duration = Duration(BigInt(days) * 86400000000000L)
```
nicolasstucki: It might be a good idea to use private functions to represent `86400000000000L` in a more explicit way and let the compiler inline and constant fold them. I see that this could become a potential source of bugs as we could easily miss some 0s. For example `NANOS_IN_SECONDS * SECONDS_IN_MINUTE * MINUTES_IN_HOUR * HOURS_IN_DAY`.

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: You should use the methods in `java.lang.Math` that take care of checking overflows as they do. For example `java.lang.Math.multiplyExact` method.


```scala
scala> java.time.Duration.ofDays(9999999999999999L)
java.lang.ArithmeticException: long overflow
  at java.lang.Math.multiplyExact(Math.java:892)
  at java.time.Duration.ofDays(Duration.java:179)
  ... 43 elided
```

nicolasstucki: You should also try to implement the `ofXXX` methods directly without the `apply`. In most cases you it will just be a couple of invocations to `java.lang.Math` methods.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+  def ofNanos(nanos: Long): Duration = Duration(BigInt(nanos))
+
+  def of(amount: Long, unit: TemporalUnit): Duration = unit match {
+    case _ if !unit.isDurationEstimated =>
+      unit.getDuration.multipliedBy(amount)
+
+    case ChronoUnit.DAYS => ofDays(amount)
+
+    case _ =>
+      throw new UnsupportedTemporalTypeException(s"Unit not supported: $unit")
```
nicolasstucki: The doc says that it should throw a `DateTimeException`.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  def from(amount: TemporalAmount): Duration = {
+    (amount.getUnits.asScala foldLeft Duration.ZERO) { (d, u) =>
```
nicolasstucki: Format issue, we use the dot to invoke the functions.
```scala
amount.getUnits.asScala.foldLeft(Duration.ZERO) { (d, u) =>
```

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,377 @@
+package java.time
+
+import scala.scalajs.js
+
+import java.time.temporal._
+
+final class LocalTime private (hour: Int, minute: Int, second: Int, nano: Int)
+    extends Temporal with TemporalAdjuster with Comparable[LocalTime]
+    with java.io.Serializable with Preconditions {
```
nicolasstucki: You can't change the inheritance definitions of a class ported from java. You need to remove `with Preconditions`, but you could redefine it as `object Preconditions {...}` and the import its method with `import Preconditions.require`.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+  require(nanos >= 0 && nanos <= 999999999, "nanos must be >= 0 and <= 999999999")
+
+  private val totalNanos: BigInt = BigInt(seconds) * 1000000000 + nanos
+
+  def get(unit: TemporalUnit) = unit match {
+    case ChronoUnit.SECONDS => seconds
+
+    case ChronoUnit.NANOS => nanos
+
+    case _ =>
```
nicolasstucki: The specs say that this method could throw a `DateTimeException`, there might be something missing. I can't see that case where it would happen, but we should investigate this, it might be for a corner case.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+
+final class Duration private (seconds: Long, nanos: Int)
+    extends TemporalAmount with Comparable[Duration]
+    with java.io.Serializable with Preconditions {
+
+  require(nanos >= 0 && nanos <= 999999999, "nanos must be >= 0 and <= 999999999")
+
+  private val totalNanos: BigInt = BigInt(seconds) * 1000000000 + nanos
+
+  def get(unit: TemporalUnit) = unit match {
```
gzm0: Public method must have explicit return type

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    case ChronoUnit.SECONDS => seconds
+
+    case ChronoUnit.NANOS => nanos
+
+    case _ =>
+      throw new UnsupportedTemporalTypeException(s"Unit not supported: $unit")
+  }
+
+  def getUnits(): java.util.List[TemporalUnit] =
+    Seq[TemporalUnit](ChronoUnit.SECONDS, ChronoUnit.NANOS).asJava
```
nicolasstucki: Try `Array[TemporalUnit](ChronoUnit.SECONDS, ChronoUnit.NANOS).asJava`, it might be a bit more efficient.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+  def withSeconds(seconds: Long): Duration =
+    new Duration(seconds, nanos)
+
+  def withNanos(nanosOfSecond: Int): Duration =
+    new Duration(seconds, nanosOfSecond)
+
+  private def plus(nanos: BigInt): Duration = Duration(totalNanos + nanos)
+
+  def plus(duration: Duration): Duration =
+    plus(BigInt(duration.getSeconds) * 1000000000 + duration.getNano)
```
nicolasstucki:
```
val sumNanos = nanos + duration.getNano
val newNanos = 
  if (sumNanos <= 999999999) sumNanos
  else sumNanos - 1000000000
val newSeconds = Math.addExact(seconds,
  if (sumNanos <= 999999999) duration.seconds
  else Math.addExact(duration.getSeconds, 1))
new Duration(newSeconds, newNanos)
```

javalib/src/main/scala/java/time/Duration.scala
ummels: I was thinking of using `java.lang.Math.multiplyExact`, but it is not yet implemented as far as I can see?

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    case _ if !unit.isDurationEstimated =>
+      plus(unit.getDuration.multipliedBy(amount))
+
+    case ChronoUnit.DAYS => plusDays(amount)
+
+    case _ =>
+      throw new UnsupportedTemporalTypeException(s"Unit not supported: $unit")
+  }
+
+  def plusDays(days: Long): Duration = plus(BigInt(86400) * days * 1000000000)
```
nicolasstucki: Use `def plus(amount: Long, unit: TemporalUnit)` and let it do the math.
```scala
def plusDays(days: Long): Duration = plus(days, ChronoUnit.DAYS)
```

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  def plusHours(hours: Long): Duration = plus(BigInt(3600) * hours * 1000000000)
```
nicolasstucki: Same

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: Same for all other `plusXXX` and `minusXXX` method

javalib/src/main/scala/java/time/Duration.scala
ummels: If you try it in the REPL, you can see that it actually throws an `UnsupportedTemporalTypeException` (which is a subclass of `DateTimeException`).

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    temporal.minus(seconds, ChronoUnit.SECONDS).minus(nanos, ChronoUnit.NANOS)
+
+  def toDays(): Long = seconds / 86400
+
+  def toHours(): Long = seconds / 3600
+
+  def toMinutes(): Long = seconds / 60
+
+  def toMillis(): Long = {
+    val (quot, rem) = totalNanos /% 1000000
```
nicolasstucki: Try with

```scala
Math.addExact(Math.multiplyExact(seconds, 1000), nanos / 1000000)
```

javalib/src/main/scala/java/time/Duration.scala
ummels: Yes, the spec is ambiguous here.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    val millis = if (rem < 0) quot - 1 else quot
+    // Reference implementation throws if millis < -9223372036854775000!
+    if (millis >= -9223372036854775000L && millis <= Long.MaxValue)
+      millis.toLong
+    else
+      throw new ArithmeticException("Long overflow")
+  }
+
+  def toNanos(): Long = {
+    // Reference implementation throws if nanos < -9223372036000000000!
```
nicolasstucki: Same

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+    if (totalNanos >= -9223372036000000000L && totalNanos <= Long.MaxValue)
+      totalNanos.toLong
+    else
+      throw new ArithmeticException("Long overflow")
+  }
+
+  def compareTo(that: Duration) = totalNanos.compare(that.totalNanos)
```
nicolasstucki:
```scala
val secCmp = seconds.compareTo(that.getSeconds)
if (secCmp == 0) nanos.compareTo(that.getNano)
else secCmp
```

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,232 @@
...
+
```
nicolasstucki: `equals` and `hashCode` should be here, not below `toString`.

javalib/src/main/scala/java/time/Duration.scala
ummels: Doesn't compile. It also returns an immutable list whereas a wrapped array would be mutable IMHO.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/LocalTime.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: I've put these constants into a new package-private object `java.time.Constant`.

javalanglib/src/main/scala/java/lang/Enum.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override final def equals(that: Any): scala.Boolean = that match {
+    case that: Enum[E] if that.getClass == getClass =>
+      _ordinal == that.ordinal() && _name == that.name()
+
+    case _ => false
+  }
+
+  override final def hashCode(): Int =
+    31 * _name.hashCode() + _ordinal.hashCode()
```
sjrd: Enum values are only ever equal to strictly themselves, so `equals` and `hashCode` can simply call `super.{equals,hashCode}` to inherit `Object`'s definitions (while still being `final`). I recommend marking them `@inline`:
```scala
@inline
override final def equals(that: Any): scala.Boolean = super.equals(that)

@inline
override final def hashCode(): Int = super.hashCode()
```

Although the JavaDoc spec is a bit scarce, experimentation in a REPL suggests that this is also what the JDK does.

javalib/src/main/scala/java/math/RoundingMode.scala
```diff
@@ -25,12 +25,8 @@ package java.math
...
 import scala.annotation.switch
 
-final class RoundingMode private (_index: Int, _name: String) {
-  def name(): String = _name
-  def ordinal(): Int = _index
-
-  override def toString(): String = name()
-}
+final class RoundingMode private (_index: Int, _name: String)
+    extends Enum[RoundingMode](_name, _index)
```
sjrd: The Boy Scout Rule suggests that the order of the parameters of `RoundingMode` be reversed to match that of `Enum`.
The same applies to `TimeUnit` and `sbt.testing.Status`.

nicolasstucki:
Jenkins failed to build because we added the `-Xfatal-warnings` flag. Here are the warnings/errors:

```
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/Preconditions.scala:6:4: No space after token if
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/Duration.scala:138:6: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/Duration.scala:195:6: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoUnit.scala:17:15: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:47:23: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:53:23: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:104:21: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:104:36: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:116:21: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/temporal/ChronoField.scala:116:35: Use an uppercase L for long literals
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/LocalTime.scala:258:6: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/LocalTime.scala:353:6: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/javalib/src/main/scala/java/time/LocalTime.scala:360:6: Public method must have explicit type
```

nicolasstucki:
Actually, those are errors thrown by `scalatype`. Therefore just format issues.

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: Ok

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,240 @@
...
+    val (quot, rem) = nanos /% NANOS_IN_SECOND
+    val seconds = if (rem < 0) quot - 1 else quot
+    val nanosOfSeconds = if (rem < 0) rem + NANOS_IN_SECOND else rem
+    if (seconds.isValidLong) new Duration(seconds.toLong, nanosOfSeconds.toInt)
+    else throw new ArithmeticException("Long overflow")
+  }
+
+  val ZERO = new Duration(0, 0)
+
+  def ofDays(days: Long): Duration = Duration(BigInt(days) * NANOS_IN_DAY)
```
nicolasstucki: You could implement all the `ofXXX` methods without the need of an intermediate `BigInt` by using `of(amount, unit)`. 
```scala
def ofDays(days: Long): Duration = of(days, ChronoUnit.DAYS)
def ofHours(hours: Long): Duration = of(days, ChronoUnit.HOURES)
...
```
Of course currently `of` is using `multipliedBy` which uses `BigInt`s, but tat should also change if we want a decent performance.

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,381 @@
...
+
+    case _: ChronoUnit =>
+      throw new UnsupportedTemporalTypeException(s"Unit not supported: $unit")
+
+    case _ => unit.addTo(this, amount).asInstanceOf[LocalTime]
+  }
+
+  def plusHours(hours: Long): LocalTime = {
+    val offset = if (hours < 0) hours % 24 + 24 else hours % 24
+    new LocalTime((hour + offset.toInt) % 24, minute, second, nano)
```
nicolasstucki: You could improve performance by doing operations on `Int` rather than longs:
```scala
val offset = (hours % 24).toInt + 24
new LocalTime((hour + offset) % 24, minute, second, nano)
```
In Scala.js `Int` operations are more performant than `Long` operations.

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  def plusMinutes(minutes: Long): LocalTime =
+    plusNanos((minutes % MINUTES_IN_DAY) * NANOS_IN_MINUTE)
+
+  def plusSeconds(seconds: Long): LocalTime =
+    plusNanos((seconds % SECONDS_IN_DAY) * 1000000000)
```
nicolasstucki: Use `NANOS_IN_MINUTE`

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,381 @@
...
+
+  def plusNanos(nanos: Long): LocalTime = {
+    val offset =
+      if (nanos < 0) nanos % NANOS_IN_DAY + NANOS_IN_DAY
+      else nanos % NANOS_IN_DAY
```
nicolasstucki: It is enough to do:
```scala
val offset = nanos % NANOS_IN_DAY + NANOS_IN_DAY
```

javalib/src/main/scala/java/time/LocalTime.scala
```diff
@@ -0,0 +1,381 @@
...
+
+      case _ => unit.between(this, endTime)
+    }
+  }
+
+  // Not implemented
+  // def format(format: DateTimeFormatter): String
+
+  // TODO
+  // def atLocalDate(date: LocalDate): LocalDateTime
```
nicolasstucki: This should be `def atDate(date: LocalDate): LocalDateTime`.

javalib/src/main/scala/java/time/Preconditions.scala
```diff
@@ -0,0 +1,9 @@
+package java.time
+
+private[time] object Preconditions {
+  // Like scala.Predef.require, but throws a DateTimeException.
+  def require(requirement: Boolean, message: => Any): Unit = {
```
nicolasstucki: Maybe it would be good to rename to avoid ambiguity. Could be `requireDateTime`.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,240 @@
...
+  def minusMinutes(minutes: Long): Duration = minus(minutes, MINUTES)
+
+  def minusSeconds(seconds: Long): Duration = minus(seconds, SECONDS)
+
+  def minusMillis(millis: Long): Duration = minus(millis, MILLIS)
+
+  def minusNanos(nanos: Long): Duration = minus(nanos, NANOS)
+
+  def multipliedBy(multiplicand: Long): Duration =
+    Duration(totalNanos * multiplicand)
```
nicolasstucki: You could do somthing like this to avoid using `BigInt`:
```scala
val mulNanos = Math.multiplyExact(multiplicand, nanos)
val mulSeconds = Math.multiplyExact(multiplicand, seconds)
val newNanos = mulNanos % NANOS_IN_SECOND
val newSeconds = Math.addExact(mulSeconds, mulNanos / NANOS_IN_SCONDS)
new Duration(newSeconds, newNanos)
```
And maybe take care of negative numbers.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  def dividedBy(divisor: Long): Duration = Duration(totalNanos / divisor)
```
nicolasstucki: Get rid of the use of `BigInt`.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  def negated(): Duration = Duration(totalNanos * -1)
```
nicolasstucki: Get rid of the use of `BigInt`.

nicolasstucki:
That is all.

javalanglib/src/main/scala/java/lang/Enum.scala
ummels: Okay, fixed.

javalib/src/main/scala/java/math/RoundingMode.scala
ummels: Done.

nicolasstucki:
Some tests are failing:
```scala
[info] java.time.Duration
[error]  x should respond to `get`
org.scalajs.jasminetest.JasmineTestException: Expected 9223372036854776000 to equal 9223372036854775000.
	at <jscode>.new jasmine.ExpectationResult([stdin]:216)
	at <jscode>.null.toEqual([stdin]:1337)
	at <jscode>.LPa.e.k(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:26)
	at <jscode>.LPa.e.t(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:17)
	at <jscode>.null.<anonymous>(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/jasmine-test-framework/src/main/scala/org/scalajs/jasminetest/JasmineTest.scala:21)
	at <jscode>.jasmine.Block.execute([stdin]:1166)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
	at <jscode>.jasmine.Queue.start([stdin]:2151)
	at <jscode>.jasmine.Spec.execute([stdin]:2478)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
...
[error]  x should respond to `getSeconds`
org.scalajs.jasminetest.JasmineTestException: Expected 9223372036854776000 to equal 9223372036854775000.
	at <jscode>.new jasmine.ExpectationResult([stdin]:216)
	at <jscode>.null.toEqual([stdin]:1337)
	at <jscode>.{anonymous}()(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:80)
	at <jscode>.WU.t(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/library/src/main/scala/scala/scalajs/runtime/AnonFunctions.scala:10)
	at <jscode>.null.<anonymous>(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/jasmine-test-framework/src/main/scala/org/scalajs/jasminetest/JasmineTest.scala:21)
	at <jscode>.jasmine.Block.execute([stdin]:1166)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
	at <jscode>.jasmine.Queue.start([stdin]:2151)
	at <jscode>.jasmine.Spec.execute([stdin]:2478)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
...
[error]  x should respond to `toMillis`
org.scalajs.jasminetest.JasmineTestException: Expected 9223372036854776000 to equal 9223372036854775000.
	at <jscode>.new jasmine.ExpectationResult([stdin]:216)
	at <jscode>.null.toEqual([stdin]:1337)
	at <jscode>.fQa.e.k(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:683)
	at <jscode>.fQa.e.t(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:671)
	at <jscode>.null.<anonymous>(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/jasmine-test-framework/src/main/scala/org/scalajs/jasminetest/JasmineTest.scala:21)
	at <jscode>.jasmine.Block.execute([stdin]:1166)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
	at <jscode>.jasmine.Queue.start([stdin]:2151)
	at <jscode>.jasmine.Spec.execute([stdin]:2478)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
[error]  x should respond to `toNanos`
org.scalajs.jasminetest.JasmineTestException: Expected 9223372036854776000 to equal 9223372036854775000.
	at <jscode>.new jasmine.ExpectationResult([stdin]:216)
	at <jscode>.null.toEqual([stdin]:1337)
	at <jscode>.gQa.e.k(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:701)
	at <jscode>.gQa.e.t(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala:692)
	at <jscode>.null.<anonymous>(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/target/scala-2.11/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/jasmine-test-framework/src/main/scala/org/scalajs/jasminetest/JasmineTest.scala:21)
	at <jscode>.jasmine.Block.execute([stdin]:1166)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
	at <jscode>.jasmine.Queue.start([stdin]:2151)
	at <jscode>.jasmine.Spec.execute([stdin]:2478)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
```

javalib/src/main/scala/java/time/LocalTime.scala
ummels: Thanks.

javalib/src/main/scala/java/time/LocalTime.scala
ummels: Done.

javalib/src/main/scala/java/time/LocalTime.scala
ummels: Ok.

javalib/src/main/scala/java/time/LocalTime.scala
ummels: Thanks.

javalib/src/main/scala/java/time/Preconditions.scala
ummels: Done.

ummels:
Yes, there is something strange going on with `Long.MaxValue`.

test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/DurationTest.scala
```diff
@@ -0,0 +1,887 @@
...
+    it("should respond to `get`") {
+      val d0 = ofSeconds(0, 1)
+
+      expect(dmin.get(SECONDS)).toEqual(Long.MinValue)
+      expect(dmin.get(NANOS)).toEqual(0)
+      expect(ZERO.get(SECONDS)).toEqual(0)
+      expect(ZERO.get(NANOS)).toEqual(0)
+      expect(d0.get(SECONDS)).toEqual(0)
+      expect(d0.get(NANOS)).toEqual(1)
+      expect(dmax.get(SECONDS)).toEqual(Long.MaxValue)
```
ummels: This test seems to fail under some circumstances because `dmax.get(SECONDS)` evaluates to `9223372036854776000` and `Long.MaxValue` evaluates to `9223372036854775000`. In fact, they both should evaluate to `9223372036854775807`, or not?

sjrd: Oh. You may not use `.toEqual` with `Long`s. You have to write it like this:
```scala
expect(dmax.get(SECONDS) == Long.MaxValue).toBeTruthy
```
otherwise both the lhs and rhs are converted to Doubles to be given to the Jasmine test framework.

ummels: I see. Although I find it strange that these tests succeed on my local machine.

sjrd: From the log, it seems they fail only in fullOpt. You might make them fail on your local machine in fullOpt as well.

ummels: Yes, I haven't tested with fullOpt so far.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -38,29 +72,133 @@ object Math {
...
+  }
+
+  def multiplyExact(a: scala.Long, b: scala.Long): scala.Long = {
+    val overflow = {
+      b > 0 && (a > Long.MAX_VALUE / b || a < Long.MIN_VALUE / b) ||
+      b < -1 && (a > Long.MIN_VALUE / b || a < Long.MAX_VALUE / b) ||
+      b == -1 && a == Long.MIN_VALUE
+    }
+    //if (!overflow) a * b // Disabled because of #1918
+    val res = BigInt(a) * b
```
ummels: This use of `BigInt` can be removed when #1918.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,271 @@
...
+    val (prodNanosQuot, prodNanosRem) = {
+      try {
+        val mulNanos = Math.multiplyExact(normalizedNanos, multiplicand)
+        (mulNanos / NANOS_IN_SECOND, (mulNanos % NANOS_IN_SECOND).toInt)
+      } catch {
+        case _: ArithmeticException =>
+          val mulNanos = BigInt(normalizedNanos) * multiplicand
+          ((mulNanos / NANOS_IN_SECOND).toLong, (mulNanos % NANOS_IN_SECOND).toInt)
+      }
+    }
```
ummels: Note that `BigInt` is only used when `Long` overflows.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,271 @@
...
+            normalizedNanos)
+        total / divisor
+      } catch {
+        case _: ArithmeticException =>
+          val total = BigInt(secondsRem) * NANOS_IN_SECOND + normalizedNanos
+          (total / divisor).toLong
+      }
+    }
+    Duration.ofSeconds(secondsQuot).plusNanos(nanos)
+  }
```
ummels: Same here.

javalib/src/main/scala/java/time/Duration.scala
```diff
@@ -0,0 +1,271 @@
...
+  import Constants._
+
+  final val ZERO = new Duration(0, 0)
+
+  private final val MIN = new Duration(Long.MinValue, 0)
+
+  private final val MAX = new Duration(Long.MaxValue, 999999999)
+
+  def ofDays(days: Long): Duration = // of(days, ChronoUnit.DAYS)
+    ofSeconds(SECONDS_IN_DAY).multipliedBy(days)
```
ummels: If I use `of(days, ChronoUnit.DAYS)` many tests are failing, because `ChronoUnit.DAYS` sometimes evaluates to `null`. This looks like a bug to me. Could one of you investigate?

ummels:
I've implemented the `...Exact` methods in `java.lang.Math` and used them in the implementation of `java.time.Duration`.

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: Wouldn't it be more direct to multiply the `seconds` and `nanos` apart. Then remove the excedent from the `nanos` multiplication and add it to the seconds.

There would be to overflow possible. The on on `seconds`, that is unavoidable and means that `multipliedBy` has overflown. The one on `nanos` must be handled explicitly because part of that could go to the seconds.

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: Clearly it fail because `of` uses `ChronoUnit.DAYS` and `ChronoUnit.DAYS` is initialised using `Duration.ofDays(1)`. 

I would suggest to remove the `from` the fields of `ChronoUnit` and implement `def getDuration(): Duration` as a switch on `ordinal`.
```scala
def getDuration(): Duration = (ordina @switch) match {
  case 0 => ... // The singleton for the duration
 ...
}
```

nicolasstucki:
Jenkins, retest this please

javalib/src/main/scala/java/time/Duration.scala
ummels: Hah, that makes sense!

javalib/src/main/scala/java/time/Duration.scala
ummels: I think that's exactly what I'm doing. `prodNanos`ist just the product of the (normalized) `nano` part with the multiplicand. The quotient `prodNanosQuot` is then added to the product of the seconds part stored in `prodSeconds`.

nicolasstucki: Now I see. That is right.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -4,25 +4,59 @@ package lang
...
 object Math {
-  final val E  = 2.718281828459045
+  final val E = 2.718281828459045
   final val PI = 3.141592653589793
 
-  @inline def abs(a: scala.Int): scala.Int = if (a < 0) -a else a
-  @inline def abs(a: scala.Long): scala.Long = if (a < 0) -a else a
-  @inline def abs(a: scala.Float): scala.Float = js.Math.abs(a).toFloat
-  @inline def abs(a: scala.Double): scala.Double = js.Math.abs(a)
+  @inline def sin(a: scala.Double): scala.Double = js.Math.sin(a)
```
sjrd: Could you avoid all the irrelevant changes, such as those reorderings and new blank lines? The diff is basically unreadable right now. Just add the new methods without touching the rest of the file.

ummels: Sorry, I tried to applied boy scout rule here since the methods were not ordered as laid out in the JavaDoc. Should I use the old ordering and just append my methods at the end?

sjrd: I understand. The boy scout rule is good to apply unless it makes a diff too hard to read. In this case, yes, please preserve the old ordering, and just append the new methods at the end.

nicolasstucki: @sjrd maybe he could split it int two commits, one where he only reorders and and where he does the changes.

ummels: @sjrd Done.

@nicolasstucki Saw your message too late, but I can still do the reordering in a new commit if you wish.

nicolasstucki:
@ummels don't worry about the method order in Math. We could do it later in a general refactoring.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -219,5 +219,100 @@ object Math {
...
+      b > 0 && a > Integer.MAX_VALUE - b || b < 0 && a < Integer.MIN_VALUE - b
+    if (!overflow) a + b
+    else throw new ArithmeticException("Integer overflow")
+  }
+
+  def addExact(a: scala.Long, b: scala.Long): scala.Long = {
+    val overflow =
+      b > 0 && a > Long.MAX_VALUE - b || b < 0 && a < Long.MIN_VALUE - b
+    if (!overflow) a + b
+    else throw new ArithmeticException("Long overflow")
```
sjrd: You can minimize the number of operations on Longs by doing something like this instead:
```scala
val r = a + b
val rSign = Long.signum(r)
if (rSign == Long.signum(a) || rSign == Long.signum(b)) r
else throw new ArithmeticException("Long overflow")
```
`Long.signum` is intrinsified by the optimizer to directly work on the internals of the `Long` representation, and returns cached "instances" of -1, 0 and 1.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -219,5 +219,100 @@ object Math {
...
+      b < 0 && a > Integer.MAX_VALUE + b || b > 0 && a < Integer.MIN_VALUE + b
+    if (!overflow) a - b
+    else throw new ArithmeticException("Integer overflow")
+  }
+
+  def subtractExact(a: scala.Long, b: scala.Long): scala.Long = {
+    val overflow =
+      b < 0 && a > Long.MAX_VALUE + b || b > 0 && a < Long.MIN_VALUE + b
+    if (!overflow) a - b
+    else throw new ArithmeticException("Long overflow")
```
sjrd: You can do something similar here:
```scala
val r = a - b
val rSign = Long.signum(r)
if (rSign == Long.signum(a) || rSign != Long.signum(b)) r
else throw new ArithmeticException("Long overflow")
```

javalib/src/main/scala/java/time/Duration.scala
ummels: What do you mean by "The singleton for the duration"? If I define `getDuration` as follows, I still run into an infinite loop because `Duration.of`calls `getDuration`again, right?
```scala
def getDuration(): Duration = (ordinal @switch) match {
  case 0 => Duration.ofNanos(1)
  ...
  case 7 => Duration.ofDays(1)
```

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Thanks. I needed to add an additional check whether `rSign == 0 && a != Long.MinValue && b != Long.MinValue`, but with that it should work.

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Thanks.

javalib/src/main/scala/java/time/Duration.scala
nicolasstucki: I mean, create private constants for `ChronoUnit.OneDay`, `ChronoUnit.OneSecond`, ... , int the companion object. They could be lazy. We want to avoid the use `Duration.ofXXX` as they may be implemented using `ChronoUnit`.
```scala
def getDuration(): Duration = (ordinal @switch) match {
  case 0 => ChronoUnit.OneDay
  ...
  case 7 => ChronoUnit.OneSecond
```

```scala
object ChronoUnit {
  private[ChronoUnit] val OneSecond = new Duration(...)
  ...
}
```

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -219,5 +219,102 @@ object Math {
...
+        resSgn == 0 && a != Integer.MIN_VALUE && b != Integer.MIN_VALUE) res
+    else throw new ArithmeticException("Integer overflow")
+  }
+
+  def addExact(a: scala.Long, b: scala.Long): scala.Long = {
+    val res = a + b
+    val resSgn = Long.signum(res)
+    if (resSgn == Long.signum(a) || resSgn == Long.signum(b) ||
+        resSgn == 0 && a != Long.MIN_VALUE && b != Long.MIN_VALUE) res
+    else throw new ArithmeticException("Long overflow")
```
sjrd: Hum, it seems I messed up ^^ Good catch!

Then maybe consider this alternative formulation where we use only the *sign bit*, rather than the sign:
```scala
val resSgnBit = res < 0
if (resSgnBit == (a < 0) || resSgnBit == (b < 0)) res
else ...
```
On second thought, this should even be faster than my earlier proposal. (and it certainly *will* be faster if we special-case `someLong < 0` in the optimizer some day).

The same applies to `Int`, and also to `subtractExact`.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -219,5 +219,102 @@ object Math {
...
+    val resSgn = Long.signum(res)
+    if (resSgn == Long.signum(a) || resSgn != Long.signum(b)) res
+    else throw new ArithmeticException("Long overflow")
+  }
+
+  def multiplyExact(a: scala.Int, b: scala.Int): scala.Int = {
+    val overflow = {
+      b > 0 && (a > Integer.MAX_VALUE / b || a < Integer.MIN_VALUE / b) ||
+          b < -1 && (a > Integer.MIN_VALUE / b || a < Integer.MAX_VALUE / b) ||
+          b == -1 && a == Integer.MIN_VALUE
```
sjrd: Don't indent the 2nd and 3rd lines. Example here: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#long-expressions-with-binary-operators

However, please do add `()` around `&&`s that have higher priority than `||`s. The higher priority of `&&` over `||` is non-obvious and hard to visually parse.

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Nice.

javalanglib/src/main/scala/java/lang/Math.scala
sjrd: Actually, I would recommend if/else's at the "top-level" to short-cut the fast path all the way to the end as fast as possible:
```scala
if (b > 0)
  (a > Integer.MAX_VALUE / b || a < Integer.MIN_VALUE / b)
else if (b < -1)
  (a > Integer.MIN_VALUE / b || a < Integer.MAX_VALUE / b)
else if (b == -1)
  (a == Integer.MIN_VALUE)
else
  false
```
The idea is that if `b > 0` and there *isn't* an overflow, then we avoid having to check `b < -1` and `b == -1` altogether.

ummels: Fixed.

javalib/src/main/scala/java/time/Duration.scala
ummels: Okay, I've put these constants into the companion object of `Duration` instead because I think they have a better place there.

ummels:
It seems tests are failing because `Math.addExact` and `Math.multiplyExact` do not exist in JDK <1.8. How can I disable compilation of the `java.time` classes for these JDK versions?

sjrd:
Ah ah! Now that's a good one ^^ I'll have to think about it.

ummels:
One option would be to put the math methods into some private object (making `java.lang.Math` just wrap these methods) so that I can use them in the implementation of `java.time.Duration` under any JDK.

nicolasstucki:
That idea might work.

ummels:
Another option would be to create a `require-jdk8` source directory inside `javalib` as done for `test-suite` and conigure SBT in a such way that it only includes those sources when on Java 8.

nicolasstucki:
The solution `require-jdk8` would not work because then you still would need an implementation of `java.time` outside of `require-jdk8` that would use the methods of `Math` that are only in jdk8. I would try to do it with your wrapper idea.

ummels:
My suggestion is not to put `java.lang.Math` into `require-jdk8`, but the `java.time` classes, which are not part of older JDK versions anyway. Of course, an advantage of the wrapper solution is that other implementations of Java classes could profit from the new methods in `java.lang.Math` as well.

@sjrd Have you made up your mind about this?

sjrd:
> My suggestion is not to put java.lang.Math into require-jdk8, but the java.time classes, which are not part of older JDK versions anyway.

But that still means that when *building* on JDK 6, we do not end up with the same published artifacts as when we build on JDK 8. That would be very dangerous and fragile.

The wrapper idea does not have that fragility, but it has other downsides:
* There actually isn't any reasonable package to put this class in. We don't want to extend `java.*`, but it needs to be accessible from `java.*`.
* With that implem, not all call paths end up in `java.lang.Math` themselves, which means that intrinsifying `java.lang.Math.addExact` would miss some other call paths.

I guess the least bad solution would be something similar to the [hack for `TypedArrayBufferBridge`](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/scala/scalajs/js/typedarray/TypedArrayBufferBridge.scala): have this in `javalanglib/`:
```scala
package java.lang

private[java] object MathJDK8Bridge {
  def addExact(a: Int, b: Int): Int = Math.addExact(a: Int, b: Int)
  ...
}
```
and this in `javalib/`:
```scala
package java.lang

private[java] object MathJDK8Bridge {
  def addExact(a: Int, b: Int): Int = sys.error("stub")
  ...
}
```
and then hack the Build to discard `MathJDK8Bridge.sjsir` coming from `javalib/`.

ummels:
I have implemented your idea with the `MathJDK8Bridge`. Now only a few builds are failing because of a binary incompatibility arising from `sbt.testing.Status` now being a subclass of `java.lang.Enum` (and two builds seem to fail to due to some timeout). Any advice on how to proceed?

sjrd:
```
[info] Scala.js test interface: found 5 potential binary incompatibilities
[error]  * the type hierarchy of class sbt.testing.Status has changed in new version. Missing types {java.lang.Object}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("sbt.testing.Status")
[error]  * method ordinal()Int in class sbt.testing.Status does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("sbt.testing.Status.ordinal")
[error]  * method name()java.lang.String in class sbt.testing.Status does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("sbt.testing.Status.name")
[error]  * method this(Int,java.lang.String)Unit in class sbt.testing.Status's type has changed; was (Int,java.lang.String)Unit, is now: (java.lang.String,Int)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.testing.Status.this")
[error]  * method toString()java.lang.String in class sbt.testing.Status does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("sbt.testing.Status.toString")
java.lang.RuntimeException: Scala.js test interface: Binary compatibility check failed!
```
I believe all those binary incompat issues are spurious.
* The `this()` is `private` in Scala, so it could not be accessed outside the compilation unit.
* The 4 other signatures are in fact inherited from `java.lang.Enum`

So I think you can blacklist them. First add this to [`project/BinaryIncompatibilities.scala`](https://github.com/scala-js/scala-js/blob/master/project/BinaryIncompatibilities.scala) (at the end):
```scala
  val TestInterface = Seq(
      // Private things, not an issue
      ProblemFilters.exclude[IncompatibleMethTypeProblem](
          "sbt.testing.Status.this"),

      // Inherited from parent class, not an issue
      ProblemFilters.exclude[MissingTypesProblem](
          "sbt.testing.Status"),
      ProblemFilters.exclude[MissingMethodProblem](
          "sbt.testing.Status.name"),
      ProblemFilters.exclude[MissingMethodProblem](
          "sbt.testing.Status.ordinal"),
      ProblemFilters.exclude[MissingMethodProblem](
          "sbt.testing.Status.toString")
  )
```
then use this in `project/Build.scala` [here](https://github.com/scala-js/scala-js/blob/master/project/Build.scala#L819):
```scala
          previousArtifactSetting,
          binaryIssueFilters ++= BinaryIncompatibilities.TestInterface
```

ummels:
Thanks. I take it that the first incompatibility about the changed type hierarchy (subclassing `java.lang.Enum` instead of `java.lang.Object`) is not a problem either?

sjrd:
No because `java.lang.Enum extends java.lang.Object`, so *transitively* `Status` still extends `Object`. I don't know why MiMa reports those as incompatibilities in the first place. Maybe it's a little bit too conservative about that (or it simply does not know; I don't think it's given the full classpath, now that I think about it).

ummels:
The `scala-2.12.0-M2` build is still failing due to a timeout. Any ideas?

sjrd:
Probably spurious. Jenkins, retest this please

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -219,5 +219,110 @@ object Math {
...
+      if (b > 0)
+        a > Long.MAX_VALUE / b || a < Long.MIN_VALUE / b
+      else if (b < -1)
+        a > Long.MIN_VALUE / b || a < Long.MAX_VALUE / b
+      else if (b == -1)
+        a == Long.MIN_VALUE
+      else
+        false
+    }
+    //if (!overflow) a * b // Disabled because of #1918
```
sjrd: For this, I think we should wait for #1923 to be merged, then rebase this PR and adapt this.

ummels: Yes, I agree.

nicolasstucki: @ummels #1923 has been merged. You can now rebase.

ummels:
I've rebased my pull request. BTW should I inline all the one-line methods?

nicolasstucki:
It is not necessary to explicitly `@inline` one liner methods. The Scala.js compiler will automatically inline all small methods. One liners are in most cases small.

nicolasstucki:
LGTM

japgolly:
Wow. Great work! :clap:

ummels:
Thanks! :blush:

ngbinh:
Awesome! Thanks! :clap:

PR #1910, opened 2015-09-22
Fix #1913: Add Thread.{interrupt,isInterrupted,interrupted}.

### LINKED ISSUE #1913: Thread.{interrupt,isInterrupted,interrupted}
`Thread.interrupt()` is sometimes used in cross-compiling codebases in cases like:
```scala
try {
  doSomething()
} catch {
  case _: InterruptedException =>
    Thread.currentThread().interrupt()
}
```
To enable that kind of code to seamlessly cross-compile, we should support `Thread.interrupt()`. In the process, `isInterrupted()` and `interrupted()` should also be implemented.

andreaTP:
Added the method ```interrupt``` to Thread class, just to achieve better source to source compatibility.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Thread.scala
+++ b/javalanglib/src/main/scala/java/lang/Thread.scala
@@ -6,9 +6,16 @@ package java.lang
  * So we use a binary signature that no Java source file can ever produce.
  */
 class Thread private (dummy: Unit) extends Runnable {
+  private var interruptedState = false
   private[this] var name: String = "main" // default name of the main thread
 
   def run(): Unit = ()
+
+  def interrupt(): Unit =
+    interruptedState = true
+
+  def isInterrupted(): scala.Boolean =
+    interruptedState
 
   final def setName(name: String): Unit =
     this.name = name
@@ -26,4 +33,10 @@ object Thread {
   private[this] val SingleThread = new Thread(())
 
   def currentThread(): Thread = SingleThread
+
+  def interrupted(): scala.Boolean = {
+    val ret = currentThread.isInterrupted
+    currentThread.interruptedState = false
+    ret
+  }
 }
--- a/test-suite/src/test/scala/org/scalajs/testsuite/javalib/lang/ThreadTest.scala
+++ b/test-suite/src/test/scala/org/scalajs/testsuite/javalib/lang/ThreadTest.scala
@@ -25,11 +25,23 @@ object ThreadTest extends JasmineTest {
     }
 
     it("Thread.currentThread().getStackTrace() should exist and not crash") {
-      java.lang.Thread.currentThread().getStackTrace()
+      Thread.currentThread().getStackTrace()
     }
 
     it("getId()") {
       expect(Thread.currentThread().getId > 0).toBeTruthy
+    }
+
+    it("interrupt() exist and the status is properly reflected") {
+      val t = Thread.currentThread()
+      expect(t.isInterrupted()).toBeFalsy
+      expect(Thread.interrupted()).toBeFalsy
+      expect(t.isInterrupted()).toBeFalsy
+      t.interrupt()
+      expect(t.isInterrupted()).toBeTruthy
+      expect(Thread.interrupted()).toBeTruthy
+      expect(t.isInterrupted()).toBeFalsy
+      expect(Thread.interrupted()).toBeFalsy
     }
   }
 


sjrd:
According to http://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#interrupt() this should set the interrupted status of the thread.

That said, I'm really wondering why we would need this. Could you provide a concrete example where it makes sense *not* to have a linking error? Ideally this should be in an issue, btw.

andreaTP:
An example use case is this:
https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/dungeon/FaultHandling.scala#L300
Basically the expected behaviour is to stop the actual Thread and continue on the others.
I think this maps quite closely, in a mono thread env, to "don't do anything" and continue the execution.

Having the flag is not a problem but it must be always ```true``` as it is not possible to really interrupt the global thread without an un-useful ```while(true) {}```.

If the linking error is desired and really wanted I definitely need a way to overcome this, at least I must be able to provide in the scope something like:
```
implicit def makeThreadInterruptable(t: Thread) = new {
  def interrupt(): Unit = ()
}
```
to avoid the linking error.

sjrd:
OK. Then I think we should also implement `interrupted()` and `isInterrupted()` for consistency, and have tests for these things.

gzm0:
Do we have any calls somewhere in the Javalib that should throw `InterruptedExceptions`? Because if so, we need to handle and test these as well.

andreaTP:
@sjrd ok, as said ```interrupted()``` will always return ```true``` do you agree?
@gzm0 couldn't be thrown because is only a facade and the Thread couldn't be really interrupted...

gzm0:
@andreaTP yes, it can:

```scala
object Test extends App {
  synchronized {
    Thread.currentThread().interrupt()
    wait() // throws InterruptedException
  }
}
```

I know that `wait` does not link in Scala.js, but if any such method were to exist, it must throw if it is invoked after a call to `interrupt`.

Further, I do not understand what you mean by facade? This is definitely not a facade type as we refer to it in JS interoperability. Could you explain what you mean?

gzm0:
I also do not agree that `interrupted` always needs to return `true`. Consider the following:

```scala
object Test extends App {
  def isInterrupted() = {
    println("isInterrupted() = " + Thread.currentThread.isInterrupted())
  }

  def interrupted() = {
    println("interrupted() = " + Thread.interrupted())
  }

  def interrupt() = {
    println("interrupt()")
    Thread.currentThread.interrupt()
  }

  isInterrupted()
  interrupted()
  isInterrupted()
  interrupt()
  isInterrupted()
  interrupted()
  isInterrupted()
}
```

This will print:

```
isInterrupted() = false
interrupted() = false
isInterrupted() = false
interrupt()
isInterrupted() = true
interrupted() = true
isInterrupted() = false
```

andreaTP:
@gzm0 the expected result in a JVM environment on the same thread will be
```
isInterrupted() = false
interrupted() = false
isInterrupted() = false
interrupt()
//blocked here
```

In JS environment as long as the Thread could not be blocked the result will be:
```
isInterrupted() = false
interrupted() = false
isInterrupted() = false
interrupt() //do nothing
isInterrupted() = false
interrupted() = false
isInterrupted() = false
```

sjrd:
I do not think we have any method that is implemented and that would have to throw an `InterruptedException`. By design (I believe), this exception is thrown by core *blocking* methods if there is an interrupt. But we don't have any blocking method, obviously. It seems that only `Object.wait`, `Thread.sleep` and `Thread.join` can throw this.

sjrd:
@andreaTP No the snippet shown by @gzm0 *on the JVM*, on the same thread, prints exactly what he showed: false false false true true false. If you read the docs of `interrupt()`, `interrupted()` and `isInterrupted()`, it is obvious that this is the expected behavior.

gzm0:
There is definitely I/O stuff that can throw `InterruptedException` (for example `InputStream.read`), but I agree that it is very unlikely that we run into a situation where we **must** throw such an exception, since we do not have blocking (and wont ever).

andreaTP:
ok, agree, sorry...I'm sick ;-)
going to do the implementation and proper tests

andreaTP:
tell me if there is anything else to do here!

javalanglib/src/main/scala/java/lang/Thread.scala
```diff
@@ -20,10 +20,24 @@ class Thread private (dummy: Unit) extends Runnable {
     scala.scalajs.runtime.StackTrace.getCurrentStackTrace()
 
   def getId(): scala.Long = 1
+
+  private var interruptedState = false
```
sjrd: Fields should be declared at the top of the class (under `name` in this case).

javalanglib/src/main/scala/java/lang/Thread.scala
```diff
@@ -20,10 +20,24 @@ class Thread private (dummy: Unit) extends Runnable {
...
+
+  def interrupt(): Unit =
+    interruptedState = true
+
+  def isInterrupted(): scala.Boolean =
+    interruptedState
```
sjrd: To match the JavaDoc's order, these two methods should be put between `run()` and `setName()`.

sjrd:
Finally, use the following as commit message, to refer to #1913 and to mention the additional methods:

    Fix #1913: Add Thread.{interrupt,isInterrupted,interrupted}.

javalanglib/src/main/scala/java/lang/Thread.scala
sjrd: Or, to follow more logically the order of the methods, *above* `name`.

andreaTP: ok

javalanglib/src/main/scala/java/lang/Thread.scala
andreaTP: ok

andreaTP:
should be ok now.
As always, thanks for the support guys! :-)

sjrd:

PR #1912, opened 2015-09-23
Fix #1911: Enable -XfatalWarnings where we can

### LINKED ISSUE #1911: Enable -XfatalWarnings where we can
Most projects do compile without warnings. We should enable `-XfatalWarnings`

gzm0:
I did not change the examples in order to not collide with another upcoming change.

### DIFF
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -163,13 +163,12 @@ object Scalajsld {
 
     new ScalaJSClosureOptimizer().optimizeCP(
         newScalaJSOptimizer(semantics, options.outputMode), cp,
-        Config(
-            output = output,
-            wantSourceMap = options.sourceMap,
-            relativizeSourceMapBase = options.relativizeSourceMap,
-            bypassLinkingErrors = options.bypassLinkingErrors,
-            checkIR = options.checkIR,
-            prettyPrint = options.prettyPrint),
+        Config(output)
+          .withWantSourceMap(options.sourceMap)
+          .withRelativizeSourceMapBase(options.relativizeSourceMap)
+          .withBypassLinkingErrors(options.bypassLinkingErrors)
+          .withCheckIR(options.checkIR)
+          .withPrettyPrint(options.prettyPrint),
         newLogger(options))
   }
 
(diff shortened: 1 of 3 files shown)


gzm0:
Review by @sjrd

nicolasstucki:
LGTM

cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
```diff
@@ -163,13 +163,12 @@ object Scalajsld {
...
         newScalaJSOptimizer(semantics, options.outputMode), cp,
-        Config(
-            output = output,
-            wantSourceMap = options.sourceMap,
-            relativizeSourceMapBase = options.relativizeSourceMap,
-            bypassLinkingErrors = options.bypassLinkingErrors,
-            checkIR = options.checkIR,
-            prettyPrint = options.prettyPrint),
+        Config(output)
+            .withWantSourceMap(options.sourceMap)
```
sjrd: This should be 2 spaces, not 4. Same for the rest of the commit of course.

project/Build.scala
```diff
@@ -321,7 +325,7 @@ object Build extends sbt.Build {
   )
 
   val commonToolsSettings = (
-      commonSettings ++ publishSettings
+      commonSettings ++ publishSettings ++ fatalWarningsSettings
```
sjrd: I believe the tools have warnings (pattern matching not being able to prove reachability or some other nonsense)

gzm0:
Updated

sjrd:
LGTM

project/Build.scala
```diff
@@ -1036,7 +1062,7 @@ object Build extends sbt.Build {
   lazy val partest: Project = Project(
       id = "partest",
       base = file("partest"),
-      settings = commonSettings ++ Seq(
+      settings = commonSettings ++ fatalWarningsSettings ++ Seq(
```
sjrd: This is the cause of the build failure. partest sources in the tools, so it builds with warnings, unfortunately.

gzm0: I am considering to turn of analysis for the ir and the tools. IIUC it does not work anyways: We have default cases everywhere (so exhaustiveness is useless) and we fail reachability everywhere. WDYT?

sjrd: If you manage to turn it off, OK. But I think I tried to, and sometimes it doesn't really turn it off even if you ask for it.

gzm0: I can't even reproduce the warning locally... The tools compile just fine

gzm0:
Updated so upstream stuff can continue. I'll investigate the patmat issues separately.

sjrd:
Is this supposed to be good to merge?

gzm0:
Yes, if it LGTY, you can merge.

sjrd:
LGTM

PR #1916, opened 2015-09-24
Avoid fatal warnings during scaladoc generation.

oleastre:
Seems #1912 prevents scaladoc compilation which is quite annoying if you try to do a publishLocal on some projects (at least javalibEx and testInterface).

As a quick fix, this pull request just disables warnings for the (Compile,doc) scope, but keeps the normal value for all other Compilation scopes.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -239,12 +239,12 @@
 
   <task id="sbtplugin-test"><![CDATA[
     # Publish Scala.js artifacts locally
+    # Then go into standalone project and test
     sbt ++2.11.7 compiler/publishLocal library/publishLocal javalibEx/publishLocal \
                  testInterface/publishLocal jasmineTestFramework/publishLocal stubs/publishLocal \
         ++2.10.5 ir/publishLocal tools/publishLocal jsEnvs/publishLocal \
-                 testAdapter/publishLocal sbtPlugin/publishLocal
-    # Go into standalone project and test
-    cd sbt-plugin-test
+                 testAdapter/publishLocal sbtPlugin/publishLocal &&
+    cd sbt-plugin-test &&
     sbt noDOM/run withDOM/run test \
         'set scalaJSStage in Global := FastOptStage' \
         jetty9/run test \
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -15,7 +15,6 @@
 
   <task id="main"><![CDATA[
     setJavaVersion $java
-    sbtretry ++$scala package packageDoc &&
     sbtretry ++$scala helloworld/run \
         'set scalaJSStage in Global := FastOptStage' \
         ++$scala helloworld/run \
@@ -183,6 +182,8 @@
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala javalibExTestSuite/test &&
     sbtretry ++$scala testSuite/test:doc compiler/test reversi/fastOptJS reversi/fullOptJS &&
+    sbtretry ++$scala compiler/compile:doc library/compile:doc javalibEx/compile:doc \
+        testInterface/compile:doc jasmineTestFramework/compile:doc &&
     sbtretry ++$scala partest/fetchScalaSource &&
     sbtretry ++$scala library/mimaReportBinaryIssues testInterface/mimaReportBinaryIssues &&
     sh ci/checksizes.sh $scala &&
@@ -204,7 +205,9 @@
         stubs/package jsEnvs/test \
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
         jsEnvs/mimaReportBinaryIssues testAdapter/mimaReportBinaryIssues \
-        stubs/mimaReportBinaryIssues cli/mimaReportBinaryIssues
+        stubs/mimaReportBinaryIssues cli/mimaReportBinaryIssues &&
+    sbt ++$scala ir/compile:doc tools/compile:doc jsEnvs/compile:doc \
+        testAdapter/compile:doc stubs/compile:doc
   ]]></task>
 
   <task id="tools-cli-stubs-sbtplugin"><![CDATA[
@@ -215,7 +218,7 @@
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
         jsEnvs/mimaReportBinaryIssues testAdapter/mimaReportBinaryIssues \
         stubs/mimaReportBinaryIssues cli/mimaReportBinaryIssues \
-        sbtPlugin/mimaReportBinaryIssues
+        sbtPlugin/mimaReportBinaryIssues &&
     sbt ++$scala library/scalastyle javalanglib/scalastyle javalib/scalastyle \
         javalibEx/scalastyle ir/scalastyle compiler/scalastyle \
         compiler/test:scalastyle tools/scalastyle tools/test:scalastyle \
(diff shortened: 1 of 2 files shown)


sjrd:
Hum, I thought we did generate ScalaDocs in our CI, precisely to avoid that kind of stuff to happen. If that's not the case, the `ci/matrix.xml` should also be updated to generate all the necessary ScalaDocs.

gzm0:
Ouch... that is bad... IMO we shouldn't do `-nowarn` though, but not set `-XfatalWarnings` where appropriate. Some of the doc warnings are very useful (mainly doc links that are not reachable)...

So as a quick fix, you can do:

```scala
scalacOptions in (Compile, doc) ~= _.filterNot(_ != "-XfatalWarnings")
```

(also, I thought our CI does `publishLocal` quite extensively...)

gzm0:
On of our first lines in CI says: `sbtretry ++$scala package packageDoc`

Unfortunately, we don't aggregate any more since quite time... Also, the sbtplugin-test is only done on the latest versions.

oleastre:
My bad, I did not think about filtering the sequence to remove the fatal warning options...
Do you want me to update this PR to make it at least a temporary fix ?

gzm0:
Yes, please update this PR so we get a temporary fix at least. If you want to add doc tasks to `ci/matrix.xml`, that'd be even more awesome of course (its basically shell like sbt invocations, not too hard to figure out).

gzm0:
If you could also file an issue and reference it, so we keep a record that this should only be temporary.

ci/matrix.xml
```diff
@@ -232,6 +232,18 @@
...
+    sbt ++$scala compiler/compile:doc library/compile:doc javalibEx/compile:doc \
+                 testInterface/compile:doc jasmineTestFramework/compile:doc stubs/compile:doc
+  ]]></task>
+
+  <task id="sbtplugin-doc-generation"><![CDATA[
+    setJavaVersion $java
+    sbt ++2.10.5 ir/compile:doc tools/compile:doc jsEnvs/compile:doc \
+                 testAdapter/compile:doc sbtPlugin/compile:doc
+  ]]></task>
+
```
gzm0: Ok. You got the idea :) Now this will fork off a totally new Jenkins task, which is probably overkill...

You can add doc generation for `ir`, `tools`, `jsEnvs`, `testAdapter` and `sbtPlugin` to the `tools-cli-stubs` / `tools-cli-stubs-sbtplugin` tasks and doc generation for `complier`, `library`, `javalibEx`, `testInterface`, `jasmineTestFramework` and `stubs` to `main` (you can also remove the line with `package` and `packageDocs` there, it is pointless).

ci/matrix.xml
```diff
@@ -320,6 +332,14 @@
...
     </run>
 
+    <!-- Check scaladoc generation -->
+    <run task="doc-generation">
+      <v n="scala">2.11.7</v>
+      <v n="java">1.8</v>
+    </run>
+
+    <run task="sbtplugin-doc-generation" />
+
```
gzm0: These two will then also become unnecessary

project/Build.scala
```diff
@@ -149,7 +149,8 @@ object Build extends sbt.Build {
   )
 
   val fatalWarningsSettings = Seq(
-      scalacOptions += "-Xfatal-warnings"
+      scalacOptions += "-Xfatal-warnings",
+      scalacOptions in (Compile, doc) ~= ( _.filterNot( _ == "-Xfatal-warnings" ) )
```
gzm0: No spaces around parens :)

gzm0:
Thanks a lot. After you have made your changes, please squash all commits in a single one. Make sure to include the following two in the commit message body:

* 311c34fbdac7ca1e7b7c08a675ad4c98cff3ebe7 (the commit that broke stuff)
* #1917 (the issue to make this entirely right)

Thanks!

oleastre:
Code changes and squash is done.
I hope I got the ci/matrix.xml as you expected it. 
(if you want the commits before squash, have a look at https://github.com/oleastre/scala-js/tree/scaladoc-fatal-warnings-commits)

ci/matrix.xml
```diff
@@ -205,6 +206,8 @@
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
         jsEnvs/mimaReportBinaryIssues testAdapter/mimaReportBinaryIssues \
         stubs/mimaReportBinaryIssues cli/mimaReportBinaryIssues
+    sbt ++$scala ir/compile:doc tools/compile:doc jsEnvs/compile:doc \
+        testAdapter/compile:doc
```
gzm0: Arrgh... I forgot to mention `stubs`... I didn't mean to have you remove it. Sorry, my bad

ci/matrix.xml
```diff
@@ -183,6 +182,8 @@
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala javalibExTestSuite/test &&
     sbtretry ++$scala testSuite/test:doc compiler/test reversi/fastOptJS reversi/fullOptJS &&
+    sbtretry ++$scala compiler/compile:doc library/compile:doc javalibEx/compile:doc \
+        testInterface/compile:doc jasmineTestFramework/compile:doc stubs/compile:doc &&
```
gzm0: I see... stubs ended up in here... I merely copied what you wrote.

Stubs belongs in the other task. The rationale is that stuff in the other task does not depend on the Scala.js compiler itself. Therefore we can compile them independently (and not compile the Scala.js compiler at all).

gzm0:
That's all. Sorry I missed the stubs thingy

oleastre:
I have to admit, I did not try to understand the whole test picture...

gzm0:
Jenkins, ok to test

gzm0:
LGTM

ci/matrix.xml
```diff
@@ -205,6 +206,8 @@
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
         jsEnvs/mimaReportBinaryIssues testAdapter/mimaReportBinaryIssues \
         stubs/mimaReportBinaryIssues cli/mimaReportBinaryIssues
+    sbt ++$scala ir/compile:doc tools/compile:doc jsEnvs/compile:doc \
```
sjrd: I think the line above this one needs a `&&` at the end.

ci/matrix.xml
```diff
@@ -225,6 +228,8 @@
         javalibExTestSuite/test:scalastyle helloworld/scalastyle \
         reversi/scalastyle testingExample/scalastyle \
         testingExample/test:scalastyle
+    sbt ++$scala ir/compile:doc tools/compile:doc jsEnvs/compile:doc \
```
sjrd: Same here.

oleastre: I'm ready to add those if needed, but before patch, the tools-cli-stubs-sbtplugin task already contained several not chained sbt commands (not separated by &&); I just copied the existing behavior. 
If needed I can also add "&&" between the mima and scalastyle tests.

sjrd: Oh. That was probably a mistake. Yes, please add `&&` between consecutive invocations of sbt.

oleastre:
Modifications done.
Note that the sbtplugin-test task does not chain the sbt publishLocal calls with the following sbt-plugin-test one.
As I understand it, the ci executes those calls sequentially, returning the latest one error code. 
As sbt publishLocal and test are not chained, the plugins tests are run against the latest successful local install (and not the current code). 
That's probably why you did not noticed the scaladoc generation error with the current ci matrix.

I added fee822797d4ebc337f65e3fc6b7680d29a33e6b5 to correct this.

sjrd:
LGTM Thanks :)

PR #1920, opened 2015-09-28
Fix #1866: Improve performance of the emitter.

### LINKED ISSUE #1866: Investigate the performance of Emitter (write output)
Following up to #1860, `Emitter (write output)` seems to be an area where we spend too much time. There might be some performance bottleneck in there that we can fix. Maybe `javascript.Printers` is to blame.

nicolasstucki:
* Implement java.lang.StringBuilder.ensureCapacity.
* Improve performance of javascript.Printers.
* Improve performance of SourceMapWriter.
* Improve preformance of ir.Utils.escapeJS.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
@@ -22,15 +22,17 @@ import Utils.escapeJS
 object Printers {
 
   /** Basically copied from scala.reflect.internal.Printers */
-  trait IndentationManager {
+  abstract class IndentationManager {
     protected val out: Writer
 
-    protected var indentMargin = 0
-    protected val indentStep = 2
-    protected var indentString = "                                        " // 40
+    private var indentMargin = 0
+    private val indentStep = 2
+    private var indentString = "                                        " // 40
 
     protected def indent(): Unit = indentMargin += indentStep
     protected def undent(): Unit = indentMargin -= indentStep
+
+    protected def getIndentMargin(): Int = indentMargin
 
     protected def println(): Unit = {
       out.write('\n')
@@ -39,51 +41,6 @@ object Printers {
       if (indentMargin > 0)
         out.write(indentString, 0, indentMargin)
     }
-
-    @tailrec
-    protected final def printSeq[A](ls: List[A])(printelem: A => Unit)(
-        printsep: A => Unit): Unit = {
-      ls match {
-        case Nil =>
-        case x :: Nil =>
-          printelem(x)
-        case x :: rest =>
-          printelem(x)
-          printsep(x)
-          printSeq(rest)(printelem)(printsep)
-      }
-    }
-
-    protected def printColumn(ts: List[Any], start: String, sep: String,
-        end: String): Unit = {
-      print(start); indent(); println()
-      printSeq(ts) { x =>
-        print(x)
-      } { _ =>
-        print(sep)
-        println()
-      }
-      undent(); println(); print(end)
-    }
-
-    protected def printRow(ts: List[Any], start: String, sep: String,
-        end: String): Unit = {
-      print(start)
-      printSeq(ts) { x =>
-        print(x)
-      } { _ =>
-        print(sep)
-      }
-      print(end)
-    }
-
-    protected def printRow(ts: List[Any], sep: String): Unit =
-      printRow(ts, "", sep, "")
-
-    protected def print(args: Any*): Unit =
-      args.foreach(printOne)
-
-    protected def printOne(arg: Any): Unit
   }
 
   class IRTreePrinter(protected val out: Writer) extends IndentationManager {
@@ -100,20 +57,55 @@ object Printers {
       }
     }
 
-    protected def printBlock(tree: Tree): Unit = {
-      val trees = tree match {
-        case Block(trees) => trees
-        case _            => List(tree)
+    protected final def printColumn(ts: List[Tree], start: String, sep: String,
+        end: String): Unit = {
+      printString(start); indent(); println()
+      var rest = ts
+      while (rest.nonEmpty) {
+        printTree(rest.head)
+        rest = rest.tail
+        if (rest.nonEmpty) {
+          printString(sep)
+          println()
+        }
       }
-      printColumn(trees, "{", ";", "}")
+      undent(); println(); printString(end)
+    }
+
+    protected final def printRow(ts: List[Any], start: String, sep: String,
+        end: String): Unit = {
+      printString(start)
+      var rest = ts
+      while (rest.nonEmpty) {
+        printOne(rest.head)
+        rest = rest.tail
+        if (rest.nonEmpty)
+          printString(sep)
+      }
+      printString(end)
+    }
+
+    protected def printBlock(tree: Tree): Unit = {
+      tree match {
+        case Block(trees) =>
+          printColumn(trees, "{", ";", "}")
+
+        case _ =>
+          print('{'); indent(); println()
+          print(tree)
+          undent(); println(); print('}')
+      }
     }
 
     protected def printSig(args: List[ParamDef], resultType: Type): Unit = {
       printRow(args, "(", ", ", ")")
-      if (resultType != NoType)
-        print(": ", resultType, " = ")
-      else
-        print(" ")
+      if (resultType != NoType) {
+        printString(": ")
+        printOne(resultType)
+        printString(" = ")
+      } else {
+        printChar(' ')
+      }
     }
 
     protected def printArgs(args: List[Tree]): Unit = {
@@ -509,11 +501,16 @@ object Printers {
 
         case JSObjectConstr(fields) =>
           print("{"); indent; println()
-          printSeq(fields) {
-            case (name, value) => print(name, ": ", value)
-          } { _ =>
-            print(",")
-            println()
+          var rest = fields
+          while (rest.nonEmpty) {
+            printOne(rest.head._1)
+            printOne(": ")
+            printOne(rest.head._2)
+            rest = rest.tail
+            if (rest.nonEmpty) {
+              printString(",")
+              println()
+            }
           }
           undent; println(); print("}")
 
@@ -628,7 +625,9 @@ object Printers {
 
         case PropertyDef(name, _, _, _) =>
           // TODO
-          print(s"<property: $name>")
+          print("<property: ")
+          print(name)
+          print('>')
 
         case ConstructorExportDef(fullName, args, body) =>
           print("export \"", escapeJS(fullName), "\"")
@@ -666,7 +665,7 @@ object Printers {
           print("[]")
 
       case RecordType(fields) =>
-        print("(")
+        print('(')
         var first = false
         for (RecordType.Field(name, _, tpe, mutable) <- fields) {
           if (first) first = false
@@ -675,7 +674,7 @@ object Printers {
             print("var ")
           print(name, ": ", tpe)
         }
-        print(")")
+        print(')')
     }
 
     protected def printIdent(ident: Ident): Unit =
@@ -692,9 +691,11 @@ object Printers {
         printString(if (arg == null) "null" else arg.toString)
     }
 
-    protected def printString(s: String): Unit = {
+    protected def printString(s: String): Unit =
       out.write(s)
-    }
+
+    protected def printChar(c: Int): Unit =
+      out.write(c)
 
     // Make it public
     override def println(): Unit = super.println()
@@ -729,7 +730,8 @@ object Printers {
 
     def printMethodInfo(methodInfo: MethodInfo): Unit = {
       import methodInfo._
-      print(escapeJS(encodedName), ":")
+      print(escapeJS(encodedName))
+      print(":")
       indent(); println()
 
       if (isStatic)
@@ -741,60 +743,90 @@ object Printers {
       if (methodsCalled.nonEmpty) {
         print("methodsCalled:")
         indent(); println()
-        printSeq(methodsCalled.toList) { case (cls, callees) =>
-          print(escapeJS(cls), ": ")
-          print(callees.map(escapeJS).mkString("[", ", ", "]"))
-        } { _ => println() }
+        val iter = methodsCalled.iterator
+        while (iter.hasNext) {
+          val (cls, callers) = iter.next()
+          print(escapeJS(cls))
+          print(": [")
+          printRow(callers, ": [", ", ", "}")
+          if (iter.hasNext)
+            println()
+        }
         undent(); println()
       }
       if (methodsCalledStatically.nonEmpty) {
         print("methodsCalledStatically:")
         indent(); println()
-        printSeq(methodsCalledStatically.toList) { case (cls, callees) =>
-          print(escapeJS(cls), ": ")
-          print(callees.map(escapeJS).mkString("[", ", ", "]"))
-        } { _ => println() }
+        val iter = methodsCalledStatically.iterator
+        while (iter.hasNext) {
+          val (cls, callers) = iter.next
+          print(escapeJS(cls))
+          printRow(callers, ": [", ", ", "}")
+          if (iter.hasNext)
+            println()
+        }
         undent(); println()
       }
       if (staticMethodsCalled.nonEmpty) {
         print("staticMethodsCalled:")
         indent(); println()
-        printSeq(staticMethodsCalled.toList) { case (cls, callees) =>
-          print(escapeJS(cls), ": ")
-          print(callees.map(escapeJS).mkString("[", ", ", "]"))
-        } { _ => println() }
+        val iter = methodsCalledStatically.iterator
+        while (iter.hasNext) {
+          val (cls, callers) = iter.next()
+          print(escapeJS(cls))
+          printRow(callers, ": [", ", ", "}")
+          if (iter.hasNext)
+            println()
+        }
         undent(); println()
       }
-      if (instantiatedClasses.nonEmpty) {
-        println("instantiatedClasses: ",
-            instantiatedClasses.map(escapeJS).mkString("[", ", ", "]"))
-      }
-      if (accessedModules.nonEmpty) {
-        println("accessedModules: ",
-            accessedModules.map(escapeJS).mkString("[", ", ", "]"))
-      }
-      if (usedInstanceTests.nonEmpty) {
-        println("usedInstanceTests: ",
-            usedInstanceTests.map(escapeJS).mkString("[", ", ", "]"))
-      }
-      if (accessedClassData.nonEmpty) {
-        println("accessedClassData: ",
-            accessedClassData.map(escapeJS).mkString("[", ", ", "]"))
-      }
+      if (instantiatedClasses.nonEmpty)
+        printRow(instantiatedClasses, "instantiatedClasses: [", ", ", "]")
+      if (accessedModules.nonEmpty)
+        printRow(accessedModules, "accessedModules: [", ", ", "]")
+      if (usedInstanceTests.nonEmpty)
+        printRow(usedInstanceTests, "usedInstanceTests: [", ", ", "]")
+      if (accessedClassData.nonEmpty)
+        printRow(accessedClassData, "accessedClassData: [", ", ", "]")
 
       undent(); println()
     }
 
-    private def println(arg1: Any, args: Any*): Unit = {
-      print((arg1 +: args): _*)
+    protected def printRow(ts: List[String], start: String, sep: String,
+        end: String): Unit = {
+      print(start)
+      var rest = ts
+      while (rest.nonEmpty) {
+        print(rest.head)
+        rest = rest.tail
+        if (rest.nonEmpty)
+          print(sep)
+      }
+      print(end)
+    }
+
+    private final def println(arg1: Any, args: Any*): Unit = {
+      printOne(arg1)
+      var i = 0
+      val len = args.length
+      while (i < len) {
+        printOne(args(i))
+        i += 1
+      }
       println()
     }
 
     protected def printOne(arg: Any): Unit = arg match {
       case classInfo: ClassInfo   => printClassInfo(classInfo)
       case methodInfo: MethodInfo => printMethodInfo(methodInfo)
-      case arg                    => out.write(arg.toString())
+      case arg                    => print(arg.toString())
     }
+
+    protected def print(s: String): Unit =
+      out.write(s)
+
+    protected def print(c: Int): Unit =
+      out.write(c)
 
     def complete(): Unit = ()
   }
--- a/ir/src/main/scala/org/scalajs/core/ir/Utils.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Utils.scala
@@ -11,6 +11,8 @@ package org.scalajs.core.ir
 
 import java.net.URI
 
+import scala.annotation.switch
+
 object Utils {
 
   /** Relativize target URI w.r.t. base URI */
@@ -47,25 +49,50 @@ object Utils {
     else new URI("file", "", uri.getPath(), uri.getQuery(), uri.getFragment())
 
   def escapeJS(str: String): String = {
+    // scalastyle:off return
     /* Note that Java and JavaScript happen to use the same encoding for
      * Unicode, namely UTF-16, which means that 1 char from Java always equals
      * 1 char in JavaScript. */
-    val builder = new StringBuilder
-    str foreach {
-      case '\\'     => builder.append("\\\\")
-      case '"'      => builder.append("\\\"")
-      case '\u0007' => builder.append("\\a")
-      case '\u0008' => builder.append("\\b")
-      case '\u0009' => builder.append("\\t")
-      case '\u000A' => builder.append("\\n")
-      case '\u000B' => builder.append("\\v")
-      case '\u000C' => builder.append("\\f")
-      case '\u000D' => builder.append("\\r")
-      case c =>
-        if (c >= 32 && c <= 126) builder.append(c.toChar) // ASCII printable characters
-        else builder.append(f"\\u$c%04x")
+    val end = str.length
+    var i = 0
+    while (i != end) {
+      val c = str.charAt(i)
+      if (c >= 32 && c <= 126 && c != '\\' && c != '"')
+        i += 1
+      else
+        return createEscapeJSString(str)
     }
-    builder.result()
+    str
+    // scalastyle:on return
+  }
+
+  private def createEscapeJSString(str: String): String = {
+    val end = str.length()
+    val builder = new java.lang.StringBuilder
+    builder.ensureCapacity(end * 2)
+    var i = 0
+    while (i != end) {
+      val c = str.charAt(i)
+      if (c >= 32 && c <= 126 && c != '\\' && c != '"') {
+        builder.append(c) // ASCII printable characters
+      } else {
+        def encode(c: Char): String = f"\\u$c%04x"
+        builder.append((c: @switch) match {
+          case '\\'     => "\\\\"
+          case '"'      => "\\\""
+          case '\u0007' => "\\a"
+          case '\u0008' => "\\b"
+          case '\u0009' => "\\t"
+          case '\u000A' => "\\n"
+          case '\u000B' => "\\v"
+          case '\u000C' => "\\f"
+          case '\u000D' => "\\r"
+          case c        => encode(c)
+        })
+      }
+      i += 1
+    }
+    builder.toString()
   }
 
   /** A ByteArrayOutput stream that allows to jump back to a given
--- a/javalanglib/src/main/scala/java/lang/StringBuilder.scala
+++ b/javalanglib/src/main/scala/java/lang/StringBuilder.scala
@@ -100,6 +100,10 @@ class StringBuilder(private var content: String) extends CharSequence
     this
   }
 
+  def ensureCapacity(minimumCapacity: Int): Unit = {
+    // Do nothing
+  }
+
   /**
    * @param start The beginning index, inclusive.
    * @param end The ending index, exclusive.
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
@@ -14,7 +14,6 @@ import scala.annotation.switch
 import scala.util.control.Breaks
 
 import java.io.Writer
-import java.net.URI
 
 import org.scalajs.core.ir
 import ir.Position
@@ -26,20 +25,29 @@ import Trees._
 
 import org.scalajs.core.tools.sourcemap.SourceMapWriter
 
+/* Printers code was hand-optimized with raw performance in mind.
+ * It was hand optimized using fastOptJS on examples/reversi and testsSuite/test
+ * as performance benchmarks. JProfiler was used to measure the performance of
+ * hotspots in this object.
+ */
 object Printers {
 
   class JSTreePrinter(protected val out: Writer) extends IndentationManager {
+
     def printTopLevelTree(tree: Tree) {
       tree match {
         case Skip() =>
           // do not print anything
-        case Block(stats) =>
-          for (stat <- stats)
-            printTopLevelTree(stat)
+        case tree: Block =>
+          var rest = tree.stats
+          while (rest.nonEmpty) {
+            printTopLevelTree(rest.head)
+            rest = rest.tail
+          }
         case _ =>
           printStat(tree)
           if (shouldPrintSepAfterTree(tree))
-            print(";")
+            print(';')
           println()
       }
     }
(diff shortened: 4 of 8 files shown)


sjrd:
Could these be 3 separate commits?

nicolasstucki:
Here are the profiled performances of methods that where affected by this commit.

[Hot Spots_testSuite_this_commit.pdf](https://github.com/scala-js/scala-js/files/3716/Hot.Spots_testSuite_this_commit.pdf)
[Hot Spots_testSuite_master.pdf](https://github.com/scala-js/scala-js/files/3717/Hot.Spots_testSuite_master.pdf)

nicolasstucki:
I can split this into 3 commits.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -25,13 +25,15 @@ object Printers {
   trait IndentationManager {
```
sjrd: Turning this into an `abstract class` might also help.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -101,19 +104,23 @@ object Printers {
...
-      if (resultType != NoType)
-        print(": ", resultType, " = ")
-      else
-        print(" ")
+      if (resultType != NoType) {
+        printString(": ")
+        printOne(resultType)
+        printString(" = ")
+      } else
+        printString(" ")
```
sjrd: `{}`

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,22 +52,46 @@ object Utils {
...
+              case '\\' => "\\\\"
+              case '"' => "\\\""
+              case '\u0007' => "\\a"
+              case '\u0008' => "\\b"
+              case '\u0009' => "\\t"
+              case '\u000A' => "\\n"
+              case '\u000B' => "\\v"
+              case '\u000C' => "\\f"
+              case '\u000D' => "\\r"
+              case c => f"\\u$c%04x"
```
sjrd: align the arrows

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,22 +52,46 @@ object Utils {
...
-      case '\u000C' => builder.append("\\f")
-      case '\u000D' => builder.append("\\r")
-      case c =>
-        if (c >= 32 && c <= 126) builder.append(c.toChar) // ASCII printable characters
-        else builder.append(f"\\u$c%04x")
+    val end = str.length()
+    var i = 0
+    while (i < end) {
+      val c = str.charAt(i)
+      if (c < 32 || c > 126)
```
sjrd: Also need to escape `\` and `"`.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,22 +52,46 @@ object Utils {
...
+        i = end + 1
+      i += 1
+    }
```
sjrd: Consider the following version with an early `return`, which is probably more readable than the "trick" to get out of the loop:
```scala
def escapeJS(str: String): String = {
  val end = str.length
  var i = 0
  while (i != end) {
    val c = str.charAt(i)
    if (c == '\\' || c == '"' || c < 32 || c > 126)
      return createEscapeJSString(str)
    i += 1
  }
  str
}
```
and put `createEscapeJSString` as a separate private method.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,22 +52,46 @@ object Utils {
...
+    if (i == end) {
+      str
+    } else {
+      def createEscapeJSString(str: String): String = {
+        val end = str.length()
+        val builder = new StringBuilder
+        builder.ensureCapacity(end)
```
sjrd: `end * 2`?

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,22 +52,46 @@ object Utils {
...
+        var i = 0
+        while (i < end) {
+          val c = str.charAt(i)
+          if (c >= 32 && c <= 126) {
+            builder.append(c) // ASCII printable characters
```
sjrd: same problem with `\` and `"`

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
sjrd: Maybe extract that last `f` format in a tiny `def`, since it's a very rare case (much more so than `\`, `"` and `\n`, for example), and it consumes quite a lot of bytecode despite its source size.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -29,17 +28,21 @@ import org.scalajs.core.tools.sourcemap.SourceMapWriter
...
+
     def printTopLevelTree(tree: Tree) {
       tree match {
-        case Skip() =>
-          // do not print anything
-        case Block(stats) =>
-          for (stat <- stats)
-            printTopLevelTree(stat)
+        case _: Skip =>
+        // do not print anything
```
sjrd: indentation issue

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -40,50 +42,51 @@ object Printers {
...
+        printOne(args(i))
+        i += 1
+      }
+    }
 
     protected def printOne(arg: Any): Unit
+
+    protected def printString(s: String): Unit
+
+    protected def printChar(c: Int): Unit
```
sjrd: I would suggest overloads rather than using different names. Overloads are resolved at compile time, and they do not need "human" specialization at call site.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -40,50 +42,51 @@ object Printers {
...
-    protected def print(args: Any*): Unit =
-      args.foreach(printOne)
+    protected final def print(args: Any*): Unit = {
+      var i = 0
+      val len = args.length
+      while (i < len) {
+        printOne(args(i))
+        i += 1
+      }
+    }
```
sjrd: Does it not help to have overloads with fixed numbers of arguments?

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -80,180 +89,253 @@ object Printers {
...
         case EmptyTree =>
-          print("<empty>")
+          printString("<empty>")
 
         // Comments
 
-        case DocComment(text) =>
-          val lines = text.split("\n").toList
+        case tree: DocComment =>
+          val lines = tree.text.split("\n").toList
```
sjrd: I very much doubt that this change (and similar ones in this file) improves things whatsoever. For case classes, the scalac compiler bypasses the `unapply` methods in pattern matching, and directly goes to `isInstanceOf` + field access.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -80,180 +89,253 @@ object Printers {
...
 
         // Control flow constructs
 
-        case Skip() =>
-          print("/*<skip>*/")
+        case _: Skip =>
+          printString("/*<skip>*/")
 
-        case tree @ Block(trees) =>
+        case tree: Block =>
```
sjrd: Note that `Block` is an exception to my earlier comment because it is *not* a case class. So here it should improve things to have `tree: Block` indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -263,138 +345,172 @@ object Printers {
...
+            case ^ => "^"
+            case << => "<<"
+            case >> => ">>"
             case >>> => ">>>"
 
-            case <  => "<"
+            case < => "<"
             case <= => "<="
-            case >  => ">"
+            case > => ">"
```
sjrd: destruction of alignment, here ^^

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -263,138 +345,172 @@ object Printers {
...
             case >= => ">="
 
             case && => "&&"
             case || => "||"
 
-            case `in`         => "in"
+            case `in` => "in"
             case `instanceof` => "instanceof"
```
sjrd: here too, actually

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -263,138 +345,172 @@ object Printers {
...
+        case tree: ClassDef =>
+          printString("class")
+          for (className <- tree.className) {
+            printChar(' ')
+            printOne(className)
+          }
+          for (parentClass <- tree.parentClass) {
+            printString(" extends ")
+            printOne(parentClass)
+          }
```
sjrd: I guess `ClassDef` is relatively rare, but ... `.isDefined` and `.get`, while we're at it?

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -263,138 +345,172 @@ object Printers {
...
+          printString(" {"); indent()
+          for (member <- tree.members) {
```
sjrd: And a loop here?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
```diff
@@ -82,11 +82,11 @@ final class Emitter(semantics: Semantics, outputMode: OutputMode) {
               emitStrongMode(orderedClasses, builder)
             case _ =>
               throw new IllegalArgumentException(
-                  "Emitting to Strong Mode requires a JSFileBuilder")
+                "Emitting to Strong Mode requires a JSFileBuilder")
```
sjrd: Indentation issue.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Emitter.scala
```diff
@@ -82,11 +82,11 @@ final class Emitter(semantics: Semantics, outputMode: OutputMode) {
...
           }
         case _ =>
-          for (classInfo <- orderedClasses)
-            emitLinkedClass(classInfo, builder)
+          for (classInfo <- orderedClasses.toArray)
+              emitLinkedClass(classInfo, builder)
```
sjrd: Hum ... why? This would do more harm than good, I believe.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -63,13 +63,20 @@ class SourceMapWriter(
...
+  private var pendingName: String = null
 
   writeHeader()
 
-  private def sourceToIndex(source: SourceFile) =
-    _srcToIndex.getOrElseUpdate(source,
-        (sources += sourceToURI(source)).size-1)
+  private def sourceToIndex(source: SourceFile): Int = {
+    if (_srcToIndex.contains(source)) {
+      return _srcToIndex(source)
```
sjrd: `return` is unnecessary here

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -79,8 +86,16 @@ class SourceMapWriter(
     Utils.fixFileURI(relURI).toASCIIString
   }
 
-  private def nameToIndex(name: String) =
-    _nameToIndex.getOrElseUpdate(name, (names += name).size-1)
+  private def nameToIndex(name: String): Int = {
+    if (_nameToIndex.contains(name)) {
+      return _nameToIndex(name)
```
sjrd: `return` is unnecessary here

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -100,10 +115,16 @@ class SourceMapWriter(
...
+    nodePosStack.push((originalPos, null))
+    startSegment(column, originalPos, null)
+  }
+
   def startNode(column: Int, originalPos: Position,
-      originalName: Option[String] = None): Unit = {
-    nodePosStack.push((originalPos, originalName))
-    startSegment(column, originalPos, originalName)
+      originalName: Option[String]): Unit = {
+    val _originalName = originalName.orNull
```
sjrd: `orNull` takes an implicit evidence identity function and then calls `getOrElse`. Don't use it. Use an `if/else` instead.

sjrd: Also call this val `originalName1` or something instead of `_originalName`.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -49,8 +49,8 @@ class SourceMapWriter(
   private val names = new ListBuffer[String]
   private val _nameToIndex = new HashMap[String, Int]
 
-  private val nodePosStack = new Stack[(Position, Option[String])]
-  nodePosStack.push((NoPosition, None))
+  private val nodePosStack = new Stack[(Position, String)]
```
sjrd: Add a comment that the Strings are nullable.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -63,13 +63,20 @@ class SourceMapWriter(
 
   private var pendingColumnInGenerated: Int = -1
   private var pendingPos: Position = NoPosition
-  private var pendingName: Option[String] = None
+  private var pendingName: String = null
```
sjrd: Add a comment that this is nullable.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -142,13 +164,14 @@ class SourceMapWriter(
     lastColumnInGenerated = pendingColumnInGenerated
 
     // If the position is NoPosition, stop here
-    if (!pendingPos.isDefined)
+    val _pendingPos = pendingPos
```
sjrd: `pendingPos1`

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -194,18 +217,26 @@ class SourceMapWriter(
...
     // Write as many base-64 digits as necessary to encode value
-    do {
-      var digit = value & VLQBaseMask
-      value = value >>> VLQBaseShift
-      if (value != 0)
-        digit |= VLQContinuationBit
-      out.write(Base64Map.charAt(digit))
-    } while (value != 0)
+    if (value < 26) {
+      out.write('A' + value0)
```
sjrd: `value`, not `value0`

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -194,18 +217,26 @@ class SourceMapWriter(
...
+    } else {
+      def writeBase64VLQSlowPath(value0: Int): Unit = {
+        var value = value0
```
sjrd: This could actually be a `@tailrec` function, now :-)

sjrd:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
nicolasstucki: It is unnecessary but the bytecode generated is 2 bytes less. In this case this is not helping, so I'll remove it. It could potentially help with JIT inlining.

sjrd:
```
[info] Fast optimizing /localhome/jenkins/a/workspace/scalajs-task-worker/tools/js/target/scala-2.10/scalajs-tools-test-fastopt.js
[error] Referring to non-existent method jl_StringBuilder.ensureCapacity__I__V
[error]   called from scm_StringBuilder.ensureCapacity__I__V
[error]   called from Lorg_scalajs_core_ir_Utils$.createEscapeJSString__p1__T__T
```
I guess you'll have to implement `ensureCapacity` in `java.lang.String{Builder,Buffer}` for the JS tools to link.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -100,20 +57,53 @@ object Printers {
...
+      printString(end)
+    }
+
     protected def printBlock(tree: Tree): Unit = {
-      val trees = tree match {
-        case Block(trees) => trees
-        case _            => List(tree)
+      tree match {
+        case Block(trees) => printColumn(trees, "{", ";", "}")
+        case _ =>
```
sjrd: I would put the `printColumn` on the next line.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -29,17 +28,21 @@ import org.scalajs.core.tools.sourcemap.SourceMapWriter
 object Printers {
 
   class JSTreePrinter(protected val out: Writer) extends IndentationManager {
+
     def printTopLevelTree(tree: Tree) {
       tree match {
-        case Skip() =>
+        case _: Skip =>
```
sjrd: Left one here

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -49,34 +52,54 @@ object Printers {
       case _                                         => true
     }
 
-    protected def printBlock(tree: Tree): Unit = {
-      val trees = tree match {
-        case Block(trees) => trees
-        case _            => List(tree)
+    protected def printRow(ts: List[Tree], start: String, sep: String,
+                                 end: String): Unit = {
```
sjrd: Indentation issue.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -212,48 +288,67 @@ object Printers {
...
           def containsOnlySelectsFromAtom(tree: Tree): Boolean = tree match {
-            case DotSelect(qual, _)     => containsOnlySelectsFromAtom(qual)
+            case DotSelect(qual, _) => containsOnlySelectsFromAtom(qual)
             case BracketSelect(qual, _) => containsOnlySelectsFromAtom(qual)
-            case VarRef(_)              => true
-            case This()                 => true
-            case _                      => false // in particular, Apply
+            case VarRef(_) => true
+            case This() => true
+            case _ => false // in particular, Apply
```
sjrd: Alignment.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -379,40 +503,42 @@ object Printers {
...
 
-    def printOne(arg: Any): Unit = arg match {
-      case tree: Tree =>
-        printTree(tree, isStat = false)
-      case ident: Ident =>
-        printIdent(ident)
-      case arg =>
-        printString(if (arg == null) "null" else arg.toString)
+    private final def print(propName: PropertyName): Unit = propName match {
+      case lit: StringLiteral => print(lit.asInstanceOf[Tree])
```
sjrd: Use type ascription `lit: Tree` instead of `asInstanceOf`

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -125,22 +162,28 @@ object Printers {
             printRow(trees, "(", ", ", ")")
```
sjrd: (unfold the lines above this line)
This `case tree @ Block(trees)` should be written as `case tree: Block`. `Block` is not a case class, so its `unapply` is "real".

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -49,34 +52,54 @@ object Printers {
...
+            rest = rest.tail
+            printStat(x)
+            if (rest.nonEmpty) {
+              if (shouldPrintSepAfterTree(x))
+                print(';')
+              println()
+            }
+          }
+
+        case _ => printStat(tree)
```
sjrd: For symmetry with the other `case`, I would put `printStat(tree)` on the next line.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -29,17 +28,21 @@ import org.scalajs.core.tools.sourcemap.SourceMapWriter
 object Printers {
```
sjrd: We should drop a comment at top of this file explaining that it is written with raw performance in mind, and hand-optimized, so that our future selves don't try to "clean it up" by factoring out stuff ;-)

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -99,14 +130,20 @@ object Printers {
...
+            print(rhs)
+          }
 
         case Let(ident, mutable, rhs) =>
-          print(if (mutable) "let " else "const ", ident)
-          if (rhs != EmptyTree)
-            print(" = ", rhs)
+          print(if (mutable) "let " else "const ")
+          print(ident)
+          if (rhs != EmptyTree) {
```
sjrd: Just thought that it would probably be a good idea to either use `rhs eq EmptyTree` or reverse the condition as `EmptyTree != rhs`. As it is now (`rhs != EmptyTree`), this requires a completely useless megamorphic dispatch on `rhs`'s `equals` method.

The same applies to all other comparisons to `EmptyTree` in this file.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -212,48 +288,67 @@ object Printers {
...
+          print('(')
+          if (op == `typeof`) {
+            print("typeof ")
+          } else {
+            print((op: @switch) match {
+                          case + => '+'
+                          case - => '-'
+                          case ~ => '~'
+                          case ! => '!'
+                        })
```
sjrd: Indentation issue.

sjrd: Also, consider the following rewriting, which don't need the special (unlikely) `if`:
```scala
(op: @switch) match {
  case +        => print('+')
  case -        => print('-')
  case ~        => print('~')
  case !        => print('!')
  case `typeof` => print("typeof ")
}
```

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -317,21 +425,29 @@ object Printers {
...
 
         case DoubleLiteral(value) =>
           if (value == 0 && 1 / value < 0)
             print("(-0)")
           else if (value >= 0)
-            print(value)
-          else
-            print("(", value, ")")
+            print(value.toString)
+          else {
```
sjrd: The two first branches should receive `{}`.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -126,7 +151,8 @@ class SourceMapWriter(
     pendingPos = originalPos
     pendingName =
       if (startColumn != pendingColumnInGenerated) originalName
-      else pendingName orElse originalName
+      else if (pendingName != null) pendingName
+      else originalName
```
sjrd: Hum ... it seems this code is just plain wrong (I mean the existing code as much as the new code). The initial test
```scala
```
```scala
```

I'm not even sure what's the purpose of these tests anymore :-s Anyway, let's not touch that in this PR, so that we don't hide change of behavior in the middle of a performance-oriented PR. I'll create an issue for this to investigate later.

sjrd: #1922

sjrd:
Having "Fix ..." in the three commits is weird. I would remove them, and just keep if in the title of the PR, which will then make its way to the merge commit message.

sjrd:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -194,18 +221,26 @@ class SourceMapWriter(
    */
   private def writeBase64VLQ(value0: Int): Unit = {
     // Sign is encoded in the least significant bit
-    var value =
+    val value =
       if (value0 < 0) ((-value0) << 1) + 1
       else value0 << 1
```
sjrd: I wonder whether a branchless equivalent of this would improve things. Assuming `writeBase64VLQ` was still visible in the top of profile data.

*If* the JVM optimizes `Math.abs(int)` as branchless, the following is branchless and does the same thing:
```
val value = (Math.abs(value0) << 1) | (value0 >>> 31)
```
because:
* If `value0 < 0`: `Math.abs(value0) == -value0` and `value0 >>> 31 == 1` so the rhs `== ((-value0) << 1) | 1 == ((-value0) << 1) + 1`
* If `value0 >= 0`: `Math.abs(value0) == value0` and `value0 >>> 31 == 0` so the rhs `== (value0 << 1) | 0 == value0 << 1`

If the JVM does not optimize `abs(int)` as branchless, we have to do it ourselves:
```scala
val signExtended = value0 >> 31
val value = (((value0 ^ signExtended) - signExtended) << 1) | (signExtended & 1)
```
Source for the branchless `abs`: https://hbfs.wordpress.com/2008/08/05/branchless-equivalents-of-simple-functions/

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -142,13 +168,14 @@ class SourceMapWriter(
...
     // Extract relevant properties of pendingPos
-    val source = pendingPos.source
-    val line = pendingPos.line
-    val column = pendingPos.column
+    val source = pendingPos1.source
+    val line = pendingPos1.line
+    val column = pendingPos1.column
 
     // Source index field
     if (source == lastSource) { // highly likely
```
sjrd: Maybe try `source eq lastSource` here? Positions coming from different `.sjsir` files are unlikely to have equal `source`s (although it's possible, because of inner classes), and all positions coming from the same `.sjsir` will be `eq`, not just `==`. So the speed improvement due to using `eq` might outweigh the increased "cache misses".

This absolutely needs profiling to decide which one is best.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
nicolasstucki: Got rid of it completely.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,20 +52,42 @@ object Utils {
...
-      case '\u000C' => builder.append("\\f")
-      case '\u000D' => builder.append("\\r")
-      case c =>
-        if (c >= 32 && c <= 126) builder.append(c.toChar) // ASCII printable characters
-        else builder.append(f"\\u$c%04x")
+    builder.ensureCapacity(end * 2)
+    var i = 0
+    while (i != end) {
+      val c = str.charAt(i)
+      if (c >= 32 && c <= 126  && c != '\\' && c != '"') {
```
sjrd: Double space.

javalanglib/src/main/scala/java/lang/StringBuilder.scala
```diff
@@ -100,6 +100,10 @@ class StringBuilder(private var content: String) extends CharSequence
     this
   }
 
+  def ensureCapacity(minimumCapacity: Int): Unit = {
+    // Do nothing
+  }
```
sjrd: Believe it or not, this needs a test :p Basically a test that just calls the method, so that we test that it links.

Also, please add the same method in `StringBuffer` as well, and its test (we keep them in sync).

sjrd:
Fun error here:
```
[info] java.lang.Thread
[info]   + getName and setName
[error] Problem executing code in tests: SyntaxError: Unexpected token 呈
SyntaxError: Unexpected token 呈
	at <jscode>.Object.parse@native(Unknown Source)
	at <jscode>.new SourceMapConsumer(/home/jenkins/node_modules/source-map-support/node_modules/source-map/lib/source-map/source-map-consumer.js:50)
	at <jscode>.mapSourcePosition(/home/jenkins/node_modules/source-map-support/source-map-support.js:120)
	at <jscode>.wrapCallSite(/home/jenkins/node_modules/source-map-support/source-map-support.js:187)
	at <jscode>.{anonymous}()(/home/jenkins/node_modules/source-map-support/source-map-support.js:223)
	at <jscode>.Array.map@native(Unknown Source)
	at <jscode>.Function.prepareStackTrace(/home/jenkins/node_modules/source-map-support/source-map-support.js:222)
	at <jscode>.jasmine.Spec.fail([stdin]:2425)
	at <jscode>.jasmine.Block.execute([stdin]:1168)
	at <jscode>.jasmine.Queue.next_([stdin]:2198)
```

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -447,15 +580,21 @@ object Printers {
...
-    override protected def printString(s: String): Unit = {
+    override protected def print(s: String): Unit = {
       // assume no EOL char in s, and assume s only has ASCII characters
-      super.printString(s)
+      super.print(s)
       column += s.length()
     }
 
+    override protected def print(c: Int): Unit = {
+      // assume c is not EOL, and assume c is an ASCII characters
```
sjrd: type: character without s

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -194,18 +239,25 @@ class SourceMapWriter(
    */
   private def writeBase64VLQ(value0: Int): Unit = {
     // Sign is encoded in the least significant bit
-    var value =
-      if (value0 < 0) ((-value0) << 1) + 1
-      else value0 << 1
+    val signExtended = value0 >> 31
+    val value = (((value0 ^ signExtended) - signExtended) << 1) | (signExtended & 1)
```
sjrd: This really really needs a good comment. I suggest:
```scala
/* The sign is encoded in the least significant bit, while the
 * absolute value is shifted one bit to the left.
 * So in theory the "definition" of `value` is:
 *   val value =
 *     if (value0 < 0) ((-value0) << 1) | 1
 *     else value0 << 1
 * The following code is a branchless version of that spec.
 * It is valid because:
 * - if value0 < 0:
 *   signExtended == value0 >> 31 == 0xffffffff
 *   value0 ^ signExtended == ~value0
 *   (value0 ^ signExtended) - signExtended == ~value0 - (-1) == -value0
 *   signExtended & 1 == 1
 *   So we get ((-value0) << 1) | 1 as required
 * - if n >= 0:
 *   signExtended == value0 >> 31 == 0
 *   value0 ^ signExtended == value0
 *   (value0 ^ signExtended) - signExtended == value0 - 0 == value0
 *   signExtended & 1 == 0
 *   So we get (value0 << 1) | 0 == value0 << 1 as required
 */
```

sjrd:
`java.lang.StringBuffer.ensureCapacity` is still missing in the first commit.

sjrd:
"Implement java.lang.StringBuilder.ensureCapacity."
-> "Implement java.lang.String**{Builder,Buffer}**.ensureCapacity."

sjrd:
Apparently the bin compat of the IR is also broken:
```
[info] Scala.js IR: found 24 potential binary incompatibilities
```
So I guess you can also disable bin compat for the IR in the same commit as when disabling it for the tools (and adapt the commit message).

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -50,20 +52,44 @@ object Utils {
     /* Note that Java and JavaScript happen to use the same encoding for
      * Unicode, namely UTF-16, which means that 1 char from Java always equals
      * 1 char in JavaScript. */
+    val end = str.length
+    var i = 0
+    // scalastyle:off return
```
sjrd: The convention I've been using for `return`s and Scalastyle is to put the `:off` and `:on` directives at the beginning and end of the method, like this:
```scala
  def escapeJS(str: String): String = {
    // scalastyle:off return
    /* Note that ... */
    val end = str.length
    ...
    str
    // scalastyle:on return
  }
```
The advantage is that, when reading the method, the very first thing you see is: "Attention! There be `return`s in this method. Watch out for the control flow!" I find this visual warning interesting :-)

sjrd:
LGTM :-)

PR #1921, opened 2015-09-28
Fix #1894: Add js.ConstructorTag[T].

### LINKED ISSUE #1894: js.constructor - support for dynamic types/classes
> sjrd :  No you have to call js.constructorOf where you know the class statically. It does not work through java.lang.Class (that would be reflection).

sjrd:
Similarly to `ClassTag[T]`, `js.ConstructorTag[T]` is an implicitly materializable wrapper for `js.constructorOf[T]`.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
@@ -90,6 +90,9 @@ trait JSDefinitions { self: JSGlobalAddons =>
     lazy val JSThisFunctionModule = JSThisFunctionClass.companionModule
       def JSThisFunction_fromFunction(arity: Int): TermSymbol = getMemberMethod(JSThisFunctionModule, newTermName("fromFunction"+arity))
 
+    lazy val JSConstructorTagModule = getRequiredModule("scala.scalajs.js.ConstructorTag")
+      lazy val JSConstructorTag_materialize = getMemberMethod(JSConstructorTagModule, newTermName("materialize"))
+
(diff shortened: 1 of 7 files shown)


sjrd:
Review by @nicolasstucki

gzm0:
Why can't this be a macro?

sjrd:
[It can](https://github.com/scala-js/scala-js/compare/master...sjrd:constructor-tag-macro), but the reflect APIs I found for Scala 2.11.x and wanted to use didn't work in 2.10.x. Since I couldn't make the typechecker to "overlook" the implicit if the macro aborts (apparently only whitebox macros can do that), there was no benefit to the macro version. Hence the compiler intrinsic, where I could also reuse the code from `js.constructorOf`.

gzm0:
Hmmm... I see the problem. How does Scalac solve this with `ClassTag`? I assume the typer just knows about them...

Maybe we should mimic the implicit not found message in the materializer intrinsic, so it at least appears to be the same.

gzm0:
You might also want to give the `ConstructorTag` an implicit not found message that mimics the one from `ClassTag`.

sjrd:
In scalac, the typer knows about `ClassTag`.

I think it would be misleading to mimic the implicit not found message. If somehow, someone crafts an example with another implicit in scope that has a lower priority than `materialize`, the mimicked message would confuse them *even more*.

I could add an `@implicitNotFound`, but it would not be used, because there is *always* an implicit that is found (because of `materialize`). So maybe that's misleading as well? Not sure about that.

sjrd:
Updated with a fix/workaround for test failures in ES6 output mode.

sjrd:
retest this please

nicolasstucki:
LGTM

PR #1923, opened 2015-09-30
Fix #1918: Fix Long.MinValue / (negative power of 2).

### LINKED ISSUE #1918: Long division returns incorrect result for Long.MinValue
When `Long.MinValue` is divided by a negative number, the result is wrong. For instance, when dividing by `-2`, the result is `4611686018427387903` instead of `4611686018427387904`. Also dividing by `-1` returns `Long.MaxValue` instead of `Long.MinValue`. The following code demonstrates the bug:

```scala
object TestApp extends JSApp {
  def main(): Unit = {
    val x = -2
    val y = Long.MinValue / x
    // Prints wrong result
    println(y)
    // Prints correct result, probably due to compiler optimization
    println(Long.MinValue / -2)
  }
}
```

sjrd:


sjrd:
Review by @gzm0, assuming you still have any recollection of how `RuntimeLong` works ^^

nicolasstucki:
LGTM

PR #1924, opened 2015-10-01
Implement ThreadFactory.

andreaTP:
...as always... just to make things cross-compile on Akka.js...


sjrd:

PR #1927, opened 2015-10-05
Reduce object allocations in SourceMapWriter.

### LINKED ISSUE #1926: Reduce the number of object allocation during emitter phase.
JProfiler is showing that during the emitter phase the number of list `List.$colon$colon`, `String` and `Tuple` objects. Many of those can be avoided.

nicolasstucki:


nicolasstucki:
Review by @sjrd

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -66,6 +66,35 @@ object Utils {
     // scalastyle:on return
   }
 
+  def printEscapeJS(str: String, out: java.io.Writer): Unit = {
```
sjrd: No chance to factor out the common parts with `createJSString`? :( If not, this needs 2 `!!!` code duplicate comments.

nicolasstucki: Actually there is an easy way using a `StringWriter`. I had it on another branch. I'll add it to this one.

sjrd: Also, how does this performs compared to testing whether any encoding is necessary first, like `escapeJS` does?

ir/src/main/scala/org/scalajs/core/ir/Position.scala
```diff
@@ -19,10 +19,13 @@ final case class Position(
...
-    source.getScheme == null && source.getRawAuthority == null &&
-    source.getRawPath == "" && source.getRawQuery == null &&
-    source.getRawFragment == null
+  def isEmpty: Boolean = isEmptyCached
+
+  private lazy val isEmptyCached: Boolean = {
+    val source1 = source
+    source1.getScheme == null && source1.getRawAuthority == null &&
+      source1.getRawPath == "" && source1.getRawQuery == null &&
+      source1.getRawFragment == null
```
sjrd: That helps? Like, seriously?

If it does, maybe instead try moving `source.getRawPath == ""` as a first test, and the rest of the comparisons in a `slowPath()` helper. Most (all) `Position`s should be non-empty because of a non-empty `getRawPath`, so just that test should be enough to answer `false`, like, all the time.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -32,8 +32,11 @@ object SourceMapWriter {
   private final val VLQBaseMask = VLQBase - 1
   private final val VLQContinuationBit = VLQBase
 
-  private def jsonString(s: String) =
-    "\"" + Utils.escapeJS(s) + "\""
+  private def jsonString(s: String, out: Writer) = {
```
sjrd: Consider renaming it to `printJSONString`.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
...
   // Strings are nullable in this stack
-  private val nodePosStack = new Stack[(Position, String)]
-  nodePosStack.push((NoPosition, null))
+  private val nodePosStack = new NodePosStack
+  nodePosStack.push(NoPosition, null)
+
+  private final class NodePosStack {
+    private var topIndex: Int = -1
+    private var posStack: Array[Position] = new Array(32)
+    private var nameStack: Array[String] = new Array(32)
```
sjrd: 32 doesn't seem very much. I think we can be more generous, here, and reduce the likelihood that copying is ever necessary.

What's the biggest index that is necessary when linking the test suite?

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
   private val _nameToIndex = new HashMap[String, Int]
 
   // Strings are nullable in this stack
-  private val nodePosStack = new Stack[(Position, String)]
-  nodePosStack.push((NoPosition, null))
+  private val nodePosStack = new NodePosStack
+  nodePosStack.push(NoPosition, null)
+
+  private final class NodePosStack {
```
sjrd: This class should be in the companion object of `SourceMapWriter`, I believe.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
...
+    private var topIndex: Int = -1
+    private var posStack: Array[Position] = new Array(32)
+    private var nameStack: Array[String] = new Array(32)
+
+    def pop(): Unit = {
+      val topIdx = topIndex
+      val newTopIdx = topIdx - 1
+      topIndex = newTopIdx
+      posStack(topIdx) = null
+      nameStack(topIdx) = null
```
sjrd: Nulling out stuff is just wasting cycles here. The strings and positions cannot be reclaimed anyway because they're still in the `Tree`. And the stack itself will be freed before the tree is removed.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
...
+
+    def topName: String =
+      nameStack(topIndex)
+
+    def push(pos: Position, originalName: String): Unit = {
+      val newTopIdx = topIndex + 1
+      topIndex = newTopIdx
+      val currentSize = posStack.length
+      if (newTopIdx >= currentSize)
+        changeStackSize(2 * currentSize, newTopIdx)
```
sjrd: `changeStackSize` is a slow path, let it recompute `currentSize` and `newTopIdx` for itself, so you don't have to do that in the fast-path `push`. Also, consider calling it `grow`, which is a pretty common name for a method that reallocates an internal array.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
...
+      posStack(newTopIdx) = pos
+      nameStack(newTopIdx) = originalName
+
```
sjrd: Spurious blank line.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -50,8 +53,48 @@ class SourceMapWriter(
...
+    }
+
+    private def changeStackSize(newSize: Int, copyLength: Int): Unit = {
+      val newPosStack = new Array[Position](newSize)
+      val newNameStack = new Array[String](newSize)
+      System.arraycopy(posStack, 0, newPosStack, 0, copyLength)
+      System.arraycopy(nameStack, 0, newNameStack, 0, copyLength)
+      posStack = newPosStack
+      nameStack = newNameStack
```
sjrd: Consider using `java.util.Arrays.copy(array, newLength)`, which basically does all of that.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -100,10 +143,10 @@ class SourceMapWriter(
   }
 
   private def writeHeader(): Unit = {
-    out.write("{\n")
-    out.write("\"version\": 3,\n")
-    out.write("\"file\": " + jsonString(generatedFile) + ",\n")
-    out.write("\"mappings\": \"")
+    val out1 = out
```
sjrd: This caching in local vals looks a bit extreme, even by the standards of the recent changes. Is there evidence that this actually improves performance? I would expect the JVM to optimize that for us.

sjrd: If it does help, it should however be in a separate commit. It's current owning commit says "reduce number of allocations".

sjrd:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
nicolasstucki: Actually in practice with the `changeStackSize` only gets invoked twice in the `testSuite`. I will initialize it with 128.

sjrd:
OK. However I think the `Fix #1926: ` part of the commit message should be removed, as it is not only this commit/PR that addresses this problem. There are other PRs as well, such as #1936. So I think we'll have to close #1926 manually with a comment referencing the PRs that have participated in it.

sjrd:
LGTM

PR #1930, opened 2015-10-05
Fix #1922: Remove unnecessary `if` branch in SourceMapWriter.startSegment

### LINKED ISSUE #1922: There's something fishy in SourceMapWriter.startSegment().
As commented in #1920, there seems to be something awfully wrong with the name index generation in source maps in `SourceMapWriter.startSegment()`. Repeating the comment here:

About the assignment to `pendingName`: hum ... it seems this code is just plain wrong. The initial test
```scala
```
```scala
```

We need to investigate this and fix it.

nicolasstucki:


nicolasstucki:
This PR might conflict with #1927. In that case a rebase on the second to be merged would be needed.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -150,8 +150,7 @@ class SourceMapWriter(
     pendingColumnInGenerated = startColumn
     pendingPos = originalPos
     pendingName =
-      if (startColumn != pendingColumnInGenerated) originalName
-      else if (pendingName != null) pendingName
+      if (pendingName != null) pendingName
       else originalName
```
sjrd: Actually, it can go further and just do
```scala
pendingName = originalName
```
The previous logic was completely broken. I have no idea how I came to write this stupidity back then. Possible explanation: I wrote that 3 days before Scala Days 2013, where I was going to present Scala.js to the world for the first time, and I wanted source maps to get in ^^

sjrd:
LGTM

PR #1931, opened 2015-10-05
Implement java.util.Objects.

### LINKED ISSUE #1922: There's something fishy in SourceMapWriter.startSegment().

nicolasstucki:


javalib/src/main/scala/java/util/Objects.scala
```diff
@@ -0,0 +1,64 @@
...
+  def equals(a: AnyRef, b: AnyRef): Boolean =
+    if (a == null) b == null
+    else a.equals(b)
+
+  @inline
+  def deepEquals(a: AnyRef, b: AnyRef) = {
+    (a, b) match {
+      case (a: Array[AnyRef], b: Array[AnyRef]) => Arrays.deepEquals(a, b)
+      case _                                    => equals(a, b)
+    }
```
sjrd: What about arrays of primitive types?

javalib/src/main/scala/java/util/Objects.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  @inline
+  def toString(o: AnyRef, nullDefault: String): String =
+    if (o == null) nullDefault
+    else o.toString
+
+  @inline
+  def compare[T](a: T, b: T, c: Comparator[_ >: T]): Int =
+    if (a == b) 0
+    else c.compare(a, b)
```
sjrd: `a eq b`. The spec is unclear about what "identical" means, but experimentation clearly shows that it means `eq`:
```scala
scala> Objects.compare((1, 2), (1, 2), new java.util.Comparator[AnyRef] { def compare(x: AnyRef, y: AnyRef) = 1 })
res3: Int = 1
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/ObjectsTest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    it("should provide `equals`") {
+      val obj = new Object
+      expect(ju.Objects.equals(obj, obj)).toBeTruthy
+      expect(ju.Objects.equals(new Object, new Object)).toBeFalsy
+      expect(ju.Objects.equals(1, 1)).toBeTruthy
+      expect(ju.Objects.equals(1, 2)).toBeFalsy
+      expect(ju.Objects.equals("abc", "abc")).toBeTruthy
+      expect(ju.Objects.equals("abc", "abd")).toBeFalsy
+    }
```
sjrd: The whole point of `equals` is to deal with `null`s, and you don't test `null`s ;-)
Same for most of the other methods.

sjrd:
`Objects` is JDK7+. The (non-JDK8) tests should be in `require-jdk7`

test-suite/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/ObjectsTestOnJDK7.Scala
```diff
@@ -0,0 +1,74 @@
+package org.scalajs.testsuite.javalib.util
```
sjrd: This file name is buggy. It ends in `.Scala` instead of `.scala`.

javalib/src/main/scala/java/util/Objects.scala
```diff
@@ -0,0 +1,78 @@
...
+
+object Objects {
+
+  @inline
+  def equals(a: AnyRef, b: AnyRef): Boolean =
+    if (a == null) b == null
+    else a.equals(b)
+
+  @inline
+  def deepEquals(a: AnyRef, b: AnyRef) = {
```
sjrd: Explicit result type, says CI.

javalib/src/main/scala/java/util/Objects.scala
```diff
@@ -0,0 +1,78 @@
...
+        case (a1: Array[Boolean], a2: Array[Boolean]) => Arrays.equals(a1, a2)
+        case (a1: Array[Float], a2: Array[Float])     => Arrays.equals(a1, a2)
+        case (a1: Array[Double], a2: Array[Double])   => Arrays.equals(a1, a2)
+        case _                                        => a === b
+      }
+    }
+  }
+
+  @inline
+  def hashCode(o: AnyRef) =
```
sjrd: Explicit result type, says CI.

sjrd:
LGTM

PR #1936, opened 2015-10-07
Add Utils.printEscapeJS

nicolasstucki:


nicolasstucki:
Extracted single commit from #1927.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -66,33 +67,39 @@ object Utils {
...
+          }
+      }
+        escapeJSEncoded(c)
       }
       i += 1
     }
-    builder.toString()
+  }
+
+  private def createEscapeJSString(str: String): String = {
```
sjrd: This method should be before `printEscapeJS`, since it is only a helper for `escapeJS`.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -66,33 +67,39 @@ object Utils {
     // scalastyle:on return
   }
 
-  private def createEscapeJSString(str: String): String = {
+  def printEscapeJS(str: String, out: java.io.Writer): Unit = {
```
sjrd: With the refactoring prompted by the introduction of this method, I think the comment in `escapeJS` should be moved in `printEscapeJS` instead.

nicolasstucki: Moved comment

sjrd:
It seems the fatal error happens for this change.

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -67,32 +65,41 @@ object Utils {
...
+            case '\u0009' => out.write('t')
+            case '\u000A' => out.write('n')
+            case '\u000B' => out.write('v')
+            case '\u000C' => out.write('f')
+            case '\u000D' => out.write('r')
+            case '"'      => out.write('\"')
+            case '\\'     => out.write('\\')
+            case c        => out.write(f"u$c%04x")
+          }
+      }
```
sjrd: Indentation issue.

nicolasstucki:
This new solution is as performant as the previous one on the JVM. But hopefully as it does relly on `String`s rather than `Char`s it will also work when compiled to JS.

tools/shared/src/main/scala/org/scalajs/core/tools/sourcemap/SourceMapWriter.scala
```diff
@@ -180,7 +182,10 @@ class SourceMapWriter(
     // New pending
     pendingColumnInGenerated = startColumn
     pendingPos = originalPos
-    pendingName = originalName
+    pendingName =
+      if (startColumn != pendingColumnInGenerated) originalName
+      else if (pendingName != null) pendingName
+      else originalName
```
sjrd: Rebase error ^^

sjrd:
LGTM

PR #1939, opened 2015-10-07
Fix #1943: Add js.Tuple

### LINKED ISSUE #1943: js.Tuple
Support js Tuple for facade types, combines a set of properties 
```javascript
{  
    0: T0;  
    1: T1;  
    ...  
    n: Tn;  
}
```
to
```scala
js.TupleN[T0,T1,...Tn]
```
See also tuple types in typescript spec:
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.3.3

Example: https://github.com/mbostock/d3/wiki/Zoom-Behavior :
```javascript
 zoom.translate([translate])
```

spaced:
Add js Tuple 
for 
```javascript
{  
    0: T0;  
    1: T1;  
    ...  
    n: Tn;  
}
```

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Tuple.scala
+++ b/library/src/main/scala/scala/scalajs/js/Tuple.scala
@@ -1,0 +1,777 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+/**
+ * All doc-comments marked as "MDN" are by Mozilla Contributors,
+ * distributed under the Creative Commons Attribution-ShareAlike license from
+ * https://developer.mozilla.org/en-US/docs/Web/Reference/API
+ */
+package scala.scalajs.js
+
+import scala.scalajs.js.annotation.JSName
+import scala.language.implicitConversions
+
+
+// scalastyle:off line.size.limit
+
+/**
+ *  A tuple "view" of 2 elements of a JavaScript [[Array]].
+ *  Combines
+ *  {{{
+ *  0: T0;
+ *  1: T1;
+ *  }}}
+ *  to
+ *  {{{
+ *  js.Tuple2[T0,T1]
+ *  }}}
+ *
+ *  Supports implicit conversion to [[scala.Tuple2]].
+ *  To use it, cast your array into a [[Tuple2]] using
+ *  {{{
+ *  val array = js.Array[Any](42, "foobar")
+ *  val tuple2 = array.asInstanceOf[js.Tuple2[Int, String]]
+ *  }}}
+ *  or convert a Scala tuple
+ *  {{{
+ *  val obj: js.Tuple2[Int, String] = (42, "foobar")
+ *  }}}
+ */
+@native
+sealed trait Tuple2[+T1, +T2] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+}
+
+object Tuple2 {
+  @inline def apply[T1, T2](_1: T1, _2: T2): Tuple2[T1, T2] =
+    Array(_1, _2).asInstanceOf[Tuple2[T1, T2]]
+
+  @inline def unapply[T1, T2](t: Tuple2[T1, T2]): Option[(T1, T2)] =
+    Some(t)
+
+  @inline implicit def fromScalaTuple2[T1, T2](t: (T1, T2)): Tuple2[T1, T2] =
+    apply(t._1, t._2)
+
+  @inline implicit def toScalaTuple2[T1, T2](t: Tuple2[T1, T2]): (T1, T2) =
+    (t._1, t._2)
+}
+
+/**
+ *  A tuple "view" of 3 elements of a JavaScript [[Array]].
+ *
+ *  @see [[Tuple2]]
+ */
+@native
+sealed trait Tuple3[+T1, +T2, +T3] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+  @JSName("2") val _3: T3 = native
+}
+
+object Tuple3 {
+  @inline def apply[T1, T2, T3](_1: T1, _2: T2, _3: T3): Tuple3[T1, T2, T3] =
+    Array(_1, _2, _3).asInstanceOf[Tuple3[T1, T2, T3]]
+
+  @inline def unapply[T1, T2, T3](t: Tuple3[T1, T2, T3]): Option[(T1, T2, T3)] =
+    Some(t)
+
+  @inline implicit def fromScalaTuple3[T1, T2, T3](t: (T1, T2, T3)): Tuple3[T1, T2, T3] =
+    apply(t._1, t._2, t._3)
+
+  @inline implicit def toScalaTuple3[T1, T2, T3](t: Tuple3[T1, T2, T3]): (T1, T2, T3) =
+    (t._1, t._2, t._3)
+}
+
+/**
+ *  A tuple "view" of 4 elements of a JavaScript [[Array]].
+ *
+ *  @see [[Tuple2]]
+ */
+@native
+sealed trait Tuple4[+T1, +T2, +T3, +T4] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+  @JSName("2") val _3: T3 = native
+  @JSName("3") val _4: T4 = native
+}
+
+object Tuple4 {
+  @inline def apply[T1, T2, T3, T4](_1: T1, _2: T2, _3: T3, _4: T4): Tuple4[T1, T2, T3, T4] =
+    Array(_1, _2, _3, _4).asInstanceOf[Tuple4[T1, T2, T3, T4]]
+
+  @inline def unapply[T1, T2, T3, T4](t: Tuple4[T1, T2, T3, T4]): Option[(T1, T2, T3, T4)] =
+    Some(t)
+
+  @inline implicit def fromScalaTuple4[T1, T2, T3, T4](t: (T1, T2, T3, T4)): Tuple4[T1, T2, T3, T4] =
+    apply(t._1, t._2, t._3, t._4)
+
+  @inline implicit def toScalaTuple4[T1, T2, T3, T4](t: Tuple4[T1, T2, T3, T4]): (T1, T2, T3, T4) =
+    (t._1, t._2, t._3, t._4)
+}
+
+
+/**
+ *  A tuple "view" of 5 elements of a JavaScript [[Array]].
+ *
+ *  @see [[Tuple2]]
+ */
+@native
+sealed trait Tuple5[+T1, +T2, +T3, +T4, +T5] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+  @JSName("2") val _3: T3 = native
+  @JSName("3") val _4: T4 = native
+  @JSName("4") val _5: T5 = native
+}
+
+object Tuple5 {
+  @inline def apply[T1, T2, T3, T4, T5](_1: T1, _2: T2, _3: T3, _4: T4, _5: T5): Tuple5[T1, T2, T3, T4, T5] =
+    Array(_1, _2, _3, _4, _5).asInstanceOf[Tuple5[T1, T2, T3, T4, T5]]
+
+  @inline def unapply[T1, T2, T3, T4, T5](t: Tuple5[T1, T2, T3, T4, T5]): Option[(T1, T2, T3, T4, T5)] =
+    Some(t)
+
+  @inline implicit def fromScalaTuple5[T1, T2, T3, T4, T5](t: (T1, T2, T3, T4, T5)): Tuple5[T1, T2, T3, T4, T5] =
+    apply(t._1, t._2, t._3, t._4, t._5)
+
+  @inline implicit def toScalaTuple5[T1, T2, T3, T4, T5](t: Tuple5[T1, T2, T3, T4, T5]): (T1, T2, T3, T4, T5) =
+    (t._1, t._2, t._3, t._4, t._5)
+}
+
+/**
+ *  A tuple "view" of 6 elements of a JavaScript [[Array]].
+ *
+ *  @see [[Tuple2]]
+ */
+@native
+sealed trait Tuple6[+T1, +T2, +T3, +T4, +T5, +T6] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+  @JSName("2") val _3: T3 = native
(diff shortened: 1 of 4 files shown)


sjrd:
Jenkins, test this please

library/src/main/scala/scala/scalajs/js/Tuple.scala
```diff
@@ -0,0 +1,389 @@
...
+    (t._1, t._2, t._3)
+}
+//same until Tuple22..
+
+
+@native
+sealed trait Tuple2[+T1, +T2] extends Object {
+  @JSName("0") val _1: T1 = native
+  @JSName("1") val _2: T2 = native
+}
```
sjrd: I'd rather you declared the things in the following order:
```scala
trait Tuple2
object Tuple2
trait Tuple3
object Tuple3
```
etc.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.scalajs.jasminetest.JasmineTest
+
+import scala.scalajs.js
+
+object TupleTest extends JasmineTest {
+
+  describe("scala.scalajs.js.Tuple") {
+
+    it("should provide an equivalent of scala tuple") {
+      val obj: js.Tuple2[Int, String] = (42, "foobar")
```
sjrd: This test is actually testing "should convert from Scala tuple".

There should also be a test directly using `apply`. Also, there should be a test defining the tuple as a `js.Array(stuff...).asInstanceOf[js.Tuple2[Int, String]`, as we want to test that the structure of a tuple is equivalent to a JS array.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,42 @@
...
+      val scalaObj: scala.Tuple2[Int, String] = obj
+
+      expect(scalaObj._1).toEqual(42)
+      expect(scalaObj._2).toEqual("foobar")
+    }
+
+    it("should unapply js tuple") {
+      val obj: js.Tuple2[Int, String] = (42, "foobar")
+      val js.Tuple2(t1, t2) = obj
```
sjrd: You should somehow test that `t1` and `t2` can be inferred as `Int` and `String`, e.g., by adding this afterwards:
```scala
val t1IsInt: Int = t1
val t2IsString: String = t2
```

sjrd: Also, this tests `unapply` in the *destructuring* use case. There should also be a test where it is used in pattern matching position.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    it("should provide an equivalent of scala tuple") {
+      val obj: js.Tuple2[Int, String] = (42, "foobar")
+
+      expect(obj._1).toEqual(42)
+      expect(obj._2).toEqual("foobar")
+    }
+
+    it("should convert to scala tuple") {
+      val obj: js.Tuple2[Int, String] = (42, "foobar")
```
sjrd: Consider
```scala
val obj = js.Tuple2(42, "foobar")
```
as a shorter form to directly create a JS tuple without going through the Scala tuple. It also makes this test independent from the conversion from Scala tuple to JS tuple.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
sjrd: Oh, and also one test asserting that a newly created `js.Tuple2(42, "foobar")` is actually a JS array, by testing `(obj: Any).isInstanceOf[js.Array[_]]`.

sjrd:
That's all.

spaced:
Tnx for your comments. Jenkins failed due timeout, my fault?
Updated commit with all tuples (1 - 22). 
Should i commit "code generator" too?
Tests updated (also Jasmin#fail added)

sjrd:
The Jenkins error was spurious. It's not your fault.

What does the generator look like? Could you also adapt it to generate unit tests for all tuples?

library/src/main/scala/scala/scalajs/js/Tuple.scala
```diff
@@ -0,0 +1,632 @@
...
+ * https://developer.mozilla.org/en-US/docs/Web/Reference/API
+ */
+package scala.scalajs.js
+
+import scala.scalajs.js.annotation.JSName
+import scala.language.implicitConversions
+
+// scalastyle:off line.size.limit
+
+
```
sjrd: Double blank line.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,73 @@
...
+      val t2IsString: String = t2
+      expect(t1IsInt).toEqual(42)
+      expect(t2IsString).toEqual("foobar")
+    }
+
+    it("should unapply JS tuple in pattern matching position") {
+      val obj = js.Tuple2(42, "foobar")
+      obj match {
+        case js.Tuple2(2, _) => fail("Not expected match")
+        case js.Tuple2(_ , "foobar") => //ok
```
sjrd: Align the `=>`'s vertically, as per our coding style.

sjrd: Although here too, you should extract `t1` and `t2` and assert that their types are correctly inferred using `t1IsInt` and `t2IsString`.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+
+    it("should be a JS array instance") {
+      val obj = js.Tuple2(42, "foobar")
+
+      expect(obj).isInstanceOf[js.Array[_]]
```
sjrd: This does not test anything. It throws away the result of `isInstanceOf`.
```scala
expect((obj: Any).isInstanceOf[js.Array[_]]).toBeTruthy
```

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,73 @@
...
+      val array= obj.asInstanceOf[js.Array[_]]
```
sjrd: Missing space before `=`

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,73 @@
...
+      expect(array.length).toEqual(2)
+    }
+
+    it("should be as JS array instance") {
```
sjrd: Copy-paste of title.

spaced:
generator is not rocket sience, just println with string substitute:
```
  def main(args: Array[String]) {
    2 until 22 foreach {i =>
      genTrait(i)
      genObject(i)
    }
  }

  def genTrait(i:Int)= {
    println("@native")
    val types= 1 to i map(t => s"+T$t") mkString ", "
    println(s"sealed trait Tuple$i[$types] extends Object {");
    1 to i map( t => s"""  @JSName("${t-1}") val _$t: T$t = native""") foreach println
    println("}")
    println()
  }
```
i could generate some tests too, what is the benefit? or just a "smoke" test for all other tuples?
like construct, convert, convert back, unapply  as one test?

sjrd:
OK maybe it's not necessary to commit the script.

> like construct, convert, convert back, unapply as one test?

Yes, something like that would be good.

spaced:
- Added "smoke" tests for all tuples
- Fixed names of implicit functions
- Added missed Tuple22

library/src/main/scala/scala/scalajs/js/Tuple.scala
```diff
@@ -0,0 +1,671 @@
...
+ */
+package scala.scalajs.js
+
+import scala.scalajs.js.annotation.JSName
+import scala.language.implicitConversions
+
+// scalastyle:off line.size.limit
+
+@native
+sealed trait Tuple1[+T1] extends Object {
```
sjrd: Do we really want `Tuple1`? I've never seen `scala.Tuple1` been used ever. I'm still wondering why it existed in the first place.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,324 @@
...
+      expect(t1IsInt).toEqual(42)
+      expect(t2IsString).toEqual("foobar")
+    }
+
+    it("should unapply JS tuple in pattern matching position") {
+      val obj = js.Tuple2(42, "foobar")
+      obj match {
+        case js.Tuple2(2, _) =>
+          fail("Not expected match")
+        case js.Tuple2(t1 , t2) =>
```
sjrd: Spurious space before `,`

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,324 @@
...
+          expect(t2IsString).toEqual("foobar")
+      }
+    }
+
+    it("should be a JS array instance") {
+      val obj = js.Tuple2(42, "foobar")
+
+      expect((obj: Any).isInstanceOf[js.Array[_]]).toBeTruthy
+      val array = obj.asInstanceOf[js.Array[_]]
+      expect(array.length).toEqual(2)
```
sjrd: I just thought you could strengthen the last line as follows:
```scala
expect(array).toEqual(js.Array(42, "foobar"))
```

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,324 @@
...
+
+      expect(jsObj._1).toEqual(1)
+      expect(scalaObj._1).toEqual(1)
+      expect(t1IsInt).toEqual(1)
+      expect(jsObj.asInstanceOf[js.Array[_]].length).toEqual(1)
+    }
+
+    it("should support tuple of 2") {
+      val jsObj = js.Tuple2("1", 2)
+      val scalaObj: scala.Tuple2[String, Int] = jsObj
```
sjrd: Consider using the usual notation for Scala tuple types: `(String, Int)`.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,324 @@
...
+      val t2IsInt:Int = js.Tuple2.unapply(jsObj).get._2
```
sjrd: Missing space after `:`

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/TupleTest.scala
```diff
@@ -0,0 +1,324 @@
...
+
+      expect(jsObj._2).toEqual(2)
+      expect(scalaObj._2).toEqual(2)
+      expect(t2IsInt).toEqual(2)
+      expect(jsObj.asInstanceOf[js.Array[_]].length).toEqual(2)
```
sjrd: You can similarly strenghten this one:
```scala
expect(jsObj).toEqual(js.Array("1", 2))
```

library/src/main/scala/scala/scalajs/js/Tuple.scala
spaced: Same thought. Found this use case (http://stackoverflow.com/questions/7207527/does-scala-have-syntax-for-0-and-1-tuples)

>One can implement a method that has a parameter of type Product, and in this case Tuple1 is the only generic way to iterate fixed size collections with multiple types without losing the type information.

make sense.

spaced:
Commit updated. Open: Javadoc with a small example

library/src/main/scala/scala/scalajs/js/Tuple.scala
spaced: On the other side. Due the implicit conversion there is  no need for iterable on js tuple so far.

sjrd: Let's keep `js.Tuple1` out for now. If some use case arises, we can reconsider later. It's easier to add stuff than remove stuff.

gzm0:
What is the use case for this? It seems like a pretty minor thing to put in the Scala.js library...

gzm0:
Ok, I realize this is useful for facade types. But we should file an issue that documents use cases for a large API change like this.

sjrd:
Indeed, we need an appropriate issue. It is indeed designed for facade types. In particular, TypeScript has built-in tuple types, which can be translated to these `js.Tuple` types.

spaced:
Issue created & linked to this PR
Javadoc added
Let me know if i can do anything else

library/src/main/scala/scala/scalajs/js/Tuple.scala
```diff
@@ -0,0 +1,697 @@
...
+ * distributed under the Creative Commons Attribution-ShareAlike license from
+ * https://developer.mozilla.org/en-US/docs/Web/Reference/API
+ */
+package scala.scalajs.js
+
+import scala.scalajs.js.annotation.JSName
+import scala.language.implicitConversions
+
+/**
+ *  Tuple "view" of a JavaScript array.
```
sjrd: This doc-comment is unfortunately useless as a ScalaDoc, since it is not attached to any class.

I suggest to attach a detailed ScalaDoc to `js.Tuple2`, and have a very simple "see also" kind of link on the other tuple types.

library/src/main/scala/scala/scalajs/js/Tuple.scala
sjrd: Don't forget to remove `js.Tuple1

spaced:
Tuple1 (and test) removed
Javadoc modified for Tuple2 trait with @see to other tuples (or should i split it to trait and object?)

library/src/main/scala/scala/scalajs/js/Tuple.scala
```diff
@@ -0,0 +1,695 @@
...
+ *  @see [[Tuple13]]
+ *  @see [[Tuple14]]
+ *  @see [[Tuple15]]
+ *  @see [[Tuple16]]
+ *  @see [[Tuple17]]
+ *  @see [[Tuple18]]
+ *  @see [[Tuple19]]
+ *  @see [[Tuple20]]
+ *  @see [[Tuple21]]
+ *  @see [[Tuple22]]
```
sjrd: Hum ... My idea was to put `@see [[Tuple2]]` from the other tuple traits, so the other way around. So that if you come to the `js.Tuple4` trait doc, you're sent to the `js.Tuple2` page which actually contains some doc.

spaced: sorry, my fault. now fixed

sjrd:

PR #1944, opened 2015-10-09
Implement floorDiv and floorMod in java.lang.Math

ummels:
Also fix test for `java.lang.toIntExact`.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Math.scala
+++ b/javalanglib/src/main/scala/java/lang/Math.scala
@@ -305,13 +305,33 @@ object Math {
     if (a >= Integer.MIN_VALUE && a <= Integer.MAX_VALUE) a.toInt
     else throw new ArithmeticException("Integer overflow")
 
+  def floorDiv(a: scala.Int, b: scala.Int): scala.Int = {
+    val quot = a / b
+    if ((a < 0) == (b < 0) || quot * b == a) quot
+    else quot - 1
+  }
+
+  def floorDiv(a: scala.Long, b: scala.Long): scala.Long = {
+    val quot = a / b
+    if ((a < 0) == (b < 0) || quot * b == a) quot
+    else quot - 1
+  }
+
+  def floorMod(a: scala.Int, b: scala.Int): scala.Int = {
+    val rem = a % b
+    if ((a < 0) == (b < 0) || rem == 0) rem
+    else rem + b
+  }
+
+  def floorMod(a: scala.Long, b: scala.Long): scala.Long = {
+    val rem = a % b
+    if ((a < 0) == (b < 0) || rem == 0) rem
+    else rem + b
+  }
+
   // TODO
 
   // def IEEEremainder(f1: scala.Double, f2: scala.Double): Double
-  // def floorDiv(a: scala.Int, b: scala.Int): scala.Int
-  // def floorDiv(a: scala.Long, b: scala.Long): scala.Long
-  // def floorMod(a: scala.Int, b: scala.Int): scala.Int
-  // def floorMod(a: scala.Long, b: scala.Long): scala.Long
   // def ulp(a: scala.Float): scala.Float
   // def copySign(magnitude: scala.Double, sign: scala.Double): scala.Double
   // def copySign(magnitude: scala.Float, sign: scala.Float): scala.Float
--- a/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/lang/MathTestOnJDK8.scala
+++ b/test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/lang/MathTestOnJDK8.scala
@@ -27,6 +27,7 @@ object MathTestOnJDK8 extends JasmineTest with ExpectExceptions {
       expect(Math.addExact(2147483646, 1) == Int.MaxValue).toBeTruthy
       expect(Math.addExact(1, 2147483646) == Int.MaxValue).toBeTruthy
       expect(Math.addExact(-1073741824, -1073741824) == Int.MinValue).toBeTruthy
+
       expectThrows[ArithmeticException](Math.addExact(Int.MinValue, -1))
       expectThrows[ArithmeticException](Math.addExact(-1, Int.MinValue))
       expectThrows[ArithmeticException](Math.addExact(Int.MinValue, Int.MinValue))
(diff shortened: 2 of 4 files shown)


javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -305,13 +305,35 @@ object Math {
     if (a >= Integer.MIN_VALUE && a <= Integer.MAX_VALUE) a.toInt
     else throw new ArithmeticException("Integer overflow")
 
+  def floorDiv(a: scala.Int, b: scala.Int): scala.Int = {
+    if (b == 0)
+      throw new ArithmeticException("/ by 0")
```
sjrd: Don't explicitly throw `ArithmeticException`s here. Let it be as undefined behavior as `a / b` itself.

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -305,13 +305,35 @@ object Math {
...
+  }
+
+  def floorDiv(a: scala.Long, b: scala.Long): scala.Long = {
+    if ((a < 0) == (b < 0) || a % b == 0) a / b
+    else a / b - 1
+  }
+
+  def floorMod(a: scala.Int, b: scala.Int): scala.Int = {
+    if (b == 0)
+      throw new ArithmeticException("/ by 0")
```
sjrd: Same here.

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Okay, wasn't sure about this since the JavaDoc explicitly says it should throw when the divisor is 0?

javalanglib/src/main/scala/java/lang/Math.scala
```diff
@@ -305,13 +305,35 @@ object Math {
...
+  def floorDiv(a: scala.Int, b: scala.Int): scala.Int = {
+    if (b == 0)
+      throw new ArithmeticException("/ by 0")
+    if ((a < 0) == (b < 0) || a % b == 0) a / b
+    else a / b - 1
+  }
+
+  def floorDiv(a: scala.Long, b: scala.Long): scala.Long = {
+    if ((a < 0) == (b < 0) || a % b == 0) a / b
+    else a / b - 1
```
sjrd: Consider this alternative formulation which trades one division for one multiplication. Multiplications are much faster than divisions, especially for longs:
```scala
val r = a / b
if ((a < 0) == (b < 0) || r * b == a) r
else r - 1
```
For symmetry, the `Int` version should do the same thing, even though the division overhead there is negligible.

javalanglib/src/main/scala/java/lang/Math.scala
sjrd: Yes, I know. When the JavaDoc says it throws a kind of exception which is an undefined behavior, it's always a judgment call whether to check and throw or not. Sometimes it's hard to decide, but here it's pretty clear that the method should behave the same way as the "non-floor" variant, i.e., the primitive operator. So it shouldn't explicitly check.

ummels: Fixed.

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Fixed.

javalanglib/src/main/scala/java/lang/Math.scala
ummels: Good, that also brings the implementation more in line with the implementation of `floorMod`.

test-suite/src/test/require-jdk8/org/scalajs/testsuite/javalib/lang/MathTestOnJDK8.scala
```diff
@@ -191,12 +196,90 @@ object MathTestOnJDK8 extends JasmineTest with ExpectExceptions {
...
+      expect(Math.floorDiv(1, Int.MaxValue)).toEqual(0)
+      expect(Math.floorDiv(1, Int.MinValue)).toEqual(-1)
+      expect(Math.floorDiv(-1, Int.MaxValue)).toEqual(-1)
+      expect(Math.floorDiv(-1, Int.MinValue)).toEqual(0)
+      expect(Math.floorDiv(Int.MaxValue, 1) == Int.MaxValue).toBeTruthy
+      expect(Math.floorDiv(Int.MaxValue, -1) == -Int.MaxValue).toBeTruthy
+      expect(Math.floorDiv(Int.MinValue, 1) == Int.MinValue).toBeTruthy
+      expect(Math.floorDiv(Int.MinValue, -1) == Int.MinValue).toBeTruthy
+
+      expect(Math.floorDiv(0L, 1L)).toEqual(0)
```
sjrd: Hum sorry, I'm only seeing this now:
for expectations on `Long`s, you should use the following "format":
```scala
expect(Math.floorDiv(0L, 1L) == 0L).toBeTruthy
```
`toEqual` is insufficiently precise on `Long`s.

ummels: Okay, I thought this is only necessary for "large" Longs.

ummels: Fixed.

sjrd:

ummels:
Test again?

sjrd:
Looks a bit too deterministic, even for a PhantomJS has been closed error. It consistently failed with the same error on all builds on JDK 8. That's not a spurious failure, although it's too late for me to diagnose right now.

ummels:
Hmm, yeah, PhantomJS seems to crash when encountering `MathTestOnJDK8`, so some change there seems to cause the crash.

sjrd:
Oh, might be the same PhantomJS bug that forced us to disable the tests for `%` in `compiler.IntTest`. See https://github.com/scala-js/scala-js/blob/2552b0410180fc2539608ecddfb92c47a72680c0/test-suite/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala#L129 and, as the comment says, #593

Try also excluding the tests for `floorMod` on PhantomJS, using the same `unless` incantation as found in `IntTest`.

ummels:
@sjrd Thanks. Can you test this again?

sjrd:

ummels:
I rebased my commit. Test again? (I guess I cannot tell Jenkins myself)

sjrd:
Jenkins, retest this please

PR #1946, opened 2015-10-10
Clean up and optimize bit operations in java.lang.Integer.

sjrd:
About `clz32`, see http://jsperf.com/polyfills-for-math-clz32
Things to note:
* The new polyfill is strictly better than the old polyfill on all VMs
* On Chrome, the intrinsified builtin is about 25x faster than the new polyfill
* On Firefox, the new polyfill is just as fast as the builtin (probably because they implemented it in JS using a similar algo)
* On Microsoft Edge (reported as Chrome 42), the new polyfill is about twice as fast as the builtin ^^


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #1948, opened 2015-10-12
Fix #1928: Defer calls to getSimpleName until needed in IR checker.

### LINKED ISSUE #1928: Calls to java.lang.Class in the IRChecker are generating too many strings.
After an execution of `fastOptJS` the profiler shows that most (around 66%) of the `String` objects are allocated by `java.lang.Class.getSimpleName`. 

screen shot 2015-10-05 at 15 06 56

This method also appears in the CPU profile hot spots.

screen shot 2015-10-05 at 15 10 41

sjrd:
Instead of eagerly fetching the simple name of tree classes, we
just store the tree node itself. We only extract the position and
class name for `ErrorContext.toString()`, which is called only if
an error is reported. In successful runs, there is no error
reported, so we avoid a lot of allocations.

ErrorContext itself is also made an AnyVal to further reduce
allocations in the IR checker.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
@@ -1060,9 +1060,16 @@ class IRChecker(unit: LinkingUnit, logger: Logger) {
 }
 
 object IRChecker {
-  private final class ErrorContext private (pos: Position, name: String) {
-    override def toString(): String =
+  private final class ErrorContext private (val treeOrLinkedClass: Any)
+      extends AnyVal {
+
+    override def toString(): String = {
+      val (pos, name) = treeOrLinkedClass match {
+        case tree: Tree               => (tree.pos, tree.getClass.getSimpleName)
+        case linkedClass: LinkedClass => (linkedClass.pos, "ClassDef")
+      }
       s"${pos.source}(${pos.line+1}:${pos.column+1}:$name)"
+    }
   }
 
   private object ErrorContext {
@@ -1070,10 +1077,10 @@ object IRChecker {
       ErrorContext(tree)
 
     def apply(tree: Tree): ErrorContext =
-      new ErrorContext(tree.pos, tree.getClass.getSimpleName)
+      new ErrorContext(tree)
 
     def apply(linkedClass: LinkedClass): ErrorContext =
-      new ErrorContext(linkedClass.pos, "ClassDef")
+      new ErrorContext(linkedClass)
   }
 
   private def isConstructorName(name: String): Boolean =


sjrd:
Review by @nicolasstucki

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
```diff
@@ -1060,9 +1060,16 @@ class IRChecker(unit: LinkingUnit, logger: Logger) {
 }
 
 object IRChecker {
-  private final class ErrorContext private (pos: Position, name: String) {
-    override def toString(): String =
+  private final class ErrorContext private (val treeOrLinkedClass: Any)
```
gzm0: Could you use an `Either` here?

gzm0: Or, if you are concerned about allocations, introduce two subclasses of `ErrorContext`? (Will also remove a pattern match)

sjrd: I was indeed concerned about allocations. Subclasses don't solve the other part of the problem, which is the allocation of the `ErrorContext` itself. It can't be an `AnyVal` if there are two subclasses.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
```diff
@@ -1060,9 +1060,16 @@ class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+      extends AnyVal {
+
+    override def toString(): String = {
+      val (pos, name) = treeOrLinkedClass match {
```
nicolasstucki: Is this tuple supposed to get stack allocated on the JVM?
If not, it might be better to lift the `s"..."` to an inner function and calling it in place of the tuple creation.

nicolasstucki:
That is all, otherwise LGTM.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
sjrd: Not necessarily. It doesn't matter at this point, because `toString` is never called in practice. It's called only if there is an IR checker error, in which case we can be slow and memory intensive and what not.

nicolasstucki: Ok

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
gzm0: I see. So I guess leaving this is fine. But please add a comment stating that this is optimized for allocations in the non-failure case (i.e. if `toString` is not called).

PR #1950, opened 2015-10-13
Adds toString to FlatJSDependency.

matthughes:
Adding toString to FlatJSDependency per https://gitter.im/scala-js/scala-js?at=561c28dc773bc4af20ff1411)

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
@@ -34,4 +34,9 @@ final class FlatJSDependency(
   @deprecated("Use relPath instead.", "0.6.1")
   val resourceName: String = relPath
 
+  override def toString(): String = s"""FlatJSDependency($origin, 
+      dependencies: ${dependencies.mkString(", ")}, 
+      commonJSName: $commonJSName, 
+      relPathMinified: $relPathMinified)""".stripMargin.replaceAll("\n", "")
+
 }
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/JSDependencyManifest.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/JSDependencyManifest.scala
@@ -38,6 +38,10 @@ final class JSDependencyManifest(
     acc = mixLast(acc, compliantSemantics.##)
     finalizeHash(acc, 4)
   }
+
+  override def toString(): String = s"""JSDependencyManifest($origin, 
+    |libDeps: ${libDeps.mkString(", ")}, requiresDOM: $requiresDOM, 
+    |compliantSemantics: $compliantSemantics)""".stripMargin.replaceAll("\n", "")
 }
 
 object JSDependencyManifest {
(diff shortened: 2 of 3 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
```diff
@@ -34,4 +34,6 @@ final class FlatJSDependency(
   @deprecated("Use relPath instead.", "0.6.1")
   val resourceName: String = relPath
 
+  override def toString: String = s"""$origin, dependencies: ${dependencies.mkString(", ")}, commonJSName: $commonJSName, relPathMinified: $relPathMinified"""
```
sjrd: A couple style things:
* `toString()` with `()`
* Try and stay under 80 characters limits; avoid going over 120 characters at all cost.

The full guidelines are here: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md

sjrd: On the content, since there are several things separated by commas, we should surround the whole thing with `FlatJSDependency( ... )`. Otherwise, when you `mkString` a list of `FlatJSDependency, you have no idea where the separations are.

Maybe take inspiration from https://github.com/scala-js/scala-js/blob/47ef7be4d5e05287334b22374cab4beb10085600/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/JSDependency.scala#L83-L94

The two same comments apply to the other two `toString()`s.

sjrd:
That's all.

matthughes:
Addressed comments.

matthughes:
Not clear to me if there is something else this PR is waiting for.

sjrd:
Hum, sorry, I forgot about it. I'll have a look now.

matthughes:
Ok, cool.  Just wanted to make sure you weren't waiting on me.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
```diff
@@ -34,4 +34,9 @@ final class FlatJSDependency(
   @deprecated("Use relPath instead.", "0.6.1")
   val resourceName: String = relPath
 
+  override def toString(): String = s"""FlatJSDependency($origin, 
+      dependencies: ${dependencies.mkString(", ")}, 
+      commonJSName: $commonJSName, 
+      relPathMinified: $relPathMinified)""".stripMargin.replaceAll("\n", "")
```
sjrd: To comply with our [coding style](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md), this should be formatted like this (also adding the missing `|`s):
```scala
  override def toString(): String = {
    s"""FlatJSDependency($origin, 
       |dependencies: ${dependencies.mkString(", ")}, 
       |commonJSName: $commonJSName, 
       |relPathMinified: $relPathMinified)""".stripMargin.replaceAll("\n", "")
  }
```
There is a problem with this formulation, though. It requires a space at the end of the lines, and that is not good at all, as most editor configs (including mine) strip trailing whitespace on saving the files. So it wouldn't be possible for me to maintain this formatting without switching editor.

The general structure that we already have in https://github.com/scala-js/scala-js/blob/47ef7be4d5e05287334b22374cab4beb10085600/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/JSDependency.scala#L83-L94 does not have this issue. I would suggest good old string concatenations instead:
```scala
  override def toString(): String = {
    s"FlatJSDependency($origin, " +
    s"dependencies: ${dependencies.mkString(", ")}, " +
    s"commonJSName: $commonJSName, " +
    s"relPathMinified: $relPathMinified)"
  }
```
which makes the spaces explicit before the `" +` ending each line.

Finally, why no `origin: ` before the `$origin` field, like the other fields?

(The same comments apply to the two other methods.)

sjrd:
Jenkins, test this please

PR #1951, opened 2015-10-13
Comment the optimized version of IRChecker.ErrorContext.

sjrd:


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
```diff
@@ -1060,6 +1060,22 @@ class IRChecker(unit: LinkingUnit, logger: Logger) {
...
 object IRChecker {
+  /** The context in which to report IR check errors.
+   *
+   *  The way this class is written is optimized for the happy path, where no
+   *  error occurs. In that case, `toString()` is never called, and we avoid
+   *  any kind of allocation.
+   *
+   *  The parameter is an `Any` for that reason. It should be an
+   *  `Either[Tree, LinkedClass]`, but that would also require an allocation of
+   *  the `Left` or `Right` (in fact, we'd love to type is as
```
gzm0: we'd love to type *it* as ...

gzm0:
LGTM otherwise

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala
sjrd: Fixed.

gzm0:
LGTM

gzm0:
You can merge this. It just seems I can't do it on mobile if CI hasn't passed. (GH says server error though...)

sjrd:
I've recently enabled *protection* of the `master` branch in this repo ([here](https://github.com/scala-js/scala-js/settings/branches)), so GitHub won't let us merge PRs that don't have a green build. Even I can't on desktop ^^ (although I could go and unprotect it, merge it, then protect it back, but that kind of defeats the purpose :-p)

gzm0:
I see. So we'll actually need a `next` branch for new versions now :)

sjrd:
No for new versions we'll unprotect the branch, and protect it back afterwards. Otherwise there's no way we can avoid having 2 commits with a release version number.

gzm0:
Jenkins, test this please

gzm0:
Really? From the comments in the options, it seems you can also push commits that have a checkmark. So opening a PR and then not merging, but pushing it should do it...

sjrd:
Ah possibly. We'll see when we get there.

PR #1952, opened 2015-10-13
Enable fatal warnings for IR, tools, partest and library

gzm0:


library/src/main/scala/scala/scalajs/macroimpls/UseAsMacros.scala
```diff
@@ -21,6 +21,7 @@ import Compat210._
  *
  *  @author Tobias Schlatter
  */
+@deprecated("Not actually deprecated, makes warnings go away", "")
 private[scalajs] object UseAsMacros {
```
gzm0: I have manually checked that using these macros does not actually trigger a deprecation warning. However, we currently do not have an automated mechanism that checks this (could change in future Scala versions).

A quick and dirty fix would be to abuse one of the example projects as a test (since they get compiled under `-XfatalWarnings` unlike the tests).

gzm0:
Review by @sjrd

sjrd:
That's all.

gzm0:
Updated

sjrd:
LGTM

sjrd:
OK apparently this protected branch stuff is a bit too extreme. It won't let me merge, even though we have a green build, because since then there have been commits on master. So it wants a rebase. With the notorious speed and reliability of our CI, this is not going to fly. I'll disable it.

PR #1953, opened 2015-10-13
Fix #1947: Rewrite calls to print.

### LINKED ISSUE #1947: ir.Printers.IRTreePrinter is printing partially to the console.
This bug was introduced while improving performance of `javascript.Printers`. When the method `def print(args: Any*)` disappeared some of the calls to `print(...)` became calls to `Predef.print`.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -92,7 +92,7 @@ object Printers {
 
         case _ =>
           print('{'); indent(); println()
-          print(tree)
+          printOne(tree)
```
sjrd: This `printOne`, as well as many others, could just as well directly call the specialized `print` for `Tree`. There should be very few calls to `printOne` left, IMO.

sjrd:
Weren't you going to un-import `Predef.{print, println}`?

sjrd:
That's all.

nicolasstucki:
There is a small problem when importing  `Predef.{print, println}` with `import Predef.{print => _, println => _}` it seems that the default `import Predef._` gets overwritten by this one. This causes a couple of other methods to fail because they can't find some implicit conversions. Such as:

```
[error] /Users/nastucki/GitHub/scala-js/ir/src/main/scala/org/scalajs/core/ir/Printers.scala:444: value until is not a member of Int
[error]           for (dim <- lengths.size until tpe.dimensions)
[error]                                    ^
[error] /Users/nastucki/GitHub/scala-js/ir/src/main/scala/org/scalajs/core/ir/Printers.scala:845: value to is not a member of Int
[error]         for (i <- 1 to dims)
[error]                     ^
```

I could rewrite everything more explicitly but I would consider this approach a bit hacky.

sjrd:
What about this:
```scala
import scala.Predef.{print => _, println => _, _}
```
(note the last `_`).

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -201,30 +242,35 @@ object Printers {
...
 
         case Throw(expr) =>
-          print("throw ", expr)
+          print("throw ")
+          print(expr)
 
         case Continue(label) =>
-          if (label.isEmpty) print("continue")
-          else print("continue ", label.get)
+            print("continue")
```
gzm0: Bad indentation?

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -500,19 +650,19 @@ object Printers {
...
 
         case JSObjectConstr(fields) =>
-          print("{"); indent; println()
+          print('{'); indent; println()
           var rest = fields
           while (rest.nonEmpty) {
-            printOne(rest.head._1)
-            printOne(": ")
-            printOne(rest.head._2)
+            print(Objects.toString(rest.head._1))
```
gzm0: `rest.head._1` is a `PropertyName`, IIUC you need to dispatch it through the right print method, otherwise we have a (potential) degradation in source maps.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -610,15 +787,21 @@ object Printers {
             print("var ")
           else
             print("val ")
-          print(name, ": ", vtpe)
+          print(Objects.toString(name))
```
gzm0: Same here: This discards the position

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -610,15 +787,21 @@ object Printers {
...
+          if (tree.optimizerHints != OptimizerHints.empty) {
+            print("@hints(")
+            print(tree.optimizerHints.bits)
+            print(") ")
+          }
           if (static)
             print("static ")
-          print("def ", name)
+          print("def ")
+          print(Objects.toString(name))
```
gzm0: Same here

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -628,7 +811,7 @@ object Printers {
         case PropertyDef(name, _, _, _) =>
           // TODO
           print("<property: ")
-          print(name)
+          print(Objects.toString(name))
```
gzm0: Same here

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -399,40 +506,67 @@ object Printers {
...
+          print(qualifier)
+          print('[')
+          print(item)
+          print(']')
 
         case JSFunctionApply(fun, args) =>
           fun match {
-            case _:JSDotSelect | _:JSBracketSelect | _:Select =>
-              print("protect(", fun, ")")
+            case _: JSDotSelect | _: JSBracketSelect | _: Select =>
```
sjrd: Coding style: no space after `:` when it's used in a `|`-separated list of alternatives in a `case`.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -717,9 +894,15 @@ object Printers {
...
       printEscapeJS(encodedName, out)
       println()
-      println("isExported: ", isExported)
-      println("kind: ", kind)
-      println("superClass: ", superClass)
+      print("isExported: ")
+      print(isExported.toString)
+      println()
+      print("kind: ")
+      print(if (kind == null) "null" else kind.toString)
```
sjrd: I'm pretty sure `kind` cannot be `null`.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -232,57 +276,97 @@ object Printers {
...
         case BinaryOp(BinaryOp.Float_-, FloatLiteral(0.0f), rhs) =>
-          print("(-", rhs, ")")
+          print("(-")
+          print(rhs)
+          print(')')
+
         case BinaryOp(BinaryOp.Double_-,
-            IntLiteral(0) | FloatLiteral(0.0f) | DoubleLiteral(0.0), rhs) =>
-          print("(-", rhs, ")")
+        IntLiteral(0) | FloatLiteral(0.0f) | DoubleLiteral(0.0), rhs) =>
```
sjrd: Indentation issue.

sjrd:
LGTM

PR #1954, opened 2015-10-13
Fix #1903: Remove noWarnMissing option from Linker

### LINKED ISSUE #1903: Remove `noWarnMissing` option of linker
The option was originally introduced to allow to run partests which do not link properly. Since we support much more of the Java lib now, this should be less of an issue.

I'll investigate the impact on partests.

gzm0:
This breaks binary compatibility of the external tools API (as scheduled for 0.6.6).

### DIFF
--- a/tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
+++ b/tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
@@ -167,8 +167,7 @@ object ScalaJSClosureOptimizer {
       val output: WritableVirtualJSFile,
       /** Cache file */
       val cache: Option[WritableVirtualTextFile],
-      /** Whether to only warn if the linker has errors. Implicitly true, if
-       *  noWarnMissing is nonEmpty */
+      /** Whether to only warn if the linker has errors. */
       val bypassLinkingErrors: Boolean,
       /** If true, performs expensive checks of the IR for the used parts. */
       val checkIR: Boolean,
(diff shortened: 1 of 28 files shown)


gzm0:
Review by @sjrd

tools/jvm/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSClosureOptimizer.scala
```diff
@@ -290,48 +230,15 @@ object ScalaJSClosureOptimizer {
...
-
-    // For binary compatibility
-    override def equals(other: Any): Boolean = super.equals(other)
-
-    // For binary compatibility
-    override def hashCode(): Int = super.hashCode()
-
-    // For binary compatibility
-    override def toString(): String =
-      productArray.mkString("Config(", ", ", ")")
```
sjrd: Maybe we want to actually keep a meaningful-ish `toString()`.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/ScalaJSOptimizer.scala
```diff
@@ -320,48 +256,15 @@ object ScalaJSOptimizer {
...
-
-    // For binary compatibility
-    override def equals(other: Any): Boolean = super.equals(other)
-
-    // For binary compatibility
-    override def hashCode(): Int = super.hashCode()
-
-    // For binary compatibility
-    override def toString(): String =
-      productArray.mkString("Config(", ", ", ")")
```
sjrd: Same here about `toString()`

sjrd:
That's all, good job :-)

I'm surprised you didn't introduce a big all-caps

> I DELETE CODE

in the PR description :p

gzm0:
Updated

sjrd:
LGTM

gzm0:
Update. One of the buglists was wrong (2.12.0-M2).

PR #1962, opened 2015-10-21
Faster implementation of RuntimeLong

sjrd:


sjrd:
Review by @gzm0 and @nicolasstucki 
For benchmarks, see https://github.com/SebsLittleHelpers/scala-js-long-benchmark

/cc @paullepoulpe and @tOverney

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -2,52 +2,114 @@ package scala.scalajs.runtime
...
-  val l: Int,
-  val m: Int,
-  val h: Int
-) extends Number with Comparable[java.lang.Long] { x =>
+import scala.scalajs.js
+import js.JSNumberOps._
+import js.JSStringOps._
+
+/** Emulates a Long on the JavaScript platform. */
+class RuntimeLong(val lo: Int, val hi: Int)
```
nicolasstucki: Is there a reason for dropping the `final`?

sjrd: No. It should have stayed indeed.

sjrd:
Jenkins, retest this please

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +117,553 @@ final class RuntimeLong(
...
+    Integer.bitCount(lo) + Integer.bitCount(hi)
 
-          val digits = rem.toInt.toString
-          val zeroPrefix =
-            if (quot.isZero) ""
-            else "000000000".substring(digits.length) // (9 - digits.length) zeros
+  def signum: RuntimeLong = {
+    val hi = this.hi
+    if (hi < 0) MinusOne
+    else if (hi == 0 && lo == 0) Zero
```
nicolasstucki: Maybe you could use `(hi | lo) == 0`.

sjrd: Ah ah, nice :-)

test-suite/src/test/scala/org/scalajs/testsuite/javalib/lang/LongTest.scala
```diff
@@ -246,6 +246,29 @@ object LongTest extends JasmineTest {
       // scalastyle:on disallow.space.before.token disallow.space.after.token
     }
 
+    it("should correctly compute leading zeros") {
+      expect(JLong.numberOfLeadingZeros(0x272d130652a160fL)).toEqual(6)
```
nicolasstucki: Should add a case where the number of leading zeros is 0.

nicolasstucki:
That is all. Otherwise LGTM.

sjrd:
Updated.

I had to adapt [`unsignedDivModHelper`](https://github.com/scala-js/scala-js/pull/1962/files#diff-50d64391eac98f26c68a9611f5656e25R431) not to be `@inline`, because otherwise its 3 copies (in `unsigned_/`, `unsigned_%` and `toUnsignedString` caused an increase of code size in Reversi). So instead of returning a 4-tuple which was supposed to be stack-allocated, it receives an additional param telling which side the caller wants. For `toUnsignedString`, we return a `js.Tuple4` of the four components.

sjrd:
retest this please

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -2,52 +2,118 @@ package scala.scalajs.runtime
...
+    extends java.lang.Number with java.io.Serializable
+    with java.lang.Comparable[java.lang.Long] { a =>
 
   import RuntimeLong._
+  import Utils._
+
+  /** Constructs a Long from an Int. */
+  def this(value: Int) = this(value, value >> 31)
+
+  // Binary compatibility for the old (l, m, h) encoding
```
gzm0: Why do we need binary compat? `RuntimeLong`s only get reified in the linker...

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+  // Shifts
+
+  /** Shift left */
+  def <<(n0: Int): RuntimeLong = {
+    val n = n0 & 63
+    val lo = this.lo
+
+    if (n == 0) this
+    else if (n < 32) new RuntimeLong(lo << n, (lo >>> -n) | (hi << n))
+    else new RuntimeLong(0, lo << n)
```
gzm0: Any reason you don't use `inline_<<` here?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+  }
+
+  /** Logical shift right */
+  def >>>(n0: Int): RuntimeLong = {
+    val n = n0 & 63
+    val hi = this.hi
 
+    if (n == 0) this
+    else if (n < 32) new RuntimeLong((lo >>> n) | (hi << -n), hi >>> n)
+    else new RuntimeLong(hi >>> n, 0)
```
gzm0: Same here. Why not `inline_>>>`?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Oh... I see... can't reuse `this`. OK

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Same here. Can't reuse this. maybe drop a comment on both these guys.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: That's if an old linker, say the sbt plugin 0.6.5, tries to link a codebase with a more recent version of `scalajs-library.jar` on the classpath.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+  @inline
+  private def inline_abs(lo: Int, hi: Int): (Boolean, Int, Int) = {
+    val neg = hi < 0
+    var absLo = lo
+    var absHi = hi
+    if (neg) {
+      absLo = -lo
+      absHi = if (lo != 0) ~hi else -hi
+    }
+    (neg, absLo, absHi)
```
gzm0: I'm assuming there is a very good reason why you use `var`s instead of two return paths. Please mention it.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+  }
 
-  override def hashCode(): Int = {
-    (this ^ (this >>> 32)).toInt
+  def +(b: RuntimeLong): RuntimeLong = {
+    val alo = a.lo
+    val blo = b.lo
+    val lo = alo + blo
+    val hi = a.hi + b.hi + (if (inlineUnsignedInt_<(lo, alo)) 1 else 0)
+    new RuntimeLong(lo, hi)
```
gzm0: Now: Why don't you use `inline_+` here?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    val lo = alo + blo
+    (lo, ahi + bhi + (if (inlineUnsignedInt_<(lo, alo)) 1 else 0))
+  }
+
+  def -(b: RuntimeLong): RuntimeLong = {
+    val alo = a.lo
+    val blo = b.lo
+    val lo = alo - blo
+    val hi = a.hi - b.hi + (if (inlineUnsignedInt_>(lo, alo)) -1 else 0)
+    new RuntimeLong(lo, hi)
```
gzm0: Same: Why not `inline_-`?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    val a3 = a.hi >>> 16
+    val b0 = b.lo & 0xFFFF
+    val b1 = b.lo >>> 16
+    val b2 = b.hi & 0xFFFF
+    val b3 = b.hi >>> 16
+
+    var c0 = a0 * b0
+
+    var c1 = c0 >>> 16
+    c1 = c1 + a1 * b0
```
gzm0: Use `+=`?

gzm0: Ah... maybe for consistency it is indeed better not to..

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    if (isZero(blo, bhi))
+      throw new ArithmeticException("/ by zero")
+
+    if (isInt32(alo, ahi)) {
+      if (isInt32(blo, bhi)) {
+        if (alo == Int.MinValue && blo == -1) new RuntimeLong(Int.MinValue, 0)
+        else new RuntimeLong(alo / blo)
+      } else {
+        // Either a == Int.MinValue && b == (Int.MaxValue + 1), or (abs(b) > abs(a))
+        if (alo == Int.MinValue && (blo == 0x80000000 && bhi == 0)) MinusOne
```
gzm0: Your usage of constants is inconsistent. Above you used `Int.MinValue` to express `Int.MaxValue + 1`. Here you use its value.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+      val (bNeg, bAbsLo, bAbsHi) = inline_abs(blo, bhi)
+      val absR = unsigned_/(aAbsLo, aAbsHi, bAbsLo, bAbsHi)
+      if (aNeg == bNeg) absR
+      else inlineLongUnary_-(absR.lo, absR.hi)
     }
-    if (b4 != 0) {
-      p4 += a0 * b4;
+  }
+
+  private def unsigned_/(alo: Int, ahi: Int, blo: Int, bhi: Int): RuntimeLong = {
```
gzm0: Maybe drop a comment that this works under the assumption that `a, b` do not fit into an integer.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    if ((ahi & 0xffe00000) == 0) {
```
gzm0: This constant occurs quite a couple of times. How about you give it a nice name?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+      if ((bhi & 0xffe0000) == 0) {
+        val aDouble = ahi * TwoPow32 + alo.toUint
```
gzm0: This might also deserve an inlined helper.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Also, that this is a well formed division (i.e. `b != `)

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+        val rDouble = aDouble / bDouble
+        new RuntimeLong(rawToInt(rDouble), rawToInt(rDouble / TwoPow32))
+      } else {
+        Zero // because b > a
+      }
+    } else {
+      if (bhi == 0 && isPowerOfTwo_IKnowItsNot0(blo)) {
+        val pow = log2OfPowerOfTwo(blo)
+        if (pow == 0) new RuntimeLong(alo, ahi)
+        else new RuntimeLong((alo >>> pow) | (ahi << -pow), ahi >>> pow)
```
gzm0: use `inline_>>>`

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Ohh... nifty... Indeed we need that :)

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
-    val c12 = (p2 & 0x3ffff) << 4;
-    val c13 = (p3 & 0x1f) << 17;
-    val c1 = c10 + c11 + c12 + c13;
+    if (isInt32(alo, ahi)) {
+      if (isInt32(blo, bhi)) {
+        if (blo != -1) new RuntimeLong(alo % blo)
+        else Zero // Work around https://github.com/ariya/phantomjs/issues/12198
+      } else {
+        // Either a == Int.MinValue && b == (Int.MaxValue + 1), or (abs(b) > abs(a))
+        if (alo == Int.MinValue && (blo == 0x80000000 && bhi == 0)) Zero
```
gzm0: Adapt the constant depending on what you chose in `/`.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+        else a // because abs(b) > abs(a)
+      }
+    } else {
+      val (aNeg, aAbsLo, aAbsHi) = inline_abs(alo, ahi)
+      val (bNeg, bAbsLo, bAbsHi) = inline_abs(blo, bhi)
```
gzm0: Use `_` instead of `bNeg` as you are not using it.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
 
-    val c22 = p2 >> 18;
-    val c23 = p3 >> 5;
-    val c24 = (p4 & 0xfff) << 8;
-    val c2 = c22 + c23 + c24;
+  private def unsigned_%(alo: Int, ahi: Int, blo: Int, bhi: Int): RuntimeLong = {
+    if ((ahi & 0xffe00000) == 0) {
+      if ((bhi & 0xffe00000) == 0) {
+        val aDouble = ahi * TwoPow32 + alo.toUint
+        val bDouble = bhi * TwoPow32 + blo.toUint
```
gzm0: Same story here. Consider having helpers for the whole "do I fit in a double story". Will make the code easier to read.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
-    // Propagate high bits from c0 -> c1, c1 -> c2
-    val c1n = c1 + (c0 >> BITS)
+  private def unsignedDivModHelper(alo: Int, ahi: Int, blo: Int, bhi: Int,
+      ask: Int): RuntimeLong | js.Tuple4[Int, Int, Int, Int] = {
+
+    var shift =
+      inlineNumberOfLeadingZeros(blo, bhi) - inlineNumberOfLeadingZeros(alo, ahi)
+    val initialBShift = inline_<<(blo, bhi, shift)
+    var bShiftLo = initialBShift._1
+    var bShiftHi = initialBShift._2
```
gzm0: use `var (bShiftLo, bShiftHi) = inline_<<(blo, bhi, shift)` (I checked that it actually does what you need :P)

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    var quotLo = 0
+    var quotHi = 0
+
+    /* Invariants:
+     *   bShift == b << shift == b * 2^shift
+     *   quot >= 0
+     *   0 <= rem < 2 * bShift
+     *   quot * b + rem == a
+     */
+    while (shift >= 0 && (remHi & 0xffe00000) != 0) {
```
gzm0: Same, here: use `fitsInDouble` or something :)

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+      if (inlineUnsigned_>=(remLo, remHi, bShiftLo, bShiftHi)) {
+        val newRem = inline_-(remLo, remHi, bShiftLo, bShiftHi)
```
gzm0: Use `val (remLo, remHi) = inline_-(remLo, remHi, bShiftLo, bShiftHi)`

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+        remHi = newRem._2
+        if (shift < 32)
+          quotLo |= (1 << shift)
+        else
+          quotHi |= (1 << shift) // == (1 << (shift - 32))
+      }
+      shift -= 1
+      val newBShift = inline_>>>(bShiftLo, bShiftHi, 1)
+      bShiftLo = newBShift._1
+      bShiftHi = newBShift._2
```
gzm0: Use `val (..., ...) =`.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+        unsignedDivModHelper(alo, ahi, blo, bhi,
+            AskRemainder).asInstanceOf[RuntimeLong]
+      }
+    }
+  }
 
-    // Propagate high bits from c0 -> c1, c1 -> c2
-    val c1n = c1 + (c0 >> BITS)
+  private def unsignedDivModHelper(alo: Int, ahi: Int, blo: Int, bhi: Int,
+      ask: Int): RuntimeLong | js.Tuple4[Int, Int, Int, Int] = {
```
gzm0: It seems there is a reason you are never using the unapply assignment syntax in this method. If this is the case, please document. Otherwise replace.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+
+    if (hi != 0) Integer.toBinaryString(hi) + padBinary32(lo)
+    else Integer.toBinaryString(lo)
   }
 
   def toHexString: String = {
-    val zeros = "000000" // 6 zeros
-    @inline def padHex(i: Int, len: Int) = {
+    val zeros = "00000000" // 8 zeros
+    @inline def padBinary8(i: Int) = {
```
gzm0: This should be called `padHex8`. But feel free to inline these. They don't make a lot of sense anymore since they are used only once.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
 
-  // helpers //
+  @deprecated("Just use `this` instead.", "0.6.6")
+  def unary_+ : RuntimeLong = this // scalastyle:ignore
 
-  @inline private def isZero = l == 0 && m == 0 && h == 0
-  @inline private def isMinValue = x.equals(MinValue)
-  @inline private def isNegative = (h & SIGN_BIT_VALUE) != 0
-  @inline private def abs = if (isNegative) -x else x
+  @deprecated("Juse `this.toString + y` instead.", "0.6.6")
```
gzm0: Just use

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+}
 
-  def numberOfLeadingZeros: Int = {
-    if (h != 0)      Integer.numberOfLeadingZeros(h) - (32 - BITS2)
-    else if (m != 0) Integer.numberOfLeadingZeros(m) - (32 - BITS) + (64 - BITS01)
-    else             Integer.numberOfLeadingZeros(l) - (32 - BITS) + (64 - BITS)
-  }
+object RuntimeLong {
+  private final val TwoPow32 = 4294967296.0
+  private final val TowPow63 = 9223372036854775808.0
```
gzm0: `TwoPow63`... You use too much autocomplete if you haven't noticed that

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
 
-  def numberOfTrailingZeros: Int = {
-    if (l != 0)      Integer.numberOfTrailingZeros(l)
-    else if (m != 0) Integer.numberOfTrailingZeros(m) + BITS
-    else             Integer.numberOfTrailingZeros(h) + BITS01
-  }
```
gzm0: maybe a little newline here?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
-            else
-              absX
-          }
-        }
-        divModHelper(newX, absY, xNegative, yNegative, xMinValue)
-      }
+      val neg = value < 0
+      val absValue = if (neg) -value else value
+      val lo = rawToInt(absValue)
+      val hi = rawToInt(absValue / TwoPow32)
```
gzm0: Actually, it seems that the `rawToInt(lo)` / `rawToInt(hi / TwoPow32)` is also a candidate for an inlined helper.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
-    val shift = y.numberOfLeadingZeros - x.numberOfLeadingZeros
-    val yShift = y << shift
+    /** Performs the JavaScript operation `(x | 0)`. */
+    @inline def rawToInt(x: Double): Int =
+      (x.asInstanceOf[js.Dynamic] | 0.asInstanceOf[js.Dynamic]).asInstanceOf[Int]
 
-    val (absQuot, absRem) = divide0(shift, yShift, x, Zero)
+    /** Tests whether the given unsigned Int is an exact power of 2. */
+    @inline def isPowerOfTwo(i: Int): Boolean =
+      i != 0 && isPowerOfTwo_IKnowItsNot0(i)
```
gzm0: This seems unused.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+    else One
+  }
 
-          toString0(quot, zeroPrefix + digits + acc)
-        }
+  def numberOfLeadingZeros: Int = {
+    val hi = this.hi
+    if (hi != 0) Integer.numberOfLeadingZeros(hi)
+    else Integer.numberOfLeadingZeros(lo) + 32
+  }
```
gzm0: use `inlineNumberOfLeadingZeros`

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
+      (b ^ 0x80000000) < (a ^ 0x80000000)
+    }
+
+    @inline
+    def inlineUnsignedInt_>=(a: Int, b: Int): Boolean =
+      (a ^ 0x80000000) >= (b ^ 0x80000000)
+
+    @inline
+    def inlineLongUnary_-(lo: Int, hi: Int): RuntimeLong =
+      new RuntimeLong(-lo, if (lo != 0) ~hi else -hi)
```
gzm0: If I were to be a pedant, I'd say: "Use `inline_unary_-`" :)

project/BinaryIncompatibilities.scala
```diff
@@ -20,6 +20,40 @@ object BinaryIncompatibilities {
...
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.SIGN_BIT_VALUE"),
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.TWO_PWR_15_DBL"),
+
+      // Synthetic methods inside RuntimeLong, not an issue
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.scala$scalajs$runtime$RuntimeLong$$isZero"),
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.scala$scalajs$runtime$RuntimeLong$/Mod"),
```
gzm0: Hahahahaha, scala's encoding is such a fail :)

project/BinaryIncompatibilities.scala
```diff
@@ -20,6 +20,40 @@ object BinaryIncompatibilities {
...
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.TenPow9"),
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.SIGN_BIT"),
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.SIGN_BIT_VALUE"),
+      ProblemFilters.exclude[MissingMethodProblem](
+          "scala.scalajs.runtime.RuntimeLong.TWO_PWR_15_DBL"),
+
+      // Synthetic methods inside RuntimeLong, not an issue
```
gzm0: What do you mean by synthetic methods? Stuff from delambdafy?

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/JSDesugaring.scala
```diff
@@ -160,6 +160,14 @@ private[javascript] object JSDesugaring {
...
     private val isStrongMode = outputMode == OutputMode.ECMAScript6StrongMode
 
+    // TODO Get rid of this when we break backward binary compatibility
+    private lazy val hasNewRuntimeLong = {
+      val rtLongClass = classEmitter.linkedClassByName(LongImpl.RuntimeLongClass)
+      rtLongClass.memberMethods.exists { linkedMethod =>
+        linkedMethod.tree.name.name == LongImpl.initFromParts
+      }
+    }
+
```
gzm0: Now this one I don't understand. You are not allowed to run a `0.6.x` linker with a `0.6.y` library if `y < x`.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/LongImpl.scala
```diff
@@ -83,11 +83,18 @@ object LongImpl {
...
+  final val initFromParts    = "init___I__I"
+  final val initFromPartsOld = "init___I__I__I"
+  final val initFromInt      = "init___I"
+
+  /* Note that initFromPartsOld depends on initFromParts in the new
+   * implementation of RuntimeLong, so this ensures that the new one is
+   * reachable.
+   * We do not directly refer to the new one, other we would crash on earlier
+   * versions of RuntimeLong that do not yet have the new constructor.
+   */
```
gzm0: Add the same TODO than in `JSDesugaring`.

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/JSDesugaring.scala
sjrd: Yes, that is allowed. In general, it is allowed to use a 0.6.x linker with any 0.6.y library, except if y > x and the emitted IR version changed between x and y.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: `inlineNumberOfLeadingZeros` forces to read `lo` from the heap even if it is not needed.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Yes, unapply assignment is not yet sufficiently optimized, especially for specialized tuple types. It creates unnecessary temporary variables. Since this is a code-size- and performance critical file, I avoid it pretty much all the time.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Yep, code size explosion. It's actually the multi-path result of a specialized tuple which is very damaging right now. So if you return `(Int, Int)` from the two branches of an if (and the condition is not constant-folded to `false`), you pay a heavy price at the moment (much lower than actually allocating the tuple, but still).

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: I don't remember. I think this one (and `-`) is an accident of history. I wrote the helpers after the entry points.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: There is `this`, but most importantly it's the multi-path result of a tuple which is not that great. Better to push the `new RuntimeLong`s inside the branches of the if/else.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: It's semantic. In `alo == Int.MinValue`, `alo` is an int32 and we compare it to the int32 value `Int.MinValue` (which is *not* `Int.MaxValue + 1` in Longs). However, `blo` is not an int32, it is the low bits of an int64, so I'm comparing it to the bit pattern I'm interested in, which represents the value `Int.MaxValue + 1`. Maybe I should have written `Int.MaxValue.toLong + 1L` instead, or `2^31`?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Indeed. I guess I could also make it an inline helper.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: `inline_>>>` would have an additional check for `pow >= 32`, which we know is not true here, but we have no well to give that information to the optimizer so that it will eliminate it.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Yep.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Actually, here, will have to keep it in that form (though with a good name for the constant), become `remHi` is a `var`. That means that inlining would cause `{ val hi$5 = remHi; (hi & 0xffe00000) != 0 }`. That would not be a problem if that did not end up in the condition of a `while`, but here, it would cause the `while` loop to be expanded as a `while (true)` if an `if` and a `break`, so a lot more code, and probably slower.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Unfortunately, that produces more code, even not using the unapply assignment.

project/BinaryIncompatibilities.scala
sjrd: Stuff. You never know what Scalac is going to emit for stuff. But they're marked as "synthetic", and that, coupled with the fact they apparently look like they come for previously private methods, generally means it's safe to ignore the changes.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Ok. I see. Maybe you should drop a comment at the top of the file that this is a concern (most of my comments seem to about about this).

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: I see. Let's leave it like that :)

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Ok

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: No need to as discussed above.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Same story for comment about unapply inlining.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Ok

project/BinaryIncompatibilities.scala
gzm0: Ok

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -55,531 +121,558 @@ final class RuntimeLong(
...
-    if (b3 != 0) {
-      p3 += a0 * b3;
-      p4 += a1 * b3;
+  private def inline_-(alo: Int, ahi: Int, blo: Int, bhi: Int): (Int, Int) = {
+    val lo = alo - blo
+    (lo, ahi - bhi + (if (inlineUnsignedInt_>(lo, alo)) -1 else 0))
+  }
+
+  def *(b: RuntimeLong): RuntimeLong = {
+    val a0 = a.lo & 0xFFFF
```
paullepoulpe: Any reason why you don't use local variables to bind a.lo/b.lo/a.hi/b.hi here ?

gzm0: Ah, ah nice :)

sjrd: Nope, I overlooked those, thanks :p

sjrd:
Updated.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: There is actually a newline.

gzm0: Ah... Seems I'm unable to read diffs now...

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -2,52 +2,153 @@ package scala.scalajs.runtime
...
-    else if (isNegative) -((-x).toDouble)
-    else l + m * TWO_PWR_22_DBL + h * TWO_PWR_44_DBL
+
+  def toDouble: Double = {
+    val lo = this.lo
+    val hi = this.hi
+
+    if (hi < 0) {
+      val (abslo, abshi) = inline_unary_-(lo, hi)
+      -(abshi.toUint * TwoPow32 + abslo.toUint) // abshi.toUint for MinValue
```
gzm0: I do not understand the comment (anymore?)

gzm0:
LGTM otherwise. (I just want an explanation on that comment :))

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: In all other conversions to Double, we do `hi * TwoPow32 + lo.toUint`. But here we also need `hi.toUint` for the very lonely case where `hi == 0x80000000` and `lo == 0` (which is `Long.MinValue`), in which case `abshi` is still `0x80000000` but it must be interpreted as unsigned to give 2^31 (not -2^31).

gzm0: Ah, ok. Makes sense.

gzm0:
LGTM

sjrd:
checksizes need (more) tihgtening

sjrd:
@nicolasstucki Anything else to say about this?

nicolasstucki:
Nothing more to say about this, LGTM.

Issue #1970, opened 2015-10-23
Fix links to JavaDoc


gzm0:
Links to the JavaDoc are broken.

See http://stackoverflow.com/questions/16934488/how-to-link-classes-from-jdk-into-scaladoc-generated-doc/ for a fix.

PR #1965, opened 2015-10-22
Make version check not choke in community build.

SethTisue:

PR #1971, opened 2015-10-23
Fix #1917: Add -Xfatal-warnings for ScalaDoc generation

### LINKED ISSUE #1917: Add -Xfatal-warnings even for scaladoc generation
For some sub-projects, scaladoc generation emit warnings and since #1912 fails.

As a temporary solution #1916 disables fatal warnings for doc compilation.

gzm0:
Note that links to the JavaDoc are not yet correct (filed as #1970).

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -56,6 +56,10 @@ object Build extends sbt.Build {
   val newScalaBinaryVersionsInThisRelease: Set[String] =
     Set()
 
+  val javaVersion = settingKey[Int](
+    "The major Java SDK version that should be assumed for compatibility. " +
+    "Defaults to what sbt is running with.")
+
   val previousArtifactSetting: Setting[_] = {
     previousArtifact := {
       val scalaV = scalaVersion.value
@@ -114,7 +118,26 @@ object Build extends sbt.Build {
           "-unchecked",
           "-feature",
           "-encoding", "utf8"
-      )
+      ),
+
+      // Scaladoc linking
+      apiURL := {
+        val name = normalizedName.value
+        Some(url(s"http://www.scala-js.org/api/$name/$scalaJSVersion/"))
+      },
+      autoAPIMappings := true,
+
+      // Add Java Scaladoc mapping (resulting links are broken, see #1970)
+      apiMappings += {
+        val rtJar = {
+          System.getProperty("sun.boot.class.path")
+            .split(java.io.File.pathSeparator)
+            .find(_.endsWith(java.io.File.separator + "rt.jar")).get
+        }
+
+        file(rtJar) -> url("http://docs.oracle.com/javase/8/docs/api/")
+      }
+
   ) ++ mimaDefaultSettings
 
   val publishSettings = Seq(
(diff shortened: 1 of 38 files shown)


gzm0:
Review by @nicolasstucki
Review by @sjrd for the principle of fixing JavaDoc later and the changes to the doc of `js.UndefOr`

nicolasstucki:
LGTM

gzm0:
Updated. Add filter for `scalacOptions` in `javalibEx`

project/Build.scala
```diff
@@ -114,7 +114,26 @@ object Build extends sbt.Build {
...
+      ),
+
+      // Scaladoc linking
+      apiURL := {
+        val name = normalizedName.value
+        Some(url(s"http://www.scala-js.org/api/$name/$scalaJSVersion/"))
+      },
+      autoAPIMappings := true,
+
+      // Add Java Scaldoc mapping (resulting links are broken, see #1970)
```
sjrd: typo Scaldoc

project/Build.scala
```diff
@@ -114,7 +114,26 @@ object Build extends sbt.Build {
...
+      apiMappings += {
+        val rtJar: String = {
```
sjrd: Remove `: String`?

sjrd:
> Review by @sjrd for the principle of fixing JavaDoc later and the changes to the doc of js.UndefOr

OK

gzm0:
Updated. Review of `Build.scala` by @sjrd

project/Build.scala
```diff
@@ -56,6 +56,16 @@ object Build extends sbt.Build {
...
     Set()
 
+  val javaVersion: Int = {
+    val v = System.getProperty("java.version")
+    v.substring(0, 3) match {
+      case "1.8" => 8
+      case "1.7" => 7
+      case "1.6" => 6
+      case _     => sys.error(s"Unknown JDK version $v")
+    }
```
sjrd: I think this val should be an `Option[Int]`, and we should do something meaningful if we have an unknown JDK, rather than blowing up (possibly printing a warning). What if someone wants to try out Scala.js on beta builds of JDK 9, for example? We recently had a similar problem with the community build (#1965), albeit with the Scala version.

sjrd: If we want to assert in our CI that we know the JDK version we're testing, we can add a *task* `ensureKnownJavaVersion` that will thrown if `javaVersion.isEmpty`, and we can call that task from the matrix.

project/Build.scala
```diff
@@ -1003,14 +1054,19 @@ object Build extends sbt.Build {
           })),
 
           sources in Test ++= {
-            def listAllScalaFilesIf(testDir: String, condition: Boolean): Seq[File] =
+            def listAllScalaFilesIf(testDir: String, condition: Boolean): Seq[File] = {
```
sjrd: Why the `{`?

sjrd:
That's all for `Build.scala`

project/Build.scala
gzm0: Indeed... Thought we need this... :)

project/Build.scala
gzm0: Ok. I wanted to issue a warning, but the I needed to make a setting out of it... Was too lazy :)

gzm0: Would the following be OK too?

```scala
            streams.value.log.warning(
                s"Unknown JDK version $v. Assuming maximal compatibility")
            Int.MaxValue
```

It is kind of hard to decide on something meaningful at usage site if you get a `None`. Thanks to the rock-solid backwards compatibility of the JDK, `Int.MinValue` and `Int.MaxValue` seem to actually be reasonable values here...

gzm0:
Updated

sjrd:
`Build.scala` LGTM

gzm0:
Updated (ditched `patmatSafeFatalWarningsSettings`)

sjrd:
Re-`Build.scala` LGTM

gzm0:
@nicolasstucki this is likely to pass CI now. Does it still LGTY?

gzm0:
@nicolasstucki would you mind taking a quick look? (Only `Build.scala` has changed since your last review).
Sorry for being pushy. I have another commit coming up behind this and this is likely to conflict with changes I am about to make (tools refactoring). Thanks!

nicolasstucki:
LGTM

PR #1973, opened 2015-10-23
Fix #1972: Always link source maps in SJS compiled code

### LINKED ISSUE #1972: SJS versions of IR and tools lack proper source map linking

gzm0:
This probabaly brings a minor performance degradation when compiling the
tests for a release version (since the relativized URIs need to be
calculated).


gzm0:
Review by @sjrd

sjrd:
LGTM. Makes me wonder why we didn't do that in the first place ...

PR #1977, opened 2015-10-27
Fix #1976: Fix Int and Long literals in ir.Printers

### LINKED ISSUE #1976: The IR printer does not print {Int|Long}Literal correctly

nicolasstucki:


nicolasstucki:
Review by @sjrd

sjrd:
LGTM

gzm0:
Is it absolutely, 100% impossible to write a test for this? It seems we are running into printer issues a lot of times (i.e. more than we should) and simple regression unit tests could mitigate this.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -681,10 +681,19 @@ object Printers {
...
+            print(')')
+          }
+
+        case LongLiteral(value) =>
+          if (value >= 0) {
+            print(value.toString)
+          } else {
+            print('(')
+            print(value.toString)
             print(')')
```
gzm0: Maybe we should append an `L` here?

nicolasstucki: Appending an `L` would append the value rather than the string representation. Couldn't it?

gzm0: Oh, yes. Probably. I mean that the end result should show an "L". No matter how you achieve it :)
(probably just adding `print('L')` is easiest).
It might just be nice to see the distinction between an int literal and a long literal. Just like in the IR we also print things like `+[int]`.

nicolasstucki: In that case I would assume that we would also add them to floats and doubles. I believe that a change of representation like this should be in another commit.

gzm0: Indeed. But we should maybe think about whether we should change it @sjrd?

sjrd: Yes, we could add `f` to floats, `d` to doubles and `L` to longs. But indeed it should be in a different commit.

nicolasstucki:
Then let me add a second commit to this PR.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
nicolasstucki: Then let me add a second commit to this PR.

gzm0:
I think you meant "suffix".

gzm0:
Otherwise LGTM

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -681,33 +681,43 @@ object Printers {
...
+            print(value.toString)
           } else {
             print('(')
-            print(value)
+            print(value.toString)
             print(')')
           }
 
+        case LongLiteral(value) =>
+          if (value < 0.0f)
```
sjrd: `0L` or just `0`, but certainly not `0.0f` here ;)

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -681,33 +681,43 @@ object Printers {
...
+            print('(')
+          print(value.toString)
+          print('L')
+          if (value < 0.0f)
```
sjrd: Same.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -681,33 +681,43 @@ object Printers {
...
+            print(')')
+
         case FloatLiteral(value) =>
           if (value == 0.0f && 1.0f / value < 0.0f) {
             print("(-0)")
```
sjrd: `(-0f)` with an `f` too.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -681,33 +681,43 @@ object Printers {
...
             print(value.toString)
-            print(')')
+            print('f')
+            if (value < 0.0f)
+              print(')')
           }
 
         case DoubleLiteral(value) =>
           if (value == 0.0 && 1.0 / value < 0.0) {
             print("(-0)")
```
sjrd: `(-0d)` with a `d` too.

sjrd:
That's all.

sjrd:
LGTM

Issue #791, opened 2014-07-02
Default arguments in the constructor of facade types fail if the companion is not a js.Any


gzm0:
How to reproduce:

``` scala
import scala.scalajs.js

object HelloWorld extends js.JSApp {

  def main(): Unit = {
    // Define JS class A
    js.eval("var A = function A(x) {}");

    // Instantiate
    new A(1)
  }
}

class A extends js.Object {
  def this(x: Int, y: Int = ???) = this()
}

// If you uncomment this, it works
//object A extends js.Object
```

Will throw a `NotImplementedError` coming from the `???`.

PR #1982, opened 2015-10-28
Fix #1811: Add support for secondary constructors in JS classes

### LINKED ISSUE #1811: Support secondary constructors in JS classes
Currently an implementation restriction wrt. #1795, coming from the implementation in #1809.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val newBody = js.Block(overloadSelection :::
+            preSuperCall :: superCall :: postSupperCall :: Nil)
+
+        val newConstructor = js.MethodDef(static = false, dispatchName, dispatchArgs,
+            jstpe.NoType, newBody)(dispatch.optimizerHints, dispatch.hash)
+
+        // Boldly get rid of everything in infoBuilder, because it's broken
+        currentClassInfoBuilder.addMethod(
+            ir.Infos.generateMethodInfo(newConstructor))
```
gzm0: Why / what is it broken? How hard is it to fix it?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          js.If(lhs, rhs, js.BooleanLiteral(false))(jstpe.BooleanType)
+      }
+    }
+
+    private def zipMap[T, U, V](xs: List[T], ys: List[U])(
+        f: (T, U) => V): List[V] = {
+      if (xs.nonEmpty && ys.nonEmpty)
+        f(xs.head, ys.head) :: zipMap(xs.tail, ys.tail)(f)
+      else
+        Nil
```
gzm0: Why this? Seems like premature optimization...

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: It's the fact that the secondary constructors call other constructors with `ApplyStatically`. While their tree is built, these calls are registered to the info builder. But these called constructors are afterwards removed, because their bodies are mangled into an only JS constructor, and hence the info builder says that it calls methods that do not, in fact, exist.

I don't know how hard exactly it would be to fix it, but it would be hacky in any case.

I think we should simply get rid of these InfoBuilders altogether, for *all* of GenJSCode. Instead, we can very simply generate the `ClassInfo` from the `ClassDef`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
             "repeated parameters.")
       }
```
sjrd: The above implementation restriction should also be lifted (for free) by this implementation. So this should disappear too, and there should be tests instead.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        infoBuilder.setIsExported(true)
+        infoBuilder.setIsStatic(false)
+
+        val infoBuilderOpt = Some(infoBuilder)
+
+        val primaryCtor =
+          genMethodWithInfoBuilder(primaryCtorTree, infoBuilderOpt).get._1
+        val secondaryCtors = secondaryCtorTrees.map { tree =>
+          genMethodWithInfoBuilder(tree, infoBuilderOpt).get._1
+        }
```
sjrd: This probably means we should defer the `partition` above until later, so we can generically handle the primary constructor and the secondary constructors at this point.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val dispatch =
+          genJSConstructorExport(constructorTrees.map(_.symbol), infoBuilder)
+
+        val isConstructor =
+          secondaryCtors.map(_.name.name).toSet + primaryCtor.name.name
```
sjrd: Another example.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        // Boldly get rid of everything in infoBuilder, because it's broken
+        currentClassInfoBuilder.addMethod(
+            ir.Infos.generateMethodInfo(newConstructor))
+
+        newConstructor
+      }
+    }
+
+    private class ConstructorGraph(overrideNum: Int, method: js.MethodDef,
```
sjrd: This should probably be called `ConstructorTree`, not graph.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          Some(method.name)
+
+        case _ => None
+      }
+
+      var overrideNum = -1
+      def mkConstructorGraph(method: js.MethodDef): ConstructorGraph = {
+        val methodName = method.name.name
+        val subCtrGraphs = secondaryCtors.collect {
+          case subCtr if superCall(subCtr.body).exists(_ == methodName) =>
```
sjrd: This computes `superCall` for the same constructor several times (I think basically n times, n being the number of constructors, so O(n²) total). You could instead compute them once in a map or similar associating a constructor's name to its called constructor name.

Better, you could simply do:
```scala
val ctorToChildren = secondaryCtors.map { ctor =>
  superCall(ctor.body).getOrElse {
    abort("Secondary constructor does not call any other constructor")
  } -> ctor
}.groupBy(_._1).mapValues(_._2)
```
and then you can do
```scala
val subCtrGraphs = ctorToChildren(methodName).map(mkConstructorGraph)
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            mkConstructorGraph(subCtr)
+        }
+        overrideNum += 1
+        new ConstructorGraph(overrideNum, method, subCtrGraphs)
+      }
+
+      mkConstructorGraph(primaryCtor)
+    }
+
+
```
sjrd: Double blank line.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val preSuperCall =
+          constructorGraph.mkPreSuperCall(overloadIdent, dispatch, isConstructor)
+
+        val superCall = primaryCtor.body
+
+        val postSupperCall =
+          constructorGraph.mkPostSuperCall(overloadIdent, isConstructor)
+
+        val newBody = js.Block(overloadSelection :::
+            preSuperCall :: superCall :: postSupperCall :: Nil)
```
sjrd: `preSuperCall` must come after `superCall`. I think this your bug on ES6: the `preSuperCall` would try to initialize fields on the object but `super` has not yet been called, so `this` is not accessible. However, this should have bothered the IR checker ... Hum?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val dispatch =
+          genJSConstructorExport(constructorTrees.map(_.symbol), infoBuilder)
+
+        val isConstructor =
+          secondaryCtors.map(_.name.name).toSet + primaryCtor.name.name
+
+        val constructorGraph =
+          mkConstructorGraph(primaryCtor, secondaryCtors, isConstructor)
+
+        val overloadIdent = js.Ident("overload")
```
sjrd: This needs to be a `freshIdent`, otherwise it will collide with a local variable named `overload` in one of the constructors.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val js.MethodDef(_, dispatchName, dispatchArgs, dispatchResultType,
+            dispatchResolution) = dispatch
```
sjrd: I suggest to move this right after `val dispatch =`, without blank line.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Ah, no, that's not it, sorry. I misunderstood what `preSuperCall` and `postSuperCall` were.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val overloadSelection = mkOverloadSelection(constructorGraph,
+            overloadIdent, dispatchResolution, isConstructor)
+
+        val preSuperCall =
+          constructorGraph.mkPreSuperCall(overloadIdent, dispatch, isConstructor)
+
+        val superCall = primaryCtor.body
```
sjrd: This should be called `primaryCtorBody`, and `pre/postSuperCall` should be named `pre/postPrimaryCtorBody`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        secondaryCtors: List[js.MethodDef], isConstructor: Set[String])(
+        implicit pos: Position): ConstructorGraph = {
+      def superCall(tree: js.Tree): Option[String] = tree match {
+        case js.Block(stats) =>
+          stats.map(superCall).collectFirst {
+            case Some(name) => name
+          }
+
+        case js.ApplyStatic(cls, method, args) if isConstructor(method.name) =>
+          Some(method.name)
```
sjrd: You should be able to use `ir.Definitions.isConstructorName(method.name)` instead. You should probably also check that `args.head` is a `This()`, and that `cls` is `currentClassType`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            case Some(node) => node
+          }
+        }
       }
 
-      genMethod(primaryCtorTree).get
+      def getOverrideNum(methodName: String): Int =
+        get(methodName).fold(-1)(_.getOverrideNum)
+
+      private[ConstructorGraph] def getOverrideNum: Int = overrideNum
```
sjrd: `[ConstructorGraph]` is not necessary (several occurrences)

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        f: (T, U) => V): List[V] = {
+      if (xs.nonEmpty && ys.nonEmpty)
+        f(xs.head, ys.head) :: zipMap(xs.tail, ys.tail)(f)
+      else
+        Nil
     }
 
+    private def mkOverloadSelection(constructorGraph: ConstructorGraph,
+        overloadIdent: js.Ident, dispatchResolution: js.Tree,
+        isConstructor: Set[String])(implicit pos: Position): List[js.Tree] = {
```
sjrd: This needs a good comment explaining what it does, and the general aspect of trees it returns.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: I did not add it as an optimization, it was a way to make code a smaller and easier to read.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      if (!constructorGraph.hasSubConstructors) {
+        dispatchResolution match {
+          case js.Block(stats) =>
+            val js.ApplyStatic(_, method, _) = stats.last
+            val refs = constructorGraph.getParamRefs(method.name)
+            val rhss = stats.collect { case js.VarDef(_, _, _, rhs) => rhs }
```
sjrd: `case ...` on the next line

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        }
+      } else {
+        val overloadRef = js.VarRef(overloadIdent)(jstpe.IntType)
+        def assignOverloadNumbers(tree: js.Tree): js.Tree = tree match {
+          case js.Block(stats) =>
+            val js.ApplyStatic(_, method, _) = stats.last
+            val num = constructorGraph.getOverrideNum(method.name)
+            val refs = overloadRef :: constructorGraph.getParamRefs(method.name)
+            val rhss = js.IntLiteral(num) ::
+              stats.collect { case js.VarDef(_, _, _, rhs) => rhs }
```
sjrd: `case ...` on the next line. Maybe like this:
```scala
val rhss = js.IntLiteral(num) :: stats.collect {
  case js.VarDef(_, _, _, rhs) => rhs
}
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            val newCases = cases.map {
+              case (literals, body) => (literals, assignOverloadNumbers(body))
+            }
+            js.Match(selector, newCases, default)(tree.tpe)
+
+          case js.If(cond, thenp, elsep) =>
+            js.If(cond, assignOverloadNumbers(thenp),
+              assignOverloadNumbers(elsep))(tree.tpe)
+
+          case _ => tree
```
sjrd: `tree` on the next line, to be consistent with the other cases

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+          case js.ApplyStatic(cls, method, args) if isConstructor(method.name) =>
+            // Calls the constructor with no arguments
+            Nil
+
+          case tree => List(tree)
+        }
+      } else {
+        val overloadRef = js.VarRef(overloadIdent)(jstpe.IntType)
+        def assignOverloadNumbers(tree: js.Tree): js.Tree = tree match {
```
sjrd: This apparently also assigns parameter vars, so that name is incorrect.

sjrd: Also, add a blank line before this line (usually there should be a blank line before a local `def`, unless it comes right after an opening `{` of course.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            js.If(cond, assignOverloadNumbers(thenp),
+              assignOverloadNumbers(elsep))(tree.tpe)
+
+          case _ => tree
+        }
+
+        val newDispatchResolution = assignOverloadNumbers(dispatchResolution)
+        val allParamDefsAsVars = constructorGraph.getAllParamDefsAsVars
+        val overrideNumDef =
+          js.VarDef(overloadIdent, jstpe.IntType, true, js.IntLiteral(0))
```
sjrd: `mutable = true`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        // Boldly get rid of everything in infoBuilder, because it's broken
+        currentClassInfoBuilder.addMethod(
+            ir.Infos.generateMethodInfo(newConstructor))
+
+        newConstructor
+      }
+    }
+
+    private class ConstructorGraph(overrideNum: Int, method: js.MethodDef,
+        subConstructors: List[ConstructorGraph])(implicit pos: Position) {
```
sjrd: This `implicit pos` feels wrong, here. I think it shouldn't be there.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      def methodName: String = method.name.name
+
+      def hasSubConstructors: Boolean = subConstructors.nonEmpty
+
+      lazy val overrideNumBounds: (Int, Int) =
+        if (subConstructors.isEmpty) (overrideNum, overrideNum)
+        else (subConstructors.last.overrideNumBounds._1, overrideNum)
```
sjrd: Shouldn't this be `first`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      private[ConstructorGraph] def getOverrideNum: Int = overrideNum
+
+      def getParamRefs(methodName: String): List[js.VarRef] =
+        get(methodName).fold(List.empty[js.VarRef])(_.getParamRefs)
+
+      private[ConstructorGraph] def getParamRefs: List[js.VarRef] =
+        method.args.map(_.ref)
+
+      def mkPreSuperCall(overrideNumIdent: js.Ident, dispatch: js.MethodDef,
```
sjrd: `dispatch` is unused, AFAICT

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -682,16 +682,304 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      def mkPostSuperCall(overrideNumIdent: js.Ident,
+          isConstructor: Set[String])(implicit pos: Position): js.Tree = {
+        val overrideNumRef = js.VarRef(overrideNumIdent)(jstpe.IntType)
+        js.Block(mkSubPostCalls(overrideNumRef, isConstructor))
+      }
+
+      def getAllParamDefsAsVars: List[js.VarDef] = {
+        val localDefs = method.args.map { pDef =>
+          js.VarDef(pDef.name, pDef.ptpe, true, jstpe.zeroOf(pDef.ptpe))
```
sjrd: `mutable = true`

ci/matrix.xml
```diff
@@ -88,18 +88,18 @@
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala testSuite/test &&
     sbtretry 'set Seq(testSuite, noIrCheckTest).map(pr => scalaJSOutputMode in pr := org.scalajs.core.tools.javascript.OutputMode.ECMAScript6)' \
-        'set Seq(testSuite, noIrCheckTest).map(pr => postLinkJSEnv in pr := NodeJSEnv(executable = "/home/jenkins/apps/iojs-2.0/bin/iojs", args = Seq("--harmony-rest-parameters")).value.withSourceMap(false))' \
+        'set Seq(testSuite, noIrCheckTest).map(pr => postLinkJSEnv in pr := NodeJSEnv(executable = "/home/jenkins/apps/node/bin/node", args = Seq("--harmony-rest-parameters")).value.withSourceMap(false))' \
```
sjrd: We should probably first have a PR dedicated to only *removing* `executable =` of all the iojs stuff. Then we can rebase this PR on top of that.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Could we get rid of the InfoBuilders first? It seems to be a rather easy change.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Oh, I see. Consider:

```scala
for ((x, y) <- xs zip xy) yield f(x,y)
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: I guess we could, yes, sure.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: Considered.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,15 +412,45 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+                  paramTypesPosterasure
+                } else {
+                  val paramIndexNoCaptures = paramIndex - numCapturesFront
+                  if (paramsTypesUncurry.size <= paramIndexNoCaptures ||
+                      isRepeatedUncurry(paramIndexNoCaptures)) {
+                    assert(isRepeatedUncurry.last)
+                    repeatedToSingle(paramsTypesUncurry.last)
+                  } else {
+                    paramsTypesUncurry(paramIndexNoCaptures)
+                  }
```
nicolasstucki: Fixed this method by special casing constructors.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,335 @@ abstract class GenJSCode extends plugins.PluginComponent
...
       val hasBadParam = enteringPhase(currentRun.uncurryPhase) {
-        sym.paramss.flatten.exists(p => p.hasDefault || isRepeated(p))
+        syms.exists(_.paramss.flatten.exists(p => p.hasDefault))
       }
       if (hasBadParam) {
         reporter.error(pos,
             "Implementation restriction: the constructor of a " +
-            "Scala.js-defined JS classes cannot have default parameters nor " +
-            "repeated parameters.")
+            "Scala.js-defined JS classes cannot have default parameters.")
```
sjrd: What is it that makes default parameters not work?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,335 @@ abstract class GenJSCode extends plugins.PluginComponent
...
       }
 
-      // Implementation restriction
-      for (tree <- secondaryCtorTrees) {
-        reporter.error(tree.pos,
-            "Implementation restriction: Scala.js-defined JS classes cannot " +
-            "have secondary constructors")
+      withNewLocalNameScope {
+
```
sjrd: Spurious blank line.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: I belive that it had something to do with the way the default parameter values are defined for constructors. I will have to check.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,335 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      ir.Definitions.isConstructorName(methodName) && cls == currentClassType
+    }
+
+    private def zipMap[T, U, V](xs: List[T], ys: List[U])(
+        f: (T, U) => V): List[V] = {
+      for ((x, y) <- xs zip ys) yield f(x, y)
+    }
+
+    /** mkOverloadSelection return a list of `stats` with that starts with:
+     * 1) The definition for the local variable that will hold the overload
```
sjrd: One more space betwen `*` and `1)`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,335 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+     *    resolution number.
+     * 2) The definitions of all local variables that are used as parameters
+     *    in all the constructors.
+     * 3) The overload resolution match/if statements. For each overload the
+     *    overload number is assigned and the parameters are casted and assigned
```
sjrd: cast~~ed~~

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: The following assertion fails: https://github.com/scala-js/scala-js/pull/1982/files#diff-356ccceef3871dbfcfa92b93f89a23c1R627. In these cases the issue seems to come from trying to access the the class module with:
```scala
val trgSym = {
  if (sym.isClassConstructor) sym.owner.companionModule.moduleClass
  else sym.owner
}
```
where `sym.isClassConstructor` is `true` and `sym.owner.companionModule` is `<none>`. 

I tried a few different ways to access the module but wasn't able. Therefore I imagine that this symbol has something broken and we might need to do something more involved to have access to the default parameters if they still actually exist.

Could the prejsinterop be messing with these modules?

nicolasstucki:
Jenkins, retest this please.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: I don't know. Let's leave it like that for now.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          }
+        }
+      }
+
+      def getOverrideNum(methodName: String): Int =
+        get(methodName).fold(-1)(_.getOverrideNum)
+
+      def rootBody: js.Tree = method.body
+
+      private def getOverrideNum: Int = overrideNum
```
sjrd: Avoid overloading with one variant taking a list of parameters and the other having no `()`. This is useless anyway, as one can use `overrideNum` directly.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      def getParamRefs(methodName: String)(implicit pos: Position): List[js.VarRef] =
+        get(methodName).fold(List.empty[js.VarRef])(_.getParamRefs)
+
+      private def getParamRefs(implicit pos: Position): List[js.VarRef] =
+        method.args.map(_.ref)
```
sjrd: Same remark about overloads with and without `()`. One of them should be renamed.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            val num = constructorTree.getOverrideNum(method.name)
+            val refs = overloadRef :: constructorTree.getParamRefs(method.name)
+            val rhss = js.IntLiteral(num) :: stats.collect {
+              case js.VarDef(_, _, _, rhs) => rhs
+            }
+            val newStats = zipMap(refs, rhss)(js.Assign(_, _))
+            js.Block(newStats)
+
+          case js.ApplyStatic(_, method, js.This() :: Nil)
+              if ir.Definitions.isConstructorName(method.name) =>
```
sjrd: Why `js.This() :: Nil` in the arguments? Shouldn't it be `js.This() :: _`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        def transformDispatch(tree: js.Tree): js.Tree = tree match {
+          case js.Block(stats) =>
+            val js.ApplyStatic(_, method, _) = stats.last
+            val num = constructorTree.getOverrideNum(method.name)
+            val refs = overloadRef :: constructorTree.getParamRefs(method.name)
+            val rhss = js.IntLiteral(num) :: stats.collect {
+              case js.VarDef(_, _, _, rhs) => rhs
+            }
+            val newStats = zipMap(refs, rhss)(js.Assign(_, _))
```
sjrd: This looks fragile and obscure. The association between the order of `VarDef`s and the `getParamRefs` is unclear.

Consider
1. Separating `js.Assign(overloadRef, js.IntLiteral)` explicitly from the others
2. Using `stats.init` instead of `stats.collect` to get the list of all `js.VarDef`s. Especially if you assume there are exactly a list of `VarDef`s followed by one `js.ApplyStatic`. (currently other kinds of statements will be silently dropped)
3. Zipping `stats.init` with `getParamRefs`, and produce one `js.Assign` for each.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+     */
+    private def mkOverloadSelection(constructorTree: ConstructorTree,
+        overloadIdent: js.Ident, dispatchResolution: js.Tree,
+        isConstructor: Set[String])(implicit pos: Position): List[js.Tree]= {
+      if (!constructorTree.hasSubConstructors) {
+        dispatchResolution match {
+          case js.Block(stats) =>
+            val js.ApplyStatic(_, method, _) = stats.last
+            val refs = constructorTree.getParamRefs(method.name)
+            val rhss = stats.collect {
```
sjrd: Same here, use `stats.init`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val newBody = js.Block(overloadSelection ::: prePrimaryCtorBody ::
+            primaryCtorBody :: postPrimaryCtorBody :: Nil)
+
+        js.MethodDef(static = false, dispatchName, dispatchArgs, jstpe.NoType,
+            newBody)(dispatch.optimizerHints, dispatch.hash)
+      }
+    }
+
+    private class ConstructorTree(overrideNum: Int, method: js.MethodDef,
+        subConstructors: List[ConstructorTree]) {
```
sjrd: This class seems to conflate two things: the tree data structure; and the construction methods. Apparently, all construction methods must only be called on the top-level node of this tree. Consider separating the construction methods into another class `JSConstructorBuilder(root: ConstructorTree)`, and keep `ConstructorTree` as a pure data structure.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        }
+      }
+
+      private def mkPostSuperCallOnSndCtr(overrideNumRef: js.VarRef,
+          isConstructor: Set[String])(implicit pos: Position): js.Tree = {
+        val postSuperCall = {
+          method.body match {
+            case js.Block(stats) =>
+              stats.dropWhile {
+                case js.ApplyStatic(_, mtd, _) if isConstructor(mtd.name) => false
```
sjrd: This seems to be the only place where `isConstructor: Set[String]` is actually used. Consider replacing it with `isConstructorName`:
```scala
case js.ApplyStatic(_, mtd, js.This() :: _) if isConstructorName(mtd.name) => false
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: Because if it had parameters it would have become a `js.Block` where it's `stats.init` correspond to the arguments.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        case (lo, hi) =>
+          val lhs = js.BinaryOp(js.BinaryOp.Num_<=, js.IntLiteral(lo), numRef)
+          val rhs = js.BinaryOp(js.BinaryOp.Num_<=, numRef, js.IntLiteral(hi))
+          js.BinaryOp(js.BinaryOp.Boolean_&, lhs, rhs)
+          js.If(lhs, rhs, js.BooleanLiteral(false))(jstpe.BooleanType)
+      }
+    }
+
+    private def isCallToContructor(cls: jstpe.ClassType,
```
sjrd: This is unused.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Ah. That's quite obscure. I think it deserves a comment.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        // Section containing the overload resolution and casts of parameters
+        val overloadSelection = mkOverloadSelection(constructorTree,
+            overloadIdent, dispatchResolution, isConstructor)
+
+        /* Section containing all the code executed before the call to `this`
+         * for every secondary constructor.
+         */
+        val prePrimaryCtorBody =
+          constructorTree.mkPreSuperCall(overloadIdent, isConstructor)
```
sjrd: Consider renaming to `mkPrePrimaryCtorBody`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val primaryCtorBody = constructorTree.rootBody
+
+        /* Section containing all the code executed after the call to this for
+         * every secondary constructor.
+         */
+        val postPrimaryCtorBody =
+          constructorTree.mkPostSuperCall(overloadIdent, isConstructor)
```
sjrd: Consider renaming to `mkPostPrimaryCtorBody`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        val newBody = js.Block(overloadSelection ::: prePrimaryCtorBody ::
+            primaryCtorBody :: postPrimaryCtorBody :: Nil)
+
+        js.MethodDef(static = false, dispatchName, dispatchArgs, jstpe.NoType,
+            newBody)(dispatch.optimizerHints, dispatch.hash)
```
sjrd: Use `None` for the hash. It is not The new method def is completely different from `dispatch`. Hashes are computed later anyway.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            zipMap(refs, rhss) { (ref, rhs) =>
+              js.VarDef(ref.ident, ref.tpe, mutable = false, rhs)
+            }
+
+          case js.ApplyStatic(_, mtd, js.This() :: Nil)
+              if ir.Definitions.isConstructorName(mtd.name) =>
+            Nil
+
+          case tree =>
+            List(tree)
```
sjrd: Can this happen? It seems this is asking for silently ignoring cases where we don't have the shape we expected.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            }
+            val newDefault = transformDispatch(default)
+            js.Match(selector, newCases, newDefault)(tree.tpe)
+
+          case js.If(cond, thenp, elsep) =>
+            js.If(cond, transformDispatch(thenp),
+              transformDispatch(elsep))(tree.tpe)
+
+          case _ =>
+            tree
```
sjrd: Can this happen? It seems this is asking for silently ignoring cases where we don't have the shape we expected.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,334 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val overrideNumDef =
+          js.VarDef(overloadIdent, jstpe.IntType, mutable = true, js.IntLiteral(0))
+
+        overrideNumDef :: allParamDefsAsVars ::: newDispatchResolution :: Nil
+      }
+    }
+
+    private def mkConstructorTree(ctors: List[js.MethodDef],
+        isConstructor: Set[String])(implicit pos: Position): ConstructorTree = {
+      def superCall(tree: js.Tree): Option[String] = tree match {
```
sjrd: The name is misleading. It's not looking for a super call, but for a call to another constructor in the same class. Maybe `findCtorForwarderCall`?

Also, it should return a `String` and crash if it doesn't find anything, since this method is only called for `secondaryCtors`, which must all have a forwarder ctor call.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: Where is `isConstructorName` defined?

nicolasstucki: Never mind.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -227,6 +227,12 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
       }
     }
 
+    def genJSConstructorExport(alts: List[Symbol]): js.MethodDef = {
+      withNewLocalNameScope {
+        genExportMethod(alts.map(ExportedSymbol), "constructor")
+      }
```
sjrd: I think this method should *not* use `withNewLocalNameScope`. Its contents will be merged/inlined with the implementations of the init constructors. So they must share the same local name scope. This deserves a comment, also noting that this method is called from a place in `GenJSCode` where it is within a local name scope.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,15 +412,45 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
 
         val altsByTypeTest = groupByWithoutHashCode(alts) {
           case ExportedSymbol(alt) =>
-            // get parameter type while resolving repeated params
+            // See the comment in genPrimitiveJSArgs for a rationale about this
+
             val tpe = enteringPhase(currentRun.uncurryPhase) {
```
sjrd: This computation has grown very large. Consider extracting the computation of this `tpe` in a separate method.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,15 +412,45 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
-                repeatedToSingle(ps.last.tpe)
+
+              lazy val paramsTypesUncurry = alt.paramss.flatten.map(_.tpe)
+              lazy val isRepeatedUncurry = alt.paramss.flatten.map(isRepeated)
+
+              def paramTypesPosterasure = enteringPhase(currentRun.posterasurePhase) {
+                alt.paramss.flatten.apply(paramIndex).tpe
+              }
+
+              if (!alt.isConstructor) {
```
sjrd: `isClassConstructor` would be slightly stricter, hence safer. (`isConstructor` also consider `$init$` methods for the initializations of traits)

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,15 +412,45 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
             val tpe = enteringPhase(currentRun.uncurryPhase) {
-              val ps = alt.paramss.flatten
-              if (ps.size <= paramIndex || isRepeated(ps(paramIndex))) {
-                assert(isRepeated(ps.last))
-                repeatedToSingle(ps.last.tpe)
+
+              lazy val paramsTypesUncurry = alt.paramss.flatten.map(_.tpe)
+              lazy val isRepeatedUncurry = alt.paramss.flatten.map(isRepeated)
+
+              def paramTypesPosterasure = enteringPhase(currentRun.posterasurePhase) {
```
sjrd: Remove the `s` in `paramTypesPosterasure`. That is only 1 type.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,15 +412,45 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+                alt.paramss.flatten.apply(paramIndex).tpe
```
sjrd: Instead of `alt.paramss.flatten`, use `alt.params`. There are a few like that.

nicolasstucki: `alt.params` does not seem to exist. I will at least factor it out.

sjrd: Ah maybe it's `alt.tpe.params`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -496,7 +532,6 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     private def genApplyForSymJSSuperCall(minArgc: Int, hasRestParam: Boolean,
         sym: Symbol): js.Tree = {
       implicit val pos = sym.pos
-
```
sjrd: useless change

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
nicolasstucki: That one exists but makes the compiler crash. I looks like those parameters are not the ones in `enteringPhase(currentRun.uncurryPhase)`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -561,9 +597,18 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
 
       val result = new mutable.ListBuffer[js.VarDef]
 
-      val funTpe = enteringPhase(currentRun.posterasurePhase)(sym.tpe)
+      val paramsPosterasure =
+        enteringPhase(currentRun.posterasurePhase)(sym.tpe).params
+      val paramsNow = sym.tpe.params
+
+      /* Thake the types from the parameters at posterasurePhase for those that
```
sjrd: Typo: Take

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -561,9 +597,18 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+       * existed at that time and add all the addidional captured parameters
+       * scince then with teir current type.
```
sjrd: "since then with their current type"? what does that mean?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
sjrd: scalac mystery ... Never mind, then.

sjrd:
That's all, I think.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      if (!jsConstructorBuilder.hasSubConstructors) {
+        dispatchResolution match {
+          /* Dispatch to constructor with no arguments.
+           * Contains trivial parameterless call to the constructor.
+           */
+          case js.ApplyStatic(_, mtd, js.This() :: Nil)
+              if ir.Definitions.isConstructorName(mtd.name) =>
+            Nil
+
+          /* Dispatch to constructor with no arguments
```
sjrd: at least one argument?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        def transformDispatch(tree: js.Tree): js.Tree = tree match {
+          /* Dispatch to constructor with no arguments.
+           * Contains trivial parameterless call to the constructor.
+           */
+          case js.ApplyStatic(_, method, js.This() :: Nil)
+            if ir.Definitions.isConstructorName(method.name) =>
+            js.Assign(overloadRef,
+              js.IntLiteral(jsConstructorBuilder.getOverrideNum(method.name)))
+
+          /* Dispatch to constructor with no arguments
```
sjrd: at least one argument

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+        /* transformDispatch takes the body of the method generated by
+         * `genJSConstructorExport` and transform it recursively.
+         */
+        def transformDispatch(tree: js.Tree): js.Tree = tree match {
+          /* Dispatch to constructor with no arguments.
+           * Contains trivial parameterless call to the constructor.
+           */
+          case js.ApplyStatic(_, method, js.This() :: Nil)
+            if ir.Definitions.isConstructorName(method.name) =>
```
sjrd: 4 spaces wrt. `case`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            val newCases = cases.map {
+              case (literals, body) => (literals, transformDispatch(body))
+            }
+            val newDefault = transformDispatch(default)
+            js.Match(selector, newCases, newDefault)(tree.tpe)
+
+          // Parameter type resolution
+          case js.If(cond, thenp, elsep) =>
+            js.If(cond, transformDispatch(thenp),
+              transformDispatch(elsep))(tree.tpe)
```
sjrd: 4 spaces

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -406,20 +409,56 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
-              val ps = alt.paramss.flatten
-              if (ps.size <= paramIndex || isRepeated(ps(paramIndex))) {
-                assert(isRepeated(ps.last))
-                repeatedToSingle(ps.last.tpe)
-              } else {
-                enteringPhase(currentRun.posterasurePhase) {
-                  ps(paramIndex).tpe
+            // See the comment in genPrimitiveJSArgs for a rationale about this
+
+            def computeType() = {
```
sjrd: Could you make this a `private def` below `genExportSameArgc`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -227,6 +227,9 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
       }
     }
 
+    def genJSConstructorExport(alts: List[Symbol]): js.MethodDef =
+        genExportMethod(alts.map(ExportedSymbol), "constructor")
```
sjrd: Indentation issue.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val preSuperCall = {
+          constructorTree.method.body match {
+            case js.Block(stats) =>
+              val beforeSuperCall = stats.takeWhile {
+                case js.ApplyStatic(_, mtd, _) => !ir.Definitions.isConstructorName(mtd.name)
+                case _                         => true
+              }
+              val superCallParams = stats.collectFirst {
+                case js.ApplyStatic(_, mtd, js.This() :: args)
+                  if ir.Definitions.isConstructorName(mtd.name) =>
```
sjrd: 4 spaces

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+                  zipMap(outputParams, args) { (ref, tree) =>
+                    js.Assign(ref, tree)
+                  }
+              }.getOrElse(Nil)
+
+              beforeSuperCall ::: superCallParams
+
+            case js.ApplyStatic(_, mtd, js.This() :: args)
+              if ir.Definitions.isConstructorName(mtd.name) =>
```
sjrd: 4 spaces

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -649,29 +649,345 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+              val beforeSuperCall = stats.takeWhile {
+                case js.ApplyStatic(_, mtd, _) => !ir.Definitions.isConstructorName(mtd.name)
+                case _                         => true
+              }
+              val superCallParams = stats.collectFirst {
+                case js.ApplyStatic(_, mtd, js.This() :: args)
+                  if ir.Definitions.isConstructorName(mtd.name) =>
+                  zipMap(outputParams, args) { (ref, tree) =>
+                    js.Assign(ref, tree)
+                  }
```
sjrd: Consider
```scala
zipMap(outputParams, args)(js.Assign(_, _))
```
like you did 10 lines below.

sjrd:
That's all.

sjrd:
LGTM :)

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: See #791

PR #1994, opened 2015-10-30
Fix #1990: Deduplicate % only for Node.js lower than v2.1.0.

nicolasstucki:

PR #1995, opened 2015-10-30
Fix #1990: Do not duplicate % for 1-arg console.log on Node.js >= 2.1.0.

### LINKED ISSUE #1990: Node.js 4's console does not deduplicate % anymore -> our test breaks
Failing build log: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/66687/console

We need to remove the hack at https://github.com/scala-js/scala-js/blob/6e3255a6c0e502c317b9d14a645dcb091a3407aa/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala#L276-L296 to get back a normal behavior.

I don't know whether we can have a "smart hack" that duplicates %'s on old Nodes, and doesn't on recent ones.

sjrd:
As of io.js v2.1.0 (which later merged back into Node.js v4), console.log() does not deduplicate % signs anymore if only one argument is given to console.log().

This commit detects the version of Node.js, and will only duplicate % signs on old versions, or if there is more than one argument given to log(). This precisely restores the previous behavior of our patching of console.log() on recent versions.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
@@ -281,11 +281,24 @@ class NodeJSEnv private (
           """
           // Hack console log to duplicate double % signs
           (function() {
+            function startsWithAnyOf(s, prefixes) {
+              for (var i = 0; i < prefixes.length; i++) {
+                // ES5 does not have .startsWith() on strings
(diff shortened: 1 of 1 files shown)


sjrd:
Supersedes #1994. Review by @nicolasstucki

sjrd:
Locally tested both with Node.js 0.10.1 and Node.js 4.2.1.

nicolasstucki:
LGTM

PR #1999, opened 2015-11-04
Fix #1975: Do not expose local-to-block defs and objects.

### LINKED ISSUE #1975: Inner function in @ScalaJSDefined method doesn't get parameter
Using Scala.js 0.6.5, Scala 2.11.7: 

It seems that when writing a @ScalaJSDefined class, when a method has inner functions, undefined values are passed to those inner functions for the methods parameters:

    @ScalaJSDefined
    class ProofDocMode extends js.Object {
     
      def token(stream : js.Dynamic) : String = {
        // stream is fine here
        def peek() : String = {
          // stream will be undefined here
          ....
        }
        peek()
        ....
      }

    }

sjrd:
Defs and objects that are local-to-block appear to be public at the level of PrepJSInterop, but they are effectively private, so they must not be exposed.

Having them exposed also triggered the bug #1975, because the exported JS methods would not properly deal with values captured by local-to-block defs.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -198,6 +198,10 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
         enterOwner(OwnerKind.NonEnumScalaMod) { super.transform(modDef) }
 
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2002, opened 2015-11-04
Fix #1905: Disallow @js.native on Scala classes

### LINKED ISSUE #1905: Using @js.native on a non-js.Any should be an error
It seems I forgot to report an error in this case:
```scala
@js.native
class Foo
```
which is very disturbing.

gzm0:


gzm0:
Review by @sjrd

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -198,25 +233,22 @@ class JSInteropTest extends DirectTest with TestHelpers {
   def onlyJSRawTraits: Unit = {
 
     """
-    @js.native
     trait A
```
sjrd: lol ^^

sjrd:
LGTM

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
gzm0: I was wondering for quite a while whether this was intended and what it should mean...

PR #2004, opened 2015-11-05
Automatically generate Infos from the Trees in GenJSCode.

sjrd:
Previously, we used mutable builders, and we had to take care to
call the appropriate mutating methods when constructing immutable
Trees. This was extremely tedious, error-prone, and has been known
to be the cause of bugs in the past.

This commit completely removes all references to Infos from
GenJSCode and GenJSExports. Instead, we focus on building the
Trees. At the end of the process, GenJSFiles can automatically
generate the appropriate ClassInfo for the ClassDef that it needs
to serialize.

This considerably simplifies the code, at the cost of an
additional pass on the Trees to generate the Infos after the fact.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -19,7 +19,6 @@ import scala.annotation.tailrec
 import org.scalajs.core.ir
 import ir.{Trees => js, Types => jstpe, ClassKind, Hashers}
 import ir.Trees.OptimizerHints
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0 and/or @nicolasstucki

gzm0:
LGTM

gzm0:
This is nice. I like removing code

nicolasstucki:
LGTM

Now I will be able to remove the `InfoBuilder` hack in https://github.com/scala-js/scala-js/pull/1982#discussion-diff-43285115R730. :)

Issue #1534, opened 2015-03-14
Desugar emits the wrong vars for a dce'ed VarDef of a Record


sjrd:
`JSDesugaring`, when faced with a `VarDef(_, SomeRecordType, _, someNothingRhs)` will emit the empty `js.VarDef` as one variable, instead of expanding the variables for the record fields.

In practice, this just declares a useless variable. The missing decls are typically compensated by another branch, which is not `Nothing`, and which declares all the fields correctly.

But in theory this could lead to references of undeclared variables. Even though the code referencing it would never be *executed* (so no `TypeError`), this could be an optimization killer for the enclosing function.

PR #2007, opened 2015-11-08
Fix #1524: Implement constant folding for +[string].

### LINKED ISSUE #1524: Constant-fold +[string]
`a +[string] b` is currently not constant-folded. This could be done very easily.

tOverney:
/cc @paullepoulpe

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
@@ -2165,6 +2165,39 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  /** Translate literals to their Scala.js String representation. */
+  private def foldToStringForString_+(tree: Tree)(implicit pos : Position): Tree = tree match {
+    case FloatLiteral(value) =>
+      foldToStringForString_+(DoubleLiteral(value.toDouble))
+
+    case DoubleLiteral(value) => 
+      jsNumberToString(value).fold(tree)(StringLiteral(_))
+
+    case LongLiteral(value)    => StringLiteral(value.toString)
+    case IntLiteral(value)     => StringLiteral(value.toString)
+    case BooleanLiteral(value) => StringLiteral(value.toString)
+    case Null()                => StringLiteral("null")
+    case Undefined()           => StringLiteral("undefined")
+    case _                     => tree
+  }
+
+  /* Following the ECMAScript 6 specification */
+  private def jsNumberToString(value: Double): Option[String] = {
+    if (1.0.toString == "1") { 
+      // We are in a JS environment, so the host .toString() is the correct one.
+      Some(value.toString) 
+    } else {
+      value match {
+        case _ if value.isNaN       => Some("NaN")
+        case 0                      => Some("0")
+        case _ if value < 0         => jsNumberToString(-value).map("-" + _)
+        case _ if value.isInfinity  => Some("Infinity")
+        case _ if value.isValidInt  => Some(value.toInt.toString)
+        case _                      => None
+      }
+    }  
+  }
+
   private def foldBinaryOp(op: BinaryOp.Code, lhs: Tree, rhs: Tree)(
       implicit pos: Position): Tree = {
     import BinaryOp._
@@ -2179,6 +2212,27 @@ private[optimizer] abstract class OptimizerCore(
           case (_: Literal, _) => foldBinaryOp(op, rhs, lhs)
           case _               => default
         }
+
+      case String_+ =>
+        val lhs1 = foldToStringForString_+(lhs)
+        val rhs1 = foldToStringForString_+(rhs)
+        @inline def stringDefault = BinaryOp(String_+, lhs1, rhs1)
+        (lhs1, rhs1) match {
+          case (StringLiteral(s1), StringLiteral(s2)) => 
+            StringLiteral(s1 + s2)
+          case (_, StringLiteral("")) =>
+            foldBinaryOp(op, rhs1, lhs1)
+          case (StringLiteral(""), _) if rhs1.tpe == StringType =>
+            rhs1
+          case (_, BinaryOp(String_+, rl, rr)) => 
+            foldBinaryOp(String_+, BinaryOp(String_+, lhs1, rl), rr)
+          case (BinaryOp(String_+, ll, StringLiteral(lr)), StringLiteral(r)) =>
+            BinaryOp(String_+, ll, StringLiteral(lr + r))
+          case (BinaryOp(String_+, StringLiteral(""), lr), _) =>
+            BinaryOp(String_+, lr, rhs1)
+          case _ =>
+            stringDefault
+      }
 
       case Int_+ =>
         (lhs, rhs) match {
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
@@ -81,6 +81,217 @@ object OptimizerTest extends JasmineTest {
 
   }
 
+  describe("+[string] constant folding") {
+    it("must not break when folding two constant strings") {
+      @inline def str: String = "I am "
+      expect(str + "constant").toEqual("I am constant")
+    }
+
+    it("must not break when folding the empty string when associated with a string") {
+      @noinline def str: String = "hello"
+      expect(str + "").toEqual("hello")
+      expect("" + str).toEqual("hello")
+    }
+
+    it("must not break when folding 1.4f and a stringLit") {
+      expect(1.4f + "hello").toEqual("1.399999976158142hello")
+      expect("hello" + 1.4f).toEqual("hello1.399999976158142")
+    }
+
+    it("must not break when folding cascading +[string]") {
+      @noinline def str: String = "awesome! 10/10"
+      expect("Scala.js" + (" is " + str)).toEqual("Scala.js is awesome! 10/10")
+      expect((str + " is ") + "Scala.js").toEqual("awesome! 10/10 is Scala.js")
+    }
+
+    it("must not break when folding a chain of +[string]") {
+      @inline def b: String = "b"
+      @inline def d: String = "d"
+      @inline def f: String = "f"
+      expect("a" + b + "c" + d + "e" + f + "g").toEqual("abcdefg")
+    }
+
+    it("must not break when folding integer in double and stringLit") {
+      expect(1.0 + "hello").toEqual("1hello")
+      expect("hello" + 1.0).toEqual("hello1")
+    }
+
+    it("must not break when folding zero and stringLit") {
+      expect(0.0 + "hello").toEqual("0hello")
+      expect("hello" + 0.0).toEqual("hello0")
+      expect(-0.0 + "hello").toEqual("0hello")
+      expect("hello" + (-0.0)).toEqual("hello0")
+    }
+
+    it("must not break when folding Infinities and stringLit") {
+      expect(Double.PositiveInfinity + "hello").toEqual("Infinityhello")
+      expect("hello" + Double.PositiveInfinity).toEqual("helloInfinity")
+      expect(Double.NegativeInfinity + "hello").toEqual("-Infinityhello")
+      expect("hello" + Double.NegativeInfinity).toEqual("hello-Infinity")
+    }
+
+    it("must not break when folding NaN and stringLit") {
+      expect(Double.NaN + "hello").toEqual("NaNhello")
+      expect("hello" + Double.NaN).toEqual("helloNaN")
+    }
+
+    unless("fullopt-stage").
+    it("must not break when folding double with decimal and stringLit") {
+      expect(1.2323919403474454E21 + "hello")
+        .toEqual("1.2323919403474454e+21hello")
+      expect("hello" + 1.2323919403474454E21)
+        .toEqual("hello1.2323919403474454e+21")
+    }
+
+    unless("fullopt-stage").
+    it("must not break when folding double that JVM would print in scientific notation and stringLit") {
+      expect(123456789012345d + "hello")
+        .toEqual("123456789012345hello")
+      expect("hello" + 123456789012345d)
+        .toEqual("hello123456789012345")
+    }
+
+    unless("fullopt-stage").
+    it("must not break when folding doubles to String"){
+      @noinline def toStringNoInline(v: Double): String = v.toString
+      @inline def test(v: Double): Unit =
+        expect(v.toString).toEqual(toStringNoInline(v))
+
+      // Special cases
+      test(0.0)
+      test(-0.0)
+      test(Double.NaN)
+      test(Double.PositiveInfinity)
+      test(Double.NegativeInfinity)
+
+      // k <= n <= 21
+      test(1.0)
+      test(12.0)
+      test(123.0)
+      test(1234.0)
+      test(12345.0)
+      test(123456.0)
+      test(1234567.0)
+      test(12345678.0)
+      test(123456789.0)
+      test(1234567890.0)
+      test(12345678901.0)
+      test(123456789012.0)
+      test(1234567890123.0)
+      test(12345678901234.0)
+      test(123456789012345.0)
+      test(1234567890123456.0)
+      test(12345678901234657.0)
+      test(123456789012345678.0)
+      test(1234567890123456789.0)
+      test(12345678901234567890.0)
+      test(123456789012345678901.0)
+
+      // 0 < n <= 21
+      test(1.42)
+      test(12.42)
+      test(123.42)
+      test(1234.42)
+      test(12345.42)
+      test(123456.42)
+      test(1234567.42)
+      test(12345678.42)
+      test(123456789.42)
+      test(1234567890.42)
+      test(12345678901.42)
+      test(123456789012.42)
+      test(1234567890123.42)
+      test(12345678901234.42)
+      test(123456789012345.42)
+      test(1234567890123456.42)
+      test(12345678901234657.42)
+      test(123456789012345678.42)
+      test(1234567890123456789.42)
+      test(12345678901234567890.42)
+      test(123456789012345678901.42)
+
+      // -6 < n <= 0
+      test(0.1)
+      test(0.01)
+      test(0.001)
+      test(0.0001)
+      test(0.00001)
+      test(0.000001)
+
+      // k == 1
+      test(1e22)
+      test(2e25)
+      test(3e50)
+      test(4e100)
+      test(5e200)
+      test(6e300)
+      test(7e307)
+      test(1e-22)
+      test(2e-25)
+      test(3e-50)
+      test(4e-100)
+      test(5e-200)
+      test(6e-300)
+      test(7e-307)
+
+      // else
+      test(1.42e22)
+      test(2.42e25)
+      test(3.42e50)
+      test(4.42e100)
+      test(5.42e200)
+      test(6.42e300)
+      test(7.42e307)
+      test(1.42e-22)
+      test(2.42e-25)
+      test(3.42e-50)
+      test(4.42e-100)
+      test(5.42e-200)
+      test(6.42e-300)
+      test(7.42e-307)
+
+      // special cases when ulp > 1
+      test(18271179521433728.0)
+      test(1.15292150460684685E18)
+      test(1234567890123456770.0)
+      test(2234567890123456770.0)
+      test(4234567890123450000.0)
+      test(149170297077708820000.0)
+      test(296938164846899230000.0)
+      test(607681513323520000000.0)
+    }
+
+    it("must not break when folding long and stringLit") {
+      expect(1L + "hello").toEqual("1hello")
+      expect("hello" + 1L).toEqual("hello1")
+    }
+
+    it("must not break when folding integer and stringLit") {
+      expect(42 + "hello").toEqual("42hello")
+      expect("hello" + 42).toEqual("hello42")
+    }
+
+    it("must not break when folding boolean and stringLit") {
+      expect(false + " is not true").toEqual("false is not true")
+      expect("false is not " + true).toEqual("false is not true")
+    }
+
+    it("must not break when folding unit and stringLit") {
+      expect(() + " is undefined?").toEqual("undefined is undefined?")
+      expect("undefined is " + ()).toEqual("undefined is undefined")
+    }
+
+    it("must not break when folding null and stringLit") {
+      expect("Damien is not " + null).toEqual("Damien is not null")
+    }
+
+    it("must not break when folding char and stringLit") {
+      expect('S' + "cala.js").toEqual("Scala.js")
+      expect("Scala.j" + 's').toEqual("Scala.js")
+    }
+
+  }
+
   @inline
   class InlineClassDependentFields(val x: Int) {
     val b = x > 3


gzm0:
What prevents you from folding `classOf[T].toString`?

gzm0:
There is already an issue for this. Please prefix the commit message with "Fix #1534: ".

gzm0:
And as always I forget to write the most important thing first: **Thank you!**

gzm0:
Jenkins, ok to test

sjrd:
> What prevents you from folding classOf[T].toString?

The result needs to depend on the `runtimeClassName` function option of the semantics, which is not tracked and difficult to track. Better not to start to depend on that from the optimizer.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
 
   }
 
+  describe("String_+ constant folding") {
+    it("must fold two constant string to the correct value") {
+      expect("I am " + "constant").toEqual("I am constant")
```
sjrd: Did you check that this is not simply constant-folded by scalac? I think not, but one way to make sure is to extract one of the two sides in an `@inline def`.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+    }
+
+    it("must fold the empty string when associated with a string") {
+      val number: String = "1" + (1.4f).toString.substring(1)
```
sjrd: It seems that this line is unnecessarily complex wrt. the title of that test. Why not simply `"hello"` or any other arbitrary string? To prevent it to be known by the optimizer, you can extract it in a `@noinline def str = "hello"`.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+      expect("" + number).toEqual("1.399999976158142")
+    }
+
+    it("must fold 1.4f and a stringLit to the correct value") {
+      expect(1.4f + "hello").toEqual("1.399999976158142hello")
+      expect("hello" + 1.4f).toEqual("hello1.399999976158142")
+    }
+
+    it("must fold cascading String + correctly") {
+      def someString(a: Int) = "awesome! " + a + "/10"
```
sjrd: `@noinline` if you intend this not to be inlined (maybe one day we will automatically inline small expressions).

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+      expect("ScalaJS" + (" is " + someString(10))).toEqual("ScalaJS is awesome! 10/10")
+      expect((someString(10) + " is ") + "ScalaJS").toEqual("awesome! 10/10 is ScalaJS")
+    }
```
sjrd: What about `"ScalaJS" + (1.4f + someString(10))"`?

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+
+    it("must fold integer in double and stringLit correctly") {
+      expect(1.0 + "hello").toEqual("1hello")
+      expect("hello" + 1.0).toEqual("hello1")
+    }
```
sjrd: Also test an actual integer, like `2`.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+
+    it("must fold double with decimal and stringLit correctly") {
+      expect(1.1 + "hello").toEqual("1.1hello")
+      expect("hello" + 1.1).toEqual("hello1.1")
+    }
```
sjrd: You should also test `0.0`, `-0.0`, `Infinity`, `-Infinity` and `NaN`.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
sjrd: Ah no sorry, that case is later.

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+      expect("false is not " + true).toEqual("false is not true")
+    }
+
+    it("must fold unit and stringLit correctly") {
+      expect(() + " is undefined?").toEqual("undefined is undefined?")
+      expect("undefined is " + ()).toEqual("undefined is undefined")
+    }
+
+    it("must fold null and stringLit correctly") {
+      expect("Damien is not " + null).toEqual("Damien is not null")
```
sjrd: Are you sure?

:-p

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
 
   }
 
+  describe("String_+ constant folding") {
```
sjrd: The "public" name of `String_+` is actually `+[string]`. That is how it is printed by the IR printer. I'd rather you use that name here (and in the commit message).

test-suite/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -81,6 +81,64 @@ object OptimizerTest extends JasmineTest {
...
+    it("must fold unit and stringLit correctly") {
+      expect(() + " is undefined?").toEqual("undefined is undefined?")
+      expect("undefined is " + ()).toEqual("undefined is undefined")
+    }
+
+    it("must fold null and stringLit correctly") {
+      expect("Damien is not " + null).toEqual("Damien is not null")
+    }
+
+  }
```
sjrd: It would be good to also have a test with `Char`. Not sure it will actually constant-fold, but it would be good to guard against a mistake, somewhere.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,18 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  private def literalToString(l: Literal): String = l match {
+    case StringLiteral(str) => str
+    case FloatLiteral(flt) => flt.toDouble.toString
```
sjrd: This also needs `isValidInt` something, otherwise `1.0f.toString` will badly constant-fold.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,18 @@ private[optimizer] abstract class OptimizerCore(
...
+    case DoubleLiteral(dbl) if dbl.isValidInt => dbl.toInt.toString
+    case DoubleLiteral(dbl) => dbl.toString
+    case LongLiteral(lng) => lng.toString
+    case IntLiteral(intgr) => intgr.toString
+    case BooleanLiteral(bln) => bln.toString
+    case Null() => "null"
+    case Undefined() => "undefined"
```
sjrd: Align the arrows. See the coding style at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#pattern-matching

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,18 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  private def literalToString(l: Literal): String = l match {
+    case StringLiteral(str) => str
```
sjrd: In general, we prefer to use the name of the field [as declared in the case class](https://github.com/scala-js/scala-js/blob/v0.6.5/ir/src/main/scala/org/scalajs/core/ir/Trees.scala#L738) unless there is a good reason not to (so `value` here instead of `str`).

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2183,6 +2195,21 @@ private[optimizer] abstract class OptimizerCore(
...
         }
 
+      case String_+ =>
+        (lhs, rhs) match {
+          case (_, ClassOf(_)) | (ClassOf(_), _) => default
+
+          case (l: Literal, r: Literal) =>
+            StringLiteral(literalToString(l) + literalToString(r))
+          case (t1, t2 @ StringLiteral("")) => foldBinaryOp(op, t2, t1)
+          case (StringLiteral(""), t2) if t2.tpe == StringType => t2
```
sjrd: Try to stay consistent over the various cases, and put the rhs of the cases on the next line as well.

sjrd:
"and test it" in the commit message should not be there. If it was not tested, it wouldn't get it in the first place ;)

gzm0:
> The result needs to depend on the `runtimeClassName` function option of the semantics, which is not tracked and difficult to track. Better not to start to depend on that from the optimizer.

What do you mean by tracked? If you are talking about incrementality, changing the semantics requires creating a new optimizer anyways.

sjrd:
Of you're right, I went through that reasoning, now I remember. The real problem is that we also need to know whether the `classOf[T]` represents a class or an interface. And the optimizer does not have any knowledge query to ask whether a `T` is a class or an interface. 

We could add one, of course, and track it. But it's quite overkill, just to be able to constant fold a `classOf[T].toString`.

gzm0:
Ah, nice one. @tOverney please add a comment explaining this. Also, a (doc) comment on `literalToString` stating that it can't handle class literals would be nice.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2183,6 +2216,43 @@ private[optimizer] abstract class OptimizerCore(
           case _               => default
         }
 
+      case String_+ =>
+        (lhs, rhs) match {
+          case (l: Literal, r: Literal) =>
```
sjrd: OK now I think I see how you can factor that much more. Instead of defining `literalToString(Literal): String`, define `foldToStringForString_+(tree: Tree): Tree` which transforms `IntLiteral(v)` into `StringLiteral(v.toString)`, for example, and returns `tree` if it cannot fold the conversion of `tree` to string.

Then, here, you can just do:
```scala
val lhs1 = foldToStringForString_+(lhs)
val rhs1 = foldToStringForString_+(rhs)
(lhs1, rhs1) match {
  ...
}
```
and now you know that you only have to deal with `StringLiteral`s. Any other stuff that you find is not foldable to a `String`, and therefore you need not worry about it. Then you will have *way* less cases here.

(note: from there you can't use `default` anymore, since at least you need to use `lhs1` and `rhs1`, not `lhs` and `rhs`, so `default`s should become `BinaryOp(String_+, lhs1, rhs1)` instead.

tOverney:
@sjrd @gzm0 Thanks for the feedback, Hopefully I've updated everything.
Regarding the real number printing, @paullepoulpe [has an implementation ready here](https://github.com/SebsLittleHelpers/num-to-string-semantics/blob/master/src/main/scala/example/ScalaJSExample.scala) to print real number in the JVM to match ECMAScript specifications.

sjrd:
I think "must not fold" in test titles is not quite the correct formulation. Because it's not a requirement that the optimizer not fold those things. The real requirement is: it must not *break* the semantics of these stuff.

(Actually "must fold" is also misleading, because the test does not actually test that it folds. Again, it only tests that the optimizer does not break things.)

sjrd:
For the `null` case you can test `(null: String) + "some constant"`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,39 @@ private[optimizer] abstract class OptimizerCore(
...
+     * knowing wether classOf[T] represent a class or an interface and
+     * there are no possibilites to get that information in the optimizer.
+     */
+    case ClassOf(_) => None
+  }
+
+  /* The JVM and ECMAScript 6 specification do not match
+   * for float and double, as such no constant folding is done on them.
+   */
+  private def nuStr(value: Double): Option[String] = {
```
sjrd: Give a better name to that function, something like `jsNumberToString`.

sjrd: Also, you could add as a first case, this:
```scala
if (1.0.toString == "1") {
  // We are in a JS environment, so the host .toString() is the correct one.
  Some(value.toString)
} else ...
```
That would make the optimizer compiled to JS *more powerful* than the one compiled on the JVM ^^

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,39 @@ private[optimizer] abstract class OptimizerCore(
...
+    if(value.isValidInt) {
+      Some(if (value == 0.0) "0" // for the -0.0 case.
+           else value.toString.stripSuffix(".0"))
+    }
+    else {
```
sjrd: `} else {`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,39 @@ private[optimizer] abstract class OptimizerCore(
...
+     * there are no possibilites to get that information in the optimizer.
+     */
+    case ClassOf(_) => None
+  }
+
+  /* The JVM and ECMAScript 6 specification do not match
+   * for float and double, as such no constant folding is done on them.
+   */
+  private def nuStr(value: Double): Option[String] = {
+    if(value.isValidInt) {
```
sjrd: Space after `if`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,39 @@ private[optimizer] abstract class OptimizerCore(
...
+      Some(if (value == 0.0) "0" // for the -0.0 case.
+           else value.toString.stripSuffix(".0"))
```
sjrd: Technically, there's an indentation issue, here (5 spaces). I don't know how to fix it without making it uglier, though. That's funny.

sjrd:
```
[error]  x must not fold double with decimal and stringLit
org.scalajs.jasminetest.JasmineTestException: Expected '1.2323919403474454E21hello' to equal '1.2323919403474454e+21hello'.
```

tOverney:
>    [error]  x must not fold double with decimal and stringLit
>    org.scalajs.jasminetest.JasmineTestException: Expected '1.2323919403474454E21hello' to equal '1.2323919403474454e+21hello'.

I forgot to test with fullOpt, I'm checking what breaks everything.

sjrd:
If it's only in fullOpt it might be a big in Closure

tOverney:
It indeed only appears in fullOpt. It works in fastOpt regardless of the js vm (node or phantom) and scala version.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+  private def foldToStringForString_+(tree: Tree)(implicit pos : Position): Tree = tree match {
+    case FloatLiteral(value)   => StringLiteral(jsNumberToString(value.toDouble))
+    case DoubleLiteral(value)  => StringLiteral(jsNumberToString(value))
+    case LongLiteral(value)    => StringLiteral(value.toString)
+    case IntLiteral(value)     => StringLiteral(value.toString)
+    case BooleanLiteral(value) => StringLiteral(value.toString)
+    case Null()                => StringLiteral("null")
+    case Undefined()           => StringLiteral("undefined")
+
+    case _ => tree
```
sjrd: Remove the blank line and align the arrow with the other arrows.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+  }
+
+  /* Following the ECMAScript 6 specification */
+  private def jsNumberToString(value: Double): String = {
+    @inline def getJVMnks : (Int, Int, BigInt) = {
```
sjrd: No space before `:` (it's a thing of the French language to have spaces before colons)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+      val s = BigInt(corrected)
+      val k = corrected.length
+
+      (n, k, s)
+    }
+
+    if (1.0.toString == "1") { 
+      // We are in a JS environment, so the host .toString() is the correct one.
+      value.toString 
+    } else value match {
```
sjrd: We don't use that style in this codebase. Put `value match {` in a nested block:
```scala
} else {
  value match {
    ...
  }
}
```

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+      case n if n.isNaN => "NaN"
+      case 0 => "0"
+      case m if m < 0 => "-" + jsNumberToString(-value)
+      case i if i.isInfinity => "Infinity"
```
sjrd: Align the arrows of these 4 cases, then add a blank line below to separate the "long" case.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+      case _ => {
```
sjrd: The `{` is not necessary.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+        val (n, k, s) = getJVMnks
+        if( k <= n && n <= 21 ) {
```
sjrd: Coding style: space after `if`, no space after `(` and before `)`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+          s.toString
```
sjrd: I think this will do the wrong thing for 10^15, for example, because `s.toString` will return it in scientific notation on the JVM for `n >= 9` or something.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+        } else if (0 < n && n <= 21) {
+          val (int, part) = s.toString.splitAt(n)
+          int + "." + part
+        } else if (-6 < n && n <= 0){
```
sjrd: Missing space before `{`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+          "0." + ("0" * -n) + s.toString
+        } else if ( k == 1 ){
```
sjrd: Spaces

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+          val exp = n - 1;
```
sjrd: `;`

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
sjrd: Ah no! It's a `BigInt`, which does the right thing. Scratch that.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2183,6 +2243,32 @@ private[optimizer] abstract class OptimizerCore(
...
 
+      case String_+ =>
+        val lhs1 = foldToStringForString_+(lhs)
+        val rhs1 = foldToStringForString_+(rhs)
+        @inline def stringDefault = BinaryOp(String_+, lhs1, rhs1)
+        (lhs1, rhs1) match {
+          case (StringLiteral(s1), StringLiteral(s2)) => 
+            StringLiteral(s1 + s2)
+          case (l, r @ StringLiteral("")) =>
+            foldBinaryOp(op, r, l)
```
sjrd: Directly use `lhs1` and `rhs1` instead of `l` and `r` (and then you don't need to capture them in the pattern). Same in the other cases.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2183,6 +2243,32 @@ private[optimizer] abstract class OptimizerCore(
...
+          case (StringLiteral(""), r) if r.tpe == StringType =>
+            r
+          case (l, BinaryOp(String_+, rl, rr)) => 
+            foldBinaryOp(String_+, BinaryOp(String_+, l, rl), rr)
+          case (l @ BinaryOp(String_+, ll, lr), StringLiteral(r)) =>
+            foldBinaryOp(String_+, ll, lr)  match {
```
sjrd: This `foldBinaryOp` is redundant. By construction, all trees "smaller" than the arguments have already been folded. So this call is always going to be `identity`.

This also means the cases below can and should be "inlined" in the `case` above. For example, the first one becomes:
```scala
case (BinaryOp(String_+, llp, StringLiteral(lrp)), StringLiteral(r)) =>
  BinaryOp(String_+, llp, StringLiteral(lrp + r))
```
and the two other cases disappear, since they basically have to fall back to `default` anyway.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2183,6 +2243,32 @@ private[optimizer] abstract class OptimizerCore(
...
+              case BinaryOp(String_+, llp, StringLiteral(lrp)) =>
+                BinaryOp(String_+, llp, StringLiteral(lrp + r))
+              case `l` =>
+                BinaryOp(String_+, l, StringLiteral(r))
+              case other =>
+                foldBinaryOp(String_+, other, StringLiteral(r))
+            }
+          case _ =>
+            default
```
sjrd: This should be `stringDefault`

sjrd: Since you normalize `o + ""` into `"" + o` and you normalize `(a + (b + c))` into `((a + b) + c)` (which are both good things, IMO), you can add this case too:
```scala
case (BinaryOp(String_+, StringLiteral(""), lr), _) =>
  BinaryOp(String_+, lr, rhs1)
```
which will allow to fold `(("" + a) + b)` into `(a + b)`.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+      val k = corrected.length
+
+      (n, k, s)
+    }
+
+    if (1.0.toString == "1") { 
+      // We are in a JS environment, so the host .toString() is the correct one.
+      value.toString 
+    } else value match {
+      case n if n.isNaN => "NaN"
```
sjrd: `case _ if value.isNaN`
No need to introduce more identifiers. Same for `m` and `i` below.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
...
+    case IntLiteral(value)     => StringLiteral(value.toString)
+    case BooleanLiteral(value) => StringLiteral(value.toString)
+    case Null()                => StringLiteral("null")
+    case Undefined()           => StringLiteral("undefined")
+
+    case _ => tree
+  }
+
+  /* Following the ECMAScript 6 specification */
+  private def jsNumberToString(value: Double): String = {
```
sjrd: If you decide to keep the full algorithm, you need to thoroughly test it, by having a good test with
```scala
@noinline def toStringNoInline(v: Double): String = v.toString
@inline def test(v: Double): Unit =
  expect(v.toString).toEqual(toStringNoInline(v))

// Specials
test(0.0)
test(-0.0)
test(Double.NaN)
test(Double.PositiveInfinity)
test(Double.NegativeInfinity)

// k <= n <= 21
test(1.0)
test(1e10)
test(1e20)
...
```
etc., with a section for each case of the algorithm (that's called whitebox testing).

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,66 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  /** Translate litterals to there ScalaJS String representation. */
```
sjrd: Typos: "literals" (1 't'), "their", and "Scala.js"

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,112 @@ private[optimizer] abstract class OptimizerCore(
...
+          if (k <= n && n <= 21) {
+            s.toString + ("0" * (n - k))
+          } else if (0 < n && n <= 21) {
+            val (int, part) = s.toString.splitAt(n)
+            int + "." + part
+          } else if (-6 < n && n <= 0) {
+            "0." + ("0" * -n) + s.toString
+          } else if (k == 1) {
+            val exp = n - 1
+            s.toString + "e" + (if(exp < 0) "-" else "+") + math.abs(exp).toInt.toString
```
sjrd: Consider
```scala
s.toString + "e" + (if (exp >= 0) "+" else "") + exp.toString
```
which avoids having to use `math.abs` then converting back to an integer.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,112 @@ private[optimizer] abstract class OptimizerCore(
...
+          } else {
+            val (head, rest) = s.toString.splitAt(1)
+            val exp = n - 1
+            head + '.' + rest + "e" + (if(exp < 0) "-" else "+") + math.abs(exp).toInt.toString
```
sjrd: Same here.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2168,6 +2168,112 @@ private[optimizer] abstract class OptimizerCore(
...
+        @inline def distance(s: BigInt): BigInt =
+          (toOriginalValue(s) - exactValue).abs
+
+        val minDistance = shortestS.map(s => distance(s)).min
+
+        val closestS = shortestS.filter(s => distance(s) == minDistance)
+
+        val newS = if(closestS.length == 1) closestS.head else closestS.filter(_ % 2 == 0).head
+
+        (n, newK, newS)
```
sjrd: I don't feel comfortable with this function. It looks like a war, way too complicated with respect to what it's trying to achieve.

If it's that complicated, I do not think it is worth endangering the correctness of the optimizer for being able to constant-fold double/float toString. I think that, in this case, we had better not do anything at all, and just bail out of constant-folding a Float/Double that does not have an `Int` value. The risk/benefit ratio is too high.

gzm0: +1. Another option would be to under specify `toString` so that we can simply `toString` again.

gzm0: We specify something like this: `toString` on `Double` / `Float` produces a string that parses into the same value when using <your favorite parser> (modulo float precision).
This is a very reasonable spec and more importantly, one we can easily test for.

sjrd: Yes, but then again, is it worth underspecifying `toString` for the benefit of being to constant-fold? I don't think so either.

gzm0: Indeed :)

paullepoulpe:
Ready for merge ?

sjrd:
At the very least it needs rebasing on top of master. Right now it doesn't merge cleanly.

paullepoulpe:
Right, sorry, there !

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2165,6 +2165,37 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  /** Translate literals to their ScalaJS String representation. */
```
sjrd: Scala.js

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2165,6 +2165,37 @@ private[optimizer] abstract class OptimizerCore(
...
+  private def foldToStringForString_+(tree: Tree)(implicit pos : Position): Tree = tree match {
+    case FloatLiteral(value)   =>
+      foldToStringForString_+(DoubleLiteral(value.toDouble))
+    case DoubleLiteral(value)  => 
+      jsNumberToString(value).map(StringLiteral(_)).getOrElse(tree)
```
sjrd: Our coding style is to use `fold` instead of `map` + `getOrElse`:
```scala
jsNumberToString(value).fold(tree)(StringLiteral(_))
```

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2165,6 +2165,37 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  /** Translate literals to their ScalaJS String representation. */
+  private def foldToStringForString_+(tree: Tree)(implicit pos : Position): Tree = tree match {
+    case FloatLiteral(value)   =>
+      foldToStringForString_+(DoubleLiteral(value.toDouble))
+    case DoubleLiteral(value)  =>
```
sjrd: For cases whose body is on the next line, do not align the arrows. Also, they should have a blank line between them.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2180,6 +2211,27 @@ private[optimizer] abstract class OptimizerCore(
...
+          case (StringLiteral(""), _) if rhs1.tpe == StringType =>
+            rhs1
+          case (_, BinaryOp(String_+, rl, rr)) => 
+            foldBinaryOp(String_+, BinaryOp(String_+, lhs1, rl), rr)
+          case (BinaryOp(String_+, ll, StringLiteral(lr)), StringLiteral(r)) =>
+            BinaryOp(String_+, ll, StringLiteral(lr + r))
+          case (BinaryOp(String_+, StringLiteral(""), lr), _) =>
+            BinaryOp(String_+, lr, rhs1)
+          case _ =>
+            stringDefault
```
sjrd: Good!

sjrd:
I wish GitHub would let me comment on the Test file. But here it is instead:

`s/ScalaJS/Scala.js`

Spurious blank line at 131

Scalastyle will not like lines 292-293. You need to wrap them with:
```scala
// scalastyle:off line.size.limit
...
// scalastyle:on line.size.limit
```

sjrd:
That's all. Good job!

tOverney:
> s/ScalaJS/Scala.js
The funny thing is that running that exact regex on the file would break some tests :)
We got the message tho!

sjrd:
Scalastyle says:
```
[info] scalastyle using config /localhome/jenkins/a/workspace/scalajs-task-worker/scalastyle-config.xml
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:144:0: Whitespace at end of line
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:163:0: Whitespace at end of line
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:291:39: Whitespace at end of line
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:294:38: Whitespace at end of line
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:86:18: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:91:20: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:102:20: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:108:18: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:109:18: Public method must have explicit type
[error] /localhome/jenkins/a/workspace/scalajs-task-worker/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala:110:18: Public method must have explicit type
```

sjrd:
Otherwise it's good.
Please also `git rebase -i master` and squash all the commits into a single commit with the commit message

    Fix #1524: Implement constant folding for +[string].

sjrd:
LGTM

sjrd:
Guess what! Google Closure Compiler wrongly constant-folds `Double`s in string concatenation! It emits the JVM version :-p I'll dig if it's a known bug. (Even Rhino does it correctly.)

In any case, you should therefore add a `unless("fullopt-stage").` on the test case that is failing. You can see [here](https://github.com/scala-js/scala-js/blob/f5ef060f176722c64e06b1163334273591836c26/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/MiscInteropTest.scala#L184) an example of the coding style we use for `unless`/`when`.

sjrd:
A new entry for the hall of fame: https://github.com/google/closure-compiler/issues/1262

sjrd:
The bootstrap test fails with the long strings, at the time of deserializing the Trees (so even before we reach the optimizer). I suggest you simply remove the test with long strings. We have enough evidence that the JS VMs can cope with the long literals, since all the "normal" tests passed.

sjrd:
LGTM

sjrd:
Yay!

PR #2008, opened 2015-11-09
Implement ReentrantLock.

andreaTP:


javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,16 @@
+package java.util.concurrent.locks
+
+class ReentrantLock(private[this] val fair: Boolean) {
```
sjrd: `extends Lock with java.io.Serializable`, and you need to define `Lock`.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,16 @@
...
+  def this() = this(false)
+
+  def lock(): Unit = locked = true
+
+  def unlock(): Unit = locked = false
+
+  def isLocked(): Boolean = locked
+
+  final def isFair: Boolean = fair
+}
```
sjrd: There are quite a few missing methods: `lockInterruptibly`, `tryLock`, `newCondition` to name a few.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  def this() = this(false)
+
+  def lock(): Unit = locked = true
+
+  def unlock(): Unit = locked = false
+
+  def isLocked(): Boolean = locked
+
+  final def isFair: Boolean = fair
```
sjrd: In a Java class, methods without parameters need to have `()`. See https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#parameterless-methods

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
sjrd: `fair` could just be `fair: Boolean`, without `private[this] val`

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,16 @@
+package java.util.concurrent.locks
+
+class ReentrantLock(private[this] val fair: Boolean) {
+
+  private var locked = false
```
sjrd: Using a boolean `locked` is not correct, because of this chain of operations:
```scala
l.lock()
l.lock()
l.unlock()
expect(l.isLocked).toBeTruthy
l.unlock()
expect(l.isLocked).toBeFalsy
```

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  def this() = this(false)
+
+  def lock(): Unit = locked = true
+
+  def unlock(): Unit = locked = false
```
sjrd: `unlock()` must throw `IllegalMonitorException` if the lock is not currently locked.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.scalajs.js
+
+import org.scalajs.jasminetest.JasmineTest
+
+object ReentrantLockTest extends JasmineTest {
+
+  describe("java.util.concurrent.locks.ReentrantLock") {
+
+    it("should have all the normal operations") {
```
sjrd: Create one `it` per method. (Well you will probably need to use `lock()/unlock()` in most `it`s, but otherwise each `it` should test one method.

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
```diff
@@ -0,0 +1,30 @@
...
+      val lock = new java.util.concurrent.locks.ReentrantLock()
+      expect(lock.isFair).toBeFalsy
```
sjrd: For example `isFair()` should definitely have its own `it`.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: sure

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok for everything but can I avoid to implement ```newCondition``` ?

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
andreaTP: ok will do that later!

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: fair

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
sjrd: Yes, but then write it commented out, with a comment "Not implemented:", like this:
```scala
  // Not implemented:
  // def newCondition(params): result
```

andreaTP: only on the interface or even in implementation?

sjrd: In both. Basically everywhere you *would* have written that line, should you actually implement it.

andreaTP:
@sjrd  could you review please?

javalib/src/main/scala/java/util/concurrent/locks/Lock.scala
```diff
@@ -0,0 +1,14 @@
...
+trait Lock {
+  def lock(): Unit
+  def lockInterruptibly(): Unit
+  def tryLock(): Boolean
+  def tryLock(time: Long, unit: TimeUnit): Boolean
+  def unlock(): Unit
+  
+  //Not implemented:
+  //def newCondition(): Condition
+}
```
sjrd: Missing EOL at EOF

sjrd:
Jekins, test this please

javalib/src/main/scala/java/util/concurrent/locks/Lock.scala
andreaTP: fixed tnx

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,48 @@
...
+class ReentrantLock(fair: Boolean) extends Lock with Serializable {
+
+  private var locked = 0
+
+  def this() = this(false)
+
+  def lock(): Unit = locked += 1
+
+  def lockInterruptibly(): Unit =
+    if (Thread.currentThread().isInterrupted()) throw new InterruptedException()
```
sjrd: The spec says that in case it's interrupted, "`InterruptedException` is thrown and *the current thread's interrupted status is cleared*.". So you should use `Thread.interrupted()` instead of `Thread.currentThread().isInterrupted()`.

Btw, an if/else in statement position should always have its branches on dedicated lines.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,48 @@
...
+    else locked += 1
+
+  def tryLock(): Boolean = {
+    locked += 1
+    true
+  }
+
+  def tryLock(time: Long, unit: TimeUnit): Boolean = {
+    if (Thread.currentThread().isInterrupted()) throw new InterruptedException()
```
sjrd: Same remarks here.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,48 @@
...
+    tryLock()
```
sjrd: Be consistent over how you decide to reuse or not existing methods. In `lockInterruptibly()` you do `locked += 1` (instead of calling `lock()`), but here you delegate to `tryLock()`. Pick one and stick to it.

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  def unlock(): Unit = {
+    if (locked <= 0) throw new IllegalMonitorStateException()
+    else locked -= 1
```
sjrd: if/else in statement position

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,48 @@
...
+  def isHeldByCurrentThread(): Boolean = isLocked()
+
+  def isLocked(): Boolean = locked > 0
+
+  final def isFair(): Boolean = fair
+
+  protected def getOwner(): Thread =
+    if (isLocked) Thread.currentThread()
+    else null
+}
```
sjrd: There are still a few missing methods: hasQueuedThreads hasQueuedThread getQueueLength getQueuedThreads hasWaiters getWaitQueueLength getWaitingThreads toString

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      expect(lock.getHoldCount()).toEqual(2)
+      lock.unlock()
+      expect(lock.isLocked).toBeTruthy
+      expect(lock.getHoldCount()).toEqual(1)
+      lock.unlock()
+      expect(lock.isLocked).toBeFalsy
+      expect(lock.getHoldCount()).toEqual(0)
+    }
+
+    it("should have all the normal operations") {
```
sjrd: This test is now redundant and can be removed (what it tests is better covered by the three tests above).

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: I know but since they can become tricky and I do not need them I will like to avoid implementing them...

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      val lock2 = new ReentrantLock(true)
+      expect(lock2.isFair).toBeTruthy
+    }
+
+    it("should throw proper exceptions on invalid lock/unlock operations") {
+      val lock = new ReentrantLock()
+      expectThrows[IllegalMonitorStateException](lock.unlock)
+      Thread.currentThread().interrupt()
+      expectThrows[InterruptedException](lock.tryLock(1L, TimeUnit.SECONDS))
+      expectThrows[InterruptedException](lock.lockInterruptibly)
```
sjrd: These tests should be part of the test cases for those three individual methods (don't group all "bad" cases together in one test; it does not really make sense).

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
andreaTP: ok

sjrd:
I'll defer the rest of the review to @nicolasstucki

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
sjrd: Then add them as comments with "Not implemented"

andreaTP: even for finals with the companion?

sjrd: I don't understand what you mean by that?

Oh, btw, you can't avoid `toString()`. That one is necessary since it's an override from a superclass, so it will link anyway if it's not implemented. But `toString()` is easy, it's like `super.toString() + "[" + (if (isLocked) "Locked by " + Thread.currentThread.name else "Unlocked") + "]"` (but better written)

andreaTP:
hope this is all

andreaTP:
?

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
sjrd: What I meant by this was the coding style: if/else in statement position should have its branches on dedicated lines:
```scala
    if (locked <= 0)
      throw new IllegalMonitorStateException()
    else
      locked -= 1
```

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
```diff
@@ -0,0 +1,62 @@
...
+      expect(lock.getHoldCount()).toEqual(2)
+      lock.unlock()
+      expect(lock.isLocked).toBeTruthy
+      expect(lock.getHoldCount()).toEqual(1)
+      lock.unlock()
+      expect(lock.isLocked).toBeFalsy
+      expect(lock.getHoldCount()).toEqual(0)
+      expectThrows[IllegalMonitorStateException](lock.unlock)
+    }
+  }
```
sjrd: You're missing tests for:
* `tryLock()`
* Successful case of `tryLock(time, unit)`
* Successful case of `lockInterruptibly()`
* `isHeldByCurrentThread()`
* `getOwner()`

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
andreaTP: ok

test-suite/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/locks/ReentrantLockTest.scala
andreaTP: ok for all but ```getOwner``` is protected...

andreaTP:
review this now tnx

sjrd:
OK, can you squash all the commits together, please?

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
```diff
@@ -0,0 +1,87 @@
...
+  //protected def getQueuedThreads(): Collection[Thread]
+
+  //Not Implemented
+  //def hasWaiters(condition: Condition): Boolean
+
+  //Not Implemented
+  //def getWaitQueueLength(condition: Condition): Int
+
+  //Not Implemented
+  //protected def getWaitingThreads(condition: Condition): Collection[Thread]
```
gzm0: Any reason you don't implement these? They all seem trivial (since there is never a queued thread).

andreaTP: do not want to implement Condition...

andreaTP:
squashed

sjrd:

javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala
gzm0: Oh,  indeed. `Condition` doesn't make a lot of sense. I meant the one without `Condition` as args.

Issue #2010, opened 2015-11-12
Object.clone() is broken in fullOpt


sjrd:
In some codebases (not the test suite), `Object.clone()` is broken in fullOpt. The returned object does not have the appropriate `prototype`, and hence its `getClass` is broken, and method calls can be broken too.

Apparently, under some circumstances, Closure does not preserve the assignment of the `constructor` field on the `prototype`.

sjrd:
Fixed in #2011

PR #2011, opened 2015-11-12
Fix #1963 and #2010: Prevent GCC to remove the `constructor` prop.

### LINKED ISSUE #1963: @ScalaJSDefined classes constructor pointing to wrong ctor in base/sub classes
Its happening in `fullOptJS` mode only 

**fullOptJS** : 

screen shot 2015-10-09 at 3 36 04 am

**fastOptJS** : 

screen shot 2015-10-09 at 3 34 59 am

Steps to reproduce : 

js code
```js

<script>
    "use strict";

    function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

    var BaseMutation = (function () {
        function BaseMutation() {
            _classCallCheck(this, BaseMutation);
        }

        BaseMutation.prototype.checkThis = function checkThis() {
            console.log(this.constructor);
            console.log(this.constructor.fragments);
        };

        return BaseMutation;
    })();
</script>
```

scala code : 

```scala
@js.native
class BaseMutation extends js.Object {
  def checkThis(): Unit = js.native
}

@ScalaJSDefined
class ScalaMutation extends BaseMutation

new ScalaMutation().checkThis() // check console this is referring to different constructors in two output modes

```

more context : https://github.com/scala-js/scala-js/issues/1902#issuecomment-146702496

Currently this is a blocker for me , i can't run relay apps in production mode! , I am happy to investigate/work on this with some pointers/guidelines ;)

sjrd:
Under some circumstances, the Closure compiler would remove the
reassignment of the `constructor` property on the prototype of
classes. This removal caused the issues seen in #1963 and #2010.

This commit fixes this issue by assigning it as StringLiteral
instead of an Ident.

The tests for this fix are put in javalibExTestSuite instead of
testSuite because, if put in testSuite, they "fail to fail",
probably due to mass effects of the testSuite codebase, which
cause Closure not to exhibit the bug.


sjrd:
Review by @nicolasstucki

javalib-ex-test-suite/src/test/scala/scala/scalajs/testsuite/javalib/lang/ObjectTestEx.scala
```diff
@@ -0,0 +1,48 @@
...
+ *  codebase than testSuite to be meaningful.
+ *
+ *  If moved to testSuite, those tests "fail to fail" due to mass effects
+ *  produced by the immensity of the testSuite codebase.
+ */
+object ObjectTestEx extends JasmineTest {
+
+  describe("java.lang.Object ex") {
+
+    it("clone()") {
```
gzm0: Reference the clone bug?

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
```diff
@@ -245,7 +245,7 @@ final class ScalaJSClassEmitter private (
       js.Block(
           inheritedCtorDef,
           js.Assign(typeVar.prototype, js.New(inheritedCtorRef, Nil)),
-          genAddToPrototype(className, js.Ident("constructor"), typeVar)
+          genAddToPrototype(className, js.StringLiteral("constructor"), typeVar)
```
gzm0: We also rely on the `constructor` property as dot select in the [`scalajsenv`][1] (for `TypedArray` backing of `scala.Array`). IIUC this case could break as well. Maybe we should add `constructor` to the externs?

[1]: https://github.com/scala-js/scala-js/blob/master/tools/scalajsenv.js#L851

sjrd: It *is* already in the externs. The reference in `scalajsenv` is not renamed. It was just the reassignments on the prototype that "disappeared" for some reason.

gzm0: Oh... OK. Strange. So this is more of a workaround than an actual bugfix in Scala.js

sjrd:
Severe impact on checksizes: 7 KB non gzipped and 2 KB gzipped on fullOpt versions. But well, it's not like we have a choice.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
sjrd: Yes, I think so.

nicolasstucki:
LGTM.

gzm0:
Ouch.. :(

PR #2012, opened 2015-11-12
Faster implementation of Random.nextInt(n) for n power of 2.

sjrd:
The new implementation avoids to use Longs to do the computation.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2015, opened 2015-11-13
Fix #1899: Remove `private` flag from anonymous js.Object accessors.

### LINKED ISSUE #1899: Some members of new js.Object {...} are not visible from JavaScript
See the following change to the example project to reproduce the bug https://github.com/bbarker/scala-js-example-app/commit/23abe08739188a7cb99d18c5d183fa2bce008e10

In particular, note the difference in behavior depending on which `def getObj():...` is commented out.

nicolasstucki:


nicolasstucki:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +964,18 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       case ctor: MethodSymbol if ctor.isPrimaryConstructor => ctor
     }
 
+  /** Identify `ddef` accessor that have the shape described in issue #1899.
+   */
+  private def isAnonymousJSObjectAccessorMethod(ddef: DefDef): Boolean = {
```
gzm0: Consider passing the symbol to this method  (rather than the tree).

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -1372,3 +1378,18 @@ object ExportObjSetterNamed_= { // scalastyle:ignore
   @JSExport
   val x = 1
 }
+
+@JSExport
```
gzm0: Why does this object need to be exported? If you don't export it, you can make it local to the `it` block. (Or am I missing some of the preconditions to actually trigger the scalac bug?)

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -1372,3 +1378,18 @@ object ExportObjSetterNamed_= { // scalastyle:ignore
...
+  def getObj0(): js.Object = new js.Object{val x = "x0"}
+
+  @JSExport
+  def getObj1(): js.Object = new js.Object{val x = "x1"}.asInstanceOf[js.Object]
+
+  @JSExport
+  def getObj2(): js.Object = {
+    val o = new js.Object{val x = "x2"}
+    o
+  }
```
gzm0: Same here: Why do the methods need to be exported? 
Also: Why do you need a separate object in the first place? (What prevents you from just putting the defs in the `if` block?)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: Makes sense.

gzm0:
Hmmm... IIUC, this makes every accessor like member public, which is obviously a trade off. I assume you guys have considered a pre typer phase that annotates only the right members.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
nicolasstucki: I put it there originally to be able to intercept the the tree of the methods looked like. To figure out the exact shape of the broken accessors. Now I should be able to inline the `def`s in the `it`. I believe the the bug will still appear.

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
nicolasstucki: There is probably no need to export.

sjrd:
What about `var`s and their setters? What about `def`s? Do these not erroneously become `private` too?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -211,6 +211,20 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
         enterOwner(OwnerKind.NonEnumScalaMod) { super.transform(modDef) }
 
+      /* Identify and fix accessor that have the shape described in issue #1899.
+       *
+       * This fix is uses an over approximation of the shape of the tree due to
```
sjrd: typo: <del>is</del> uses

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -211,6 +211,20 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+       * loss of information from bug SI-9487. As such the fix introduces a
+       * difference in semantics for new js.Object {...}. Any private val,
```
sjrd: "a difference in semantics" is a big too gentle. I would write "introduces another bug"

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +966,56 @@ object ExportsTest extends JasmineTest {
       expect(jsPackage.ExportObjSetterNamed_=().x).toBe(1)
     }
 
+    it("should export js.Function values of new js.Object{...} - #1899") {
```
sjrd: The title of the test has virtually nothing to do with the test. I suggest:

    should expose public members of new js.Object {...} despite SI-9487 - #1899

test-suite/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +966,56 @@ object ExportsTest extends JasmineTest {
...
+      val dynObj = obj.asInstanceOf[js.Dynamic]
+      expect(dynObj.x1).toEqual("x1") 
+      expect(dynObj.x2).toEqual("x2")
+      expect(dynObj.x3).toEqual(js.undefined)
+
+      expect(dynObj.y1).toEqual("y1")
+      expect(dynObj.y2).toEqual("y2")
+      expect(dynObj.y3).toEqual(js.undefined)
+
+      expect(dynObj.z1).not.toEqual(js.undefined)
```
sjrd: Use `.toBeDefined` for these things instead.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +971,15 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       case ctor: MethodSymbol if ctor.isPrimaryConstructor => ctor
     }
 
+  /** Identify `ddef` accessor that have the shape described in issue #1899.
+   *  Note: This is an over approximation, read comment at call site.
+   */
+  private def isAnonymousJSObjectAccessorMethod(ddef: DefDef): Boolean = {
```
sjrd: `sym: Symbol` as parameter, as suggested in the previous version of this PR?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -211,6 +211,20 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+       * loss of information from bug SI-9487. As such the fix introduces a
+       * difference in semantics for new js.Object {...}. Any private val,
+       * var and def will become public, but private[this] can still be used to
+       * to make them private.
+       */
+      case ddef: DefDef if isAnonymousJSObjectAccessorMethod(ddef) =>
+        val fixedDDef = treeCopy.DefDef(ddef, ddef.mods &~ Flag.PRIVATE,
+            ddef.name, ddef.tparams, ddef.vparamss, ddef.tpt, ddef.rhs)
+        fixedDDef.symbol.resetFlag(Flag.PRIVATE)
+        transform(fixedDDef)
```
sjrd: I suggest moving this case as a first step *within* the `case vddef: ValOrDefDef if enclosingOwner is OwnerKind.RawJSType =>`, to avoid the self-call to `transform`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +971,15 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
     }
 
+  /** Identify `ddef` accessor that have the shape described in issue #1899.
+   *  Note: This is an over approximation, read comment at call site.
+   */
+  private def isAnonymousJSObjectAccessorMethod(ddef: DefDef): Boolean = {
+    val sym = ddef.symbol
+    val cls = sym.owner
+    sym.isPrivate && !sym.isPrivateLocal &&
+    cls.isAnonymousClass && cls.parentSymbols.contains(JSObjectClass)
```
sjrd: The last condition is wrong. The same treatment must also be applied to subclasses of other classes extending `js.Any`.

The appropriate test would be taken care of by `enclosingOwner is RawJSType`

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: I change it back to a tree while trying the `originalTree` scheme that did not work and then forgot to make it a symbol again.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +966,58 @@ object ExportsTest extends JasmineTest {
...
+      expect(dynObj.x2).toEqual("x2")
+      expect(dynObj.x3).not.toBeDefined
+
+      expect(dynObj.y1).toEqual("y1")
+      expect(dynObj.y2).toEqual("y2")
+      expect(dynObj.y3).not.toBeDefined
+      expect(dynObj.z1).toBeDefined
+
+
+
```
sjrd: Triple blank line.

sjrd:
Otherwise it's good. But I agree with @gzm0 that we should evaluate the feasibility of a fix in scalac before proceeding here, as well as the alternative solution with a pre-typer phase.

nicolasstucki:
I also agree with that pre-typer idea. An approach that I would consider is transforming any method that only contains a `new ...` to a block.

A method like this:
```scala
def foo(...): Bar = new Baz { ... }
```
would become:
```scala
def foo(...): Bar = {
  val o = new Baz { ... }
  o
}
```
A shape compiles correctly in scalac and will be easily optimized way. 

I assume that this conversion must be possible pre-typer as it does not involve any knowledge of the types.

sjrd:
I'm pretty sure such a conversion would change the way the typechecker infers some types. And it's very dangerous as it will also affect non-`js.Any` objects.

The most sensible thing to do IMO would be to define a new implementation-specific annotation `@WasPublic` or something. Then the pre-typer phase will add that annotation (with its fully qualified name `_root_.scala.scalajs.runtime.WasPublic` to all public (and protected?) members of anonymous classes. Then `PrepJSInterop` can use those annotations to re-publicize such members in classes extending `js.Any`. And it will also *remove* the annotation from everywhere. That way, there is no impact on non-`js.Any` classes.

gzm0:
@sjrd that was exactly what I had in mind. However, I'm not sure if we need to actually remove the annotations, since we can make them non-static (just like we do for stubs).
Maybe it is worth investigating which of these two things is faster (we are traversing all the trees anyways, so removal might indeed be cheaper).

PR #2016, opened 2015-11-13
Fix #1878: Make comparison between Int and Float work on Floats.

### LINKED ISSUE #1878: Comparing Int to Float should cast the Int to a Float
When comparing an `Int` and a `Float`, such as `x < y` with `x: Int` and `y: Float`, our compiler generates wrong code. It upcasts both `x` and `y` to `Double` and performs a `Double` comparison. In Scala/JVM, the `Int` is converted to a `Float` and then the comparison is on `Float`s.

This *can* cause different results, for example `(Int.MaxValue - 1) < Int.MaxValue.toFloat`. The JVM says `false`, but the current version of Scala.js says `true`.

sjrd:
Previously, such comparisons would work at the Double level,
contrary to their specification. This can cause discrepancies in
the rare cases where the imprecision of casting the Int to Float
makes it equal to the other operand.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2469,44 +2469,57 @@ abstract class GenJSCode extends plugins.PluginComponent
 
         // Binary operation
         case List(lsrc_in, rsrc_in) =>
-          def convertArg(tree: js.Tree, tpe: Type) = {
-            val kind = toTypeKind(tpe)
+          val leftKind = toTypeKind(args(0).tpe)
+          val rightKind = toTypeKind(args(1).tpe)
 
-            // If we end up with a long, target must be float or double
-            val fromLong =
+          val opType = (leftKind, rightKind) match {
+            case (DoubleKind, _) | (_, DoubleKind) => jstpe.DoubleType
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2469,44 +2469,57 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+             * so we can first eliminate the Long case by converting to Double.
+             */
+            val notLong =
               if (kind == LongKind) js.UnaryOp(js.UnaryOp.LongToDouble, tree)
               else tree
 
-            if (resultType != jstpe.FloatType) fromLong
-            else if (kind == FloatKind) fromLong
-            else js.UnaryOp(js.UnaryOp.DoubleToFloat, fromLong)
+            if (opType != jstpe.FloatType) notLong
```
gzm0: Could we make this `opType == jstpe.DoubleType` instead? Seems easier to understand.

gzm0: Argh... Of course we can't, since we might not be handling a double at all...

gzm0:
LGTM

PR #2019, opened 2015-11-15
Fix #1865: Upgrade to MiMa 0.1.8 and reenable it on 2.12.

### LINKED ISSUE #1865: MiMa on 2.12.0-M2 is so noisy that it is useless
See https://groups.google.com/forum/#!topic/scala-internals/l3-hcg18crw

I will disable MiMa checking on 2.12 artifacts until some reasonable solution is found. This issue is meant to remember that.

sjrd:
MiMa 0.1.8 fixes the issue https://github.com/typesafehub/migration-manager/issues/78 which was responsible for the high level of noise in 2.12 artifacts.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -89,9 +89,6 @@ object Build extends sbt.Build {
       } else if (newScalaBinaryVersionsInThisRelease.contains(scalaBinaryV)) {
         // New in this release, no binary compatibility to comply to
         None
-      } else if (scalaBinaryV == "2.12.0-M3") {
-        // See #1865: MiMa is much too noisy with 2.12.0-M3 to be useful
-        None
       } else {
         val thisProjectID = projectID.value
         val previousCrossVersion = thisProjectID.crossVersion match {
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @nicolasstucki.
I locally tested that the upgrade of MiMa from the v0.6.5 tag indeed solved the issue for which we had disabled it.

project/BinaryIncompatibilities.scala
```diff
@@ -41,17 +41,7 @@ object BinaryIncompatibilities {
       ProblemFilters.exclude[MissingMethodProblem](
           "scala.scalajs.runtime.RuntimeLong.SIGN_BIT_VALUE"),
       ProblemFilters.exclude[MissingMethodProblem](
-          "scala.scalajs.runtime.RuntimeLong.TWO_PWR_15_DBL"),
-
-      // Synthetic methods inside RuntimeLong, not an issue
-      ProblemFilters.exclude[MissingMethodProblem](
-          "scala.scalajs.runtime.RuntimeLong.scala$scalajs$runtime$RuntimeLong$$isZero"),
```
nicolasstucki: Why are these one not an issue anymore? Is it some fix in `sbt-mima-plugin 0.1.8` or is it something that got fixed along the way through another commit?

sjrd: It's a fix in MiMa. Now all synthetic methods are ignored by default.

sjrd:
Rebased on top of master, after a long time :p

nicolasstucki:
LGTM

PR #2024, opened 2015-11-16
Fix #2013: Make testSuite a cross-compiling project.

### LINKED ISSUE #2013: Convert testSuite into a cross-compiling project
Now that we have JUnit support for Scala.js, we should convert testSuite into a cross-compiling pair of projects for the JVM and JS. We will then be able to put tests that should run the same on the JVM and on JS as JUnit in the shared sources of the testSuite. Having the tests run on the JVM will ensure that the *tests* themselves are correct, in the sense that they model the JVM specification/behavior.

nicolasstucki:


nicolasstucki:
Review by @sjrd

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
```diff
@@ -290,6 +290,16 @@ object CrossProject extends CrossProjectExtra {
...
       crossType: CrossType): CrossProject = {
+    val crossProject = withOutScalaJSPlugin(jvmId, jsId, base, crossType)
+    crossProject.js.enablePlugins(ScalaJSPlugin)
+    crossProject
+  }
+
+  /** This is an unsafe variant of CrossProject.apply that assumes that the
+   *  ScalaJSPlugin will be added manually later on.
+   */
+  def withOutScalaJSPlugin(jvmId: String, jsId: String, base: File,
```
gzm0: without is written as a single word in english. So I would not camel case there.

project/Build.scala
```diff
@@ -1088,93 +1089,117 @@ object Build extends sbt.Build {
...
+        "org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided",
+
+      testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a")
+    ).
+    jvmSettings(
+      name := "Scala.js test suite on JVM",
+      libraryDependencies += "com.novocode" % "junit-interface" % "0.11" % "test",
+      sources in Test ++= listAdditionalSources("jvm",
+          (sourceDirectory in Test).value, scalaBinaryVersion.value,
+          scalacOptions.value, javaVersion.value)
```
gzm0: If you are basically fully re-implementing a cross project, why do you use one? It seems you are not getting anything from the implementation of `CrossProject`.

If you really want to go down the path of adding `withoutScalaJSPlugin` you need to change the other build definitions we have as well (probably in an earlier commit). However, I'd advocate for implementing cross building ourselves in the Scala.js build itself (just like we do for tools).

This also makes sense, since we probably won't ever have JVM specific sources in this project (no point of testing mainline Scala in the Scala.js build).

ci/matrix.xml
```diff
@@ -44,6 +44,7 @@
         'set scalaJSOptimizerOptions in testingExample ~= (_.withDisableOptimizer(true))' \
         'set scalaJSStage in Global := FastOptStage' \
         ++$scala testingExample/test &&
+    sbtretry ++$scala testSuiteJVM/test testSuite/clean &&
```
sjrd: testSuite**JVM**/clean (although I don't think it's necessary at all, actually)

project/Build.scala
sjrd: I agree. Let's stick to the self-cross-building in our own build.

test-suite/shared/src/test/scala/org/scalajs/testsuite/junit/JUnitAnnotationsTest.scala
```diff
@@ -1,7 +1,7 @@
 package org.scalajs.testsuite.junit
 
-import org.junit._
 import org.junit.Assert._
+import org.junit._
```
sjrd: Looks like a useless diff.

test-suite/shared/src/test/scala/org/scalajs/testsuite/junit/JUnitAssertionsTest.scala
```diff
@@ -1,9 +1,8 @@
 package org.scalajs.testsuite.junit
 
-import org.junit.Test
-
-import org.junit.Assert._
 import org.hamcrest.CoreMatchers._
+import org.junit.Assert._
+import org.junit.Test
```
sjrd: Useless diff.

sjrd:
That's all for now, though I skipped `Build.scala` for now.

project/Build.scala
```diff
@@ -1176,6 +1193,17 @@ object Build extends sbt.Build {
     compiler % "plugin", library, jUnitRuntime % "test", jasmineTestFramework % "test"
   )
 
+  lazy val testSuiteJVM = Project(
+    id = "testSuiteJVM",
+    base = file("test-suite/shared"),
+    settings = commonSettings ++ testSuiteCommonSettings(addShared = false) ++ Seq(
```
sjrd: Why `addShared = false`? Doesn't that defeat the purpose of this thing?

project/Build.scala
```diff
@@ -1176,6 +1193,17 @@ object Build extends sbt.Build {
     compiler % "plugin", library, jUnitRuntime % "test", jasmineTestFramework % "test"
   )
 
+  lazy val testSuiteJVM = Project(
+    id = "testSuiteJVM",
+    base = file("test-suite/shared"),
```
sjrd: Oh, because you put `testSuiteJVM` directly in `shared` ... That does not seem very clean. I'd rather `testSuiteJVM` was rooted in `test-suite/jvm`, and that you do add the shared files all the time.

project/Build.scala
nicolasstucki: Hm. It is a naming issue. It should probably be `isJSTest`.

project/Build.scala
```diff
@@ -1088,84 +1088,101 @@ object Build extends sbt.Build {
...
+        listScalaFilesIf(testDir / "require-jdk8", javaVersion.value >= 8) ++
+        listScalaFilesIf(testDir / "require-sam", hasSAM)
+      }
 
-            val outFile = dir / "SourceMapTest.scala"
-            IO.write(outFile, replaced.replace("0/*<testCount>*/", i.toString))
-            Seq(outFile)
-          },
+      if (addShared) {
+        val sharedTestDir = testDir / "../../../shared/src/test"
```
sjrd: IIRC that kind of path causes problems when generating Eclipse or IntelliJ paths (one of those). Prefer `testDir.getParentFile.getParentFile.getParentFile / "shared/src/test"` (not very pretty but more stable).

project/Build.scala
nicolasstucki: Makes sense. I was just imitating the structure of `CrossProject`s which ends up being confusing.

project/Build.scala
```diff
@@ -1088,84 +1088,101 @@ object Build extends sbt.Build {
...
-                  s"; if (TC.is($i)) { throw new TestException($lNo) } ;"
+    sources in Test ++= {
+      def listScalaFilesIf(testDir: File, condition: Boolean): Seq[File] =
+        if (condition) (testDir ** "*.scala").get
+        else Nil
 
-              i += 1
+      val hasSAM = {
+        scalaBinaryVersion.value != "2.10" &&
+          scalacOptions.value.contains("-Xexperimental")
```
sjrd: Indentation issue (the second line should not be more indented that the first line of the condition)

sjrd:
That's all.

test-suite/jvm/src/test/require-jdk7/org/scalajs/testsuite/javalib/lang/CharacterTestOnJDK7.scala
```diff
@@ -0,0 +1,335 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.javalib.lang
```
sjrd: How come this file ended up in `/jvm/`, now? It should be in `/shared/`.

PR #2031, opened 2015-11-18
Port testSuite/../compiler to JUnit and cross compile.

nicolasstucki:


nicolasstucki:
Review by @sjrd

nicolasstucki:
I split it into two commits to make the git history more comprehensible.

project/Build.scala
```diff
@@ -1187,7 +1187,7 @@ object Build extends sbt.Build {
     compiler % "plugin", library, jUnitRuntime % "test", jasmineTestFramework % "test"
   )
 
-  lazy val testSuiteJVM = Project(
+  lazy val testSuiteJVM: Project = Project(
```
sjrd: Looks like a completely unrelated change, here.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/BooleanJSTest.scala
```diff
@@ -7,32 +7,16 @@
...
-      testC(-1.toChar, -1)
-      testC(200.toChar, 200)
-      testC(60000.toChar, -5536)
-    }
+import scala.scalajs.js
 
+class BooleanJSTest {
+  @Test
+  def `primitive_operations_on_booleans_should_return_boolean`(): Unit = {
+    assertEquals(js.typeOf(true & false), "boolean")
```
sjrd: Lol, these tests are completely useless: they're constant-folded *by scalac*. We're not at all testing those operations are the IR level, nor, a fortiori, at the JS level -_-'

We should fix that as a follow up PR (let's not mix a change in behavior like that in the middle of a purely refactoring PR)

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/BooleanJSTest.scala
```diff
@@ -7,32 +7,16 @@
...
-
-      testC(-1.toChar, -1)
-      testC(200.toChar, 200)
-      testC(60000.toChar, -5536)
-    }
+import scala.scalajs.js
 
+class BooleanJSTest {
+  @Test
+  def `primitive_operations_on_booleans_should_return_boolean`(): Unit = {
```
sjrd: If you use backquotes anyway, you might as well keep actual spaces, no? Or does that produce really awful names at runtime, for JUnit's point of view?

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/FloatTest.scala
```diff
@@ -0,0 +1,45 @@
...
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.compiler
+
+import org.junit.Test
+import org.junit.Assert._
+
+class FloatTest {
+  @Test
+  def `Float_toInt`(): Unit = {
```
sjrd: `Float_toInt` looks redundant to me, since we're in `FloatTest`. I suggest just `toInt`. If you agree, we should follow the same convention everywhere.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala
```diff
@@ -0,0 +1,184 @@
...
+ * sound tests. We always perform the same operation, on the same operands,
+ * once in a way constant folding understands, and once in a way it doesn't.
+ * Since constant folding is performed on the JVM, we know it has the right
+ * semantics.
+ */
+
+class IntTest {
+  import IntTest._
+
+  @Test def `should_support_unary_-`(): Unit = {
```
sjrd: In general, with JUnit, I would suggest we use the method under test as the test method name. At least for tests that target one identifiable method. So here I would suggest `def unary_-()`. It would also reduce the need for backquotes.

sjrd: Although maybe for symbolic method names, we should use `unary_minus` instead?

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/LongTest.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Assert._
+
+/** Tests the compiler re-patching of native longs to
+ *  scala.scalajs.runtime.Long
+ *  see org.scalajs.testsuite.jsinterop.RuntimeLongTest
+ *  for a test of the implementation itself
+ */
+class LongTest {
+  @Test def `should_correctly_handle_literals`(): Unit = {
+    assertTrue(5L + 100L == 105L)
```
sjrd: With JUnit, we can finally use `assertEquals` even for `Long` comparison. We should do that.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/BooleanJSTest.scala
nicolasstucki: The problem is that the spaces are encoded as `$u0020`. This makes the test logs unreadable.

sjrd:
That's all.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/BooleanJSTest.scala
sjrd: Yes, that's what I thought. Let's keep underscores when we need test phrases, then.

sjrd:
LGTM

PR #2035, opened 2015-11-18
Fix #2034: Fix Assert.expectThrows return value.

### LINKED ISSUE #2034: JUnit Assert.{assert|expect}Throws never succeed.

nicolasstucki:


nicolasstucki:
Review by @sjrd

test-suite/shared/src/test/scala/org/scalajs/testsuite/junit/JUnitAssertionsTest.scala
```diff
@@ -326,4 +326,41 @@ class JUnitAssertionsTest {
     testIfAsserts(assertThat(0, instanceOf[Int](classOf[Double])), ShallNotPass)
   }
 
+  @Test
+  def testExpectThrows(): Unit = {
+    testIfAsserts(expectThrows(classOf[Exception],
+        throwingRunnable {throw new Exception}))
```
sjrd: space inside braces

test-suite/shared/src/test/scala/org/scalajs/testsuite/junit/JUnitAssertionsTest.scala
```diff
@@ -326,4 +326,41 @@ class JUnitAssertionsTest {
...
+    testIfAsserts {
+      val ex = expectThrows(classOf[Exception],
+          throwingRunnable {throw new Exception("abc")})
+      assertEquals(ex.getMessage, "abc")
+    }
+
+    testIfAsserts(expectThrows(classOf[IndexOutOfBoundsException],
+        throwingRunnable {throw new Exception}), ShallNotPass)
+    testIfAsserts(expectThrows(classOf[IndexOutOfBoundsException],
+      throwingRunnable { }), ShallNotPass)
```
sjrd: 4 spaces

sjrd:
That's all.

junit-runtime/src/main/scala/org/junit/Assert.scala
```diff
@@ -291,34 +291,37 @@ object Assert {
...
+  //          val assertionError = new AssertionError(mismatchMessage)
+  //          assertionError.initCause(actualThrown)
+  //          throw assertionError
+  //        }
+  //    }
+  //  }
+  //
+  //  trait ThrowingRunnable {
+  //    def run(): Unit
+  //  }
```
sjrd: I suggest using `/*` at the beginning and `*/` at the end, on separate lines, so that not all of these lines are affected by a diff. *Especially* since the actual fix that is in there is therefore completely invisible in the diff.

test-suite/shared/src/test/scala/org/scalajs/testsuite/junit/JUnitAssertionsTest.scala
```diff
@@ -326,4 +326,81 @@ class JUnitAssertionsTest {
...
+    testIfAsserts(assertThrows(classOf[IndexOutOfBoundsException],
+        throwingRunnable { throw new Exception }), ShallNotPass)
+    testIfAsserts(assertThrows(classOf[IndexOutOfBoundsException],
+        throwingRunnable { }), ShallNotPass)
+  }
+
+  /** Dummy implementation of Assert.assertThrows to be used until JUnit 4.13 is
+   *  released. See org.junit.Assert.scala in jUnitRuntime.
+   */
+  private def assertThrows(expectedThrowable: Class[_ <: Throwable],
```
sjrd: We should put these in `org.scalajs.testsuite.util.AssertThrows` or something.
Also, they're *not* "dummy", that's not the right word. They're real implementations since they actually do their job.

sjrd: Actually the right word is *backport*, I believe.

sjrd:
LGTM

PR #2046, opened 2015-11-22
Add implementation for java.net.URLDecoder

paullepoulpe:


sjrd:
Jenkins, ok to test

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
+package java.net
+
+import java.io.UnsupportedEncodingException
+import java.nio.{CharBuffer, ByteBuffer}
+import java.nio.charset.Charset
+
+object URLDecoder {
+  def decode(s: String, enc: String): String = {
+
+    if (!Charset.isSupported(enc)) throw new UnsupportedEncodingException(enc)
```
sjrd: This test is redundant. `Charset.forName(enc)` will throw `UnsupportedEncodingException` if given an unsupported encoding.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
+package java.net
+
+import java.io.UnsupportedEncodingException
+import java.nio.{CharBuffer, ByteBuffer}
+import java.nio.charset.Charset
+
+object URLDecoder {
+  def decode(s: String, enc: String): String = {
```
sjrd: I see no reason not to also implement the (deprecated) overload without the explicit encoding. It's a one-liner.

sjrd: Ideally, both `decode(String)` and `decode(String, String)` would call an internal `decodeImpl(String, Charset)` which takes the instance of `Charset`.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
...
+          val chunks = s.substring(i).grouped(3).takeWhile(_.charAt(0) == '%')
+          try {
+            val hexValues = chunks.map(_.substring(1))
+            val bytes = hexValues.map(Integer.parseInt(_, 16).toByte).toArray
+            val decoded = charsetDecoder.decode(ByteBuffer.wrap(bytes))
+            charBuffer.append(decoded)
+            i += 3 * bytes.length
+          } catch {
+            case _: NumberFormatException =>
+              throw new IllegalArgumentException("URLDecoder: Illegal hex characters in escape (%) pattern")
```
sjrd: You also need to handle `java.nio.charset.CharacterCodingException`, which can happen if a sequence of `%`-encoded bytes is not valid for the given charset.

sjrd: Also, about the try/catch, it is customary to put one big try/catch around the loop, instead of opening/closing a try/catch on every iteration.

javalib/src/main/scala/java/net/URLDecoder.scala
paullepoulpe: well it actually throws an UnsupportedCharsetException 
https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/nio/charset/Charset.scala#L62

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
...
+          charBuffer.append(' ')
+          i += 1
+
+        case '%' if i + 3 < len =>
+          val chunks = s.substring(i).grouped(3).takeWhile(_.charAt(0) == '%')
+          try {
+            val hexValues = chunks.map(_.substring(1))
+            val bytes = hexValues.map(Integer.parseInt(_, 16).toByte).toArray
+            val decoded = charsetDecoder.decode(ByteBuffer.wrap(bytes))
+            charBuffer.append(decoded)
```
sjrd: This is going to be prohibitively slow for such an algorithm. In the Java lib, we tend to reasonably favor performance.

A few tips:
* (Lazily) allocate one `ByteBuffer` large enough to hold all possible bytes once (hence of size `len / 3`) and reuse it.
* Fill in the `ByteBuffer` directly from the `parseInt`s instead of allocating arrays all the time.
* Don't use `substring(1)` + `parseInt`. Directly use `charAt(i)` and `Character.digit(char, 16)` twice.
* Fill in the buffer with consecutive bytes using another `while` loop.
* Use the overload of `CharsetDecoder.decode` that takes an output `CharBuffer` to directly decode into `charBuffer` rather than allocating temporary `CharBuffer`s in the process. You'll have to `flush()` and `reset()` after calling `decode()` in that case; but since by construction you can't overflow, it's not more difficult than that.

javalib/src/main/scala/java/net/URLDecoder.scala
sjrd: Ouch indeed, nice catch.

Well, then you should put the `throw` on the next line, because [ifs and if/elses in statement position should always have their branch(es) on dedicated lines](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#ifelse)

paullepoulpe: Also, this doesn't work because the doc specifies that the function should throw only if the encoding is actually needed.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
...
+    val charBuffer = CharBuffer.allocate(len)
+
+    var i = 0
+    while (i < len) {
+      s.charAt(i) match {
+        case '+' =>
+          charBuffer.append(' ')
+          i += 1
+
+        case '%' if i + 3 < len =>
```
sjrd: This test makes you handling of incorrect input inconsistent. `"%a"` will return `"%a"`, but `"%at"` will throw. The JavaDoc says we can do either, but we need to be consistent over all types of malformedness.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,46 @@
...
+          }
+
+        case c =>
+          charBuffer.append(c)
+          i += 1
+      }
+    }
+
+    charBuffer.limit(charBuffer.position)
+    charBuffer.rewind()
```
sjrd: I think you can trade those two lines for simply `charBuffer.flip()`. See http://docs.oracle.com/javase/8/docs/api/java/nio/Buffer.html#flip--

javalib/src/main/scala/java/net/URLDecoder.scala
paullepoulpe: True, good catch!

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,74 @@
...
+import java.net.URLDecoder
+import java.io.UnsupportedEncodingException
+import org.scalajs.jasminetest.JasmineTest
+
+object URLDecoderTest extends JasmineTest {
+
+  describe("URLDecoderTest"){
+
+    it("returns the empty string unchanged"){
+      expect(URLDecoder.decode("", "utf-8")).toEqual("")
```
sjrd: Declare a `private final val utf8 = "utf-8"`.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,74 @@
+package org.scalajs.testsuite.javalib.math
+
+import java.net.URLDecoder
+import java.io.UnsupportedEncodingException
+import org.scalajs.jasminetest.JasmineTest
+
+object URLDecoderTest extends JasmineTest {
```
sjrd: New tests should be written in JUnit in `test-suite/shared/`, if they're supposed to behave in the same way on JS and JVM.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,74 @@
+package org.scalajs.testsuite.javalib.math
```
sjrd: Wrong package.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,74 @@
...
+          false
+        } catch {
+          case _: UnsupportedEncodingException => true
+          case _: Throwable => false
+        }
+      }.toBeTruthy
+    }
+
+    it("works with iso-8859-1"){
+      expect(URLDecoder.decode("a%20b%20c", "iso-8859-1")).toEqual("a b c")
```
sjrd: Well you should test some ISO-8859-1-specific code units, then, i.e., characters in the range 128-255. That line would "fail to fail" if `decode()` ignored the second argument and used UTF-8 regardless.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,74 @@
...
+      expect(URLDecoder.decode("", "utf-8")).toEqual("")
+    }
+
+    it("transforms + in spaces"){
+      expect(URLDecoder.decode("a+b+c", "utf-8")).toEqual("a b c")
+    }
+
+    it("transforms encoded caracters correctly"){
+      expect(URLDecoder.decode("a%20b%20c", "utf-8")).toEqual("a b c")
+    }
```
sjrd: You should also test sequences of code units longer than 1 byte (such as `%c3%9f`). You can find more of those at https://github.com/scala-js/scala-js/blob/master/test-suite/js/src/test/scala/org/scalajs/testsuite/niocharset/UTF8Test.scala#L23-L38

And also sequences of more than 1 code point (such as `%20%20`)

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    // multi byte codepoint
+    test("a%c3%9fc", "aßc")
+
+    // consecutive characters
+    test("a%20%20c", "a  c")
+
+    // illegal codepoints
+    illegalArgument("a%b%c")
+    illegalArgument("%-1")
```
nicolasstucki: This test fails on the JVM when running on JDK6.

paullepoulpe: I fixed this, but it seems that the oracle-jdk doesn't follow the spec. 

The spec says: There are two possible ways in which this decoder could deal with illegal strings. It could either leave illegal characters alone or it could throw an IllegalArgumentException. Which approach the decoder takes is left to the implementation

Whereas, what the oracle-jdk does is that it replaces invalid characters with U+FFFD.

From that, I don't know how the tests should be written. Should we base the tests on the specs and skip the ones with illegal characters when run on the JVM. Or should we check for the current behavior URLDecoder in the oracle-jdk (as I did for now).

paullepoulpe:
I fixed the test that was failing, but it seems that the oracle-jdk doesn't follow the spec. 

The spec says: There are two possible ways in which this decoder could deal with illegal strings. It could either leave illegal characters alone or it could throw an IllegalArgumentException. Which approach the decoder takes is left to the implementation

Whereas, what the oracle-jdk does is that it replaces invalid characters with U+FFFD.

From that, I don't know how the tests should be written. Should we base the tests on the specs and skip the ones with illegal characters when run on the JVM. Or should we check for the current behavior URLDecoder in the oracle-jdk (as I did for now).

nicolasstucki:
As this test is passing on JDK8 you could at least add it as a special case in the `requires-jdk8` with a comment explaining why it does not succeed in the other versions.

paullepoulpe:
For that test yes, but what about this input?  
```
"%c3%28"
```
This input still has a behaviour that doesn't follow the spec on JDK8.

nicolasstucki:
This is unfortunate. If the implementation of the JDK has a bug we need to reproduce it and add a good comment in the implementation.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    // other charsets
+    test("a%20%A3%20c", "a £ c", enc = "iso-8859-1")
+    test("a%20b%20c", "a b c", enc = "us-ascii")
+    test("a%00%20b%00%20c", "a b c", enc = "utf-16be")
+    test("a%20%00b%20%00c", "a b c", enc = "utf-16le")
+    test("a%fe%ff%00%20b%fe%ff%00%20c", "a b c", enc = "utf-16")
+  }
+}
+
```
nicolasstucki: Files should only have one blank line at the end.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,84 @@
...
+        } catch {
+          case _: IllegalArgumentException => true
+        }
+
+      assertTrue(thrown)
+    }
+
+    @inline
+    def unsupportedEncoding(encoded: String, enc: String = utf8): Unit = {
+      val thrown =
```
nicolasstucki: Should have braces `{..}` for the body on the `val thrown = { ...` because it has more than two lines.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  @Test
+  def decodeTest(): Unit = {
+    @inline
+    def test(encoded: String, expected: String, enc: String = utf8): Unit =
+      assertEquals(URLDecoder.decode(encoded, enc), expected)
+
+    @inline
+    def illegalArgumentOrReplacement(encoded: String, enc: String = utf8): Unit = {
+      val thrown =
```
nicolasstucki: Same as below.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,76 @@
...
+  private def decodeImpl(s: String, charset: => Charset): String = {
+
+    val len = s.length
+    lazy val charsetDecoder = charset.newDecoder()
+
+    lazy val byteBuffer = ByteBuffer.allocate(len / 3)
+    val charBuffer = CharBuffer.allocate(len)
+
+    lazy val illegalHex =
+      new IllegalArgumentException("URLDecoder: Illegal hex characters in escape (%) pattern")
```
nicolasstucki: This line is too long.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,76 @@
...
+import java.nio.charset.{Charset, MalformedInputException}
+
+object URLDecoder {
+
+  @deprecated("", "")
+  def decode(s: String): String = decodeImpl(s, Charset.defaultCharset)
+
+  def decode(s: String, enc: String): String = {
+    /* An exception is thrown only if
+    the character encoding needs to be consulted */
```
nicolasstucki: For comments the format is:
```scala
/* An exception is thrown only if
 * the character encoding needs to be consulted 
 */
```

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,76 @@
+package java.net
+
+import java.io.UnsupportedEncodingException
+import java.nio.{CharBuffer, ByteBuffer}
+import java.nio.charset.{Charset, MalformedInputException}
+
+object URLDecoder {
+
+  @deprecated("", "")
```
nicolasstucki: Fill in the version.

nicolasstucki:
That's all. 
Otherwise LGTM.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  private def decodeImpl(s: String, charset: => Charset): String = {
+
+    val len = s.length
+    lazy val charsetDecoder = charset.newDecoder()
+
+    lazy val byteBuffer = ByteBuffer.allocate(len / 3)
+    val charBuffer = CharBuffer.allocate(len)
+
+    lazy val illegalHex =
```
sjrd: Consider `def throwIllegalHex() =` that throws. That's what we typically use in those cases.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+          i += 1
+
+        case '%' if i + 3 > len =>
+          throw illegalHex
+
+        case '%' =>
+          byteBuffer.clear()
+          charsetDecoder.reset()
+
+          while(i + 3 <= len && s.charAt(i) == '%'){
```
sjrd: Space before `(` and before `{`

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+            val c1 = Character.digit(s.charAt(i + 1), 16)
+            val c2 = Character.digit(s.charAt(i + 2), 16)
+
+            if(c1 < 0 || c2 < 0)
```
sjrd: Same.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+              throw illegalHex
+
+            byteBuffer.put(((c1 << 4) + c2).toByte)
+            i += 3
+          }
+
+          byteBuffer.flip()
+          val result = charsetDecoder.decode(byteBuffer, charBuffer, true)
+
```
sjrd: You need to `flush()` the `charsetDecoder` after decoding.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+          if(result.isError)
```
sjrd: Space before `(`

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+        throw new UnsupportedEncodingException(enc)
+      else
+        Charset.forName(enc)
+    }
+
+    decodeImpl(s, charset)
+  }
+
+  private def decodeImpl(s: String, charset: => Charset): String = {
+
```
sjrd: I suggest removing this blank line.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,78 @@
...
+        case '+' =>
+          charBuffer.append(' ')
+          i += 1
+
+        case '%' if i + 3 > len =>
+          throw illegalHex
+
+        case '%' =>
+          byteBuffer.clear()
+          charsetDecoder.reset()
```
sjrd: `charsetDecoder` is a `lazy val`. Accessing it every time has a cost. Consider doing `val decoder = charsetDecoder` above this line, and then use `decoder`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+      assertEquals(URLDecoder.decode(encoded, enc), expected)
+
+    @inline
+    def illegalArgumentOrReplacement(encoded: String, enc: String = utf8): Unit = {
+      val thrown = {
+        try {
+          val res = URLDecoder.decode(encoded, enc)
+
+          /* it is valid to return the Unicode replacement character (U+FFFD)
+            when encountering an invalid codepoint */
```
sjrd: Comment style:
```scala
  /* It is valid ...
   * ... invalid codepoint.
   */
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+      val thrown = {
+        try {
+          URLDecoder.decode(encoded, enc)
+          false
+        } catch {
+          case _: UnsupportedEncodingException => true
+        }
+      }
+
+      assertTrue(thrown)
```
sjrd: `assertThrows`?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+          res.contains(ReplacementChar)
+        } catch {
+          case _: IllegalArgumentException => true
+        }
+      }
+
+      assertTrue(thrown)
+    }
+
+    @inline
```
sjrd: No need for `@inline` here.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+  private final val utf8 = "utf-8"
+  private final val ReplacementChar = '\uFFFD'
+
+  @Test
+  def decodeTest(): Unit = {
+    @inline
+    def test(encoded: String, expected: String, enc: String = utf8): Unit =
+      assertEquals(URLDecoder.decode(encoded, enc), expected)
+
+    @inline
```
sjrd: No need for `@inline` here.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+import java.io.UnsupportedEncodingException
+
+class URLDecoderTest {
+
+  private final val utf8 = "utf-8"
+  private final val ReplacementChar = '\uFFFD'
+
+  @Test
+  def decodeTest(): Unit = {
+    @inline
```
sjrd: No need for `@inline` here.

javalib/src/main/scala/java/net/URLDecoder.scala
sjrd: Same with `byteBuffer`, actually.

sjrd:
That's all.

javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private def decodeImpl(s: String, charset: => Charset): String = {
+    val len = s.length
+    lazy val charsetDecoder = charset.newDecoder()
+
+    lazy val byteBuffer = ByteBuffer.allocate(len / 3)
+    val charBuffer = CharBuffer.allocate(len)
+
+    def throwIllegalHex() =
```
sjrd: This def with two lines in the body should have `{}`

sjrd:
That's all.

sjrd:

sjrd:
retest this please

PR #2058, opened 2015-11-26
Fix #1899: Fix public members of anonymous js.Any classes.

### LINKED ISSUE #1899: Some members of new js.Object {...} are not visible from JavaScript

nicolasstucki:
The fix requires an additional phase (<del>`prepwaspublicbeforetyper`</del> `jspretyper`)
between `parser` and `namer` that adds @WasPublicBeforeTyper annotations
on every public `ValDef` or `DefDef` of anonymous classes.
The `jsinterop` will additionally fix the methods that should be public
in the js.Any anonymous classes.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
@@ -1,0 +1,132 @@
+/* Scala.js compiler
+ * Copyright 2013 LAMP/EPFL
+ * @author Nicolas Stucki
+ */
+
+package org.scalajs.core.compiler
+
+import scala.tools.nsc
+import nsc._
+
+/** This `jspretyper` phase prepares a fix for issue SI-9487 in the case of
+ *  anonymous classes that extend js.Any.
+ *
+ *  During `typer`, due to a bug (SI-9487), Scalac transfroms some public method
+ *  definitions of a quite specific syntactic form of anonymous classes into
+ *  private methods. This affects both methods and field accessors. This phase
+ *  identifies any anonymous class and adds a `@WasPublicBeforeTyper` annotation
+ *  on its public methods and fields. After the `typer` in `jsinterop` the
+ *  anonymous classes are fixed if they extend js.Any using the annotations as
+ *  reference.
+ *
+ *  As an example:
+ *  {{{
+ *  class $anon extends ... {
+ *    val foo = ???
+ *    var bar = ???
+ *    def baz = ???
+ *    private val foo2 = ???
+ *    private var bar2 = ???
+ *    private def baz2 = ???
+ *  }
+ *  }}}
+ *
+ *  Would become:
+ *  {{{
+ *  class $anon extends ... {
+ *    @WasPublicBeforeTyper val foo = ???
+ *    @WasPublicBeforeTyper var bar = ???
+ *    @WasPublicBeforeTyper def baz = ???
+ *    private val foo2 = ???
+ *    private var bar2 = ???
+ *    private def baz2 = ???
+ *  }
+ *  }}}
+ *
+ *  And after `typer` (if has SI-9487) will be:
+ *  {{{
+ *  class $anon extends ... {
+ *    @WasPublicBeforeTyper private[this] var foo = ???
+ *    private <stable> <accessor> def foo = ??? // Needs fix
+ *
+ *    @WasPublicBeforeTyper private[this] var bar = ???
+ *    private <accessor> def bar = ??? // Needs fix
+ *    private <accessor> def bar_=(...) = ??? // Needs fix
+ *
+ *    @WasPublicBeforeTyper private def baz = ??? // Needs fix
+ *    ...
+ *  }
+ *  }}}
+ *
+ *  @author Nicolas Stucki
+ */
+abstract class PreTyperComponent
+    extends plugins.PluginComponent with transform.Transform with Compat210Component {
+
+  import global._
+
+  val phaseName = "jspretyper"
+
+  override protected def newTransformer(unit: CompilationUnit): Transformer =
+    new PreTyperTransformer(unit)
+
+  class PreTyperTransformer(unit: CompilationUnit) extends Transformer {
+    override def transform(tree: Tree): Tree = tree match {
+      case tree: ClassDef if needsAnnotations(tree) =>
+        val newBody = tree.impl.body.map {
+          case vdef: ValDef if needsAnnotations(vdef) =>
+            treeCopy.ValDef(vdef, withWasPublic(vdef.mods), vdef.name,
+                vdef.tpt, transform(vdef.rhs))
+
+          case ddef: DefDef if needsAnnotations(ddef) =>
+            treeCopy.DefDef(ddef, withWasPublic(ddef.mods), ddef.name,
+                ddef.tparams, ddef.vparamss, ddef.tpt, transform(ddef.rhs))
+
+          case member => transform(member)
+        }
+        val newImpl =
+          treeCopy.Template(tree.impl, tree.impl.parents, tree.impl.self, newBody)
+        treeCopy.ClassDef(tree, tree.mods, tree.name, tree.tparams, newImpl)
+
+      case tree: Template =>
+        /* Avoid filtering out members that are EmptyTree during this transform.
+         *
+         * run/macro-term-declared-in-trait is an example of code where they
+         * should not be cleaned.
+         */
+        val newBody = tree.body.map(transform)
+        treeCopy.Template(tree, tree.parents, tree.self, newBody)
+
+      case _ => super.transform(tree)
+    }
+  }
+
+  private def needsAnnotations(classDef: ClassDef): Boolean = {
+    classDef.name == nme.ANON_CLASS_NAME.toTypeName &&
+    classDef.impl.body.exists {
+      case vdef: ValDef => needsAnnotations(vdef)
+      case ddef: DefDef => needsAnnotations(ddef)
+      case _ => false
+    }
+  }
+
+  private def needsAnnotations(vdef: ValDef): Boolean =
+    vdef.mods.isPublic
+
+  private def needsAnnotations(ddef: DefDef): Boolean =
+    ddef.mods.isPublic && ddef.name != nme.CONSTRUCTOR
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(List(anonymousClassMethodWasPublicAnnotation))
+
+  private val scalajs = newTermName("scalajs")
+  private val js = newTermName("js")
+  private val wasPublicBeforeTyper = newTypeName("WasPublicBeforeTyper")
+
+  private def anonymousClassMethodWasPublicAnnotation: Tree = {
+    val runtimePackage = Select(Select(Select(Select(Ident(nme.ROOTPKG),
+        nme.scala_), scalajs), js), nme.annotation)
+    val cls = Select(runtimePackage, wasPublicBeforeTyper)
+    Apply(Select(New(cls), nme.CONSTRUCTOR), Nil)
+  }
+}
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -215,9 +215,13 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       case vddef: ValOrDefDef if vddef.symbol.isLocalToBlock =>
         super.transform(tree)
 
-      // Catch ValOrDefDef in js.Any
-      case vddef: ValOrDefDef if enclosingOwner is OwnerKind.RawJSType =>
-        transformValOrDefDefInRawJSType(vddef)
+      // Catch ValDef in js.Any
+      case vdef: ValDef if enclosingOwner is OwnerKind.RawJSType =>
+        transformValOrDefDefInRawJSType(vdef)
+
+      // Catch DefDef in js.Any
+      case ddef: DefDef if enclosingOwner is OwnerKind.RawJSType =>
+        transformValOrDefDefInRawJSType(fixPublicBeforeTyper(ddef))
 
       // Exporter generation
       case ddef: DefDef =>
@@ -942,6 +946,8 @@ abstract class PrepJSInterop extends plugins.PluginComponent
   }
 
   private lazy val ScalaEnumClass = getRequiredClass("scala.Enumeration")
+  private lazy val WasPublicBeforeTyperClass =
+    getRequiredClass("scala.scalajs.js.annotation.WasPublicBeforeTyper")
 
   /** checks if the primary constructor of the ClassDef `cldef` does not
    *  take any arguments
@@ -957,6 +963,25 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       case ctor: MethodSymbol if ctor.isPrimaryConstructor => ctor
     }
 
+  private def wasPublicBeforeTyper(sym: Symbol): Boolean =
+    sym.hasAnnotation(WasPublicBeforeTyperClass)
+
+  private def fixPublicBeforeTyper(ddef: DefDef): DefDef = {
+    // This method assumes that isJSAny(ddef.symbol.owner) is true
+    val sym = ddef.symbol
+    val needsFix = {
+      sym.isPrivate &&
+      (wasPublicBeforeTyper(sym) ||
+          (sym.isAccessor && wasPublicBeforeTyper(sym.accessed)))
+    }
+    if (needsFix) {
+      sym.resetFlag(Flag.PRIVATE)
+      treeCopy.DefDef(ddef, ddef.mods &~ Flag.PRIVATE, ddef.name, ddef.tparams,
+          ddef.vparamss, ddef.tpt, ddef.rhs)
+    } else {
+      ddef
+    }
+  }
 }
 
 object PrepJSInterop {
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
@@ -10,10 +10,12 @@ package org.scalajs.testsuite.jsinterop
 import scala.scalajs.js
 import js.annotation._
 import org.scalajs.jasminetest.{JasmineTest, TestSuiteContext}
+import org.scalajs.testsuite.utils.ExpectExceptions
 
 import scala.annotation.meta
+import scala.scalajs.js.TypeError
 
-object ExportsTest extends JasmineTest {
+object ExportsTest extends JasmineTest with ExpectExceptions {
 
   /** This package in the JS (export) namespace */
   val jsPackage = js.Dynamic.global.org.scalajs.testsuite.jsinterop
@@ -966,6 +968,141 @@ object ExportsTest extends JasmineTest {
       expect(jsPackage.ExportObjSetterNamed_=().x).toBe(1)
     }
 
+    it("should expose public members of new js.Object{...} - #1899") {
+
+      // Test that the bug is fixed for js.Any classes.
+
+      def testExposure(obj: js.Object): Unit = {
+        expect(obj).toBeDefined
+        expect(obj.hasOwnProperty("x1")).toBeTruthy
+        expect(obj.hasOwnProperty("y1")).toBeTruthy
+        expect(obj.hasOwnProperty("x2")).toBeFalsy
+        expect(obj.hasOwnProperty("y2")).toBeFalsy
+        expect(obj.hasOwnProperty("x3")).toBeFalsy
+        expect(obj.hasOwnProperty("y3")).toBeFalsy
+
+        val dynObj = obj.asInstanceOf[js.Dynamic]
+        expect(dynObj.x1).toEqual("x1")
+        expect(dynObj.x2).not.toBeDefined
+        expect(dynObj.x3).not.toBeDefined
+
+        expect(dynObj.y1).toEqual("y1")
+        expect(dynObj.y2).not.toBeDefined
+        expect(dynObj.y3).not.toBeDefined
+
+        expect(dynObj.z1()).toEqual("z1")
+        expect(dynObj.z2).not.toBeDefined
+        expect(dynObj.z2).not.toBeDefined
+        expect(dynObj.z3).not.toBeDefined
+
+        dynObj.y1 = "y1+"
+        dynObj.y2 = "y2+"
+        dynObj.y3 = "y3+"
+        expect(dynObj.y1).toEqual("y1+")
+        expect(dynObj.y2).toEqual("y2+")
+        expect(dynObj.y3).toEqual("y3+")
+        expect(dynObj.checkOriginalY1()).toEqual("y1+")
+        expect(dynObj.checkOriginalY2()).toEqual("y2")
+        expect(dynObj.checkOriginalY3()).toEqual("y3")
+      }
+
+      def getJSObj(): js.Object = new js.Object {
+        val x1 = "x1"
+        var y1 = "y1"
+        def z1() = "z1"
+        private val x2 = "x2"
+        private var y2 = "y2"
+        private def z2() = "z2"
+        private[this] val x3 = "x3"
+        private[this] var y3 = "y3"
+        private[this] def z3() = "z3"
+        def checkOriginalY1() = y1
+        def checkOriginalY2() = y2
+        def checkOriginalY3() = y3
+      }
+
+      @ScalaJSDefined
+      class JSClass extends js.Object
+
+      def getJSObj2(): js.Object = new JSClass {
+        val x1 = "x1"
+        var y1 = "y1"
+        def z1() = "z1"
+        private val x2 = "x2"
+        private var y2 = "y2"
+        private def z2() = "z2"
+        private[this] val x3 = "x3"
+        private[this] var y3 = "y3"
+        private[this] def z3() = "z3"
+        def checkOriginalY1() = y1
+        def checkOriginalY2() = y2
+        def checkOriginalY3() = y3
+      }
+
+      @ScalaJSDefined
+      abstract class JSAbstractClass extends js.Object
+
+      def getJSObj3(): js.Object = new JSAbstractClass {
+        val x1 = "x1"
+        var y1 = "y1"
+        def z1() = "z1"
+        private val x2 = "x2"
+        private var y2 = "y2"
+        private def z2() = "z2"
+        private[this] val x3 = "x3"
+        private[this] var y3 = "y3"
+        private[this] def z3() = "z3"
+        def checkOriginalY1() = y1
+        def checkOriginalY2() = y2
+        def checkOriginalY3() = y3
+      }
+
+      @ScalaJSDefined
+      abstract class JSTrait extends js.Object
+
+      def getJSObj4(): js.Object = new JSTrait {
+        val x1 = "x1"
+        var y1 = "y1"
+        def z1() = "z1"
+        private val x2 = "x2"
+        private var y2 = "y2"
+        private def z2() = "z2"
+        private[this] val x3 = "x3"
+        private[this] var y3 = "y3"
+        private[this] def z3() = "z3"
+        def checkOriginalY1() = y1
+        def checkOriginalY2() = y2
+        def checkOriginalY3() = y3
+      }
+
+      testExposure(getJSObj())
+      testExposure(getJSObj2())
+      testExposure(getJSObj3())
+      testExposure(getJSObj4())
+
+      // Test that non js.Any classes were unaffected by the fix.
+
+      def getObj(): AnyRef = new {
+        val x1 = "x1"
+        var y1 = "y1"
+        def z1() = "z1"
+        private val x2 = "x2"
+        private var y2 = "y2"
+        private def z2() = "z2"
+        private[this] val x3 = "x3"
+        private[this] var y3 = "y3"
+        private[this] def z3() = "z3"
+      }
+
+      import scala.language.reflectiveCalls
+
+      val obj2 = getObj().asInstanceOf[{ val x1: String; var y1: String; def z1(): String }]
+
+      expectThrows[Throwable](obj2.x1)
+      expectThrows[Throwable](obj2.y1)
+      expectThrows[Throwable](obj2.y1 = "y1+")
+      expectThrows[Throwable](obj2.z1)
+    }
   } // describe
 
   describe("@JSExportDescendentObjects") {
--- a/compiler/src/main/scala/org/scalajs/core/compiler/ScalaJSPlugin.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/ScalaJSPlugin.scala
@@ -25,10 +25,12 @@ class ScalaJSPlugin(val global: Global) extends NscPlugin {
   val name = "scalajs"
   val description = "Compile to JavaScript"
   val components = {
-    if (global.forScaladoc)
+    if (global.forScaladoc) {
       List[NscPluginComponent](PrepInteropComponent)
-    else
-      List[NscPluginComponent](PrepInteropComponent, GenCodeComponent)
+    } else {
+      List[NscPluginComponent](PreTyperComponentComponent, PrepInteropComponent,
+          GenCodeComponent)
+    }
   }
 
   /** Called when the JS ASTs are generated. Override for testing */
@@ -61,6 +63,12 @@ class ScalaJSPlugin(val global: Global) extends NscPlugin {
    */
   def registerModuleExports(sym: Symbol): Unit =
     PrepInteropComponent.registerModuleExports(sym)
+
+  object PreTyperComponentComponent extends {
+    val global: ScalaJSPlugin.this.global.type = ScalaJSPlugin.this.global
+    val runsAfter = List("parser")
+    override val runsBefore = List("namer")
+  } with PreTyperComponent
 
   object PrepInteropComponent extends {
     val global: ScalaJSPlugin.this.global.type = ScalaJSPlugin.this.global
--- a/library/src/main/scala/scala/scalajs/js/annotation/WasPublicBeforeTyper.scala
+++ b/library/src/main/scala/scala/scalajs/js/annotation/WasPublicBeforeTyper.scala
@@ -1,0 +1,10 @@
+package scala.scalajs.js.annotation
+
+/** IMPLEMENTATION DETAIL: Marks public members of anonymous classes before typer.
+ *
+ *  This annotation is added automatically by the compiler to all public
+ *  members of anonymous classes.
+ *
+ *  Do not use this annotation yourself.
+ */
+class WasPublicBeforeTyper extends scala.annotation.Annotation
--- a/partest-suite/src/test/resources/scala/tools/partest/scalajs/2.11.0/neg/t6446-additional.check
+++ b/partest-suite/src/test/resources/scala/tools/partest/scalajs/2.11.0/neg/t6446-additional.check
@@ -1,41 +1,31 @@
     phase name  id  description
     ----------  --  -----------
         parser   1  parse source into ASTs, perform simple desugaring
-         namer   2  resolve names, attach symbols to named trees
-packageobjects   3  load package objects
-         typer   4  the meat and potatoes: type the trees
-     jsinterop   5  
-        patmat   6  translate match expressions
-superaccessors   7  add super accessors in traits and nested classes
-    extmethods   8  add extension methods for inline classes
-       pickler   9  serialize symbol tables
-     refchecks  10  reference/override checking, translate nested objects
-       uncurry  11  uncurry, translate function values to anonymous classes
-     tailcalls  12  replace tail calls by jumps
-    specialize  13  @specialized-driven class and method specialization
- explicitouter  14  this refs to outer pointers
-       erasure  15  erase types, add interfaces for traits
-   posterasure  16  clean up erased inline classes
-      lazyvals  17  allocate bitmaps, translate lazy vals into lazified defs
-    lambdalift  18  move nested functions to top level
-  constructors  19  move field definitions into constructors
-       flatten  20  eliminate inner classes
-         mixin  21  mixin composition
-        jscode  22  
-       cleanup  23  platform-specific cleanups, generate reflective calls
-    delambdafy  24  remove lambdas
-         icode  25  generate portable intermediate code
-#partest !-optimise
-           jvm  26  generate JVM bytecode
-       ploogin  27  A sample phase that does so many things it's kind of hard...
-      terminal  28  the last phase during a compilation run
-#partest -optimise
-       inliner  26  optimization: do inlining
-inlinehandlers  27  optimization: inline exception handlers
-      closelim  28  optimization: eliminate uncalled closures
-      constopt  29  optimization: optimize null and other constants
-           dce  30  optimization: eliminate dead code
-           jvm  31  generate JVM bytecode
-       ploogin  32  A sample phase that does so many things it's kind of hard...
-      terminal  33  the last phase during a compilation run
-#partest
+    jspretyper   2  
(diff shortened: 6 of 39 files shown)


nicolasstucki:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+package org.scalajs.core.compiler
+
+import scala.tools.nsc
+import nsc._
+
+/**
+ * @author Nicolas Stucki
+ */
+abstract class PrepWasPublicBeforeTyper extends plugins.PluginComponent with transform.Transform
+    with Compat210Component {
```
sjrd: `extends` should already be on the next line.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  import global._
+
+  val phaseName = "prepwaspublicbeforetyper"
```
sjrd: This name is long. Doesn't it destroy the output of `-Xshow-phases`?

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+
+package org.scalajs.core.compiler
+
+import scala.tools.nsc
+import nsc._
+
+/**
+ * @author Nicolas Stucki
+ */
+abstract class PrepWasPublicBeforeTyper extends plugins.PluginComponent with transform.Transform
```
sjrd: This class should have a doc-comment explaining what it does. Showing the kind of snippets it transforms, and how it transforms them.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+        val newTree =
+          treeCopy.ClassDef(tree, tree.mods, tree.name, tree.tparams, newImpl)
+        super.transform(newTree)
+
+      case _ => super.transform(tree)
+    }
+  }
+
+  private def needsAnnotations(classDef: ClassDef): Boolean = {
+    classDef.name.toString == "$anon" &&
```
sjrd: Can we find this magical constant somewhere in `nme`?

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+    classDef.impl.body.exists {
+      case vdef: ValDef => needsAnnotations(vdef)
+      case ddef: DefDef => needsAnnotations(ddef)
+      case _ => false
```
sjrd: Align the arrows.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+    ddef.mods.isPublic && ddef.name != nme.CONSTRUCTOR
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(anonymousClassMethodWasPublicAnnotations)
+
+  private lazy val anonymousClassMethodWasPublicAnnotations: List[Tree] = {
+    val runtimePackage =
+      Select(Select(Ident(newTermName("scala")), newTermName("scalajs")), newTermName("runtime"))
+    val cls = Select(runtimePackage, newTypeName("WasPublicBeforeTyper"))
+    List(Apply(Select(New(cls), nme.CONSTRUCTOR), Nil))
```
sjrd: I think it's risky to reuse Trees. I suggest making this a `def` to create new Trees every time. (the `newTermName`s could be cached, though)

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  private def needsAnnotations(ddef: DefDef): Boolean =
+    ddef.mods.isPublic && ddef.name != nme.CONSTRUCTOR
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(anonymousClassMethodWasPublicAnnotations)
+
+  private lazy val anonymousClassMethodWasPublicAnnotations: List[Tree] = {
+    val runtimePackage =
+      Select(Select(Ident(newTermName("scala")), newTermName("scalajs")), newTermName("runtime"))
```
sjrd: Start from `_root_`. I think there's a `nme` for that. (There's also a `nme` for `scala`)

library/src/main/scala/scala/scalajs/runtime/WasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,5 @@
+package scala.scalajs.runtime
+
+import scala.annotation.StaticAnnotation
+
+class WasPublicBeforeTyper extends StaticAnnotation
```
sjrd: This should receive some Scaladoc telling not to use it. Something similar to what we have for `@RawJSType`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
sjrd: That doc should also explain that we do this to work around SI-something from scalac.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
gzm0: consider prefixing this with `js` like all phases we have.

library/src/main/scala/scala/scalajs/runtime/WasPublicBeforeTyper.scala
gzm0: Also, this should not be static: We don't want this to be retained in `.class` files.

gzm0: And: Is `runtime` really the right location for this? This is as little runtime as it gets...

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
gzm0: How about: `jspretyper` (and the class then: `PreTyperComponent`). I think it is a good idea to keep the class name general. If we need more pre typer transforms, we won't add another phase, will we?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+      case vddef: ValOrDefDef if needsCleanUp(vddef) =>
+        vddef.symbol.filterAnnotations(!_.matches(WasPublicBeforeTyperClass))
+        vddef
+
+      case member =>
+        member
+    }
+
+    treeCopy.Template(template, template.parents, template.self, newBody)
+  }
```
gzm0: What is preventing you to do this when you match a `DefDef` / `ValDef` in the main transformer? It seems you are reimplementing a transform here...

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
     }
 
+  private def needsFixOrCleanup(template: Template): Boolean = {
+    val body = template.body
+    if (body == null) {
+      false
+    } else {
+      body.exists { member =>
+        val sym = member.symbol
+        if (sym != null) sym.hasAnnotation(WasPublicBeforeTyperClass)
```
sjrd: Are there really `null` `body`s and `symbol`s? Wouldn't they be `EmptyTree` and `NoSymbol`, respectively?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+  }
+
+  private def fixPublicBeforeTyper(template: Template): Template = {
+    val body = template.body
+
+    def needsCleanUp(vddef: ValOrDefDef): Boolean =
+      vddef.symbol.hasAnnotation(WasPublicBeforeTyperClass)
+
+    val accessorsNeedFix = body.collect {
+      case vdef: ValDef if needsCleanUp(vdef) => vdef.name.toString.trim
```
sjrd: No operation in `nme` to do that?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+    }.toSet
+
+    def needsFix(ddef: DefDef): Boolean = {
+      lazy val isAccessorThatNeedsFix = {
+        val accessorName = ddef.name.toString
+        val fieldName = {
+          if (accessorName.endsWith("_$eq"))
+            accessorName.substring(0, accessorName.length - 4)
```
sjrd: No operation in `nme` to do that?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+    def needsCleanUp(vddef: ValOrDefDef): Boolean =
+      vddef.symbol.hasAnnotation(WasPublicBeforeTyperClass)
+
+    val accessorsNeedFix = body.collect {
+      case vdef: ValDef if needsCleanUp(vdef) => vdef.name.toString.trim
+    }.toSet
+
+    def needsFix(ddef: DefDef): Boolean = {
+      lazy val isAccessorThatNeedsFix = {
+        val accessorName = ddef.name.toString
```
sjrd: Try not to convert to string. Work with names instead.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+      sym.isPrivate && isJSAny(template.symbol.owner) &&
+      (sym.isAccessor && isAccessorThatNeedsFix) ||
+        sym.hasAnnotation(WasPublicBeforeTyperClass)
+    }
+
+    val newBody = body.map {
+      case ddef: DefDef if needsFix(ddef) =>
+        val sym = ddef.symbol
+        sym.filterAnnotations(!_.matches(WasPublicBeforeTyperClass))
+        sym.resetFlag(Flag.PRIVATE)
```
sjrd: Removing `private` should only be done for members of `js.Any` classes. We don't want to meddle with Scala anonymous classes.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,64 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+    def needsFix(ddef: DefDef): Boolean = {
+      lazy val isAccessorThatNeedsFix = {
+        val accessorName = ddef.name.toString
+        val fieldName = {
+          if (accessorName.endsWith("_$eq"))
+            accessorName.substring(0, accessorName.length - 4)
+          else
+            accessorName
+        }
+        accessorsNeedFix(fieldName)
```
sjrd: Instead of collecting field names, can't we collect the field symbols, and simply use `ddef.symbol.accessed`?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: However, the annotation should be removed from all classes (`js.Any` and non-`js.Any`).

sjrd:
That's all, I think.

Did you measure how much longer does it take to compile the Scala lib after this change?

gzm0:
Oh, another thing: You will need to adapt the checkfiles of the partests that list the compiler phases. There is a list with comments at the end of `Whitelisted.txt`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepWasPublicBeforeTyper.scala
nicolasstucki: I like `jspretyper`, I will use it.

library/src/main/scala/scala/scalajs/runtime/WasPublicBeforeTyper.scala
nicolasstucki: It is just because we did not have any other good locations to put it. It probably should be in its own library that is only used by the plugin, but this is a bit of overkill. Maybe in the future we could take out stuff that is only used by the plugin.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: They should, but I was getting both :(

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: The `if needsFix(ddef)` is checking that `isJSAny(template.symbol.owner)`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: Yes, that works :)

nicolasstucki:
@gzm0 how do I run partests that have `#partest`?
I assume that I'm missing a flag or simply running it from the wrong place.

For example `neg/t7494-no-options` which I would update to: 
```
error: Error: ploogin takes no options
    phase name  id  description
    ----------  --  -----------
        parser   1  parse source into ASTs, perform simple desugaring
    jspretyper   2
         namer   3  resolve names, attach symbols to named trees
packageobjects   4  load package objects
         typer   5  the meat and potatoes: type the trees
     jsinterop   6
        patmat   7  translate match expressions
superaccessors   8  add super accessors in traits and nested classes
    extmethods   9  add extension methods for inline classes
       pickler  10  serialize symbol tables
     refchecks  11  reference/override checking, translate nested objects
       uncurry  12  uncurry, translate function values to anonymous classes
     tailcalls  13  replace tail calls by jumps
    specialize  14  @specialized-driven class and method specialization
 explicitouter  15  this refs to outer pointers
       erasure  16  erase types, add interfaces for traits
   posterasure  17  clean up erased inline classes
      lazyvals  18  allocate bitmaps, translate lazy vals into lazified defs
    lambdalift  19  move nested functions to top level
  constructors  20  move field definitions into constructors
       flatten  21  eliminate inner classes
         mixin  22  mixin composition
        jscode  23
       cleanup  24  platform-specific cleanups, generate reflective calls
    delambdafy  25  remove lambdas
         icode  26  generate portable intermediate code
#partest !-optimise
           jvm  27  generate JVM bytecode
       ploogin  28  A sample phase that does so many things it's kind of hard...
      terminal  29  the last phase during a compilation run
#partest -optimise
       inliner  27  optimization: do inlining
inlinehandlers  28  optimization: inline exception handlers
      closelim  29  optimization: eliminate uncalled closures
      constopt  30  optimization: optimize null and other constants
           dce  31  optimization: eliminate dead code
           jvm  32  generate JVM bytecode
       ploogin  33  A sample phase that does so many things it's kind of hard...
      terminal  34  the last phase during a compilation run
#partest
```

Fails with:
```
> partestSuite/testOnly -- --showDiff neg/t7494-no-options
...
!! 1 - neg/t7494-no-options                      [output differs]
% diff /Users/nastucki/GitHub/scala-js/partest/fetchedSources/2.11.7/test/files/neg/t7494-no-options-neg.log /Users/nastucki/GitHub/scala-js/partest-suite/target/scala-2.11/test-classes/scala/tools/partest/scalajs/2.11.7/neg/t7494-no-options.check
@@ -27,6 +27,17 @@ superaccessors   8  add super accessors in traits and nested classes
        cleanup  24  platform-specific cleanups, generate reflective calls
     delambdafy  25  remove lambdas
          icode  26  generate portable intermediate code
+#partest !-optimise
            jvm  27  generate JVM bytecode
        ploogin  28  A sample phase that does so many things it's kind of hard...
       terminal  29  the last phase during a compilation run
+#partest -optimise
+       inliner  27  optimization: do inlining
+inlinehandlers  28  optimization: inline exception handlers
+      closelim  29  optimization: eliminate uncalled closures
+      constopt  30  optimization: optimize null and other constants
+           dce  31  optimization: eliminate dead code
+           jvm  32  generate JVM bytecode
+       ploogin  33  A sample phase that does so many things it's kind of hard...
+      terminal  34  the last phase during a compilation run
+#partest
...
```

gzm0:
That the diff contains the "pragmas" is a known bug. Are you sure your whitespace is correct? You might not need a newline at the end of the file (or have two of them).

nicolasstucki:
I only inserted one line (with the new phase), that one has no problems in the diff as you can see. I was quite careful to not add spaces while editing the phase numbers. If I add a new line at the end it will be part of the diff, therefore it looks like the end of the file is fine. I also tried to remove the lines that should be removed by the "pragmas" but then it got weirder output.

With:
```
error: Error: ploogin takes no options
    phase name  id  description
    ----------  --  -----------
        parser   1  parse source into ASTs, perform simple desugaring
    jspretyper   2
         namer   3  resolve names, attach symbols to named trees
packageobjects   4  load package objects
         typer   5  the meat and potatoes: type the trees
     jsinterop   6
        patmat   7  translate match expressions
superaccessors   8  add super accessors in traits and nested classes
    extmethods   9  add extension methods for inline classes
       pickler  10  serialize symbol tables
     refchecks  11  reference/override checking, translate nested objects
       uncurry  12  uncurry, translate function values to anonymous classes
     tailcalls  13  replace tail calls by jumps
    specialize  14  @specialized-driven class and method specialization
 explicitouter  15  this refs to outer pointers
       erasure  16  erase types, add interfaces for traits
   posterasure  17  clean up erased inline classes
      lazyvals  18  allocate bitmaps, translate lazy vals into lazified defs
    lambdalift  19  move nested functions to top level
  constructors  20  move field definitions into constructors
       flatten  21  eliminate inner classes
         mixin  22  mixin composition
        jscode  23
       cleanup  24  platform-specific cleanups, generate reflective calls
    delambdafy  25  remove lambdas
         icode  26  generate portable intermediate code
           jvm  27  generate JVM bytecode
       ploogin  28  A sample phase that does so many things it's kind of hard...
      terminal  29  the last phase during a compilation run
```

I get 
```
...
!! 1 - neg/t7494-no-options                      [output differs]
% diff /Users/nastucki/GitHub/scala-js/partest/fetchedSources/2.11.7/test/files/neg/t7494-no-options-neg.log /Users/nastucki/GitHub/scala-js/partest-suite/target/scala-2.11/test-classes/scala/tools/partest/scalajs/2.11.7/neg/t7494-no-options.check

[error] Failed: Total 4, Failed 2, Errors 0, Passed 2
```

Even tried copying the log and replace of the check with it and I get the same.

sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/3246/

gzm0:
@nicolasstucki could you solve the partest problem?

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+
+package org.scalajs.core.compiler
+
+import scala.tools.nsc
+import nsc._
+
+/** This `jspretyper` phase prepares a fix for issue SI-9487 in the case of
+ *  anonymous classes that extend js.Any.
+ *
+ *  The issue SI-9487 transforms durring `typer` some public method definitions
```
gzm0: during

gzm0: Also consider: During `typer`, Scalac transforms some [...] (SI-9487).
Or: Due to a bug (SI-9487), Scalac transfroms [...] during typer.

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+ *  of a quite specific syntactic form of anonymous classes int private methods.
+ *  This affects both methods and field accessors. This phase identifies any
+ *  anonymous class and add a `@WasPublicBeforeTyper` annotation on its public
```
gzm0: adds

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+ *  methods and fields. After the `typer` in `jsinterop` the anonymous classes
+ *  are fixed if they are js.Any using the annotations as reference.
```
gzm0: if they extend js.Any

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+ *  Additionally in `jsinterop` all the `@WasPublicBeforeTyper` are removed.
```
gzm0: .. `@WasPublicBeforeTyper` annotations are removed

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+        val newTree =
+          treeCopy.ClassDef(tree, tree.mods, tree.name, tree.tparams, newImpl)
+        super.transform(newTree)
+
+      case _ => super.transform(tree)
+    }
+  }
+
+  private def needsAnnotations(classDef: ClassDef): Boolean = {
+    classDef.name.toString == "$anon" &&
```
gzm0: Use `ANON_CLASS_NAME`

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+  override protected def newTransformer(unit: CompilationUnit): Transformer =
+    new PreTyperTransformer(unit)
+
+  class PreTyperTransformer(unit: CompilationUnit) extends Transformer {
+    override def transform(tree: Tree): Tree = tree match {
+      case tree: ClassDef if needsAnnotations(tree) =>
+        val newBody = tree.impl.body.map {
+          case vdef: ValDef if needsAnnotations(vdef) =>
+            treeCopy.ValDef(vdef, withWasPublic(vdef.mods), vdef.name,
+                vdef.tpt, vdef.rhs)
```
gzm0: The `rhs` needs a transform as well here (nested anonymous classes).

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+
+          case ddef: DefDef if needsAnnotations(ddef) =>
+            treeCopy.DefDef(ddef, withWasPublic(ddef.mods), ddef.name,
+                ddef.tparams, ddef.vparamss, ddef.tpt, ddef.rhs)
```
gzm0: The `rhs` needs a transform as well here (nested anonymous classes).

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  private def needsAnnotations(ddef: DefDef): Boolean =
+    ddef.mods.isPublic && ddef.name != nme.CONSTRUCTOR
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(anonymousClassMethodWasPublicAnnotations)
+
+  private lazy val anonymousClassMethodWasPublicAnnotations: List[Tree] = {
+    val runtimePackage =
+      Select(Select(Ident(newTermName("scala")), newTermName("scalajs")), newTermName("runtime"))
```
gzm0: Use `nme.scala_`.

gzm0: Use `nme.runtime`.

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,121 @@
...
+    val cls = Select(runtimePackage, newTypeName("WasPublicBeforeTyper"))
+    List(Apply(Select(New(cls), nme.CONSTRUCTOR), Nil))
```
gzm0: Create definitions for `newTermName("scalajs")` and `newTermName("WasPublicBeforeTyper")` (since this is going to become a def).

nicolasstucki:
There are still some nightly tests failing

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: So I guess my second comment applies. Currently you don't remove the annotation from Scala classes.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +968,80 @@ object ExportsTest extends JasmineTest {
       expect(jsPackage.ExportObjSetterNamed_=().x).toBe(1)
     }
 
+    it("should export js.Function values of new js.Object{...} - #1899") {
```
sjrd: should expose public members of ...

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +968,80 @@ object ExportsTest extends JasmineTest {
...
+
+      // Test that the bug is fixed for js.Any classes.
+
+      def getObj(): js.Object = new js.Object {
```
sjrd: It would also be good to test with an anonymous subclass of a strict subclass of `js.Object`.

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,125 @@
...
+            treeCopy.DefDef(ddef, withWasPublic(ddef.mods), ddef.name,
+                ddef.tparams, ddef.vparamss, ddef.tpt, transform(ddef.rhs))
+
+          case member => member
+        }
+        val newImpl =
+          treeCopy.Template(tree.impl, tree.impl.parents, tree.impl.self, newBody)
+        val newTree =
+          treeCopy.ClassDef(tree, tree.mods, tree.name, tree.tparams, newImpl)
+        super.transform(newTree)
```
gzm0: You don't need `super.transform` anymore now.

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,125 @@
...
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(List(anonymousClassMethodWasPublicAnnotation))
+
+  private val scalajs = newTermName("scalajs")
+  private val wasPublicBeforeTyper = newTypeName("WasPublicBeforeTyper")
+
+  private def anonymousClassMethodWasPublicAnnotation: Tree = {
+    val runtimePackage =
+      Select(Select(Ident(nme.scala_), scalajs), nme.runtime)
+    val cls = Select(runtimePackage,wasPublicBeforeTyper)
```
gzm0: Space after `,`[

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,63 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+    val needsFix: DefDef => Boolean = {
+      if (isInJSAny) {
+        new Function1[DefDef, Boolean] {
+          val accessorsNeedFix = body.collect {
+            case vdef: ValDef if needsCleanUp(vdef) => vdef.symbol
+          }.toSet
+          def apply(ddef: DefDef): Boolean = {
+            val sym = ddef.symbol
+            sym.isPrivate &&
+            (sym.isAccessor && accessorsNeedFix(ddef.symbol.accessed)) ||
```
gzm0: Replace `accessorsNeedFix(...)` by `sym.accessed.hasAnnotation(WasPublicBeforeTyper)`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
gzm0: I have read through this again. There is really no reason I can see that you do not do this in the main `case ddef: DefDef =>` in the main transform. Could you please clarify why you did implement this in your own transform?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -966,6 +968,141 @@ object ExportsTest extends JasmineTest {
...
+        def checkOriginalY2() = y2
+        def checkOriginalY3() = y3
+      }
+
+      testExposure(getJSObj())
+      testExposure(getJSObj2())
+      testExposure(getJSObj3())
+      testExposure(getJSObj4())
+
+      // Test that non js.Any classes where unaffectedd by the fix.
```
sjrd: Typo: were

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
nicolasstucki: I need it to transform members that have not been transformed yet by `case member => member`. But I will remove it from there and use `case member => transform(member)` instead.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: That does not work because when I reach the accessor method the annotation has been erased. Note that the members are processed in order and that variable definitions come before their accessors. Alternatively I could do the transformation in two passes, first handle the methods and then the variables.

gzm0: Oh, I see. Btw. I still do not understand why we need to remove the annotations. They will not be retained in the `.class` files anyways.

nicolasstucki: I will do it in two phases, the code will be cleaner.

sjrd:
> Btw. I still do not understand why we need to remove the annotations. They will not be retained in the .class files anyways.

Principle of least garbage, I guess? They'll still be retained in the Scala symbol table. But it's true that the symbol table for anonymous classes cannot be accessed anyway. Except by macros ... but then macros have already been processed at this point, so they saw the annotations anyway.

We try leaving them be, I guess.

gzm0:
> We try leaving them be, I guess.

It seems that if we are concerned about efficiency and we'll need two passes to remove them (current code is essentially a hand-implemented transformer pass), it might be worthwhile to at least consider to leave them.

I really wouldn't worry about annotations breaking subsequent phases. This is probably one of the few things scalac is resilient to :P

nicolasstucki:
If we decide to leave them, I could implement the transformation directly on each method instead of each template. This will remove an additional pass on the template.

nicolasstucki:
It has been decided, I will not clear the annotations.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -957,6 +963,38 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       case ctor: MethodSymbol if ctor.isPrimaryConstructor => ctor
     }
 
+  private def needsFixOrCleanup(template: Template): Boolean = {
```
sjrd: This method is not used anymore :)

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(List(anonymousClassMethodWasPublicAnnotation))
+
+  private val scalajs = newTermName("scalajs")
+  private val wasPublicBeforeTyper = newTypeName("WasPublicBeforeTyper")
+
+  private lazy val anonymousClassMethodWasPublicAnnotation: Tree = {
+    val runtimePackage =
+      Select(Select(Ident(nme.scala_), scalajs), nme.runtime)
```
sjrd: Should prefix that even by a `Select(Ident(nme.ROOTPKG), ...`. Otherwise this can be broken if there is something named `scala` in scope.

sjrd:
The commit message has obsolete stuff (it says annotations will be removed).

That's all.

gzm0:
> It has been decided, I will not clear the annotations.

This sounds like you have found more background than what we discussed on this PR. Could you please share it with this PR thread?

nicolasstucki:
@gzm0 we decided to keep the annotations around as you suggested because as you said they will be erased automatically later on. Additionally this allowed the implementation of the fix to be much simpler, now we only need to take the `ddef` from ` case ddef: DefDef if enclosingOwner is OwnerKind.RawJSType =>` (a case that already existed as a `ValOrDefDef`) and fix them directly there if needed. We don't have to worry about `Template`s and `ValDef`s anymore.

gzm0:
Ok. Thanks for clarifying.

sjrd:
Is there still an issue in the partests?

nicolasstucki:
I'll execute them locally to check again.

nicolasstucki:
The partests are not passing, but I just found out why. There are some whitespaces missing at the end of some lines in the check files.

sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/3314/

nicolasstucki:
The partests are failing on `run/macro-term-declared-in-trait` with the following diff: 
```scala
@@ -3,7 +3,8 @@ prefix = Expr[Nothing]({
     def <init>(): <$anon: Base> = {
       $anon.super.<init>();
       ()
-    }
+    };
+    <empty>
   };
   new $anon()
 })
```

If the tree is not transformed by `PreTyperComponent` the issue disappears.
```scala
class PreTyperTransformer(unit: CompilationUnit) extends Transformer {
  override def transform(tree: Tree): Tree = tree
}
```
From the other side, if the phase is trivialized to an empty transformation the issue is still there:
```scala
class PreTyperTransformer(unit: CompilationUnit) extends Transformer {
  override def transform(tree: Tree): Tree = super.transform(tree)
}
```

I will probably have to have more cases in my transformation to avoid cleaning the empty trees during this transformation.

nicolasstucki:
I fixed the partest issue using this [case](https://github.com/scala-js/scala-js/commit/c40b6f1b382571c7d2358013418cc702567f1472#diff-a7084dd83dcc67fe8ad7289d5d9c2a88R92). No other changes.

@sjrd could you re-run the partests to make sure that there is no rogue configuration where it fails.

sjrd:
Weird ... OK.

sjrd:
Nightly here: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/3317/

nicolasstucki:
Good, all partests succeeded.

library/src/main/scala/scala/scalajs/runtime/WasPublicBeforeTyper.scala
```diff
@@ -0,0 +1,3 @@
+package scala.scalajs.runtime
+
+class WasPublicBeforeTyper extends scala.annotation.Annotation
```
gzm0: I still do not understand why this is in `runtime`. If this is in `runtime`, so should `ExposedJSMember`.

sjrd: True. Let's put in in `js.annotation` like everything else, then.

nicolasstucki: Done

sjrd:
LGTM

gzm0:
Woooot!! Now I actually have something new to present in Japan :P

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  private def withWasPublic(mods: Modifiers): Modifiers =
+    mods.withAnnotations(List(anonymousClassMethodWasPublicAnnotation))
+
+  private val scalajs = newTermName("scalajs")
+  private val wasPublicBeforeTyper = newTypeName("WasPublicBeforeTyper")
+
+  private lazy val anonymousClassMethodWasPublicAnnotation: Tree = {
+    val runtimePackage = Select(Select(Select(Select(Ident(nme.ROOTPKG),
+        nme.scala_), scalajs), TermName("js")), nme.annotation)
```
sjrd: `newTermName`

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
```diff
@@ -0,0 +1,132 @@
...
+ *  anonymous classes that extend js.Any.
+ *
+ *  During `typer`, due to a bug (SI-9487), Scalac transfroms some public method
+ *  definitions of a quite specific syntactic form of anonymous classes into
+ *  private methods. This affects both methods and field accessors. This phase
+ *  identifies any anonymous class and adds a `@WasPublicBeforeTyper` annotation
+ *  on its public methods and fields. After the `typer` in `jsinterop` the
+ *  anonymous classes are fixed if they extend js.Any using the annotations as
+ *  reference. Additionally in `jsinterop` all the `@WasPublicBeforeTyper`
+ *  annotations are removed.
```
sjrd: The last sentence is obsolete and should be removed.

sjrd:
Also typo in the commit message: missing 'n' in anonymous.

compiler/src/main/scala/org/scalajs/core/compiler/PreTyperComponent.scala
gzm0: This should be a val as well

nicolasstucki:
Dear Jenkins, retest this please.

sjrd:
LGTM

nicolasstucki:
Dear Jenkins, retest this please.

nicolasstucki:
Dear Jenkins, retest this please.

PR #2059, opened 2015-11-27
Fix #2053: Always use UTF8 in PhantomJSEnv

### LINKED ISSUE #2053: Use UFT8 in PhantomJSEnv
`FileWriter` is used throughout PhantomJSEnv, which defaults to the system encoding. If a user has not set the java system property `file.encoding` to UTF-8 and a dependency contains a UTF-8 character, builds may fail with the error "org.scalajs.jsenv.ExternalJSEnv$NonZeroExitException: PhantomJS2 exited with code 2". 

In my case, this was due to https://github.com/widok/scala-js-momentjs pulling in `moment-with-locales.min.js`, which contains many UTF-8 characters and using them in regexes. The resulting errors were `Uncaught SyntaxError: Invalid regular expression: /?|?/: Nothing to repeat` due to regexes where the non-ascii characters were all represented by question marks.

gzm0:


gzm0:
Review by @sjrd

sjrd:
LGTM

sjrd:
retest this please

PR #2070, opened 2015-11-29
Prove that `js.|[X, Y]` is `js.Any` (given that `X` and `Y` are, of c…

elacin:
There is also this thing that seem to crop up every once in a while that i would have expected to just work.
Im not sure if i need to keep those `ClassTag`s around, or if a simple cast would do the trick

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Union.scala
+++ b/library/src/main/scala/scala/scalajs/js/Union.scala
@@ -10,6 +10,7 @@
 package scala.scalajs.js
 
 import scala.language.implicitConversions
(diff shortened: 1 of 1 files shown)


sjrd:
This will silently not work (i.e., blow up at run-time) if `A` and/or `B` are traits inheriting from `js.Any`. Also, it's silently unsound if `A` and `B` have the same erasure, such as `List[Int]` and `List[String]`.

I don't think this should get in.

elacin:
Awesome review there, @sjrd - right on both counts - it definitely blows up. Thanks, closing this

PR #2081, opened 2015-12-02
Create independent CI tasks for testSuite.

nicolasstucki:
Now testSuite runs are in tasks outside of the main one.
There is one task for ECMAScript5, one for ECMAScript6 and
one for ECMAScript6StrongMode.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -45,6 +45,21 @@
         'set scalaJSStage in Global := FastOptStage' \
         ++$scala testingExample/test &&
     sbtretry ++$scala testSuiteJVM/test testSuiteJVM/clean &&
+    sbtretry 'set scalaJSStage in Global := FastOptStage' \
+        ++$scala javalibExTestSuite/test \
+        'set scalaJSStage in Global := FullOptStage' \
+        ++$scala javalibExTestSuite/test &&
+    sbtretry ++$scala testSuite/test:doc compiler/test reversi/fastOptJS reversi/fullOptJS &&
+    sbtretry ++$scala compiler/compile:doc library/compile:doc javalibEx/compile:doc \
+        testInterface/compile:doc jasmineTestFramework/compile:doc &&
+    sbtretry ++$scala partest/fetchScalaSource &&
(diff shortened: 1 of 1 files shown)


ci/matrix.xml
```diff
@@ -132,7 +151,11 @@
         'set scalaJSOptimizerOptions in testSuite ~= (_.withDisableOptimizer(true))' \
         'set scalaJSStage in Global := FastOptStage' \
         ++$scala testSuite/test \
-        testSuite/clean &&
+        testSuite/clean
```
gzm0: Do we really need a clean at the end of a task?

nicolasstucki: We might do it only once at the end of each task with a single `sbt clean`.

gzm0: Why should we do it at all? IIUC, the cleans where there, when options changed. So at the end of the task, they don't seem to make sense at all.

sjrd: Indeed. We need `clean` in between invocations of the linker when the options change. We don't a `clean` at the end of a job. That said, it doesn't bother me to leave it there, for uniformity.

nicolasstucki:
Jenkins, retest this please.

nicolasstucki:
Jenkins, retest this please.

sjrd:
The second build took 5h51 minutes, which is very close to the total limit of 6 hours :s

gzm0:
Is the goal of this to make the builds faster? If so, we need more machines :)

nicolasstucki:
The immediate goal is not necesarly to make it faster, but to remove the bottleneck from `main`. Most of the time the Jenkins is waiting for one single. 

Those 6 hours were also taken while executing the other 8 PR tests that I lunched at that time.

sjrd:
The second commit is wrong. It does everything it shouldn't and nothing it should. We *need* the cleans between invocations of the test with different options (such as setting semantics etc.). But don't need the clean at the end of the job.

I suggest keeping the first commit as it was. As I said, I'm not bothered with finishing by cleaning because it's more symmetric with the other invocations, and we are less likely to add a new configuration while forgetting to add the clean.

sjrd:
LGTM

PR #2090, opened 2015-12-03
Fix #2040: Remove "utf_8" from Charset.CharsetMap.

### LINKED ISSUE #2040: Charset.forName("UTF_8") should throw.
On JDK8 `Charset.forName("UTF_8")` throws, but on Scala.js it succeeds.
Note that it is the underscore notation that is not supported.

nicolasstucki:


sjrd:
LGTM

PR #2091, opened 2015-12-04
Fix #2048: Handle URI.{_authority|_path} ""/undefined cases.

### LINKED ISSUE #2048: Methods from java.net.URI return wrong empty values.
Methods:
- getAuthority
- getPath
- getRawAuthority

are returning different empty values in JS and JVM. In some cases they return `""` when they should return `null` and in other cases it is the other way around. 

Tests for these cases exist in #2047 with the `!executingInJVM ` filter.

nicolasstucki:


gzm0:
Could you please make sure, that `normalize`, `relativize` and `resolve` work properly given this behavior? (I did assume the current behavior while writing them).

nicolasstucki:
`normalize`, `relativize` and `resolve` do have the correct behaviour with `""`. I added a few tests on them just to be sure.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -135,13 +135,15 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
     case _ => false
   }
 
-  def getAuthority(): String = _authority.map(decodeComponent).orNull
+  def getAuthority(): String =
+    _authority.filter(_ != "").map(decodeComponent).orNull
```
sjrd: What happens to all the other uses of `_authority` in the file? Don't they need this treatment as well? Would it make sense to add `.filter(_ != "")` immediately when creating `_authority` in the first place?

gzm0: Very good point. A couple of things use `_authority`. Notably:

- comparison
- `parseServerAuthority`
- `relativize`
- `resolve`

nicolasstucki: If I apply the filter directly on `_authority` other parts of `URI` start failing because the internal representation expects that format. I will look at what else needs to change.

nicolasstucki: It seems like for all for all the methods the current definition of `_authority` is the correct one (i.e. it can be `""` or `undefined`). They actually rely on this differentiation in several places using `_authority.isDefined` and the current behaviour is consistent with the JVM implementation.

gzm0: Could you please add the examples you used for testing to the tests? As you have certainly observed, the current tests would not expose a behavioral difference between having an empty authority and no authority.

gzm0:
There is a typo in the commit message (Handel vs Handle).

lihaoyi:
`Authority` is also mis-spelled as `Autority` =) Not sure if anyone noticed but nobody mentioned it and if you care about handel you probably care about that too

gzm0:
@lihaoyi of course we care! (You really should have followed this project long enough to know that typos like that are much more important than actually getting code in :P).

javalib/src/main/scala/java/net/URI.scala
nicolasstucki: Note that the only case where the distinction is necessary is in the computation of `_path` and the test cases cover non empty strings, empty strings and `null`.

nicolasstucki: @gzm0, is there some particular combination that you saw that I could be missing?

nicolasstucki:
Last failure was due to #2095.

javalib/src/main/scala/java/net/URI.scala
gzm0: Yes. For example (results not shown):

```scala
relativize("file:///a", "file:///a/b/")
relativize("file:/a", "file:///a/b/")
resolve("http://foo/b", "///a")
resolve("http://foo/b", "/a")
```

gzm0: For example:

```
scala> new java.net.URI("http://foo/a").resolve("///b")
res4: java.net.URI = http://foo/b
```

I'm pretty sure the current implementation would return `http:///b` or even `///b`.

nicolasstucki: Yes, you are right, those are problematic.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -41,18 +41,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
   private val _host = fld(AbsHost, RelHost)
   private val _port = fld(AbsPort, RelPort).fold(-1)(_.toInt)
 
   private val _path = {
-    if (_isAbsolute) {
-      if (_authority.isDefined) fld(AbsNetPath)
-      else fld(AbsAbsPath)
+    if (fld(AbsAuthority, RelAuthority).isDefined) {
+      val netPath = fld(if (_isAbsolute) AbsNetPath else RelNetPath)
+      netPath orElse ("": js.UndefOr[String])
```
gzm0: replace by: `fld(AbsNetPath, RelNetPath)`
`net_path` contains the authority. So if the authority is non-empty, so is `net_path`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -121,7 +116,13 @@ class URITest {
   @Test def should_parse_relative_URIs_with_registry_based_authority(): Unit = {
     expectURI(new URI("//foo:bar"), false, false)(
         authority = "foo:bar",
+        path = "",
         schemeSpecificPart = "//foo:bar")()
+    expectURI(new URI("http://foo:bar"), true, false)(
+      authority = "foo:bar",
+      path = "",
+      scheme = "http",
+      schemeSpecificPart = "//foo:bar")()
```
gzm0: This is not a relative URI as the title says

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -273,6 +279,16 @@ class URITest {
     resTest("http:g", "http:g")
   }
 
+  @Test def should_provide_resolve_issue_2048(): Unit = {
```
gzm0: Consider adding something more descriptive and just linking the bug in a comment. It's nice to understand what is tested without going to the bugtracker.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -41,18 +41,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+  private val _authority = fld(AbsAuthority, RelAuthority).filter(_ != "")
   private val _userInfo = fld(AbsUserInfo, RelUserInfo)
   private val _host = fld(AbsHost, RelHost)
   private val _port = fld(AbsPort, RelPort).fold(-1)(_.toInt)
 
   private val _path = {
-    if (_isAbsolute) {
-      if (_authority.isDefined) fld(AbsNetPath)
-      else fld(AbsAbsPath)
+    if (fld(AbsAuthority, RelAuthority).isDefined) {
```
gzm0: Please refactor this to a `_hasNetPath` field.

javalib/src/main/scala/java/net/URI.scala
gzm0: Last comment is wrong. But you can replace this with:

```scala
fld(if (_isAbsolute) AbsNetPath else RelNetPath).getOrElse("")
```

Maybe you need to annotate `_path` with its type, but that's still nicer than having to `"": js.UndefOr[String]`.

javalib/src/main/scala/java/net/URI.scala
nicolasstucki: Why a field? We use it only once for the initialization of `_path`.

gzm0: Yeah. You are right. So make it a local val. But like this it is very hard to understand what your intent is.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -41,19 +41,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
-    if (_isAbsolute) {
-      if (_authority.isDefined) fld(AbsNetPath)
-      else fld(AbsAbsPath)
-    } else {
-      if (_authority.isDefined) fld(RelNetPath)
-      else fld(RelAbsPath) orElse fld(RelRelPath)
-    }
+    val useNetPath = fld(AbsAuthority, RelAuthority).isDefined
+    if (useNetPath)
+      fld(AbsNetPath, RelNetPath) orElse ("": js.UndefOr[String])
```
gzm0: Why don't you use `getOrElse` here?

nicolasstucki: Then this would be of type `String` instead of `UndefOr[String]`. Then I would either put it back in an `UndefOr[String]` or change `_path` to `String` which would need changes all around the file.

Additionally type inference seems to broken with that:
```scala
[error] /Users/nastucki/GitHub/scala-js/javalib/src/main/scala/java/net/URI.scala:52: error: type mismatch;
[error]  found   : Object
[error]  required: scala.scalajs.js.UndefOr[String]
[error]       fld(AbsNetPath, RelNetPath).getOrElse("")
```
It Might be related to #2095.

gzm0: Ok. So let's leave it like that.

nicolasstucki:
Rebased on #2107 and removed the now unnecessary `: js.UndefOr[String]` in the `_path` initialization.

sjrd:
@gzm0 Any other comments, here?

gzm0:
No. LGTM

PR #2092, opened 2015-12-04
Fix #2049: Remove NPE tests on buf in InputStream.read(buf, ...).

### LINKED ISSUE #2049: InputStream.read is throwing wrong type of exception.
Tests in `InputStreamTest` filter out this case with `executingInJVM`.

nicolasstucki:


sjrd:
IMO that's the kind of NPE we don't want to handle specially. They fall into the UB case (and throw a `TypeError` instead, in this case).

The IndexOutOfBoundExceptions should be thrown reliably, though. Weren't they before?

gzm0:
> IMO that's the kind of NPE we don't want to handle specially

I agree. The original code **would** already NPE under proper JVM semantics. So if we add a semantic flag at a later point in time to do this, everything would be fine.

nicolasstucki:
Unfortunately the original code was failing with a `JavaScriptException`.

nicolasstucki:
This is what I get if I remove the explicit NPE:

```
error] Test org.scalajs.testsuite.javalib.io.InputStreamTest.should_provide_a_default_implementation_of_read_to_an_array failed: java.lang.AssertionError: unexpected exception type thrown;NullPointerException JavaScriptException, took 0.008999936 sec
[error]     at org.scalajs.testsuite.utils.AssertThrows$.expectThrows(/Users/nastucki/GitHub/scala-js/test-suite/js/target/scala-2.11/file:/Users/nastucki/GitHub/scala-js/test-suite/shared/src/test/scala/org/scalajs/testsuite/utils/AssertThrows.scala:24)
[error]     at org.scalajs.testsuite.utils.AssertThrows$.expectThrows(/Users/nastucki/GitHub/scala-js/test-suite/js/target/scala-2.11/file:/Users/nastucki/GitHub/scala-js/test-suite/shared/src/test/scala/org/scalajs/testsuite/utils/AssertThrows.scala:51)
[error]     at org.scalajs.testsuite.javalib.io.InputStreamTest.should$undprovide$unda$unddefault$undimplementation$undof$undread$undto$undan$undarray(/Users/nastucki/GitHub/scala-js/test-suite/js/target/scala-2.11/file:/Users/nastucki/GitHub/scala-js/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/InputStreamTest.scala:40)
```

nicolasstucki:
The original code would also fail on these two new cases that I just added. 
```scala
    expectThrows(classOf[NullPointerException], stream.read(null, -1, 10))
    expectThrows(classOf[NullPointerException], stream.read(null, 0, -1))
```

sjrd:
Yes, a `TypeError` is a JS exception, which is wrapped as `JavaScriptException` when caught (because we need a subtype of `Throwable`). This is the expected behavior.

Excepting NPE is undefined behavior in Scala.js.

sjrd:
LGTM

PR #2093, opened 2015-12-04
Fix #2089: StringReader.ready returns always true or throws.

### LINKED ISSUE #2089: StringReader.ready should not return false when at the end.

nicolasstucki:


sjrd:
LGTM

PR #2096, opened 2015-12-08
Port and cross-compile tests for java.time

ummels:

PR #2098, opened 2015-12-09
Fix #2097: LocalDate.range returns incorrect result

### LINKED ISSUE #2097: LocalDate.range returns incorrect result
Given an instance `d` of `LocalDate`, `d.range(ChronoField.DAY_OF_MONTH)` returns a range of 1 - 28/31, when on the JVM the upper bound is adapted to the length of the month of `d`. For instance, `LocalDate.of(2012, 2, 1).range(ChronoField.DAY_OF_MONTH)` should return a range of 1 - 29.

ummels:


test-suite/js/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/temporal/TemporalAccessorTest.scala
```diff
@@ -17,7 +17,7 @@ trait TemporalAccessorTest extends JasmineTest with ExpectExceptions {
     it("should respond to `range`") {
       for {
         accessor <- samples
-        field <- ChronoField.values()
+        field <- ChronoField.values() if field != ChronoField.DAY_OF_MONTH
```
ummels: Just a hack to make the test pass. The correct test is included in #2096.

nicolasstucki: Put the correct test directly in this commit.

ummels: Fixed. Would be better to have the test in `LocalDateTest.scala`, but we need to add the exception here anyway and it's only temporarily anyway.

nicolasstucki:
Jenkins, test this please.

nicolasstucki:
LGTM

PR #2101, opened 2015-12-10
Improve the story for setting up an Eclipse development environment.

sjrd:


sjrd:
Review by whoever feels like it ;)

.gitignore
```diff
@@ -8,3 +8,4 @@ target/
 /cli/pack/
 /.idea/
 /.idea_modules/
+bin/
```
gzm0: ?

DEVELOPING.md
```diff
@@ -70,6 +70,21 @@ or, more typically,
...
 
+## Eclipse
+
+If you want to develop in Eclipse, use
+[sbteclipse](https://github.com/typesafehub/sbteclipse). Projects as created by
+the build by default are not suited for Eclipse. You can create *somewhat*
+appropriate projects with:
+
+    $ sbt tools/sources
+    $ GENERATING_ECLISE=true sbt "eclipse with-source=true"
```
gzm0: `GENERATING_ECLIPSE`?

.gitignore
sjrd: The new versions of sbteclipse always compile into `bin/` folders rather than `classDirectory.value`.

DEVELOPING.md
sjrd: Indeed

project/Build.scala
```diff
@@ -573,7 +601,6 @@ object Build extends sbt.Build {
           normalizedName := "sbt-scalajs",
           name in bintray := "sbt-scalajs-plugin", // "sbt-scalajs" was taken
           sbtPlugin := true,
-          scalaVersion := "2.10.6",
```
gzm0: Doesn't this make 2.11.7 the default?

project/Build.scala
```diff
@@ -459,7 +484,10 @@ object Build extends sbt.Build {
           },
           exportJars := true
       )
-  )
+  ).configure { p =>
+    if (isGeneratingEclipse) p.dependsOn(irProject)
+    else p
+  }
```
gzm0: Consider making a `dependsOnSource`

project/Build.scala
```diff
@@ -598,7 +625,8 @@ object Build extends sbt.Build {
 
   lazy val delambdafySetting = {
     scalacOptions ++= (
-        if (scalaBinaryVersion.value == "2.10") Seq()
+        if (isGeneratingEclipse) Seq()
+        else if (scalaBinaryVersion.value == "2.10") Seq()
```
gzm0: I do not understand when you use `if (x || y) Seq() else Seq(...)` and when you use if (x) Seq() else if (y) Seq() else Seq(...)`

project/Build.scala
```diff
@@ -322,13 +332,26 @@ object Build extends sbt.Build {
...
   )
 
-  /** Depend library as if (exportJars in library) was set to false */
-  val compileWithLibrarySetting = {
-    internalDependencyClasspath in Compile ++= {
-      val prods = (products in (library, Compile)).value
-      val analysis = (compile in (library, Compile)).value
+  implicit class ProjectOps(val project: Project) extends AnyVal {
+    /** Depends on the compiler */
+    def dependsOnCompiler: Project =
```
gzm0: Consider calling this `withScalaJSCompiler` or something. If we already have a dedicated method, it seems to me that "depending" is not the best term.

project/Build.scala
```diff
@@ -322,13 +332,26 @@ object Build extends sbt.Build {
...
+      if (isGeneratingEclipse) project
+      else project.dependsOn(compiler % "plugin")
 
-      prods.map(p => Classpaths.analyzed(p, analysis))
+    /** Depend library as if (exportJars in library) was set to false */
+    def compileWithLibrary = {
```
gzm0: This however, is much more of a normal dependency :) (don't you think so?)

project/Build.scala
```diff
@@ -322,13 +332,26 @@ object Build extends sbt.Build {
...
-      val prods = (products in (library, Compile)).value
-      val analysis = (compile in (library, Compile)).value
+  implicit class ProjectOps(val project: Project) extends AnyVal {
+    /** Depends on the compiler */
+    def dependsOnCompiler: Project =
+      if (isGeneratingEclipse) project
+      else project.dependsOn(compiler % "plugin")
 
-      prods.map(p => Classpaths.analyzed(p, analysis))
+    /** Depend library as if (exportJars in library) was set to false */
```
gzm0: Depend on

project/Build.scala
sjrd: Yes, it does. And that allows `sbteclipse` (and IntelliJ, apparently) to generate a somewhat OK-ish project. Without that it would just blow up with dependency resolution hell.

This does not make actual building harder. It was always necessary to do `++2.10.6` before compiling the sbt plugin anyway, because of the dependent projects (tools and so on).

.gitignore
gzm0: And you keep telling people that IDE specific ignores should not live in project local `.gitignore`s???

project/Build.scala
sjrd: Makes sense.

project/Build.scala
gzm0: Ok. Fair enough. And its not like anyone compiles it from sbt anyways... just `reload` :P

project/Build.scala
sjrd: I use if/else if when using `||` makes the line too long ^^
But I should probably use if/else if everywhere, indeed.

.gitignore
gzm0: Oh... I see we have the idea ignores as well. Ok fair enough. However, you should probably anchor them at the top level.

sjrd: I can't anchor it at the top-level. It's like `target/`. There's one in every subproject.

The IDE-specific gitignores made it into this repo before I was strict enough to reject them.

gzm0: Ok

sjrd:
Updated.

project/Build.scala
```diff
@@ -322,13 +333,37 @@ object Build extends sbt.Build {
...
-  /** Depend library as if (exportJars in library) was set to false */
-  val compileWithLibrarySetting = {
-    internalDependencyClasspath in Compile ++= {
-      val prods = (products in (library, Compile)).value
-      val analysis = (compile in (library, Compile)).value
+  implicit class ProjectOps(val project: Project) extends AnyVal {
+    /** Uses the Scala.js compiler plugin. */
+    def withScalaJSCompiler: Project =
+      if (isGeneratingEclipse) project
+      else project.dependsOn(compiler % "plugin")
```
gzm0: We should probably consolidate this with `myScalaJSSettings` using a `Project` transformer (later, different PR).

gzm0:
LGTM

Issue #1997, opened 2015-11-02
Need @JSSymName(..) annotation


bblfish:
In order to implement iterators one needs a way to tie a a method to a symbol as explained in [my comment on issue 1141: Add js.Iterator according to ES6's iterator protocol](https://github.com/scala-js/scala-js/issues/1141#issuecomment-152741877).
Something like this perhaps:

```
@js.native
trait Iterable[+A] extends js.Object {
  @JSSymName("Symbol.iterator")
  def iterator(): Iterator[A] = js.native
}
```

or perhaps

```
@js.native
trait Iterable[+A] extends js.Object {
  @JSName(Symbol.iterator)
  def iterator(): Iterator[A] = js.native
}
```

when a `Symbol` class has been created.

Workarounds using using `JSBracketAccess` do not seem to work: I tried in this [JSIterable](https://github.com/read-write-web/rww-scala-js/blob/akka.js/src/main/scala/org/scalajs/js/collection/JSIterable.scala) class, and somehow am always getting a function back when I want a value.  The context ends up missing as discussed [on gitter](https://gitter.im/scala-js/scala-js?at=56363fc6643ae8fb655b9af4).

sjrd:
Define `JSIterable` as:
```scala
@js.native
trait JSIterable[+A] extends js.Object {
  @JSBracketAccess
  def method[R](symbol: js.Any)(): R = js.native
}
```

bblfish:
With the changes to [JSIterable](https://github.com/read-write-web/rww-scala-js/blob/6dfbd31a37676022034ca4de750782d93d926489/src/main/scala/org/scalajs/js/collection/JSIterable.scala)  that is called by [ServiceWorkerAuth](https://github.com/read-write-web/rww-scala-js/blob/6dfbd31a37676022034ca4de750782d93d926489/src/main/scala/rww/auth/ServiceWorkerAuth.scala) I get the following output in the console:
![screen shot 2015-11-02 at 21 34 27](https://cloud.githubusercontent.com/assets/124506/10894983/831d1fce-81a9-11e5-9f71-1c06e84141b5.png)

bblfish:
Ah the error was elsewhere: my logger was creating the function. This diff fixed it
https://github.com/read-write-web/rww-scala-js/commit/c8ae2c226728b6d1b4fcdbf3929c888a1974fba5

bblfish:
[...]

PR #2110, opened 2015-12-12
DO NOT MERGE For discussion: add the ES6 standard library

sjrd:
This PR is more like a Request For Comments on the best design for the integration of the ES6 standard library to the typings in `scala.scalajs.js`.

Several things are definitely worth discussing:
- We should probably add the Scala collection API on `js.Map` and `js.Set`
- Should we put these things in a separate namespace?
- <del>Typings of `Promise` and `Thenable`.</del>
- Typings of `Generator`.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Generator.scala
+++ b/library/src/main/scala/scala/scalajs/js/Generator.scala
@@ -1,0 +1,60 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.js
+
+import scala.language.implicitConversions
+
+import scala.annotation.implicitNotFound
+import scala.annotation.unchecked.uncheckedVariance
+
+import scala.scalajs.js
+import js.annotation._
+
+/** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+ *  JavaScript Generator, such as the result of generator function.
+ *
+ *  @tparam V The type of elements returned by `next()`
+ *  @tparam I The type of values expected by `next()` as parameter
+ *  @tparam R The type of the result/done value.
+ */
+@ScalaJSDefined
+trait Generator[+V, -I, +R] extends js.Object {
+  def next(value: I): Generator.Entry[V, R]
+
+  def `return`[S](result: S): Generator.Entry[Nothing, S]
+
+  def `throw`(ex: scala.Any): Generator.Entry[V, R]
+}
+
+object Generator {
+  /** The type of a simple Generator that takes no input in `next()` and
+   *  returns no value.
+   */
+  type Iter[+V] = Generator[V, Unit, Unit]
+
+  implicit def asIterator[V](g: Generator[V, Unit, _]): js.Iterator[V] =
+    g.asInstanceOf[js.Iterator[V]]
+
+  implicit def toIteratorOps[V](g: Generator[V, Unit, _]): js.Iterator.IteratorOps[V] =
+    asIterator(g)
+
+  @ScalaJSDefined
+  trait Entry[+V, +R] extends js.Object {
+    def value: V @uncheckedVariance | R @uncheckedVariance
+    def done: Boolean
+  }
+
+  object Entry {
+    implicit class EntryOps[+V, +R](val self: Entry[V, R]) extends AnyVal {
+      @inline def iterValue: V = self.value.asInstanceOf[V]
+      @inline def doneValue: R = self.value.asInstanceOf[R]
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


library/src/main/scala/scala/scalajs/js/Generator.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import scala.scalajs.js
+import js.annotation._
+
+/** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+ *  JavaScript Generator, such as the result of generator function.
+ *
+ *  @tparam V The type of elements returned by `next()`
+ *  @tparam I The type of values expected by `next()` as parameter
+ *  @tparam R The type of the result/done value.
```
gzm0: I do not understand what you gain by making the distinction between `V` and `R` here...

library/src/main/scala/scala/scalajs/js/Generator.scala
```diff
@@ -0,0 +1,60 @@
...
+ */
+@ScalaJSDefined
+trait Generator[+V, -I, +R] extends js.Object {
+  def next(value: I): Generator.Entry[V, R]
```
gzm0: IIUC, `value` should be optional here.

library/src/main/scala/scala/scalajs/js/Iterable.scala
```diff
@@ -0,0 +1,41 @@
...
+      extends scala.collection.Iterable[A] {
+
+    @inline
+    def jsIterator(): Iterator[A] =
+      self.asInstanceOf[IteratorCallAccess[A]].iteratorCall(Symbol.iterator)()
+
+    @inline
+    def iterator: scala.collection.Iterator[A] = self.jsIterator()
+  }
+}
```
gzm0: If this is a direction JavaScript will take (i.e. defining methods based on Symbols), we need language support for this...

library/src/main/scala/scala/scalajs/js/Iterable.scala
```diff
@@ -0,0 +1,41 @@
...
+  @inline
+  implicit final class IterableOps[+A](val self: Iterable[A])
+      extends scala.collection.Iterable[A] {
+
+    @inline
+    def jsIterator(): Iterator[A] =
+      self.asInstanceOf[IteratorCallAccess[A]].iteratorCall(Symbol.iterator)()
+
+    @inline
+    def iterator: scala.collection.Iterator[A] = self.jsIterator()
```
gzm0: This shouldn't be needed (but pimped through `IterableOps`).

gzm0: This *is* iterable ops... silly me

library/src/main/scala/scala/scalajs/js/Iterable.scala
sjrd: Yes, it is. There's a dozen or so so-called well-known symbols, that allow classes to implement methods that interact with the language. `Symbol.iterator` is one of them (the most widely supported so far) and underlies `for-or` loop.

We have a ticket for language support at #1997.

library/src/main/scala/scala/scalajs/js/Generator.scala
sjrd: Many common generators yield values of some type (`V`), but do not `return` anything, hence `R` is `Unit` most of the time. But they *can* return something else.

library/src/main/scala/scala/scalajs/js/Generator.scala
sjrd: Nope. It is only optional if `I >: Unit`, but that cannot be expressed. Ideally, for `I >: Unit`, the implicit conversion to `Iterator[V]` would kick in, and you would be able to call `next()` from `Iterator[V]`. Unfortunately, the automatic `()` insertion kicks in first, and issues a deprecation warning -_-'. For once, I wish it hadn't been deprecated, since it models so perfectly well the fact that not passing an argument in JS yields `undefined` on the other side :-p

library/src/main/scala/scala/scalajs/js/Generator.scala
gzm0: Yes, I more meant that in the typing you have so far, `R` and `V` always appear together. They are actually only relevant to define (or compose) generators. Do you already have plans for that?

library/src/main/scala/scala/scalajs/js/Generator.scala
gzm0: I see. I understand this `next` thing better now. So essentially, the drawback of this typing is that you must pass in a value for the first `next()` call, right?

library/src/main/scala/scala/scalajs/js/Generator.scala
sjrd: I don't understand. They are used separately in `EntryOps`, and more importantly only `V` is relevant when "upcasting" to `js.Iterator[V]`.

library/src/main/scala/scala/scalajs/js/Generator.scala
sjrd: Hah! There is that, too. The first `next()` call. I was actually talking about a common subset of generators that don't take any input, i.e., they accept `undefined`/`()`, but that's an even trickier one. It's true that `next()`'s first call has an asymmetry in that its parameter is always discarded no matter what. I will think more about this.

library/src/main/scala/scala/scalajs/js/Generator.scala
gzm0: True. For the conversion to iterator it is definitely important.

gzm0:
We should probably revisit this now. I think anything but `Generator` is pretty undisputed.

hepin1989:
ECMA-262, 8th edition, June 2017 http://www.ecma-international.org/ecma-262/

hepin1989:
`Object.assign`

sjrd:
Closing as this would probably have to be redone from scratch at this point.

PR #2112, opened 2015-12-14
Fix #2111: Make JUnit annotations static.

### LINKED ISSUE #2111: JUnit annotations should be static.
The issue arises when compiling a test hierarchy where some parent class was compiled before and is not being recompiled. For example, if we compile
```scala
class A {
  @Test def someTestOfA(): Unit = ...
}
```
And later add a new class `B` and compile
```scala 
class B extends A
```
then `B` will not be transformed because `A` has lost the `@Test` annotation.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

sjrd:
Can we write tests for this? Like defining an abstract base class with an `@Test` method in the `main/` directory, then subclass it in the `test/` configuration?

nicolasstucki:
Good idea

nicolasstucki:
No, that would not work as the JUnit plugin is only active in the test mode. I would have to write a second project with tests and depend on it. I'm not sure if that is worth it.

sjrd:
Does the plugin need to run on abstract classes? I thought it did not.

nicolasstucki:
No, it does not, I missed that. I'll implement it that way.

sjrd:
LGTM

PR #2120, opened 2015-12-16
Add junit bootstrapper classes to the currentRun.symSource map

lrytz:
Classes being compiled in the current compilation run are expected to
be found in this map. In this particualr case, the backend would not
build the InlineInfo for these classes and later crash when the
optimizer is enabled.

### DIFF
--- a/junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
+++ b/junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
@@ -241,6 +241,8 @@ class ScalaJSJUnitPlugin(val global: Global) extends NscPlugin {
         scalaJSPlugin.registerModuleExports(bootSym)
         bootClazz.setSymbol(bootSym)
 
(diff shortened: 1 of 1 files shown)


lrytz:
Review by @nicolasstucki. Similar in scala/scala: https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/transform/Delambdafy.scala#L298-L299, https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/transform/AddInterfaces.scala#L107-L108

lrytz:
This should fix the crashes observed in a community build with the 2.12-optimizer enabled (https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/269/consoleFull, discussion at https://gitter.im/scala-js/scala-js?at=567143223fdae127593c4082.

I verified the fix locally. A community build that includes this fix for scala-js is running here: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/270/console

sjrd:
Dear Jenkins, ok to test

nicolasstucki:
LGTM

PR #2123, opened 2015-12-18
Fix #2109: LocalDate.until sometimes returns wrong result

### LINKED ISSUE #2109: LocalDate.until sometimes returns wrong result
`LocalDate.until(d1, d2)` sometimes returns a wrong result. For instance `LocalDate.of(2013, 3, 1).until(LocalDate.of(2013, 1, 28))` returns a period of -1 months and -1 days, when it should return a period of -1 months and -4 days.

ummels:
Also add private lazy val for proleptic month.

### DIFF
--- a/javalib/src/main/scala/java/time/LocalDate.scala
+++ b/javalib/src/main/scala/java/time/LocalDate.scala
@@ -24,6 +24,7 @@ final class LocalDate private (year: Int, month: Month, dayOfMonth: Int)
 
   private lazy val dayOfYear =
     month.firstDayOfYear(_isLeapYear) + dayOfMonth - 1
+
   private lazy val epochDay = {
     val year1 = year - 1970
     val daysBeforeYear = daysBeforeYears(MathJDK8Bridge.floorMod(year1, 400))._2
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/time/LocalDate.scala
```diff
@@ -307,17 +310,22 @@ final class LocalDate private (year: Int, month: Month, dayOfMonth: Int)
...
+      if (dmonths > 0 && ddays < 0) -1
+      else if (dmonths < 0 && ddays > 0) 1
+      else 0
+    }
+    val months = dmonths + corr
+    val days = {
+      if (corr < 0) plus(months, MONTHS).until(other, DAYS).toInt
+      else if (corr > 0) ddays - other.lengthOfMonth
+      else ddays
     }
```
ummels: This seems to be the only way to stay consistent with the JDK implementation.

sjrd:
Jenkins, test this please

javalib/src/main/scala/java/time/LocalDate.scala
```diff
@@ -31,9 +32,12 @@ final class LocalDate private (year: Int, month: Month, dayOfMonth: Int)
       MathJDK8Bridge.floorDiv(year1, 400).toLong * daysInFourHundredYears
     offset + daysBeforeYear + dayOfYear - 1
   }
+
   private lazy val dayOfWeek =
     MathJDK8Bridge.floorMod(epochDay + 3, 7).toInt + 1
 
+  private lazy val prolepticMonth = year.toLong * 12 + getMonthValue - 1
```
nicolasstucki: Do we really need to keep this as a field? Wouldn't it be better as a `def`?

nicolasstucki:
In general LGTM. Will just wait on the answer for the field question before merging.

javalib/src/main/scala/java/time/LocalDate.scala
ummels: Yeah, I guess we can make it a `def` since only a few methods rely on it.

ummels:
Updated. Please retest.

nicolasstucki:
Jenkins, retest this please.

PR #2126, opened 2015-12-20
Fix #2125: Make null.asInstanceOf[Nothing] throw.

### LINKED ISSUE #2125: null.asInstanceOf[Nothing] should throw UBE
The following code produces random behaviors, depending on how much we allow the optimizer to do:
```scala
  @noinline def foo(): String = null
  @noinline def bar(): Nothing = foo().asInstanceOf[Nothing]
  def main() {
    val n = bar()
    println(n)
  }
}
```
With the optimizer, this executes without error and does not print anything. Without the optimizer, it prints `undefined` (not `null`).

In Scala/JVM, `null.asInstanceOf[Nothing]` "succeeds" and returns a `null` value. We implemented the same behavior, but for us it is not sound wrt the IR type system, because we actually have a `nothing` type. And if we have an expression `nothing`, our type system means that the expression *cannot* complete successfully, an assumption we use for local dead code elimination and other reasoning.

So, we should make the cast of `null.asInstanceOf[Nothing]` throw an `UndefinedBehaviorError` of `ClassCastException`. Note that it should *always* throw a UBE even with compliant asInstanceOfs. With unchecked, casts are erased anyway so nothing changes.

sjrd:


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2128, opened 2015-12-22
Added initial implementation for MonthDay

alonsodomin:
Trying to contribute to the JDK8 time port to ScalaJS. An implementation of `MonthDay` in this PR.

### DIFF
--- a/javalib/src/main/scala/java/time/MonthDay.scala
+++ b/javalib/src/main/scala/java/time/MonthDay.scala
@@ -1,0 +1,158 @@
+package java.time
+
+import java.time.chrono.IsoChronology
+
+import scala.scalajs.js
+
+import java.time.temporal._
+
+/** Created by alonsodomin on 22/12/2015. */
+final class MonthDay private (month: Int, day: Int)
+    extends TemporalAccessor with TemporalAdjuster with Comparable[MonthDay]
+    with java.io.Serializable {
+
+  import Preconditions._
+  import ChronoField._
+
+  requireDateTime(month >= 1 && month <= 12, s"Invalid month value: $month")
+  requireDateTime(day >= 1 && day <= 31, s"Invalid day value: $day")
+
+  def isSupported(field: TemporalField): Boolean = field match {
+    case _: ChronoField => field == DAY_OF_MONTH || field == MONTH_OF_YEAR
+    case null           => false
+    case _              => field.isSupportedBy(this)
+  }
+
+  override def range(field: TemporalField): ValueRange = field match {
+    case DAY_OF_MONTH => ValueRange.of(1, getMonth().minLength(), getMonth().maxLength())
+    case _            => super.range(field)
+  }
+
+  // Implemented by TemporalAccessor
+  // def get(field: TemporalField): Int
+
+  def getLong(field: TemporalField): Long = field match {
+    case DAY_OF_MONTH  => day
+    case MONTH_OF_YEAR => month
+
+    case _: ChronoField =>
+      throw new UnsupportedTemporalTypeException(s"Field not supported: $field")
+
+    case _ => field.getFrom(this)
+  }
+
+  def getMonthValue(): Int = month
+
+  def getMonth(): Month = Month.of(month)
+
+  def getDayOfMonth(): Int = day
+
+  override def equals(other: Any): Boolean = other match {
+    case that: MonthDay =>
+      that.getMonthValue == getMonthValue() && that.getDayOfMonth == getDayOfMonth()
+
+    case _ => false
+  }
+
+  override def hashCode(): Int = {
+    val state = Seq(month, day)
+    state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+  }
+
+  def isValidYear(year: Int): Boolean = {
+    if (month == 2 && day == 29)
+      IsoChronology.INSTANCE.isLeapYear(year)
+    else
+      true
+  }
+
+  def `with`(month: Month): MonthDay = {
+    require(month != null, s"'month' can not be null")
+    if (month == getMonth()) {
+      this
+    } else {
+      val dayOfMonth = js.Math.min(day, month.maxLength())
+      MonthDay.of(month, dayOfMonth)
+    }
+  }
+
+  def withMonth(month: Int): MonthDay = `with`(Month.of(month))
+
+  def withDayOfMonth(dayOfMonth: Int): MonthDay = {
+    requireDateTime(dayOfMonth <= getMonth().maxLength(),
+        s"The day $dayOfMonth is invalid for mont $month")
+    if (dayOfMonth == day) this
+    else MonthDay.of(month, dayOfMonth)
+  }
+
+  // Not implemented
+  // def query[R](query: TemporalQuery[R]): R
+
+  def adjustInto(temporal: Temporal): Temporal = {
+    val temporalWithMonth = temporal.`with`(MONTH_OF_YEAR, month)
+    val dayOfMonth: Int = js.Math.min(day,
+        temporalWithMonth.range(DAY_OF_MONTH).getMaximum.toInt)
+    temporalWithMonth.`with`(DAY_OF_MONTH, dayOfMonth)
+  }
+
+  def atYear(year: Int): LocalDate =
+    LocalDate.of(year, month, if (isValidYear(year)) day else 28)
+
+  def compareTo(that: MonthDay): Int = {
+    if (month == that.getMonthValue) day - that.getDayOfMonth
+    else month - that.getMonthValue
+  }
+
+  def isAfter(that: MonthDay): Boolean = compareTo(that) > 0
+
+  def isBefore(that: MonthDay): Boolean = compareTo(that) < 0
+
+  override def toString(): String = f"--$month%02d-$day%02d"
+
+  // Not implemented
+  // def format(formatter: DateTimeFormatter): String
+
+}
+
+object MonthDay {
+
+  import Preconditions._
+  import ChronoField._
+
+  def now(): MonthDay = {
+    val date = LocalDate.now()
+    of(date.getMonthValue, date.getDayOfMonth)
+  }
+
+  // Not implemented
+  // def now(zoneId: ZoneId): MonthDay
+
+  // Not implemented
+  // def now(clock: Clock): MonthDay
+
+  def of(month: Month, dayOfMonth: Int): MonthDay = {
+    if (month == null)
+      throw new NullPointerException("month")
+    DAY_OF_MONTH.checkValidValue(dayOfMonth)
+    requireDateTime(dayOfMonth <= month.maxLength(),
+        s"The day $dayOfMonth is invalid for month $month")
+    new MonthDay(month.getValue, dayOfMonth)
+  }
+
+  def of(month: Int, dayOfMonth: Int): MonthDay = {
+    MONTH_OF_YEAR.checkValidIntValue(month)
+    of(Month.of(month), dayOfMonth)
+  }
+
+  def from(accessor: TemporalAccessor): MonthDay = accessor match {
+    case md: MonthDay => md
+    case _ => of(accessor.get(MONTH_OF_YEAR), accessor.get(DAY_OF_MONTH))
+  }
+
+  // Not implemented
+  // def parse(text: CharSequence): MonthDay
+
+  // Not implemented
+  // def parse(text: CharSequence, formatter: DateTimeFormatter): MonthDay
+
+}
--- a/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
+++ b/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
@@ -1,0 +1,208 @@
+package org.scalajs.testsuite.javalib.time
+
+import java.time.chrono.IsoChronology
+import java.time.temporal.{TemporalField, UnsupportedTemporalTypeException, ValueRange, ChronoField}
+import java.time.{DateTimeException, LocalDate, Month, MonthDay}
+
+import org.junit.Assert._
+import org.junit.Test
+import org.scalajs.testsuite.utils.AssertThrows._
+
+/** Created by alonsodomin on 22/12/2015. */
+class MonthDayTest extends TemporalAccessorTest[MonthDay] {
+  import ChronoField._
+
+  final val min = MonthDay.of(Month.JANUARY, 1)
+  final val max = MonthDay.of(Month.DECEMBER, 31)
+  final val leapMonth = MonthDay.of(Month.FEBRUARY, 29)
+
+  val samples = Month.values().map(month => (month, month.minLength(), month.maxLength())).flatMap {
+    case (month, minDay, maxDay) =>
+      if (minDay != maxDay) Seq(MonthDay.of(month, 1), MonthDay.of(month, minDay), MonthDay.of(month, maxDay))
+      else Seq(MonthDay.of(month, 1), MonthDay.of(month, minDay))
+  }.toSeq
+
+  val yearSamples = 2000 to 2020
+  def leapYear(year: Int): Boolean = IsoChronology.INSTANCE.isLeapYear(year)
+
+  override def isSupported(field: ChronoField): Boolean =
+    field == ChronoField.MONTH_OF_YEAR || field == ChronoField.DAY_OF_MONTH
+
+  override def expectedRangeFor(accessor: MonthDay, field: TemporalField): ValueRange = {
+    field match {
+      case DAY_OF_MONTH =>
+        ValueRange.of(1, accessor.getMonth.minLength(), accessor.getMonth.maxLength())
+
+      case _ =>
+        super.expectedRangeFor(accessor, field)
+    }
+  }
+
+  @Test def getLong(): Unit = {
+    assertEquals(1L, min.getLong(MONTH_OF_YEAR))
+    assertEquals(1L, min.getLong(DAY_OF_MONTH))
+
+    assertEquals(12L, max.getLong(MONTH_OF_YEAR))
+    assertEquals(31L, max.getLong(DAY_OF_MONTH))
+  }
+
+  @Test def getMonthValue(): Unit = {
+    assertEquals(1, min.getMonthValue)
+    assertEquals(12, max.getMonthValue)
+  }
+
+  @Test def getDayOfMonth(): Unit = {
+    assertEquals(1, min.getDayOfMonth)
+    assertEquals(31, max.getDayOfMonth)
+  }
+
+  @Test def getMonth(): Unit = {
+    assertEquals(Month.JANUARY, min.getMonth)
+    assertEquals(Month.DECEMBER, max.getMonth)
+  }
+
+  @Test def isValidYear(): Unit = {
+    for {
+      t <- samples
+      y <- yearSamples
+    } {
+      if (t == leapMonth) {
+        assertEquals(leapYear(y), t.isValidYear(y))
+      } else {
+        assertTrue(t.isValidYear(y))
+      }
+    }
+  }
+
+  @Test def `with`(): Unit = {
+    assertEquals(min, min.`with`(Month.JANUARY))
+    assertEquals(MonthDay.of(Month.FEBRUARY, 1), min.`with`(Month.FEBRUARY))
+    assertEquals(max, max.`with`(Month.DECEMBER))
+    assertEquals(MonthDay.of(Month.NOVEMBER, 30), max.`with`(Month.NOVEMBER))
+    assertEquals(MonthDay.of(Month.FEBRUARY, 29), max.`with`(Month.FEBRUARY))
+  }
+
+  @Test def withMonth(): Unit = {
+    assertEquals(min, min.withMonth(1))
+    assertEquals(MonthDay.of(Month.FEBRUARY, 1), min.withMonth(2))
+    assertEquals(max, max.withMonth(12))
+    assertEquals(MonthDay.of(Month.NOVEMBER, 30), max.withMonth(11))
+    assertEquals(MonthDay.of(Month.FEBRUARY, 29), max.withMonth(2))
+
+    for (t <- samples) {
+      expectThrows(classOf[DateTimeException], t.withMonth(Int.MinValue))
+      expectThrows(classOf[DateTimeException], t.withMonth(Int.MaxValue))
+      expectThrows(classOf[DateTimeException], t.withMonth(0))
+    }
+  }
+
+  @Test def withDayOfMonth(): Unit = {
+    assertEquals(min, min.withDayOfMonth(1))
+    assertEquals(MonthDay.of(Month.JANUARY, 31), min.withDayOfMonth(31))
+
+    assertEquals(max, max.withDayOfMonth(31))
+    assertEquals(MonthDay.of(Month.DECEMBER, 1), max.withDayOfMonth(1))
+
+    for (t <- samples) {
+      expectThrows(classOf[DateTimeException], t.withDayOfMonth(Int.MinValue))
+      expectThrows(classOf[DateTimeException], t.withDayOfMonth(Int.MaxValue))
+      expectThrows(classOf[DateTimeException], t.withDayOfMonth(t.getMonth.maxLength() + 1))
+    }
+  }
+
+  @Test def adjustInto(): Unit = {
+    // Intentionally using a leap year here to be able to test the full sample
+    val leapYearDate = LocalDate.of(2016, 1, 1)
+    for (t <- samples) {
+      val expectedDate = LocalDate.of(leapYearDate.getYear,
+          t.getMonthValue, t.getDayOfMonth)
+      assertEquals(expectedDate, t.adjustInto(leapYearDate))
+    }
+
+    val nonLeapYearDate = LocalDate.of(2015, 1, 1)
+    assertEquals(LocalDate.of(2015, 2, 28), leapMonth.adjustInto(nonLeapYearDate))
+  }
+
+  @Test def atYear(): Unit = {
+    val years = Seq(-999999999, 0, 1, 999999999)
+    for {
+      y <- years
+      t <- samples if !(t.getMonthValue == 2 && t.getDayOfMonth == 29)
+    } {
+      assertEquals(LocalDate.of(y, t.getMonthValue, t.getDayOfMonth), t.atYear(y))
+    }
+
+    val invalidYears = Seq(Int.MinValue, -1000000000, 1000000000, Int.MaxValue)
+    for (t <- samples; y <- invalidYears) {
+      expectThrows(classOf[DateTimeException], t.atYear(y))
+    }
+
+    for (y <- yearSamples) {
+      val expectedDay = if (leapYear(y)) 29 else 28
+      assertEquals(LocalDate.of(y, 2, expectedDay), leapMonth.atYear(y))
+    }
+  }
+
+  @Test def compareTo(): Unit = {
+    assertTrue(min.compareTo(min) == 0)
+    assertTrue(min.compareTo(max) < 0)
+    assertTrue(max.compareTo(min) > 0)
+    assertTrue(max.compareTo(max) == 0)
+  }
+
+  @Test def isAfter(): Unit = {
+    assertFalse(min.isAfter(min))
+    assertFalse(min.isAfter(max))
+    assertTrue(max.isAfter(min))
+    assertFalse(max.isAfter(max))
+  }
+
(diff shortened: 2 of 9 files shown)


sjrd:
Thanks. Since, AFAICT, you're a first-time contributor, could please confirm that you have not looked at the sources of OpenJDK/OracleJDK to implement this?

alonsodomin:
Yes, I can confirm I didn't look at the official sources.

sjrd:
Jenkins, test this please

nicolasstucki:
@alonsodomin you will need to add some tests in `testSuite/shared/require-jdk8/org/scalajs/testsuite/javalib/time`.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import scala.scalajs.js
+
+import java.time.temporal._
+
+/**
+  * Created by alonsodomin on 22/12/2015.
+  */
+final class MonthDay private (month: Int, day: Int)
+  extends TemporalAccessor with TemporalAdjuster with Comparable[MonthDay] with java.io.Serializable {
```
nicolasstucki: Put two more spaces before the extends and break the line before 80 characters:

```scala
final class MonthDay private (month: Int, day: Int)
    extends TemporalAccessor with TemporalAdjuster with Comparable[MonthDay] 
    with java.io.Serializable {

  import ...
```

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+    state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+  }
+
+  // TODO
+  // def isValidYear(year: Int): Boolean
+
+  def `with`(month: Month): MonthDay = {
+    require(month != null, s"'month' can not be null")
+    val dayOfMonth = js.Math.min(day, month.maxLength())
+    MonthDay.of(month, dayOfMonth)
```
nicolasstucki: Return `this` if `month == this.month`.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  def withMonth(month: Int): MonthDay = `with`(Month.of(month))
+
+  def withDayOfMonth(dayOfMonth: Int): MonthDay = {
+    requireDateTime(dayOfMonth <= getMonth().maxLength(), s"The day $dayOfMonth is invalid for mont $month")
+    MonthDay.of(month, dayOfMonth)
```
nicolasstucki: Similar but on day.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+    if (month == that.getMonthValue) day - that.getDayOfMonth
+    else month - that.getMonthValue
+  }
+
+  def isAfter(that: MonthDay): Boolean = compareTo(that) > 0
+
+  def isBefore(that: MonthDay): Boolean = compareTo(that) < 0
+
+  override def toString(): String = f"--$month%02d-$day%02d"
+
```
nicolasstucki: Remove this blank line.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+  def withDayOfMonth(dayOfMonth: Int): MonthDay = {
+    requireDateTime(dayOfMonth <= getMonth().maxLength(), s"The day $dayOfMonth is invalid for mont $month")
+    MonthDay.of(month, dayOfMonth)
+  }
+
+  // Not implemented
+  // def query[R](query: TemporalQuery[R]): R
+
+  def adjustInto(temporal: Temporal): Temporal = temporal.`with`(MONTH_OF_YEAR, month).`with`(DAY_OF_MONTH, day)
```
nicolasstucki: Line is too long.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,130 @@
...
+  // Not implemented
+  // def now(zoneId: ZoneId): MonthDay
+
+  // Not implemented
+  // def now(clock: Clock): MonthDay
+
+  def of(month: Month, dayOfMonth: Int): MonthDay = {
+    require(month != null, s"'month' can not be null")
+    DAY_OF_MONTH.checkValidValue(month.getValue)
+    requireDateTime(dayOfMonth <= month.maxLength(), s"The day $dayOfMonth is invalid for month $month")
```
nicolasstucki: Line is too long.

nicolasstucki:
That is all.

alonsodomin:
Comments addressed, please review

nicolasstucki:
Jenkins, retest this please.

alonsodomin:
Fixed failing test

alonsodomin:
Can anyone review the fix please?

sjrd:
Jenkins, retest this please. 

Note that it's the holidays. Expect the reviews to be delayed longer than usual.

sjrd:
Jenkins, retest this please

alonsodomin:
It seems that a few tests failed in previous build when running against the JDK implementation. Made some changes to code to align it with the expected behaviour in the JDK so feel free to trigger a build when you have time. Thanks.

nicolasstucki:
Jenkins, retest this please.

alonsodomin:
Do we know what made that last build run for so long? I couldn't look at the Jenkins logs straight away and now it's not easy to find the actual worker for the matrix build that may have some information about the build...

sjrd:
Jenkins, retest this please

sjrd:
On that day several builds ill-behaved. So it was most likely not specific to this PR.

alonsodomin:
Thanks, now I see the failure, there is diferent behaviour of `atYear` between my impl and the JDK one. Just added a new commit to fix it.

nicolasstucki:
Jenkins, retest this please.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+  def isValidYear(year: Int): Boolean = {
+    if (month == 2 && day == 29)
+      IsoChronology.INSTANCE.isLeapYear(year)
+    else
+      true
+  }
+
+  def `with`(month: Month): MonthDay = {
+    require(month != null, s"'month' can not be null")
+    if (month == getMonth()) this
```
nicolasstucki: Put the `{}` on the `if` because the `else` has them.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+  def getMonthValue(): Int = month
+
+  def getMonth(): Month = Month.of(month)
+
+  def getDayOfMonth(): Int = day
+
+  override def equals(other: Any): Boolean = other match {
+    case that: MonthDay =>
+      that.getMonthValue == getMonthValue() && that.getDayOfMonth == getDayOfMonth()
+    case _ => false
```
nicolasstucki: Add a line before this `case _ =>`

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+      val dayOfMonth = js.Math.min(day, month.maxLength())
+      MonthDay.of(month, dayOfMonth)
+    }
+  }
+
+  def withMonth(month: Int): MonthDay = `with`(Month.of(month))
+
+  def withDayOfMonth(dayOfMonth: Int): MonthDay = {
+    requireDateTime(dayOfMonth <= getMonth().maxLength(),
+      s"The day $dayOfMonth is invalid for mont $month")
```
nicolasstucki: Please add 4 spaces (double indentation) when continuing with parameters.
```scala
requireDateTime(dayOfMonth <= getMonth().maxLength(),
    s"The day $dayOfMonth is invalid for mont $month")
```

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+    else MonthDay.of(month, dayOfMonth)
+  }
+
+  // Not implemented
+  // def query[R](query: TemporalQuery[R]): R
+
+  def adjustInto(temporal: Temporal): Temporal = {
+    val temporalWithMonth = temporal.`with`(MONTH_OF_YEAR, month)
+    val dayOfMonth: Int = js.Math.min(day,
+      temporalWithMonth.range(DAY_OF_MONTH).getMaximum.toInt)
```
nicolasstucki: Same

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+    temporalWithMonth.`with`(DAY_OF_MONTH, dayOfMonth)
+  }
+
+  def atYear(year: Int): LocalDate = LocalDate.of(year, month,
+    if (isValidYear(year)) day else 28
```
nicolasstucki: Reformat to:
```scala
def atYear(year: Int): LocalDate = 
  LocalDate.of(year, month, if (isValidYear(year)) day else 28)
```

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,159 @@
...
+
+  // Not implemented
+  // def now(clock: Clock): MonthDay
+
+  def of(month: Month, dayOfMonth: Int): MonthDay = {
+    if (month == null)
+      throw new NullPointerException("month")
+    DAY_OF_MONTH.checkValidValue(dayOfMonth)
+    requireDateTime(dayOfMonth <= month.maxLength(),
+      s"The day $dayOfMonth is invalid for month $month")
```
nicolasstucki: Double indentation.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+  final val leapMonth = of(Month.FEBRUARY, 29)
+
+  val samples = Month.values().map(month => (month, month.minLength(), month.maxLength())).flatMap {
+    case (month, minDay, maxDay) =>
+      if (minDay != maxDay) Seq(of(month, 1), of(month, minDay), of(month, maxDay))
+      else Seq(of(month, 1), of(month, minDay))
+  }.toSeq
+
+  val yearSamples = 2000 to 2020
+  val leapYearLifted = (year: Int) => IsoChronology.INSTANCE.isLeapYear(year)
```
nicolasstucki: Use a `def`

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  override def isSupported(field: ChronoField): Boolean =
+    field == ChronoField.MONTH_OF_YEAR || field == ChronoField.DAY_OF_MONTH
+
+  @Test override def test_range(): Unit = {
```
nicolasstucki: Remove the `test_` from the names of the test methods.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+      if (isSupported(field)) {
+        assertEquals(expected, t.range(field))
+      } else {
+        expectThrows(classOf[UnsupportedTemporalTypeException], t.range(field))
+      }
+    }
+  }
+
+  @Test def test_getLong(): Unit = {
+    assertEquals(min.getLong(MONTH_OF_YEAR), 1L)
```
nicolasstucki: All your `assertEquals` receive the parameters in the wrong order. It is defined as `assertEquals(expected, actual)`.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+      expectThrows(classOf[DateTimeException], t.withDayOfMonth(Int.MaxValue))
+      expectThrows(classOf[DateTimeException], t.withDayOfMonth(t.getMonth.maxLength() + 1))
+    }
+  }
+
+  @Test def test_adjustInto(): Unit = {
+    // Intentionally using a leap year here to be able to test the full sample
+    val leapYearDate = LocalDate.of(2016, 1, 1)
+    for (t <- samples) {
+      assertEquals(LocalDate.of(leapYearDate.getYear, t.getMonthValue, t.getDayOfMonth), t.adjustInto(leapYearDate))
```
nicolasstucki: line is too long.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    for (y <- yearSamples) {
+      val expectedDay = if (leapYearLifted(y)) 29 else 28
+      assertEquals(LocalDate.of(y, 2, expectedDay), leapMonth.atYear(y))
+    }
+  }
+
+  @Test def test_compareTo(): Unit = {
+    assert(min.compareTo(min) == 0)
```
nicolasstucki: Use `assertEquals`.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+    assert(min.compareTo(max) < 0)
```
nicolasstucki: Use `assertTrue` from `org.junit.Assert`

nicolasstucki: Same for all other `assert(boolean)`

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  @Test def test_isAfter(): Unit = {
+    assert(!min.isAfter(min))
+    assert(!min.isAfter(max))
+    assert(max.isAfter(min))
+    assert(!max.isAfter(max))
+  }
+
+  @Test def test_isBefore(): Unit = {
+    assert(!min.isBefore(min))
```
nicolasstucki: Use `assertFalse`

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,218 @@
...
+  @Test def test_of(): Unit = {
+    expectThrows(classOf[DateTimeException], of(Int.MinValue, 1))
+    expectThrows(classOf[DateTimeException], of(Int.MaxValue, 1))
+
+    assertEquals(min, of(1, 1))
+    assertEquals(max, of(12, 31))
+  }
+
+  @Test def test_from(): Unit = {
+    assert(from(min) eq min)
```
nicolasstucki: Use `assertSame`.
```scala
assertSame(min, from(min))
```

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
nicolasstucki: Actually not all of them are inverted, only most of them.

nicolasstucki:
You can also squash all the commits into one.

nicolasstucki:
That is all.

javalib/src/main/scala/java/time/MonthDay.scala
alonsodomin: Do you mean the `if` in the `isValidYear` method? Neither `if` or `else` clauses have braces already, I can add the braces to both if that is what you mean...

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
alonsodomin: Won't that break the "naming consistency"? I mean, I not a fan myself of prefixing the methods with `test_` but I chose to do so because all other JUnit tests in the same package use that same convention

nicolasstucki: Those will eventually change to this same format.

javalib/src/main/scala/java/time/MonthDay.scala
nicolasstucki: No, I mean in the `with` method. In this case the `else` has braces and therefore the if also needs them.

You currently have:
```scala
if (month == getMonth()) this
else {
  val dayOfMonth = js.Math.min(day, month.maxLength())
  MonthDay.of(month, dayOfMonth)
}
```

But it should be:
You currently have:
```scala
if (month == getMonth()) {
  this
} else {
  val dayOfMonth = js.Math.min(day, month.maxLength())
  MonthDay.of(month, dayOfMonth)
}
```

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
alonsodomin: Fair enough. Although I need keep the prefix in `test_range` because it's overriding a method from the parent test class, otherwise the test would fail.

nicolasstucki: I see.

nicolasstucki: Why do tests fail in that case?

nicolasstucki: You should probably use this implementation of `test_range` in `TemporalAccessorTest` rather than overriding it. Or a version that tests all of these cases.

alonsodomin: Well, I had to override the `range` method defined in `TemporalAccessor` because the test was originally failing when testing the output of that `range` method against the JDK MonthDay class.

The scenario is, the default `ValueRange` for the `TemporalField` `DAY_OF_MONTH` gives a minimum value of `1`, a minimum length of `28` and a maximum length of `31`. So the `range` method in `TemporalAccessor` will return always that value range regardless of the actual month value stored in the `TemporalAccessor` concrete implementation.

This means that when asked for the `DAY_OF_MONTH` range of a `MonthDay` representing any day of January, the `range` method will return `ValueRange.of(1, 28, 31)`. Accordingly it will return the same `ValueRange` when asked for the range of that same field for a `MonthDay` representing any day of Feburary.

However the JDK seems to behave differently as the test failed with the first `MonthDay` sample, saying that it was expecting a `ValueRange.of(1, 28, 31)` but it got a `ValueRange.of(1, 31)` instead. Similar behaviour happens with any February sample but in this case, whilst the `test_range` method in the `TemporalAccessorTest` class was expecting a `ValueRange.of(1, 28, 31)` it got a `ValueRange.of(1, 28, 29)`.

Due to this I decided to override the `range` method to make it aware of the actual month represented by the `MonthDay` instance, and this forced me to override the `test_range` method in the test as well.

nicolasstucki: Ok, but still you should this test should be done in the `TemporalAccessorTest.test_range`. Then you will need to special case this behaviour in `TemporalAccessorTest.test_range` using your definition of `expected` and with (if needed) an `.isInstanceOf[MonthDay]` test on the sample.

I imagine something like this should work in  `TemporalAccessorTest`:
```scala
@Test def range(): Unit = {
  for {
    accessor <- samples
    field <- ChronoField.values
  } {
    if (accessor.isSupported(field)) {
      val expected = {
        // Put smallish comment on why this is a special case.
        if (field == DAY_OF_MONTH && accessor.isInstanceOf[MonthDay]) 
          ValueRange.of(1, t.getMonth.minLength, t.getMonth.maxLength)
        else 
          field.range
      }
      assertEquals(expected, accessor.range(field))
    } else {
      expectThrows(classOf[UnsupportedTemporalTypeException], accessor.range(field))
    }
  }
}
```

alonsodomin: Just wondering how many _special_ cases would leak into that impl. `LocalDateTest`, for example, also overrides that method to add _a few_ special cases, not only for the `DAY_OF_MONTH` field but for other fields as well. Merging all the possible special cases in the parent class as more classes are added sounds to me like a maintenance nightmare.

If we want to keep it DRY then the base method should probably be refactored to consume some sort of function `TemporalField => ValueRange` that will return the expected range for that specific field in the concrete test implementation. That function could perfectly be a method in `TemporalTest` that implementors need to override to provide the special cases.

```scala
def expectedRangeFor(field: TemporalField): ValueRange = field.range

@Test def range(): Unit = {
  for {
    accessor <- samples
    field <- ChronoField.values
  } {
    if (accessor.isSupported(field)) {
      val expected = expectedRangeFor(field)
      assertEquals(expected, accessor.range(field))
    } else {
      expectThrows(classOf[UnsupportedTemporalTypeException], accessor.range(field))
    }
  }
}
```

nicolasstucki: I agree. Your solution with `expectedRangeFor` is what we need.

alonsodomin:
As discussed previously, the `TemporalAccessorTest`'s `range` method has been refactored as `final` and using a `expectedRangeFor` method to allow specific test class implementations hook in special cases for the ranges of the temporal fields.

I still need to squash all the commits into a single one but this should be ready for review.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/TemporalAccessorTest.scala
```diff
@@ -7,8 +7,8 @@ import org.junit.Test
 import org.junit.Assert._
 import org.scalajs.testsuite.utils.AssertThrows._
 
-abstract class TemporalAccessorTest {
-  val samples: Seq[TemporalAccessor]
+abstract class TemporalAccessorTest[A <: TemporalAccessor] {
+  val samples: Seq[A]
```
nicolasstucki: I think that adding this `A <: TemporalAccessor` is overkill. You could just use `val samples: Seq[TemporalAccessor]` and then make it more specific in each implementation (as `Seq` is covariant).

alonsodomin: Yeah, thought the same myself... I didn't really like to add a type parameter to the test class but the reason for that `A` was driven more by the new:

```scala
def expectedRangeFor(accessor: A, field: TemporalField): ValueRange = field.range()
```

The _expected range_ for a particular `TemporalField` depends not only on the field, but in the actual sample instance being tested at a particular point. And then the implementation of the _special cases_ depends on fields/methods declared on the `TemporalAccessor` concrete implementation.

This is a snippet from `MonthDayTest`:

```scala
override def expectedRangeFor(accessor: MonthDay, field: TemporalField): ValueRange = {
    field match {
      case DAY_OF_MONTH =>
        ValueRange.of(1, accessor.getMonth.minLength(), accessor.getMonth.maxLength())

      case _ =>
        super.expectedRangeFor(accessor, field)
    }
  }
```

This is a snippet from `LocalDateTest`:

```scala
override def expectedRangeFor(accessor: LocalDate, field: TemporalField): ValueRange = {
    field match {
      case DAY_OF_MONTH => ValueRange.of(1, accessor.lengthOfMonth)
      case DAY_OF_YEAR  => ValueRange.of(1, accessor.lengthOfYear)

      case ALIGNED_WEEK_OF_MONTH =>
        ValueRange.of(1, if (accessor.lengthOfMonth > 28) 5 else 4)

      case YEAR_OF_ERA =>
        val maxYear = if (accessor.getEra == IsoEra.CE) 999999999 else 1000000000
        ValueRange.of(1, maxYear)

      case _ =>
        super.expectedRangeFor(accessor, field)
    }
  }
```

None of the methods `MonthDay.getMonth`, `LocalDate.lengthOfMonth`, `LocalDate.lengthOfYear` are accessible from the `TemporalAccessor` type as they are implemented by the different concrete classes. So I ended up adding the type parameter to basically avoid an `accessor.asInstanceOf[MonthDay]` cast in the concrete test classes.

nicolasstucki: Oh, now I see. Then it is fine. You could give it a more meaningful name than `A` like `TA` or `TempAcc`. 

You can also go ahead and squash the commits.

alonsodomin: Sure, will rename it. Since I'm also removing the `test_` prefix from the `test_range` method in the `TemporalAccessorTest` class, probably it would make sense to remove that prefix in the renaming methods (including the abstract `TemporalTest` as well), just to help _keep the house clean_ if that sounds good to you too.

nicolasstucki: Yes, it does.

alonsodomin:
Squashed the commits, for some reason I couldn´t do it in just one most likely because whilst this branch was alive I merged the upstream master into it... stupid thing to do as I now have a branch with conflicts...

nicolasstucki:
Jenkins, retest this please

sjrd:
The commit message should be amended.

alonsodomin:
Removed right now the bits that were referencing commits from other branches. Is that what you meant or should I rewrite whole commit message as a one-liner?

nicolasstucki:
You should write a one liner in this case. Make it `Add initial implementation for MonthDay.`.

alonsodomin:
Amended

nicolasstucki:
Jenkins, retest this please.

javalib/src/main/scala/java/time/MonthDay.scala
```diff
@@ -0,0 +1,160 @@
...
+
+import java.time.chrono.IsoChronology
+
+import scala.scalajs.js
+
+import java.time.temporal._
+
+/**
+  * Created by alonsodomin on 22/12/2015.
+  */
```
nicolasstucki: Oh, I missed this formatting issue with comments, we align to the first `*` and start in the first line, i.e.:
```scala
/** Created by alonsodomin on 22/12/2015.
 */
```

nicolasstucki: Actually this one should be a one liner.
```scala
/** Created by alonsodomin on 22/12/2015. */
```

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/time/MonthDayTest.scala
```diff
@@ -0,0 +1,210 @@
...
+import java.time.temporal.{TemporalField, UnsupportedTemporalTypeException, ValueRange, ChronoField}
+import java.time.{DateTimeException, LocalDate, Month, MonthDay}
+
+import org.junit.Assert._
+import org.junit.Test
+import org.scalajs.testsuite.utils.AssertThrows._
+
+/**
+  * Created by alonsodomin on 22/12/2015.
+  */
```
nicolasstucki: Same

nicolasstucki:
That is all. Otherwise LGTM.

alonsodomin:
No prob, both comments updated

nicolasstucki:
Oh, could you just squash the two commits. This is the last thing to do.

alonsodomin:
oh, sorry. No prob, commits squashed

nicolasstucki:
Jenkins, retest this please.

Issue #1419, opened 2014-12-24
IRChecker does not verify method existence on receiver


gzm0:
It does check existence for fields.

sjrd:
Now that #3831 is merged, we can delay this to post-v1.0.0.

gzm0:
After #4410, it is IMO clear that this is the Analyzers job (and IIUC it does this already). So unless you object, @sjrd, I'd propose to close this as wontfix.

sjrd:
Sounds good to me. 😉

PR #2130, opened 2015-12-22
Redesign constructors

sjrd:
~~CI only.~~

This is a relatively major redesign of the specifications of constructors at the IR level. There are two parts:

* Make them return `NoType` at the IR level. The `return this` necessary to chain them with `new` is added at desugar time.
* Do not consider constructors as inherited anymore (as is the case on the JVM).

The changes are accompanied by appropriate backward binary compatibility hacks.

The second change will be instrumental to @tOverney's work on merging `init` methods into the JS constructors when possible.


sjrd:
Now ready for review.
Review by @gzm0, please.

Also /cc @tOverney, as you will probably have to rebase your branch on top of this to get full soundness.

sjrd:
Note that the CI failures in the intermediate commits are spurious.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -720,14 +729,27 @@ object Serializers {
...
-            result
+            result1
+          }
+          if (useHacks065 && result2.resultType != NoType &&
+              isConstructorName(result2.name.name)) {
+            val patchedBody = result2.body match {
+              case This()                 => Skip()(result2.body.pos)
+              case Block(stats :+ This()) => Block(stats)(result2.body.pos)
+              case body                   => body
+            }
```
gzm0: Why do you need to do this? Just to decrease JS size, or is there another reason?

sjrd: It is "just" to reduce JS size, but quite dramatically so. It is necessary for elidable module accessors to stay elidable.

gzm0: Ok. Maybe drop a comment. It is hard to understand why you need to do this (in a hack), but using the escape hatch (just not patching the body) is fine as well.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/StackTraceTest.scala
```diff
@@ -69,26 +69,34 @@ object StackTraceTest extends JasmineTest {
     when("nodejs").
     unlessAny("fullopt-stage", "strong-mode").
     it("decode class name and method name") {
-      verifyClassMethodNames("Foo" -> "f") {
-        new Foo().f(25)
-      }
+      val Error = js.constructorOf[js.Error]
+      val oldStackTraceLimit = Error.stackTraceLimit
+      Error.stackTraceLimit = 20
```
gzm0: What makes this necessary?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -283,16 +289,51 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+        if (!initialLink) {
+          createNonExistentMethod(ctorName)
+        } else {
+          val inherited = lookupMethod(ctorName)
+          if (inherited.owner eq this) {
+            // Can happen only for non-existent constructors, at this point
+            assert(inherited.nonExistent)
+            inherited
+          } else {
+            val nonExistent = !initialLink
```
gzm0: Fruitless test.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -243,6 +270,43 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
         version)
   }
 
+  private def synthesizeInheritedConstructor(
+      classInfo: Analysis.ClassInfo, methodInfo: Analysis.MethodInfo,
+      getTree: TreeProvider, analysis: Analysis)(
+      implicit pos: Position): MethodDef = {
+    val encodedName = methodInfo.encodedName
```
gzm0: Consider replacing this by a constant. This code really only ever makes sense for ctors. So no point to generalize. It's just confusing.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -243,6 +270,43 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
...
+
+    @tailrec
+    def findInheritedMethodDef(ancestorInfo: Analysis.ClassInfo): MethodDef = {
+      val inherited = ancestorInfo.methodInfos(methodInfo.encodedName)
```
gzm0: Same here.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -111,6 +111,9 @@ private[optimizer] abstract class OptimizerCore(
           disableOptimisticOptimizations = true
           transformIsolatedBody(Some(myself), thisType, params, resultType, body)
       }
+      val newBody =
+        if (name.name == "init___") tryElimStoreModule(newBody1)
```
gzm0: Why does `isConstructorName` not work here? (I think I am missing something w.r.t. to ctor names).

gzm0:
I understand correctly, that you do not yet avoid emitting trivial ctors in the Emitter and you will follow up with changes that do so, right?

gzm0:
That's all

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
sjrd: Don't forget overloads. `init___` is not the only constructor name. There are plenty of them, and they're definitely not constant.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/StackTraceTest.scala
sjrd: When the optimizer is disabled, the new constructors in the chain cannot be eliminated, and caused one of the stack traces tested here to be longer. In particular, the `Baz.<init>` constructor *just* made it to the 11th depth. The test was failing because Node.js only records 10 frames by default.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: Indeed. Not to mention unused `val` ^^

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
sjrd: Oh it would *work*, but would do useless work. Since we're after `StoreModule`s, this is only relevant to the constructor odd module classes, and those are always called `init___`. It's an early filter.

sjrd:
> I understand correctly, that you do not yet avoid emitting trivial ctors in the Emitter and you will follow up with changes that do so, right?

You're right. Although, in fact, I believe @tOverney's work will be enough to get rid of virtually all the new init methods. There are surprisingly few constructors that only call super, except those that call a chain of trivial constructors up until `Object.init`. Those are now optimized down to `Skip` due do inlining. Such constructors will completely disappear with the optimization he's working on.

sjrd:
I also did not want to do that yet because the implementation will git-conflict with his progress, whereas all of this does not.

Btw, see the branch https://github.com/SebsLittleHelpers/scala-js/tree/final-classes

test-suite/js/src/test/scala/org/scalajs/testsuite/library/StackTraceTest.scala
gzm0: Ok.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
gzm0: Ah... Indeed... :P

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
gzm0: Yep... Forgot about overloads :P

gzm0:
Ok. In that case LGTM otherwise (still TODO: comment in hack, unused val in Analyzer)

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
sjrd: I must have dreamed ^^ Or it was so during a transient stage of my implementation. I just checked again, and removing this patch on the body does not cause reversi-fastopt.js to grow. In fact, without the patch it is completely unchanged wrt. the parent commit :-p So I'll remove it.

sjrd:
Updated.

gzm0:
Hold on, don't you have to modify the IRChecker as well?

sjrd:
> Hold on, don't you have to modify the IRChecker as well?

What part, and for which commit/change?

gzm0:
In "Constructors are not inherited anymore at the IR level." you are tightening the constraints on constructor calls (they now *must* exist in the target class). The `IRChecker` is past the linker (that synthesizes the missing ctors). So you could check for it.

gzm0:
Essentially [here](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/IRChecker.scala#L610), `isStatic` should be true?

gzm0:
(Which static is `isStatic` there? :P)

gzm0:
Hahahaha, #1419 anyone?

sjrd:
Exactly, that is the job of #1419. Currently, the Analyzer is actually checking for this. This is also the reason I deactivated the hack for the refiner, and I emit warnings in that case (I haven't seen any such warning).

sjrd:
>  (Which static is isStatic there? :P)

That is *static*, not "statically". So this is only for static methods. A constructor is still not a static method.

gzm0:
LGTM

gzm0:
Is this good to merge or do we need to wait on CI in #2133?

sjrd:
We don't need to catch on. The other commits will literally behave the same as in my original commit-by-commit check, since only the deserialization hack changed, and that code is only exercised during that commit.

PR #2139, opened 2015-12-27
Synthesize reflective proxies at link time.

sjrd:
~~PR in progress to tests one commit at a time. All commits are in [no-refl-proxy-in-sjsir-2](https://github.com/scala-js/scala-js/compare/master...sjrd:no-refl-proxy-in-sjsir-2)~~

Instead of relying on reflective proxies stored in sjsir files by the compiler, we synthesize them at link time, on demand. During deserialization of sjsir files, we throw away all the reflective proxies. The Analyzer identifies the reflective proxies needed to satisfy calls of reflective methods, and the Linker synthesizes the bridges.

The compiler does not need to generate reflective proxies anymore, which simplifies it a bit.

More importantly, this reduces the amount of virtually useless stuff stored in sjsir files. This improves their size, but also the time it takes to deserialize them, and the memory footprint for the deserialized reflective proxies that are never used.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
@@ -15,7 +15,7 @@ import scala.collection.mutable
 
 import org.scalajs.core.ir
 import ir.{ClassKind, Definitions, Infos}
-import Definitions.{isConstructorName, isReflProxyName}
+import Definitions._
 
 import org.scalajs.core.tools.sem._
 import org.scalajs.core.tools.javascript.{LongImpl, OutputMode}
@@ -23,14 +23,16 @@ import org.scalajs.core.tools.javascript.{LongImpl, OutputMode}
 import ScalaJSOptimizer._
 
 final class Analyzer(semantics: Semantics, outputMode: OutputMode,
-    reachOptimizerSymbols: Boolean, initialLink: Boolean) extends Analysis {
+    reachOptimizerSymbols: Boolean,
+    allowAddingSyntheticMethods: Boolean) extends Analysis {
   import Analyzer._
   import Analysis._
 
-  @deprecated("Use the overload with an explicit `initialLink` flag", "0.6.6")
+  @deprecated("Use the overload with an explicit `allowAddingSyntheticMethods` flag", "0.6.6")
   def this(semantics: Semantics, outputMode: OutputMode,
       reachOptimizerSymbols: Boolean) = {
-    this(semantics, outputMode, reachOptimizerSymbols, initialLink = true)
+    this(semantics, outputMode, reachOptimizerSymbols,
+        allowAddingSyntheticMethods = true)
   }
 
   @deprecated("Use the overload with an explicit output mode", "0.6.3")
@@ -297,7 +299,7 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
        * during the initial link. In a refiner, this must not happen anymore.
        */
       methodInfos.get(ctorName).getOrElse {
-        if (!initialLink) {
+        if (!allowAddingSyntheticMethods) {
           createNonExistentMethod(ctorName)
         } else {
           val inherited = lookupMethod(ctorName)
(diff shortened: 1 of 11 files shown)


sjrd:
Hum ... surprisingly this ends up eating *more* memory :s The toolsJS tests all failed with OOMs (or Node.js' equivalent thereof).

sjrd:
Review by @gzm0

gzm0:
Did you run the partest suite on this one?

sjrd:
I didn't, I was waiting for the regular CI to complete (there is some binary incompat in CLI to be fixed). But indeed we should definitely run a nighlty on this one.

gzm0:
Could you rename the `initialLink` option from the `Analyzer` to `allowAddSyntheticMembers`? I didn't realize you wanted to extend this to a non-hack (thinking of default methods it makes perfect sense to do so).
However, `initialLink` is a name that completely breaks abstractions. (The `Analyzer` doesn't need to know *why* you want to generate synthetic members, but *that* you want to).

sjrd:
OK. And yes, default methods should come next :p

sjrd:
Updated.

sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/3439/

sjrd:
Nightly has a couple of spurious failures, but all partest jobs have passed :-)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -350,6 +352,124 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+    }
+
+    private def methodResultType(methodName: String): ir.Types.ReferenceType = {
+      val typeName = methodName.substring(methodName.lastIndexOf("__") + 2)
+      val arrayDepth = typeName.indexWhere(_ != 'A')
+      if (arrayDepth == 0)
+        ir.Types.ClassType(typeName)
+      else
+        ir.Types.ArrayType(typeName.substring(arrayDepth), arrayDepth)
+    }
```
gzm0: Is this really necessary? Why can't you take the result type from the method? Also, what happens to primitive types?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -350,6 +352,124 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+
+      val targets = candidates.filterNot { c =>
+        val resultType = methodResultType(c.encodedName)
+        candidates.exists { other =>
+          (other ne c) &&
+          isMoreSpecific(methodResultType(other.encodedName), resultType)
+        }
+      }
+
+      targets.headOption
```
gzm0: You should maybe retain the comment why and that this is just as broken than on the JVM (i.e. `targets` may contain multiple elements here, etc.)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: Ok, I see now that you have `methodResultType` to shortcut converting primitive IR types to their `ClassType` form. Consider making a helper in the IR for this. It really seems something more fundamental to me than that we can just do some string mangling to get it done...

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -307,6 +297,63 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
...
+  private def synthesizeReflectiveProxy(
+      classInfo: Analysis.ClassInfo, methodInfo: Analysis.MethodInfo,
+      targetName: String, getTree: TreeProvider, analysis: Analysis)(
+      implicit pos: Position): MethodDef = {
+    val encodedName = methodInfo.encodedName
+
+    val targetMDef = findInheritedMethodDef(classInfo, targetName, getTree)
+
+    val targetIdent = Ident(targetMDef.name.name,
+        targetMDef.name.asInstanceOf[Ident].originalName)
```
gzm0: `targetMDef.name.asInstanceOf[Ident].copy()(pos)` (or something like that?)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: Ahhhh... What if I told you: There are no IR types :P
This is just a poor mans version of what we do in the IR checker as well. It's a bit strange we have to rely on string un-mangling, but lets leave it for now.
Please just drop a comment on what happens with things that have primitive types. (`IntType` -> `ClassType("I")`)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
```diff
@@ -307,6 +297,63 @@ final class Linker(semantics: Semantics, outputMode: OutputMode,
         inheritedMDef.hash) // over-approximation
   }
 
+  private def synthesizeReflectiveProxy(
+      classInfo: Analysis.ClassInfo, methodInfo: Analysis.MethodInfo,
+      targetName: String, getTree: TreeProvider, analysis: Analysis)(
+      implicit pos: Position): MethodDef = {
```
gzm0: I wouldn't position this at the position of the containing class def, but at the position of the method this is calling. It will give much better debugging experience when stepping.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -350,6 +352,124 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+              case _ =>
+                ancestorInfo.findProxyMatch(proxyName) match {
+                  case Some(target) =>
+                    val targetName = target.encodedName
+                    Some(ancestorInfo.createReflProxy(proxyName, targetName))
+
+                  case None =>
+                    loop(ancestorInfo.superClass)
+                }
+            }
```
gzm0: This lookup does not what we have done before:

> Cases like the upper currently fail on the JVM backend at runtime. The Scala.js backend uses the following rules for selection (which will also cause runtime failures):
 - If a proxy with the same signature (method name and parameters) exists in the superclass, no proxy is generated (proxy is inherited)
 - If no proxy exists in the superclass, a proxy is generated for the first method with matching signatures.

So IIUC, the following example will change call selection with this change:

```scala
class Base {
  def foo(x: Option[Int]): String = "a"
}

class Sub extends Base {
  def foo(x: Option[String]): Int = 1
}

val x: { def foo(x: Option[String]): Any } = new Sub()
x.foo(Some(1))
```

Before this change (if the doc is correct), the last line would succeed (inherited proxy). After this change, `Sub.foo` would be elected for proxy generation and the call would cause a `ClassCastException`.

Please test that whatever behavior we have afterwards is consistent with Scala JVM.

gzm0:
That's all. Sorry it took so long.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: What happens to primitive types is actually nothing special, here. That's always what happens for primitive types when encoded in a `ReferenceType`. This is for example used in `ClassOf` nodes. Besides, primitive types are not the only ones having a behavior different from what they would be as a full-fledged `Type`: raw JS types are also kept as `ClassType`s, whereas as `Type`s they become `Any`.

gzm0: Hmmm... So IIUC you are essentially just abusing the type case classes so you can store some stuff until it hits `isMoreSpecific`. Would the types you generate be valid on an IR level?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: I admit I did not look at the previous implementation. I just looked anew at the output on the JVM after `cleanup`, saw it used `getMethod`, and looked at its specification.

With `Base` and `Sub` defined as above, we get this funny behavior:
```scala
scala> val x: { def foo(x: Option[Int]): Any } = new Sub
scala> x.foo(Some(1))
res1: Int = 1

scala> val y: { def foo(x: Option[String]): Any } = new Sub
scala> y.foo(Some("hello"))
res2: Any = 1
```
In both cases, `Sub.foo` is called. (There is no CCE because the `Option` param is not unwrapped.) I'm pretty sure this is consistent with my implementation. I'll add a test for it.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -350,6 +352,124 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+          leftDepth == rightDepth && classIsMoreSpecific(leftBase, rightBase)
+        case (ArrayType(_, _), ClassType(ObjectClass)) =>
+          true
+        case _ =>
+          false
+      }
+    }
+
+    private def createReflProxy(proxyName: String,
+        targetName: String): MethodInfo = {
```
gzm0: I forgot to mention that: Consider asserting that we are of a sensible class type (Scala classes and modules only, no traits, etc.)

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: Yep. So in fact the previous implementation was wrong in this respect.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: I agree it's sort of an abuse, but it's an abuse we've used for a long time in `ClassOf` nodes (at least, I think there are other cases). It is not a valid type for a term, but it is a valid so-called "reference-type", which identifies a class/interface/primitive type/array type exactly as it would be identified on the JVM (without the JS-type erasure kicking in, nor `Byte`, `Short`, `Char` being erased to `int`).

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: OK. It's pretty explicit by the spec of `java.lang.Class.getMethod`, but it doesn't hurt to restate it, indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: Ah, I see. Maybe that's why I never understood the difference between `IntType` and `ClassType("I")`. So let's leave it.

Btw. IIUC another place of abuse is the `ArrayType` itself. One cannot encode `Array[Int]` without falling back to using `ClassType`s.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: Good point.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Linker.scala
sjrd: Indeed.

sjrd:
Updated.

gzm0:
LGTM

PR #2140, opened 2015-12-28
Fix #1887: Fix HTML runners of the test suite in fullOpt.

### LINKED ISSUE #1887: HTML runners of the test suite are broken in fullOpt
They currently fail one test about "Emitted classes should have a meaningful `name` property", which is not supposed to run in fullOpt mode.

sjrd:
Some tests must be disabled in fullOpt mode. From sbt, they are disabled by Jasmine tags, but in the HTML runners, the appropriate tags were not set. This commit fixes this.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2141, opened 2016-01-01
Fix #1269: Add support for default methods.

### LINKED ISSUE #1269: Default methods in the IR
They will be used by future versions of Scala and Dotty.

sjrd:
Let's see what fails at this point ...

The design at large is as follows:
* In the IR, interfaces can have concrete methods; they are default methods.
* The `Analyzer` resolves all default method lookups, as specified by the JVM (see the [invokespecial opcode specification](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.invokespecial))
  * It generates synthetic bridges in classes to the default methods
  * For `ApplyStatically` to default methods, it also generates bridges as default methods in subinterfaces when lookup resolves to an inherited default method
* After the first linking by the `Linker`, default methods can still exist in interfaces, but they are not inherited anymore. They must be explicitly called with `ApplyStatically`, with the target `cls` being an interface directly defining that default method.
* The `GenIncOptimizer` is enhanced to deal with default methods. They mostly behave like static methods, except that they have a `thisType`, and they live in a separate namespace
* The `OptimizerCore` is remarkably untouched
* The desugaring produces top-level `$f_Interface__method__signature` methods for default methods (similar to static methods) with an explicit `$thiz` parameter. `ApplyStatically` calls to default methods are transformed to call these top-level functions.

To *emit* default methods in the first place, we add an annotation `@JavaDefaultMethod`, which can be used on concrete trait methods in the JDK implementation for things that are Java default methods. The effect of that annotation is Scaladoc'umented. Briefly it creates a default method bridge to the impl class' static method, and removes mixed in methods. Its effective semantics are *very much* implementation-defined, and hence it should only be used inside the JDK as a special stop-gap hack to produce default methods.


sjrd:
@gzm0 @nicolasstucki Any comments on the design described above?

sjrd:
Note-to-self: strong mode fails with the optimizer disabled. Probably the desugaring of default methods is wrong (either definition or call site).

gzm0:
Overall design sounds good.

sjrd:
It's now complete and good to review (still need to squash of course).

Review by @gzm0. Don't review individual commits; just look at the entire diff for the PR.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -734,13 +733,36 @@ abstract class GenJSCode extends plugins.PluginComponent
         if (scalaPrimitives.isPrimitive(sym)) {
           None
         } else if (sym.isDeferred || sym.owner.isInterface) {
+          val body = if (sym.hasAnnotation(JavaDefaultMethodAnnotation)) {
+            /* For an interface method with @JavaDefaultMethod, make it a
+             * default method calling the impl class method.
+             */
```
gzm0: So you are doing this because it is hard to recover the impl class method body at this point, right? This will not be necessary once scalac emits default methods, right?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -734,13 +733,36 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+              }
+            }
+            genTraitImplApply(implMethodSym,
+                js.This()(currentClassType) :: jsParams.map(_.ref))
+          } else {
+            js.EmptyTree
+          }
           Some(js.MethodDef(static = false, methodName,
-              jsParams, currentClassType, js.EmptyTree)(
+              jsParams, toIRType(sym.tpe.resultType), body)(
```
gzm0: I do not understand this... Was the return type set to the class type??? That's just plain wrong?!?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Exactly. Not to mention some other code could still call the impl class method body *statically*, so we can't just remove it altogether. They both need to exist, one being a forwarder to the other. We might as well keep the impl class version as the full version, since it already exists as is. At the end of the day, the optimizer removes all these indirections.

When scalac will emit default methods, we'll just have to make sure *not* to do this, but just emit the default methods. The specifics will depend on how scalac gives default methods to us, so it is fruitless to try and anticipate the implementation at the moment.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Er, yes, that was just plain wrong, indeed ^^

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Ok. That's what I thought.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: In that case the IR-Checker has a pretty serious bug. :)

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
```diff
@@ -66,6 +69,11 @@ final class ScalaJSClassEmitter private (
     linkingUnit.classDefs.map(c => c.encodedName -> c).toMap
   }
 
+  private[javascript] def isInterface(className: String): Boolean = {
+    linkingUnit != null &&
+    linkedClassByName(className).kind == ClassKind.Interface
```
gzm0: Is a class recreated in the incremental optimizer if just its kind changes?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: I thought about that, and in fact it doesn't. But it's a limitation of our approach to IR checking. We only IR-check after linking (for good reasons, but still). But after linking, there are no abstract methods anymore, since abstract methods cannot be reached (by definition). Therefore, the IR checker never even sees these methods, and therefore does not error out on them.

Addressing this will require rethinking the IR checker strategy.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
sjrd: Yes, all sorts of stuff is completely shuffled around when a thing switches from class to interface and vice versa. In the incremental optimizer, classes and interfaces live in completely different namespaces, so they are necessarily recreated.

I can see, in theory, a failure, though. In theory, a third-party class would have an `ApplyStatically` to `C.foo`. If `C` switches from class to interface or the other way around, the IR does not prescribe that this call needs to change. Therefore, the result could be broken. However, in practice, only subclasses and subtraits can call statically (due to the rules of the Scala *language*), and in scalac, it makes an awful lot of difference internally whether you extend a class or mix in a trait/interface, which means it will always be recompiled, leading to a different input IR anyway.

There are 3 possible ways forward about this:
* Do nothing, and accept the theoretical flaw
* Throw away the emitter cache when an entity switches from class to interface or the other way around (should not happen often in practical situations)
* Use a different IR node for statically calling a default method, for example `ApplyDefaultStatically`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: OK.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/ScalaJSClassEmitter.scala
gzm0: I would go with solution 1 +  clearly document the assumptions the emitter makes and how they are satisfied. Well might need a more principled way of addressing this in the future anyways.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/JSDesugaring.scala
```diff
@@ -1576,8 +1576,24 @@ private[javascript] object JSDesugaring {
...
-          js.Apply(fun DOT "call", (receiver :: args) map transformExpr)
+          val className = cls.className
+          val transformedArgs = (receiver :: args) map transformExpr
+
+          if (classEmitter.isInterface(className)) {
+            val Ident(methodName, origName) = method
+            if (isStrongMode) {
+              js.Apply(js.DotSelect(envField("c", className),
+                  js.Ident("$f_" + methodName, origName)(method.pos)),
+                  transformedArgs)
```
gzm0: Why do we put this in the static namespace of the class in strong mode? Just because?
It seems that in strong mode we now have another collision domain here (default implementation / static method) that we do not have in non-strong mode.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analysis.scala
```diff
@@ -76,6 +76,7 @@ object Analysis {
     // TODO Get rid of InheritedConstructor when we can break binary compat
     final case object InheritedConstructor extends MethodSyntheticKind
     final case class ReflectiveProxy(target: String) extends MethodSyntheticKind
+    final case class DefaultBridge(targetInterface: String) extends MethodSyntheticKind
```
gzm0: This might actually be a good spot to document the shapes of all these synthetic methods (in a later commit / PR). Just had that thought.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -336,8 +336,9 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
     }
 
     def tryLookupMethod(methodName: String): Option[MethodInfo] = {
-      assert(isScalaClass,
+      assert(isScalaClass || isInterface,
           s"Cannot call lookupMethod($methodName) on non Scala class $this")
+
       @tailrec
       def loop(ancestorInfo: ClassInfo): Option[MethodInfo] = {
```
gzm0: This probably needs a better name now... How about `lookupInherited` or something like that?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -349,7 +350,80 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+      }
+
+      /* Deviation from the spec: if there are several targets, the spec
+       * chooses one arbitrarily. We fail the linking instead, so that we can
+       * more easily discover errors. Note that the Java *language* and
+       * compiler do not let this happen (in the absence of manipulated
+       * classpaths, of course).
+       */
+      if (notShadowed.size == 1) Some(notShadowed.head)
+      else None
```
gzm0: Could we have a better error here than "missing method"?

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -349,7 +350,80 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
           None
         }
       }
-      loop(this)
+      val existing =
+        if (isScalaClass) loop(this)
+        else methodInfos.get(methodName).filter(!_.isAbstract)
```
gzm0: So if I read this correctly, static calls on classes have inheritance while on interfaces they do not. Why is this so? Is this a parallel to JVM bytecode?

gzm0:
Ok. Done with a first pass. I'll probably need to go for this once more, since there are a couple of things I haven't grasped yet.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/JSDesugaring.scala
sjrd: In strong mode, we need top-level functions to be static methods of a class, because only consecutive classes can see each other. Whether they are static methods or default methods at the IR level, they should become top-level functions, but in strong mode they become ES static methods.

There's no collision because no IR method can start with `$f_`.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: OK

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analysis.scala
sjrd: Good point.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: Hum ... Not easily, I think :-s

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: It is so because, eventually, in JS, default methods cannot be made to inherit through any JS mechanism. Class methods are inherited through JS prototype chains, but interface methods cannot be inherited at the JS level.

If we chose to keep them inherited at the IR level, then desugaring would have to introduce indirections instead of the analyzer. But those indirections could not be removed by the optimizer like those produced by the analyzer are (through simple inlining). Not to mention `GenIncOptimizer` would be considerably more complicated.

In the JVM bytecode, they are obviously inherited (otherwise, default methods would exist at the bytecode level). That's why we allow them to be inherited before the analyzer (at the .sjsir level). The analyzer resolves the calls once and for all, creating statically linked bridges from there.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: It's even harder than I thought. It would require substantially changing the way `tryLookupXYZ` methods work, and by extension, `lookupXYZ` methods too. I would rather not go there without a compelling reason, i.e., at least one manifestation of a bug that would cause this to happen.

gzm0: Ok

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/JSDesugaring.scala
gzm0: Ok.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
```diff
@@ -336,20 +336,94 @@ final class Analyzer(semantics: Semantics, outputMode: OutputMode,
...
+      } else if (existing.exists(m => !m.isDefaultBridge || m.owner == this)) {
+        /* If we found a non-bridge, it must be the right target.
+         * If we found a bridge directly in this class/interface, it must also
+         * be the right target.
+         */
+        existing
+      } else {
+        // Try and find the target of a possible default bridge
+        findDefaultTarget(methodName).fold {
+          existing
```
gzm0: Shouldn't we be sure that `existing` is `None` here? By contradiction, `existing = Some(m)`. If `!m.isDefaultBridge`, we should never end up here. Therefore, `m.isDefaultBridge`. But we didn't find a default target from the subclass. That shouldn't happen.

Unless that reasoning is wrong, consider `assert(existing.isEmpty)`.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: Couldn't we just do `_error += ...` right here? Of course that would still trigger missing method errors at the call site (potentially), but it is IMO still better than just silently ignoring (potentially) that we encountered something we really shouldn't...

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: No, because *try*LookupMethod must not error out if it doesn't find anything. Its contract is to return `None`. That's why we need to change how this set of methods works.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
sjrd: Absolutely.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala
gzm0: It can still return `None` after recording the error. Because if we reach this point, there *is* an error in the IR no matter what the intent of the caller of `tryLookupMethod` was.
Sorry to insist that hard on this one. I just think it is really hard to figure out what is going on if you messed up your classpath and the linker tells you that the implementation for a method is missing that is not even abstract anywhere...

sjrd: There are cases where `tryLookupMethod` return `None` that are not errors. For example when called from ` tryLookupReflProxy`.

sjrd: However, this case in particular should indeed always be an error. I didn't think about it like that.

In fact it could even emit an error then return a `Some(notShadowed.head)` anyway, which would prevent the subsequent missing methods error. It would even be meaningful when bypassing linker errors.

gzm0: And align with the JVM spec ;) So no one can complain anymore :P

sjrd: Yes, although it turns out the JVM itself does not respect its own spec. Manipulating the classpath to create a conflict causes this runtime exception:
```
Exception in thread "main" java.lang.IncompatibleClassChangeError: Conflicting default methods: B.test A.test
        at Test.test(Test.java)
        at Test.main(Test.java:3)
```
So in fact our erroring out is compatible with the real JVM :p

sjrd: Ah but there's no `From` available at this point ... Which makes sense, since lookups were never supposed to error out. Even non-try lookups do not fail. They create non-existent methods, and it is only when they are reached that they produce errors.

So it should create a conflicting default bridge (similar to a non-existent method) which, when reached, would produce an error. But even that is hard, because try-lookups should not create non-existent stuff, only `lookupMethod` is supposed to produce erroneous methods. I guess `tryLookupMethod` should then return a `List[MethodInfo]` rather than an `Option[MethodInfo]`, and it is then the job of `lookupMethod` to produce an erroneous method if it receives more than one target. But that is likely going to slow down the analyzer as a whole, which is already a bottleneck of the pipeline.

I don't like this ... I'm tempted to just follow the JVM spec and return `notShadowed.headOption` without error. Or possibly send a warning to `Console.err` (we don't even have a logger here).

gzm0: Just use a `FromCore`. We already have a case like that: https://github.com/sjrd/scala-js/blob/default-methods/tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/Analyzer.scala#L191
With the tools refactoring, the `FromCore` will even track the component.

sjrd: OK

sjrd:
Updated, squashed, and rebased.

gzm0:
That's all.

gzm0:
LGTM after the list is fixed. Feel free to merge.

sjrd:
List fixed.

PR #2143, opened 2016-01-04
Reduce the number of classes in j.u.Collections.

sjrd:
Many methods of j.u.Collections wrapping collections specify that
they return a serializable collection if the wrapped collection
is serializable. The previous implementation would explicitly test
whether it was an instance of Serializable, and return different
classes inheriting from Serializable or not depending on that.

That is unnecessary. To be serializable, a class need not only be
an instance of Serializable. It must also contain only serializable
fields. Since the wrapped collection is captured by the wrapper,
it is a field, and the serializable property is carried over. We
can therefore always return collections that are instances of
Serializable, which reduces the number of classes in Collections.

This analysis is confirmed by experimentation on the JVM. Those
methods always return instances of Serializable.

### DIFF
--- a/javalib/src/main/scala/java/util/Collections.scala
+++ b/javalib/src/main/scala/java/util/Collections.scala
@@ -336,101 +336,45 @@ object Collections {
     }
   }
 
-  def unmodifiableCollection[T](c: Collection[_ <: T]): Collection[T] = {
-    class BasicImmutableCollection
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2146, opened 2016-01-06
Reboot sbt more often in CI.

sjrd:
In particular, always reboot it between a pre-link, fastOpt, and
fullOpt runs. This has at several benefits:

* The logs more clearly indicate what linking stage has failed.
* In case of retry, only the failed stage is retried.
* It gives more time to every single stage test.
* It consumes less RAM, as the states for fastOpt and fullOpt
  are not retained in memory at the same time.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -15,10 +15,10 @@
 
   <task id="main"><![CDATA[
     setJavaVersion $java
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2148, opened 2016-01-07
Decode class and method names in the traces of the linker.

sjrd:
This way, we show meaningful names to the developer, instead of
internal encodings.

PR #2150, opened 2016-01-08
Do not make ReferenceType extend Type anymore.

sjrd:
A ReferenceType is fundamentally *not* a Type. It so happens that
all concrete implementations of ReferenceType also implement Type,
but those two concepts must not be confused.

There are a few cases were it is OK-ish to treat a ReferenceType
as a Type (such as serializing/hashing), for which we introduced
casts.

This refactoring highlights a suspicious usage of ReferenceType as
Type in the OptimizerCore core. This commit leaves them as is,
with casts.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Types.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Types.scala
@@ -15,7 +15,19 @@ import Trees._
 
 object Types {
 
-  /** Type of an expression in the IR. */
+  /** Type of an term (expression or statement) in the IR.
+   *
+   *  There is a many-to-one relationship from [[ReferenceType]]s to types,
+   *  because:
+   *
+   *  - `scala.Byte`, `scala.Short` and `scala.Int` collapse to [[IntType]]
+   *  - `java.lang.Object` and raw JS types all collapse to [[AnyType]]
+   *
+   *  In fact, there are two `Type`s that do not have any real equivalent in
+   *  reference types: [[StringType]] and [[UndefType]], as they refer to the
+   *  non-null variants of `java.lang.String` and `scala.runtime.BoxedUnit`,
+   *  respectively.
+   */
   abstract sealed class Type {
     def show(): String = {
       val writer = new java.io.StringWriter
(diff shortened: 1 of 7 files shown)


sjrd:
Review by @gz

sjrd:
Review by @gzm0

gzm0:
It seems I am lacking fundamental knowledge here. What is the difference between a type and a reference type? (I'm assuming this is not the reference type in "value type" v.s. "reference type").

sjrd:
A `ReferenceType` has exactly the same level of precision as a JVM type. There is a one-to-one relationship between a `ReferenceType` and an instance of `java.lang.Class`. This means that:
* All primitive types have their reference type (including `scala.Byte` and `scala.Short`), and they are different from their boxed versions.
* Raw JS types are not erased to `any`
* Array types are like on the JVM

A `ReferenceType` therefore uniquely identifies a `classOf[T]`. It is also the reference types that are used in method signatures, and which therefore dictate JVM/IR overloading.

A `Type` is the type of an expression (or statement) in the IR. There is a many-to-one relationship from reference types to types, because:
* `scala.Byte`, `scala.Short` and `scala.Int` collapse to `int`
* `java.lang.Object` and raw JS types all collapse to `any`

(in fact, there are two `Type`s that do not have a real equivalent in reference types: `StringType` and `UndefinedType`, as they refer to the non-null variants of `java.lang.String` and `scala.runtime.BoxedUnit`, respectively)

ir/src/main/scala/org/scalajs/core/ir/Types.scala
```diff
@@ -84,13 +84,14 @@ object Types {
   case object NullType extends Type
 
   /** Reference types (allowed for classOf[], is/asInstanceOf[]). */
-  sealed abstract class ReferenceType extends Type
+  sealed trait ReferenceType
```
gzm0: Any reason you do not add a `toType` method here that essentially returns `this` (with `this` having a self-type of `Type with ReferenceType`)? It seems its less fragile than the casts but essentially does the same.

gzm0:
Could you also paste what you wrote about reference types into the documentation of the `Types` object?

gzm0:
That's all.

ir/src/main/scala/org/scalajs/core/ir/Types.scala
sjrd: I would argue that it would be *more* fragile, because it would lead to people assuming that it is always safe and meaningful to convert a `ReferenceType` into a `Type`, which it is not.

The *proper* `toType` conversion would require a predicate `isJSType: /* className: */ String => Boolean` to know whether `ClassType(className)` has to erase to `AnyType` or not.

sjrd:
> Could you also paste what you wrote about reference types into the documentation of the Types object?

Sure.

ir/src/main/scala/org/scalajs/core/ir/Types.scala
gzm0: Ok, I see. But why then don't you mark the cast in `asInstanceOf` as suspicious? IIUC we're doing the following wrong:

```scala
AsInstanceOf(???, ClassType("scala.Int")).tpe // -> ClassType("scala.Int")
```

But it should be `IntType`.

sjrd: `AsInstanceOf(???, ClassType("scala.Int"))` (which would actually be `AsInstanceOf(???, ClassType("I"))`) is not valid IR. It would be `Unbox(???, 'I')` which has `tpe = IntType`.

sjrd: This whole business is super hacky. We should definitely re-design this aspect of the IR for the next major version. At least with this PR we make it obvious that it's hacky ...

gzm0: Ok, fair enough.

gzm0:
Ok. So LGTM modulo the comment on `Types`.

sjrd:
Updated with the documentation.

sjrd:
Rebased on top of #2148. There was a compile error with the automatic clean merge.

gzm0:
LGTM

PR #2155, opened 2016-01-09
Fix #1949: Add/fix toString() methods for tools.jsdep.*.

### LINKED ISSUE #1949: Missing toString() methods in tools.jsdeps.*
A few classes in https://github.com/scala-js/scala-js/tree/master/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep miss proper `toString()` implementations. This is a pity, as they sometimes appear in error messages, and can also be displayed with `show` commands in sbt.
* [`FlatJSDependency`](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala)
* [`JSDependencyManifest`](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/JSDependencyManifest.scala)
* [`ResolutionInfo`](https://github.com/scala-js/scala-js/blob/master/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/ResolutionInfo.scala)

sjrd:


sjrd:
Review by @nicolasstucki

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
```diff
@@ -34,4 +34,16 @@ final class FlatJSDependency(
...
 
+  override def toString(): String = {
+    val b = new StringBuilder
+    b ++= s"FlatJSDependency(origin=$origin, relPath=$relPath"
+    if (dependencies.nonEmpty)
+      b ++= s", dependencies=$dependencies"
+    if (commonJSName.nonEmpty)
+      b ++= s", commonJSName=$commonJSName"
+    if (relPathMinified.nonEmpty)
+      b ++= s", relPathMinified=$relPathMinified"
```
nicolasstucki: You are using both `StringBuilder` and string interpolators, I would suggest to only use the builder. This would be to gain performance and I would argue that there is no loss in expressivity or style in this particular case.

nicolasstucki: And clearly the same in all other cases.

nicolasstucki:
That's all, otherwise LGTM

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/FlatJSDependency.scala
sjrd: I was following the style of `toString()` methods in other (existing) classes of the [jsdep](https://github.com/scala-js/scala-js/tree/master/tools/shared/src/main/scala/org/scalajs/core/tools/jsdep) package.

Should I change them all?

I'm not convinced by the performance argument, here. These methods are essentially never called, except via a manual interaction with the sbt CLI.

> there is no loss in expressivity or style in this particular case

Could you show an example of the style you would expect?

nicolasstucki: Yes, best to maintain the implementation similar. The performance argument is quite weak.

PR #2160, opened 2016-01-10
Optimize the extraction of `lo` and `hi` parts of a Long.

sjrd:
This optimization introduces two new rewritings:
* `(x >>> 32).toInt` or `(x >> 32).toInt` -> `x.hi()`
* `x.toInt` -> `x.lo()`

The calls to `hi()` and `lo()` are also inlined, giving, at the moment, `x.hi$2` and `x.lo$2`, respectively.

This allows user code extracting the 32 high and low bits of a Long to do so efficiently.

With that optimization, we can then remove several intrinsics from `java.lang.Long`.

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
@@ -576,8 +576,9 @@ final class RuntimeLong(val lo: Int, val hi: Int)
     else js.Tuple4(quotLo, quotHi, remLo, remHi)
   }
 
-  // Support for intrinsics
+  // TODO Remove these. They were support for intrinsics before 0.6.6.
 
+  @deprecated("Use java.lang.Long.toBinaryString instead.", "0.6.6")
   def toBinaryString: String = {
     val zeros = "00000000000000000000000000000000" // 32 zeros
     @inline def padBinary32(i: Int) = {
@@ -592,6 +593,7 @@ final class RuntimeLong(val lo: Int, val hi: Int)
     else Integer.toBinaryString(lo)
   }
 
+  @deprecated("Use java.lang.Long.toHexString instead.", "0.6.6")
   def toHexString: String = {
     val zeros = "00000000" // 8 zeros
     @inline def padHex8(i: Int) = {
@@ -606,6 +608,7 @@ final class RuntimeLong(val lo: Int, val hi: Int)
     else Integer.toHexString(lo)
   }
 
+  @deprecated("Use java.lang.Long.toOctalString instead.", "0.6.6")
   def toOctalString: String = {
     val zeros = "0000000000" // 10 zeros
     @inline def padOctal10(i: Int) = {
@@ -625,9 +628,11 @@ final class RuntimeLong(val lo: Int, val hi: Int)
     else Integer.toOctalString(lp)
   }
 
+  @deprecated("Use java.lang.Long.bitCount instead.", "0.6.6")
   def bitCount: Int =
     Integer.bitCount(lo) + Integer.bitCount(hi)
 
+  @deprecated("Use java.lang.Long.signum instead.", "0.6.6")
   def signum: RuntimeLong = {
     val hi = this.hi
     if (hi < 0) MinusOne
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2150,7 +2136,30 @@ private[optimizer] abstract class OptimizerCore(
...
 
-          case _ => default
+          case BinaryOp(BinaryOp.Long_>> | BinaryOp.Long_>>>,
+              x, IntLiteral(32)) =>
+            // x.hi__I()
+            staticCall(LongImpl.RuntimeLongClass, LongImpl.hi).fold[Tree] {
+              default
+            } { target =>
+              trampoline {
+                implicit val scope = Scope.Empty // hack
```
gzm0: Please explain the hack.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2150,7 +2136,30 @@ private[optimizer] abstract class OptimizerCore(
...
+              }
+            }
+
+          case _ =>
+            // arg.lo__I()
+            staticCall(LongImpl.RuntimeLongClass, LongImpl.lo).fold[Tree] {
+              default
+            } { target =>
+              trampoline {
+                implicit val scope = Scope.Empty // hack
```
gzm0: Please explain the hack.

gzm0:
That's all.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/optimizer/OptimizerCore.scala
```diff
@@ -2150,7 +2136,42 @@ private[optimizer] abstract class OptimizerCore(
...
+            // x.hi__I()
+            staticCall(LongImpl.RuntimeLongClass, LongImpl.hi).fold[Tree] {
+              default
+            } { target =>
+              trampoline {
+                /* Hack: use an empty scope to inline x.hi__I().
+                 * `inline` requires a scope to keep track of the set methods
+                 * currently being inlined, and prevent infinite inlining of
+                 * recursive methods. In theory we should provide a proper
+                 * scope, but we don't have one here. It give an empty scope
```
gzm0: Typo.

gzm0:
LGTM otherwise.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -576,8 +576,9 @@ final class RuntimeLong(val lo: Int, val hi: Int)
     else js.Tuple4(quotLo, quotHi, remLo, remHi)
   }
 
-  // Support for intrinsics
+  // TODO Remove these. They were support for intrinsics before 0.6.6.
 
+  @deprecated("Use java.lang.Long.toBinaryString instead.", "0.6.6")
```
gzm0: Actually, why do we need to keep these around?

gzm0: (you probably have explained this already 5 times to me... :-/)

sjrd: Well, last time I explained, the reason was: an 0.6.5 sbt plugin could link an 0.6.6 library, and would therefore expect those to be there. However, since 0.6.6 breaks forward binary compatibility, that cannot happen. So in theory, we could remove them. In practice, we would have to add exceptions in MiMa for the `library`, and I'd rather avoid that if we can avoid it. If there is no good reason to break the binary compatibility of the library, we shouldn't, I believe (even if it's in `runtime`).

sjrd:
Fixed the typo.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
gzm0: Ok. Fair enough.

gzm0:
LGTM

PR #2165, opened 2016-01-11
Fix #1892: Officially deprecate package objects extending js.Any.

### LINKED ISSUE #1892: @js.native cannot be applied to package objects

sjrd:


sjrd:
Review by @nicolasstucki or @gzm0

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -458,7 +458,10 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
       val isJSNative = !sym.hasAnnotation(ScalaJSDefinedAnnotation)
 
-      if (isJSNative && !isJSAnonFun &&
+      if (sym.isPackageObjectClass) {
+        reporter.warning(implDef.pos,
+            "package objects inheriting from js.Any are deprecated")
```
gzm0: - Consider adding: "Use a normal object instead."
- Add a full stop at the and
- Consider starting with upper case (our error messages seem to do so, I don't know about scalac)

gzm0:
That's all.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: So:

    Package objects inheriting from js.Any are deprecated. Use a normal object instead.

?

gzm0: Yep.

sjrd:
Updated.

gzm0:
LGTM

nicolasstucki:
LGTM

PR #2166, opened 2016-01-11
Fix #1983: Warn if @ScalaJSDefined overrides js names differ.

### LINKED ISSUE #1983: Missing @JSName annotation warning after overriding a @ScalaJSDefined method
There should be at least an information that ```@JSName``` annotation has to be reapplied after overriding a method from a ```@ScalaJSDefined``` class. Best would be if the annotation is carried over, because I think it's a common pattern in ```@ScalaJSDefined``` classes to rename the scala method name from a facade to provide a custom method with the actual method name and call the renamed one. 
```
@ScalaJSDefined
abstract class A extends js.Object {
  @JSName("_test")
  def test(): Int
  @JSName("test")
  def _test(): String = test().toString
}
val a = new A {
  override def test() = 1
}
```
This leads to a ```Conflicting properties and methods for ...``` Error because the ```@JSName``` annotation is not carried over.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

nicolasstucki:
Jenkins, retest this please

sjrd:
Here are a few more things to try:
```scala
class A {
  def foo: Int = 5
}

trait B extends A {
  @JSName("bar")
  def foo: Int
}

class C extends B
```
```scala
class A {
  @JSName("bar")
  def foo: Int = 5
}

trait B extends A {
  def foo: Int
}

class C extends B
```
```scala
class A[T] {
  @JSName("bar")
  def foo(x: T): T = x
}

class B extends A[Int] {
  override def foo(x: Int): Int = x
}
```
```scala
trait A[T] {
  @JSName("bar")
  def foo(x: T): T = x
}

class B extends A[Int] {
  override def foo(x: Int): Int = x
}
```
```scala
class A[T] {
  @JSName("bar")
  def foo(x: T): T = x
}

trait B extends A[Int] {
  def foo(x: Int): Int

class C extends B {
  override def foo(x: Int): Int = x
}
```
```scala
class A[T] {
  def foo(x: T): T = x
}

trait B extends A[Int] {
  @JSName("bar")
  def foo(x: Int): Int

class C extends B {
  override def foo(x: Int): Int = x
}
```

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -719,6 +719,18 @@ abstract class PrepJSInterop extends plugins.PluginComponent
                 "A Scala.js-defined JS trait cannot contain private members")
           }
         }
+
+        // @ScalaJSDefined overrides with different js names - issue #1983
+        sym.allOverriddenSymbols.headOption match {
```
sjrd: There's no reason to limit this test to non-native JS classes. I suggest to apply it to *all* JS types (native and Scala.js-defined).

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -585,6 +585,22 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
         checkJSNameLiteral(sym)
 
+      // Check for overrides with different JS names - issue #1983
+      for (overridingPair <- new overridingPairs.Cursor(sym).iterator) {
+        val lowJSName = jsInterop.jsNameOf(overridingPair.low)
+        val highJSName = jsInterop.jsNameOf(overridingPair.high)
+        if (lowJSName != highJSName) {
+          reporter.warning(implDef.pos,
+              s"""
+                |A Scala.js-defined JS class member is overriding another with a different exported name.
```
sjrd: A member of a JS class is overriding another member with a different JS name.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -585,6 +585,22 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+                |
+                |${overridingPair.lowString} with exported JSName '$lowJSName'
```
sjrd: ~~exported~~

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -585,6 +585,22 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+                |    is conflicting with
+                |${overridingPair.highString} with exported JSName '$highJSName'
```
sjrd: ~~exported~~

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -585,6 +585,22 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       if (shouldCheckLiterals)
         checkJSNameLiteral(sym)
 
+      // Check for overrides with different JS names - issue #1983
+      for (overridingPair <- new overridingPairs.Cursor(sym).iterator) {
```
sjrd: That is cool!

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -585,6 +585,22 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+        val lowJSName = jsInterop.jsNameOf(overridingPair.low)
+        val highJSName = jsInterop.jsNameOf(overridingPair.high)
+        if (lowJSName != highJSName) {
+          reporter.warning(implDef.pos,
```
sjrd: Could we use the `low`'s `pos` when `low.owner == sym`?

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -750,4 +750,381 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+    """
+      |newSource1.scala:11: warning:
+      |A Scala.js-defined JS class member is overriding another with a different exported name.
+      |
+      |override def bar(): Int at line 13 with exported JSName 'baz'
+      |    is conflicting with
+      |def bar(): Int in class A with exported JSName 'foo'
+      |
+      |    class B extends A {
+      |          ^
```
sjrd: For example of the position problem, ideally this should point to `B.bar`, not to `B` itself.

sjrd:
That's all.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: In 2.10 the `Cursor` doesn't have the `iterator` method, but at least it exists.

compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
```diff
@@ -100,6 +100,43 @@ trait Compat210Component {
...
+    }
+
+    class SymbolPair(val low: Symbol, val high: Symbol)
+
+    /** To make this compat code compile in 2.11 as the fields `overriding` and
+     *  `overridden` are only present in 2.10.
+     */
+    private implicit class Cursor210toCursor211(cursor: overridingPairs.Cursor) {
+      def overriding: Symbol = throw new UnsupportedOperationException
+      def overridden: Symbol = throw new UnsupportedOperationException
```
sjrd: Use `sys.error("infinite loop in Compat")`, like at other similar places in this file.

compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
```diff
@@ -100,6 +100,43 @@ trait Compat210Component {
     }
   }
 
+  // Compat to support: new overridingPairs.Cursor(sym).iterator
+
+  implicit class OverridingPairsCursor2Iterable(cursor: overridingPairs.Cursor) {
+    def iterator: Iterator[SymbolPair] = new Iterator[SymbolPair] {
+
```
sjrd: I would remove this blank line.

compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
```diff
@@ -100,6 +100,43 @@ trait Compat210Component {
...
+        new SymbolPair(low, high)
+      }
+
+      private def skipIgnoredEntries(): Unit = {
+        while (ignoreCurrentEntry)
+          cursor.next()
+      }
+
+      private def ignoreCurrentEntry: Boolean =
+        cursor.overriding.name == nme.isInstanceOf_
```
sjrd: This warrants a comment :p

compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
```diff
@@ -100,6 +100,43 @@ trait Compat210Component {
...
+      def next(): SymbolPair = {
+        val low = cursor.overriding
+        val high = cursor.overridden
+        cursor.next()
+        skipIgnoredEntries()
+        new SymbolPair(low, high)
+      }
+
+      private def skipIgnoredEntries(): Unit = {
+        while (ignoreCurrentEntry)
```
sjrd: I think this should have `cursor.hasNext &&`

sjrd:
That's all (+ squash the commits)

sjrd:
Modulo squashing, LGTM

nicolasstucki:
Squashed

sjrd:
LGTM

sjrd:
retest this please

PR #2169, opened 2016-01-12
Fix #1889: Override the dependency of partest on scala-library.jar.

### LINKED ISSUE #1889: Investigate 3 partests sometimes failing on weekly
* `run/bitsets.scala`
* `neg/t8266-invalid-interp.scala`
* `neg/t7848-interp-warn.scala`

They don't emit the same warnings as what they're supposed to.

sjrd:
For some reason, the transitive dependency of scala-partest on
scala-library.jar leaks into the classpath used to compile the
partests. Since scala-partest 1.0.9 depends on scala-library
2.11.6, that version overrides the scala-library for the current
Scala version when testing on 2.11.5 and earlier. This causes
wrong warnings to be reported, and the check files are not happy.

This commit fixes this problem by overriding transitive
dependencies and forcing scala-library with the version we're
testing.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1384,6 +1384,9 @@ object Build extends sbt.Build {
           fork in Test := true,
           javaOptions in Test += "-Xmx1G",
 
+          // Override the dependency of partest - see #1889
+          dependencyOverrides += "org.scala-lang" % "scala-library" % scalaVersion.value % "test",
+
           testFrameworks ++= {
             if (shouldPartest.value)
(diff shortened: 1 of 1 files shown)


sjrd:
I will run a weekly on this one when Jenkins is a bit less overloaded.

nicolasstucki:
LGTM

sjrd:
Weekly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/3527/

sjrd:
Weekly actually passed :-)

PR #2176, opened 2016-01-17
Fix #1998: Add a semantics flag for development v production mode.

### LINKED ISSUE #1998: Dev / Production mode
Do we have a means to produce different JS (use different Scala trees) according to configurable compile-time settings? Maybe some args passed through SBT?

Some context:  In Babel there's an optimisation available which they only switch on when making a release: http://babeljs.io/blog/2015/03/31/5.0.0/#inline-elements

I'd like to support it in scalajs-react but I don't have a means of letting users specify dev/prod mode at compile-time. Up until now I've been using `-Xelide-below` and tricks with `@elidable` which work well enough, but I don't think I should use that to support something larger like an optimisation setting.

Any ideas?

sjrd:
Semantics receive an additional flag `productionMode`, which is,
by default, false in fastOpt stage in true in fullOpt stage. It
can be queried with
`scala.scalajs.runtime.{production,development}Mode`
to write user-level code that optimizes itself for production,
for example by discarding expensive run-time sanity checks.

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/package.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/package.scala
@@ -6,38 +6,10 @@ import scala.collection.GenTraversableOnce
 
 package object runtime {
 
-  /** Returns true if we are assuming that the target platform supports
-   *  ECMAScript 6, false otherwise.
-   *
-   *  This ends up being constant-folded to a constant at link-time. So
-   *  constant-folding, inlining, and other local optimizations can be
-   *  leveraged with this "constant" to write polyfills that can be
-   *  dead-code-eliminated.
-   *
-   *  A typical usage of this method is:
-   *  {{{
-   *  if (assumingES6 || featureTest())
-   *    useES6Feature()
-   *  else
-   *    usePolyfill()
-   *  }}}
-   *
-   *  At link-time, `assumingES6` will either be a constant false, in which
-   *  case the above snippet folds into
-   *  {{{
-   *  if (featureTest())
-   *    useES6Feature()
-   *  else
-   *    usePolyfill()
-   *  }}}
-   *  or a constant true, in which case it folds into
-   *  {{{
-   *  useES6Feature()
-   *  }}}
-   */
+  @deprecated("Use scala.scalajs.LinkingInfo.assumingES6 instead.", "0.6.6")
   @inline
   def assumingES6: Boolean =
-    linkingInfo.assumingES6
+    scala.scalajs.LinkingInfo.assumingES6
 
   def wrapJavaScriptException(e: Any): Throwable = e match {
     case e: Throwable => e
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -110,6 +110,12 @@ object BinaryIncompatibilities {
   )
 
   val Library = Seq(
+      // In theory, breaking, but this is an interface in runtime that no one should extend
+      ProblemFilters.exclude[MissingMethodProblem](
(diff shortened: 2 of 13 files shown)


sjrd:
Review by @gzm0

library/src/main/scala/scala/scalajs/runtime/package.scala
```diff
@@ -6,6 +6,68 @@ import scala.collection.GenTraversableOnce
...
+   *  }}}
+   *  or a constant false, in which case it is dead-code-eliminated away,
+   *  yielding maximum performance in production.
+   *
+   *  @see [[productionMode]]
+   */
+  @inline
+  def developmentMode: Boolean =
+    !productionMode
+
```
gzm0: We expect this to be used in user code, right? So the `runtime` package is arguably a bad place since we don't give any guarantees about how it evolves, no?

project/BinaryIncompatibilities.scala
```diff
@@ -110,6 +110,12 @@ object BinaryIncompatibilities {
   )
 
   val Library = Seq(
+      // In theory, breaking, but this is an interface in runtime that no one should extend
```
gzm0: Yep, this confirms my suspicion :P

library/src/main/scala/scala/scalajs/runtime/package.scala
sjrd: Yes, but that's also where we put `assumingES6`. I don't think there exists a better place right now. We could create one, but I'm not sure where?

gzm0: Is `assumingES6` supposed to be used by non-internal code?

gzm0: I think we should create a place, since otherwise it is very confusing for users what they may rely upon and what not. For instance, many methods in this package object should definitely not be called in user code. So maybe we need to create `scala.scalajs.LinkingInfo` / `linking_info`?

sjrd: > Is assumingES6 supposed to be used by non-internal code?

There are reasonable use cases for that, yes.

> I think we should create a place

I agree.

> `scala.scalajs.LinkingInfo` / `linking_info`?

`scala.scalajs.LinkingInfo` is the least bad I've seen/thought about so far.

gzm0: One thing I was wondering is whether we should put it in the `org.scalajs` namespace...

sjrd: No, we shouldn't. It's core, as it is allowed to see `runtime`.

gzm0: Ok. So `LinkingInfo` it is I guess...

sjrd:
Updated.

project/Build.scala
```diff
@@ -1112,7 +1112,11 @@ object Build extends sbt.Build {
 
         val envTags = envTagsFor((resolvedJSEnv in Test).value)
 
-        val sems = (scalaJSSemantics in Test).value
+        val sems = (scalaJSStage in Test).value match {
+          case FastOptStage => (scalaJSSemantics in (Test, fastOptJS)).value
+          case FullOptStage => (scalaJSSemantics in (Test, fullOptJS)).value
```
gzm0: Hmmm... Should we make `dispatchSettingKeySettings` public?

gzm0: Ah, nevermind... Wouldn't solve this anyways...

gzm0:
LGTM

sjrd:
retest this please

PR #2181, opened 2016-01-18
Fix #2162: Unify error handling in the Analyzer.

### LINKED ISSUE #2162: Unify error handling in Analyzer
There are three ways the analyzer currently handles errors (i.e. bad state on the classpath, not bad internal state):

- Populate `errors`
- `sys.error("Fatal error: could not find java.lang.Object on the classpath")`
- Throw `CyclicDependencyException`

This is a strange interface and we should unify it taking advantage of the binary compat break.

sjrd:


sjrd:
Review by @gzm0 

The fatal error cases were tested locally.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -87,11 +124,20 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
...
+      val fatal = !bypassLinkingErrors || analysis.errors.exists {
+        case _: Analysis.MissingJavaLangObjectClass => true
+        case _: Analysis.CycleInInheritanceChain    => true
+        case _                                      => false
+      }
 
-    if (analysis.errors.nonEmpty && !bypassLinkingErrors)
-      sys.error("There were linking errors")
+      val linkingErrLevel = if (fatal) Level.Error else Level.Warn
+      analysis.errors.foreach(logError(_, logger, linkingErrLevel))
```
gzm0: Shouldn't the error level be individual for each error message?

sjrd: It's arguable. 

But anyway, in practice it wouldn't change anything, because either you get fatal errors, or you get non fatal errors. There is no way you can get a mix of both.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -165,6 +173,7 @@ private final class Analyzer(semantics: Semantics,
   private class ClassInfo(data: Infos.ClassInfo) extends Analysis.ClassInfo {
     private[this] var _linking = false
```
gzm0: Drop this and use `_linkingFrom` instead.

sjrd: I can't. That's what I tried initially, but I failed to discover cycles and it went into infinite loop. There is a time during which `_linking` is `true` but `_linkingFrom` needs to stay `null`.

gzm0: :P No, not really. The problem is more that `_linkingFrom == null` can mean two things then:

- I'm not linking
- I'm a linking root

Not being able to distinguish these cases is problematic. Not because we can't detect cycles, but because we can't reconstruct the cycles.

I think it would be cleaner if you made `_linkingFrom` an `Option` and then essentially have proper tri-state (with the `null`).

gzm0: Or, don't condition on `_linkingFrom` at all anymore. (why do you need the `_linkingFrom == null` check)?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -182,22 +191,29 @@ private final class Analyzer(semantics: Semantics,
...
+          _linking = true
           linkClassesImpl()
-        } catch {
-          case CyclicDependencyException(chain) =>
-            throw CyclicDependencyException(encodedName :: chain)
+        } else {
+          def makeCycle(from: ClassInfo, acc: List[ClassInfo]): List[ClassInfo] = {
+            if (from == linkingFrom) from :: acc
+            else makeCycle(from._linkingFrom, from :: acc)
+          }
```
gzm0: I don't think this does the right thing. It walks up the dependency tree that has already been established and then checks for the `linkingFrom`. There is no guarantee we reach that:

Consier: `A <- B <- C <- B`. When we reach `B` for the second time, we get the following invocation order:

```
makeCycle(B, Nil)
makeCycle(A, B :: Nil)
makeCycle(null, A :: B :: Nil)
NPE
```

The proper invocation is `makeCycle(linkingFrom, Nil)` and the abort condition is `from eq this`:

```
makeCycle(C, Nil)
makeCycle(B, C :: Nil)
--> B :: C :: Nil
```

This is probably also the reason why you need to set `_linkingFrom` even if `_linking`. You can then "detect" loops if the loop involves the root linking node.

gzm0: Hmmm... I'm thinking of this the complete wrong way... My initial class hierarchy doesn't make sense... `B` has two different parents.
Need to think more.

sjrd: I guess this algorithm is just too complicated. I suggest to revert back to the throwing thing, but just catch it one last time in the top-level `Analyzer.linkClasses`, and from there, create exactly one `CycleInInheritanceChain` error to put in `_errors`. It's fatal anyway, so we don't really care that we won't discover *other* potential cycles on the classpath.

gzm0: That's probably easier. Optionally you could also have `linkClasses` an `Option[List[ClassInfo]]` so we don't have to use exceptions for control flow... But I think in this case its actually ok.

sjrd:
Updated.

gzm0:
LGTM

PR #2142, opened 2016-01-02
Inline init method in the constructor

tOverney:
When effectively final and only 1 constructor.

cc/ @paullepoulpe @sjrd

PR #2183, opened 2016-01-18
Further tools refactoring.

sjrd:
This commit mostly shuffles classes around in packages for a better
organization and stricter encapsulation of concerns:

* Everything related to Emitting JavaScript code from the IR is
  moved to backend.emitter.
* The OutputMode is not visible at the generic LinkerBackend level
  nor at the GenLinker level. It is a specificity of the two
  existing backends, but one could create a separate backend having
  nothing to do with those.

### DIFF
--- a/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
+++ b/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
@@ -20,13 +20,14 @@ import com.google.javascript.jscomp.{
 
 import org.scalajs.core.tools.corelib.CoreJSLibs
 import org.scalajs.core.tools.io._
-import org.scalajs.core.tools.javascript.OutputMode
+import org.scalajs.core.tools.javascript.ESLevel
 import org.scalajs.core.tools.logging.Logger
 import org.scalajs.core.tools.sem.Semantics
 
 import org.scalajs.core.tools.linker.LinkingUnit
 import org.scalajs.core.tools.linker.analyzer.SymbolRequirement
-import org.scalajs.core.tools.linker.backend.{LinkerBackend, Emitter}
+import org.scalajs.core.tools.linker.backend.{OutputMode, LinkerBackend}
+import org.scalajs.core.tools.linker.backend.emitter.Emitter
 
 /** The Closure backend of the Scala.js linker.
  *
@@ -35,12 +36,12 @@ import org.scalajs.core.tools.linker.backend.{LinkerBackend, Emitter}
  */
 final class ClosureLinkerBackend(
     semantics: Semantics,
-    outputMode: OutputMode,
     withSourceMap: Boolean,
     config: LinkerBackend.Config
-) extends LinkerBackend(semantics, outputMode, withSourceMap, config) {
+) extends LinkerBackend(semantics, ESLevel.ES5, withSourceMap, config) {
 
-  private[this] val emitter = new Emitter(semantics, outputMode)
+  private[this] val emitter =
+    new Emitter(semantics, OutputMode.ECMAScript51Isolated)
 
   val symbolRequirements: SymbolRequirement = emitter.symbolRequirements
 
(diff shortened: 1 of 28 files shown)


sjrd:
Review by @gzm0 

My review of #2142 made me realize that `JSDesugaring` and `ScalaJSClassEmitter` are much more coupled to `Emitter` than to `javascript.*`. In the process of isolating emitter-related stuff, I realized `OutputMode` was not a general property of a back-end. It is a configuration option of the two back-ends we happen to have so far, but it is not general.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -38,7 +38,7 @@ final class ClosureLinkerBackend(
     outputMode: OutputMode,
```
gzm0: Can we really chose the `outputMode` here?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/LinkerBackend.scala
```diff
@@ -12,9 +12,9 @@ package org.scalajs.core.tools.linker.backend
 import java.net.URI
 
 import org.scalajs.core.tools.io.WritableVirtualJSFile
-import org.scalajs.core.tools.javascript.OutputMode
 import org.scalajs.core.tools.logging.Logger
 import org.scalajs.core.tools.sem.Semantics
+import org.scalajs.core.tools.javascript.ESLevel
```
gzm0: Why the reordering?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -22,11 +22,12 @@ import Trees._
 import Types._
 
 import org.scalajs.core.tools.sem._
-import org.scalajs.core.tools.javascript.{ESLevel, LongImpl}
+import org.scalajs.core.tools.javascript.ESLevel
 import org.scalajs.core.tools.logging._
 
 import org.scalajs.core.tools.linker._
 import org.scalajs.core.tools.linker.analyzer.SymbolRequirement
+import org.scalajs.core.tools.linker.backend.emitter.LongImpl
```
gzm0: This makes a nice caveat apparent: The optimizer is not agnostic to the backend used because of longs :)

gzm0:
That's all.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: You can choose es 5 global or isolated. And assuming we can upgrade gcc someday, es 6 will also become an option.

gzm0: But no matter what you chose, you'll get isolated in the end, no? So how we desugar internally before feeding to GCC really is an implementation detail, no?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
sjrd: Yep ^^

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: Also, if we currently can't have ES5, we should just not have the option and add it later...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/LinkerBackend.scala
sjrd: An accident of the history of the refactoring. Will fix.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: Makes sense. I'll remove it.

sjrd:
Updated.

gzm0:
LGTM

PR #2189, opened 2016-01-20
Last tweaks to the tools and JS envs

sjrd:


sjrd:
Review by @gzm0

sjrd:
With these changes, the generated Scaladocs of the Tools and the JS Envs actually make sense :p

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/CoreJSLibs.scala
```diff
@@ -20,7 +20,8 @@ import org.scalajs.core.tools.linker.backend.OutputMode
 import scala.collection.immutable.Seq
 import scala.collection.mutable
 
-object CoreJSLibs {
+// The only reason this is not private[backend] is that Rhino needs it
```
gzm0: `private[emitter]`?

gzm0:
One thing I'm a bit weary of, is that we privilege ourselves to use deep internals of the linking pipeline to implement one of the jsEnvs. Maybe it would be worth leaving a comment on one (or some) of these `private[scalajs]` entry points saying something the like: "If you would like to implement a lazy-loading JSEnv (like RhinoJSEnv) and have reached this point, **don't** move your package or parts thereof to `scalajs`, but contact us and we'll spec and create a clean interface to do so.".

In any case, I completely agree that what we have in this PR is sufficient for the moment (i.e. no need to do fancy interfacing for lazy loading etc. at this point).

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/CoreJSLibs.scala
sjrd: It is also needed by the Closure backend

sjrd:
Yes, ideally a clean interface would be needed. It will be easier to add one later, now that everything is private, when we need it. 

Honestly I don't think anyone else but us will ever write a lazy loading env, so that issue will probably not arise anytime soon.

sjrd:
@gzm0 I'm not sure whether you want me to do anything on this PR?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/CoreJSLibs.scala
gzm0: Ah, ok.

gzm0:
> I'm not sure whether you want me to do anything on this PR?

Nah, let's just leave it.

gzm0:
LGTM

PR #2191, opened 2016-01-21
Update developer documentation.

sjrd:


sjrd:
Review by @nicolasstucki

DEVELOPING.md
```diff
@@ -38,12 +38,14 @@ you typically need to
 
     > reload
 
-To test in fastOpt or fullOpt stage, use the usual Scala.js stage setting:
+To test with Node.js instead of Rhino, or to test in fullOpt stage, use the
+usual Scala.js settings:
 
-    > set scalaJSStage in Global := FastOptStage
+    > set scalaJSUseRhino in Global := false
+    > set scalaJSStage in Global := FullOptStage
```
nicolasstucki: I would say that this is not so clear. You are explaining two different cases and then the two commands which could be read as executing the two commands one after the other.

sjrd:
Updated.

nicolasstucki:
LGTM

PR #2194, opened 2016-01-22
Exclude JUnit from json-simple's transitive dependencies

xuwei-k:
https://code.google.com/p/json-simple/issues/detail?id=91
http://repo1.maven.org/maven2/com/googlecode/json-simple/json-simple/1.1.1/json-simple-1.1.1.pom

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -534,7 +534,7 @@ object Build extends sbt.Build {
       settings = commonToolsSettings ++ Seq(
           libraryDependencies ++= Seq(
               "com.google.javascript" % "closure-compiler" % "v20130603",
-              "com.googlecode.json-simple" % "json-simple" % "1.1.1",
+              "com.googlecode.json-simple" % "json-simple" % "1.1.1" exclude("junit", "junit"),
               "com.novocode" % "junit-interface" % "0.9" % "test"
           )
       )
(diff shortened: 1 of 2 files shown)


sjrd:
Will have to be rebased on top of the Towards 0.6.7 commit when it lands in master.

sjrd:
This should now be rebased on master.

Could you also explain why this change is needed, and what it does, exactly? I never experienced any problem.

xuwei-k:
I just rebased on master.

Hmm, I can't explain with a specific example, but I thinks this is a just unnecessary dependency.

sjrd:
Jenkins, test this please

gzm0:
I agree that it doesn't make sense at all that the compile configuration (what's that in maven's / ivy's terms?) depends on `junit`.

sjrd:
Could you make the commit message a bit clearer. Something like:

    Exclude JUnit from json-simple's transitive dependencies

    json-simple erroneously declares a dependency on JUnit, see:
    https://code.google.com/archive/p/json-simple/issues/91
    http://repo1.maven.org/maven2/com/googlecode/json-simple/json-simple/1.1.1/json-simple-1.1.1.pom

Also, as such, does this automatically exclude that double-transitive dependency from libraries that would depend on `scalajs-tools`? Or would they all have to exclude it? How does that work?

sjrd:
Jenkins, retest this please

sjrd:
LGTM

PR #2196, opened 2016-01-23
Towards 0.6.7.

sjrd:
For reference, the Towards 0.6.6 commit was d0f87774796bc7d9d79a97acfedfa0139cd7ea09

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -48,7 +48,7 @@ object Build extends sbt.Build {
   val shouldPartest = settingKey[Boolean](
     "Whether we should partest the current scala version (and fail if we can't)")
 
-  val previousVersion = "0.6.5"
+  val previousVersion = "0.6.6"
   val previousSJSBinaryVersion =
     ScalaJSCrossVersion.binaryScalaJSVersion(previousVersion)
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @nicolasstucki for fast-forward push to master (not through GitHub's merge button).

project/Build.scala
```diff
@@ -98,10 +98,12 @@ object Build extends sbt.Build {
           case ScalaJSCrossVersion.binary => previousBinaryCrossVersion
           case crossVersion               => crossVersion
         }
+        val prevExtraAttributes =
+          thisProjectID.extraAttributes.filterKeys(_ != "e:info.apiURL")
```
sjrd: This is actually a fix for the issue I experienced in #2019, which came back hard on this commit for all Scala versions, not just 2.12.x.

nicolasstucki: Then I suggest adding a comment that mentions why we need to filter that key.

nicolasstucki:
That is all, otherwise LGTM

sjrd:
Updated.

nicolasstucki:
LGTM

Issue #2202, opened 2016-01-26
Concurrent use of `scalaJSLinker` in sbt plugin


gzm0:
The Scala.js sbt plugin does not ensure that a project's linker instance is not used concurrently. A typical scenario where this arises is the following (discovered thanks to @reid-spencer):

```scala
val client = project.enablePlugins(ScalaJSPlugin, ScalaJSPlay)
val server = project.enablePlugins(PlayScalaJS).settings(scalaJSProjects += client)
val root = project.aggregate(client, server)
```
Then invoke:
```sh
sbt test
```

sbt now concurrently invokes the `client/test:loadedJSEnv` and the `client/test:fastOptJS` task. The former in order to detect the necessary tests for the `client` project, the latter to satisfy `server`'s dependency<sup id="a1">[1](#f1)</sup>.

These tasks in turn invoke `linkUnit`, `link` respectively on `client/test:scalaJSLinker` concurrently. This is not an allowed use of the linker. "Fortunately" an assert in `GenIncOptimizer` caught this:

```
java.lang.AssertionError: assertion failed
        at scala.Predef$.assert(Predef.scala:165)
        at org.scalajs.core.tools.linker.frontend.optimizer.GenIncOptimizer.withLogger(GenIncOptimizer.scala:97)
        at org.scalajs.core.tools.linker.frontend.optimizer.GenIncOptimizer.update(GenIncOptimizer.scala:107)
        at org.scalajs.core.tools.linker.frontend.LinkerFrontend$$anonfun$5.apply(LinkerFrontend.scala:71)
        at org.scalajs.core.tools.linker.frontend.LinkerFrontend$$anonfun$5.apply(LinkerFrontend.scala:71)
        at org.scalajs.core.tools.logging.Logger$class.time(Logger.scala:28)
        at org.scalajs.sbtplugin.Implicits$SbtLoggerWrapper.time(Implicits.scala:9)
```

# Workarounds

- Disable Rhino entirely (`scalaJSUseRhino in Global` := false). This will cause `loadedJSEnv` to depend on the `fastOptJS` task as well and sbt will properly linearize it.
- Disable sbt's parallel execution (`parallelExecution := false`). Exact required configuration and whether this works at all is untested.

# Possible fixes
## Teach sbt to not run the tasks concurrently
[...]

PR #2203, opened 2016-01-26
Protect linker against concurrent usage.

gzm0:
This gives better error diagnosis if the linker is used concurrently. (The sbt plugin is vulnerable to do this. See #2202).


sjrd:
LGTM

PR #2204, opened 2016-01-26
Fix #2202: Prevent concurrent use of linker in sbt

### LINKED ISSUE #2202: Concurrent use of `scalaJSLinker` in sbt plugin

gzm0:
We use sbt's tag mechanism to prevent concurrent execution.

The test failed 60 consecutive times before the change and succeeded 60
consecutive times after it on a 2 core machine.

### DIFF
--- a/sbt-plugin-test/build.sbt
+++ b/sbt-plugin-test/build.sbt
@@ -27,7 +27,11 @@ lazy val noDOM = project.settings(baseSettings: _*).
     scalaJSOutputWrapper := (
         "// Scala.js - noDOM sbt test\n//\n// Compiled with Scala.js\n",
         "// End of Scala.js generated script")
-  )
+  ).
+  /* This hopefully exposes concurrent uses of the linker. If it fails/gets
+   * flaky, there is a bug somewhere - #2202
+   */
+  settings(inConfig(Compile)(run <<= run.dependsOn(fastOptJS, loadedJSEnv)): _*)
 
 lazy val withDOM = project.settings(baseSettings: _*).
   enablePlugins(ScalaJSPlugin).
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -59,6 +59,15 @@ object ScalaJSPluginInternal {
   val scalaJSLinker = SettingKey[ClearableLinker]("scalaJSLinker",
       "Scala.js internal: Setting to persist a linker", KeyRanks.Invisible)
 
+  /** A tag to indicate that a task is using the value of [[scalaJSLinker]]
+   *
+   *  This setting's value should always be retrieved from the same scope than
+   *  [[scalaJSLinker]] was retrieved from.
+   */
+  val usesScalaJSLinkerTag = SettingKey[Tags.Tag]("usesScalaJSLinkerTag",
+      "Scala.js internal: Tag to indicate that a task uses the link or " +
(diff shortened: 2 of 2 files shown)


gzm0:
Review by @sjrd

sjrd:
LGTM wow ... The things one can do with sbt :-p

gzm0:
Yep. To be honest I'm astonished this worked. But they use the tagging mechanisms to limit concurrent task execution internally themselves :)

gzm0:
Even gives readable output when "showing" (added line breaks):
```
[info] 	List(Limit all to 4, Limit forked-test-group to 1,
Limit uses-scalajs-linker-jetty9-compile-fastopt to 1,
Limit uses-scalajs-linker-jetty9-compile-fullopt to 1,
Limit uses-scalajs-linker-jetty9-test-fastopt to 1,
Limit uses-scalajs-linker-jetty9-test-fullopt to 1,
// snip
)
```

PR #2209, opened 2016-01-29
ES6 Promises

sjrd:


sjrd:
Pre-review by @gzm0 and @nicolasstucki 

I would love suggestions on how to test this. [I tried mocking `Promise`](https://github.com/sjrd/scala-js/compare/promises...sjrd:promise-mock), but by the time the tests install the mock, `QueueExecutionContext` has already instantiated its `resolvedUnitPromise`, and therefore the mock is ignored.

mseddon:
LGTM actually.  Certainly an improvement from `scala-js-dom`'s implementation.  `Thenable` is an interesting addition.

The only 'issue' with compatibility with `dom` I can see is that in that library, Promise signatures adopt Scala's `andThen` and `recover` naming, but I've never been particularly keen on that anyway.  Other than that- it's mostly a drop in replacement with slightly tighter typing in certain situations.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -51,4 +54,33 @@ object JSConverters {
...
+    /** Converts the Future to a JavaScript [[Promise]].
+     *
+     *  Attention! The nature of the [[Promise]] class, from the ECMAScript
+     *  specification, makes this method inherently un-typeable, because it is
+     *  not type parametric.
+     *
+     *  The signature of the `toJSPromise` method is only valid
+     *  <i>provided that</i> the values of `A` do not have a `then` method.
+     */
+    def toJSPromise: Promise[A] = {
```
gzm0: I'm wondering if we should implicit foo to actually resolve `Thenable`s properly here. Of course that doesn't give you any guarantee, since your object might have a `then ` method, but at least you can do something if you know:

```scala
val f0: Future[js.Dynamic] = ???
// Assume I know `f0`s result has a `then` method
val f1 = f0.asInstanceOf[Future[Thenable[js.Dynamic]]
f1.toJSPromise // sound
```

sjrd: So you want:

* `(f: Future[Thenable[A]]).toJSPromise` typed as `js.Promise[A]`, and
* `(f: Future[A]).toJSPromise` typed as `js.Promise[A]`?

I guess that makes sense. I'll try to come up with some implicit voodoo.

gzm0: Exactly. Not that much voodoo required. Just a low priority on any and a high on Thenable :)

gzm0:
> I would love suggestions on how to test this.

I think there is a reasonable case to be made that we would want an execution context to capture the current "implementation". That means we need a factory method (and of course a default).

Maybe we should adapt the interface, and allow to create a new `JSQueueExecutionContext` that captures the current implementation of `Promise` / `setTimeout` at usage site. Then the mocking would actually work properly. It would also be useful for people that use mechanisms that modify these parts of the JS runtime system.

sjrd:
Yep, we could do that.

sjrd:
Updated with an additional commit allowing more control over `QueueExecutionContext`s, and then more tests.

sjrd:
Updated. I addressed the magical `Future[Thenable[A]].toJSPromise` case, and I added decent tests.

sjrd:
Note that "Nicer implementation ..." should probably be squashed with its parent commit "Open up the API ..."

sjrd:
Rebased, rearranged commits in a meaningful way. Now ready for a real review.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -51,4 +62,35 @@ object JSConverters {
       arr: scala.Array[T]): JSRichGenTraversableOnce[T] =
     new JSRichGenTraversableOnce(arr)
 
+  @inline
+  implicit def JSRichFutureThenable[A](f: Future[Thenable[A]]): JSRichFuture[A] =
+    new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
```
gzm0: What if `A` itself is a `Thenable`? I thought JS promises flatten all the way...

sjrd: That's why there is a big "Attention unsound!" comment ... These types are only valid when values of `A` don't have a `then` method.

But since a `Thenable[Thenable[A]]` is unsound to begin with, there's no point in dealing with `f: Future[Thenable[Thenable[A]]]`.

gzm0: Why is a `Thenable[Thenable[A]]` unsound? The typing for `Thenable[A]` is unsound if `A <: Thenable[B]`. But `Thenable[Thenable[A]]` should be just fine if the relevant code handles the flattening correctly (and I think it should).

sjrd: `Thenable[Thenable[A]]` is unsound because, assuming you have a `t` of that type, then in `t.then(x => e)`, `x` will in fact be of type `A` (because of flattening). But the type signature of `Thenable.then` says that `x` should be `Thenable[A]`.

Another way to look at it is: by construction, there cannot be any `t: Thenable[Thenable[A]]`. Any attempt to construct a `Thenable[Thenable[A]]` would in fact result in a `Thenable[A]` instead, because of flattening.

sjrd: A good read about these problems is the es-discuss thread: https://esdiscuss.org/topic/promise-cast-and-promise-resolve
In particular, pay attention to Andreas Rossberg's interventions, e.g., at https://esdiscuss.org/topic/promise-cast-and-promise-resolve#content-12

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
```diff
@@ -1,22 +1,55 @@
...
   }
 
-  def reportFailure(t: Throwable): Unit =
-    t.printStackTrace()
+  private final class PromisesExecutionContext extends ExecutionContextExecutor {
+    private val resolvedUnitPromise = new js.Promise[Unit]({
+      (resolve: js.Function1[Unit | js.Thenable[Unit], _],
+          reject: js.Function1[Any, _]) =>
+        resolve(())
+    })
```
gzm0: Why don't you use `js.Promise.resolve[Unit](())` here?

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
```diff
@@ -1,22 +1,55 @@
...
+  def promises(): ExecutionContextExecutor =
+    new PromisesExecutionContext
+
+  def apply(): ExecutionContextExecutor =
+    if (js.isUndefined(js.Dynamic.global.Promise)) timeouts()
+    else promises()
+
+  private final class TimeoutsExecutionContext extends ExecutionContextExecutor {
+    def execute(runnable: Runnable): Unit = {
+      js.Dynamic.global.setTimeout({ () =>
```
gzm0: This is not consistent with `PromisesExecutionContext` as it doesn't capture `setTimeout`. I don't know how much sense it makes to do so, but maybe it is a good idea for consistency.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/AsyncTest.scala
```diff
@@ -136,4 +199,86 @@ object AsyncTest extends JasmineTest {
...
+      PromiseMock.withMockedPromise { processQueue =>
+        implicit val ec = QueueExecutionContext.promises()
+
+        val p = new js.Promise[Int]({
+          (resolve: js.Function1[Int | js.Thenable[Int], _], reject: js.Function1[Any, _]) =>
+            resolve(42)
+        })
+
+        val f = p.toFuture
+        val fAssertType: Future[Int] = f
```
gzm0: Hahahaha, local type inference FTW :) Try doing this in a global inferencer without influencing the `p.toFuture` call.

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
sjrd: Hum, there is no good reason to capture `setTimeout` here. The other guy "accidentally" captures promises because it wants to allocate exactly one promise for `()`. Not because it caches the `Promise` constructor per se.

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
sjrd: Because it didn't cross my mind ^^ I should do that indeed.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
```diff
@@ -0,0 +1,225 @@
...
+
+    private[this] var fulfillReactions = js.Array[js.Function1[A, Any]]()
+    private[this] var rejectReactions = js.Array[js.Function1[Any, Any]]()
+
+    init(executor)
+
+    private[this] def init(
+        executor: js.Function2[js.Function1[A | Thenable[A], _], js.Function1[scala.Any, _], _]) = {
+      tryCatchAny[Unit] {
+        executor(resolve _, reject _)
```
gzm0: Shouldn't the executor also run on the event queue?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
```diff
@@ -0,0 +1,225 @@
...
+    private[this] def fulfill(value: A): Unit = {
+      assert(state == Pending)
+      state = Fulfilled(value)
+      clearAndTriggerReactions(fulfillReactions, value)
+    }
+
+    private[this] def clearAndTriggerReactions[A](
+        reactions: js.Array[js.Function1[A, Any]],
+        argument: A): Unit = {
+
```
gzm0: `assert(state != Pending)`?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
sjrd: No, I checked. It's in the spec for the `Promise` constructor. On mobile right now so I can't give a link.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
gzm0: Ok... I got this now... \*hairpull\*
Thanks for your patience :P

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
gzm0: I know. But as it is now, there is no way to have a "stable" execution context. I.e. one that doesn't depend on global state. Usually it is not good if stuff depends on global state. Since we don't really have a choice here, we could go with the nicer: "We capture state when you call xyz", rather than: "State just updates under the hood, tough luck, pal"...

OTOH, I don't know if there are VMs that do not like if one captures `setTimeout` (also, it is not evident whether we must capture global to have a proper `this` argument. And if we do, the whole thing is kind of bogus.).

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
gzm0: Ok. Fair enough.

gzm0:
That's all. I haven't validated this with the ES6 spec. The implementation just seems reasonable. Please let me know if you want me to validate with the spec.

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
sjrd: Also, `setTimeout` is not *supposed* to be be global state. Other than mocking, no reasonable code should ever change `setTimeout`.

By the same argument, we should capture `Math` in `java.lang.Math`, `Date` in `java.util.Date`, `Object` in ... er, boom! So many places we should capture `Object` ...

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
sjrd: Yep, makes sense.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/PromiseMock.scala
sjrd: http://www.ecma-international.org/ecma-262/6.0/#sec-promise-executor

I'll add the reference as a comment on `def init`, like I did for some other methods.

library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala
gzm0: Ok. Agreed :)

sjrd:
Updated.

> Please let me know if you want me to validate with the spec.

Only if you want more hair pulling :-p I don't think that's necessary.

gzm0:
LGTM

sjrd:
Ping @nicolasstucki ?

nicolasstucki:
LGTM

PR #2213, opened 2016-01-31
Fix #2212: Fix printing of optimizer hints

### LINKED ISSUE #2212: IRTreePrinter prints closing paren in front of class name
Culprit is this (refactored) `if`:
https://github.com/scala-js/scala-js/blob/master/ir/src/main/scala/org/scalajs/core/ir/Printers.scala#L758-L760

gzm0:


gzm0:
Review by @sjrd

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -754,10 +754,7 @@ object Printers {
 
         case tree: ClassDef =>
           val ClassDef(name, kind, superClass, interfaces, jsName, defs) = tree
-          if (tree.optimizerHints != OptimizerHints.empty)
-            print("@hints(")
-            print(tree.optimizerHints.bits)
-            print(") ")
+          print(tree.optimizerHints)
```
sjrd: Fail! ^^

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -895,6 +888,15 @@ object Printers {
     protected def print(c: Int): Unit =
       out.write(c)
 
+    protected def print(optimizerHints: OptimizerHints)(
+        implicit dummy: DummyImplicit): Unit = {
+      if (optimizerHints != OptimizerHints.empty) {
+        print("@hints(")
+        print(optimizerHints.bits)
```
sjrd: Hum that's weird too, actually. Isn't it calling the `print(c: Int)` just above? There's probably a `.toString()` missing.

sjrd:
That's all.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
gzm0: Yep:

```
@hints() def productionMode__Z(): boolean
```

gzm0:
Updated. Tested locally.

sjrd:
LGTM

PR #2214, opened 2016-01-31
Minimize helloworld

sjrd:
|fastOptJS|fullOptJS
---|---|---
Before | 100.333 | 17.666
After | 91.671 | 12.902
Ratio | 91% | 73%


sjrd:
Review by @nicolasstucki

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2972,16 +2973,6 @@ private[optimizer] abstract class OptimizerCore(
...
 
-  private def finishTransformCheckNull(preTrans: PreTransform)(
-      implicit pos: Position): Tree = {
-    if (preTrans.tpe.isNullable) {
-      val transformed = finishTransformExpr(preTrans)
-      CallHelper("checkNonNull", transformed)(transformed.tpe)
-    } else {
-      finishTransformExpr(preTrans)
-    }
-  }
```
sjrd: In case you wonder, this method was already dead code.

gzm0:
Nice! I just looked at this in preparation for my talk and thought there is still to much we reach :)
When do we start dce-ing the scalajsenv itself? ;)

sjrd:
> When do we start dce-ing the scalajsenv itself? ;)

We probably don't need to do that. So far it seems Closure dce's scalajsenv just fine.

... unless of course we want to get rid of Closure :-p But we're not ready for that, yet.

sjrd:
If I perform dce "manually" on the pretty-printed version of helloworld-opt.js, I go from 17.304 to 15.236. So potentially we can still get a ratio of 88% if we dce scalajsenv ourselves.

By manual dce I mean I progressively removed stuff that was "obviously" not used anywhere, as in, it would be pretty obvious for a reachability analysis to "see" it.

Main thing that goes away: Scala arrays. Currently they are retained because of `$TypeData["newArrayOfThisClass"]`, which is always kept by Closure because it's a bracket access. But with our own dce we could remove it when we know that `java.lang.Class.newArrayOfThisClass` is not reachable. We can similarly remove `"getSuperclass"`, `"getComponentType"` and `"getFakeInstance"`.

The biggest remaining sources of "legit" baggage are:
* `Object.toString`. That awful thing uses `GetClass`, `Class.{getName,isInterface,isPrimitive}`, and worst of all `Object.hashCode`, which reaches all `hashCode`s in the world, including `identityHashCode`.
* `identityHashCode` needs the hash code computation of `Double`s, which is the awful polyfill for `doubleToLongBits`
* `RuntimeLong.toString` cannot be eliminated, and it uses part of the division algorithm.

To go further, one thing we could do is provide a semantics flag to break `Object.toString`. This would liberate us from hash codes and most of `java.lang.Class` (maybe all of it if we get our dce right).

scalalib/overrides/scala/util/DynamicVariable.scala
```diff
@@ -0,0 +1,70 @@
...
+   * This removes the dependency on ThreadLocal and InheritableThreadLocal from the
+   * Hello World.
+   */
+  private[this] var v = init
+
+  /** Retrieve the current value */
+  def value: T = v
+
+  /** Set the value of the variable while executing the specified
+    * thunk.
```
nicolasstucki: Alignment issue.

scalalib/overrides/scala/util/DynamicVariable.scala
```diff
@@ -0,0 +1,70 @@
...
+  def withValue[S](newval: T)(thunk: => S): S = {
+    val oldval = v
+    v = newval
+
+    try thunk
+    finally v = oldval
+  }
+
+  /** Change the currently bound value, discarding the old value.
+    * Usually withValue() gives better semantics.
```
nicolasstucki: Alignment issue.

scalalib/overrides/scala/util/DynamicVariable.scala
sjrd: It's not our code. This comes from the original file in scala/scala.

scalalib/overrides/scala/util/DynamicVariable.scala
sjrd: Same.

tools/scalajsenv.js
```diff
@@ -272,7 +264,7 @@ ScalaJS.objectGetClass = function(instance) {
       return ScalaJS.d.sr_BoxedUnit.getClassOf();
     default:
       if (instance === null)
-        ScalaJS.throwNullPointerException();
+        return instance.getClass__jl_Class();
```
nicolasstucki: Why is this change from a throw to a return?

scalalib/overrides/scala/util/DynamicVariable.scala
nicolasstucki: Ok

tools/scalajsenv.js
sjrd: `instance.getClass__jl_Class()` will necessarily throw because `instance` is `null` at this point. The `return` is therefore dead code, strictly speaking, but it is there by symmetry with the other branches. This is a pattern that is often used in this file. See e.g. in `objectClone` just below.

nicolasstucki:
LGTM

PR #2221, opened 2016-02-02
Fix #2220: Deal with 2.12 Function lambdas inside JS classes.

### LINKED ISSUE #2220: Compiler crash in 2.12 with a lambda inside a Scala.js-defined JS class
`patchFunBodyWithBoxes` has a spurious assertion at line [`GenJSCode.scala`:4438](https://github.com/scala-js/scala-js/blob/v0.6.6/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L4438):
```scala
assert(methodSym.isClassConstructor == isScalaJSDefinedJSClass(methodSym.owner))
```
This is in fact not true when there is a lambda inside a method of a Scala.js-defined JS class. But it doesn't need to be true. It's fine for the lhs to be false and the rhs be true. In fact, I think since the support for overloaded constructors, the lhs should *always* be false.

sjrd:
There were two problems:

* A wrong assertion.
* Calling the method as if it were a Scala method (rather than
  a non-exposed JS method, which is in fact a static method).


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2230, opened 2016-02-06
Fix #2005: Add cross-version shared source directories.

### LINKED ISSUE #2005: Scala cross-version support for shared source directory
Since version 0.13.8, sbt adds Scala version specific source directories to `unmanagedSources`, see https://github.com/sbt/sbt/pull/1799 or https://github.com/sbt/sbt/blob/0.13.8/notes/0.13.8.markdown#cross-version-support-for-scala-sources. That means that besides `src/main/scala` `src/main/scala-2.x` is added to the list of source directories. This is extremely useful for projects that target multiple Scala versions but want to use different sources per version. 

Unfortunately a Scala.js `CrossProject` with `CrossType.Full` does not yet add Scala version specific directories for the `shared` source directory. This is `unmanagedSourceDirectories` for one of my projects:
```
> coreJVM/unmanagedSourceDirectories
[info] List(
  $PROJECT_ROOT/core/jvm/src/main/scala-2.11,
  $PROJECT_ROOT/core/jvm/src/main/scala,
  $PROJECT_ROOT/core/jvm/src/main/java,
  $PROJECT_ROOT/core/shared/src/main/scala)
> coreJS/unmanagedSourceDirectories
[info] List(
  $PROJECT_ROOT/core/js/src/main/scala-2.11,
  $PROJECT_ROOT/core/js/src/main/scala,
  $PROJECT_ROOT/core/js/src/main/java,
  $PROJECT_ROOT/core/shared/src/main/scala)
```
Ideally both lists should also contain `$PROJECT_ROOT/core/shared/src/main/scala-2.11`.

sjrd:


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2231, opened 2016-02-06
Fix #2197: Annotate `@JSName` with `@field @getter @setter`.

### LINKED ISSUE #2197: @JSName fails for field in trait
```
@ScalaJSDefined
trait Stuff extends js.Object {
  @JSName("id")
  val stuffId: Int
}
```

Gives warning message in SJS 0.6.5:
```
no valid targets for annotation on value programId - it is discarded unused. You may specify targets with meta-annotations, e.g. @(scala.scalajs.js.annotation.JSName("id") @getter)
[warn]   @JSName("id")
```

Tests are present in https://github.com/scala-js/scala-js/blob/master/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala#L648-L687, but does not cover traits.

Error is also present for `abstract class`.

sjrd:
This causes it to be automatically replicated to the getter and/or
setter of fields. This is especially important for abstract vals
and vars in Scala.js-defined JS traits, which do not have any
field on which the Scala compiler could put the annotation.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -672,8 +672,12 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       /* If this is an accessor, copy a potential @JSName annotation from
        * the field since otherwise it will get lost for traits (since they
        * have no fields).
+       *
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2232, opened 2016-02-06
Optimize bracket-select with a valid identifier as dot-select.

### LINKED ISSUE #2217: this["require"]() and fastOptJS
As explained on https://github.com/facebook/react-native/issues/465 some scala-js interfaces cannot be used with `fastOptJS` due to lack of `this["require"]()` syntax support. Any chance to have now a solution on scala-js side?

sjrd:
This is done by the default back-end. The Closure back-end explicitly deactivates this optimization, as it would break the advanced optimizations of Closure.

Fix #2217 as a side-effect.


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2245,14 +2247,15 @@ private[emitter] object JSDesugaring {
...
     if (linkedClass.kind == ClassKind.JSClass) {
       encodeClassVar(linkedClass.encodedName)
     } else {
       require(linkedClass.jsName.isDefined)
       linkedClass.jsName.get.split("\\.").foldLeft(envField("g")) {
-        (prev, part) => js.BracketSelect(prev, js.StringLiteral(part))
+        (prev, part) => classEmitter.genBracketSelect(prev, js.StringLiteral(part))
       }
     }
   }
```
gzm0: Move this into an instance of `JSDesugaring`. Refactor call sites as necessary (or create a forwarder if really necessary).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -66,6 +70,28 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+         * specificities of eval().
+         */
+        js.DotSelect(qual, js.Ident(name))
+      case _ =>
+        js.BracketSelect(qual, item)
+    }
+  }
+
+  private[emitter] def genIdentBracketSelect(qual: js.Tree, item: String)(
+      implicit pos: Position): js.Tree = {
```
gzm0: `require(item != "eval")`? Seems like thing that could be easily missed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -66,6 +70,28 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    if (internalOptions.optimizeBracketSelects)
+      js.DotSelect(qual, js.Ident(item))
+    else
+      js.BracketSelect(qual, js.StringLiteral(item))
+  }
+
```
gzm0: Actually, I don't understand why these methods are in `ScalaJSClassEmitter`. They have nothing to do with class emission specifically and it shows: The fact that you need to pass a `ScalaJSClassEmitter` to a `JSDesugaring` to generate constructors (for merely calling back the `emitter` again...).

sjrd: I could move them to `object JSDesugaring` with an explicit parameter `internalOptions`, if you prefer.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
sjrd: OK.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: OK. 

The reason is here in the first place is that the class emitter also needs it. So yes, a forwarder will be necessary.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
gzm0: I think it is reasonable to make `JSDesugaring` a class one has to construct with `internalOptions` as parameter. WDYT?

sjrd: Yes, let's do that. Good idea.

sjrd:
Updated (and rebased).

gzm0:
LGTM

sjrd:
retest this please

PR #2233, opened 2016-02-06
Fix #2222: Restore statesInUse when we do a rollback.

### LINKED ISSUE #2222: Linker/optimiser performance in 0.6.6
Linker/optimiser is doing weird stuff and taking a *looong* time.

Here is a graph of CPU load x time:
CPU load x time

Here is a stack trace when low:
[stack2.txt](https://github.com/scala-js/scala-js/files/114981/stack2.txt)

Here is a stack trace when high:
[stack3.txt](https://github.com/scala-js/scala-js/files/114982/stack3.txt)

You could probably reproduce this - I was working on scalajs-react and using this command: 
```
~test/test-only -- japgolly.scalajs.react.experimental
```

sjrd:
This contains the growth of the `statesInUse` list for methods performing a lot of rollbacks.

### DIFF
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -6,6 +6,9 @@ object BinaryIncompatibilities {
   )
 
   val Tools = Seq(
+      // private, not an issue
+      ProblemFilters.exclude[MissingMethodProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.OptimizerCore#RollbackException.this")
   )
 
   val JSEnvs = Seq(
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0 but **do not merge**. I've haven't yet *thought* about this. I only tested. It seems not to break anything and it definitely improves/solves the mentioned bug. It literally divides the "Optimizer part" running time by 3 on the problematic codebase.

sjrd:
OK now I have thought about it. I'm pretty sure it's fine. states created during a `try` cannot be reused if that try is rollbacked. It's like `freshLocalName`.

gzm0:
LGTM

PR #2235, opened 2016-02-11
Towards 0.6.8.

sjrd:
FTR, the Towards 0.6.7 commit was e9b19e32fb8a96c8a14e57894e5054083f06194d

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
@@ -10,7 +10,7 @@ object ScalaJSVersions {
    */
 
   /** Scala.js version. */
-  val current: String = "0.6.7"
+  val current: String = "0.6.8-SNAPSHOT"
 
   /** true iff the Scala.js version is a snapshot version. */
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @nicolasstucki for fast-forward on master. Do not merge.

ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
```diff
@@ -10,7 +10,7 @@ object ScalaJSVersions {
    */
 
   /** Scala.js version. */
-  val current: String = "0.6.7"
+  val current: String = "0.6.8-SNAPSHOT"
```
nicolasstucki: I'm guessing that `binarySupported` was not updated on purpose. Is that right?

nicolasstucki:
That is all. Otherwise LGTM.

ir/src/main/scala/org/scalajs/core/ir/ScalaJSVersions.scala
sjrd: Yes. 0.6.7 is forward binary compatible with 0.6.6. It emits the same version of the IR.

PR #2236, opened 2016-02-11
Fix #2226: Remove timeout from PhantomJSEnv socket connection.

### LINKED ISSUE #2226: JsEnv timeouts start too early
My travis builds often fail because  PhantomJSEnv timeouts halfway through the tests. The fix for this was to change my `.travis.yml` from
```yml
script:
- sbt ++$TRAVIS_SCALA_VERSION test
```
to
```yml
script:
- sbt ++$TRAVIS_SCALA_VERSION test:fastOptJS
- sbt ++$TRAVIS_SCALA_VERSION test
```

This, coupled with the fact that I never get timeouts locally in long-running SBT sessions, implies that whatever timeout is in-place (*I know we've discussed this before but I never come away with a proper understanding - sorry*) begin when the batch of commands are issued. This should ideally be changed to begin when tests truly commence.

That being said, the workaround is trivial so if a fix isn't worth the effort, that's cool - feel free to close.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

gzm0:
As I already stated in the relevant issue, I heavily disagree with having a timeout by default.

If we implement timeout logic, it should be implemented in general for `ComJSRunner` (probably on a per-message retrieval level).
For asynchronous testing, deciding timeouts should be entirely left up to the testing framework (maybe with support from the testing interface runtime). How long a test takes is not a property of your build, its a property of the test and it belongs there.

nicolasstucki:
I could simply leave the [line](https://github.com/scala-js/scala-js/pull/2236/files#diff-934daf20d0bc5326d2b1e6e11a55b09dR51) as `connection.setMaxIdleTime(-1)` and remove the rest.

gzm0:
> I could simply leave the line as connection.setMaxIdleTime(-1) and remove the rest.

I think that would actually be more appropriate (assuming Jetty then actually doesn't timeout).

@sjrd?

sjrd:
As I said on the issue, yes, I agree that having no timeout at all is the most consistent thing to do.

Also, this should be rebased on master now that the Towards 0.6.8 commit is there.

sjrd:
LGTM

PR #2237, opened 2016-02-11
Test environment with sys props

nicolasstucki:


nicolasstucki:
Review by @sjrd

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -6,7 +6,35 @@ import scala.scalajs.js
...
 object System {
+
+  private[this] var systemProperties = loadSystemProperties()
+
+  private def loadSystemProperties(): ju.Properties = {
+    val sysProp = new ju.Properties()
+    sysProp.setProperty("java.version", "1.8")
+    sysProp.setProperty("java.vm.specification.version", "1.8")
+    sysProp.setProperty("java.vm.specification.vendor", "Oracle Corporation")
+    sysProp.setProperty("java.vm.specification.name", "Java Virtual Machine Specification")
```
gzm0: What's the difference between the Java specification and the JVM specification?

sjrd: See https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getProperties%28%29

`java.vm.specification.version` is the version of the [JVMS](https://docs.oracle.com/javase/specs/) that is supported by this implementation.

`java.version` is the version of the JRE, i.e., the implementation itself. That is most usefully understood as the version of the JDK being implemented.

sjrd: For example, before we implemented default methods, we should have used 1.7 for the VM specification version. But still 1.8 for the JRE, since we do (strive to) implement the methods of JDK 8.

gzm0: Oh, OK. I see. And there is no VM version? Strange...

gzm0: Ah, there is. We should set it :)

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -6,7 +6,35 @@ import scala.scalajs.js
 import js.Dynamic.global
 import scala.scalajs.LinkingInfo.assumingES6
 
+import java.{util => ju}
+import scalajs.runtime.environmentInfo
+
 object System {
+
+  private[this] var systemProperties = loadSystemProperties()
```
sjrd: This is bad for code size, as any `println` call will cause this code to be reachable.

You'll have to use a "manual" lazy initialization, here:
```scala
private[this] var systemProperties: ju.Properties = _

def getProperties(): ju.Properties = {
  if (systemProperties == null)
    systemProperties = loadSystemProperties()
  systemPropertis
}
```
Alternatively, you can put this `var` in an `object` of its own (similarly to `IDHashCode`)

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -207,11 +235,40 @@ object System {
...
+    systemProperties.remove(key).asInstanceOf[String]
+  }
+
+  def setProperty(key: String, value: String): String = {
+    // Option(getSecurityManager()).foreach(_.checkPermission(ju.PropertyPermission(key, "write")))
+    systemProperties.setProperty(key, value).asInstanceOf[String]
+  }
+
+  // If implemented, commented code in property related methods must be uncommented
+  //  def getSecurityManager(): SecurityManager
```
sjrd: This is misleading. There are zillions of places in the already implemented parts of the JDK that assume no security manager. This comment suggests that only the above methods will have to be updated, but that's not true.

I recommend removing all comments related to security managers.

javalib/src/main/scala/java/util/Hashtable.scala
```diff
@@ -0,0 +1,96 @@
...
+
+import scala.collection.mutable
+import scala.collection.JavaConversions._
+
+class Hashtable[K, V] private (inner: mutable.HashMap[Box[Any], V])
+    extends ju.Dictionary[K,V] with ju.Map[K, V] with Cloneable with Serializable {
+
+  def this() = {
+    this(mutable.HashMap.empty[Box[Any], V])
+  }
```
sjrd: No `{}`

javalanglib/src/main/scala/java/lang/System.scala
nicolasstucki: I do set it, but the code is in another place because I load it directly from the plugin:
https://github.com/scala-js/scala-js/pull/2237/files#diff-f74255631d0aae4a116f32f2cb6385ccR643

gzm0: And if we are not running with sbt the version is just automagically gone? Probably not a good idea. I think we should propagate the version through the linking infos.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +634,29 @@ object ScalaJSPluginInternal {
     }
   }
 
+  protected def javaSysProps2JSLib(javaOptions: Seq[String]): VirtualJSFile = {
+    val javaSysPropsPattern = "-D([^=]*)=(.*)".r
+    val javaSysProps = javaOptions.map {
+      case javaSysPropsPattern(propName, propValue) => (propName, propValue)
+    }
+    val defaultSysProps = List(
+      "java.vm.version" -> ScalaJSVersions.current,
```
sjrd: As @gzm0 said, this should go through linking info, as it's dependent on the tools; not the sbt plugin, and it should be available even we run in a different way.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +634,29 @@ object ScalaJSPluginInternal {
...
+      "os.name" -> java.lang.System.getProperty("os.name"),
+      "os.arch" -> java.lang.System.getProperty("os.arch"),
+      "os.version" -> java.lang.System.getProperty("os.version")
```
sjrd: `os` stuff should not be present, IMO. They would disappear if we run with something else than the sbt plugin. But there's no way we can set that from the tools, either. So I would just not support them.

gzm0: +1

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -0,0 +1,64 @@
...
+  // def storeToXML(os: OutputStream, comment: String, encoding: String): Unit
+
+  def getProperty(key: String): String = {
+    get(key) match {
+      case value: String => value
+
+      case _ =>
+        if (defaults != null) defaults.getProperty(key)
+        else null
+    }
```
sjrd: I think it would make more sense to implement `getProperty(value)` in terms of `getProperty(value, defaultValue = null)`, and have the latter call `defaults.getProperty(value, defaultValue)`.

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -0,0 +1,64 @@
...
+    JavaConversions.asJavaEnumeration(thisSet.iterator() ++ filteredDefaults)
+  }
+
+  def stringPropertyNames(): ju.Set[String] = {
+    val set = new ju.HashSet[String]
+    propertyNames().foreach {
+      case key: String => set.add(key)
+      case _           => // Ignore key
+    }
+    ju.Collections.synchronizedSet(set)
```
sjrd: uh? Why not just `set`?

library/src/main/scala/scala/scalajs/runtime/EnvironmentInfo.scala
```diff
@@ -51,4 +51,9 @@ trait EnvironmentInfo extends js.Object {
   def sourceMapper: js.UndefOr[js.Function1[ // scalastyle:ignore
     js.Array[JSStackTraceElem], js.Array[JSStackTraceElem]]] = js.native
 
+  /** Method used to insert Java system properties.
```
sjrd: It's not a method; it's a dictionary. I suggest:

> Dictionary of system properties to add to `java.lang.System.getProperties()`

project/Build.scala
```diff
@@ -1069,6 +1069,49 @@ object Build extends sbt.Build {
   // Testing
 
   val testTagSettings = Seq(
+      javaOptions in Test ++= {
```
sjrd: The common logic between `javaOptions` and `testOptions` should somehow be factored out.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -1,5 +1,6 @@
 package org.scalajs.sbtplugin
 
+import sbt.KeyRanks._
```
sjrd: Is this used?

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -575,7 +576,8 @@ object ScalaJSPluginInternal {
 
       loadedJSEnv <<= Def.taskDyn {
         val log = streams.value.log
-        val libs = resolvedJSDependencies.value.data
+        val libs = resolvedJSDependencies.value.data :+
+            ResolvedJSDependency.minimal(javaSysProps2JSLib(javaOptions.value))
```
sjrd: Could we avoid adding any `VirtualFile` if there turns out not to be any `-D` option? I guess `javaSysProps2JSLib` should return an `Option[VirtualJSFile]`

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +634,29 @@ object ScalaJSPluginInternal {
...
+    val defaultSysProps = List(
+      "java.vm.version" -> ScalaJSVersions.current,
+      "os.name" -> java.lang.System.getProperty("os.name"),
+      "os.arch" -> java.lang.System.getProperty("os.arch"),
+      "os.version" -> java.lang.System.getProperty("os.version")
+    )
+    val allProperties = (defaultSysProps.iterator ++ javaSysProps).toMap
+    val formattedProps = allProperties.map {
+      case (propName, propValue) =>
+        "\"" + escapeJS(propName) + "\": \"" + escapeJS(propValue) + '"'
```
sjrd: I suggest `"\""` at the end too, to be consistent.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+    assertNotNull(get("java.vm.version"))
+    assertNotNull(get("os.name"))
+    assertNotNull(get("os.version"))
+    assertNotNull(get("os.arch"))
+    assertNotNull(get("scalajs.compliant-moduleinit"))
+    assertNotNull(get("scalajs.compliant-asinstanceofs"))
+    assertNotNull(get("scalajs.production-mode"))
+
+    val inNode = get("scalajs.node") == "true"
+    val inPhantom = get("scalajs.phantom") == "true"
```
sjrd: `inPhantomJS` and `scalajs.phantomjs`
Contrary to Node.js, PhantomJS is always spelled with the JS

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+    val inRhino = get("scalajs.rhino") == "true"
+    val hasTypedArrays = get("scalajs.typedarray") == "true"
```
sjrd: `hasTypedArrays` is misleading. The property `typedarray` is enabled independently of whether the VM has typed arrays. It is set when we want to test typed array tests. It is possible that the VM supports typed arrays but `scalajs.typedarray` is `false`.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+
+    if (inNode) {
+      val process = js.Dynamic.global.process
+      assertFalse(js.isUndefined(process))
+      assertEquals("node", process.title)
```
sjrd: I am pretty sure this can be wrong. For example if Node.js is launched via a symbolic link `nodejs`. This test should be removed.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+      assertFalse(inPhantom || inRhino)
+      assertTrue(hasTypedArrays)
+    } else if (inPhantom) {
+      assertFalse(js.isUndefined(js.Dynamic.global.callPhantom))
+      assertFalse(inNode || inRhino)
+      assertFalse(hasTypedArrays)
+    } else if (inRhino) {
+      // TODO find some platform specific test for Rhino
```
sjrd: The existence of `global.Packages`

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+      assertFalse(inNode || inPhantom)
+      assertFalse(hasTypedArrays)
+    } else {
+      fail("No platform tag found.")
+    }
+
+    assertEquals(inNode, Platform.executingInNodeJS)
+    assertEquals(inPhantom, Platform.executingInPhantomJS)
+    assertEquals(inRhino, Platform.executingInRhino)
+    assertEquals(hasTypedArrays, Platform.hasTypedArrays)
```
sjrd: The last assert is wrong, I think, for the same reason about typed arrays.

test-suite/jvm/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
```diff
@@ -15,4 +15,18 @@ object Platform {
...
+  def hasTypedArrays: Boolean = false
+
+  def isInFastOpt: Boolean = false
+  def isInFullOpt: Boolean = false
+  def isInStrongMode: Boolean = false
+  def isInProductionMode: Boolean = false
+
+  def hasCompliantAsInstanceOfs: Boolean = true
+  def hasCompliantModule: Boolean = true
+  def hasStrictFloats: Boolean = true
```
sjrd: Are these used?

js-envs/src/main/scala/org/scalajs/jsenv/rhino/RhinoJSEnv.scala
```diff
@@ -362,14 +362,22 @@ final class RhinoJSEnv private (
...
+        case oldScalaJSenv: Scriptable =>
+          oldScalaJSenv
+      }
 
       scalaJSenv.addFunction("sourceMapper", args => {
         val trace = Context.toObject(args(0), scope)
         loader.mapStackTrace(trace, context, scope)
       })
-
-      ScriptableObject.putProperty(scope, "__ScalaJSEnv", scalaJSenv)
```
gzm0: I think this change actually deserves a bug, a separate fix/commit and a test.

sjrd:
That's all for now.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: I think assembling this file should be a separate task. The current implementation harms composability.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +634,29 @@ object ScalaJSPluginInternal {
     }
   }
 
+  protected def javaSysProps2JSLib(javaOptions: Seq[String]): VirtualJSFile = {
+    val javaSysPropsPattern = "-D([^=]*)=(.*)".r
+    val javaSysProps = javaOptions.map {
+      case javaSysPropsPattern(propName, propValue) => (propName, propValue)
+    }
```
gzm0: I don't like this. It gives the impression that we support all Java options. Either:

- Add your own key (safe, less compatible)
- Fail if you encounter options we don't support (technically a breaking change, practically not a problem since we forced people to have `fork := false` so `javaOptions` was useless anyways).

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,56 @@ class SystemJSTest {
...
+    assertEquals("1.8", get("java.version"))
+    assertEquals("1.8", get("java.vm.specification.version"))
+    assertEquals("Oracle Corporation", get("java.vm.specification.vendor"))
+    assertEquals("Scala.js", get("java.vm.name"))
+    assertEquals("1.8", get("java.specification.version"))
+    assertEquals("/", get("file.separator"))
+    assertEquals(":", get("path.separator"))
+    assertEquals("\n", get("line.separator"))
+
+    // Added via __ScalaJSEnv
```
gzm0: I think it is worth mentioning from where they are added to `__ScalaJSEnv`.

javalib/src/main/scala/java/util/Properties.scala
nicolasstucki: Agreed

javalib/src/main/scala/java/util/Properties.scala
nicolasstucki: I was just following the JVM implementation:
```scala
scala> p
res5: java.util.Properties = {b=2, a=1}

scala> p.stringPropertyNames.getClass
res6: Class[_ <: java.util.Set[String]] = class java.util.Collections$SynchronizedSet
```

sjrd: I would follow the spec, here. It doesn't say anything about synchronization. Since we don't need synchronizing anyway, I would just return `set` and avoid depending on all the code that hides behind `synchronizedSet()`.

test-suite/jvm/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
nicolasstucki: Not yet, but they will allow me to port the jasmine tests that use these conditions.

project/Build.scala
nicolasstucki: There would be only:

```scala
val sems = (scalaJSStage in Test).value match {
  case FastOptStage => (scalaJSSemantics in (Test, fastOptJS)).value
  case FullOptStage => (scalaJSSemantics in (Test, fullOptJS)).value
}
```
And it would change all line indentations there. Do you really want me to do itt?

sjrd: I guess we don't have the same definition of "common logic" :p

What I'm saying is that both `javaOptions` and `testOptions` perform the same tests to then decide whether or not a "flag" is on or off. For example, regarding `"fastopt-stage"` and `"fullopt-stage"`, in `javaOptions` we have:
```scala
        val stageProp = (scalaJSStage in Test).value match {
          case FastOptStage => Seq(scalaJSProp("fastopt-stage", "true"))
          case FullOptStage => Seq(scalaJSProp("fullopt-stage", "true"))
        }
```
and in `testOptions` we have:
```scala
        val stageTag = Tests.Argument((scalaJSStage in Test).value match {
          case FastOptStage => "-tfastopt-stage"
          case FullOptStage => "-tfullopt-stage"
        })
```
The test and the flags that are produced are the same.

There could be a custom task returning a `Set[String]` (the enabled flags, e.g., `Set("fastopt-stage")`), which is computed based on the common logic. It is then straightforward for `javaOptions` and `testOptions` to convert this `Set[String]` into `-D` options and `Test.Argument`s, respectively.

js-envs/src/main/scala/org/scalajs/jsenv/rhino/RhinoJSEnv.scala
nicolasstucki: Added as a fix for #2239.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,58 @@ class SystemJSTest {
...
+    assertEquals("1.8", get("java.vm.specification.version"))
+    assertEquals("Oracle Corporation", get("java.vm.specification.vendor"))
+    assertEquals("Scala.js", get("java.vm.name"))
+    assertEquals("1.8", get("java.specification.version"))
+    assertEquals("/", get("file.separator"))
+    assertEquals(":", get("path.separator"))
+    assertEquals("\n", get("line.separator"))
+    assertNotNull(get("java.vm.version"))
+
+    // Added via __ScalaJSEnv in ScalaJSPluginInternal
```
sjrd: Hum ... It's `Build.scala`, not `ScalaJSPluginInternal.scala`, is it?

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,58 @@ class SystemJSTest {
...
+    if (inNode) {
+      val process = js.Dynamic.global.process
+      assertFalse(js.isUndefined(process))
+      assertFalse(inPhantomJS || inRhino)
+    } else if (inPhantomJS) {
+      assertFalse(js.isUndefined(js.Dynamic.global.callPhantom))
+      assertFalse(inNode || inRhino)
+    } else if (inRhino) {
+      // TODO find some platform specific test for Rhino
+      assertFalse(inNode || inPhantomJS)
```
sjrd: Apparently my comment disappeared, so here it is again: you can test for the presence of `js.Dynamic.global.Packages`.

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -207,11 +210,54 @@ object System {
...
+    }
+  }
+
+  def getProperties(): ju.Properties =
+    SystemProperties.value
+
+  def setProperties(properties: ju.Properties): Unit = {
+    if (properties == null)
+      throw new scala.NullPointerException
+    SystemProperties.value = properties
```
sjrd: JavaDoc says:
> If the argument is null, then the current set of system properties is forgotten.

Experimentation suggests that it does recreate a new set out of `loadSystemProperties()` (not an empty map). So it should probably be:
```scala
SystemProperties.value =
  if (properties != null) properties
  else SystemProperties.loadSystemProperties()
```

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +633,24 @@ object ScalaJSPluginInternal {
     }
   }
 
+  protected def javaSysProps2JSLib(javaOptions: Seq[String]): Option[VirtualJSFile] = {
```
sjrd: This should probably be in its own `Task`, so that it can be reused/overridden by users. It seems a bit too monolithic right now.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,58 @@ class SystemJSTest {
       assertEquals(x1FirstHash, System.identityHashCode(x1))
     }
   }
+
+  @Test def systemProperties(): Unit = {
+    def get(key: String): String = java.lang.System.getProperty(key)
+
+    // Defined in System.scala
+    assertEquals("1.8", get("java.version"))
```
sjrd: In the same commit as those, there should be one hard-coded one `-D` property in the `javaOptions` of the test-suite to test that the logic in `javaSysProps2JSLib` works.

library/src/main/scala/scala/scalajs/LinkingInfo.scala
```diff
@@ -108,4 +108,17 @@ object LinkingInfo {
...
+   *  constant-folding, inlining, and other local optimizations can be
+   *  leveraged with this "constant" to write polyfills that can be
+   *  dead-code-eliminated.
+   *
+   *  Used to define the `java.vm.version` property in `java.lang.System`.
+   *
+   */
+  @inline
+  def version: String =
+    linkingInfo.version
```
sjrd: Since it is accessible with `java.vm.version`, I wouldn't expose this here at all.

library/src/main/scala/scala/scalajs/runtime/LinkingInfo.scala
```diff
@@ -10,6 +10,9 @@ trait LinkingInfo extends js.Object {
 
   /** Whether we are assuming ECMAScript 6 support or not. */
   val assumingES6: Boolean = js.native
+
+  /** Version of the linker */
+  val version: String = js.native
```
sjrd: Consider renaming to `linkerVersion`. `version` is a bit too general, even within `LinkingInfo`.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -57,4 +59,58 @@ class SystemJSTest {
...
+    assertEquals("1.8", get("java.vm.specification.version"))
+    assertEquals("Oracle Corporation", get("java.vm.specification.vendor"))
+    assertEquals("Scala.js", get("java.vm.name"))
+    assertEquals("1.8", get("java.specification.version"))
+    assertEquals("/", get("file.separator"))
+    assertEquals(":", get("path.separator"))
+    assertEquals("\n", get("line.separator"))
+    assertNotNull(get("java.vm.version"))
```
sjrd: You can make this more precise with
```scala
assertEquals(scala.scalajs.runtime.LinkingInfo.linkerVersion, get("java.vm.version"))
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/CoreJSLibs.scala
```diff
@@ -111,7 +111,8 @@ private[scalajs] object CoreJSLibs {
       else "" // blank line preserves line numbers in source maps
     }
 
-    val content = lines.mkString("", "\n", "\n")
+    val content = lines.mkString("", "\n", "\n").
+        replace("{{LINKER_VERSION}}", ScalaJSVersions.current)
```
sjrd: `replace(` should go at the end of the previous line.
Or the `.` should go at the beginning of the second line, with 2 spaces.

project/Build.scala
```diff
@@ -1065,83 +1065,105 @@ object Build extends sbt.Build {
...
-                    "source-map-support). To deactivate source map " +
-                    s"tests, do: set jsEnv in $projectId := " +
-                    "NodeJSEnv().value.withSourceMap(false)")
-                }
-                baseArgs :+ "-tsource-maps"
-              } else
-                baseArgs
-            }
+  val testTagSettings = {
+    class OptionTags(stage: Stage, sems: Semantics, jsEnv: JSEnv, projectId: String) {
```
sjrd: Looks like this could simply be a `def makeOptionTags(stage, ...): Seq[String]`. (and have callers call `map` instead of `as`)

project/Build.scala
```diff
@@ -1065,83 +1065,105 @@ object Build extends sbt.Build {
...
-          case FullOptStage => "-tfullopt-stage"
-        })
+        val stage = (scalaJSStage in Test).value
+        val sems = stage match {
+          case FastOptStage => (scalaJSSemantics in (Test, fastOptJS)).value
+          case FullOptStage => (scalaJSSemantics in (Test, fullOptJS)).value
+        }
+        val jsEnv = (resolvedJSEnv in Test).value
+        val optionsTags =
+          new OptionTags(stage, sems, jsEnv, thisProject.value.id)
```
sjrd: You can factor out this code (from `val stage =` to here) as a custom `Task`.

sjrd:
That's all.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +633,24 @@ object ScalaJSPluginInternal {
...
+    val javaSysPropsPattern = "-D([^=]*)=(.*)".r
+    val formattedProps = javaOptions.collect {
+      case javaSysPropsPattern(propName, propValue) =>
+        "\"" + escapeJS(propName) + "\": \"" + escapeJS(propValue) + "\""
+    }
```
gzm0: I still don't agree to this. Options we cannot handle should fail.

sjrd: I am actually not that sure. This method fills in the `System.getProperties()` map. It takes the `javaOptions` that are of interest to it. There *could* be another task/setting somewhere, that would take other parts of `javaOptions` and do other things with it. For example, one could imagine capturing `-Xmx` options and reroute them as `args` to `NodeJSEnv`.

Various `javaOptions` could have to be handled by various parts of our setting pipeline, and/or users' builds. The extraction of system properties should not break everything else.

gzm0: That is an implementation detail of the sbt plugin. I do not say it must fail here. I just say if you write:

```scala
javaOptions += "-foo"
```

running must fail.

Just like it does on the JVM:

```
> set javaOptions += "-foo"
// chatter
> set fork := true
// chatter
> run
// chatter
[error] Unrecognized option: -foo
[error] Error: Could not create the Java Virtual Machine.
[error] Error: A fatal exception has occurred. Program will exit.
```

sjrd: My point is: *where* should it fail? And how would that place make sure it understands all "handlers"?

gzm0: For now, it can fail here, since we know that the Scala.js sbt plugin does not have other handlers. Using `javaOptions` to change different parts of the pipeline is just plain wrong (and will fail in JVM projects since java will complain).

If we need to understand other options, we probably need to introduce an intermediary task that parses these into an "JvmOptions" structure. This task can then fail, downstream tasks use the structured result.

sjrd: OK that makes sense.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -223,6 +223,10 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalajsp = InputKey[Unit]("scalajsp",
         "Prints the content of a .sjsir file in human readable form.",
         CTask)
+
+    val scalaJSJavaSystemProperties = TaskKey[Seq[(String, String)]]("scalaJSJavaSysProps",
```
sjrd: The name as a string should be exactly the same as the identifier of the `val`.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +573,17 @@ object ScalaJSPluginInternal {
         }
       },
 
+      scalaJSJavaSystemProperties := scalaJSJavaSystemProperties.?.value.getOrElse {
+        val javaSysPropsPattern = "-D([^=]*)=(.*)".r
+        (javaOptions in Test).value.collect {
```
sjrd: `in Test` should be removed.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +573,17 @@ object ScalaJSPluginInternal {
...
+          case javaSysPropsPattern(propName, propValue) => (propName, propValue)
+        }
```
sjrd: This should fail on `case _ =>`, following the previous discussion.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -632,6 +640,25 @@ object ScalaJSPluginInternal {
     }
   }
 
+  protected def javaSysProps2JSLib(
+      javaOptions: Seq[(String, String)]): Option[ResolvedJSDependency] = {
```
sjrd: This should still be in its own task to modularize its constructions wrt `loadedJSEnv`. I suggest `scalaJSConfigurationLibs: TaskKey[Seq[ResolvedJSDependency]]`, initialized to `Nil` in the `projectBaseSettings` (or even in `Global`?), and in `configSettings` we would `++=` the content of this method.

That will allow users to inject their own configuration files very easily, if required.

library/src/main/scala/scala/scalajs/LinkingInfo.scala
```diff
@@ -108,4 +108,17 @@ object LinkingInfo {
...
+   *  constant-folding, inlining, and other local optimizations can be
+   *  leveraged with this "constant" to write polyfills that can be
+   *  dead-code-eliminated.
+   *
+   *  Used to define the `java.vm.version` property in `java.lang.System`.
+   *
+   */
+  @inline
+  def linkerVersion: String =
+    linkingInfo.linkerVersion
```
sjrd: My previous comment disappeared:

> Since it is accessible with `java.vm.version`, I wouldn't expose this here at all.

It should of course stay in `runtime.LinkerInfo`, but it shouldn't be part of the public API for `scalajs.LinkingInfo`.

project/Build.scala
```diff
@@ -1065,39 +1065,66 @@ object Build extends sbt.Build {
...
+        "Task that lists all test options for javaOptions and testOptions.",
+        KeyRanks.Invisible)
+
+    Seq(
+      testOptionTags := {
+        val stage = (scalaJSStage in Test).value
+        def stageTag: Seq[String] = stage match {
+          case FastOptStage => Seq("fastopt-stage")
+          case FullOptStage => Seq("fullopt-stage")
+          case _            => Seq.empty
```
sjrd: `case _ =>` is unnecessary. It cannot happen.

project/Build.scala
```diff
@@ -1065,39 +1065,66 @@ object Build extends sbt.Build {
...
+        }
+
+        val sems = stage match {
+          case FastOptStage => (scalaJSSemantics in (Test, fastOptJS)).value
+          case FullOptStage => (scalaJSSemantics in (Test, fullOptJS)).value
+        }
+
+        def semTags: Seq[String] = {
```
sjrd: Maybe move back `sems` and `semTags` down below, where they were before. So that the diff is more understandable.

project/Build.scala
```diff
@@ -1065,39 +1065,66 @@ object Build extends sbt.Build {
...
-      testOptions in Test ++= {
+  val testTagSettings = {
+    val testOptionTags = TaskKey[Seq[String]]("testOptionTags",
+        "Task that lists all test options for javaOptions and testOptions.",
+        KeyRanks.Invisible)
+
+    Seq(
+      testOptionTags := {
+        val stage = (scalaJSStage in Test).value
+        def stageTag: Seq[String] = stage match {
```
sjrd: And similarly move `stageTag` after `semTags`.

sjrd:
That's all.

nicolasstucki:
Dear Jenkins, retest this please.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +575,38 @@ object ScalaJSPluginInternal {
...
+          case (propName, propValue) =>
+            "\"" + escapeJS(propName) + "\": \"" + escapeJS(propValue) + "\""
+        }
+        val formattedPropsList =
+          if (formattedProps.isEmpty) Nil
+          else Seq(formattedProps)
+        formattedPropsList.map { props =>
+          val code = {
+            "var __ScalaJSEnv = (typeof __ScalaJSEnv === \"object\" && __ScalaJSEnv) ? __ScalaJSEnv : {};\n" +
+                "__ScalaJSEnv.javaSystemProperties = {" + props.mkString(", ") + "};"
```
sjrd: No indentation here.
Also add an `\n` at the end of the JS code.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +575,38 @@ object ScalaJSPluginInternal {
...
+        javaOptions.value.collect {
+          case javaSysPropsPattern(propName, propValue) => (propName, propValue)
+          case opt =>
+            sys.error("Scala.js javaOptions can only be \"-D<key>=<value>\"," +
+                " but received: " + opt)
+        }
+      },
+
+      scalaJSConfigurationLibs ++= {
+        val formattedProps = scalaJSJavaSystemProperties.value.map {
```
sjrd: Consider short-circuiting the empty case from the start:
```scala
val javaSystemProperties = scalaJSJavaSystemProperties.value
if (javaSystemProperties.isEmpty) {
  Nil
} else {
  val formattedProps = ...
  ...
  Some(ResolvedJSDependency ...)
}
```

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +575,38 @@ object ScalaJSPluginInternal {
         }
       },
 
+      scalaJSJavaSystemProperties ++= {
+        val javaSysPropsPattern = "-D([^=]*)=(.*)".r
+        javaOptions.value.collect {
```
sjrd: collect -> map

project/BinaryIncompatibilities.scala
```diff
@@ -26,6 +26,12 @@ object BinaryIncompatibilities {
   )
 
   val Library = Seq(
+    ProblemFilters.exclude[MissingMethodProblem](
+        "scala.scalajs.runtime.EnvironmentInfo.javaSystemProperties"),
+    ProblemFilters.exclude[MissingMethodProblem](
+        "scala.scalajs.runtime.LinkingInfo.linkerVersion"),
+    ProblemFilters.exclude[MissingMethodProblem](
+        "scala.scalajs.runtime.LinkingInfo.scala$scalajs$runtime$LinkingInfo$_setter_$linkerVersion_=")
```
sjrd: Those require a comment explaining why it's fine to exclude them. Here it would be:

    // In theory, breaking, but this is an interface in runtime that no one should extend

like we did in https://github.com/scala-js/scala-js/blob/v0.6.6/project/BinaryIncompatibilities.scala#L121

library/src/main/scala/scala/scalajs/LinkingInfo.scala
```diff
@@ -107,5 +107,4 @@ object LinkingInfo {
   @inline
   def assumingES6: Boolean =
     linkingInfo.assumingES6
-
```
sjrd: Spurious diff.

library/src/main/scala/scala/scalajs/runtime/LinkingInfo.scala
```diff
@@ -10,6 +10,9 @@ trait LinkingInfo extends js.Object {
 
   /** Whether we are assuming ECMAScript 6 support or not. */
   val assumingES6: Boolean = js.native
+
+  /** Version of the linker */
+  val linkerVersion: String = js.native
```
sjrd: Actually, I just realized this must be a `js.UndefOr[String]`. If we link a library 0.6.8 with a linker 0.6.7 (which is possible as long as 0.6.8 remains forward compatible), the `linkerVersion` will not be there yet. `loadSystemProperties()` should only populate `java.vm.version` if this field is not undefined.

sjrd:
That's all.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -573,9 +575,38 @@ object ScalaJSPluginInternal {
...
+          case javaSysPropsPattern(propName, propValue) => (propName, propValue)
+          case opt =>
+            sys.error("Scala.js javaOptions can only be \"-D<key>=<value>\"," +
+                " but received: " + opt)
+        }
+      },
+
+      scalaJSConfigurationLibs ++= {
+        val javaSystemProperties = scalaJSJavaSystemProperties.value
+        if(javaSystemProperties.isEmpty) {
```
sjrd: Missing space after `if`

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
```diff
@@ -24,4 +24,21 @@ object Platform {
...
+
+  def executingInRhino: Boolean = sysProp("rhino")
+  def executingInNodeJS: Boolean = sysProp("nodejs")
+  def executingInPhantomJS: Boolean = sysProp("phantomjs")
+  def typedArrays: Boolean = sysProp("typedarray")
+
+  def isInFastOpt: Boolean = sysProp("fastopt-stage")
+  def isInFullOpt: Boolean = sysProp("fullopt-stage")
+  def isInStrongMode: Boolean = sysProp("strong-mode")
+  def isInProductionMode: Boolean = sysProp("strong-mode")
```
sjrd: Oops. `isInStrongMode` should not be there + `isInProductionMode` is testing `"strong-mode"`.

test-suite/jvm/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
```diff
@@ -15,4 +15,18 @@ object Platform {
...
   }
+
+  def executingInRhino: Boolean = false
+  def executingInNodeJS: Boolean = false
+  def executingInPhantomJS: Boolean = false
+  def hasTypedArrays: Boolean = false
+
+  def isInFastOpt: Boolean = false
+  def isInFullOpt: Boolean = false
+  def isInStrongMode: Boolean = false
```
sjrd: `isInStrongMode` should not be there.

nicolasstucki:
Dear Jenkins, retest this please.

sjrd:
LGTM :)

Issue #1881, opened 2015-08-27
Stop recommending PhantomJS


japgolly:
Scala.JS recommends Node.JS for DOM-less testing, and PhantomJS for testing with DOM.

Over the past year, PhantomJS has been nothing but a source of trouble for me. There are 1600+ open issues on its project page, many are serious and have been open without activity for years. In fact it doesn't get much activity at all. There are a few open issues within Scala.JS too (#1555, #1457, #1702, more?).

We shouldn't be directing Scala.JS users to use PhantomJS. Some other solution should be discovered and recommended.

I'm guessing node/io.js + jsdom (_@sjrd | @gzm0, if you've tried this I'd love to hear your experiences_), or maybe something using `xvfb` as suggested [here](https://groups.google.com/forum/#!folder/Scala/scala-js/d7S56w2VUpw). No answer yet, but I think it should be recognised that PhantomJS doesn't really work, and we're going to be setting the community up for eventual and significant pain, if we find a better solution here.

mdedetrich:
tbh, I don't think any solution would be better than `PhantomJS`. Emulating the DOM **correctly** is hard, every browser does it differently, and with stuff continuously moving in HTML5 land, its hard to be ontop of especially with a spec that isn't really set in stone.

I don't have an issue in recommending something that is better, I am just not sure if a better exists

japgolly:
It's not just niche/advanced/edge features that PhantomJS can't handle. It doesn't even support focusing (`.focus()`, `.blur()`, `document.activeElement`). :disappointed:

mdedetrich:
I don't doubt that at all, my main point is that I am not aware of a better `Phantom.js` existing. Doing this stuff correctly is ridiculously hard, and you can use the old "standards" argument as to why thats the case.

Also issue count isn't an objective measure of how stable/correct implementation is. Once any non trivial project gets to a certain size/number of users, you get that many issues (a significant amount of those issues also turn out to be non issues)

japgolly:
[...]

PR #2245, opened 2016-02-17
DO NOT MERGE Fix #2168: Add SeleniumJSEnv supporting Firefox and Chrome.

### LINKED ISSUE #2168: Feature request: Selenium support
As I've rambled, ranted and cried about previously, certain things are impossible to simulate/test with {Node,Phantom}.JS. It would be really awesome if Scala.JS came with similar support for Selenium so I could run tests in a real browser.

nicolasstucki:


nicolasstucki:
Do not review yet.

nicolasstucki:
Review by @sjrd.

Note that this will need a rebase on #2237 and needs Firefox on Jenkins.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -9,6 +9,7 @@
 
 package org.scalajs.sbtplugin
 
+import org.scalajs.jsenv.selenium.{SeleniumBrowser, SeleniumJSEnv}
```
nicolasstucki: This import must go with the other ones.

js-envs/src/main/scala/org/scalajs/jsenv/selenium/SeleniumComJSRunner.scala
```diff
@@ -0,0 +1,102 @@
...
+         |    popOutMsg: function() {
+         |      if (sendMsgBufOut.length == 0) {
+         |         sendMsgBufOut = sendMsgBufIn.reverse();
+         |         sendMsgBufIn = [];
+         |      }
+         |      return sendMsgBufOut.pop();
+         |    },
+         |    recvMessage: function(msg) {
+         |      onReceive(msg);
+         |    }
```
gzm0: Consider putting these two on a different, more obscurely named object so it is clear they are not for public consumption.

nicolasstucki: I'm considering changing it to:
```javascript
this.scalajsSeleniumComJSRunnerChanel = {
  popOutMsg: function() { ... },
  recvMessage: function(msg) { ... }
}
```

sjrd: Even better, make those functions in the scope of the enclosing lambda (where `var onReceive` is). That way they'll be truly private.

ci/matrix.xml
```diff
@@ -158,6 +173,13 @@
         'set scalaJSStage in Global := FullOptStage' \
         'set scalaJSOptimizerOptions in testSuite ~= (_.withDisableOptimizer(true))' \
         ++$scala testSuite/test \
+        testSuite/clean &&
+    sbtretry 'set inScope(ThisScope in testSuite)(jsEnv := SeleniumJSEnv(SeleniumFirefox).value)' \
+        ++$scala testSuite/test \
+        testSuite/clean &&
+    sbtretry 'set inScope(ThisScope in testSuite)(jsEnv := SeleniumJSEnv(SeleniumFirefox).value)' \
+        'set scalaJSStage in Global := FullOptStage' \
+        ++$scala testSuite/test \
```
sjrd: These changes should probably not be there. You're in the ES6 section, but they don't test ES6. These two can be removed, I think, they're duplicates of the ones in the ES5 section anyway.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -110,6 +112,49 @@ object ScalaJSPlugin extends AutoPlugin {
...
+    def SeleniumFirefox(): SeleniumBrowser =
+      selenium.Firefox
+
+    /** Google Chrome browser configuration for SeleniumJSEnv(browser).
+     *
+     *  @param chromedriverPath Path to the Selenium chrome driver.
+     *    Can be downloaded from http://chromedriver.storage.googleapis.com/index.html
+     */
+    def SeleniumChrome(chromedriverPath: String): SeleniumBrowser = {
+      System.setProperty("webdriver.chrome.driver", chromedriverPath)
```
sjrd: Ouch, that's a nasty side-effect right there. Any way to make this a little more contained?

js-envs/src/main/scala/org/scalajs/jsenv/selenium/Firefox.scala
```diff
@@ -0,0 +1,51 @@
...
+    new MemVirtualJSFile("hijackConsole.js").withContent(
+      """
+        |(function () {
+        |  var console_log_hijack = [];
+        |  var oldLog = console.log;
+        |  console.log = function (msg) {
+        |    console_log_hijack.push(msg);
+        |    oldLog.apply(console, arguments);
+        |  };
+        |  this.sclajsPopHijackedConsoleLog = function () {
```
sjrd: Typo: `sclajs`

js-envs/src/main/scala/org/scalajs/jsenv/selenium/Firefox.scala
```diff
@@ -0,0 +1,51 @@
...
+      """.stripMargin
+    )
+  }
+
+  private class FirefoxDriver extends BrowserDriver {
+    protected def newDriver(): RemoteWebDriver =
+      new org.openqa.selenium.firefox.FirefoxDriver()
+
+    def newConsoleLogsIterator(): Iterator[String] = {
+      getWebDriver.executeScript("return sclajsPopHijackedConsoleLog()") match {
```
sjrd: Same typo, obviously.

js-envs/src/main/scala/org/scalajs/jsenv/selenium/Firefox.scala
```diff
@@ -0,0 +1,51 @@
...
+        case log: util.ArrayList[_] => log.iterator().map(_.toString)
+        case _ => throw new Exception("Ill formatted message.")
```
sjrd: Align the arrows, or (if too long) put the body of the cases on the next line.

sjrd:
Given the hard conditions in which this needs to be tested, I'm wondering whether we could test these on Travis. It appears to have support for Selenium: https://docs.travis-ci.com/user/gui-and-headless-browsers/ (including local Firefox + remote "Sauce Labs" cluster of things).

GitHub is able to handle CI results from various sources, so it should be doable.

(Or, even crazier, we might want to consider putting Selenium JS envs in a dedicated repo+artifact ?)

Mostly thinking out loud, here.

js-envs/src/main/scala/org/scalajs/jsenv/selenium/SeleniumComJSRunner.scala
nicolasstucki: They need to be visible from a global perspective, because it is another script that invokes them.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
nicolasstucki: I could just remove the parameter and let the user add it to the `javaOptions` as a `-D`.

gzm0:
I was going to ask: why does this need to be in the core? (Unless we want to use it ourselves in here).

js-envs/src/main/scala/org/scalajs/jsenv/selenium/SeleniumComJSRunner.scala
lihaoyi: Probably should be `RunnerChannel` (with two `n`s) too unless it's a new perfume line :stuck_out_tongue:

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: @sjrd This kind of stuff is where my utter dislike for system properties comes from :)

js-envs/src/main/scala/org/scalajs/jsenv/selenium/SeleniumComJSRunner.scala
nicolasstucki: Clearly, thanks

nicolasstucki:
I reactivated the Travis sevice to test Selenium on Firefox up there. I removed those tests from the Jenkins CI for now.

nicolasstucki:
Hmm, Travis does not support oraclejdk6 (only openjdk6). Should it be tested on openjdk6? Or do we just ignore it for this particular case.

nicolasstucki:
It seems like the `testSuite/test` overloads Travis by creating too many instances of Firefox and the start dying. But the `testingExample/test` seems to work fine, this is the one that we really need because it is the one that tests the DOM. We could still do a manual test on the `testSuite` before releases (we also need this for Chrome).

gzm0:
More seriously: Why does this need to go into Scala.js core? There seems nothing inherent at all here.

Having this separate does not require substantial user overhead either.

nicolasstucki:
The only reason why we would keep it in the core would be to address #1881 with this environment in the future. If not, I agree that it would be better in its own repo.

gzm0:
Didn't you try out alternatives in Node.js that support the DOM? Also, quite honestly, I don't think it is that of a big deal if we don't support (efficient) DOM testing in core. Currently, doing so requires manual work anyways (installing PhantomJS).
So if at a later point in time, we just load Selenium reflectively and tell you you should put it on the classpath if it isn't there, it is not really any worse. But I really think this would make our life much easier (also we don't force people to pull in even more stuff, if they don't care about selenium).

sjrd:
Eventually, I think we should even move PhantomJS support outside of core. Rhino and Node.js are the only things *we* care about.

nicolasstucki:
@gzm0 I did try an alternative using `jsdom` on Node.js but the DOM support was incomplete and hard to manage in the environment.

It is evident that this will end up in a separate repo. I would propose to call it `scala-js-env-selenium` so that later on, if we extract Phantom from the core, we can call it `scala-js-env-phantom` and have a common prefix.

gzm0:
> scala-js-env-selenium
SGTM

PR #2260, opened 2016-02-26
Fix #2259: Reachable default methods cause isNeededAtAll == true.

### LINKED ISSUE #2259: isNeededAtAll does not take reachable default methods into account
This came up while porting Dotty to Scala.js. The following program links, but crashes the optimizer because `MyTrait` is not identified as `isNeededAtAll`:
```scala
package hello

import scala.scalajs.js

trait MyTrait {
  val x = 5
  def foo(y: Int) = x //+ y
}

object world extends js.JSApp with MyTrait {
  def main(): Unit = {
    println("hello dotty.js!")
    println(foo(4))
  }
}
```
So despite `foo` being reachable, `MyTrait` is completely eliminated. Adding a stupid
```scala
println(classOf[MyTrait])
```
works around the issue, because now `MyTrait` `isNeededAtAll`, and the default methods are correctly kept.

I am not sure yet whether it's possible to reproduce the bug in core Scala.js (for scalac).

sjrd:


sjrd:
Review by @gzm0

gzm0:
LGTM

Issue #2269, opened 2016-02-27
Remove primitives


sjrd:
This is a meta-issue, collecting all the issues trying to get rid of compiler primitives:
- [x] #1265 - `js.Dynamic.literal`
- [x] #2261 - `j.l.Object.getClass`
- [x] #2262 - `js.Any.fromFunctionN` and `js.ThisFunction.fromFunctionN`
- [x] #2263 - `js.Object.propertiesOf`
- [x] #2264 - `js.Object.hasProperty`
- [x] ~~#2266 - `js.Array.apply`~~
- [x] #2267 - `runtime.environmentInfo`
- [x] #2268 - `BoxedUnit.TYPE` ~~(maybe `BoxedUnit.UNIT`?)~~

What survives the purge?
- `js.Dynamic.newInstance`
- `js.Dictionary.delete`
- `js.typeOf`
- `js.debugger`
- (`js.native`, not really a primitive, it's for compile errors only)
- `BoxedUnit.UNIT` ~~?~~
- `runtime.constructorOf`
- `runtime.linkingInfo`
- `js.Array.apply`

sjrd:
Which makes me think `js.Dictionary.delete` should probably be `js.delete(js.Dynamic, String)`...

mrmiywj:
Maybe I can pick up some of these? Or any ideas about something easier for new bees?

sjrd:
The remaining issues here have non-trivial consequences on a) the design of the language and b) backward compatibility issues. So I don't think they're a good fit to get into the project.

sjrd:
All done.

PR #2273, opened 2016-03-01
Internal-only removal of compiler primitives

sjrd:
This is a subset of the removals of primitives of #2269 that are not externally visible.


sjrd:
One commit missing; waiting for CI to start all jobs before pushing it. The full list is visible at https://github.com/scala-js/scala-js/compare/master...sjrd:less-primitives-complete

Review by @gzm0

gzm0:
Did you verify that we have enough tests to cover `43f74b5`?
- Super calls to `getClass`
- Calls to `getClass` in non-hijackable context
- Calls to `getClass` in hijackable context on non-hijacked value
- Calls to `getClass` in hijackable context on hijacked value

library/src/main/scala/scala/scalajs/js/Any.scala
```diff
@@ -91,29 +91,30 @@ object Any extends LowPrioAnyImplicits {
...
-  implicit def fromFunction21[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, R](f: scala.Function21[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, R]): Function21[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, R] = sys.error("stub")
-  implicit def fromFunction22[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R](f: scala.Function22[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R]): Function22[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R] = sys.error("stub")
+  // identity() is important! It prevents the tail-rec treatment
```
gzm0: Could you quickly explain this? (More out of interest :P)

gzm0:
Do you think there is a way we can test the `js.{Any,ThisFunction}.fromFunctionN` change? I can see how testing this is hard, but it also seems a bit crazy to rely on Scalac doing the right thing under an experimental flag for our core library. How hard would it be to generate IR manually to call these guys?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
```diff
@@ -14,6 +14,42 @@ import scala.runtime.BoxedUnit
...
+    assertNotSame(classOf[java.lang.Void], scala.runtime.BoxedUnit.TYPE)
+    assertNotSame(classOf[java.lang.Void], java.lang.Void.TYPE)
+    assertNotSame(classOf[java.lang.Boolean], java.lang.Boolean.TYPE)
+    assertNotSame(classOf[java.lang.Character], java.lang.Character.TYPE)
+    assertNotSame(classOf[java.lang.Byte], java.lang.Byte.TYPE)
+    assertNotSame(classOf[java.lang.Short], java.lang.Short.TYPE)
+    assertNotSame(classOf[java.lang.Integer], java.lang.Integer.TYPE)
+    assertNotSame(classOf[java.lang.Long], java.lang.Long.TYPE)
+    assertNotSame(classOf[java.lang.Float], java.lang.Float.TYPE)
+    assertNotSame(classOf[java.lang.Double], java.lang.Double.TYPE)
```
gzm0: This is because we don't box and differs from Scala JVM, right?

gzm0:
That's all.

library/src/main/scala/scala/scalajs/js/Any.scala
sjrd: Without inline:
```scala
implicit def fromFunction1[T1, R](f: T1 => R): js.Function1[T1, R] = (x1: T1) => f(x1)
```
expands to
```scala
def fromFunction1[T1, R](f: T1 => R): js.Function1[T1, R] = fromFunction1((x1: T1) => f(x1))
```
where the recursive call to `fromFunction1` is in tail position. So after the `tailcalls` phase, this expands to (roughly):
```scala
def fromFunction1[T1, R](f: T1 => R): js.Function1[T1, R] = {
  loop: while (true) {
    val newF = (x1: T1) => f1
    f = newF // re-assign the parameter
    continue loop
  }
}
```
and hence the back-end sees only an infinite loop, which it compiles as such.

With `identity()`, after implicit expansion we get:
```scala
def fromFunction1[T1, R](f: T1 => R): js.Function1[T1, R] =
  identity(fromFunction1((x1: T1) => f(x1)))
```
where the recursive call to `fromFunction1` is *not* in tail-call position. Therefore, `tailcalls` does not transform the function. The back-end then sees a *call* to `fromFunction1`, *which is a primitive*, and then applies its primitive treatment to it, short-circuiting the infinite recursion.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
sjrd: This is a cross-compiled test. Everything here behaves exactly the same on JS and JVM.

library/src/main/scala/scala/scalajs/js/Any.scala
gzm0: Oh, I see. In this case maybe this deserves a bigger comment explaining how the primitive chaining works. I expected this to only produce a sensible result under `-Xexperimental`.

Also, does this strange artifact happen as well with SAM treatment or just with vanilla 2.1x?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
gzm0: Ahh... Yes. Makes sense. (Should read the docs :P)

sjrd:
Did you verify that we have enough tests to cover the j.l.Object.getClass change?

> Super calls to getClass

https://github.com/scala-js/scala-js/blob/c03909dd2f4739d27a8395545ad3576e3e6d0153/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/ReflectionTest.scala#L92

> Calls to getClass in non-hijackable context

https://github.com/scala-js/scala-js/blob/c03909dd2f4739d27a8395545ad3576e3e6d0153/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/ReflectionTest.scala#L95

> Calls to getClass in hijackable context on non-hijacked value

It seems we don't have an explicit test for this. I'll add one. That said, there is `java.lang.{Object,Throwable}.toString()` which definitely use it.

> Calls to getClass in hijackable context on hijacked value

https://github.com/scala-js/scala-js/blob/c03909dd2f4739d27a8395545ad3576e3e6d0153/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/ReflectionTest.scala#L99

And calls to getClass in hijackable context with a JS value:

https://github.com/scala-js/scala-js/blob/c0a00912df3f4b487cedd0a0d07251af8f6d9dd0/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala#L225

sjrd:
> Do you think there is a way we can test the `js.{Any,ThisFunction}.fromFunctionN` change? I can see how testing this is hard, but it also seems a bit crazy to rely on Scalac doing the right thing under an experimental flag for our core library. How hard would it be to generate IR manually to call these guys?

It's not under the experimental flag. We don't compile our library with `-Xexperimental`. In this case it's the current behavior of triggering the implicit conversion. With the flag, it can only "get better", because the implicit conversion will not be triggered (and we do have tests that lambda expressions in SAM position work fine with `-Xexperimental` because we compile the whole test suite with the flag).

We don't have any infrastructure to link/run custom IR. It's always the same problem, I guess. We should invest in that, but it's out of scope for this PR. Whether we want to delay this PR after we manage to set up such an infrastructure is another question, of course ...

library/src/main/scala/scala/scalajs/js/Any.scala
sjrd: The artifact does not happen with the SAM treatment, because the (recursive) implicit conversion is not necessary.

gzm0:
> We don't have any infrastructure to link/run custom IR. It's always the same problem, I guess. We should invest in that, but it's out of scope for this PR. Whether we want to delay this PR after we manage to set up such an infrastructure is another question, of course ...

I think it is OK to let this through, since it is so simple. However, I anticipate things like this (i.e. stuff we cannot test otherwise) to occur more, now that we'll have two backends to generate IR. Considering that the necessary infrastructure is pretty easy to write (unit test that loads library, feeds it to the linker, runs, asserts), I think we should envision to write something like that in the near term.

sjrd:
Updated. Comments addressed + updated checksizes in the `getClass` commit. Inlining the method sometimes requires to extract the argument in a temporary variable. This gave about 90 bytes to the fastopt version.

gzm0:
LGTM

PR #2281, opened 2016-03-09
Fix #2243: fix JSDependencies cache path.

### LINKED ISSUE #2243: -jsdeps.min.js not produced anymore after upgrading from 0.6.5 to 0.6.6 and 0.6.7
In version 0.6.5, running `sbt fastOptJS fullOptJS` used to produce both -jsdeps.js and -jsdeps.min.js.

After upgrading to 0.6.6 and 0.6.7, the -jsdeps.min.js stopped being generated.

After further investigating, I could see that

* running `sbt fastOptJS` (without `fullOptJS`) would produce just the -jsdeps.js -- as expected.
* running `sbt fullOptJS` (without `fastOptJS`) would produce just the -jsdeps.min.js, but not -jsdeps.js.
* running `sbt fastOptJS fullOptJS` would produce just the -jsdeps.js, but not -jsdeps.min.js.

Please check this repository for reproducing the bug: https://github.com/gabriel-bezerra/scalajs-067-minified-jsdeps-bug

nicolasstucki:


nicolasstucki:
Review by @sjrd.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -431,8 +431,8 @@ object ScalaJSPluginInternal {
       packageJSDependenciesSetting(packageJSDependencies, "package-js-deps", _.lib),
 
       artifactPath in packageMinifiedJSDependencies :=
-        ((crossTarget in packageJSDependencies).value /
-            ((moduleName in packageJSDependencies).value + "-jsdeps.min.js")),
+        ((crossTarget in packageMinifiedJSDependencies).value /
+            ((moduleName in packageMinifiedJSDependencies).value + "-jsdeps.min.js")),
```
sjrd: Looks like this fixes another (unreported) issue. There should be an issue to reference and fix it in a different commit.

sjrd:
Could we have a test for this? Something in `sbt-plugin-test` like:
```scala
testRegressionIssue2243 := {
  val _ = Def.sequential(packageJSDependencies in Compile, packageMinifiedJSDependencies in Compile).value
  assert((artifactPath in (Compile, packageJSDependencies))).value.exists)
  assert((artifactPath in (Compile, packageMinifiedJSDependencies))).value.exists)
}
```

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
nicolasstucki: Extracted as fix for #2282.

sbt-plugin-test/build.sbt
```diff
@@ -182,10 +182,30 @@ lazy val jsDependenciesTest = project.settings(versionSettings: _*).
...
+lazy val jsNoDependenciesTest = project.settings(versionSettings: _*).
+    enablePlugins(ScalaJSPlugin).
+    settings(inConfig(Compile)(Seq(
+      packageJSDependencies <<= packageJSDependencies.dependsOn(Def.task {
+        // Regression test for issue #2243
+        val pack = (packageJSDependencies in Compile).value
+        val minPack = (packageMinifiedJSDependencies in Compile).value
+        assert((artifactPath in (Compile, packageJSDependencies)).value.exists)
+        assert((artifactPath in (Compile, packageMinifiedJSDependencies)).value.exists)
+        streams.value.log.info("Regression test for issue #2243 passed")
```
nicolasstucki: Added two sets of regression tests, one for the case where there are dependencies and one for the case where there are none.

sbt-plugin-test/build.sbt
```diff
@@ -182,10 +182,30 @@ lazy val jsDependenciesTest = project.settings(versionSettings: _*).
...
+      val minPack = (packageMinifiedJSDependencies in Compile).value
+      assert((artifactPath in (Compile, packageJSDependencies)).value.exists)
+      assert((artifactPath in (Compile, packageMinifiedJSDependencies)).value.exists)
+      streams.value.log.info("Regression test for issue #2243 passed")
     })
   )): _*).
   dependsOn(jetty9) // depends on jQuery
 
+lazy val jsNoDependenciesTest = project.settings(versionSettings: _*).
+    enablePlugins(ScalaJSPlugin).
```
sjrd: 2 spaces only (and all lines below)

sbt-plugin-test/build.sbt
```diff
@@ -182,10 +182,30 @@ lazy val jsDependenciesTest = project.settings(versionSettings: _*).
           "compressed/history.js appears before foo.js")
 
       streams.value.log.info("jsDependencies resolution test passed")
+
+      // Regression test for issue #2243
+      val pack = (packageJSDependencies in Compile).value
```
sjrd: Hem ... this code is in a task that `packageJSDependencies in Compile` depends on. You have a very weird circular dependency, here. I have no idea why sbt let you do that, but I don't think it's good to rely on it.

I suggest putting this code in a separate (custom) task.

sbt-plugin-test/build.sbt
```diff
@@ -182,10 +182,30 @@ lazy val jsDependenciesTest = project.settings(versionSettings: _*).
...
+      val minPack = (packageMinifiedJSDependencies in Compile).value
```
sjrd: Also I would encourage using `Def.sequential` to encourage sbt not to run both this things in parallel, since otherwise the test could accidentally work due to a race.

sbt-plugin-test/build.sbt
sjrd: The two are literally copy-pasted. This should be factored out.

sjrd:
That's all.

ci/matrix.xml
```diff
@@ -224,7 +224,9 @@
         'set scalaJSUseRhino in Global := false' \
         jetty9/run test \
         jsDependenciesTest/packageJSDependencies \
-        jsDependenciesTest/packageMinifiedJSDependencies
+        jsDependenciesTest/packageMinifiedJSDependencies \
+        jsDependenciesTest/regretionTestForIssue2243 \
+        jsNoDependenciesTest/regretionTestForIssue2243
```
sjrd: Er ... typo :-) it's "regression" with ss, not t

sjrd:
That's all.

sjrd:
LGTM

PR #2283, opened 2016-03-09
Sbt-plugin sharedSrcDir use Seq instead of Option.

MikaelMayer:
We would like to use cross-projects with possibly more source files than just one.

### DIFF
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
@@ -315,11 +315,9 @@ object CrossProject extends CrossProjectExtra {
   )
 
   // Inspired by sbt's Defaults.makeCrossSources
-  private def makeCrossSources(sharedSrcDir: Option[File],
+  private def makeCrossSources(sharedSrcDir: Seq[File],
       scalaBinaryVersion: String, cross: Boolean): Seq[File] = {
-    sharedSrcDir.fold[Seq[File]] {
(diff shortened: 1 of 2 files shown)


sjrd:
Is it really a property of a semi-general `CrossType` that you want?
Why doesn't simply `.settings(unmanagedSourceDirectories in Compile += something)` cover your use case?

sjrd:
Jenkins, test this please

(I expect that this will break binary compatibility)

sjrd:
test this please

gzm0:
I agree with @sjrd that using `unmanagedSourceDirectories` is the way to go... You would have to do the same for a normal sbt project...

MikaelMayer:
unmanagedSourceDirectories is the way I tried for the project but none of
the sources I added was available from the javascript project nor the jvm
project.

I have the sbt command in my build file :

lazy val shared = crossproject ... ( ... file("shared")).settings(
...
unmanagedSourceDirectories in Compile += (baseDirectory.value /
"../../leon/library" ))

The folder structure is:

/shared/src/main/scala/...
/leon/library/...

It did not complain or anything, it just ignored the secondary source
folder.

This is why I wanted to define another type of crossType, but source
folders have to be options.
Any advice appreciated.
Le 9 mars 2016 17:52, "Tobias Schlatter" <notifications@github.com> a
écrit :

sjrd:
@MikaelMayer Check that you have the right path with

    > show sharedJS/compile:unmanagedSourceDirectories

in sbt. You can also

    > show sharedJS/compile:sources

to check the exact set of .scala files that are used.

PR #2289, opened 2016-03-11
Fix #2014: Add project to test jUnit tests from scala/scala.

### LINKED ISSUE #2014: Run the JUnit tests from scala/scala
Now that we have JUnit support, we should run [the JUnit tests from scala/scala](https://github.com/scala/scala/tree/2.11.x/test/junit/scala), compiled with Scala.js. This is similar to the fact that we run partest tests compiled with Scala.js.

nicolasstucki:


nicolasstucki:
Review by @sjrd

ci/matrix.xml
```diff
@@ -80,53 +80,53 @@
         ++$scala noIrCheckTest/test \
         noIrCheckTest/clean &&
     sbtretry 'set scalaJSOptimizerOptions in testSuite ~= (_.withDisableOptimizer(true))' \
-        ++$scala testSuite/test &&
+        ++$scala testSuite/test scalaTestSuite/test &&
```
sjrd: You also need to apply the `scalaJSOptimizerOptions` to `scalaTestSuite`, otherwise this is the same as above. The same comment applies to almost all lines from here.

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+    base = file("scala-test-suite"),
+    settings = commonSettings ++ myScalaJSSettings ++ Seq(
+      publishArtifact in Compile := false,
+
+      testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a"),
+
+      unmanagedSources in Test ++= {
+        def loadList(listName: String): Set[String] = {
+          val listsDir = (resourceDirectory in Test).value / scalaVersion.value
+          val buff = scala.io.Source.fromFile(listsDir / listName)
```
sjrd: What if the file does not exist? Like for 2.10 versions?

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+          if (file.isDirectory) {
+            file.listFiles().foldLeft((acc, listed, listedErr)) { (foldAcc, f) =>
+              listFiles(f, foldAcc._1, foldAcc._2, foldAcc._3)
+            }
+          } else {
+            file.relativeTo(jUnitTestsPath) match {
+              case Some(relPath) if relPath.getName.endsWith(".scala") =>
+                val relativePathString = relPath.toString
+                val blacklisted = blackList(relativePathString)
+                if (!(blacklisted ^ whiteList(relativePathString)))
```
sjrd: `if (blacklisted == whiteList(relativePathString))`?

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+            case line if !line.startsWith("#") && line.nonEmpty => line
+          }
+          lines.toSet
+        }
+
+        val blackList: Set[String] = loadList("BlacklistedTests.txt")
+        val whiteList: Set[String] = loadList("WhitelistedTests.txt")
+
+        val jUnitTestsPath =
+          (artifactPath in fetchScalaSource in partest).value / "test" / "junit"
```
sjrd: Use `(fetchScalaSource in partest).value` instead. And then you get the dependency for free so the `val _ =` later is not needed.

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+
+        def listFiles(file: File, acc: List[File], listed: List[String],
+            listedErr: List[String]): (List[File], List[String], List[String]) = {
+          if (file.isDirectory) {
+            file.listFiles().foldLeft((acc, listed, listedErr)) { (foldAcc, f) =>
+              listFiles(f, foldAcc._1, foldAcc._2, foldAcc._3)
+            }
```
sjrd: Use `(topLevelDir ** "*.scala").get` to get all files recursively from `topLevelDir`. Don't hand-roll your own directory recursion.

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+                  (file :: acc, relativePathString :: listed, listedErr)
+
+              case _ =>
+                (acc, listed, listedErr)
+            }
+          }
+        }
+
+        val _ = (fetchScalaSource in partest).value
+        val (whitelistedFiles, listed, listedErr) =
```
sjrd: `listed` -> `blacklisted`?

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+          listFiles(jUnitTestsPath, Nil, Nil, Nil)
+        val wronglyListed = blackList.union(whiteList).diff(listed.toSet)
```
sjrd: How can this be non-empty, given that `listFiles` already puts erroneous entries in `listedErr`.

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+
+        if (listedErr.nonEmpty || wronglyListed.nonEmpty) {
+          val msg = new StringBuffer("Errors in black or white lists.\n")
+          val listedErrSet = listedErr.toSet
+          val (inBoth, inNone) = listedErrSet.span(blackList.apply)
```
sjrd: Did you mean `partition`?

project/Build.scala
```diff
@@ -1419,4 +1424,85 @@ object Build extends sbt.Build {
...
+          val listsDir = (resourceDirectory in Test).value / scalaVersion.value
+          val buff = scala.io.Source.fromFile(listsDir / listName)
+          val lines = buff.getLines().collect {
+            case line if !line.startsWith("#") && line.nonEmpty => line
+          }
+          lines.toSet
+        }
+
+        val blackList: Set[String] = loadList("BlacklistedTests.txt")
+        val whiteList: Set[String] = loadList("WhitelistedTests.txt")
```
sjrd: Lowercase `l` in `blacklist` and `whitelist`

sjrd:
That's all.

project/Build.scala
nicolasstucki: It fails with an exception that says that the file does not exist. Considering that it also fails if the file contains incoherent contents, this seams like the equivalent behaviour.

project/Build.scala
nicolasstucki: Hmm, yes.

project/Build.scala
nicolasstucki: No, listed is the list of all Scala sources found in the directory. It will be more explicit once I use `(topLevelDir ** "*.scala").get`.

project/Build.scala
nicolasstucki: Yes

project/Build.scala
nicolasstucki: `listedErr` only contains sources refs that have a corresponding file. `wronglyListed` gives the list that does not have corresponding files in the directory.

ci/matrix.xml
```diff
@@ -79,86 +79,86 @@
...
+    sbtretry 'set scalacOptions in $testSuite += "-Xexperimental"' \
         'set scalaJSUseRhino in Global := false' \
-        ++$scala testSuite/test &&
-    sbtretry 'set scalacOptions in testSuite += "-Xexperimental"' \
+        ++$scala $testSuite/test &&
+    sbtretry 'set scalacOptions in $testSuite += "-Xexperimental"' \
         'set scalaJSUseRhino in Global := false' \
         'set scalaJSStage in Global := FullOptStage' \
-        ++$scala testSuite/test
+        ++$scala $testSuite/test scalaTestSuite/test
```
nicolasstucki: I missed one `scalaTestSuite/test` here.

nicolasstucki:
Note that I had to set `testSuite` in the task as a variable `$testSuite` the previous way would have failed nightly tests on 2.10.

sjrd:
> Note that I had to set `testSuite` in the task as a variable `$testSuite` the previous way would have failed nightly tests on 2.10.

Wasn't it easier to simply have `sources in scalaTestSuite` be a `Seq.empty` in the case of 2.10?

nicolasstucki:
That could be simpler. I also considered the fact that I would have to duplicate each `set xyz in testSuite` to have also the `set xyz in scalaTestSuite` in each statement of the testing tasks. This approach seems harder to maintain at long term.

nicolasstucki:
This PR conflicts with #2294, the second one being merged will need a rebase on the first one.

nicolasstucki:
Rebases and adapted ci matrix to Scala 2.11.8.

scala-test-suite/src/test/resources/2.11.8/WhitelistedTests.txt
```diff
@@ -0,0 +1,27 @@
...
+scala/math/OrderingTest.scala
+scala/runtime/ScalaRunTimeTest.scala
+scala/tools/nsc/SampleTest.scala
+scala/tools/testing/AssertUtil.scala
+scala/tools/testing/TempDir.scala
+scala/util/RandomTest.scala
+scala/util/SpecVersionTest.scala
+scala/util/TryTest.scala
+
+
```
sjrd: Spurious blank lines.

project/Build.scala
```diff
@@ -1423,4 +1428,84 @@ object Build extends sbt.Build {
...
+            msg.append(nonexistentWhitelisted).append('\n')
+          }
+          throw new AssertionError(msg.toString)
+        }
+
+        scalaScalaJUnitSources.collect {
+          case fTup if whitelist(fTup._1) => fTup._2
+        }
+      }
+
```
sjrd: Spurious blank line.

ci/matrix.xml
```diff
@@ -19,7 +19,7 @@
     sbtretry 'set scalaJSUseRhino in Global := false' \
         ++$scala helloworld/run &&
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
-	'set scalaJSUseRhino in Global := false' \
+	    'set scalaJSUseRhino in Global := false' \
```
sjrd: Indentation issue.

ci/matrix.xml
```diff
@@ -65,13 +65,13 @@
...
-    sbtretry ++$scala testSuite/test &&
+    sbtretry ++$scala $testSuite/test &&
     sbtretry 'set scalaJSUseRhino in Global := false' \
-        ++$scala testSuite/test &&
+        ++$scala $testSuite/test &&
     sbtretry 'set scalaJSUseRhino in Global := false' \
-	'set scalaJSStage in Global := FullOptStage' \
-        ++$scala testSuite/test \
-        testSuite/clean &&
+	    'set scalaJSStage in Global := FullOptStage' \
```
sjrd: Indentation issue.

sjrd:
That's all for the code.

I'm a bit worried about the explosion of testing jobs, though :-s
Maybe it's not worth testing JUnit tests for all minor Scala versions. I suggest we only do it for the latest 2.11 and the latest 2.12, i.e., 2.11.8 and 2.12.0-M3 at the moment.

ci/matrix.xml
nicolasstucki: There was a thab there hidden at the beginning of that line, I'll clean up all the tabs in this file.

sjrd:
LGTM

PR #2291, opened 2016-03-13
Cleanup state management in GenJSCode

gzm0:
This is pre-work to be able to generate classes in a nested fashion (e.g. generate an anonymous class while we are generating another one).


gzm0:
Review by @sjrd

gzm0:
Obsoleted by #2292

PR #2292, opened 2016-03-13
Fix #2009: No own class for anonymous SJS-defined classes

### LINKED ISSUE #2009: Proposal: No own class for anonymous SJS-defined objects
## Context

#1795 introduced Scala.js-defined JS classes, traits, and objects. And, in particular, specified that anonymous classes extending `js.Any` would be Scala.js-defined by default. For example, in
```scala
@ScalaJSDefined
trait Foo extends js.Object {
  def x: Int
  def y: Any
}

val obj = new js.Object with Foo {
  val x = 5
  val y = "hello"
}
```
the anonymous class is automatically Scala.js-defined.

As currently specified, an actual JavaScript class is created for this anonymous class, so the ES6 semantics look like this:
```javascript
class $anon extends Object {
  constructor() {
    super();
    this.x = 5;
    this.y = "hello";
  }
}

var obj = new $anon();
```
These semantics have two issues:
* They cause a lot of additional code in JS, compared to the "natural" way of writing the above code, which would be `var obj = {x: 5; y: "hello"}`
* As reported in #2006, such objects do not count as "plain objects" because an "invisible" prototype wound its way into the object's prototype chain.

More globally, the semantics are just not what would be *expected* by default by the developer.

## The proposal at large

I propose to refine the specification from #1795 for *anonymous* classes extending `js.Any`. Instead of simply stating that they be Scala.js-defined JS classes by default, we state the following. An `new` expression for an anonymous class extending `js.Any` is *sugar* for instantiating the *parent* class, then adding fields *on the object itself*. There will be no additional class/prototype.

For example, the above Scala.js example "desugars" into the following Scala.js code (I'm intentionally evasive on what `T` and `U` are, at this point):
```scala
val obj = {
  val o: T = new js.Object().asInstanceOf[T]
  o.asInstanceOf[js.Dynamic].x = 5
  o.asInstanceOf[js.Dynamic].y = "hello"
  o.asInstanceOf[U]
}
```
which would, according to existing semantics, eventually become:
```javascript
var o = {}; // new js.Object() produces {}
o.x = 5;
o.y = "hello";
var obj = o;
```
which is immensely closer to the ideal code. In fact, experience suggests that Closure is able to optimize that into:
```javascript
var o = {x: 5, y: "hello"};
```
which *is* the ideal code.

### Type `T` of `o` and type `U` of the result

The types `T` and `U` don't really matter to understand practically the result of the desugaring. For all practical purposes, `T` and `U` can be thought as `js.Dynamic`, with appropriate casts everywhere.

However, we need to carefully craft them to guarantee that the result of the desugaring is still well-typed from scalac's point of view.

* The type `T` of the temporary `val o` is the *self type* of the anonymous class, as seen from within it, but rewritten not to refer to the `$anon` class symbol itself (since it won't exist after desugaring).
* The type `U` of the result is the `tpe` of the scalac `Block` defining the anonymous class.

In practice, `T` and `U` should both look like `js.Object with Foo { val x: Int; val y: String }`, but there might be some weird scalac details. 

### Parent class and interfaces

* Since a Scala.js-defined JS class must inherit from a JS class, there is always a super class to instantiate.
* Since they can only mix in Scala.js-defined JS traits which are guaranteed to be pure interfaces (no concrete term member), there is nothing to actually mix in during the desugaring.
* Super constructor parameters are appropriately forwarded.

Essentially, the "header"
```scala
new ParentClass(arg1, ..., argN) with Trait1 with Trait2
```
desugars into
```scala
val o: T = new ParentClass(arg1, ..., argN).asInstanceOf[T]
```

### Fields

* `val`s and `var`s in the anonymous class desugar to adding the corresponding fields on the object being constructed.
* References to `val`s and `var`s are desugared into reading the fields from the object

```scala
val x = something
var y = x + somethingElse
```
desugars into
```scala
o.x = something
o.y = o.x + somethingElse
```
or, to be precise, as `o` is a `T` and we are past typer:
```scala
o.asInstanceOf[js.Dynamic].updateDynamic("x")(something)
o.asInstanceOf[js.Dynamic].updateDynamic("y")(
    o.asInstanceOf[js.Dynamic].selectDynamic("x") + somethingElse)
```

### Other statements in the class body

Other statements in the class body, i.e., in the constructor of the anonymous class, are desugared in order, appropriately interleaved with the definition of fields.

### References to `this`

References to `this` in the body of the anonymous class desugar into the temporary `val` of the object being constructed, i.e., `o`.

### Methods

`def`s with parentheses desugar into *fields* added on the object, which are `js.ThisFunction` values:
```scala
def m(a: Int, b: Int): Int = a + b + this.x
```
desugars into
```scala
o.m = js.ThisFunction.fromFunction3({ ($this: o.type, a: Int, b: Int) =>
  a + b + $this.asInstanceOf[js.Dynamic].selectDynamic("x")
})
```
Note that, in the body of `def`s, references to `this` are desugared into `$this` (not `o`), and that `$this` has type `o.type`.

### Getters and setters

Getters (`def`s without `()`) and setters (`def`s whose name ends in `_=`) are desugared into JS getters and setters on the object `o` itself, invoking `Object.defineProperty`:
```scala
def foo: Int = x
def foo_=(v: Int): Unit = x = v
```
is desugared into
```scala
js.Dynamic.global.Object.defineProperty(o, "foo", js.Dynamic.literal(
    "get" -> js.ThisFunction.fromFunction1({ ($this: o.type) => $this.x }),
    "set" -> js.ThisFunction.fromFunction2({ ($this: o.type, v: Int) => $this.x = v })
))
```

## Implementation

This desugaring should be implemented as an additional compiler transformation on the scalac `Tree`s. It must happen before `explicitouter`, which comes before `erasure`. I think the best place is to put it between `refchecks` and `uncurry`.

The implementation will be *very* difficult. I do not expect this feature to be shipped any time soon. And I personally will not have the time to implement it. But it seems *possible*, and the spec looks very nice to me.

gzm0:


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/util/VarBox.scala
```diff
@@ -0,0 +1,9 @@
+package org.scalajs.core.compiler.util
+
+import language.implicitConversions
+
+class VarBox[T](var value: T)
+
+object VarBox {
+  implicit def toValue[T](box: VarBox[T]): T = box.value
```
sjrd: This implicit bothers me, because `value` is fundamentally mutable. I think it should be removed, as well as `toValueFromBox`. And instead write `.value` explicitly when we want to read the value.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -124,39 +125,85 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+    private val paramAccessorLocals = new ScopedVar(Map.empty[Symbol, js.ParamDef])
 
-    val tryingToGenMethodAsJSFunction = new ScopedVar[Boolean](false)
-    class CancelGenMethodAsJSFunction(message: String)
+    private class CancelGenMethodAsJSFunction(message: String)
         extends Throwable(message) with scala.util.control.ControlThrowable
 
-    // Rewriting of anonymous function classes ---------------------------------
+    //// For matches
+    private val countsOfReturnsToMatchEnd = new ScopedVar[VarBox[Int]]
```
sjrd: I don't think dropping the `Map` is valid. In case of nested matches, or worse, in the case of a `match` in the *guard* of another `match`, I'm pretty sure it's possible to have jumps to a non-directly enclosing matchEnd. I think this should be a `ScopedVar[mutable.Map]`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -124,39 +125,85 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-    val currentClassSym          = new ScopedVar[Symbol]
-    val currentMethodSym         = new ScopedVar[Symbol]
-    val thisLocalVarIdent        = new ScopedVar[Option[js.Ident]](None)
-    val fakeTailJumpParamRepl    = new ScopedVar[(Symbol, Symbol)]((NoSymbol, NoSymbol))
-    val enclosingLabelDefParams  = new ScopedVar(Map.empty[Symbol, List[Symbol]])
-    val mutableLocalVars         = new ScopedVar[mutable.Set[Symbol]]
-    val mutatedLocalVars         = new ScopedVar[mutable.Set[Symbol]]
-    val unexpectedMutatedFields  = new ScopedVar[mutable.Set[Symbol]]
-    val paramAccessorLocals      = new ScopedVar(Map.empty[Symbol, js.ParamDef])
+    //// Per class body
```
sjrd: Why 4 `/`s?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -124,39 +125,85 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+    val currentClassSym = new ScopedVar[Symbol]
```
sjrd: `private`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -201,29 +246,29 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-        /* First gen and record lambdas for js.FunctionN and js.ThisFunctionN.
-         * Since they are SAMs, there cannot be dependencies within this set,
-         * and hence we are sure we can record them before they are used,
-         * which is critical for these.
+        /* There are three types of anonymous classes we want to generate
+         * only when we need them so we can inline them at construction site:
+         *
+         * - lambdas for js.FunctionN and js.ThisFunctionN (SAMs).
+         * - lambdas for scala.FunctionN. Note that these may fail.
+         * - anonymous Scala.js defined classes. We'll inline the constructor.
```
sjrd: Scala.js-defined **JS** classes

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -201,29 +246,29 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+         *
+         * Note that the first two are optimizations (we'd rather have a cheap
+         * JS lambda, than a fully fleged scala class).
```
sjrd: The first one is a requirement, too. We cannot generate an actual class extending `js.(This)FunctionN`. We must recover them into lambdas.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -264,6 +309,9 @@ abstract class GenJSCode extends plugins.PluginComponent
           }
         }
 
+        assert(lazilyGeneratedAnonClasses.isEmpty, "Unused anonymous classes: " +
+            lazilyGeneratedAnonClasses.keys.map(_.fullName).mkString)
```
sjrd: `mkString` without any argument does not separate the elements at all, IIRC.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      val origJsClass =
+        nestedGenerateClass(sym)(genScalaJSDefinedJSClass(classDef))
+
+      // Partition members into the ones we inline and the ones we keep
+      val retainedMembers = ListBuffer.empty[js.Tree]
+      var constructor: Option[js.MethodDef] = None
+      val exportedMembers = ListBuffer.empty[js.Tree]
+
+      origJsClass.defs.foreach {
+        case _: js.FieldDef => // ignore. we get it from the ctor
```
sjrd: You actually need to keep them, because of fields with no initializer (`var f: Int = _`)

sjrd: And, even for the others, to maintain initialization behavior with 0s and nulls.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          defs = retainedMembers.toList)(
+          origJsClass.optimizerHints)(origJsClass.pos)
+
+      // Construct inline class definition
+      assert(constructor.isDefined, "No ctor found")
+      val js.MethodDef(_, _, ctorParams, _, ctorBody) = constructor.get
+
+      assert(ctorParams.size == args.size)
+      val argDefs = for ((param, arg) <- ctorParams zip args)
+        yield js.VarDef(param.name, param.ptpe, param.mutable, arg)(param.pos)
```
sjrd: Don't you need `freshLocalIdent`s, here?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      val selfName = freshLocalIdent("self")(pos)
```
sjrd: `this`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        case mdef: js.MethodDef =>
+          implicit val pos = mdef.pos
+          val name = mdef.name.asInstanceOf[js.StringLiteral]
+          val impl = lambda(mdef.args, mdef.body)
+          js.Assign(js.JSBracketSelect(selfRef, name), impl)
+
+        case pdef: js.PropertyDef =>
+          implicit val pos = pdef.pos
+          val name = pdef.name.asInstanceOf[js.StringLiteral]
+          val jsObject = js.JSDotSelect(genLoadGlobal(), js.Ident("Object"))
```
sjrd: `JSBracketSelect`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          def accessor(tree: js.Tree, name: String)(genBody: js.Tree => js.Tree) =
+            if (tree == js.EmptyTree) Nil
+            else List(js.StringLiteral(name) -> genBody(tree))
+
+          val accessorObj = js.JSObjectConstr(
+              accessor(pdef.getterBody, "get")(lambda(params = Nil, _)) ++
+              accessor(pdef.setterBody, "set")(lambda(pdef.setterArg :: Nil, _))
+          )
+
+          js.JSDotMethodApply(jsObject, js.Ident("defineProperty"),
```
sjrd: `JSBracketMethodApply`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          val jsObject = js.JSDotSelect(genLoadGlobal(), js.Ident("Object"))
+
+          def accessor(tree: js.Tree, name: String)(genBody: js.Tree => js.Tree) =
+            if (tree == js.EmptyTree) Nil
+            else List(js.StringLiteral(name) -> genBody(tree))
+
+          val accessorObj = js.JSObjectConstr(
+              accessor(pdef.getterBody, "get")(lambda(params = Nil, _)) ++
+              accessor(pdef.setterBody, "set")(lambda(pdef.setterArg :: Nil, _))
+          )
```
sjrd: You should also set `configurable` and `enumerable` to `true`, to be consistent with the simple `=` for methods.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +532,133 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+          js.JSDotMethodApply(jsObject, js.Ident("defineProperty"),
+              List(selfRef, name, accessorObj))
+
+        case tree =>
+          sys.error("Unexpected tree: " + tree)
+      }
+
+      val inlinedCtorStats = new ir.Transformers.Transformer {
+        override def transform(tree: js.Tree, isStat: Boolean) = tree match {
```
sjrd: That's a public method. It should have an explicit result type.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: In general, what if the constructor defines local variables with names that clash with local variables in the current method?

One easy solution would be to emit an immediately-applied `js.Closure` for the inline class-that-became-an-object. Something like:
```scala
js.JSFunctionApply(
    js.Closure(Nil, ctorParams, transformedCtorBody, Nil),
    actualArgsToTheConstructor)
```
or, in IR syntax:
```scala
(lambda<>(...ctorParams: any): any = {
  someStatements;
  val this$1 = new SuperClass(...parentArgs);
  exportMethodsAndProperties;
  someMoreStatements;
  this$1
})(...actualArgsToTheConstructor)
```
It's not optimal, but we can let the optimizer deal with inlining the `js.Closure` later. It's good at it ^^

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4070,31 +4255,26 @@ abstract class GenJSCode extends plugins.PluginComponent
      *
      *  so that, at instantiation point, we can write:
      *
-     *    new AnonFunctionN(functionMaker(this, captured1, ..., capturedM))
+     *    new AnonFunctionN(function)
+     *
+     *  this latter tree is returned in case of success.
```
sjrd: "this latter" is weird. "the latter"?

library/src/main/scala/scala/scalajs/js/annotation/SJSDefinedAnonymousClass.scala
```diff
@@ -0,0 +1,10 @@
+package scala.scalajs.js.annotation
+
+/** IMPLEMENTATION DETAIL: Marks Scala.js defined anonymous classes.
```
sjrd: "Marks anonymous Scala.js-defined JS classes."?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Such a rewriting would also protect from applying the transform twice on a nested anonymous JS class (where the constructor of the outer one defines an anonymous inner one!) -> probably needs a test, that one ^^

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -305,6 +305,94 @@ class ScalaJSDefinedTest {
...
+    assertEquals(3, obj.asInstanceOf[js.Dynamic].z)
+  }
+
+  @Test def anonymous_class_using_own_method_in_ctor(): Unit = {
+    val obj = new js.Object {
+      val y = inc(0)
+      def inc(x: Int) = x + 1
+    }
+
+    assertEquals(obj.asInstanceOf[js.Dynamic].y, 1)
```
sjrd: The constant should on the left.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -305,6 +305,94 @@ class ScalaJSDefinedTest {
...
+    val inner = outer.inner
+    assertEquals(outer.x, 1)
+    inner.inc()
+    assertEquals(outer.x, 2)
+  }
+
+  @Test def nested_anonymous_classes_and_lambdas(): Unit = {
+    def call(f: Int => js.Any) = f(1)
+
+    // Also check, that f's capture is properly transformed.
```
sjrd: No `,` before "that"

sjrd:
That's all.

compiler/src/main/scala/org/scalajs/core/compiler/util/VarBox.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Dropped them.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Can't make this private. `GenJSExports` is using this.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Indeed.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Sure.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Done.

library/src/main/scala/scala/scalajs/js/annotation/SJSDefinedAnonymousClass.scala
gzm0: Done.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
gzm0: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: > probably needs a test, that one ^^

Did you have something more in mind than `nested_anonymous_classes`?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
gzm0: Done.

gzm0:
Note to self: Tests to add:

- Uninitialized field
- Field definition order
- Name collision (local name / ctor param)

gzm0:
Updated.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -124,39 +125,84 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          thisLocalVarIdent := null,
+          fakeTailJumpParamRepl := null,
+          enclosingLabelDefParams := null,
+          isModuleInitialized := null,
+          undefinedDefaultParams := null,
+          mutableLocalVars := null,
+          mutatedLocalVars := null,
+          tryingToGenMethodAsJSFunction := false,
+          paramAccessorLocals := Map.empty,
+          countsOfReturnsToMatchEnd := null
```
gzm0: maybe reorder this?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -264,6 +307,9 @@ abstract class GenJSCode extends plugins.PluginComponent
           }
         }
 
+        assert(lazilyGeneratedAnonClasses.isEmpty, "Unused anonymous classes: " +
+            lazilyGeneratedAnonClasses.keys.map(_.fullName).mkString(", "))
```
gzm0: This is not a valid check anymore, since a class might be used during `GenMethodAsJSFunction` and not be removed from the map.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +530,157 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          classMembers += property
+
+        case tree =>
+          sys.error("Unexpected tree: " + tree)
+      }
+
+      // Only emit static members to class
+      generatedClasses += sym -> origJsClass.copy(
+          defs = staticMembers.toList)(
+          origJsClass.optimizerHints)(origJsClass.pos)
```
sjrd: You might also want to turn the class into a normal Scala class extending `Object`, here. So maybe it's easier to create a new `ClassDef` from scratch, rather than copying `origJsClass`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -486,6 +530,157 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      // Construct inline class definition
+      assert(constructor.isDefined, "No ctor found")
+      val js.MethodDef(_, _, ctorParams, _, ctorBody) = constructor.get
```
sjrd: We typically use the `.getOrElse(throw Something)` idiom instead of `assert(x.isDefined)`

sjrd:
That's all.

gzm0:
Updated.

gzm0:
It seems the IR checker does not like that I change the class type... @sjrd do you know what is going on?

sjrd:
Ah yes, I understand. The static methods generated for the actual implementations of the member methods take an explicit `$this` parameter, which of course is an `AnyType` because it's a JS value. But the *signature* of the method uses the full name of the class. The IR checker therefore complains that the `$this` param should conform to the IR equivalent of the type found in the signature. And since that type is now a Scala class type, the IR type cannot be `AnyType`.

I think the best solution is to use a `ClassKind.RawJSType` then, instead of a `ClassKind.Class`. Basically with the same structure as the one produced for a raw JS trait such as `js.Any`.

gzm0:
Updated.

sjrd:
LGTM :-D

PR #2293, opened 2016-03-14
Port test suite source map test.

nicolasstucki:

Issue #2248, opened 2016-02-19
Run JUnit tests in HTML test-suite runner


gzm0:
Our HTML test-suite runner currently relies on `run-jasmine-tests.js` which (so says the name), does not run JUnit tests.

sjrd:
This would be addressed by #1709, wouldn't it?

gzm0:
Yes. I thought about #1709 as something more user facing, but maybe it is a good way to get some experience with writing HTML runners.

gzm0:
I'll take a stab at implementing something simple that exports detected tests from sbt and then runs them in the browser today.

Issue #1709, opened 2015-06-11
Provide an HTML runner for tests


gzm0:
This will allow to:

- Run tests in the browser
- Debug tests in the browser (since they run in the browser)
- Have a nice HTML report

This can be completely agnostic of the underlying test framework and hopefully even implemented outside of scala-js core.

The only issue being test detection which is currently done by sbt and hard to port to JavaScript.

sjrd:
We could have sbt do the detection and generate a json report of what tests to execute.

gzm0:
Good thinking. I guess that would be useful in general. On the other hand, we might need some sbt foo anyway to do this in a convenient way...

gzm0:
Ironically, the biggest difficulty seems to be where to put the HTML file that the users can open :S

An additional sbt plugin would of course trivially fix that, but I was trying to avoid yet another plugin  ...

sjrd:
What's wrong with `crossTarget.value`, i.e., `target/scala-2.11/`?

gzm0:
If the HTML runner is a library, the HTML file is just inside the JAR as a resource and wont get unpacked. If we want to generate/unpack/place it, we need sbt instrumentation. This can either be done via a plugin or by telling the user. (We could also tell the user to just create/place the file...)

I don't like either of these options...

sjrd:
Oh, because you want to put implement that outside of core? I don't see why.

gzm0:
If it doesn't have to be in core, it shouldn't. Also, it would be nice to rely at least on the statically typed DOM and maybe jQuery to do this...

sjrd:
In that case, I'm pretty sure we do need an additional sbt plugin. I don't think asking people to do it by hand is an option, if we can instead provide an sbt plugin that does it for them.

mysticfall:
Maybe it won't be a very helpful comment, but I had the exact same requirements for my project.

So I had to implement a HTML based test UI, with which you can run individual tests and get visual feedback.

I think it looks and works quite nice, but it depends on specific libraries and there's no test case detection. So, I'm definitely interested in learning how to do it in more generalized/standardized way.

InTheNow:
FWIW - I sort-of added support for [saucelabs](https://github.com/InTheNow/zcheck/tree/master/saucelabs) and got the tests running there. Any solution might want to take this into scope. 
[...]

PR #2296, opened 2016-03-16
Port tools/.../TestRunner to JUnit.

nicolasstucki:


nicolasstucki:
Review by @sjrd.

nicolasstucki:
Commit extracted from #2293.

tools/js/src/test/scala/org/scalajs/core/tools/test/js/TestRunner.scala
```diff
@@ -15,8 +15,8 @@ object TestRunner {
 
   @JSExport
   def runTests(): Unit = {
-    val framework = new JasmineFramework()
-    val runner = framework.runner(Array("-ttypedarray"), Array(),
+    val framework = new JUnitFramework()
+    val runner = framework.runner(Array(), Array("scalajs.typedarray=true"),
```
sjrd: Could we instead simply do `System.setProperty` here?

sjrd:
The [`TestDetector`](https://github.com/scala-js/scala-js/blob/master/test-suite/js/src/test/scala/org/scalajs/testsuite/utils/TestDetector.scala) also needs to be adapted to detect JUnit boostrappers instead of Jasmine tests. Otherwise this doesn't do anything anymore.

sjrd:
And [`run-jasmine-test.js`](https://github.com/scala-js/scala-js/blob/master/test-suite/js/run-jasmine-tests.js) also needs work, as it uses the `TestDetector`.

Which means there's an impact on the HTML test runners. And which means this is linked to #2248.

gzm0:
Unfortunately this isn't that easy. Jasmine comes with a "test scheduler" for the browser. JUnit doesn't. So we almost need a fully fledged #1709 to deal with that. Feel free to steal the ticket from me if you need this now.

sjrd:
> Jasmine comes with a "test scheduler" for the browser. JUnit doesn't.

I guess, as a first step, we could simply run everything in "the main thread" (blockingly), using a stupid simple console reporter. We'd have to open the console window of the browser to see the results, but at least we would have *something*.

tools/js/src/test/scala/org/scalajs/core/tools/test/js/TestRunner.scala
nicolasstucki: We could.

test-suite/js/scalajs-test-suite-2.10.html
```diff
@@ -3,15 +3,24 @@
 <head>
   <title>Scala.js Test Suite - Jasmine HTML reporter</title>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
-  <link rel="stylesheet" type="text/css" href="http://cdn.jsdelivr.net/jasmine/1.3.1/jasmine.css"></link>
 </head>
 <body>
 <script type="text/javascript" src="http://cdn.jsdelivr.net/jasmine/1.3.1/jasmine.js"></script>
```
nicolasstucki: Note that this is kept because there are still a couple of tests that use info from the jasmine framework. This will be removed in the next PR.

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/JUnitTests.scala
```diff
@@ -0,0 +1,57 @@
...
+      testDef match {
+        case testDef: JUnitTestBootstrapper =>
+          val runner = new JUnitSlaveRunner(Array(), Array(), classLoader,
+              (m: String) => masterRunner.receiveMessage(m), runSettings)
+          val name = testDef.toString.split("\\$scalajs\\$junit\\$bootstrapper").head
+          val eventHandler = new EventHandler {
+            def handle(event: Event): Unit = ()
+          }
+          val richLogger = new RichLogger(loggers, runSettings, name)
+          val taskDef = new TaskDef(name, JUnitFingerprint, false, Array())
```
sjrd: Instead of exposing `JUnitFingerprint` in an ad hoc way, could we have `val framework = new JUnitFramework` and then use `framework.fingerprints.head`, like we do in `TestRunner`?

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/JUnitTests.scala
```diff
@@ -0,0 +1,57 @@
...
+          val executor = new JUnitExecuteTest(taskDef, runner, testDef,
+            richLogger, eventHandler)
```
sjrd: 4 spaces

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/TestDetector.scala
```diff
@@ -22,11 +22,14 @@ object TestDetector {
       "SJSDefinedAutoExportedIgnoreClassClass"
   ).map(basePackage + ".jsinterop." + _)
 
-  def detectTestNames(): List[String] = detectTestsInternal().map(_._2).toList
+  def detectTestNames(): List[String] = detectTestsInternal().map(_._2)
 
   @JSExport
   def loadDetectedTests(): Unit = detectTestsInternal().foreach(_._1())
 
+  def getDetectedTests(): List[js.Dynamic] =
```
sjrd: Rename to `detectTests()`?

test-suite/js/scalajs-test-suite-2.10-fastopt.html
```diff
@@ -3,15 +3,24 @@
...
+    }
+  };
+</script>
 <script type="text/javascript" src="./target/scala-2.10/scalajs-test-suite-test-fastopt.js"></script>
-<script type="text/javascript">var isFullOpt = false;</script>
-<script type="text/javascript" src="run-jasmine-tests.js"></script>
+<script type="text/javascript" src="run-junit-tests.js"></script>
+
+Scala.js test suite. Open the console to see the output.
+Reload the page with the console open if the logs where not shown.
```
sjrd: These two lines should be in a `<p>` tag.

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/JUnitTests.scala
```diff
@@ -0,0 +1,57 @@
...
+          JUnitTask.printStart(richLogger, runner)
+          val startTime = System.nanoTime
+          executor.executeTests()
+          val time = System.nanoTime - startTime
+          JUnitTask.printEnd(richLogger, runner, time)
```
sjrd: This looks like a vague copy-paste of `JUnitTask.execute`. Why can't we use `new JUnitTask(...).execute(...)` directly? That would also allow not to expose `printStart` and `printEnd`.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitBaseRunner.scala
```diff
@@ -19,12 +19,12 @@ abstract class JUnitBaseRunner(
...
-  private[junit] def taskFailedCount: Int = failedCount
-  private[junit] def taskIgnoredCount: Int = ignoredCount
-  private[junit] def taskSkippedCount: Int = skippedCount
-  private[junit] def taskTotalCount: Int = totalCount
+  def taskDoneCount: Int = doneCount
+  def taskPassedCount: Int = passedCount
+  def taskFailedCount: Int = failedCount
+  def taskIgnoredCount: Int = ignoredCount
+  def taskSkippedCount: Int = skippedCount
+  def taskTotalCount: Int = totalCount
```
sjrd: Why were these things made public? They do not seem to be used anywhere in this PR.

sjrd:
That's all.

gzm0:
What's the status of this thing? I am looking into generating a test description JSON on the sbt side that can be pulled in to launch the tests. This will be completely agnostic to the testing framework that is used (since it uses the `sbt.testing` interface). Also, it will not rely on SJS internal hacks anymore (depending on how we emit things etc.).
With that will come a nice(r) HTML runner that eventually we can hopefully make publicly available.

However, given my pace, it will probably take a couple of weeks until this is fleshed out (prototype HTML runner is working though). So just wanted to make sure we are in sync about this.

sjrd:
> I am looking into generating a test description JSON on the sbt side that can be pulled in to launch the tests.

That would definitely supersede the special-case that this PR intends to fix, of course. This is mostly important to get some HTML tests to run again, before we have the more complete and generic solution.

gzm0:
All the stuff you are making public inside JUnit (if you still dobafter @sjrd's comments), can you make it `private[scalajs]` so we can make it private again once we move to a general solution?

sjrd:
I guess this PR is completely superseded by #2316?

PR #2310, opened 2016-03-29
Fix #2307: Avoid emitting dangling breaks to eliminated labels.

### LINKED ISSUE #2307: Wrong compilation of the operation on shapeless HList
The following(slightly simplified)snippet compiles and works as expected in scala but after compilation to scalaJs and launching I see the following exception:
```org.scalajs.jsenv.ExternalJSEnv$NonZeroExitException: Node.js exited with code 8```

```scala
import shapeless._, record._, syntax.singleton._

trait Crud[L <: HList, K, R] extends Serializable {
  type Out <: HList
  type V
  def apply(t: L, u: V=>R): (Out, V)
}

class UnsafeCrud(i: Int) extends Crud[HList, Any, Any] {
  type V = Any
  type Out = HList

  def unsafeUpdate(l: HList, i: Int, f: Any => Any): (HList, Any) = {
    def loop(l: HList, i: Int, prefix: List[Any]): (List[Any], HList, Any) =
      l match {
        case hd :: (tl: HList) => if (i == 0) (prefix, f(hd) :: tl, hd) else loop(tl, i - 1, hd :: prefix)
        case _ => throw new Exception("...")
      }

    val (prefix, suffix, v) = loop(l, i, Nil)
    prefix.foldLeft(suffix) { (tl, hd) => hd :: tl } -> v
  }

  def apply(l: HList, f: V => Any): (HList, Any) = unsafeUpdate(l, i, f)
}
type R = Record.`'i -> Int, 's -> String, 'c -> Char, 'j -> Int`.T
val r = 'i ->> 10 :: 's ->> "foo" :: 'c ->> 'x' :: 'j ->> 42 :: HNil

new UnsafeCrud(0).asInstanceOf[Crud[R,Witness.`'i`.T, String]].apply(r, _ => "newStr")
```

sjrd:


sjrd:
Review by @gzm0

sjrd:
Ping @gzm0 ?

gzm0:
Sorry. This one slipped past my filters.

gzm0:
Reviewing now.

gzm0:
So IIUC, the fixes here are the pendant to things like: https://github.com/sjrd/scala-js/blob/fix-dangling-break/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala#L1061

gzm0:
If so LGTM (otherwise more explanation required why we need change to label usage elimination and no change to label elimination).

sjrd:
> So IIUC, the fixes here are the pendant to things like

Yes, exactly.

gzm0:
LGTM in that case :)

PR #2312, opened 2016-03-30
Fix #2311: Minimal migration away from sbt's Build trait

### LINKED ISSUE #2311: Migrate away from our Build.scala build definition
It seems sbt folks have simply decided they'd drop support for `Build.scala` build definitions: https://github.com/sbt/sbt/pull/2524

So we should at least investigate now whether it's possible for our build to survive in those conditions.

dwijnand:
A minimal difference change that leaves most existing code where it is, for ease of review and inspiring confidence.

Fixes #2311


sjrd:
Dear Jenkins, ok to test

project/Build.scala
```diff
@@ -106,7 +106,7 @@ object Build extends sbt.Build {
     }
   }
 
-  val commonSettings = Seq(
+  lazy val commonSettings = inThisBuild(baseSettings) ++ Seq(
```
sjrd: That is not ideal. It will replicate `baseSettings` as many times as there are projects.

It would be better to simply write `inThisBuild(Build.baseSettings)` as an independent line in `build.sbt`.

dwijnand: Makes sense. Done.

dwijnand: It could also stay named "settings" or alternatively be named "buildSettings" (or "thisBuildSettings"?). Let me know if you prefer that.

sjrd:
```
<set>:1: error: not found: value makeCompliant
scalaJSSemantics in testSuite ~= makeCompliant
                                 ^
[error] Type error in expression
```
I guess you'll have to amend `ci/matrix.xml` to add a bunch of `Build.` before things like `makeCompliant`. Previously they were automatically in scope in the command line interface. It's a bit of a downgrade in usability from the command line, but as long as setting keys and projects are still in scope, it's not too bad.

project/Build.scala
sjrd: Maybe `thisBuildSettings` is the most explicit, so I'd recommend that one.

dwijnand:
I was thinking of adding forwarders to build.sbt. WDYT? I don't mind
personally.

sjrd:
> I was thinking of adding forwarders to build.sbt. WDYT?

I'm not sure what you mean. Maybe you're referring to some feature of sbt that I do not know yet. Please do as you were thinking, I'm curious :-p

dwijnand:
It's very advanced "sbt" technique of "val makeCompliant =
Build.makeCompliant" in build.sbt ;-)

Clearly dropping the Build trait pushes for a strong preference for things
being mostly in build.sbt (or using local AutoPlugins).

sjrd:
> It's very advanced "sbt" technique of "val makeCompliant = Build.makeCompliant" in build.sbt ;-)

Nice. I did not know that would put the val in the command line scope.

dwijnand:
It doesn't. I was tricked by the fact that it clearly does for `Build`.. :(

dwijnand:
@sjrd I created a local AutoPlugin to expose the values. WDYT of this approach? Happy to iterate.

sjrd:
Ah ah! Cute hack. Yes that sounds reasonable :)

project/Build.scala
```diff
@@ -38,7 +38,20 @@ import org.scalajs.core.tools.linker.backend.OutputMode
...
 import sbtassembly.AssemblyPlugin.autoImport._
 
-object Build extends sbt.Build {
+/*
+ * In sbt 0.13 the Build trait would expose all vals to the shell, where you can use them in "set a := b"
+ * like expressions. This re-exposes them.
+ */
+object ExposedValues extends AutoPlugin {
+  override def requires = plugins.JvmPlugin
+  override def trigger = allRequirements
```
sjrd: I don't think the trigger is necessary, nor is `requires`. We don't need the plugin to be applied on any project. We just want its `autoImport`s.

project/Build.scala
```diff
@@ -38,7 +38,20 @@ import org.scalajs.core.tools.linker.backend.OutputMode
...
+  object autoImport {
+    val makeCompliant = Build.makeCompliant
+    val ES6NodeArgs   = Build.ES6NodeArgs
```
sjrd: We do not use vertical alignment in our coding style.

dwijnand: Sure. Shall I drop it from build.sbt too?

project/Build.scala
```diff
@@ -38,7 +38,20 @@ import org.scalajs.core.tools.linker.backend.OutputMode
 
 import sbtassembly.AssemblyPlugin.autoImport._
 
-object Build extends sbt.Build {
+/*
+ * In sbt 0.13 the Build trait would expose all vals to the shell, where you can use them in "set a := b"
+ * like expressions. This re-exposes them.
```
sjrd: Try to keep line length under 80 characters, as is the soft rule in our coding style. By the way, we start our multi-line comments on the first line, so it should look like this:
```scala
/* In sbt 0.13. the Build trait ...
 * ... This re-exposes them.
 */
```
(yes, we are coding style nazis ^^ see https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md)

dwijnand: I don't mind following coding styles.

Out of curiosity, what's the rational for starting on the first line? My rational for second line is weak: habit :D

project/Build.scala
dwijnand: Good call - just checked and it works (TIL).

dwijnand:
Pushed updates.

project/Build.scala
sjrd: Good question. I think it's mostly habit in this case too, but it is a codified habit. After all, most coding style rules are habits. I could argue that it is more consistent with the Scaladoc comments, which also start on the first line, but I do not think that is why I started doing it that way in the first place.

project/Build.scala
sjrd: I am less opinionated about the `build.sbt` in this case. I would lean towards following the coding style anyway, and not use vertical alignment there either.

dwijnand:
Looks like all the runs for this PR were aborted?

sjrd:
Yes, I aborted them. They had not started yet, so I canceled the intermediary commits.

sjrd:
Oh no, wait, there are many more cancellations than there were supposed to be. I'll wipe everything and start over. Give me a sec'.

dwijnand:
Cool. Thanks.

sjrd:
Jenkins, retest this please

sjrd:
Looks like it passes :)

Could you please squash all the commits into one, with the following commit message:

    Fix #2311: Minimal migration away from sbt's Build trait

    A minimal difference change that leaves most existing code where it is,
    for ease of review and inspiring confidence.

(with `Fix #2311` at the beginning of the title line)

dwijnand:
Done.

sjrd:
LGTM Thanks!

gzm0:
@sjrd Do we have a story on how we clean this up as we go? While I understand that this is a necessary change so we can move forward in sbt's versions, I can hardly imagine that we want it to stay this way.

sjrd:
> I can hardly imagine that we want it to stay this way.

Why not? The code in this PR looks reasonable. We can keep editing `Build.scala` in an IDE, which is important given that it's 1 kLoC-long. If sbt 1.0 is happy with this pattern, I'm happy too.

gzm0:
> Why not?

* We keep two lists of projects that we need to keep in sync
* `thisBuildSettings` becomes a "magic" value in `Build.scala` (w.r.t. to other settings in the file)
* `Build.scala` seems poorly named (we named it that way because sbt auto-imported things)
* We need an auto plugin just to make our CI matrix work. We should either:
    * Have a normal location where we put all these (and live with longer names in the matrix)
    * Simplify everything in the matrix (`RetryingComJSEnv` and `OutputMode` comes to mind).
  
    This is admittedly not too related to this change, but IMHO becomes more flagrant when we need ad-hoc plugins to support it.

sjrd:
> We keep two lists of projects that we need to keep in sync

True. That is not very nice. But at least it can hardly *not* be in sync and have the build still, well, build.

> `thisBuildSettings` becomes a "magic" value in `Build.scala` (w.r.t. to other settings in the file)

I do not understand that argument. It's a public API, nothing else. Just like the `projects` are, in fact. And like `BinaryIncompatibilities.Tools` is, etc.

> `Build.scala` seems poorly named (we named it that way because sbt auto-imported things)

Hum, yes. But in a sense it still contains basically the entire build. I guess it could be `BuildDefinitions` or something like this.

> We need an auto plugin just to make our CI matrix work.

Previously we had "magic" `val`s in `Build` just to make our CI matrix work. I don't really understand the difference.

I would have preferred to keep the everything-in-`Build.scala`, which had its beauty. Now we have two choices: either this, or move all of `Build.scala` to `build.sbt`. I dislike the latter mostly because I don't want to edit a very large `.sbt` file which contains a lot of non-trivial logic. Also, that does not solve the put-things-in-cli-scope problem. The former has its problems, as you mention, but it allows to keep our logic in a true `.scala` file.

That said, I'm still open to see what we would get with everything-in-`build.sbt`. I'm just not convinced that `.sbt` is a good abstraction for large and complex builds like ours.

gzm0:
Ok. Let's not block on this and move ahead with this PR. I can come up with a proposal on how to change it later if necessary.

sjrd:
OK let's merge this, then.

PR #2316, opened 2016-04-01
Add test detection JSON report with HTML runner.

### LINKED ISSUE #2248: Run JUnit tests in HTML test-suite runner

gzm0:
Test detection done on sbt side is now reported via JSON to the HTML
runner. The HTML runner is therefore fully agnostic to the exact
Scala.js testing framework that is in use.
The same principle is applied in the bootstrapping test.

This fixes #2248 (Run JUnit tests in HTML test-suite runner) and takes
the first steps towards #1709 (Provide an [outside facing] HTML runner
for tests).

Not yet supported is picking up framework flags from sbt. This can be
easily added by extending the mechanism if necessary.

HTML runner looks like this:
HTML Runner Screenshot


gzm0:
WDYT? @nicolasstucki @sjrd

sjrd:
On holiday until next Sunday with nothing but my phone to (sporadically) access Internet. So I'll have a look at this when I'm back.

gzm0:
Sure. No problem.
Enjoy your vacation!
On Apr 3, 2016 11:07 PM, Sébastien Doeraene <notifications@github.com> wrote:On holiday until next Sunday with nothing but my phone to (sporadically) access Internet. So I'll have a look at this when I'm back.

nicolasstucki:
Also on holiday ;)

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/HTMLRunner.scala
```diff
@@ -0,0 +1,136 @@
...
+    def trace(t: Throwable): Unit = UI.log(t.toString, "error")
+  }
+
+  private object EventCounter extends EventHandler {
+    val counts = mutable.Map.empty[Status, Int].withDefaultValue(0)
+    def handle(event: Event): Unit = counts(event.status) += 1
+  }
+
+  def main(): Unit = {
+    System.setProperty("scalajs.testsuite.testtag", "testtag.value")
```
nicolasstucki: This tag property should go directly in the `__ScalaJSEnv`. It is used in the test that checks that system properties are loaded correctly from the environment.

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/HTMLRunner.scala
```diff
@@ -0,0 +1,136 @@
...
+    System.setProperty("scalajs.browser", "true")
+    System.setProperty("scalajs.typedarray", "true")
```
nicolasstucki: Is this true for every browser?

gzm0: This isn't for every browser, but the browsers we test in. So yes, it was an assumption we had in our runners before.

nicolasstucki: Ok

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
     base = file("test-suite/js"),
     settings = commonSettings ++ myScalaJSSettings ++ testTagSettings ++
       testSuiteCommonSettings(isJSTest = true) ++ Seq(
-        name := "Scala.js test suite",
+        name := "Scala.js test suite"
+      ) ++ inConfig(Test)(Seq(
+        artifactPath in jsTestDefinitions :=
+          crossTarget.value / "test_definitions.js",
```
sjrd: Nitpick: use `-` instead of `_`. AFAIK we've always used `-` in file names so far, and never `_`.

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
...
+
+        // Generate a JS file that contains defined tests and testFrameworks.
+        // The file will be called "test_definitions.js" and defines:
+        // - var definedTests
+        // - var testFrameworkNames
```
sjrd: `/* ... */` for multi-line comments, according to our style guide ^^

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
...
+        jsTestDefinitions := {
```
sjrd: I guess the job done by this task should eventually belong to the sbt plugin to address #1709?

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
...
+
+          val allFrameworks =
+            loadedTestFrameworks.value.map(_._1.implClassNames.toList)
+
+          val writer = new BufferedWriter(new FileWriter(outFile))
+
+          try {
+            writer.write("var definedTests = ")
+            writeJSON(allTests.toJSON, writer)
+            writer.write(";");
```
sjrd: `\n`?

Also spurious `;`

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
...
+
+            writer.write("var testFrameworkNames = ")
+            writeJSON(allFrameworks.toList.toJSON, writer)
+            writer.write(";")
```
sjrd: `\n`?

project/Build.scala
```diff
@@ -1216,8 +1226,50 @@ object Build extends sbt.Build {
...
+          } finally {
+            writer.close()
+          }
+
+          outFile
+        },
+        jsTestDefinitions <<= jsTestDefinitions triggeredBy definedTests,
```
sjrd: Why?

test-suite/js/scalajs-test-suite-2.10-fastopt.html
```diff
@@ -1,17 +1,16 @@
...
-  <title>Scala.js Test Suite - Jasmine HTML reporter</title>
+  <title>Scala.js Test Suite</title>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
-  <link rel="stylesheet" type="text/css" href="http://cdn.jsdelivr.net/jasmine/1.3.1/jasmine.css"></link>
+  <link rel="stylesheet" type="text/css" href="test-stylesheet.css">
+  <script type="text/javascript">
+    var __ScalaJSEnv = {"javaSystemProperties": { "scalajs.fastopt-stage": "true" }};
+  </script>
+  <script type="text/javascript" src="./target/scala-2.10/scalajs-test-suite-test-jsdeps.js"></script>
+  <script type="text/javascript" src="./target/scala-2.10/test_definitions.js"></script>
```
sjrd: At https://github.com/scala-js/scala-js/pull/2316/files#diff-215113124f1de02f228327ba7abb45f7R565 you add `test_definitions.js` to the `jsDependencies`, but here you include `jsdeps` + `test_definitions`. That does not look good.

test-suite/js/scalajs-test-suite-2.10-fastopt.html
```diff
@@ -1,17 +1,16 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>Scala.js Test Suite - Jasmine HTML reporter</title>
+  <title>Scala.js Test Suite</title>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
-  <link rel="stylesheet" type="text/css" href="http://cdn.jsdelivr.net/jasmine/1.3.1/jasmine.css"></link>
+  <link rel="stylesheet" type="text/css" href="test-stylesheet.css">
+  <script type="text/javascript">
+    var __ScalaJSEnv = {"javaSystemProperties": { "scalajs.fastopt-stage": "true" }};
```
sjrd: Shouldn't you also add `scalajs.browser`?

sjrd: And `scalajs.typedarray`?

test-suite/js/src/test/scala/org/scalajs/testsuite/utils/TestDetector.scala
```diff
@@ -1,61 +1,61 @@
...
+        a.isModule == b.isModule && a.superclassName == b.superclassName
+
+      case (a: AnnotatedFingerprint, b: AnnotatedFingerprint) =>
+        a.isModule == b.isModule && a.annotationName == b.annotationName
+
+      case _ => false
+    }
+  }
+
+  @js.annotation.JSName("foo") // needed, otherwise compiler complains
```
sjrd: IIUC you can remove this if you rebase on master, now.

test-suite/js/test-stylesheet.css
```diff
@@ -0,0 +1,37 @@
+body {
+    margin: 1em;
```
sjrd: Our only existing .css file https://github.com/scala-js/scala-js/blob/master/assets/additional-doc-styles.css uses 2-space indentation. We should consistently keep that.

sjrd:
That's all. Looks nice :)

project/Build.scala
gzm0: Yes, indeed. But I want us to get some experience first in using this HTML runner. Eventually, this should indeed go into the sbt plugin.

project/Build.scala
gzm0: Otherwise the "test_definitions.js" wont be generated for the `testSuite` project.

test-suite/js/scalajs-test-suite-2.10-fastopt.html
gzm0: I add it to `toolsJS/test:jsDependencies` but not to `testSuite/test:jsDependencies`. The problem is that `jsTestDefinitions` depends on `jsDependencies`. Maybe it is nicer if I introduce a `testtime` configuration to mitigate that (would also make the triggeredBy obsolete). WDYT?

test-suite/js/scalajs-test-suite-2.10-fastopt.html
gzm0: Currently I set these in the runner itself, but it is arguably nicer if I do it here.

test-suite/js/scalajs-test-suite-2.10-fastopt.html
sjrd: Ah I see.

A separate configuration is probably not the way to go, because it would have to be duplicated for "every" test configuration that people otherwise use (such as `it`, for example).

Let's keep it the way you did for now. Hopefully we'll get better insight when we eventually generalize to #1709.

tools/js/src/test/scala/org/scalajs/core/tools/test/js/TestRunner.scala
```diff
@@ -15,11 +14,10 @@ object TestRunner {
...
   def runTests(): Unit = {
-    val framework = new JasmineFramework()
-    val runner = framework.runner(Array("-ttypedarray"), Array(),
-        new ScalaJSClassLoader(js.Dynamic.global))
-
-    val tasks = runner.tasks(taskDefs(framework.fingerprints.head).toArray)
+    System.setProperty("scalajs.testsuite.testtag", "testtag.value")
+    System.setProperty("scalajs.nodejs", "true")
+    System.setProperty("scalajs.typedarray", "true")
+    System.setProperty("scalajs.fullopt-stage", "true")
```
gzm0: This should be fastopt-stage

gzm0:
Updated.

sjrd:
LGTM

gzm0:
I still need to test the runner manually, but the fullOpt OOMs on my machine :(

sjrd:
Do you mean the `fullOptJS` task? How much RAM do you allocate for sbt? I give it at least 3GB.

gzm0:
Yes. `testSuite/test:fullOptJS`. I run with `-Xmx4G`...

sjrd:
In 2.11.7 fastOpt (and fullOpt), I see this in the log:
```
Error while loading test class: org.scalajs.testsuite.jsinterop.?7

    at org.scalajs.testinterface.TestUtils$.loadModule(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:26432)

    at org.scalajs.junit.JUnitTask.execute(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:138857)

    at org.scalajs.junit.JUnitTask.execute(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:138842)

    at org.scalajs.testsuite.utils.HTMLRunner$$anonfun$org$scalajs$testsuite$utils$HTMLRunner$$scheduleTask$1.apply$mcV$sp(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:295167)

    at org.scalajs.testsuite.utils.HTMLRunner$$anonfun$org$scalajs$testsuite$utils$HTMLRunner$$scheduleTask$1.apply(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:295176)

    at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:230586)

    at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:230581)

    at scala.scalajs.concurrent.QueueExecutionContext$PromisesExecutionContext.execute(file:///C:/Users/Sepi/Documents/Projets/scalajs/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:272325)
```
(in red), but then the tests pass.

Comment?

sjrd:
Also, this happens both in Firefox and Chrome Canary, as well as with 2.10.6.

gzm0:
Yeah, this is probably an encoding issue. I thought it is resolved but apparently not. The problem is that the `FileWriter` in the build does not specify the encoding. Also, we cannot specify the encoding with which the JS file is read in the browser (this would normally be an HTTP header). 
So the best is probably if we do not rely on any non-ASCII characters in this file, but this in turn requires us to adapt the JSON serializer.

sjrd:
My concern is not really that there's an error (we could disable that test in browser and fix it later). My main concern is that does not fail the whole test suite. The banner at the end said "All test passed".

gzm0:
Indeed. That is a problem. Judging from the stacktrace, this lies inside our JUnit implementation:
https://github.com/scala-js/scala-js/blob/master/junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala#L41

This logs a failure but does not report it to the event handler.

sjrd:
OK so that's another bug #2333.
So then as such this PR is fine for me. Still OK for you?

gzm0:
Yeah, let's go ahead with this, so we have at least something and refine later. SG?

sjrd:
LGTM

PR #2317, opened 2016-04-02
Implement overload of java.lang.Character.toChars

kxbmap:


gzm0:
Jenkins, ok to test

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -528,15 +528,48 @@ object Character {
...
 
+  def toChars(codePoint: Int, dst: Array[Char], dstIndex: Int): Int = {
+    if (!isValidCodePoint(codePoint))
+      throw new IllegalArgumentException()
+
+    if (dst == null)
+      throw new NullPointerException()
+
+    if (dstIndex < 0)
+      throw new ArrayIndexOutOfBoundsException(dstIndex.toString)
```
sjrd: These two tests are not necessary, nor is `checkIndex`. The accesses to `dst` inside `toSurrogate` will be erroneous and (in theory) cause the same exceptions anyway.

Of course NPEs and IndexOOBs are undefined behavior, so in practice the exceptions will not be thrown as is, but then also we don't care precisely because they are undefined behavior.

So just keep the `isValidCodePoint` test. Remove all the other checks.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -77,6 +77,52 @@ class CharacterTest {
...
+    expectThrows(classOf[NullPointerException], Character.toChars(0x61, null, 0))
+
+    val e1 = expectThrows(classOf[IndexOutOfBoundsException], Character.toChars(0x61, new Array(2), -1))
+    assertEquals("-1", e1.getMessage)
+    val e2 = expectThrows(classOf[IndexOutOfBoundsException], Character.toChars(0x61, new Array(2), 2))
+    assertEquals("2", e2.getMessage)
+    val e3 = expectThrows(classOf[IndexOutOfBoundsException], Character.toChars(0x10FFFF, new Array(2), -1))
+    assertEquals("-1", e3.getMessage)
+    val e4 = expectThrows(classOf[IndexOutOfBoundsException], Character.toChars(0x10FFFF, new Array(2), 2))
+    assertEquals("3", e4.getMessage)
```
sjrd: Of course, you'll also have to remove the tests for NPE and IOOB (which are not valid in Scala.js because those kinds of exceptions are UB)

sjrd:
That's all.

kxbmap:
Thanks. Updated.

sjrd:
LGTM, Thanks!

Issue #2325, opened 2016-04-11
Upgrade to Scala 2.12.0-M4


sjrd:
Should replace 2.12.0-M3.

PR #2328, opened 2016-04-12
Fix #2309: Support LMF-capable Function nodes of 2.12.0-M4.

### LINKED ISSUE #2309: Adapt to Scala 2.12.0-M4: a Function node can have any SAM type, not just FunctionN
The community build has [revealed](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/357/consoleFull) a problematic interaction between https://github.com/scala/scala/pull/4971 and scala-js:

```
[scala-js] Running "test" in: jasmineTestFramework
[scala-js] java.lang.RuntimeException: First argument to init must be a function
[scala-js] 	at scala.sys.package$.error(package.scala:27)
[scala-js] 	at org.scalajs.jsenv.rhino.RhinoJSEnv$$anonfun$org$scalajs$jsenv$rhino$RhinoJSEnv$$setupCom$2.apply(RhinoJSEnv.scala:344)
[scala-js] 	at org.scalajs.jsenv.rhino.RhinoJSEnv$$anonfun$org$scalajs$jsenv$rhino$RhinoJSEnv$$setupCom$2.apply(RhinoJSEnv.scala:338)
[scala-js] 	at org.scalajs.jsenv.rhino.package$ScriptableObjectOps$$anon$1.call(package.scala:32)
[scala-js] 	at org.mozilla.javascript.Interpreter.interpretLoop(Interpreter.java:1473)
```

@retronym suggests this may be because Function nodes are now allowed to have any SAM type. During uncurry, we expand those that LambdaMetaFactory can't handle into an anonymous subclass as before. The others will reach the scala-js plugin.

The most recent commit in the PR is available from `"pr-scala snapshots" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/"` as `2.12.0-3ae3903-SNAPSHOT`.

We're hoping to release M4 very soon (in the next day or two) -- please let us know if you consider this a showstopper!

sjrd:
Starting from Scala 2.12.0-M4, Function nodes typed as any `LambdaMetaFactory`-capable SAM type can reach the back-end. This commit takes care of them by synthesizing an anonymous class in the back-end.

This should not be merged until #2325 is addressed, so that we have CI tests for this, although I tested locally.

There is one subtle suboptimal aspect in the current solution: in the presence of incremental compilation, when *removing* an LMF-enabled lambda, the corresponding synthetic `.sjsir` file will *not* be removed. In general, removal of obsolete .sjsir files is done by the specialized `incOptions` defined here:
https://github.com/scala-js/scala-js/blob/v0.6.8/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L117
So far it always worked because an .sjsir file always has an associated .class file. With this commit, however, this is not the case anymore. It is not possible, in general, to fix `incOptions` in this case.
This is not too much of a problem because the obsolete .sjsir files will not be referenced by other code on the classpath, and they are never exported (by construction), so the linking step will ignore them. But it is not super nice.

The only way to properly fix this would be to defer the generation of synthetic SAM classes to linking. We could introduce a new node `LambdaMetaFactory(intfType, methodIdent, bodyJSFunction)` or something like that. This could bring performance and code size improvements, too, as the linker could generated a unique wrapper class for all lambdas of a given `intfType`.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -129,6 +129,7 @@ abstract class GenJSCode extends plugins.PluginComponent
     // Per class body
     val currentClassSym = new ScopedVar[Symbol]
     private val unexpectedMutatedFields = new ScopedVar[mutable.Set[Symbol]]
+    private val generatedSAMWrapperCount = new ScopedVar[VarBox[Int]]
 
     private def currentClassType = encodeClassType(currentClassSym)
 
@@ -164,6 +165,7 @@ abstract class GenJSCode extends plugins.PluginComponent
       withScopedVars(
           currentClassSym := clsSym,
           unexpectedMutatedFields := mutable.Set.empty,
+          generatedSAMWrapperCount := null,
           currentMethodSym := null,
           thisLocalVarIdent := null,
           fakeTailJumpParamRepl := null,
@@ -181,7 +183,8 @@ abstract class GenJSCode extends plugins.PluginComponent
     // Global class generation state -------------------------------------------
 
     private val lazilyGeneratedAnonClasses = mutable.Map.empty[Symbol, ClassDef]
-    private val generatedClasses = ListBuffer.empty[(Symbol, js.ClassDef)]
+    private val generatedClasses =
+      ListBuffer.empty[(Symbol, Option[String], js.ClassDef)]
 
     private def consumeLazilyGeneratedAnonClass(sym: Symbol): ClassDef = {
       /* If we are trying to generate an method as JSFunction, we cannot
@@ -284,8 +287,9 @@ abstract class GenJSCode extends plugins.PluginComponent
 
           if (!isPrimitive && !isRawJSImplClass) {
             withScopedVars(
-                currentClassSym         := sym,
-                unexpectedMutatedFields := mutable.Set.empty
+                currentClassSym          := sym,
+                unexpectedMutatedFields  := mutable.Set.empty,
+                generatedSAMWrapperCount := new VarBox(0)
             ) {
               val tree = if (isRawJSType(sym.tpe)) {
                 assert(!isRawJSFunctionDef(sym),
@@ -302,16 +306,16 @@ abstract class GenJSCode extends plugins.PluginComponent
                 genClass(cd)
               }
 
-              generatedClasses += ((sym, tree))
+              generatedClasses += ((sym, None, tree))
             }
           }
         }
 
-        val clDefs = generatedClasses.map(_._2).toList
+        val clDefs = generatedClasses.map(_._3).toList
         generatedJSAST(clDefs)
 
-        for ((sym, tree) <- generatedClasses) {
-          genIRFile(cunit, sym, tree)
+        for ((sym, suffix, tree) <- generatedClasses) {
+          genIRFile(cunit, sym, suffix, tree)
         }
       } finally {
         lazilyGeneratedAnonClasses.clear()
@@ -587,7 +591,7 @@ abstract class GenJSCode extends plugins.PluginComponent
             staticMembers.toList)(origJsClass.optimizerHints)
       }
 
-      generatedClasses += sym -> newClassDef
+      generatedClasses += ((sym, None, newClassDef))
 
       // Construct inline class definition
       val js.MethodDef(_, _, ctorParams, _, ctorBody) = constructor.getOrElse(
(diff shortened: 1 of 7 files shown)


sjrd:
Review by @gzm0

sjrd:
Rebased.

sjrd:

sjrd:
Ping @gzm0. (not sure whether you've overlooked this or you're delaying it for when you have more time)

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4601,7 +4631,103 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      }
+    }
+
+    private def synthesizeSAMWrapper(funSym: Symbol, sam: Symbol)(
+        implicit pos: Position): String = {
+      val intfName = encodeClassFullName(funSym)
+
+      val suffix = {
+        val newCount = generatedSAMWrapperCount.value + 1
+        generatedSAMWrapperCount.value = newCount
```
gzm0: Consider:

```scala
generatedSAMWrapperCount.value += 1
"$$Lambda$" + generatedSAMWrapperCount.value
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4601,7 +4631,103 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        "$$Lambda$" + newCount // LambdaMetaFactory names classes like this
+      }
+      val generatedClassName = encodeClassFullName(currentClassSym) + suffix
```
gzm0: Use `intfName`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4601,7 +4631,103 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+      val genClassType = jstpe.ClassType(generatedClassName)
```
gzm0: Consider just `classType`. There is only one class here.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4601,7 +4631,103 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      // def this(f: Any) = { this.f$1 = f; super() }
+      val ctorDef = {
+        val fParamDef = js.ParamDef(js.Ident("f"), jstpe.AnyType,
+            mutable = false, rest = false)
+        js.MethodDef(static = false, js.Ident("init___O"), List(fParamDef),
+            jstpe.NoType,
+            js.Block(List(
+                js.Assign(
+                    js.Select(js.This()(genClassType), funFieldIdent)(jstpe.AnyType),
+                    fParamDef.ref),
```
gzm0: Consider aligning `funFieldIdent` and `fParamDef`: Use either both times just `f` or both times `fun`.

project/Build.scala
```diff
@@ -1232,15 +1232,24 @@ object Build {
...
       /* Can't add require-sam as unmanagedSourceDirectories because of the use
        * of scalacOptions. Hence sources are added individually.
        * Note that a testSuite/test will not trigger a compile when sources are
        * modified in require-sam
        */
-      if (isJSTest && scalaBinaryVersion.value != "2.10" &&
-          scalacOptions.value.contains("-Xexperimental")) {
-        val sourceDir = (sourceDirectory in Test).value / "require-sam"
-        (sourceDir ** "*.scala").get
+      if (isJSTest && supportsSAM) {
```
gzm0: This should probably be just `supportsSAM`.

gzm0:
That's all. Sorry for the delay. I indeed overlooked this.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: No, that's not the same thing. `currentClass` is the enclosing class. `intfName` is the target interface to implement.

project/Build.scala
sjrd: Er yes indeed. Good catch.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Indeed... :-/

sjrd:
Updated.

gzm0:
LGTM

PR #2330, opened 2016-04-13
Fix #2325: Build for 2.12.0-M4.

nicolasstucki:
Review by @sjrd

PR #2334, opened 2016-04-14
Multi-inline default method bridges

sjrd:
This recovers the regression in code size with 2.12.0-M4, and then more.


sjrd:
Review last commit by @nicolasstucki but do not merge. This depends on #2330.

sjrd:
Rebased.

sjrd:
retest this please

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1173,40 +1173,10 @@ private[optimizer] abstract class OptimizerCore(
...
-                    throw new AssertionError("Invalid forwarder shape: " + body)
-                }
-                if (!areAllTheSame) {
-                  // Not all doing the same thing
-                  treeNotInlined
-                } else {
-                  inline(allocationSites, Some(treceiver), targs, reference,
-                      isStat, usePreTransform)(cont)
-                }
+              if (canMultiInline(impls)) {
```
nicolasstucki: Why is this a nested `if` rather than an `else if`?

sjrd: Because the outer `if` distinguishes between 0, 1, and 2+ targets. Within the 2+ targets case, we test whether we can multi-inline them or not.

nicolasstucki:
LGTM

PR #2336, opened 2016-04-15
Fix #2335: Fix the fast path condition of BigDecimal.longValueExact.

### LINKED ISSUE #2335: run/equality.scala is broken on master since April 12
PRs that were merged on April 11: #2317 (unlikely), #2321 (unlikely) and #2314 (by elimination, most likely culprit).

sjrd:


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

sjrd:
Updated. The previous fix was also incorrect and broke other tests.

nicolasstucki:
It seams that we have issues with zero, we should check if the scale and presition are really what we where expecting with some tests.

sjrd:
They are, but in fact there is no guarantee that the `scale`'th digit after the period is non-zero. So the simple check for non-whole number that we did was not valid. I simply removed that one, as `toBigIntegerExact` itself already has a fast-path for non-whole numbers.

nicolasstucki:
Jenkins, retest this please.

nicolasstucki:
LGTM

sjrd:
retest this please

PR #2340, opened 2016-04-18
DO NOT MERGE Fix JUnit log order and non verbose mode.

nicolasstucki:
Review by @gzm0

### DIFF
--- a/junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
+++ b/junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
@@ -11,6 +11,8 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
     classMetadata: JUnitTestBootstrapper, richLogger: RichLogger,
     eventHandler: EventHandler) {
 
+  private val verbose = runner.runSettings.verbose
+
   lazy val packageName = fullyQualifiedName.split('.').init.mkString(".")
   lazy val className = fullyQualifiedName.split('.').last
 
@@ -28,9 +30,13 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
         true
     }
 
+    def logTestIgnored(name: String): Unit = {
+      if (verbose) logFormattedInfo(name, "ignored")
+      else logFormattedDebug(name, "ignored")
+    }
+
     if (assumptionViolated) {
-      if (runner.runSettings.verbose)
-        logFormattedInfo(null, "ignored")
+      logTestIgnored(null)
       taskSkipped()
     } else {
       def runWithOrWithoutQuietMode[T](block: => T): T = {
@@ -47,8 +53,7 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
         for (method <- jUnitMetadata.testMethods) {
           method.getIgnoreAnnotation match {
             case Some(ign) =>
-              if (runner.runSettings.verbose)
-                logFormattedInfo(method.name, "ignored")
+              logTestIgnored(method.name)
               ignoreTest(method.name)
 
             case None =>
@@ -72,11 +77,15 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
     val t0 = System.nanoTime
     def getTimeInSeconds(): Double = (System.nanoTime - t0).toDouble / 1000000000
 
+    def logTestStarted(name: String): Unit = {
+      if (verbose) logFormattedInfo(name, "started")
+      else logFormattedDebug(name, "started")
+    }
+
     def executeTestMethods(): Unit = {
       val expectedException = testAnnotation.expected
       try {
-        if (runner.runSettings.verbose)
-          logFormattedInfo(methodName, "started")
+        logTestStarted(methodName)
 
         classMetadata.invoke(testClassInstance, methodName)
 
@@ -151,7 +160,7 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
 
     try {
       for (method <- jUnitMetadata.beforeMethod) {
-        logFormattedInfo(methodName, "started")
+        logTestStarted(methodName)
         classMetadata.invoke(testClassInstance, method.name)
       }
       executeTestMethods()
@@ -202,6 +211,12 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
     val fMethod = if (method != null) c(method, NNAME2) else null
     richLogger.info(
         formatLayout("Test ", packageName, c(className, NNAME1), fMethod, msg))
+  }
+
+  private[this] def logFormattedDebug(method: String, msg: String): Unit = {
+    val fMethod = if (method != null) c(method, NNAME2) else null
+    richLogger.debug(
+      formatLayout("Test ", packageName, c(className, NNAME1), fMethod, msg))
   }
 
   private[this] def logFormattedWarn(prefix: String, method: String,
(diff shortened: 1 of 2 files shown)


junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
       method: JUnitMethodMetadata) = {
     val jUnitMetadata = classMetadata.metadata()
     val testClassInstance = classMetadata.newInstance()
+    val methodName = method.name
+    val testAnnotation = method.getTestAnnotation.get
 
     val t0 = System.nanoTime
-    def timeInSeconds() = (System.nanoTime - t0).toDouble / 1000000000
+    def timeInSeconds = (System.nanoTime - t0).toDouble / 1000000000
```
gzm0: Don't remove parens on side-effecting methods.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
-          false
-        } catch {
-          case ex: Throwable =>
-            executedWithExceptions(method.name, testAnnotation, timeInSeconds(), ex)
-            true
+        classMetadata.invoke(testClassInstance, methodName)
+
+        if (expectedException == classOf[org.junit.Test.None]) {
+          taskPassed(methodName)
+          executeAfterMethods()
```
gzm0: This is wrong (although seems to reproduce the old behavior). `@AfterClass` methods always need to be run: http://junit.sourceforge.net/javadoc/org/junit/AfterClass.html

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
+      } catch {
+        case ex: Throwable =>
+          if (classOf[AssumptionViolatedException].isInstance(ex) ||
+              classOf[internal.AssumptionViolatedException].isInstance(ex)) {
+            logAssertionWarning(methodName, ex, timeInSeconds)
+            taskSkipped()
+          } else if (expectedException.isInstance(ex)) {
+            taskPassed(methodName)
+            executeAfterMethods()
+          } else if (expectedException == classOf[org.junit.Test.None]) {
```
gzm0: Are you sure this is the right order? If someone throws a `org.junit.Test.None` (yes, it is throwable), it will fall in the previous case, not this one.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
+    }
 
-      if (!testMethodFailed) {
-        try {
-          for (method <- jUnitMetadata.afterMethod)
-            classMetadata.invoke(testClassInstance, method.name)
+    def executeAfterMethods(): Unit = {
+      try {
+        for (method <- jUnitMetadata.afterMethod)
+          classMetadata.invoke(testClassInstance, methodName)
```
gzm0: `method.name`?

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
-        logFormattedInfo(methodName, "started")
-      taskPassed(methodName)
-    } else {
-      val msg = {
-        s"failed: Expected exception: ${testAnnotation.expected} " +
-        s"took $timeInSeconds sec"
+    try {
+      for (method <- jUnitMetadata.beforeMethod) {
+        if (runner.runSettings.verbose)
+          logFormattedInfo(methodName, "started")
```
gzm0: Aren't you logging "started" twice now?

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,116 +64,101 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
-        else None
-      }
-      logFormattedError(methodName, msg, exOpt)
-      taskFailed(methodName)
-    } else {
-      val msg = s"failed: ${ex.getClass}, took $timeInSeconds sec"
-      logFormattedError(methodName, msg, Some(ex))
-      taskFailed(methodName)
+      case ex: internal.AssumptionViolatedException =>
+        logAssertionWarning(methodName, ex, timeInSeconds)
```
gzm0: What about other exceptions? (I realize this is not changing the existing behavior, but it seems odd, since it can cause `executeTests` to throw).

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -199,8 +199,12 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
   private[this] def logFormattedInfo(method: String, msg: String): Unit = {
     val fMethod = if (method != null) c(method, NNAME2) else null
-    richLogger.info(
-        formatLayout("Test ", packageName, c(className, NNAME1), fMethod, msg))
+    val formatted = formatLayout("Test ", packageName, c(className, NNAME1),
+        fMethod, msg)
+    if (verbose)
+      richLogger.info(formatted)
+    else
+      richLogger.debug(formatted)
```
gzm0: Hum... That now makes everything that goes to info go to debug if `-v` is not turned on. Are you sure this is intended? The documentation of `-v` says:

> Log "test run started" / "test started" / "test run finished" events on log level "info" instead of "debug".

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
nicolasstucki: `org.junit.Test.None` is a class that can't be instantiated by the user [Test.scala#L21](https://github.com/scala-js/scala-js/blob/master/junit-runtime/src/main/scala/org/junit/Test.scala#L21) and in fact is never instantiated be the framework either. As such there should never be a case where `expectedException.isInstance(ex: Test.None)` and the order is irrelevant.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
nicolasstucki: Cleary ;)
And another one bellow.

gzm0:
It seems there are many degrees of freedom here. Honestly I think this code is too complex not to be guarded by tests. Do we have any tests of this code?

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
nicolasstucki: Actually it is only for the `started` and `ignored` ones. I should change the name of the function or add another level of indirection. The implementation follows the original implementation rather than the documentation to have a similar output (modulo stack traces).

nicolasstucki:
We do not have any test for this. We should figure out a reliable way of creating tests on the output that can be run on `JS` and `JVM`.

gzm0:
So I guess what I'm saying is that there is no point fixing nifty details like this, if we don't test for it. Mainly because the aggregate of:

- How it works
- How it is supposed to work

Is too complex for a developer to just grasp when something is changed.

PR #2346, opened 2016-04-20
Fix #2345: Bring hash implem in Statics up to date wrt BoxesRunTime.

### LINKED ISSUE #2345: Different hashCode in JS and JVM for case class with big Long
I have a case class with the long value. If this long is less than `Int.MaxValue` I have the same results of `hashCode` invocation in the JVM and in the JS runtimes.
But if this long is bigger than `Int.MaxValue`, I have different results in different runtimes

Minimal example:
```scala
case class A(long: Long)

val big = 1461126709984L
val small = 1L

val bigClass = new A(big)
val smallClass = new A(small)

println(big.hashCode())
println(bigClass.hashCode())
println(small.hashCode())
println(smallClass.hashCode())
```
Results for JVM:
```
837829556
259268522
1
818387364
```
Results for JS:
```
837829556
509963985
1
818387364
```
Only `bigClass.hashCode()` is different.

sjrd:
The methods in `Statics` are used by the code generator for the `hashCode()` of case classes. They must be in sync with the implementations in `BoxesRunTime`.

The code duplication should be treated at the source in scala/scala.

### DIFF
--- a/library-aux/src/main/scala/scala/runtime/Statics.scala
+++ b/library-aux/src/main/scala/scala/runtime/Statics.scala
@@ -34,56 +34,46 @@ object Statics {
   }
 
   def longHash(lv: Long): Int = {
-    lv.toInt
-    /*
-    val iv = lv.toInt | 0
-    if (iv == lv) iv
-    else (lv ^ (lv >>> 32)).toInt | 0
-    */
+    val lo = lv.toInt
+    val hi = (lv >>> 32).toInt
+    if (hi == (lo >> 31)) lo // it is in the Int range
+    else lo ^ hi
   }
 
   def doubleHash(dv: Double): Int = {
-    dv.toInt
-    /*
-    val iv = dv.toInt | 0
-    if (iv == dv)
-      return iv
-
-    val fv = dv.toFloat
-    if (fv == dv)
-      return java.lang.Float.floatToIntBits(fv)
-
-    val lv = dv.toLong
-    if (lv == dv)
-      return lv.toInt | 0
-
-    val lv2 == java.lang.Double.doubleToLongBits(dv)
-    return (lv2 ^ (lv2 >>> 32)).toInt | 0
-    */
+    /* This implementation is based on what 2.12.0-M5+ does on the JVM.
+     * The 2.10/2.11 implementation on the JVM was not consistent with that of
+     * BoxesRunTime, and most importantly was not consistent with the hash of
+     * Long values.
+     *
+     * In Scala.js, we always use the version consistent with BoxesRunTime.
+     * Note that, for values that happen to be valid floats but not valid
+     * longs, this implementation is *not* consistent with the JVM (just like
+     * that of BoxesRunTime).
+     */
+    val iv = dv.toInt
+    if (iv == dv) {
+      iv
+    } else {
+      // don't test the case dv.toFloat == dv
+      val lv = dv.toLong
+      if (lv == dv)
+        lv.hashCode()
+      else
+        dv.hashCode()
+    }
   }
 
   def floatHash(fv: Float): Int = {
-    fv.toInt
-    /*
-    val iv = fv.toInt
-    if (iv == fv)
-      return iv
-
-    val lv = fv.toLong
-    if (lv == fv)
-      return (lv ^ (lv >>> 32)).toInt | 0
-
-    return java.lang.Float.floatToIntBits(fv)
-    */
+    doubleHash(fv.toDouble)
   }
 
   def anyHash(x: Any): Int = {
     x match {
-      case null => 0
-      case x: Long => longHash(x)
+      case null      => 0
       case x: Double => doubleHash(x)
-      case x: Float => floatHash(x)
-      case _ => x.hashCode()
+      case x: Long   => longHash(x)
+      case _         => x.hashCode()
     }
   }
 }
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -115,6 +115,7 @@ abstract class GenJSCode extends plugins.PluginComponent
   override def newPhase(p: Phase): StdPhase = new JSCodePhase(p)
 
   private object jsnme { // scalastyle:ignore
+    val anyHash = newTermName("anyHash")
     val arg_outer = newTermName("arg$outer")
     val newString = newTermName("newString")
   }
@@ -3082,13 +3083,19 @@ abstract class GenJSCode extends plugins.PluginComponent
       js.BinaryOp(js.BinaryOp.String_+, lhs, rhs)
     }
 
-    /** Gen JS code for a call to Any.## */
+    /** Gen JS code for a call to `Any.##`.
+     *
+     *  This method unconditionally generates a call to `Statics.anyHash`.
+     *  On the JVM, `anyHash` is only called as of 2.12.0-M5. Previous versions
+     *  emitted a call to `ScalaRunTime.hash`. However, since our `anyHash`
+     *  is always consistent with `ScalaRunTime.hash`, we always use it.
+     */
     private def genScalaHash(tree: Apply, receiver: Tree): js.Tree = {
       implicit val pos = tree.pos
 
-      val instance = genLoadModule(ScalaRunTimeModule)
+      val instance = genLoadModule(RuntimeStaticsModule)
       val arguments = List(genExpr(receiver))
-      val sym = getMember(ScalaRunTimeModule, nme.hash_)
+      val sym = getMember(RuntimeStaticsModule, jsnme.anyHash)
 
       genApplyMethod(instance, sym, arguments)
     }
--- a/scalalib/overrides/scala/runtime/BoxesRunTime.scala
+++ b/scalalib/overrides/scala/runtime/BoxesRunTime.scala
@@ -81,45 +81,28 @@ object BoxesRunTime {
     }
   }
 
-  def hashFromLong(n: java.lang.Long): Int = {
-    val iv = n.intValue()
-    if (iv == n.longValue()) iv
-    else n.hashCode()
-  }
+  @inline
+  def hashFromLong(n: java.lang.Long): Int =
+    Statics.longHash(n.asInstanceOf[Long])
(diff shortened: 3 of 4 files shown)


sjrd:
Review by @gzm0.

library-aux/src/main/scala/scala/runtime/Statics.scala
```diff
@@ -34,56 +34,35 @@ object Statics {
...
-    return (lv2 ^ (lv2 >>> 32)).toInt | 0
-    */
+    val iv = dv.toInt
+    if (iv == dv) {
+      iv
+    } else {
+      // don't test the case dv.toFloat == dv
+      val lv = dv.toLong
+      if (lv == dv)
+        lv.hashCode()
```
gzm0: This seems to deviate from the implementation in scala/scala if `lv > Int.MaxValue`. Is this intentional?

library-aux/src/main/scala/scala/runtime/Statics.scala
```diff
@@ -34,56 +34,35 @@ object Statics {
...
 
   def longHash(lv: Long): Int = {
-    lv.toInt
-    /*
-    val iv = lv.toInt | 0
+    val iv = lv.toInt
     if (iv == lv) iv
-    else (lv ^ (lv >>> 32)).toInt | 0
-    */
+    else iv ^ (lv >>> 32).toInt
```
gzm0: `lv.hashCode()`?

gzm0:
That's all.

library-aux/src/main/scala/scala/runtime/Statics.scala
sjrd: Are you talking about `(lv2 ^ (lv2 >>> 32)).toInt`? That's exactly what `lv.hashCode()` returns.

library-aux/src/main/scala/scala/runtime/Statics.scala
sjrd: I was trying to avoid re fetching `lo` from memory. But maybe that's a tiny bit overkill ^^

sjrd: In fact I realized there's a better implementation of this method, that does not involve allocating a new `RuntimeLong`:
```scala
val lo = lv.toInt
val hi = (lv >>> 32).toInt
if (hi == (lo >> 31)) lo
else lo ^ hi
```

library-aux/src/main/scala/scala/runtime/Statics.scala
sjrd: Oh no. I see now that the JVM impl does `lv.toInt` in that case. I actually believe that's a bug in the JVM version ^^ Because it wouldn't return something consistent with `longHash`.

I'll have to try and observe the bug tomorrow.

sjrd: What I implemented here is supposed to be consistent with [our `BoxesRunTime` implementation](https://github.com/scala-js/scala-js/blob/5c622745f20841b91d37aac567eec886a722e976/scalalib/overrides/scala/runtime/BoxesRunTime.scala#L90-L104), which, itself, is supposed to be consistent with [the JVM one](https://github.com/scala/scala/blob/6090f53ee9ab2677f732ea5cc7144cb9684b2593/src/library/scala/runtime/BoxesRunTime.java#L233-L244).

And here, I think the JVM implementation in `Statics` is not consistent with theirs in `BoxesRunTime`.

gzm0: In that case, can you make a test that makes this apparent? (And make it JS only, or make Scala.js bug compatible).

library-aux/src/main/scala/scala/runtime/Statics.scala
gzm0: Fair enough :)

library-aux/src/main/scala/scala/runtime/Statics.scala
sjrd: I confirmed there was an inconsistency on the JVM in https://github.com/scala/scala/pull/5110
I suggest we wait for that PR to be discussed before proceeding here.

gzm0: SGTM.

scottcarey:
````
val lo = lv.toInt
val hi = (lv >>> 32).toInt
if (hi == (lo >> 31)) lo
else lo ^ hi
````

there is an if statement in there, lets get rid of it.  branching is expensive! 

So, the reason the hash is not just:
````
(lv ^ (lv >>> 32))
````

is because it has to be consistent with the integer hash when they equal.  YUCK.

The implementation above that returns `lo ^ hi` also does not change when the 32nd bit flips (e.g. the hash is the same if the number is 2^40 + 2^32 or if it is just 2^40 -- in both cases `lo` is 0 and `hi` is 2^8.   Not using all 64 bits in the hash is bad.

Is the only requirement to make it consistent with the hash for int when they are equal value?  We can fix the above and also get rid of the conditional, if we are allowed to change the output.

So, the hash must be `lv.toInt` whenever the magnitude of the long is not larger than an int.  this happens whenever the top 33 bits are all 0's (for positive numbers) or all 1's (for negative numbers).

So we're going to use `lo` no matter what.  The question is what `hi` should look like.  The 64th bit is already present in `lo`, because we used `toInt`.  So we need to deal with bits 32 to 63. 

If the top 33 bits are all 0's, or all 1's, the long is equivalent to the int produced by `toInt`.   To unconditionally xor, we have to make all those 1's zeroes in this case only.

This operation does that:

````
((lv >>> 31) & 0xFFFFFFFF) + (lo >>> 31) // the mask is unsigned truncation, with toInt we lose a bit
````

Note the original `lv`'s 64th bit (sign bit) is now the 32nd bit on `lo`.

If the 64th bit is 1, and bits 32 to 63 are all 1's, this becomes all zeroes as the +1 rolls it over to zero.
If the 64th bit is 0, and bits 32 to 63 are all 0's, this remains all zeroes.

We keep all bits in the hash this way as well.  The only collision introduced is exactly when the long value is no larger than an int.

I propose:
````
val lo = lv.toInt // contains  original bits:  64 (in position 32) and 31 to 1
val hi = ((lv >>> 31) & 0xFFFFFFFF) // contains original bits 63 to 32
val hi2 = hi + (lo >>> 31) // add 1 if a negative number, if 'hi' is all 1's they become 0's
lo ^ hi2
````

**Comment by sjrd** [comment]
> Is the only requirement to make it consistent with the hash for int when they are equal value?

*Upstream*, yes. But in Scala.js we have the requirement to stay consistent with whatever hash is computed on the JVM. You could suggest your proposal to scala/scala, and if accepted there, we could then import it.

But we cannot have the liberty to use a different algorithm if it does not return the same hash as the original one for all longs.

I had thought of another, simpler branch-less algorithm earlier, but I have to reject it for the same reason. Here it is, FTR.
From the original:
```scala
if (hi == (lo >> 31)) lo
else lo ^ hi
```
we can rewrite as:
```scala
val magic = if (hi == (lo >> 31)) 0 else hi
lo ^ magic
```
Now, the only branch we really care about is the `0`. We can change the `hi` if we want. We can thus rewrite as:
```scala
val magic = hi ^ (lo >> 31)
lo ^ magic
```
because `magic` will be 0 when `hi == (lo >> 31)`. We can then inline `magic` and end with:
```scala
lo ^ hi ^ (lo >> 31)
```
As the original algorithm, it disregards bit n°31, though. But it contains fewer operations and, *most importantly* no operation on `Long`s, which are extremely expensive in Scala.js (much more than a branch); yours contains the `>>> 31` and the `& 0xFFFFFFFF` before we can cast to int.

scottcarey:
Yeah, the JVM sees "((lv >>> 31) & 0xFFFFFFFF)" and goes "oh! unsigned cast to int!" so it doesn't even tell the CPU to mask, it just truncates, emitting an unsigned cast operator to the cpu rather than computing a mask on 64 bit values then casting.  In short, this is one very fast CPU instruction on the JVM, javascript probably sees it as a 64 bit mask and then a truncation to int.

Your algorithm ends up being the same as mine minus a small difference.

Because of the identity `~x + 1 = (-1) * x`, adding 1 and flipping all bits are the same thing in the case where its all 1's, the one case we care about here.

You are doing `x ^ (lo >> 31)` which is the same as  `if (lo >>> 31 == 1) ~x else x`.

By the identity above `(-1) * ~x = x  + 1`.  I'm doing `x + 1` and you are doing `~x`, the difference is multiplying by -1 -- which in the case of all 1's, is still all 1's.  in all other cases our bits are flipped for 'hi'.

Well other than yours dropping a bit.  But that can be fixed, I think:

```` scala
val lo = lv.toInt
val hi = (lv >>> 31).toInt // the same as ((lv >>> 31) & 0xFFFFFFFF)
lo ^ (hi ^ (lo >> 31))
````

Now the only difference between yours and mine is `hi ^ (lo >> 31)` vs `hi + (lo >>> 1)`

Both are identical when `(lo >>> 1) == 0`.  Neither are consistent with java.util.Long.hashCode

scottcarey:
`val hi = (lv >>> 31).toInt // the same as ((lv >>> 31) & 0xFFFFFFFF)`

actually, not the same.  We lose one bit using toInt.

scottcarey:
actually, not the same. We lose one bit using toInt.

I was wrong, we do not lose a bit here, this is simple truncation to the lower 32 bits.  Thus, my versions above are all flawed.  The two variations discussed (corrected) are:

````java
int lo = (int) lv;
int hi = (int)(lv >>> 32);
lo ^ (hi + (lo >>> 31));
````


````java
int lo = (int) lv;
int hi = (int)(lv >>> 32);
lo ^ (hi ^ (lo >> 31));
````

I suspect we can make one of these two consistent with the current implementation.

library-aux/src/main/scala/scala/runtime/Statics.scala
sjrd: The PR was merged in Scala. I suggest we anticipatively use this new version even in 2.10 and 2.11 in `Statics`, since it only affects `Double`, and hash code for `Double`s are not preserved wrt to the JVM anyway.

sjrd:
Updated.

gzm0:
LGTM

PR #2351, opened 2016-04-23
Type ES6 Promises

julienrf:
This is an attempt in typing ES6 Promises so that we don’t get static types that are inconsistent with the actual runtime types.

The PR is still missing some documentation, I’d like to first get your feedback on this work. As you can see in the diff it simplifies or removes some type ascriptions, but, on the other hand, the type signatures of `Promise`’s methods are harder to understand (due to the `CanBuildFrom`-like pattern).

The changes make `scala_concurrent_FutureToJSPromise_basic_case` and `scala_concurrent_FutureToJSPromise_thenable_case` tests fail with the following error:

~~~
failed: An undefined behavior was detected: undefined is not an instance of scala.scalajs.js.Thenable$Returning
~~~

I don’t understand why we get such an `undefined` value. Any idea?

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Promise.scala
+++ b/library/src/main/scala/scala/scalajs/js/Promise.scala
@@ -37,26 +37,35 @@ import scala.concurrent.Future
  *  directly use the methods of `Future` on `Promise`s.
  */
 @js.native
-class Promise[+A](
+class Promise[+A] private[js] (
     executor: js.Function2[js.Function1[A | Thenable[A], _], js.Function1[scala.Any, _], _])
     extends js.Object with js.Thenable[A] {
 
-  def `then`[B](
-      onFulfilled: js.Function1[A, B | Thenable[B]],
-      onRejected: js.UndefOr[js.Function1[scala.Any, B | Thenable[B]]] = js.undefined): Thenable[B] = js.native
+  def `then`[B, C](
+      onFulfilled: js.Function1[A, B],
+      onRejected: js.UndefOr[js.Function1[scala.Any, B]] = js.undefined
+  )(implicit
+      ev: Thenable.Returning[B, C]
+  ): Thenable[C] = js.native
 
-  def `then`[B >: A](
+  def `then`[B >: A, C](
       onFulfilled: Unit,
-      onRejected: js.UndefOr[js.Function1[scala.Any, B | Thenable[B]]]): Thenable[B] = js.native
+      onRejected: js.UndefOr[js.Function1[scala.Any, B]]
+  )(implicit
+      ev: Thenable.Returning[B, C]
+  ): Thenable[C] = js.native
 
-  def `catch`[B >: A](
-      onRejected: js.UndefOr[js.Function1[scala.Any, B | Thenable[B]]] = js.undefined): Promise[B] = js.native
+  def `catch`[B >: A, C](
+      onRejected: js.UndefOr[js.Function1[scala.Any, B]] = js.undefined
+  )(implicit
+    ev: Thenable.Returning[B, C]
+  ): Promise[C] = js.native
 }
 
 @js.native
 object Promise extends js.Object {
   /** Returns a new [[Promise]] completed with the specified `value`. */
-  def resolve[A](value: A | Thenable[A]): Promise[A] = js.native
+  def resolve[A, B](value: A)(implicit ev: Thenable.Returning[A, B]): Promise[B] = js.native
 
   /** Returns a new [[Promise]] failed with the specified `reason`. */
   def reject(reason: scala.Any): Promise[Nothing] = js.native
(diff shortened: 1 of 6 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/AsyncTest.scala
```diff
@@ -221,15 +221,47 @@ class AsyncTest {
...
+
+      val promiseInt =
+        js.JSPromise[Int]((resolve: js.Function1[Int, _], error: js.Function1[Any, _]) => resolve(42))
+      typed[js.Thenable[Int]](promiseInt)
+
+      val promiseFromThenable =
+        js.JSPromise[Int](
+          (resolve: js.Function1[js.Thenable[Int], _], error: js.Function1[Any, _]) => resolve(promiseInt)
+        )
+      typed[js.Promise[Int]](promiseFromThenable)
```
julienrf: Resolving a `Promise` with a `Thenable[Int]` gives a `Promise[Int]`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/AsyncTest.scala
```diff
@@ -221,15 +221,47 @@ class AsyncTest {
...
+
+      typed[js.Thenable[Int]](promiseInt.`then`((x: Int) => x + 1))
+
+      typed[js.Thenable[Int]](promiseInt.`then`((x: Int) => promiseInt))
```
julienrf: Returning a `Thenable[Int]` in the function supplied to `then` gives a `Thenable[Int]`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/AsyncTest.scala
```diff
@@ -221,15 +221,47 @@ class AsyncTest {
...
+
+      typed[js.Thenable[String | Int]] {
+        promiseInt.`then`((x: Int) => "foo": String | js.Promise[Int])
+      }
+
+      typed[js.Thenable[String | Int]] {
+        promiseInt.`then`((x: Int) => 0: js.Promise[String] | Int)
+      }
```
julienrf: Same here with or types.

sjrd:

library/src/main/scala/scala/scalajs/js/Promise.scala
```diff
@@ -37,26 +37,35 @@ import scala.concurrent.Future
...
     extends js.Object with js.Thenable[A] {
 
-  def `then`[B](
-      onFulfilled: js.Function1[A, B | Thenable[B]],
-      onRejected: js.UndefOr[js.Function1[scala.Any, B | Thenable[B]]] = js.undefined): Thenable[B] = js.native
+  def `then`[B, C](
+      onFulfilled: js.Function1[A, B],
+      onRejected: js.UndefOr[js.Function1[scala.Any, B]] = js.undefined
+  )(implicit
+      ev: Thenable.Returning[B, C]
```
sjrd: You can't do this. An actual value of `Returning` will be given as a third parameter when calling `then`. For some `Thenable`s, this could have an unwanted effect (if they use the third param for options or something).

library/src/main/scala/scala/scalajs/js/Promise.scala
```diff
@@ -37,26 +37,35 @@ import scala.concurrent.Future
  *  directly use the methods of `Future` on `Promise`s.
  */
 @js.native
-class Promise[+A](
+class Promise[+A] private[js] (
```
sjrd: You can't hide the constructor. It must be accessible for subclasses. (because yes, `Promise` is subclassable in ES)

library/src/main/scala/scala/scalajs/js/Thenable.scala
```diff
@@ -80,4 +84,23 @@ object Thenable {
...
+  sealed trait Returning[A, B]
+
+  trait Returning1 {
+    implicit def default[A]: Returning[A, A] = new Returning[A, A] {}
+  }
+
+  object Returning extends Returning1 {
+
+    implicit def flattened[A, T[X] <: Thenable[X]]: Returning[T[A], A] =
+      new Returning[T[A], A] {}
```
sjrd: I'm pretty sure this can still be broken. Consider:
```scala
def mapP[A, B](p: js.Promise[A], f: A => B): js.Promise[B] =
  p.`then`(x => f(x))

val p = js.Promise(5)
val p2: js.Promise[js.Promise[Int]] = mapP(p, x => js.Promise(x))
```

sjrd:
Also note that, since this a public API, any source or binary incompatible change will have to wait for the next major version; and if source incompatible, we'll have to find some way to provide a migration path. So unless you manage to come up with something compatible, do not expect this to be merged anytime soon.

We're *really* serious about compatibility in Scala.js.

library/src/main/scala/scala/scalajs/js/Thenable.scala
julienrf: Your implementation of `mapP` does not compile: you have to provide an evidence that resolving a `B` effectively returns a `js.Promise[B]`. You can do that by adding an implicit parameter of type `Thenable.Returning[B, C]`, and it will fix the return type of `mapP` to `js.Promise[C]`, not `js.Promise[B]`.

In other words, and that’s the point of this PR, you can not implement `Functor[js.Promise]` because `js.Promise` breaks parametricity.

julienrf:
Thanks for your feedback @sjrd!

I think that the fact that this PR adds an additional parameter to the `then` method and makes the `js.Promise` constructor private are indeed serious issues… So maybe this work is not worth it.

PR #2352, opened 2016-04-24
Fix #2350: Add an sbt setting isScalaJSProject.

### LINKED ISSUE #2350: Provide a way to determine if an sbt Project is a Scala.js project
It would be convenient if it were possible to determine if an sbt Project is a Scala.js project.

Perhaps something like this (which I have for now in user-land):

```scala
import sbt._
import sbt.Keys._

object IsScalaJsProjectAutoPlugin extends AutoPlugin {
  override def requires = plugins.JvmPlugin
  override def trigger = allRequirements

  object autoImport {
    val isScalaJsProject = settingKey[Boolean]("Determines if a project is a Scala.js project")
  }
  import autoImport._

  override def globalSettings: Seq[Setting[_]] = Seq(isScalaJsProject := false)

  override def projectSettings: Seq[Setting[_]] = Seq(
    isScalaJsProject := libraryDependencies.value.exists(_.name == "scalajs-library")
  )
}
```

sjrd:
This sbt setting is true if and only if the current project is a Scala.js project. It is not meant to be written to by users (only to be read).

### DIFF
--- a/sbt-plugin-test/build.sbt
+++ b/sbt-plugin-test/build.sbt
@@ -113,13 +113,25 @@ lazy val multiTest = crossProject.
     // Make FrameworkDetector resilient to other output - #1572
     jsDependencies in Test += ProvidedJS / "consoleWriter.js",
     testOptions += Tests.Argument(
-      TestFramework("com.novocode.junit.JUnitFramework"), "-v", "-a")
+      TestFramework("com.novocode.junit.JUnitFramework"), "-v", "-a"),
+
+    // Test isScalaJSProject (as a setting, it's evaluated when loading the build)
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

dwijnand:
"It is not meant to be written to by users (only to be read)." I would love if there were a way to define immutable definitions in sbt..

Is `isScalaJSProject` false or undefined for projects that don't enable `ScalaJSPlugin`?

sjrd:
> Is isScalaJSProject false or undefined for projects that don't enable ScalaJSPlugin?

It is `false`.

PR #2353, opened 2016-04-25
Fix #2333: Add Failure event for JUnit plugin failures.

### LINKED ISSUE #2333: JUnit runner does not report failures as events
See https://github.com/scala-js/scala-js/pull/2316#issuecomment-209833657

nicolasstucki:


nicolasstucki:
Review by @sjrd

gzm0:
As far as I am concerned, the same comment as to #2340 applies here: As long as we don't test (and therefore define) the expected behavior, changes/fixes to this logic seem pretty meaningless...

sjrd:
@gzm0 In an ideal world, true. However in practice, this is hard. @nicolasstucki has been working on testing these things, and the work so far demonstrates that this will take significant effort. Meanwhile, we have a completely broken framework which is worth fixing, even if only manual testing can verify the fix at this point.

Maybe the fixes in #2340 are not critical enough to apply this reasoning, but this one IMO it is.

gzm0:
> Maybe the fixes in #2340 are not critical enough to apply this reasoning, but for this one IMO it is.

Point taken. This one is indeed too important.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
```diff
@@ -17,29 +17,38 @@ final class JUnitTask(val taskDef: TaskDef, runner: JUnitBaseRunner)
...
-            taskDef.fullyQualifiedName + ", expected " + bootstrapperName +
-            " to extend JUnitTestBootstrapper")
+        errorWhileLoadingClass(
+            s"expected $bootstrapperName to extend JUnitTestBootstrapper")
 
       case Failure(exception) =>
-        richLogger.error("Error while loading test class: " +
-            taskDef.fullyQualifiedName, exception)
+        errorWhileLoadingClass(exception.getMessage)
+        exception.printStackTrace()
```
gzm0: You should log the stack trace to the logger.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
```diff
@@ -17,29 +17,38 @@ final class JUnitTask(val taskDef: TaskDef, runner: JUnitBaseRunner)
...
       case Success(classMetadata: JUnitTestBootstrapper) =>
         new JUnitExecuteTest(taskDef, runner, classMetadata,
             richLogger, eventHandler).executeTests()
 
       case Success(_) =>
-        richLogger.error("Error while loading test class: " +
-            taskDef.fullyQualifiedName + ", expected " + bootstrapperName +
-            " to extend JUnitTestBootstrapper")
+        errorWhileLoadingClass(
+            s"expected $bootstrapperName to extend JUnitTestBootstrapper")
```
gzm0: Consider also reporting the class of what you got. If this case ever occurs, it is probably because of a name clash. In that case reporting the class might be useful.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
```diff
@@ -17,29 +17,38 @@ final class JUnitTask(val taskDef: TaskDef, runner: JUnitBaseRunner)
...
     if (runner.runSettings.verbose)
       richLogger.info(c("Test run started", INFO))
 
-    val bootstrapperName = taskDef.fullyQualifiedName + "$scalajs$junit$bootstrapper"
+    val bootstrapperName = fullClassName + "$scalajs$junit$bootstrapper"
 
     val startTime = System.nanoTime
 
+    def errorWhileLoadingClass(msg: String): Unit = {
+      val msg = {
```
sjrd: Double definition of `msg` (it's also the param of the `def`)

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
gzm0: And you should report the `Throwable` to the event handler.

sjrd:
That's all.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
sjrd: Hum ... That's `$bootstrapperName`, by construction.
Unless you mean the complete list of ancestors of `$bootstrapperName`, but that is not available in Scala.js (so far, as `Class.getInterfaces` is not implemented).

gzm0: ... indeed :P

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
nicolasstucki: Oh, last minute refactoring gone wrong :(

junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
```diff
@@ -17,29 +17,33 @@ final class JUnitTask(val taskDef: TaskDef, runner: JUnitBaseRunner)
...
 
-    val bootstrapperName = taskDef.fullyQualifiedName + "$scalajs$junit$bootstrapper"
+    val bootstrapperName = fullClassName + "$scalajs$junit$bootstrapper"
 
     val startTime = System.nanoTime
 
+    def errorWhileLoadingClass(t: Throwable): Unit = {
+      richLogger.error("Error while loading test class: " + fullClassName, t)
+      val selector = new TestSelector(fullClassName)
+      eventHandler.handle(new JUnitEvent(taskDef, Status.Failure, selector))
```
sjrd: IIUC @gzm0's earlier comment, you should give `t` as an additional parameter to `JUnitEvent` so that the event also has a reference to the throwable.

sjrd:
That's all.

sjrd:
LGTM

PR #2356, opened 2016-04-25
Fix #2272: Write unit tests for the IR printer.

### LINKED ISSUE #2272: Write unit tests for the IR printers
See https://github.com/scala-js/scala-js/pull/2258#issuecomment-189242905

sjrd:
This includes a couple of tweaks to what is actually printed, to make things more beautiful.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -181,16 +181,18 @@
 
   <task id="bootstrap"><![CDATA[
     setJavaVersion $java
-    sbt 'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
-        ++$scala toolsJS/test &&
-    sbt 'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
+    sbt 'set scalaJSUseRhino in Global := false' \
+        'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
+        ++$scala irJS/test toolsJS/test &&
+    sbt 'set scalaJSUseRhino in Global := false' \
+        'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
         'set scalaJSStage in Global := FullOptStage' \
-        ++$scala toolsJS/test
+        ++$scala irJS/test toolsJS/test
   ]]></task>
 
   <task id="tools-cli-stubs"><![CDATA[
     setJavaVersion $java
-    sbt ++$scala tools/package tools/test cli/package cli/assembly \
+    sbt ++$scala tools/package ir/test tools/test cli/package cli/assembly \
         stubs/package jsEnvs/test \
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
         jsEnvs/mimaReportBinaryIssues testAdapter/mimaReportBinaryIssues \
@@ -202,7 +204,7 @@
 
   <task id="tools-cli-stubs-sbtplugin"><![CDATA[
     setJavaVersion $java
-    sbt ++$scala tools/package tools/test cli/package cli/assembly \
+    sbt ++$scala tools/package ir/test tools/test cli/package cli/assembly \
         stubs/package jsEnvs/test \
         sbtPlugin/package \
         ir/mimaReportBinaryIssues tools/mimaReportBinaryIssues \
--- a/ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
+++ b/ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
@@ -1,0 +1,1008 @@
+package org.scalajs.core.ir
+
+import scala.language.implicitConversions
+
+import org.junit.Test
+import org.junit.Assert._
+
+import Definitions._
+import Printers._
+import Trees._
+import Types._
+
+import OptimizerHints.{empty => NoOptHints}
+
+class PrintersTest {
+  private implicit val dummyPos = Position.NoPosition
+
+  private def assertPrintEquals(expected: String, tree: Tree): Unit =
+    assertPrintEqualsImpl(expected, _.print(tree))
+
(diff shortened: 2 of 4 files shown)


sjrd:
I guess @gzm0 will want to review this :p

gzm0:
Thanks for doing this! I'll try and review this tonight.

ci/matrix.xml
```diff
@@ -181,16 +181,18 @@
 
   <task id="bootstrap"><![CDATA[
     setJavaVersion $java
-    sbt 'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
-        ++$scala toolsJS/test &&
-    sbt 'set jsEnv in toolsJS := NodeJSEnv(args = Seq("--max_old_space_size=2048")).value.withSourceMap(false)' \
+    sbt 'set scalaJSUseRhino in Global := false' \
```
gzm0: Is this really necessary, or do you just want it for completeness?

sjrd: It's necessary (as the previous failure showed).
That's because `irJS` needs `zip.js` which does not boot in Rhino.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -0,0 +1,999 @@
...
+
+  @Test def printEmpty(): Unit = {
+    assertPrintEquals("<empty>", EmptyTree)
+  }
+
+  @Test def printVarDef(): Unit = {
+    assertPrintEquals("val x: int = 5",
+        VarDef("x", IntType, mutable = false, i(5)))
+    assertPrintEquals("var x: int = 5",
+        VarDef("x", IntType, mutable = true, i(5)))
```
gzm0: Please test the case of an empty rhs (we shouldn't print the equality sign in that case).

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -0,0 +1,999 @@
...
+          |      10
+          |    };
+          |  default:
+          |    11;
+          |}
+        """,
+        Match(ref("x", IntType), List(
+            List(i(5)) -> i(6),
+            List(i(7), i(8)) -> Block(i(9), i(10))),
+            i(11))(IntType))
```
gzm0: Please also check with a missing default case.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
sjrd: OK, even though, in fact, a `VarDef` with en empty rhs is never valid IR.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -0,0 +1,999 @@
...
+  }
+
+  @Test def printCallHelper(): Unit = {
+    assertPrintEquals("help(x, y)",
+        CallHelper("help", List(ref("x", AnyType), ref("y", AnyType)))(IntType))
+  }
+
+  @Test def printJSNew(): Unit = {
+    assertPrintEquals("new C()", JSNew(ref("C", AnyType), Nil))
+    assertPrintEquals("new C(4, 5)", JSNew(ref("C", AnyType), List(i(4), i(5))))
```
gzm0: Could you also add one that exercises the nesting of `containsOnlySelectsFromAtom`?

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -0,0 +1,999 @@
...
+            IntType, i(5))(NoOptHints, None))
+  }
+
+  @Test def printPropertyDef(): Unit = {
+    assertPrintEquals(
+        """<property: "prop">""",
+        PropertyDef(StringLiteral("prop"),
+            i(5),
+            ParamDef("x", AnyType, mutable = false, rest = false),
+            EmptyTree))
```
gzm0: Hahahaha, `Printers.scala` still reads: `// TODO` here :) Filed as #2357

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
gzm0: Note that I think there is actually a bug there: The printer checks for `EmptyTree`, but `EmptyTree` in the default value is not (and should not be allowed). Just like in an `if`, it should simply be `Skip`.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
gzm0: Oh, really? In that case you can remove the special case from the printing code I guess.

gzm0:
That's all. Thanks!

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
sjrd: OK yes it's better.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
sjrd: Indeed, it is not valid IR not to have a default case. I'll also remove the `if` in the printer.

gzm0: Should we have a special case in the printer if the default case is a skip?

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
sjrd: OK

sjrd:
Updated.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
sjrd: No. There are other cases where we do not specialize `skip`s, e.g., the body of `while` loops. This should not happen often, so we don't bother.

The cases we do specialize are things that happen often in typical Scala code. An empty default is very rare.

gzm0:
LGTM. Thanks!

PR #2362, opened 2016-04-26
Fix #2358: Normalize the dir separator to '/' in relative paths.

### LINKED ISSUE #2358: ProvidedJS requires OS dependent path separator
In order to include `src/main/resources/js/dependency.js` on a build that is ran in Windows, we have to specify `ProvidedJS / "js\\dependency.js"`. 

`ProvidedJS / "js/dependency.js"` does not work. It results in the following error:

```
> show resolvedJSDependencies
[trace] Stack trace suppressed: run last compile:resolvedJSDependencies for the full output.
[error] (compile:resolvedJSDependencies) org.scalajs.core.tools.jsdep.JSLibResolveException: Some references to JS libraries could not be resolved:
[error] - Missing JS library: js/dependency.js
```

since the `String` returned by `FileVirtualJSFile.relativePath` contains a backslash.

sjrd:
When reading from a directory on the classpath, we need to normalize the relative paths so that they always use '/' as directory separator.

This commit contains no test. We already have the tests that highlight the bug, but they only do it on Windows, and our CI does not run on Windows.

### DIFF
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -342,8 +342,9 @@ object ScalaJSPluginInternal {
         results ++= collectJar(vf)
       } else if (cpEntry.isDirectory) {
         for {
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @gzm0 or @nicolasstucki 

Tested locally on Windows.

gzm0:
LGTM

PR #2364, opened 2016-04-27
Remove wrong overrides of check files for partest 2.11.8.

sjrd:
Running those partests actually produce exactly the output expected by the JVM check files, so these overrides should not be there.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2365, opened 2016-04-27
Improve linking errors display

sjrd:


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -115,7 +116,18 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
       }
 
       val linkingErrLevel = if (fatal) Level.Error else Level.Warn
-      analysis.errors.foreach(logError(_, logger, linkingErrLevel))
+      val maxDisplayErrors = {
+        val propName = "org.scalajs.core.tools.linker.maxlinkingerrors"
+        Try(System.getProperty(propName, "20").toInt).getOrElse(20).max(1)
```
gzm0: How is this not a global, mutable, untyped variable?

It strikes me as odd, that we have nice injection for the loggers but simply use a system property to determine the maximum number of errors to show...

I'm assuming you have considered a lot of other options before resorting to a system property. Can you give me some insights?

sjrd: > How is this not a global, mutable, untyped variable?

It's *slightly* not as bad as that. System properties are not *supposed* to be mutated. An almost reasonable way to use them is to set them once at "process" creation. But yes, it's definitely global, untyped, and potentially mutable.

The alternative was not to offer any means to configure this thing. Yes, we could add a new options object for `BaseLinker` for this option, but then we'd have to trail that option through the layer of abstractions of the linker that we have. Unlike all the other things we allow to configure, this one doesn't even have a use case. Complicating the interface through several layers for something that will never be set, and only has a "visual" impact for a human, is not worth it.

Another possible way to configure it would be a `DynamicVariable` in the companion object of `BaseLinker`. Still global, still mutable, but at least thread-safe and typed. But, for a human, harder to configure.

gzm0: Have you considered adding some option to `logger`? It seems to me that this is really a property of the logger and not of the `BaseLinker`.

Between a `DynamicVariable` and the `SystemProperty` I'd go with the `SystemProperty` as well, because the typing of `DynamicVariable` doesn't buy much but makes configuration much harder.

sjrd: Our logger abstraction is much too generic to be bothered with such an option. How would you imagine configuring that at the logger level?

gzm0:
LGTM

PR #2366, opened 2016-04-27
Implement System.getenv

andreaTP:
just to avoid linking errors.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/System.scala
+++ b/javalanglib/src/main/scala/java/lang/System.scala
@@ -260,8 +260,15 @@ object System {
   def setProperty(key: String, value: String): String =
     SystemProperties.value.setProperty(key, value).asInstanceOf[String]
 
-  //def getenv(): java.util.Map[String,String]
-  //def getenv(name: String): String
+  def getenv(): ju.Map[String, String] =
+    ju.Collections.emptyMap()
+
+  def getenv(name: String): String = {
+    if (name eq null)
+      throw new NullPointerException
+
+    null
+  }
 
   def exit(status: scala.Int): Unit = Runtime.getRuntime().exit(status)
   def gc(): Unit = Runtime.getRuntime().gc()
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
@@ -12,6 +12,7 @@ import language.implicitConversions
 import org.junit.Test
 import org.junit.Assert._
 
+import org.scalajs.testsuite.utils.AssertThrows.expectThrows
 import org.scalajs.testsuite.utils.Platform.executingInJVM
 
 class SystemTest {
@@ -146,5 +147,15 @@ class SystemTest {
 
       assertEquals(().hashCode(), System.identityHashCode(()))
     }
+  }
+
+  @Test def getenv_should_return_an_unmodifiable_map(): Unit = {
+    assertTrue(System.getenv().isInstanceOf[java.util.Map[String, String]])
+
+    expectThrows(classOf[Exception], System.getenv.put("", ""))
+  }
+
+  @Test def getenv_should_link_and_does_not_throw(): Unit = {
+    assertEquals(null, System.getenv(":${PATH}"))
   }
 }


javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -260,8 +260,15 @@ object System {
   def setProperty(key: String, value: String): String =
     SystemProperties.value.setProperty(key, value).asInstanceOf[String]
 
-  //def getenv(): java.util.Map[String,String]
-  //def getenv(name: String): String
+  def getenv(): ju.Map[String,String] =
```
sjrd: Missing space after `,`

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -260,8 +260,15 @@ object System {
...
+    new ju.HashMap[String, String]()
+  
+  def getenv(name: String): String = {
+    if (name eq null) {
```
sjrd: No need for braces here.

sjrd:
This should come with rudimentary tests, asserting that the methods do link and don't behave in a crazy way (e.g., verifying that map returning by `getenv()` "works", as in, it's not `null`).

The map returned by `getenv()` should be *unmodifiable*. There is something for that in `java.util.Collections`: https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#emptyMap%28%29

javalanglib/src/main/scala/java/lang/System.scala
andreaTP: ok

javalanglib/src/main/scala/java/lang/System.scala
andreaTP: ok

andreaTP:
done and rebased

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
```diff
@@ -147,4 +148,10 @@ class SystemTest {
       assertEquals(().hashCode(), System.identityHashCode(()))
     }
   }
+
+  @Test def getenv_should_return_an_unmodifiable_map(): Unit = {
+    assertTrue(System.getenv().isInstanceOf[java.util.Map[String, String]])
+
+    expectThrows(classOf[Exception], System.getenv.put("", ""))
+  }
 }
```
sjrd:
```scala
  @Test def getenvString(): Unit = {
    // Test that the method links and does not throw
    assertEquals(null, System.getenv("ANYVAR"))
  }
```

andreaTP: this could be a problem ... depends on the presence of ```ANYVAR``` variable present in the environment when running tests on the JVM.

Should we do this assumption?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
```diff
@@ -147,4 +148,16 @@ class SystemTest {
...
     }
   }
+
+  @Test def getenv_should_return_an_unmodifiable_map(): Unit = {
+    assertTrue(System.getenv().isInstanceOf[java.util.Map[String, String]])
+
+    expectThrows(classOf[Exception], System.getenv.put("", ""))
+  }
+
+  @Test def getenv_should_link_and_does_not_throw(): Unit = {
```
andreaTP: is this acceptable? this should work on jvm as well as on js.

sjrd: Ah, I didn't realize the JVM case ^^ It's probably better to check something improbable such as `THISVARISNEVERINTHEENV` or something like that, and actually check for `null`.

andreaTP: probably this should do the job! ^^

sjrd:
LGTM. Will need to be rebased after the "Towards 0.6.10" commit, so I don't launch tests now. (our CI is already overloaded by the extensive release tests)

sjrd:
This can now be rebased on top of master.

andreaTP:
rebased

sjrd:

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -260,8 +260,15 @@ object System {
   def setProperty(key: String, value: String): String =
     SystemProperties.value.setProperty(key, value).asInstanceOf[String]
 
-  //def getenv(): java.util.Map[String,String]
-  //def getenv(name: String): String
+  def getenv(): ju.Map[String, String] =
+    ju.Collections.emptyMap()
+
```
sjrd:
```
[error] javalanglib/src/main/scala/java/lang/System.scala:265:0: Whitespace at end of line
```

andreaTP: ...err... =_='

andreaTP:
fixed... I want a CI that automatically fix EOL whitespaces :-P

sjrd:
LGTM, Jenkins, retest this please

PR #2379, opened 2016-05-02
Add JUnit output tests and fix the issues.

nicolasstucki:


nicolasstucki:
Review by @sjrd and @gzm0.

All fixes were done after the creation of the testing tool, as such I saw them failing before I created the fixes.

junit-test/run-tests.py
```diff
@@ -0,0 +1,53 @@
...
+jUnitTestJS = "jUnitTestJS"
+jUnitTestJVM = "jUnitTestJVM"
+
+cmd = ['sbt']
+
+def setFlagsInScope(flags, scope):
+    args = map(lambda x: '"-' + x + '"', flags)
+    cmd.append(
+        "set inScope(ThisScope in " + scope +")(jUnitTestArgs := Seq(" + \
+        ', '.join(args) + "))")
```
sjrd: `\` is not necessary inside the `()` of the call.
Also, consider using a `%`-formatted string:
```python
    cmd.append(
        "set inScope(ThisScope in %s)(jUnitTestArgs := Seq(%s))" %
        (scope, ", ".join(args)))
```

junit-test/run-tests.py
```diff
@@ -0,0 +1,53 @@
...
+
+def testInSope(task, scope):
+    cmd.append(scope + "/" + task)
+
+for test in tests:
+
```
sjrd: Spurious blank line.

project/Build.scala
```diff
@@ -1091,6 +1092,58 @@ object Build {
...
+      publishArtifact in Compile := false,
+      parallelExecution in Test := false,
+      unmanagedSourceDirectories in Test +=
+          baseDirectory.value.getParentFile / "shared/src/test/scala",
+
+      jUnitTestArgs := Seq("-v"),
+      testOptions += Tests.Argument(TestFrameworks.JUnit, jUnitTestArgs.value: _*),
+
+      checkJUnitTestFailuresOutput in Test := {
+        val _ = (testOnly in Test).toTask(" org.scalajs.junit.failure.*").failure.value
```
sjrd: Is there a way we could check that running *any* of the failure tests does indeed fail, rather than testing that when running *all* failure tests?

project/Build.scala
```diff
@@ -1091,6 +1092,58 @@ object Build {
...
+      },
+
+      test in Test := {
+        val _ = (testOnly in Test).
+            toTask(" org.scalajs.junit.success.* org.scalajs.junit.namedecode.*").value
```
sjrd: That's a funky style. I suggest extracting the string in a temp val first:
```scala
val taskArgs = " org.scalajs.junit.success.* org.scalajs.junit.namedecode.*"
val _ = (testOnly in Test).toTask(taskArgs).value
```

project/JUnitTestLogger.scala
```diff
@@ -0,0 +1,165 @@
...
+        (level == Level.Error && msg.contains("Failed: Total"))
+      }
+
+      if (isTestRunFinished)
+        testRunFinished()
+      else if (areTestsFinished)
+        testsFinished()
+
+      log.append(logMsg)
+
```
sjrd: Spurious blank line.

project/JUnitTestLogger.scala
```diff
@@ -0,0 +1,165 @@
...
+          msg.replaceAll(", \\d+(\\.\\d*)?s", "")
+        else if (msg.matches(".*, took \\d+(\\.\\d*)? sec"))
+          msg.substring(0, msg.lastIndexOf(", took "))
+        else
+          msg
+      }
+      val logMsg = s"[$level] $msgWithoutTimes"
+
+      val isTestRunFinished = {
+        /* The first option is always printed at the start of a new test,
```
sjrd: option -> alternative? message?

project/JUnitTestLogger.scala
```diff
@@ -0,0 +1,165 @@
...
+      warnIgnored(referenceOutputIgnored, "reference")
+      warnIgnored(escapedColorLogIgnored, "actual")
+
+      val patch = DiffUtils.diff(referenceOutput2, escapedColorLog2)
+      val deltas = patch.getDeltas
+
+      if (deltas.isEmpty) {
+        loggerOutput(s"$SUCCESS JUnit output matched for $config")
+      } else {
+        loggerOutput(s"$ERROR JUnit output missmatch for $config")
```
sjrd: typo: mismatch

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -261,6 +261,12 @@ object ScalaJSPlugin extends AutoPlugin {
         "scalaJSJavaSystemProperties",
         "List of arguments to pass to the Scala.js Java System.properties.",
         CTask)
+
+    val jUnitTestArgs = SettingKey[Seq[String]]("jUnitTestArgs",
+        "List of flags to pass to the JUnit framework in jUnitTest.", CSetting)
+
+    val checkJUnitTestFailuresOutput = TaskKey[Unit]("checkJUnitTestFailuresOutput",
+        "Executes tests in org.scalajs.junit.failure expecting them to fail.", DTask)
```
sjrd: Those should definitely not be in the published plugin. They are for our own build only. The should be in [`ExposedValues`](https://github.com/scala-js/scala-js/blob/v0.6.9/project/Build.scala#L45)

project/JUnitTestLogger.scala
nicolasstucki: Yes, should be message.

sjrd:
In [the CI log](https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/114413/console), I see the following where the Python script should have run:
```
java.io.IOException: No locks available
	at sun.nio.ch.FileDispatcherImpl.lock0(Native Method)
	at sun.nio.ch.FileDispatcherImpl.lock(FileDispatcherImpl.java:91)
	at sun.nio.ch.FileChannelImpl.tryLock(FileChannelImpl.java:1095)
	at java.nio.channels.FileChannel.tryLock(FileChannel.java:1154)
	at xsbt.boot.Locks$GlobalLock.withChannel$1(Locks.scala:86)
	at xsbt.boot.Locks$GlobalLock.xsbt$boot$Locks$GlobalLock$$withChannelRetries$1(Locks.scala:78)
	at xsbt.boot.Locks$GlobalLock$$anonfun$withFileLock$1.apply(Locks.scala:97)
	at xsbt.boot.Using$.withResource(Using.scala:10)
	at xsbt.boot.Using$.apply(Using.scala:9)
	at xsbt.boot.Locks$GlobalLock.ignoringDeadlockAvoided(Locks.scala:58)
	at xsbt.boot.Locks$GlobalLock.withLock(Locks.scala:48)
	at xsbt.boot.Locks$.apply0(Locks.scala:31)
	at xsbt.boot.Locks$.apply(Locks.scala:28)
	at xsbt.boot.Launch.locked(Launch.scala:238)
	at xsbt.boot.Launch.app(Launch.scala:147)
	at xsbt.boot.Launch.app(Launch.scala:145)
	at xsbt.boot.Launch$.run(Launch.scala:102)
	at xsbt.boot.Launch$$anonfun$apply$1.apply(Launch.scala:35)
	at xsbt.boot.Launch$.launch(Launch.scala:117)
	at xsbt.boot.Launch$.apply(Launch.scala:18)
	at xsbt.boot.Boot$.runImpl(Boot.scala:41)
	at xsbt.boot.Boot$.main(Boot.scala:17)
	at xsbt.boot.Boot.main(Boot.scala)
Error during sbt execution: java.io.IOException: No locks available
```
but then it goes as if it succeeded. That does not look good.

project/Build.scala
nicolasstucki: Oh, I can not do that. `toTask` macro expects its argument to be a `String` literal. 
```scala
[error] /Users/nastucki/GitHub/scala-js/project/Build.scala:1110: Illegal dynamic reference: taskArgs
[error]         val _ = (testOnly in Test).toTask(taskArgs).value
```

sjrd: Arf ... Well in that case the only reasonable way to adhere to the coding style is this:
```scala
val _ = {
  (testOnly in Test)
    .toTask(" org.scalajs.junit.success.* org.scalajs.junit.namedecode.*").value
}
```

nicolasstucki:
Jenkins, retest this please.

gzm0:
I have been thinking a lot about these tests and I really dislike that we need to introduce two new languages (as far as I am concerned, the mini-test language is a language as well) and another dependency on the CI servers (python) that is unmanaged.

Further, I am also not a fan, that test code and assertions are so far apart from each other.

Therefore I would like to propose another solution that uses JUnit to test JUnit.

```scala
import sbt.testing.{Logger, EventListener, TaskDef, Tasks}

abstract class JUnitTest {
  val jUnitFramework = new com.novocode.junit.JUnitFramework()

  val suiteUnderTestName = getClass.getName.stripSuffix("Assertions")

  // records everything including invocation order and result.
  val recorder: Logger with EventListener = ???

  // appropriate class loader for platform, needs platform extension
  val classLoader = ???
  val classLoaderJVM = getClass.getClassLoader
  val classLoaderJS =
    new org.scalajs.testinterface.ScalaJSClassLoader(js.Dynamic.global)

  // To be overriden by subclasses
  def frameworkArgs(): Array[String] = Array.empty

  // run all tasks and the tasks they generate, needs platform extension
  private def executeLoop(tasks: Array[Task]): Unit = ???

  private def executeLoopJVM(): Unit = {
    if (tasks.nonEmpty) {
      executeLoop(tasks.flatMap(_.execute(recorder, Array(recorder))))
    }
  }

  private def executeLoopJS(): Unit = {
    if (tasks.nonEmpty) {
      executeLoop(tasks.flatMap { task => 
        /* This is a hack, assuming the asynchronous continuation will be called before
         * the outer call returns. This is terrible in general, but we do not have
         * another choice, since JUnit does not support async testing.
         * However, since JUnit does not support async testing, this is a safe 
         * assumption to make (module JUnit implementation details, which we control).
         * If JUnit ever were to support async testing, we could also change this here.
         */
        var newTasks: Array[Task] = null
        task.execute(recorder, Array(recorder), newTasks = _)
        newTasks
      })
    }
  }

  @Before
  def setUp(): Unit = {
    val runner = jUnitFramework.runner(frameworkArgs, Array.empty, classLoader)
    val tasks = runner.tasks(new TaskDef(suiteUnderTestName,
        jUnitFramework.fingerprints.head, explicitlySpecified = true,
        selectors = Array.empty))
    executeLoop(tasks)
    recorder.recordDone(runner.done())
  }
}
```

Now, a test could read:

```scala
class AssertTrueTest {
  @Test def test(): Unit = {
    assertTrue(false)
  }
}

class AssertTrueTestAssertions {
  @Test def recordsFailure(): Unit = {
    assertEquals(recorder.failureCount, 1)
  }
}
```

The only thing left to do is to tell sbt not to invoke tests that do not end in "Assertions":

```scala
testOptions in Test += Seq(Tests.Filter(_.endsWith("Assertions")))
```

IMHO, this will lead to much more readable and maintainable tests. Furthermore, we also do not rely on sbt's behavior in our tests but can test against the true interface of the testing framework itself.

WDYT?

sjrd:
Makes sense. It's definitely worth a try.

nicolasstucki:
I will try this approach.
PS The current approach with the sbt extra seams fail most of the times on Jenkins. I have not been able to make it work in a stable way on that platform.

nicolasstucki:
@sjrd @gzm0 here is a version based on @gzm0 idea. It is clearly the correct way to achieve these tests. 

The test are checking the same characteristic of the output as before, plus the exact moment when events are triggered.

nicolasstucki:
Jenkins, retest this please.

junit-test/output-js/src/main/scala/org/scalajs/junit/utils/JUnitTestPlatformImpl.scala
```diff
@@ -0,0 +1,27 @@
+package org.scalajs.junit.utils
+
+import sbt.testing._
+
+import scala.scalajs.js
+
+object JUnitTestPlatformImpl {
+
+  def getClassLoader: ClassLoader =
+      new org.scalajs.testinterface.ScalaJSClassLoader(js.Dynamic.global)
```
sjrd: 2 spaces

junit-test/output-jvm/src/main/scala/org/scalajs/junit/JUnitTestLoggerOutputCheck.scala
```diff
@@ -0,0 +1,33 @@
+package org.scalajs.junit
+
+import java.io.File
+
+object JUnitTestLoggerOutputCheck extends App {
```
sjrd: This file seems to be dead code.

junit-test/shared/src/test/scala/org/scalajs/junit/AssertEquals2Test.scala
```diff
@@ -0,0 +1,31 @@
...
+
+class AssertEquals2TestAssertions extends JUnitTest with FailureFrameworkArgs {
+
+  override val expectedFail: Int = 1
+  override val expectedTotal: Int = 1
+
+  protected def expectedOutput(context: OutputContext): List[Output] = {
+    import context._
+    List(
+      testRunStartedOutput,
```
sjrd: 4 spaces.
This is probably the same in all `expectedOutput` methods.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/FailureFrameworkArgs.scala
```diff
@@ -0,0 +1,19 @@
+package org.scalajs.junit.utils
+
+trait FailureFrameworkArgs {
+  protected def frameworkArgss: List[List[String]] = List(
+    Nil,
```
sjrd: 4 spaces here too

junit-test/shared/src/test/scala/org/scalajs/junit/utils/SuccessFrameworkArgs.scala
```diff
@@ -0,0 +1,12 @@
+package org.scalajs.junit.utils
+
+trait SuccessFrameworkArgs {
+  protected def frameworkArgss: List[List[String]] = List(
+    Nil,
```
sjrd: 4 spaces here too.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+      if (noColor) str
+      else MAGENTA + str + NORMAL
+
+    private def cyan(str: String): String =
+      if (noColor) str
+      else CYAN + str + NORMAL
+
+  }
+
+  protected class JunitTestRecorder(context: OutputContext)
```
sjrd: Uppercase `U` in `JUnit`

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  protected sealed trait Output
+  protected case class Info(msg: String) extends Output
+  protected case class Warn(msg: String) extends Output
+  protected case class Error(msg: String) extends Output
+  protected case class Debug(msg: String) extends Output
+  protected case class Trace(msg: String) extends Output
+  protected case class Event(status: String) extends Output
+  protected case class Done(msg: String) extends Output
```
sjrd: case classes should be `final`

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+    def formattedTestClass: String = {
+      val (pack, cls) =
+        suiteUnderTestName.splitAt(suiteUnderTestName.lastIndexOf('.'))
+      pack + '.' + yellow(cls.tail)
+    }
+
+    def formattedAssumptionViolatedException: String =
+      "org.junit.internal." + red("AssumptionViolatedException")
+
+
```
sjrd: Double blank line.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+    def trace(t: Throwable): Unit = buff += Trace(t.getMessage)
+
+    def handle(event: sbt.testing.Event): Unit = {
+      buff += Event(event.status().toString)
+    }
+
+    def recordDone(msg: String): Unit = {
+      buff += Done(msg)
+      checkOutput(buff.result())
```
sjrd: IMO `checkOutput` should be called explicitly from `testJUnitOutput`, rather than implicitly through `recordDone`.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+      val enc1 =
+        msg.replaceFirst(", took \\d+(\\.\\d*)? sec", s", took $TIME_TAG sec")
+      if (!enc1.contains("Test run finished")) enc1
+      else enc1.replaceFirst("\\d+(\\.\\d*)?s", TIME_TAG + "s")
+    }
+
+    private def checkOutput(actual: List[Output]): Unit = {
+      val expected = expectedOutput(context)
+      type ListTuple = (List[Output], List[Output])
+      def minimizeDiff(lists: ListTuple): ListTuple = {
```
sjrd: Blank line above this line.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+        lists match {
+          case (x::xs, y::ys) if x == y => minimizeDiff(xs, ys)
```
sjrd: Spaces around `::`

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+    private def encodeTimes(msg: String): String = {
+      val enc1 =
+        msg.replaceFirst(", took \\d+(\\.\\d*)? sec", s", took $TIME_TAG sec")
+      if (!enc1.contains("Test run finished")) enc1
+      else enc1.replaceFirst("\\d+(\\.\\d*)?s", TIME_TAG + "s")
+    }
+
+    private def checkOutput(actual: List[Output]): Unit = {
+      val expected = expectedOutput(context)
+      type ListTuple = (List[Output], List[Output])
```
sjrd: Blank space above this line.

sjrd: Or move `val expected` right before `val (ex, ac)`

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
sjrd: Also don't rely on auto-tupling. Add `()` around `xs, ys`.
Alternatively, have `minimizeDiff` take two parameters of type `List[Output]` rather than a `ListTuple`.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
sjrd: This method should be annotated with `@tailrec`, I believe.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,236 @@
...
+      def minimizeDiff(lists: ListTuple): ListTuple = {
+        lists match {
+          case (x::xs, y::ys) if x == y => minimizeDiff(xs, ys)
+          case _                        => lists
+        }
+      }
+
+      val (ex, ac) = minimizeDiff((expected, actual))
```
sjrd: Shouldn't you also try and minimize the diff from the end? For example add
```scala
val (ex2, ac2) = minimizeDiff((ex.reverse, ac.reverse))
val ex3 = ex2.reverse
val ac3 = ac3.reverse
```

project/Build.scala
```diff
@@ -461,6 +461,7 @@ object Build {
               clean in stubs, clean in cli,
               clean in testInterface, clean in jasmineTestFramework,
               clean in jUnitRuntime, clean in jUnitPlugin,
+              clean in jUnitTestOutputsJS, clean in jUnitTestOutputsJS,
```
sjrd: JS twice. One of them should be JVM

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
       fatalWarningsSettings ++ Seq(name := "Scala.js JUnit test runtime")
   ).withScalaJSCompiler.dependsOn(testInterface)
 
+  val commonJUnitTestOutputsSettings = commonSettings ++ fatalWarningsSettings ++ Seq(
+      publishArtifact in Compile := false,
+      parallelExecution in Test := false,
+      unmanagedSourceDirectories in Test +=
+          baseDirectory.value.getParentFile / "shared/src/test/scala",
```
sjrd: 2 spaces

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
...
+      testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a", "-s"),
+      test in Test := {
+        val _ = {
+          (testOnly in Test).toTask(" *Assertions").result.value
+        }
```
sjrd: Doesn't the `val _ = ...` fit on one line?

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
...
+      }
+  )
+
+  lazy val jUnitTestOutputsJS = Project(
+    id = "jUnitTestOutputsJS",
```
sjrd: 4 spaces

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
...
+    base = file("junit-test/output-js"),
+    settings = commonJUnitTestOutputsSettings ++ myScalaJSSettings ++ Seq(
+      name := "Tests for Scala.js JUnit output in JS."
+    )
+  ).withScalaJSCompiler.withScalaJSJUnitPlugin.dependsOn(jUnitRuntime, testInterface)
+
+
+  lazy val jUnitTestOutputsJVM = Project(
+    id = "jUnitTestOutputsJVM",
```
sjrd: 4 spaces

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
...
+    base = file("junit-test/output-jvm"),
+    settings = commonJUnitTestOutputsSettings ++ Seq(
+      name := "Tests for Scala.js JUnit output in JVM.",
+      libraryDependencies ++= Seq(
+          "org.scala-sbt" % "test-interface" % "1.0",
```
sjrd: `% "test"`?

project/Build.scala
```diff
@@ -1092,6 +1093,40 @@ object Build {
...
+      }
+  )
+
+  lazy val jUnitTestOutputsJS = Project(
+    id = "jUnitTestOutputsJS",
+    base = file("junit-test/output-js"),
+    settings = commonJUnitTestOutputsSettings ++ myScalaJSSettings ++ Seq(
+      name := "Tests for Scala.js JUnit output in JS."
+    )
+  ).withScalaJSCompiler.withScalaJSJUnitPlugin.dependsOn(jUnitRuntime, testInterface)
```
sjrd: `% "test"` for both dependencies?

project/Build.scala
sjrd: And, actually, why not manipulating `testFilter` instead, as suggested by @gzm0?

junit-test/shared/src/test/scala/org/scalajs/junit/MethodNameDecodeTest.scala
```diff
@@ -0,0 +1,35 @@
+package org.scalajs.junit
+
+import org.junit.Test
+import org.scalajs.junit.utils.JUnitTest
+
+class MethodNameDecodeTest {
+  @Test def `abcd ∆ƒ 😀 * #&$`(): Unit = ()
```
sjrd: I actually recommend using `\uD83D\uDE00` instead of the actual astral character. IDEs don't like astral characters. Since `\u` processing happens "pre-lexer" in Scala, it's fine to do so.

junit-test/shared/src/test/scala/org/scalajs/junit/MultiAssumeFail1Test.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Assume._
+import org.junit.Test
+
+import org.scalajs.junit.utils.{JUnitTest, SuccessFrameworkArgs}
+
+class MultiAssumeFail1Test {
+  @Test def multiTest1(): Unit = {
+    assumeTrue("This assume should not pass", false)
+  }
+  @Test def multiTest2(): Unit = ()
```
sjrd: Blank line above this line.

sjrd:
That's all I have for the last commit.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
nicolasstucki: If I I get a compile error saying `The outer reference in this type test cannot be checked at run time.`. But I can use `private final` as they also never escape this class.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
nicolasstucki: Yes, but then the diff becomes a useless because it is impossible to know where it should be in the list. I solved this problem by printing in grey all the elements that were dropped by `minimizeDiff`. 

The output will be like:
screen shot 2016-05-20 at 11 56 44

Where the diff is clearly the additional `Event(Failure)`. I'll also remove the color from the non diff lines.

sjrd: Maybe you could keep 3 lines above and below the differences? (like the default behavior of `diff -u`)

nicolasstucki:
This PR might be ready to merge.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,147 +73,145 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
+        logFormattedDebug(decodedMethodName,
+            s"finished, took ${getTimeInSeconds()} sec")
+
+        if (expectedException == classOf[org.junit.Test.None]) {
+          testPassed(methodName)
+          executeAfterMethods()
+        } else {
+          val msg = {
+            s"failed: Expected exception: " + expectedException +
+                s"took ${getTimeInSeconds()} sec"
```
sjrd: 0 space.

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,147 +73,145 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
-          true
-
-        case ex: internal.AssumptionViolatedException =>
-          logFormattedInfo(method.name, "started")
-          logAssertionWarning(method.name, ex, timeInSeconds())
-          true
+        case ex: Throwable =>
+          val timeInSeconds = getTimeInSeconds()
+          if (classOf[AssumptionViolatedException].isInstance(ex) ||
+              classOf[internal.AssumptionViolatedException].isInstance(ex)) {
```
sjrd: Why not `ex.isInstanceOf[AssumptionViolatedException] || ex.isInstanceOf[internal.AssumptionViolatedException`?

junit-runtime/src/main/scala/org/scalajs/junit/JUnitExecuteTest.scala
```diff
@@ -64,147 +73,145 @@ final class JUnitExecuteTest(taskDef: TaskDef, runner: JUnitBaseRunner,
...
+            logAssertionWarning(decodedMethodName, ex, timeInSeconds)
+            testSkipped()
+          } else if (expectedException.isInstance(ex)) {
+            testPassed(methodName)
+            executeAfterMethods()
+          } else if (expectedException == classOf[org.junit.Test.None]) {
+            val isAssertion = ex.isInstanceOf[AssertionError]
+            val failedMsg = new StringBuilder
```
sjrd: Consider a `var failedMsg = ""`

sjrd: Ah nvm, that was like that before. Let's not change it in an unrelated commit.

ci/matrix.xml
```diff
@@ -78,6 +78,9 @@
 
   <task id="test-suite-ecma-script5"><![CDATA[
     setJavaVersion $java
+    sbtretry ++$scala jUnitTestOutputsJVM/test jUnitTestOutputsJS/test \
+        'set scalaJSUseRhino in Global := false' jUnitTestOutputsJS/test \
+        'set scalaJSStage in Global := FullOptStage' jUnitTestOutputsJS/test &
```
sjrd: Shouldn't this be `&&`?

ci/matrix.xml
```diff
@@ -143,6 +146,9 @@
 
   <task id="test-suite-ecma-script6"><![CDATA[
     setJavaVersion $java
+    sbtretry ++$scala jUnitTestOutputsJVM/test jUnitTestOutputsJS/test \
+        'set scalaJSUseRhino in Global := false' jUnitTestOutputsJS/test \
+        'set scalaJSStage in Global := FullOptStage' jUnitTestOutputsJS/test &
```
sjrd: Shouldn't this be `&&`?

sjrd:
That's all.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  protected sealed trait Output
+  private final case class Info(msg: String) extends Output
+  private final case class Warn(msg: String) extends Output
+  private final case class Error(msg: String) extends Output
+  private final case class Debug(msg: String) extends Output
+  private final case class Trace(msg: String) extends Output
+  private final case class Event(status: String) extends Output
+  private final case class Done(msg: String) extends Output
```
nicolasstucki: On `2.12.0-M2` we get the following issue since I added back the `final`, I will have to remove the `final`.
```scala
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:60: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Info(msg: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:61: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Warn(msg: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:62: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Error(msg: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:63: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Debug(msg: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:64: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Trace(msg: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:65: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Event(status: String) extends Output
[error]                            ^
[error] /Users/nastucki/GitHub/scala-js/junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala:66: The outer reference in this type test cannot be checked at run time.
[error]   private final case class Done(msg: String) extends Output
[error]                            ^
```

sjrd: I suggest you put them in the companion object. Together with `Output` as well (which must then be public). You can add a type alias in the class for `Output`: `type Output = JUnitTest.Output` for convenience in subclasses.

case classes nested in classes and traits always have tricky behaviors.

nicolasstucki: That solution seams to work.

sjrd:
LGTM

PR #2388, opened 2016-05-12
Fix #2387: Emit compile error for compiler annotations.

### LINKED ISSUE #2387: Compile error to outlaw user-space use of RawJSType
The compiler should prevent developers from using `@RawJSType` (and other compiler-inserted annotations).

nicolasstucki:


compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
```diff
@@ -127,6 +128,10 @@ trait JSDefinitions { self: JSGlobalAddons =>
     lazy val BoxesRunTime_boxToCharacter = getMemberMethod(BoxesRunTimeModule, newTermName("boxToCharacter"))
     lazy val BoxesRunTime_unboxToChar    = getMemberMethod(BoxesRunTimeModule, newTermName("unboxToChar"))
 
+    lazy val compilerInternalAnnotations = List(
+      RawJSTypeAnnot, JSFullNameAnnotation, SJSDefinedAnonymousClassAnnotation,
+      WasPublicBeforeTyperAnnotation, ExposedJSMemberAnnot
+    )
```
sjrd: This should be defined in `PrepJSInterop`. (also 4 spaces, and `)` at the end of the previous line)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -136,220 +136,224 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
-        val sym = cldef.symbol
-
-        if (sym.hasAnnotation(JSNativeAnnotation)) {
-          reporter.error(cldef.pos, "Traits and classes not extending js.Any " +
-              "may not have a @js.native annotation")
-        }
+    override def transform(tree: Tree): Tree = {
+      if (tree.symbol != null)
+        checkInternalAnnotations(tree.symbol)
+      postTransform { tree match {
```
sjrd: Consider using this opportunity to also change this to:
```scala
val preTransformedTree = tree match {
  ...
}
postTransform(preTransformedTree)
```
to get rid of the `} }` at the end of this match, which is not a style that we use.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1031,17 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       ddef
     }
   }
+
+  private def checkInternalAnnotations(sym: Symbol): Unit = {
+    for {
+      ann <- compilerInternalAnnotations
+      annSym <- sym.getAnnotation(ann)
+      if annSym.pos.isDefined
```
sjrd: This is a bit dangerous. It would fail to report an error simply because the position was not set, for some reason. Why does this test exist at all?

compiler/src/test/scala/org/scalajs/core/compiler/test/InternalAnnotationsTest.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.scalajs.core.compiler.test.util._
+
+// scalastyle:off line.size.limit
+
+class InternalAnnotationsTest extends DirectTest with TestHelpers {
+
+  override def preamble: String =
+    "import scala.scalajs.js, js.annotation._"
+
+
```
sjrd: Double blank line.

compiler/src/test/scala/org/scalajs/core/compiler/test/InternalAnnotationsTest.scala
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  def exposedJSMember: Unit = {
+    test("ExposedJSMember")
+  }
+
+
```
sjrd: Double blank line.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1031,17 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       ddef
     }
   }
+
+  private def checkInternalAnnotations(sym: Symbol): Unit = {
+    for {
+      ann <- compilerInternalAnnotations
```
sjrd: You'll have to make an exception for `@RawJSType` on `js.UndefOr` and `js.|`, I'm afraid. These two union types are implemented partly as a hack by artificially making them raw-js-type.

Alternative: have the compiler mark them `@RawJSType`.

sjrd:
That's all.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: In practice, the position is only defined if the annotation came from the file. I added it to filter out the `@WasPublicBeforeTyper` that were generated in `PreTyperComponent`.

sjrd: Ouch. I'm afraid things generated by a macro might also have missing positions and the like.

If there is no better way to deal with `@WasPublicBeforeTyper`, then you should only apply this filter for that particular annotation + write a comment.

nicolasstucki: I will probably have to check for `@WasPublicBeforeTyper` in `PreTyperComponent` and only base the filter by the name.

gzm0: Does the typer expand the annotation trees into fully qualified names? If not, you can check for the fully qualified name in `PreTyperComponent` and for the type in `PrepJSInterop` and whitelist fully qualified names.

That solution would not have any false positives.

nicolasstucki: Unfortunately in `PreTyperComponent` the name is still not fully qualified. 
For the ones detected in `PrepJSInterop` the type works perfectly.

nicolasstucki: To make things even worse, in `PreTyperComponent` we also have the issue of aliasing.

gzm0: > Unfortunately in PreTyperComponent the name is still not fully qualified. 

Of course not. It needs the typer to find out the fully qualified name. But if the typer doesn't change the trees, you know that it is not a tree generated by the compiler if it is not fully qualified.

nicolasstucki: Ok, that would eliminate most of them. There would remain the case where the user writes it fully qualified, any ideas for that one?

gzm0: Yes, fail in `PreTyperComponent`. Because if the name is fully qualified, you are sure it actually refers to the one you want to exclude.

nicolasstucki: Oh, yes. Thanks.

nicolasstucki: Typer does change the trees. :(

nicolasstucki: And the `original` tree is `null`.

sjrd:
LGTM, but I'll also give @gzm0 a chance to look at it.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -136,220 +136,229 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+                List(Literal(Constant("x")))
+            )
+          }
+          typer.typed(newTree, Mode.FUNmode, tree.tpe)
+
+        case _ => super.transform(tree)
+      }
 
-      case _ => super.transform(tree)
-    } }
```
gzm0: I'm assuming all this is just indented and unchanged.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1036,27 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+  private def checkInternalAnnotations(tree: Tree): Unit = {
+    def isPreTyper(annotation: AnnotationInfo) =
+      annotation.symbol == WasPublicBeforeTyperClass && annotation.pos.isDefined
+
+    def isPostTyper(annotation: AnnotationInfo) = {
+      annotation.symbol == ExposedJSMemberAnnot ||
+      annotation.symbol == JSFullNameAnnotation ||
+      annotation.symbol == RawJSTypeAnnot ||
+      annotation.symbol == SJSDefinedAnonymousClassAnnotation
+    }
```
gzm0: I'm ok with the names `isPreTyper` and `isPostTyper`, but please add a comment why you need two of them and the shortcuts you take (i.e. a summary of the discussion we had). In that comment, please mention that `@WasPublicBeforeTyper` could still be inserted by a macro, but we can't reliably check for that with reasonable effort.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -136,220 +136,229 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
 
-        if (shouldPrepareExports && sym.isTrait) {
-          // Check that interface/trait is not exported
-          for {
-            exp <- exportsOf(sym)
-            if !exp.ignoreInvalid
-          } reporter.error(exp.pos, "You may not export a trait")
-        }
+          if (sym == UndefOrClass || sym == UnionClass)
+            sym.addAnnotation(RawJSTypeAnnot)
```
gzm0: Ah, this is new :)

nicolasstucki: Yes, it is the only new code in `transform` apart from the call to `checkInternalAnnotations`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
nicolasstucki: Done

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1036,41 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       ddef
     }
   }
+
+  private def checkInternalAnnotations(tree: Tree): Unit = {
+    /* The compiler internal annotations are categorized into two:
```
gzm0: groups:

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1036,41 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+     * 1) postTyper: are annotations that are inserted in any phase
+     *   after the typer.
```
gzm0: I would omit "are". Also, indentation issue.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1036,41 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+     * 2) preTyper: are annotations that are inserted before typer
+     *   (such as @WasPublicBeforeTyper), as the type is required to find them,
+     *   the check must be after typer. We use an heuristic to find the
+     *   instances of these annotations based on the existence of their position
+     *   in the source file. Unfortunately it is still possible for macros to
+     *   generate such annotations trees without positions.
+     *
+     * WARNING: This check must be done in the earliest phase after typer that
+     * inserts scalajs complier internal annotations.
```
gzm0: What do you intend with this comment? If you put it, explain why (although it is kind of obvious). (I do not understand why someone would move the check though).

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1027,6 +1036,41 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
     }
   }
+
+  private def checkInternalAnnotations(tree: Tree): Unit = {
+    /* The compiler internal annotations are categorized into two:
+     * 1) postTyper: are annotations that are inserted in any phase
+     *   after the typer.
+     * 2) preTyper: are annotations that are inserted before typer
+     *   (such as @WasPublicBeforeTyper), as the type is required to find them,
+     *   the check must be after typer. We use an heuristic to find the
```
gzm0: I think in english "h" does not count as vowel.

gzm0:
LGTM @sjrd?

sjrd:
LGTM

PR #2394, opened 2016-05-16
Fix #2357: Implement printing for ir.PropertyDef

### LINKED ISSUE #2357: Fully implement printing for ir.PropertyDef

gzm0:


gzm0:
Review by @sjrd

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -975,11 +975,40 @@ class PrintersTest {
 
   @Test def printPropertyDef(): Unit = {
     assertPrintEquals(
-        """<property: "prop">""",
+        """
+          |get "prop"(): int = {
```
sjrd: Getters are supposed to always return `any`, since they are called from JS. The displayed result type should be `any`.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -809,11 +809,23 @@ object Printers {
...
-        case PropertyDef(name, _, _, _) =>
-          // TODO
-          print("<property: ")
-          print(name)
-          print('>')
+        case PropertyDef(name, getterBody, setterArg, setterBody) =>
+          if (getterBody != EmptyTree) {
+            print("get ")
+            print(name)
+            printSig(Nil, getterBody.tpe)
```
sjrd: In other words: `getterBody.tpe` -> `AnyType`

sjrd:
That's all.

gzm0:
Updated

sjrd:
LGTM

PR #2395, opened 2016-05-16
Add CrossProject.configureAll & rename/alias/deprecate configure/configureCross

dwijnand:
WDYT?

### DIFF
--- a/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
+++ b/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
@@ -243,8 +243,17 @@ final class CrossProject private (
   def configs(cs: Configuration*): CrossProject =
     copy(jvm.configs(cs: _*), js.configs(cs: _*))
 
-  def configure(transforms: (CrossProject => CrossProject)*): CrossProject =
+  def configureCross(transforms: (CrossProject => CrossProject)*): CrossProject =
     transforms.foldLeft(this)((p, t) => t(p))
+
+  @deprecated("Use configureCross instead.", "0.6.10")
+  def configure(transforms: (CrossProject => CrossProject)*): CrossProject =
+    configureCross(transforms: _*)
+
+  // TODO: rename to "configure" when compatibility can be broken (1.0.0)
+  //       and the existing deprecated "configure" is removed
+  def configureAll(transforms: (Project => Project)*): CrossProject =
+    copy(jvm.configure(transforms: _*), js.configure(transforms: _*))
 
   def dependsOn(deps: CrossClasspathDependency*): CrossProject =
     copy(jvm.dependsOn(deps.map(_.jvm): _*), js.dependsOn(deps.map(_.js): _*))


dwijnand:
Actually more generally it would be useful if `copy` were public, even if it need be named something like 'copyAndWatchYourStep' or 'unsafeCopy' or whatever.

sjrd:

@gzm0 Do we have any infrastructure to test additional methods on `CrossProject`?

> Actually more generally it would be useful if `copy` were public, even if it need be named something like 'copyAndWatchYourStep' or 'unsafeCopy' or whatever.

That really doesn't look like a good idea to me.

sjrd:

dwijnand:
>>Actually more generally it would be useful if copy were public, even if it need be named something like 'copyAndWatchYourStep' or 'unsafeCopy' or whatever.

> That really doesn't look like a good idea to me.

The reason is I'd like to use the existing 'configure' method, with some `Project => Project` functions that I can get to auto-lift to `CrossProject => CrossProject`. That way I can consume my straight-forward `Project => Project` functions, but also be more JVM/JS specific where needed (eg adding `junit-interface` vs enabling `ScalaJSJUnitPlugin`).

dwijnand:
Doh.. just realised this is already achievable in user-land with `.jvmConfigure` and `.jsConfigure`.

japgolly:
My 2 cents: I'd like to see this in place as `.bothConfigure`. I'd like to see all the `.{jvm,js}Xxx` methods become `.{jvm,js,both}Xxx`.

gzm0:
> Do we have any infrastructure to test additional methods on CrossProject?

Other than the `sbt-plugin-test`, I don't think so.

>  I'd like to see this in place as .bothConfigure. I'd like to see all the .{jvm,js}Xxx methods become .{jvm,js,both}Xxx.

Very valid point to have it everywhere if we have it. Suggest to replace `both` with `all`. Just in case this ever extends to another platform (native) :)

sjrd:
Or existing convention is actually `{js,jvm,}`. See for example `settings`.

sjrd:
To keep the conventions, we could make this an overload of `configure` with a `DummyImplicit` to fight back the double definition after erasure.

gzm0:
> To keep the conventions, we could make this an overload of configure with a DummyImplicit to fight back the double definition after erasure.

Indeed. Maybe we should even deprecate the other one and give it a better name?

sjrd:
Possibly, but I don't have a better name to suggest.

dwijnand:
crossConfigure

sjrd:
Actually, I think that `configure` is the right name for both methods, and they should meaningfully be overloaded. In both cases you configure the cross-project itself. You're just allowed to give two different types of functions as argument, but essentially the two methods are doing the same thing.

gzm0:
> Actually, I think that configure is the right name for both methods, and they should meaningfully be overloaded.

Doesn't that mess up type-inference for argument lambdas? It'd be pretty annoying, if you had to write the type in the lambda all the time, no?

sjrd:
Oh ... Yes, it does. Well that's not an option, then.

sjrd:
Summarizing:
* They cannot both be called `configure`, because that messes up parameter type inference
* For consistency with other methods, ideally `configure` should be the one acting on `Project`
* The best name I've seen for the one acting on `CrossProject` would be `crossConfigure`. I would actually suggest `configureCross` as it configures the cross-project (it does not cross-configure anything).

Since we cannot remove the current `configure` for now, we cannot introduce the new `configure(Project)` now either (as, even if the old one is deprecated, it would also mess up param type inference). I therefore suggest the following plan of action:

1. In 0.6.10:
  * rename `configure(CrossProject)` into `configureCross`. Leave the old one as a deprecated alias.
  * introduce `configureAll(Project)`.
2. In 1.0.0:
  * remove the deprecated `configure(CrossProject)`
  * rename `configureAll` to `configure`, with a deprecated alias

dwijnand:
Sounds good, and I agree to that reasoning for configureCross.

dwijnand:
Updated.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala
```diff
@@ -243,9 +243,18 @@ final class CrossProject private (
   def configs(cs: Configuration*): CrossProject =
     copy(jvm.configs(cs: _*), js.configs(cs: _*))
 
-  def configure(transforms: (CrossProject => CrossProject)*): CrossProject =
+  def configureCross(transforms: (CrossProject => CrossProject)*): CrossProject =
     transforms.foldLeft(this)((p, t) => t(p))
 
+  @deprecated("Use configureCross instead.", "0.6.9")
```
sjrd: This should be 0.6.10 ;)

sjrd:
That's all.

dwijnand:
Derp. Updated.

sjrd:

dwijnand:
nice thanks

PR #2397, opened 2016-05-17
Add implicit class to convert scala Future to js.Promise

axos88:
Adds the posibility to do

Future { ... }.toJSPromise

for cases where you need to call JavaScript code that expects a promise (Such as RxJS observable flatMaps)


sjrd:
This already exists in `JSConverters`

PR #2403, opened 2016-05-22
Enable JSPromise conversion for "anything Future-like"

axos88:
This allows chaining the implicit conversions such as going from a Q.Promise to a JSPromise, as in:
```scala
httpService.get[T](url)toJSPromise
```

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/JSConverters.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSConverters.scala
@@ -21,7 +21,7 @@ import scala.scalajs.runtime.genTraversableOnce2jsArray
 sealed abstract class JSConvertersLowPrioImplicits { this: JSConverters.type =>
 
   @inline
-  implicit def JSRichFutureNonThenable[A](f: Future[A]): JSRichFuture[A] =
+  implicit def JSRichFutureNonThenable[A, A1](f: A1)(implicit ev: A1 => Future[A]): JSRichFuture[A] =
     new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
 
 }
@@ -63,7 +63,7 @@ object JSConverters extends JSConvertersLowPrioImplicits {
     new JSRichGenTraversableOnce(arr)
 
   @inline
-  implicit def JSRichFutureThenable[A](f: Future[Thenable[A]]): JSRichFuture[A] =
+  implicit def JSRichFutureThenable[A, A1](f: A1)(implicit ev: A1 => Future[Thenable[A]]): JSRichFuture[A] =
     new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
 
   final class JSRichFuture[A](val self: Future[A | Thenable[A]]) extends AnyVal {
--- a/library/src/main/scala/scala/scalajs/js/JSConverters.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSConverters.scala
@@ -22,7 +22,7 @@ sealed abstract class JSConvertersLowPrioImplicits { this: JSConverters.type =>
 
   @inline
   implicit def JSRichFutureNonThenable[A, A1](f: A1)(implicit ev: A1 => Future[A]): JSRichFuture[A] =
-    new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
+    new JSRichFuture[A](ev(f).asInstanceOf[Future[A | Thenable[A]]])
 
 }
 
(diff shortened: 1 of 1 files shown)


sjrd:
Thanks for your contribution, however I do not believe this is a good idea. Implicit chaining should be enabled with care, as it can quickly become difficult to track.

IMO it is better to explicitly enable such a chain with a user-defined implicit like this:
```scala
implicit def RichQPromise[A](p: Q.Promise[A]): js.JSConverters.JSRichFuture[A] =
  js.JSConverters.JSRichFutureNonThenable(p)
```
Also, your current implementation is not precise enough: since it is not possible to have a `Q.Promise[Thenable[A]]` (for the same reason that there cannot be a `Thenable[Thenable[A]]`, there is no need for the `Thenable`/`NonThenable` distinction.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -21,7 +21,7 @@ import scala.scalajs.runtime.genTraversableOnce2jsArray
 sealed abstract class JSConvertersLowPrioImplicits { this: JSConverters.type =>
 
   @inline
-  implicit def JSRichFutureNonThenable[A](f: Future[A]): JSRichFuture[A] =
+  implicit def JSRichFutureNonThenable[A, A1](f: A1)(implicit ev: A1 => Future[A]): JSRichFuture[A] =
     new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
```
sjrd: This completely by-passes `ev`. The cast applies without `ev` kicking in first. At least this should be `ev(f).asInstanceOf[...]`.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -63,7 +63,7 @@ object JSConverters extends JSConvertersLowPrioImplicits {
     new JSRichGenTraversableOnce(arr)
 
   @inline
-  implicit def JSRichFutureThenable[A](f: Future[Thenable[A]]): JSRichFuture[A] =
+  implicit def JSRichFutureThenable[A, A1](f: A1)(implicit ev: A1 => Future[Thenable[A]]): JSRichFuture[A] =
     new JSRichFuture[A](f.asInstanceOf[Future[A | Thenable[A]]])
```
sjrd: Same problem with `ev` missing.

axos88:
ev is missing, hold on

library/src/main/scala/scala/scalajs/js/JSConverters.scala
axos88: Yeah, I just discovered this. Will update in a bit

axos88:
I agree that implicit chaining should be taken with care however in this case I find it hard to find a case where it would casue surprises.

It is used only for extending "anything that can be converted to a Future" to be able to be converted to the JS equivalent JSPromise as well. That conversion however needs to be called explicitly, so converting it to a Future before converting it to the JSPromise must be the developer's intent.

sjrd:
Hum sorry I may not have been clear enough. When I said only the case `NonThenable` was relevant, that applied only to `Q.Promise`. For general convertibles, both are necessary.

I am still not convinced this is a good idea. `Q.Promise` can have a better translation to `js.Promise` than this one.

Also, this change would require tests. And I'm afraid in any case or won't be binary compatible anyway.

axos88:
Can you rephrase your last sentence?

sjrd:
Oops sorry, phone auto correction. I edited my earlier comment.

sjrd:
> `Q.Promise` can have a better translation to `js.Promise` than this one.

Actually even better than I thought. You can convert a `p: Q.Promise` to `js.Promise` natively with `js.Promise.resolve(p)`. So it's really better to define a conversion such as
```scala
implicit class QPromiseOps[A](val self: Q.Promise[A]) extends AnyVal {
  def toJSPromise: js.Promise[A] = js.Promise.resolve(self)
}
```

mejine:
extends AnyVal {
  def toJSPromise: js.Promise[A] = js.Promise.resolve(self)

sjrd:
@mejine Hum ... did you send your comment too soon? I don't understand it.

mejine:
Hum sorry ??

sjrd:
Well I did not understand your comment. It seems to be a copy-paste of part of my own comment. Basically it looks like your comment is missing some sentence or explanation or, well, something. Would you mind elaborating a bit on what you were trying to say?

axos88:
Hmm. Have to check out wether this works with failed promises as well.

PR #2405, opened 2016-05-23
Fix #2400: Reimplement Arrays.sort with a stable sort.

### LINKED ISSUE #2400: java.util.Arrays.sort is not stable.
Found this bug from the test `scala.util.SortingTest` in `scalaTestSuite`.

```scala
case class N(i: Int, j: Int)

val xs = Array.tabulate(16)(i => N(0, i))
println(xs.mkString("[", ", ", "]"))
    
val sxs = {
  val temp = xs.clone;
  val cmp = new Comparator[N] {
    def compare(o1: N, o2: N): Int =
      if (o1.i < o2.i) -1 else if (o1.i > o2.i) 1 else 0
  }
  java.util.Arrays.sort(temp, cmp)
  temp
}
println(sxs.mkString("[", ", ", "]"))
```

will print:
```
[N(0,0), N(0,1), N(0,2), N(0,3), N(0,4), N(0,5), N(0,6), N(0,7), N(0,8), N(0,9), N(0,10), N(0,11), N(0,12), N(0,13), N(0,14), N(0,15)]
[N(0,1), N(0,2), N(0,3), N(0,4), N(0,5), N(0,6), N(0,7), N(0,8), N(0,0), N(0,15), N(0,9), N(0,10), N(0,11), N(0,12), N(0,13), N(0,14)]
```

nicolasstucki:


nicolasstucki:
Review by @sjrd

nicolasstucki:
Oops, forgot to update all the `WhitelistedTests` and `BlacklistedTests`.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -80,121 +80,68 @@ object Arrays {
...
+    else
       insertionSort(a, i0, iN)
+  }
+
+  @noinline
+  private def stableSplitMerge[@specialized K](a: Array[K], b: Array[K],
+      i0: Int, iN: Int)(implicit ord: Ordering[K]): Unit = {
+    val length = iN - i0
+    if (length > inPlaceSortThreshold) {
+      val iM = i0 + (length / 2)
```
sjrd: Use `(i0 + iN) >>> 1 // Unsigned div by 2` which does not have the overflow problem.

sjrd: Ah never mind. `length / 2` doesn't overflow either.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -80,121 +80,68 @@ object Arrays {
...
+          merge(k + 1, i + 1, j)
       } else {
-        quickSort(a, iB, iN)  // True recursion
-        quickSort(a, i0, iA)  // Should be tail recursion
+        b(k) = a(j)
+        if (k < iN)
+          merge(k + 1, i, j + 1)
       }
     }
+    merge(i0, i0, iM)
```
sjrd: IMO this would be more readable with a good old while loop.

Also give meaningful names to the variables, e.g.,
* `k` -> `outIndex`
* `i` -> leftInIndex`
* `j` -> rightInIndex`

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -237,97 +184,48 @@ object Arrays {
...
+          merge(k + 1, i + 1, j)
       } else {
-        quickSortAnyRef(a, iB, iN)  // True recursion
-        quickSortAnyRef(a, i0, iA)  // Should be tail recursion
+        b(k) = a(j)
+        if (k < iN)
+          merge(k + 1, i, j + 1)
       }
     }
+    merge(i0, i0, iM)
```
sjrd: Same remarks here, obviously.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/ArraysTest.scala
```diff
@@ -83,6 +83,25 @@ class ArraysTest {
...
+    }
+
+    def isStable(a: scala.Array[N]): Boolean =
+      a.zipWithIndex.forall { case (n, i) => n.i == i }
+
+    for (size <- Seq(2, 4, 5, 15, 16, 19, 32, 33, 63, 64, 65, 4356)) {
+      val sxs = scala.Array.tabulate(size)(i => N(i))
+      Arrays.sort(sxs, cmp)
+
+      assertTrue(s"Arrays.sort wasn't stable on array of size $sxs.'",
```
sjrd: `of size $sxs` -> `of size $size`?

sjrd:
That's all.

nicolasstucki:
I had to disable `scala/scala` `scala/util/SortingTest.scala` test because it times out on Rhino.

sjrd:
LGTM

PR #2410, opened 2016-05-23
java.util.logging implementation

cquiroz:
This PR contains a partial implementation of the java.util.logging package. JULI is not so widely used but still is part of the platform and we have several packages using it. Having support in the platform would make porting much easier. This PR includes all the clases on the package as it is practical and tests

Some Notes:
* The PR is missing some of the handlers (Like `SocketHandler` or `FileHandler`) that don't make sense in the js environment.
* Another missing class is `LogManager`, used to configure logging but it relies heavily on configuration files and reflection. Instead you can manually configure on code the output handlers, formats, levels, etc
* Console handler uses `System.err` as indicated on the javadocs. That is a poor choice for the browser environment but I don't think we are allowed to change it
* I have a handler that uses Ajax to post to the backend but I believe this wouldn't belong to the core scala.js. Would it be a good addition to scala-js-dom?

Please let me know if this is an acceptable contribution to the platform and if it needs any changes on code formatting, history, etc


sjrd:
Jenkins, test this please

sjrd:
Woof! That's some massive contribution.

IMO, for such large and independent parts of the JDK, it would be best to create separate repositories, like we did with [`scalajs-java-time`](https://github.com/scala-js/scala-js-java-time). This has two main advantages:
* For users: if you don't need these parts, you do not pay the price of their footprint on your classpath and in the memory of the linker.
* For us: it does not put an unnecessary burden on our already overloaded CI (we can easily off-load the CI of other repositories to Travis)

We can start discussing the code here though, while we debate this.

sjrd:
Review by @nicolasstucki

cquiroz:
I think the build failed because the method `getGlobal` is available only since java 1.7 thus the jvm test fails. How do you go about those cases? By not implementing the method? Or not testing? Any other alternative?

sjrd:
Put is tests in a `require-jdk7` subdirectory. There are existing cases like this in the repo.

cquiroz:
I moved the test files and tested locally on JDK 6, JDK 7 and JDK 8

sjrd:
Jenkins, retest this please

javalib/src/main/scala/java/util/logging/ConsoleHandler.scala
```diff
@@ -0,0 +1,19 @@
+package java.util.logging
+
+// Unfortunately the spec require the use of System.err
+class ConsoleHandler extends StreamHandler(System.err, new SimpleFormatter) {
```
nicolasstucki: The `Formatter` should be loaded from the `LogManager` and default to `new SimpleFormatter` if it is not defined there. The doc says: 


**Configuration**: By default each ConsoleHandler is initialized using the following LogManager configuration properties. If properties are not defined (or have invalid values) then the specified default values are used.
...
java.util.logging.ConsoleHandler.formatter specifies the name of a Formatter class to use (defaults to java.util.logging.SimpleFormatter).
...

javalib/src/main/scala/java/util/logging/ErrorManager.scala
```diff
@@ -0,0 +1,25 @@
...
+
+class ErrorManager() {
+  // The spec is a bit vague. This will implement the most
+  // obvious interpretation of outputting only once
+  private[this] var called = false
+
+  def error(msg: String, ex: Exception, code: Int): Unit = {
+    if (!called) {
+      System.err.println(msg)
+      ex.printStackTrace(System.err)
```
nicolasstucki: `msg` and `ex` may be null. You also need to print code at the start. Look at the output from the JVM:

```scala
scala> new java.util.logging.ErrorManager().error(null, null, 42)
java.util.logging.ErrorManager: 42

scala> new java.util.logging.ErrorManager().error("abc", null, 42)
java.util.logging.ErrorManager: 42: abc

scala> new java.util.logging.ErrorManager().error("abc", new Exception, 42)
java.util.logging.ErrorManager: 42: abc
java.lang.Exception
	at $line37.$read$$iw$$iw$.<init>(<console>:11)
	at $line37.$read$$iw$$iw$.<clinit>(<console>)
	at $line37.$eval$.$print$lzycompute(<console>:7)
	at $line37.$eval$.$print(<console>:6)
	at $line37.$eval.$print(<console>)
	...
```

javalib/src/main/scala/java/util/logging/Formatter.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  def format(record: LogRecord): String
+
+  def getHead(h: Handler): String = ""
+
+  def getTail(h: Handler): String = ""
+
+  def formatMessage(record: LogRecord): String =
+    // The Java spec uses java.text formatting not available in Scala.js
+    if (Option(record.getParameters).exists(_.nonEmpty)) {
```
nicolasstucki: This is overkill and it hides the intent, just use
```scala
val recordParams = record.getParameters
if (recordParams != null && recordParams.nonEmpty) {
  ...
```

javalib/src/main/scala/java/util/logging/Formatter.scala
```diff
@@ -0,0 +1,22 @@
...
+      // Simple text replacement
+      val msg = record.getMessage
+      record.getParameters.zipWithIndex.foldRight(msg) { case ((x, i), t) =>
+        t.replace(s"{$i}", x.toString)
```
nicolasstucki: This transformation will create a linear amount of `Strings` (in respect to number of parameters). Consider using a `StringBuilder` and a single pass through `msg`.

javalib/src/main/scala/java/util/logging/Handler.scala
```diff
@@ -0,0 +1,54 @@
+package java.util.logging
+
+import java.nio.charset.{Charset, UnsupportedCharsetException}
+
+abstract class Handler protected () {
+  private[this] var level:Option[Level] = Some(Level.ALL)
+  private[this] var filter:Option[Filter] = None
+  private[this] var formatter:Option[Formatter] = None
+  private[this] var encoding:Option[String] = None
+  private[this] var errorManager:Option[ErrorManager] = None
```
nicolasstucki: Missing space between the `:` and `Option[ErrorManager]`. The same applies for the method types bellow.

javalib/src/main/scala/java/util/logging/Handler.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def close():Unit
+
+  def setFormatter(formatter: Formatter):Unit =
+    this.formatter = Option(formatter)
+
+  def getFormatter():Formatter = formatter.orNull
+
+  def setEncoding(encoding: String):Unit =
+    if (Charset.isSupported(encoding)) this.encoding = Option(encoding)
```
nicolasstucki: `encoding` may be `null`, in that case the default encoding should be used.

javalib/src/main/scala/java/util/logging/Handler.scala
```diff
@@ -0,0 +1,54 @@
...
+  def setLevel(level: Level):Unit =
+    this.level = Option(level)
+
+  def getLevel():Level = level.orNull
+
+  def isLoggable(record:LogRecord):Boolean =
+    level.forall(_.intValue() <= record.getLevel.intValue()) &&
+      filter.forall(_.isLoggable(record))
+
+  protected def encodingOrDefault:Option[Charset] =
```
nicolasstucki: You should not add new `protected` methods to java classes.

javalib/src/main/scala/java/util/logging/Level.scala
```diff
@@ -0,0 +1,48 @@
...
+object Level {
+  val OFF: Level     = new Level("OFF", Int.MaxValue)
+  val SEVERE: Level  = new Level("SEVERE", 1000)
+  val WARNING: Level = new Level("WARNING", 900)
+  val INFO: Level    = new Level("INFO", 800)
+  val CONFIG: Level  = new Level("CONFIG", 700)
+  val FINE: Level    = new Level("FINE", 500)
+  val FINER: Level   = new Level("FINER", 400)
+  val FINEST: Level  = new Level("FINEST", 300)
+  val ALL: Level     = new Level("ALL", Int.MinValue)
```
nicolasstucki: Do not vertically align the `=` in `val` definitions.

javalib/src/main/scala/java/util/logging/Level.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private val knownLevels = List(OFF, SEVERE, WARNING, INFO, CONFIG, FINE,
```
nicolasstucki: Make this one `lazy`.

nicolasstucki: Consider creating a map instead, and having a trivial implementation for `parse`.
```scala
private lazy val knownLevels = Map[String, Level](
    OFF.name -> OFF, OFF.intValue -> OFF,
    SEVERE.name -> SEVERE, SEVERE.intValue -> SEVERE,
    WARNING.name -> WARNING, WARNING.intValue -> WARNING,
    ...
)
```

javalib/src/main/scala/java/util/logging/Level.scala
```diff
@@ -0,0 +1,48 @@
...
+    if (name == null) throw new NullPointerException("Name cannot be null")
+    knownLevels.find(l => l.getName == name || l.intValue().toString == name)
+      .fold(throw new IllegalArgumentException(""))(l => l)
+  }
+}
+
+class Level protected (private[this] val name: String,
+    private[this] val value: Int,
+    private[this] val resourceBundleName: String) {
+  if (name == null) throw new NullPointerException("Name cannot be null")
```
nicolasstucki: break the line before the `throw`

javalib/src/main/scala/java/util/logging/Level.scala
```diff
@@ -0,0 +1,48 @@
...
+  override def toString():String = name
+
+  def intValue():Int = value
+
+  override def equals(ox: Any):Boolean = ox match {
+    case l: Level => intValue() == l.intValue()
+    case _        => false
+  }
+
+  override def hashCode():Int = value.hashCode()
```
nicolasstucki: In this case `value.hashCode()` is equal to `value`.

javalib/src/main/scala/java/util/logging/LogRecord.scala
```diff
@@ -0,0 +1,83 @@
+package java.util.logging
+
+import java.util.concurrent.atomic.AtomicLong
+
+object LogRecord {
+  protected[logging] var sequence:AtomicLong = new AtomicLong(0)
```
nicolasstucki: Use a plain `Long`, it is impossible to have concurrency issues.

javalib/src/main/scala/java/util/logging/LogRecord.scala
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class LogRecord protected[logging] (private[this] var level: Level,
+    private[this] var msg: String,
+    private[this] var sourceClassName: Option[String] = None,
+    private[this] var sourceMethodName: Option[String] = None,
+    private[this] var params: List[AnyRef] = Nil,
+    private[this] var thrown: Option[Throwable] = None,
+    private[this] var loggerName: Option[String] = None) {
```
nicolasstucki: you should not define a second constructor on this class. Put all the `var` definitions in the class and use `(level: Level, msg: String)` as the only constructor.

javalib/src/main/scala/java/util/logging/LogRecord.scala
```diff
@@ -0,0 +1,83 @@
...
+  def getMessage(): String = msg
+
+  def setMessage(message: String): Unit = msg = message
+
+  def getParameters(): Array[AnyRef] =
+    if (params.isEmpty) null
+    else params.toArray
+
+  def setParameters(parameters: Array[AnyRef]): Unit =
+    if (parameters == null) this.params = Nil
```
nicolasstucki: Experimentation on the JVM shows that `params` should be the reference `parameters`. It is possible to change the parameters after `setParameters` by changing the array.

javalib/src/main/scala/java/util/logging/LogRecord.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  def setThreadID(threadID: Int): Unit = this.threadId = threadID
+
+  def getMillis(): Long = millis
+
+  def setMillis(millis: Long): Unit = this.millis = millis
+
+  def getThrown(): Throwable = thrown.orNull
+
+  def setThrown(thrown: Throwable): Unit = this.thrown = Option(thrown)
```
nicolasstucki: Consider keeping the reference to the `Throwable` directly rather than boxing it in a `Option[Throwable]` as you do not use any useful method of `Option`. There are several places where this would make sense.

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+
+  def getAnonymousLogger():Logger =
+    // No references to anonymous loggers are kept
+    new Logger(null, None, None, true, Some(rootLogger))
+
+  // Not implemented, no resource bundle in scala.js
+  //def getAnonymousLogger(resourceBundle: String):Logger
+}
+
+class Logger private (protected val name: String,
```
nicolasstucki: `name, `resourceBundle` and the rest of this constructors `var`s should not be protected fields of `Logger`.

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+  //def getLogger(name: String, resourceBundle: String): Logger
+
+  protected def findParent(logger: Logger): Option[Logger] = {
+    def go(s: List[String]): Option[Logger] = s match {
+      case Nil                                    =>
+        None
+      case b if loggers.contains(b.mkString(".")) =>
+        loggers.get(b.mkString("."))
+      case b                                      =>
+        go(b.dropRight(1))
```
nicolasstucki: Use the following format:
```scala
def go(s: List[String]): Option[Logger] = s match {
  case Nil =>
    None

  case b if loggers.contains(b.mkString(".")) =>
    loggers.get(b.mkString("."))

  case b =>
    go(b.dropRight(1))
}
```

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+    if (name == null)
+      throw new NullPointerException("Logger name cannot be null")
+    loggers.getOrElseUpdate(name,
+      new Logger(name, None, None, true, Some(rootLogger)))
+  }
+
+  // Not implemented, no resource bundle in scala.js
+  //def getLogger(name: String, resourceBundle: String): Logger
+
+  protected def findParent(logger: Logger): Option[Logger] = {
```
nicolasstucki: THis should be `private`.

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+        None
+      case b if loggers.contains(b.mkString(".")) =>
+        loggers.get(b.mkString("."))
+      case b                                      =>
+        go(b.dropRight(1))
+    }
+    go(Option(logger.getName).getOrElse("").split("\\.").toList.dropRight(1))
+  }
+
+  def getAnonymousLogger():Logger =
```
nicolasstucki: Add the `{ }` because the comments makes it a 2-liner.

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+    this(name, Option(resourceBundle), None, true)
+
+  // Not implemented, no resource bundle
+  //def getResourceBundle():ResourceBundle = ???
+
+  // Find the effective level
+  private def levelR: Option[Level] = {
+    @tailrec
+    def go(logger: Logger): Option[Level] = logger.level match {
+      case l @ Some(_)                 => l
```
nicolasstucki: You should not have access to `level`. Consider:
```scala
private def levelR: Level = {
  @tailrec 
  def go(logger: Logger): Level = {
    if (logger.level != null) logger.level
    else if (getParent() != null) go(logger.getParent)
    else null
  }
  go(this)
}
```

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    go(this)
+  }
+
+  def getResourceBundleName():String = resourceBundle.orNull
+
+  def setFilter(filter: Filter): Unit = this.filter = Option(filter)
+
+  def getFilter(): Filter = filter.orNull
```
nicolasstucki: Should avoid boxing `filter`:
```scala
 private[this] var filter: Filter = _
 def setFilter(filter: Filter): Unit = this.filter = filter
 def getFilter(): Filter = filter
```

This also applies to other fields.

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,252 @@
...
+  def logp(level: Level, sourceClass: String, sourceMethod: String,
+      msg: String, params: Array[AnyRef]): Unit =
+    log(new LogRecord(level, msg,
+      sourceClassName = Option(sourceClass),
+      sourceMethodName = Option(sourceMethod),
+      params = params.toList,
+      loggerName = Option(name)))
+
+  def logp(level: Level, sourceClass: String, sourceMethod: String,
+      msg: String, thrown: Throwable): Unit =
```
nicolasstucki: Add `{ }` to all the methods that have a body of 2 lines or more (exception with `if/else` expressions).

javalib/src/main/scala/java/util/logging/ConsoleHandler.scala
cquiroz: There is no `LogManager` in this port. I take that as that the properties are not defined so we use the default values. That's why I'm defaulting to `new SimpleFormatter`. It is not clear to me what's are you pointing wrong in your comment

javalib/src/main/scala/java/util/logging/Logger.scala
cquiroz: The reason for the boxing is that I prefer not to have nulls in my code. If keeping the nulls is the preferred way I'll change that

javalib/src/main/scala/java/util/logging/SimpleFormatter.scala
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Date
+
+class SimpleFormatter extends Formatter {
+  // The default format is implementation specific
+  private val defaultFormat = "[%4$s] %1s - %3$s - %5$s"
+
+  def format(record: LogRecord): String = {
+    // As per spec we check the property or use a default
+    val fmt = Option(
+        System.getProperty("java.util.logging.SimpleFormatter.format"))
```
nicolasstucki:
```scala
val fmt = System.getProperty("java.util.logging.SimpleFormatter.format", defaultFormat)

File: javalib/src/main/scala/java/util/logging/StreamHandler.scala [hunk 24]
```diff
@@ -0,0 +1,88 @@
+package java.util.logging
+
+import java.io.OutputStream
+
+class StreamHandler(private[this] val _out: OutputStream,
```
nicolasstucki: Why the `private[this] val`?

javalib/src/main/scala/java/util/logging/StreamHandler.scala
```diff
@@ -0,0 +1,88 @@
...
+    private[this] val _formatter: Formatter) extends Handler {
+
+  // Required by javadoc but it is unspecified what to do if formatter is null
+  def this() = this(null, null)
+
+  protected var out:Option[OutputStream] = Option(_out)
```
nicolasstucki: This should not be `protected`, it is not in the documentation. Consider not boxing the `_out`.

nicolasstucki: You are also duplicating the fields that you need to keep one reference.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/HandlerTest.scala
```diff
@@ -0,0 +1,40 @@
...
+  class TestHandler extends Handler {
+    override def flush(): Unit = {}
+
+    override def publish(record: LogRecord): Unit = {}
+
+    override def close(): Unit = {}
+  }
+
+  class TestFilter extends Filter {
+    override def isLoggable(record: LogRecord): Boolean = true
```
nicolasstucki: Should also test when it is not loggable, I suggest adding a `logable: Boolean` parameter to the constructor and writing some tests for when it is false.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/HandlerTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    assertNotNull(h.getFilter)
+  }
+
+  @Test def test_handler_loggable(): Unit = {
+    val h = new TestHandler()
+    assertTrue(h.isLoggable(new LogRecord(Level.INFO, "message")))
+    h.setLevel(Level.SEVERE)
+    assertFalse(h.isLoggable(new LogRecord(Level.INFO, "message")))
+  }
+}
```
nicolasstucki: You should add tests for `getEncoding`, `reportError`, `setLevel`, `getLevel`, `isLoggable`...

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LevelTest.scala
```diff
@@ -0,0 +1,44 @@
...
+    assertEquals(300, Level.FINEST.intValue())
+
+    assertEquals("ALL", Level.ALL.getName)
+    assertEquals(Int.MinValue, Level.ALL.intValue())
+  }
+
+  @Test def test_equals_hash_code(): Unit = {
+    assertEquals(Level.SEVERE, Level.SEVERE)
+    assertNotEquals(Level.SEVERE, Level.WARNING)
+    assertEquals(Level.SEVERE.hashCode(), Level.SEVERE.hashCode())
```
nicolasstucki: Should also add some `assertSame` tests.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LogRecordTest.scala
```diff
@@ -0,0 +1,55 @@
...
+    assertEquals(Thread.currentThread().getId, record.getThreadID.toLong)
+    record.setThreadID(5)
+    assertEquals(5L, record.getThreadID.toLong)
+
+    assertTrue(record.getMillis <= System.currentTimeMillis())
+    record.setMillis(1)
+    assertEquals(1L, record.getMillis)
+
+    assertNull(record.getThrown)
+    record.setThrown(new RuntimeException("testmsg"))
```
nicolasstucki: Test `record.setThrown(null)`

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LogRecordTest.scala
```diff
@@ -0,0 +1,55 @@
...
+    assertEquals("msg", record.getMessage)
+    record.setMessage("newmsg")
+    assertEquals("newmsg", record.getMessage)
+    // You can set message to null
+    record.setMessage(null)
+    assertNull(record.getMessage)
+
+    assertNull(record.getParameters)
+    record.setParameters(Array[AnyRef]("value"))
+    assertArrayEquals(Array[AnyRef]("value"), record.getParameters)
```
nicolasstucki: Test 
```scala
val arr = Array[AnyRef]("value")
record.setParameters(arr)
arr(0) = "other value"
assertArrayEquals(Array[AnyRef]("other value"), record.getParameters)
```

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LoggerTest.scala
```diff
@@ -0,0 +1,362 @@
...
+    assertNotNull(logger.getParent)
+  }
+
+  @Test def test_static(): Unit = {
+    assertEquals("global", Logger.GLOBAL_LOGGER_NAME)
+  }
+
+  @Test def test_get_logger(): Unit = {
+    assertEquals(s"$Prefix.TestLogger",
+      Logger.getLogger(s"$Prefix.TestLogger").getName)
```
nicolasstucki: 4 spaces for parameters in a new line.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LoggerTest.scala
```diff
@@ -0,0 +1,362 @@
...
+
+import java.util.logging._
+
+import org.junit.{Ignore, Test}
+import org.junit.Assert._
+
+class LoggerTest {
+  // We add a prefix to all loggers. This avoids some errors when running tests
+  // more than once as the loggers are global
+  val Prefix = System.currentTimeMillis().toString
```
nicolasstucki: Use lowercase for the first letter of `val`/`var`/`def` members.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LoggerTest.scala
```diff
@@ -0,0 +1,362 @@
...
+
+    logger.log(Level.FINE, "Log")
+    assertEquals(Some(Level.FINE), testHandler.lastRecord.map(_.getLevel))
+    assertEquals(Some("Log"), testHandler.lastRecord.map(_.getMessage))
+
+    logger.log(Level.FINE, "Log", "param")
+    assertEquals(Some(Level.FINE), testHandler.lastRecord.map(_.getLevel))
+    assertEquals(Some("Log"), testHandler.lastRecord.map(_.getMessage))
+    assertArrayEquals(Array[AnyRef]("param"),
+      testHandler.lastRecord.map(_.getParameters).getOrElse(Array.empty))
```
nicolasstucki: 4 spaces

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LoggerTest.scala
```diff
@@ -0,0 +1,362 @@
...
+    val logger = Logger.getLogger(s"$Prefix.TestLogger6")
+    val testHandler = new TestHandler
+    logger.setLevel(Level.ALL)
+    logger.addHandler(testHandler)
+    logger.setUseParentHandlers(false)
+
+    val lrFine = new LogRecord(Level.FINE, "Log")
+    logger.log(lrFine)
+    assertEquals(Some(Level.FINE), testHandler.lastRecord.map(_.getLevel))
+    assertEquals(Some("Log"), testHandler.lastRecord.map(_.getMessage))
```
nicolasstucki: Consider changing `var lastRecord: LogRecord`, `assertEquals("Log", testHandler.lastRecord.getMessage)` and assume that in the assets they are non null. If they happens to be null then the test will fail with the NPE that will have more information about the source of the problem.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/LoggerTest.scala
```diff
@@ -0,0 +1,362 @@
...
+      testHandler.lastRecord.map(_.getSourceMethodName))
+    assertArrayEquals(Array[AnyRef]("param"),
+      testHandler.lastRecord.map(_.getParameters).getOrElse(Array.empty))
+
+    logger.entering("cls", "method", Array[AnyRef]("param1", "param2"))
+    assertEquals(Some(Level.FINER), testHandler.lastRecord.map(_.getLevel))
+    assertEquals(Some("ENTRY {0} {1}"), testHandler.lastRecord.map(_.getMessage))
+    assertEquals(Some("cls"), testHandler.lastRecord.map(_.getSourceClassName))
+    assertEquals(Some("method"),
+      testHandler.lastRecord.map(_.getSourceMethodName))
```
nicolasstucki: 4 space and many more occurrences.

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/StreamHandlerTest.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.ByteArrayOutputStream
+import java.util.logging._
+
+import org.junit.Test
+import org.junit.Assert._
+
+class StreamHandlerTest {
+  object TestFormatter extends SimpleFormatter {
+
```
nicolasstucki: remove this line

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/StreamHandlerTest.scala
```diff
@@ -0,0 +1,60 @@
...
+    override def getHead(h: Handler): String = "header"
+
+    override def getTail(h: Handler): String = "footer"
+
```
nicolasstucki: and this line

test-suite/shared/src/test/require-jdk7/org/scalajs/testsuite/javalib/util/logging/SimpleFormatterTest.scala
```diff
@@ -0,0 +1,50 @@
...
+    assertTrue(f.format(r).contains(msg))
+
+    val r1 = new LogRecord(Level.INFO, msg)
+    r1.setParameters(Array("param1"))
+    assertTrue(f.format(r1).contains("message with params param1 {1}"))
+
+    val r2 = new LogRecord(Level.INFO, msg)
+    r2.setParameters(Array("param1", new java.lang.Integer(20)))
+    assertTrue(f.format(r2).contains("message with params param1 20"))
+
```
nicolasstucki: remove this line

nicolasstucki:
That is all.

Given that this is a Java library I highly suggest to avoid using `None` as a replacement for `null` and `Some(ref)` for non null references. Just use the references directly and handle `null`s like they would in java libraries, this will be more performant and will help us fail with NPE in the same places it would in the JVM.

cquiroz:
@nicolasstucki Thanks for the thorough review. I'll update the code with your suggestions. Perhaps on the contributing documentation it should be mention to avoid `Option`. It just feels the natural way to handle `null`

nicolasstucki:
The `Option`s thing is more related to the internal logic of the java classes. In general we would prefer using nice abstraction like `Option`.

cquiroz:
I integrated your comments from the PR and added more test coverage as requested

sjrd:

javalib/src/main/scala/java/util/logging/Logger.scala
```diff
@@ -0,0 +1,319 @@
...
+      else if (logger.getParent == null) null
+      else go(logger.getParent)
+    }
+
+    go(this)
+  }
+
+  // Not implemented, no resource bundle
+  //def getResourceBundle():ResourceBundle = ???
+
```
sjrd: CI says: Whitespace at end of line.

cquiroz:
Fixed whitespace

nicolasstucki:
Jenkins, please retest this

sjrd:
The correct incantation is: retest this please

javalib/src/main/scala/java/util/logging/ConsoleHandler.scala
nicolasstucki: Then ignore this.

nicolasstucki:
LGTM.
@cquiroz could you squash the commits into one (`Implement parts of java.util.logging.`)

cquiroz:
@nicolasstucki Thanks. Should it also be rebased onto master?

nicolasstucki:
Yes, rebase it.

sjrd:
I still think we should move this to a separate repository, like scalajs-java-time.
@nicolasstucki Could you set up such a repo with an initial commit? It should be easy to copy-paste the infrasture of scalajs-java-time. Then @cquiroz could target a PR against that repo.

nicolasstucki:
Ok, I will create the new repository.

nicolasstucki:
@cquiroz could you make a PR with this code on https://github.com/scala-js/scala-js-java-logging

sjrd:
Superseded by https://github.com/scala-js/scala-js-java-logging/pull/2

PR #2411, opened 2016-05-23
DO NOT MERGE: Fix #2066 : Improve performance of isInstanceOf

### LINKED ISSUE #2066: Optimize isInstanceOf on Scala objects
Hi,

Thanks

paullepoulpe:
Waiting for @tOverney's changes to the IncEmitter to fix incrementality (see discussion on #2142 ).

### DIFF
--- a/tools/scalajsenv.js
+++ b/tools/scalajsenv.js
@@ -144,13 +144,27 @@ const $idHashCodeMap = ScalaJS.g["WeakMap"] ? new ScalaJS.g["WeakMap"]() : null;
 //!endif
 //!endif
 
+ScalaJS.ClassData = {};
+
 // Core mechanism
 
-ScalaJS.makeIsArrayOfPrimitive = function(primitiveData) {
+ScalaJS.expandSubtypeArray = function(tags){
+  let expanded = [];
+  let len = 0;
+  for (let i = 0; i < tags.length; i += 2) {
+    const start = tags[i];
+    const end = tags[i + 1];
+    while(len < start){ len = expanded["push"](0) };
+    while(len <= end){ len = expanded["push"](1) };
+  }
+  return expanded;
+};
+
+ScalaJS.makeIsArrayOfPrimitive = function(primitiveTag) {
   return function(obj, depth) {
-    return !!(obj && obj.$classData &&
-      (obj.$classData.arrayDepth === depth) &&
-      (obj.$classData.arrayBase === primitiveData));
+     return !!(obj && (obj.$typeTag < 0) &&
+        (((obj.$typeTag >>> 23) & 255) === depth) &&
+        ((obj.$typeTag & 8388607) === primitiveTag));
   }
 };
 
@@ -165,24 +179,10 @@ ScalaJS.makeAsArrayOfPrimitive = function(isInstanceOfFunction, arrayEncodedName
 };
 //!endif
 
-/** Encode a property name for runtime manipulation
-  *  Usage:
-  *    env.propertyName({someProp:0})
-  *  Returns:
-  *    "someProp"
-  *  Useful when the property is renamed by a global optimizer (like Closure)
-  *  but we must still get hold of a string of that name for runtime
-  * reflection.
-  */
-ScalaJS.propertyName = function(obj) {
-  for (const prop in obj)
-    return prop;
-};
-
 // Runtime functions
 
 ScalaJS.isScalaJSObject = function(obj) {
-  return !!(obj && obj.$classData);
+  return !!(obj && obj.$typeTag);
 };
 
 //!if asInstanceOfs != Unchecked
@@ -270,7 +270,7 @@ ScalaJS.objectGetClass = function(instance) {
       else if (ScalaJS.is.sjsr_RuntimeLong(instance))
         return ScalaJS.d.jl_Long.getClassOf();
       else if (ScalaJS.isScalaJSObject(instance))
-        return instance.$classData.getClassOf();
+        return ScalaJS.ClassData[instance.$typeTag].getClassOf();
       else
         return null; // Exception?
   }
@@ -739,9 +739,10 @@ ScalaJS.TypeData.prototype.initPrim = function(
 //!else
 initPrim(
 //!endif
-    zero, arrayEncodedName, displayName) {
+    zero, arrayEncodedName, displayName, typeTag) {
   // Runtime support
-  this.ancestors = {};
+  this.ancestors = [];
+  this.typeTag = typeTag;
   this.componentData = null;
   this.zero = zero;
   this.arrayEncodedName = arrayEncodedName;
@@ -760,22 +761,23 @@ ScalaJS.TypeData.prototype.initClass = function(
 //!else
 initClass(
 //!endif
-    internalNameObj, isInterface, fullName,
-    ancestors, isRawJSType, parentData, isInstance, isArrayOf) {
-  const internalName = ScalaJS.propertyName(internalNameObj);
-
+    isInterface, fullName,
+    ancestors, typeTag, isRawJSType, parentData, isInstance, isArrayOf) {
   isInstance = isInstance || function(obj) {
-    return !!(obj && obj.$classData && obj.$classData.ancestors[internalName]);
+    return ScalaJS.isScalaJSObject(obj) &&
+      (ScalaJS.ClassData[obj.$typeTag].ancestors["indexOf"](typeTag) >= 0);
   };
 
   isArrayOf = isArrayOf || function(obj, depth) {
-    return !!(obj && obj.$classData && (obj.$classData.arrayDepth === depth)
-      && obj.$classData.arrayBase.ancestors[internalName])
+    const data = ScalaJS.isScalaJSObject(obj) ? ScalaJS.ClassData[obj.$typeTag] : null;
+    return !!(data && (data.arrayDepth === depth)
+      && (data.arrayBase.ancestors["indexOf"](typeTag) >= 0));
   };
 
   // Runtime support
   this.parentData = parentData;
   this.ancestors = ancestors;
+  this.typeTag = typeTag;
   this.arrayEncodedName = "L"+fullName+";";
   this.isArrayOf = isArrayOf;
 
@@ -853,8 +855,6 @@ initArray(
   };
 //!endif
 
-  ArrayClass.prototype.$classData = this;
-
   // Don't generate reflective call proxies. The compiler special cases
   // reflective calls to methods on scala.Array
 
@@ -864,6 +864,12 @@ initArray(
   const componentBase = componentData.arrayBase || componentData;
   const arrayDepth = componentData.arrayDepth + 1;
 
+  const typeTag = (((1 << 8) | arrayDepth) << 23) | componentBase.typeTag;
+
+  ArrayClass.prototype.$typeTag = typeTag;
+  ScalaJS.ClassData[typeTag] = this;
+  this.typeTag = typeTag;
+
   const isInstance = function(obj) {
     return componentBase.isArrayOf(obj, arrayDepth);
   }
@@ -871,7 +877,7 @@ initArray(
   // Runtime support
   this.constr = ArrayClass;
   this.parentData = ScalaJS.d.O;
-  this.ancestors = {O: 1, jl_Cloneable: 1, Ljava_io_Serializable: 1};
+  this.ancestors = [9, 10, 11]; // O, jl_Cloneable, Ljava_io_Serializable
   this.componentData = componentData;
   this.arrayBase = componentBase;
   this.arrayDepth = arrayDepth;
@@ -933,7 +939,7 @@ ScalaJS.TypeData.prototype["getFakeInstance"] = function() {
   else if (this === ScalaJS.d.sr_BoxedUnit)
     return void 0;
   else
-    return {$classData: this};
+    return {$typeTag: this.typeTag};
 };
 
 //!if outputMode != ECMAScript6
@@ -968,40 +974,40 @@ ScalaJS.TypeData.prototype["newArrayOfThisClass"] = function(lengths) {
 
 // Create primitive types
 
-ScalaJS.d.V = new ScalaJS.TypeData().initPrim(undefined, "V", "void");
-ScalaJS.d.Z = new ScalaJS.TypeData().initPrim(false, "Z", "boolean");
-ScalaJS.d.C = new ScalaJS.TypeData().initPrim(0, "C", "char");
-ScalaJS.d.B = new ScalaJS.TypeData().initPrim(0, "B", "byte");
-ScalaJS.d.S = new ScalaJS.TypeData().initPrim(0, "S", "short");
-ScalaJS.d.I = new ScalaJS.TypeData().initPrim(0, "I", "int");
-ScalaJS.d.J = new ScalaJS.TypeData().initPrim("longZero", "J", "long");
-ScalaJS.d.F = new ScalaJS.TypeData().initPrim(0.0, "F", "float");
-ScalaJS.d.D = new ScalaJS.TypeData().initPrim(0.0, "D", "double");
+ScalaJS.d.V = new ScalaJS.TypeData().initPrim(undefined, "V", "void", 0);
+ScalaJS.d.Z = new ScalaJS.TypeData().initPrim(false, "Z", "boolean", 1);
+ScalaJS.d.C = new ScalaJS.TypeData().initPrim(0, "C", "char", 2);
+ScalaJS.d.B = new ScalaJS.TypeData().initPrim(0, "B", "byte", 3);
+ScalaJS.d.S = new ScalaJS.TypeData().initPrim(0, "S", "short", 4);
+ScalaJS.d.I = new ScalaJS.TypeData().initPrim(0, "I", "int", 5);
+ScalaJS.d.J = new ScalaJS.TypeData().initPrim("longZero", "J", "long", 6);
+ScalaJS.d.F = new ScalaJS.TypeData().initPrim(0.0, "F", "float", 7);
+ScalaJS.d.D = new ScalaJS.TypeData().initPrim(0.0, "D", "double", 8);
 
 // Instance tests for array of primitives
 
-ScalaJS.isArrayOf.Z = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.Z);
+ScalaJS.isArrayOf.Z = ScalaJS.makeIsArrayOfPrimitive(1);
 ScalaJS.d.Z.isArrayOf = ScalaJS.isArrayOf.Z;
 
-ScalaJS.isArrayOf.C = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.C);
+ScalaJS.isArrayOf.C = ScalaJS.makeIsArrayOfPrimitive(2);
 ScalaJS.d.C.isArrayOf = ScalaJS.isArrayOf.C;
 
-ScalaJS.isArrayOf.B = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.B);
+ScalaJS.isArrayOf.B = ScalaJS.makeIsArrayOfPrimitive(3);
 ScalaJS.d.B.isArrayOf = ScalaJS.isArrayOf.B;
 
-ScalaJS.isArrayOf.S = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.S);
+ScalaJS.isArrayOf.S = ScalaJS.makeIsArrayOfPrimitive(4);
 ScalaJS.d.S.isArrayOf = ScalaJS.isArrayOf.S;
 
-ScalaJS.isArrayOf.I = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.I);
+ScalaJS.isArrayOf.I = ScalaJS.makeIsArrayOfPrimitive(5);
 ScalaJS.d.I.isArrayOf = ScalaJS.isArrayOf.I;
 
-ScalaJS.isArrayOf.J = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.J);
+ScalaJS.isArrayOf.J = ScalaJS.makeIsArrayOfPrimitive(6);
 ScalaJS.d.J.isArrayOf = ScalaJS.isArrayOf.J;
 
-ScalaJS.isArrayOf.F = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.F);
+ScalaJS.isArrayOf.F = ScalaJS.makeIsArrayOfPrimitive(7);
 ScalaJS.d.F.isArrayOf = ScalaJS.isArrayOf.F;
 
-ScalaJS.isArrayOf.D = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.D);
+ScalaJS.isArrayOf.D = ScalaJS.makeIsArrayOfPrimitive(8);
 ScalaJS.d.D.isArrayOf = ScalaJS.isArrayOf.D;
 
 //!if asInstanceOfs != Unchecked
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
@@ -185,6 +185,9 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
           classEmitter.genInstanceTests(linkedClass),
           classEmitter.genArrayInstanceTests(linkedClass)
       )(linkedClass.pos)))
+      if (classEmitter.needsSubtypeArray(className)) {
+        addTree(classEmitter.genSubtypeArray(linkedClass))
+      }
     }
 
     if (linkedClass.hasRuntimeTypeInfo) {
@@ -192,9 +195,11 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
           classEmitter.genTypeData(linkedClass)))
     }
 
-    if (linkedClass.hasInstances && kind.isClass && linkedClass.hasRuntimeTypeInfo)
+    if (linkedClass.hasInstances && kind.isClass && linkedClass.hasRuntimeTypeInfo) {
       addTree(classTreeCache.setTypeData.getOrElseUpdate(
           classEmitter.genSetTypeData(linkedClass)))
+      addTree(classEmitter.genSetTypeTag(linkedClass))
+    }
 
     if (linkedClass.kind.hasModuleAccessor)
       addTree(classTreeCache.moduleAccessor.getOrElseUpdate(
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
@@ -9,11 +9,14 @@
 
 package org.scalajs.core.tools.linker.backend.emitter
 
+import scala.annotation.tailrec
+
 import org.scalajs.core.ir._
 import Position._
 import Transformers._
 import org.scalajs.core.ir.Trees._
 import Types._
+import Definitions._
 
 import org.scalajs.core.tools.sem._
 import CheckedBehavior.Unchecked
@@ -72,6 +75,74 @@ private[scalajs] final class ScalaJSClassEmitter(
 
   private[emitter] def semantics: Semantics = linkingUnit.semantics
 
+  private[emitter] final val ReservedTags = Map("V" -> 0, "Z" -> 1, "C" -> 2,
+      "B" -> 3, "S" -> 4, "I" -> 5, "J" -> 6, "F" -> 7, "D" -> 8,
+      ObjectClass -> 9, CloneableClass -> 10, SerializableClass -> 11)
+
+  private[emitter] final val MaxPrimitiveTag = 8
+  private[emitter] final val MaxReservedTag = 11
+  private[emitter] final val MaxTagValue = (1 << 23) - 1
+
+  private lazy val typeTags: Map[LinkedClass, Int] = {
+    val (nonRoots, roots) = linkingUnit.classDefs.partition { cls =>
+      cls.superClass.exists(parent => linkedClassByName.contains(parent.name))
+    }
+
+    val childrenMap =
+      nonRoots.groupBy(_.superClass.get.name).withDefaultValue(Nil)
+
+    val tags = Map.newBuilder[LinkedClass, Int]
+    var curTag = MaxReservedTag + 1
+
+    def computeTypeTag(cls: LinkedClass): Unit = {
+      if (ReservedTags.contains(cls.name.name)) {
+        tags += cls -> ReservedTags(cls.name.name)
+      } else {
+        tags += cls -> curTag
+        curTag += 1
+      }
+      childrenMap(cls.name.name).sortBy(_.name.name).foreach(computeTypeTag)
+    }
+
+    roots.sortBy(_.name.name).foreach(computeTypeTag)
+
+    val result = tags.result
+
+    // Sanity check: all classDefs must have a tag
+    assert(linkingUnit.classDefs.forall(result.contains))
+    result
+  }
+
+  private lazy val subtypeIntervals: Map[String, List[(Int, Int)]] = {
+    val parentToChilds = for {
+      child <- typeTags.keys
+      parent <- child.ancestors
+    } yield {
+      parent -> child
+    }
+
+    parentToChilds.groupBy(_._1).map {
+      case (parent, parentChilds) =>
+        val tags = parentChilds.map(t => typeTags(t._2)).toList.sorted
+        val intervals = tags.foldLeft[List[(Int, Int)]](Nil) {
+          case ((start, end) :: rest, tag) if tag == end + 1 =>
+            (start, tag) :: rest
+          case (intervals, tag) =>
+            (tag, tag) :: intervals
+        }.reverse
+        parent -> intervals
+    }
+  }
+
+  private[emitter] def needsSubtypeArray(className: String): Boolean = {
+    val numComparaisons = subtypeIntervals(className).map {
+      case (start, end) if start == end => 1
+      case _                            => 2
+    }.sum
+
+    numComparaisons > 3
+  }
+
   private implicit def implicitOutputMode: OutputMode = outputMode
 
   def genDeclareTypeData(tree: LinkedClass): js.Tree = {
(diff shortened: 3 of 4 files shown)


paullepoulpe:
Comments on the rest are welcome though

sjrd:
Jenkins, ok to test

sjrd:
Nothing works with Rhino ;-) You'll probably have to tweak [the Rhino JS env](https://github.com/scala-js/scala-js/tree/master/js-envs/src/main/scala/org/scalajs/jsenv/rhino), because it interacts with the internals of the `Emitter`.

tools/scalajsenv.js
```diff
@@ -144,13 +144,28 @@ const $idHashCodeMap = ScalaJS.g["WeakMap"] ? new ScalaJS.g["WeakMap"]() : null;
...
 // Core mechanism
 
-ScalaJS.makeIsArrayOfPrimitive = function(primitiveData) {
+ScalaJS.expandSubtypeArray = function(tags){
+  let expanded = [];
+  let len = 0;
+  for (let i = 0; i < tags.length; i++) {
+    let interval = tags[i];
+    let start = interval[0];
+    let end = interval[1];
```
sjrd: You can win a bit of space (and possibly memory usage) by flattening `tags`. Instead of having
```js
tags = [[3, 6], [15, 17]]
```
you could have
```js
tags = [3, 6, 15, 17]
```
Less brackets. Then the loop becomes:
```js
for (let i = 0; i < tags.length; i += 2) {
  const start = tags[i];
  const end = tags[i + 1];
  ...
}
```
(btw, use `const`s where possible)

tools/scalajsenv.js
```diff
@@ -761,21 +777,24 @@ ScalaJS.TypeData.prototype.initClass = function(
 initClass(
 //!endif
     internalNameObj, isInterface, fullName,
-    ancestors, isRawJSType, parentData, isInstance, isArrayOf) {
+    ancestors, typeTag, isRawJSType, parentData, isInstance, isArrayOf) {
   const internalName = ScalaJS.propertyName(internalNameObj);
```
sjrd: `internalName` and, consequently, `internalNameObj`, are not needed anymore. They should be removed. :p

sjrd: Oh, and `ScalaJS.propertyName` can be removed too.

tools/scalajsenv.js
```diff
@@ -761,21 +777,24 @@ ScalaJS.TypeData.prototype.initClass = function(
...
 
   isInstance = isInstance || function(obj) {
-    return !!(obj && obj.$classData && obj.$classData.ancestors[internalName]);
+    let data = obj && ScalaJS.ClassData[obj.$typeTag];
```
sjrd: This is awkward if `$typeTag` does not exist on `obj` (happens for primitives and JS objects). It would be good to use
```js
ScalaJS.isScalaJSObject(obj) ? ScalaJS.ClassData[obj.$typeTag] : null;
```

tools/scalajsenv.js
```diff
@@ -761,21 +777,24 @@ ScalaJS.TypeData.prototype.initClass = function(
...
+    return !!(data && (data.ancestors["indexOf"](typeTag) >= 0))
```
sjrd: In fact, you can also rewrite the two lines as
```js
return ScalaJS.isScalaJSObject(obj) &&
  (ScalaJS.ClassData[obj.$typeTag].ancestors["indexOf"](typeTag) >= 0);
```

tools/scalajsenv.js
```diff
@@ -761,21 +777,24 @@ ScalaJS.TypeData.prototype.initClass = function(
...
   };
 
   isArrayOf = isArrayOf || function(obj, depth) {
-    return !!(obj && obj.$classData && (obj.$classData.arrayDepth === depth)
-      && obj.$classData.arrayBase.ancestors[internalName])
+    let data = obj && ScalaJS.ClassData[obj.$typeTag];
```
sjrd: Same remark as above on how to get `data` (although here you can't rewrite as a "one-liner", I think)

tools/scalajsenv.js
```diff
@@ -864,14 +881,21 @@ initArray(
   const componentBase = componentData.arrayBase || componentData;
   const arrayDepth = componentData.arrayDepth + 1;
 
+  const typeTag = (((1 << 8) | arrayDepth) << 23) | componentBase.typeTag;
+
+  ArrayClass.prototype.$typeTag = typeTag;
+  ScalaJS.ClassData[typeTag] = this;
```
sjrd: Could we do away without this line? In other words, can we avoid having to store the `TypeData` of array types in `ScalaJS.ClassData`?
My concern is that this will turn `ClassTag` into a sparse array, considerably hurting its performance (access is not constant-time anymore).

tools/scalajsenv.js
```diff
@@ -864,14 +881,21 @@ initArray(
...
   const isInstance = function(obj) {
     return componentBase.isArrayOf(obj, arrayDepth);
   }
 
   // Runtime support
   this.constr = ArrayClass;
   this.parentData = ScalaJS.d.O;
-  this.ancestors = {O: 1, jl_Cloneable: 1, Ljava_io_Serializable: 1};
+  // O, jl_Cloneable, Ljava_io_Serializable
+  this.ancestors = [9, 10, 11];
```
sjrd: Can you fit the comment at the end of the line and stay under 80 characters?

tools/scalajsenv.js
```diff
@@ -968,40 +992,40 @@ ScalaJS.TypeData.prototype["newArrayOfThisClass"] = function(lengths) {
...
+ScalaJS.d.S = new ScalaJS.TypeData().initPrim(0, "S", "short", 4);
+ScalaJS.d.I = new ScalaJS.TypeData().initPrim(0, "I", "int", 5);
+ScalaJS.d.J = new ScalaJS.TypeData().initPrim("longZero", "J", "long", 6);
+ScalaJS.d.F = new ScalaJS.TypeData().initPrim(0.0, "F", "float", 7);
+ScalaJS.d.D = new ScalaJS.TypeData().initPrim(0.0, "D", "double", 8);
 
 // Instance tests for array of primitives
 
-ScalaJS.isArrayOf.Z = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.Z);
+ScalaJS.isArrayOf.Z = ScalaJS.makeIsArrayOfPrimitive(ScalaJS.d.Z.typeTag);
```
sjrd: Consider hard-coding the constants here. Hopefully this could help GCC to dce away `ScalaJS.d.V` (and similar) if the data are not accessed by the `isArrayOf`s are needed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -185,16 +185,21 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
           classEmitter.genInstanceTests(linkedClass),
           classEmitter.genArrayInstanceTests(linkedClass)
       )(linkedClass.pos)))
+      if(classEmitter.needsSubtypeArray(className)){
+        addTree(classEmitter.genSubtypeArray(linkedClass))
+      }
```
sjrd: This change must be accompanied by the equivalent change in [`ScalaJSClassEmitter.genClassDef`](https://github.com/scala-js/scala-js/blob/v0.6.9/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala#L96).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -185,16 +185,21 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
     if (linkedClass.hasRuntimeTypeInfo) {
       addTree(classTreeCache.typeData.getOrElseUpdate(
           classEmitter.genTypeData(linkedClass)))
     }
 
-    if (linkedClass.hasInstances && kind.isClass && linkedClass.hasRuntimeTypeInfo)
+    if (linkedClass.hasInstances && kind.isClass && linkedClass.hasRuntimeTypeInfo){
       addTree(classTreeCache.setTypeData.getOrElseUpdate(
           classEmitter.genSetTypeData(linkedClass)))
+      addTree(classEmitter.genSetTypeTag(linkedClass))
```
sjrd: Same thing here.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -745,14 +858,27 @@ private[scalajs] final class ScalaJSClassEmitter(
...
     import TreeDSL._
 
     implicit val pos = tree.pos
+    val tag = typeTags(tree)
 
     assert(tree.kind.isClass)
 
-    encodeClassVar(tree.name.name).prototype DOT "$classData" :=
+    js.BracketSelect(envField("ClassData"), js.IntLiteral(tag)) :=
       envField("d", tree.name.name)
```
sjrd: It's probably best to keep it the way you did in this PR. But I'm noting here to remember that we should try to assign the entire `ClassData` array in one go as an array literal, rather than building it up like this. This would be smaller.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -745,14 +858,27 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+  def genSetTypeTag(tree: LinkedClass): js.Tree = {
+    import TreeDSL._
+
+    implicit val pos = tree.pos
+
+    assert(tree.kind.isClass)
+    val tag = typeTags(tree)
+
+    encodeClassVar(tree.name.name).prototype DOT "$typeTag" :=
+        js.IntLiteral(tag)
```
sjrd: 2 spaces only

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -654,6 +743,25 @@ private[scalajs] final class ScalaJSClassEmitter(
     ancestors DOT className
   }
 
+  def genSubtypeArray(tree: LinkedClass): js.Tree = {
+    import Definitions._
+    import TreeDSL._
```
sjrd: I think these two imports are useless.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -616,12 +705,12 @@ private[scalajs] final class ScalaJSClassEmitter(
...
-              genIsScalaJSObject(obj) &&
-              ((obj DOT "$classData" DOT "arrayDepth") === depth) &&
-              genIsClassNameInAncestors(className,
-                  obj DOT "$classData" DOT "arrayBase" DOT "ancestors")
-            })))
+            val tagVar = genLet(js.Ident("tag"), mutable = false, obj && (obj DOT "$typeTag"))
+            val tag = tagVar.ref
+            val test = js.BinaryOp(JSBinaryOp.<, tag, js.IntLiteral(0)) &&
+              (((tag >> 23) & 255) === depth) &&
+              genIntervalsTest(className, tag & 8388607)
```
sjrd: This magic constant 8388607 deserves a `final val` somewhere.

Also, the coding style for multi-line `&&` chains would be:
```scala
val test = {
  js.BinaryOp(JSBinaryOp.<, tag, js.IntLiteral(0)) &&
  (((tag >> 23) & 255) === depth) &&
  genIntervalsTest(className, tag & 8388607)
}
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -590,7 +679,7 @@ private[scalajs] final class ScalaJSClassEmitter(
         js.Function(List(objParam, depthParam), className match {
           case Definitions.ObjectClass =>
             val dataVarDef = genLet(Ident("data"), mutable = false, {
-              obj && (obj DOT "$classData")
+              obj && js.BracketSelect(envField("ClassData"), obj DOT "$typeTag")
```
sjrd: There's no need to go through `ClassData`. You have all the necessary information in `obj DOT "$typeTag"`. Indeed, all you need is to compute `arrayDepth` and `arrayBaseTag`.

You can replace the currently used expression `data.arrayBase["isPrimitive"]` by `arrayBaseTag < 9`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -616,12 +705,12 @@ private[scalajs] final class ScalaJSClassEmitter(
...
-              ((obj DOT "$classData" DOT "arrayDepth") === depth) &&
-              genIsClassNameInAncestors(className,
-                  obj DOT "$classData" DOT "arrayBase" DOT "ancestors")
-            })))
+            val tagVar = genLet(js.Ident("tag"), mutable = false, obj && (obj DOT "$typeTag"))
+            val tag = tagVar.ref
+            val test = js.BinaryOp(JSBinaryOp.<, tag, js.IntLiteral(0)) &&
+              (((tag >> 23) & 255) === depth) &&
+              genIntervalsTest(className, tag & 8388607)
+            js.Block(tagVar, js.Return(!(!test)))
```
sjrd: Is `!!` necessary? It seems all the components of `test` are always booleans, aren't they?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -497,30 +584,32 @@ private[scalajs] final class ScalaJSClassEmitter(
...
-              var test = {
-                genIsScalaJSObject(obj) &&
-                genIsClassNameInAncestors(className,
-                    obj DOT "$classData" DOT "ancestors")
-              }
+              val tagVar = genLet(js.Ident("tag"), mutable = false, obj && (obj DOT "$typeTag"))
+              val tag = tagVar.ref
+              var test = genIntervalsTest(className, tag)
 
+              if(isAncestorOfPseudoArrayClass)
```
sjrd: Missing space after `if`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -478,6 +547,24 @@ private[scalajs] final class ScalaJSClassEmitter(
     }
   }
 
+  private[tools] def genIntervalsTest(className: String, tag: js.Tree)
+      (implicit pos: Position): js.Tree = {
```
sjrd: The opening `(` should be at the end of the previous line.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -478,6 +547,24 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    import TreeDSL._
+
+    val intervals = subtypeIntervals(className)
+
+    if(intervals.nonEmpty && !needsSubtypeArray(className)){
```
sjrd: Missing space after `if` and before `{`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -478,6 +547,24 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+      intervals.map {
+        case (a, b) if a == b => tag === js.IntLiteral(a)
+        case (lo, hi) =>
+          js.BinaryOp(JSBinaryOp.>=, tag, js.IntLiteral(lo)) &&
+          js.BinaryOp(JSBinaryOp.<=, tag, js.IntLiteral(hi))
+      } reduce(_ || _)
```
sjrd: `}.reduce` with a `.`. We don't use infix syntax on alphanumeric methods anymore.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -478,6 +547,24 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    } else {
+      js.BracketSelect(envField("Is", className), tag)
```
sjrd: Hum ... What happens when `intervals.isEmpty && !needsSubtypeArray(className)`? This seems like it would try to access an `Is_xyz` that is not declared. What does it mean when `intervals.isEmpty`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -478,6 +547,24 @@ private[scalajs] final class ScalaJSClassEmitter(
...
 
+  private[tools] def genIntervalsTest(className: String, tag: js.Tree)
+      (implicit pos: Position): js.Tree = {
+    import TreeDSL._
+
+    val intervals = subtypeIntervals(className)
+
+    if(intervals.nonEmpty && !needsSubtypeArray(className)){
+      intervals.map {
+        case (a, b) if a == b => tag === js.IntLiteral(a)
```
sjrd: Put the body of the case on its own line, to be consistent with the other case.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
 
   private[emitter] def semantics: Semantics = linkingUnit.semantics
 
+  private[emitter] lazy val typeTags: Map[LinkedClass, Int] = {
```
sjrd: Looks like this and other members can be `private` rather than `private[emitter]`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    import Definitions.ReservedTags
+
+    val childrenMap =
+      linkingUnit.classDefs.groupBy(_.superClass.map(_.name)).withDefaultValue(Nil)
+
+    // Some classes may have parents that have been removed
+    val orphans = for(cls <- linkingUnit.classDefs;
+                    parent <- cls.superClass
+                    if !linkedClassByName.isDefinedAt(parent.name)) yield cls
```
sjrd: Check the coding style of `for` loops: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#for-comprehensions

sjrd: Prefer `contains()` over `isDefinedAt` for `Map`s. `isDefinedAt` implies working with a `PartialFunction`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+
+    val roots = childrenMap(None) ++ orphans
```
sjrd: If you consider that `Object` is in fact an orphan, you can compute all of this with less passes:
```scala
val (nonRoots, roots) = linkingUnit.classDefs.partition { cls =>
  cls.superClass.exists(linkedClassByName.contains(_))
}

val childrenMap =
  nonRoots.groupBy(_.superClass.get.name).withDefaultValue(Nil)
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+        getTypeTags(nextTag, children ++ rest, accu + (cur -> curTag))
+
+      case Nil =>
+        // Sanity check: all classDefs must have a tag
+        assert(linkingUnit.classDefs.forall(accu.isDefinedAt))
+        accu
+    }
+
+    // The smallest tags are reserved
+    getTypeTags(ReservedTags.values.max + 1, roots)
```
sjrd: Consider defining a hard-coded `final val MaxReservedTag` where you define `ReservedTags`. This operation is needlessly costly to compute something that is an all-time constant.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+
+    // Some classes may have parents that have been removed
+    val orphans = for(cls <- linkingUnit.classDefs;
+                    parent <- cls.superClass
+                    if !linkedClassByName.isDefinedAt(parent.name)) yield cls
+
+    val roots = childrenMap(None) ++ orphans
+
+    @tailrec def getTypeTags(curTag: Int, stack: Seq[LinkedClass],
+      accu: Map[LinkedClass, Int] = Map.empty): Map[LinkedClass, Int] = stack match {
```
sjrd: 4 spaces

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+      case cur +: rest =>
+        val nextTag = curTag + 1
+        val children = childrenMap(Some(cur.name.name))
+        getTypeTags(nextTag, children ++ rest, accu + (cur -> curTag))
+
+      case Nil =>
+        // Sanity check: all classDefs must have a tag
+        assert(linkingUnit.classDefs.forall(accu.isDefinedAt))
+        accu
+    }
```
sjrd: The reified `stack` is a useless complication. You should simply write a (non-tail-)recursive function that walks the tree. The tree's depth is never going so big that we can avoid a StackOverflowError by using an on-heap stack.

This also avoids all the manipulations of these `Seq`s, which are not super efficient.

Finally, I suggest using a `Map.newBuilder[LinkedClass, Int]` and mutable additions to it, instead of trying to stay functional for the sake of it.

ir/src/main/scala/org/scalajs/core/ir/Definitions.scala
```diff
@@ -68,6 +68,10 @@ object Definitions {
 
   val ExportedConstructorsName = "__exportedInits"
 
+  val ReservedTags = Map("V" -> 0, "Z" -> 1, "C" -> 2, "B" -> 3,
+    "S" -> 4, "I" -> 5, "J" -> 6, "F" -> 7, "D" -> 8,
+    ObjectClass -> 9, CloneableClass -> 10, SerializableClass -> 11)
```
sjrd: This is an `Emitter`-specific implementation detail. It should be somewhere in `emitter` and be `private[emitter]`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
sjrd: On a different topic, this algorithm is not *stable*, in the sense that nothing guarantees that, on two consecutive runs, it will give the same tags to the same classes. This could be annoying as compiling the same codebase twice does not give the same .js file.

It would be good to stabilize it, for example by sorting `roots` and `children` by `_.name.name`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+          case ((start, end) :: rest, tag) if tag == end + 1 => (start, tag) :: rest
+          case (intervals, tag) => (tag, tag) :: intervals
+        }.reverse
+    }
+  }
+
+  private[emitter] def needsSubtypeArray(className: String): Boolean = {
+    val numComparaisons = subtypeIntervals(className).map {
+      case (start, end) if start == end => 1
+      case _ => 2
```
sjrd: Align the `=>`'s

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+        assert(linkingUnit.classDefs.forall(accu.isDefinedAt))
+        accu
+    }
+
+    // The smallest tags are reserved
+    getTypeTags(ReservedTags.values.max + 1, roots)
+  }
+
+  private[emitter] lazy val subtypeTags: Map[String, List[Int]] = {
+    val descendsMap = for(
```
sjrd: This should be named `descendentsMap`. No need to shorten that name.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+        child <- typeTags.keys;
+        parent <- child.ancestors) yield parent -> child
```
sjrd: for coding style

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
sjrd: Actually, it's not even a Map. So it should be called `parentToChilds` or something like that.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+
+    descendsMap.groupBy {
+      case (parent, child) => parent
+    } map {
+      case (parent, parentChilds) =>
```
sjrd: `children`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    getTypeTags(ReservedTags.values.max + 1, roots)
+  }
+
+  private[emitter] lazy val subtypeTags: Map[String, List[Int]] = {
+    val descendsMap = for(
+        child <- typeTags.keys;
+        parent <- child.ancestors) yield parent -> child
+
+    descendsMap.groupBy {
+      case (parent, child) => parent
```
sjrd: `.groupBy(_._1)`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+        // Sanity check: all classDefs must have a tag
+        assert(linkingUnit.classDefs.forall(accu.isDefinedAt))
+        accu
+    }
+
+    // The smallest tags are reserved
+    getTypeTags(ReservedTags.values.max + 1, roots)
+  }
+
+  private[emitter] lazy val subtypeTags: Map[String, List[Int]] = {
```
sjrd: This is only used in `subtypeIntervals`. Its computation should be moved as an inner `val` or `def` within `subtypeIntervals`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+    } map {
+      case (parent, parentChilds) =>
+        parent -> parentChilds.map(t => typeTags(t._2)).toList.sorted
+    }
+  }
+
+  private[emitter] lazy val subtypeIntervals: Map[String, List[(Int, Int)]] = {
+    subtypeTags.map {
+      case (parent, tags) =>
+        parent -> tags.foldLeft(List[(Int, Int)]()){
```
sjrd: Prefer writing an explicit type for `foldLeft` as `.foldLeft[List[(Int, Int)]](Nil) {`.
btw, missing space before `{`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +74,73 @@ private[scalajs] final class ScalaJSClassEmitter(
...
+          case ((start, end) :: rest, tag) if tag == end + 1 => (start, tag) :: rest
+          case (intervals, tag) => (tag, tag) :: intervals
```
sjrd: Align the arrows, or put the case bodies on their own lines if that's too long.

sjrd:
That's all I have for now.

tools/scalajsenv.js
paullepoulpe: We could store array class tags in a different array, that array would still need to be sparse though (because of nested arrays). But it would not hurt the performance of all non-array objects.

tools/scalajsenv.js
paullepoulpe: Yup

tools/scalajsenv.js
paullepoulpe: Should this be a separate commit ?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
paullepoulpe: Correct

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
paullepoulpe: That should've been an assert, every class has at least its own tag in it's interval list. intervals can never be empty

tools/scalajsenv.js
sjrd: OK. Let's keep the existing version, then. In the future we can refine it on the basis of benchmarks.

tools/scalajsenv.js
sjrd: No it should be the same commit, since your commit turns this into dead code, so it should also remove it.

tools/scalajsenv.js
```diff
@@ -144,13 +144,27 @@ const $idHashCodeMap = ScalaJS.g["WeakMap"] ? new ScalaJS.g["WeakMap"]() : null;
...
+  }
+  return expanded;
+};
+
+ScalaJS.makeIsArrayOfPrimitive = function(primitiveTag) {
   return function(obj, depth) {
-    return !!(obj && obj.$classData &&
-      (obj.$classData.arrayDepth === depth) &&
-      (obj.$classData.arrayBase === primitiveData));
+     return !!(obj && (obj.$typeTag < 0) &&
```
paullepoulpe: Not sure we need this part: `(obj.$typeTag < 0)`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -72,6 +75,74 @@ private[scalajs] final class ScalaJSClassEmitter(
 
   private[emitter] def semantics: Semantics = linkingUnit.semantics
 
+  private[emitter] val ReservedTags = Map("V" -> 0, "Z" -> 1, "C" -> 2,
```
paullepoulpe: I am not quite sure this is the best place to put this.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -610,18 +705,20 @@ private[scalajs] final class ScalaJSClassEmitter(
...
-            js.Return(!(!({
-              genIsScalaJSObject(obj) &&
-              ((obj DOT "$classData" DOT "arrayDepth") === depth) &&
-              genIsClassNameInAncestors(className,
-                  obj DOT "$classData" DOT "arrayBase" DOT "ancestors")
-            })))
+            val tagVarDef = genLet(js.Ident("tag"), mutable = false, obj && (obj DOT "$typeTag"))
+            val tag = tagVarDef.ref
+            val test = {
+              js.BinaryOp(JSBinaryOp.<, tag, js.IntLiteral(0)) &&
```
paullepoulpe: Same thing here, I think this is useless

sjrd:
Closing as this would probably have to be redone from scratch at this point.

PR #2414, opened 2016-05-26
Fix #2381: Implement regex.Matcher.group(String).

### LINKED ISSUE #2381: Implement java.util.regex.Matcher.group(String).
See https://github.com/scala/scala/pull/4990#issuecomment-216760517

JavaDoc: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#group%28java.lang.String%29

nicolasstucki:


nicolasstucki:
Review by @sjrd.

javalib/src/main/scala/java/util/regex/Matcher.scala
```diff
@@ -195,6 +195,8 @@ final class Matcher private[regex] (
 
   def group(group: Int): String = ensureLastMatch(group).orNull
 
+  def group(name: String): String = throw new NoSuchElementException
```
sjrd: Er ... It's supposed to return an `IllegalArgumentException`: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#group%28java.lang.String%29

Also ideally it should throw an `IllegalStateException` instead if no match was attempted so far (like calling `ensureLastMatch`).

sjrd:
LGTM

PR #2416, opened 2016-05-27
DO NOT MERGE Prepare switch with guards

sjrd:
This is for CI only. I intend to redo this (the second commit of this PR) on stage during my talk at Scala Days Berlin.
No review needed at the moment.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2512,21 +2512,22 @@ abstract class GenJSCode extends plugins.PluginComponent
       val expr = genExpr(selector)
       val resultType = toIRType(tree.tpe)
 
-      val List(defaultBody0) = for {
-        CaseDef(Ident(nme.WILDCARD), EmptyTree, body) <- cases
-      } yield body
-
-      val (defaultBody, defaultLabelSym) = defaultBody0 match {
-        case LabelDef(_, Nil, rhs) if hasSynthCaseSymbol(defaultBody0) =>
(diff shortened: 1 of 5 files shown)


sjrd:
Erf ... GCC wrongly optimizes away [`RegressionTest.switch_match_with_2_guards_for_the_same_value_issue_1589`](https://github.com/sjrd/scala-js/blob/32fd845a7f8049b35f5ca9fee618102c91636602/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala#L419-L429) as
```js
throw $c_s_MatchError(14);
```
even though our fastOpt code is correct:
```scala
$c_Lorg_scalajs_testsuite_compiler_RegressionTest.prototype.switch$undmatch$undwith$und2$undguards$undfor$undthe$undsame$undvalue$undissue$und1589__V = (function() {
  var b = this.genB$1__p1__I();
  var x1 = (b >> 4);
  x: {
    var x$2;
    $default: {
      switch (x1) {
        case 14: {
          if ((b === 224)) {
            var x$2 = 4;
            break x
          } else if ((b === 225)) {
            var x$2 = 5;
            break x
          } else {
            break $default
          };
          break
        }
        default: {
          /*<skip>*/
        }
      }
    };
    throw new $c_s_MatchError().init___O(x1)
  };
  var this$1 = $m_Lorg_junit_Assert$();
  this$1.assertEquals__T__O__O__V(null, 5, x$2)
});
```
The online version of GCC does not appear to have this bug, so it was probably fixed since the old version we're using.

Consequently, it is unlikely we can merge this improvement anytime soon. But I'll probably use this example anyway for my Scala Days talk, because it is really a nice example of hacking the compiler/.

PR #2417, opened 2016-05-27
Fix #2401: Allow package objects to have non named native classes.

### LINKED ISSUE #2401: Allow native objects in package objects without @JSName
These two snippets are equivalent:

1:
```scala
package foo.bar

import scalajs.js

@js.native
object Ah extends js.Object
```

2:
```scala
package foo

import scalajs.js

package object bar {
  @js.native
  object Ah extends js.Object
}
```

Yet the second snippet fails compilation with
```
[error] ah.scala:7: Native JS objects inside non-native objects must have an @JSName annotation
[error]   object Ah extends js.Object
```

nicolasstucki:


nicolasstucki:
Review by @sjrd

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSNativeObjectInPackage.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.scalajs.js.annotation.JSName
+
+package object jsNativeObjectInPackage {
+  @js.native
+  object Foo extends js.Object
+
+  @js.native
+  @JSName("Bar")
+  object Baz extends js.Object
+}
```
sjrd: You don't seem to use `Foo` nor `Baz` in the tests.
Also, you should test classes.

test-suite/js/src/test/resources/ScalaJSDefinedTestNatives.js
```diff
@@ -50,4 +50,6 @@
     }
   };
   $g.ConstructorDefaultParam = ConstructorDefaultParam;
+  $g.Foo = {};
+  $g.Bar = {};
```
sjrd: This enters JS' global scope. They should have longer names that won't clash with anything.

sjrd:
That's all.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -597,6 +597,21 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+    """.hasNoWarns
+
+    """
+    package object A {
+      @js.native
+      class B extends js.Object
+    }
+    """.hasNoWarns
+
     """
```
sjrd: You should add tests for `class`es here, too.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSNativeInPackage.scala
```diff
@@ -0,0 +1,70 @@
+package org.scalajs.testsuite.jsinterop
+
+import org.junit.Test
+import org.junit.Assert._
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSName
+
+package object jsNativeInPackage {
```
sjrd: Uh oh. I just realized this package object has the same name as the class below except for the case. This will make Windows go berserk.

You should rename the package object (and, btw, give it an all-lowercase name). I suggest `packageobjectwithnatives`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSNativeInPackage.scala
```diff
@@ -0,0 +1,70 @@
...
+    assertEquals("foo", gJSNativeClassInPackageFoo.foo())
+    assertEquals("foo", new JSNativeClassInPackageFoo().foo())
+  }
+
+  @Test def testClassJSName(): Unit = {
+    val gJSNativeClassInPackageBarCtr = global.JSNativeClassInPackageBar
+    val gJSNativeClassInPackageBazCtr = global.JSNativeClassInPackageBaz
+    assertFalse(js.isUndefined(gJSNativeClassInPackageBarCtr))
+    assertTrue(js.isUndefined(gJSNativeClassInPackageBazCtr))
+
```
sjrd: Another one:
```scala
assertSame(js.constructorOf[JSNativeClassInPackageBaz], gJSNativeClassInPackageBarCtr)
```

sjrd:
That's all.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
nicolasstucki: Isn't that the test for `class`:
```scala
package object A {
  @js.native
  class B extends js.Object
}
```

Or do you have another test in mind?

sjrd: Er no you're right. I was just being blind, I guess ^^

sjrd:
LGTM

PR #2418, opened 2016-05-27
Sbt-plugin: raw IR AST printer

piotrMocz:
Added the `--raw` (or `-r` in short) option to scalajsp, which prints the AST nodes comprising the given IR file.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
@@ -1083,9 +1083,9 @@ object Printers {
     }
 
     private final def printList(ts: List[Tree]): Unit = {
-      if (ts.isEmpty)
+      if (ts.isEmpty) {
         print("List ()")
-      else {
+      } else {
         println("List (")
         indent()
         var rest = ts
@@ -1106,9 +1106,9 @@ object Printers {
     }
 
     private final def printIdentList(is: List[Ident]): Unit = {
-      if (is.isEmpty)
+      if (is.isEmpty) {
         print("List ()")
-      else {
+      } else {
         println("List (")
         indent()
         var rest = is
@@ -1129,7 +1129,7 @@ object Printers {
     }
 
     private final def print(ident: Ident): Unit =
-      print(s"Ident(${ident.name}, ${ident.originalName}")
+      print(s"Ident(${ident.name}, ${ident.originalName})")
 
     private final def println(ident: Ident): Unit = {
       print(ident)
@@ -1161,7 +1161,7 @@ object Printers {
           indent()
           println(ident)
           println(vtpe)
-          println(mutable.toString)
+          println("mutable = " + mutable.toString)
           print(rhs); paren()
           undent()
 
@@ -1171,8 +1171,8 @@ object Printers {
           indent()
           println(ident)
           println(ptpe)
-          println(mutable.toString)
-          print(rest.toString); paren()
+          println("mutable = " + mutable.toString)
+          print("rest = " + rest.toString); paren()
           undent()
 
         // Control flow constructs
@@ -1288,10 +1288,8 @@ object Printers {
           undent()
 
         case LoadModule(cls) =>
-          println("LoadModule (")
-          indent()
+          print("LoadModule (")
           print(cls); paren()
-          undent()
 
         case StoreModule(cls, value) =>
           println("StoreModule (")
@@ -1586,13 +1584,15 @@ object Printers {
         case FieldDef(name, vtpe, mutable) =>
           println("FieldDef (")
           indent()
+          println(name.name)
           println(vtpe)
-          print(mutable.toString + ")")
+          print("mutable = " + mutable.toString + ")")
           undent()
 
         case MethodDef(static, name, args, resultType, body) =>
           println("MethodDef (")
           indent()
+          println("static = " + static.toString)
           println(name.toString)
           printlnList(args)
           println(resultType)
--- a/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
@@ -1073,7 +1073,7 @@ object Printers {
       out.write("\n")
   }
 
-  class IRASTPrinter(protected val out: Writer) extends RawIndentationManager {
+  class RawIRPrinter(protected val out: Writer) extends RawIndentationManager {
     /** Pretty-prints the IR's AST nodes. */
     def printTopLevelTree(tree: Tree): Unit =
       println(tree)
@@ -1209,12 +1209,12 @@ object Printers {
           out.write(")")
           undent()
 
-        case ifExpr@If(cond, thenp, elsep) =>
+        case ifExpr: If =>
           println("If(")
           indent()
-          println(cond)
-          println(thenp)
-          print(elsep)
+          println(ifExpr.cond)
+          println(ifExpr.thenp)
+          print(ifExpr.elsep)
           out.write(")(\n")
           print(ifExpr.tpe)
           out.write(")")
@@ -1299,9 +1299,11 @@ object Printers {
           undent()
 
         case LoadModule(cls) =>
-          print("LoadModule(")
+          println("LoadModule(")
+          indent()
           print(cls)
           out.write(")")
+          undent()
 
         case StoreModule(cls, value) =>
           println("StoreModule(")
--- a/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Printers.scala
@@ -1242,14 +1242,14 @@ object Printers {
           out.write(")")
           undent()
 
-        case Try(block, errVar, handler, finalizer) =>
+        case t: Try =>
           println("Try(")
           indent()
-          println(block)
-          println(errVar)
-          println(handler)
-          print(finalizer)
-          out.write(")")
+          println(t.block)
+          println(t.errVar)
+          println(t.handler)
+          print(t.finalizer)
+          out.write(")(" + t.tpe + ")")
           undent()
 
         case Throw(expr) =>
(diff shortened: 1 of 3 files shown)


sjrd:
Jenkins, ok to test

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
     def complete(): Unit = ()
   }
 
+  abstract class RawIndentationManager extends IndentationManager {
```
sjrd: It seems this class does not keep any behavior from `IndentationManager`, so I don't see the point of extending it.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+    protected def println(s: String): Unit =
+      print(s + "\n")
+
+    override protected def println(): Unit =
+      out.write("\n")
+  }
+
+  class IRASTPrinter(protected val out: Writer) extends RawIndentationManager {
+    /** Pretty-prints the IR's AST nodes.
+      */
```
sjrd: See https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#scaladoc for the Scaladoc style to use.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+      println(tree)
+
+    private final def println(tree: Tree): Unit = {
+      print(tree)
+      out.write("\n")
+    }
+
+    private final def printList(ts: List[Tree]): Unit = {
+      if (ts.isEmpty) {
+        print("List ()")
```
sjrd: In general, I think it would be best not to have a space before `(`. That would be more consistent with the coding style we use in the actual code.

This comment applies to all other occurrences of this pattern in this file.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+      out.write("\n")
+  }
+
+  class IRASTPrinter(protected val out: Writer) extends RawIndentationManager {
+    /** Pretty-prints the IR's AST nodes.
+      */
+    def printTopLevelTree(tree: Tree): Unit =
+      println(tree)
+
+    private final def println(tree: Tree): Unit = {
```
sjrd: `final` is redundant for `private` methods. Consider removing it.

sjrd:
You should add tests, similarly to those found in https://github.com/scala-js/scala-js/blob/master/ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala (probably in a separate file `RawPrintersTest.scala`).

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+      print(tpe.toString)
+
+
+    private final def println(tp: Type): Unit = {
+      print(tp)
+      out.write("\n")
+    }
+
+    private final def paren(): Unit = {
+      out.write(")")
```
sjrd: This looks like over-abstraction to me. I suggest writing `out.write(")")` explicitly instead of calling `paren()` (btw, `paren` is not a good name because we've got no idea whether it's a opening or closing paren)

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+
+        // Definitions
+
+        case VarDef(ident, vtpe, mutable, rhs) =>
+          println("VarDef (")
+          indent()
+          println(ident)
+          println(vtpe)
+          println("mutable = " + mutable.toString)
+          print(rhs); paren()
```
sjrd: Consider writing `paren()`/`out.write("(")` on its own line, instead of using a `;`

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+          paren()
+          undent()
+
+        case If(cond, thenp, elsep) =>
+          println("If (")
+          indent()
+          println(cond)
+          println(thenp)
+          print(elsep); paren()
+          undent()
```
sjrd: For trees like `If` where the `tpe` is an explicit parameter to the constructor of `If` (as opposed to being inferred), I would suggest you also print it explicitly, so that an `If` looks like this:
```scala
If(
  BooleanLiteral(true),
  IntLiteral(2),
  IntLiteral(3))(
  IntType)
```

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+          }
+          paren()
+          undent()
+
+        case Match(selector, cases, default) =>
+          println("Match (")
+          indent()
+          println(selector)
+
+          cases foreach { pair =>
```
sjrd: Use dot syntax: `cases.foreach { pair =>`

sjrd: Or even better, a for comprehension:
```scala
for ((alts, body) <- cases) {
  print(alts.toString()); println(":")
  indent(); print(body); undent()
}
```

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+            print(pair._1.toString()); println(":")
+            indent(); print(pair._2); undent()
+          }
+
+          print(default); paren()
+          undent()
+
+
+        case Debugger() =>
+          print("Debugger")
```
sjrd: Explicitly write the `()`. Otherwise it looks like a `case object` instead of a `case class`

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+        case JSArrayConstr(items) =>
+          println("JSArrayConstr (")
+          indent()
+          printList(items); paren()
+          undent()
+
+        case JSObjectConstr(fields) =>
+          println("JSObjectConstr (")
+          indent()
+          fields foreach { pair =>
```
sjrd: Use a for comprehension.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+          undent()
+
+        case LoadJSModule(cls) =>
+          println("LoadJSModule (")
+          indent()
+          print(cls); paren()
+          undent()
+
+        case JSSpread(items) =>
+          println("JSSpreas (")
```
sjrd: Typo: `JSSpread`

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+          println("JSObjectConstr (")
+          indent()
+          fields foreach { pair =>
+            print(pair._1); print(": "); println(pair._2)
+          }
+          paren()
+          undent()
+
+        case JSLinkingInfo() =>
+          print("JSLinkingInfo")
```
sjrd: Explicit `()` (there are a couple more like this)

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -28,13 +28,12 @@ import org.scalajs.jsenv.phantomjs.{PhantomJSEnv, PhantomJettyClassLoader}
...
 import org.scalajs.core.ir.Utils.escapeJS
 import org.scalajs.core.ir.ScalaJSVersions
-import org.scalajs.core.ir.Printers.{InfoPrinter, IRTreePrinter}
+import org.scalajs.core.ir.Printers.{IRASTPrinter, IRTreePrinter, InfoPrinter}
 
 import org.scalajs.testadapter.ScalaJSFramework
 
 import scala.util.Try
 import scala.collection.mutable
-
```
sjrd: Do not remove this line.

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -270,12 +269,15 @@ object ScalaJSPluginInternal {
...
+        infos: Boolean = false,
+        raw: Boolean = false
     )
 
     val optionsParser: Parser[Options] = {
       token(OptSpace ~> (
-          (literal("-i") | "--infos") ^^^ ((_: Options).copy(infos = true))
+          (literal("-i") | "--infos") ^^^ ((_: Options).copy(infos = true)) |
+            (literal("-r") | "--raw") ^^^ ((_: Options).copy(raw = true))
+
```
sjrd: For parser combinators, prefer this style:
```scala
      token(OptSpace ~> (
          (literal("-i") | "--infos") ^^^ ((_: Options).copy(infos = true))
        | (literal("-r") | "--raw") ^^^ ((_: Options).copy(raw = true))
      )
```
(you also have a spurious blank line, here)

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -1052,4 +1052,601 @@ object Printers {
...
+    }
+
+    protected def println(s: String): Unit =
+      print(s + "\n")
+
+    override protected def println(): Unit =
+      out.write("\n")
+  }
+
+  class IRASTPrinter(protected val out: Writer) extends RawIndentationManager {
```
sjrd: Consider naming this `RawIRPrinter`.

sjrd:
The `--raw`/`-r` option should also be added to the command-line `scalajsp` whose code is in https://github.com/scala-js/scala-js/blob/master/cli/src/main/scala/org/scalajs/cli/Scalajsp.scala

sjrd:
That's all, I believe.

sjrd:
Closing for lack of follow up.

PR #2420, opened 2016-05-28
Enable local dce of vals across their scope boundaries.

sjrd:
Consider this snippet:
```scala
@inline class Foo(a: Int, b: Int) {
  val x = a + b
  val y = a - b
}

val foo = new Foo(5, someNonConstant)
println(foo.y)
```
Previously, even though the optimizer could completely virtualize `foo` and replace `foo.y` by a local `val y`, it would also declare a local `val x`, although it is not used. In other words, it was not able to dce `x` away.

That is because, when pre-transforming `new Foo()`, the local vals `x` and `y` that are created are confined to a scope like this:
```scala
val foo = {
  val x = 5 + someNonConstant
  val y = 5 - someNonConstant
  <virtual>(x, y)
}
```
When escaping the scoped block, the optimizer had to assume that both `x` and `y` are used, because they are "contained" in the virtual object that escapes the scope. This prevents from dce'ing them away.

This commit fixes this limitation by enhancing `PreTransBlock`. A `PreTransBlock` can now keep virtual bindings in addition to completely transformed statements. Only when we force the `PreTransBlock` (or keep only its side-effects) do we test what bindings have actually been used. Those that are not can be dce'ed away.

The rhs of `foo` therefore becomes a virtual block (a `PreTransBlock`) which keeps the bindings of `x` and `y` virtual. When we later discover that this `PreTransBlock` need not be forced (because `foo` itself is not used), the optimizer tests whether `x` and `y` are used. `y` is used at this point, because it was used to replace `foo.y`. `x`, on the other hand, is not used, and can be removed (along with the side-effect-free parts of its rhs).

Lifting this restriction significantly improves the optimized version of `Range.foreach`. This is because, since 0e64158871d3ba57a426dcc6c859fd93b0d825e8, `numRangeElements` is not needed anymore for `foreach`. And `numRangeElements` is the one field of the initialization of `Range` that requires non-trivial operations on `Long`s: they are completely dce'ed away with the new optimizations.

Implementing these new opportunities for optimizations made it harder to support the optimizations of `Long.toInt` and `Long +/- Long` across `val`s (the thing that was stored in `ReplaceWithVarRef.longOpTree`). The current commit does not support filling in that field with `Some(_)` when appropriate. It is a small regression, in theory. But in practice, this had been added mostly to supported optimizing the `Long` operations involved in computing `numRangeElements` as `Int` operations, in the first place. Since now, `numRangeElements` is completely removed anyway, it does not really matter. This commit keeps the ability of the rest of the optimizer to *use* the `longOpTree` information if it is present, even though it is never set anymore. That is because I have ideas about how to recover the ability to optimize expressions across `val`s, in a more generic way than `longOpTree`, and I want to keep a compiling reference of what optimizations we used to be able to perform before.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -901,7 +901,7 @@ private[optimizer] abstract class OptimizerCore(
               cont(PreTransTree(transformedStat, RefinedType.Nothing))
             else {
               pretransformList(rest) { trest =>
-                cont(PreTransBlock(transformedStat :: Nil, trest))
+                cont(PreTransBlock(transformedStat, trest))
               }
             }
         }
@@ -1004,12 +1004,13 @@ private[optimizer] abstract class OptimizerCore(
   private def resolveLocalDef(preTrans: PreTransform): PreTransGenTree = {
     implicit val pos = preTrans.pos
     preTrans match {
-      case PreTransBlock(stats, result) =>
+      case PreTransBlock(bindingsAndStats, result) =>
         resolveLocalDef(result) match {
           case PreTransRecordTree(tree, tpe, cancelFun) =>
-            PreTransRecordTree(Block(stats :+ tree), tpe, cancelFun)
+            PreTransRecordTree(finishTransformBindings(bindingsAndStats, tree),
+                tpe, cancelFun)
           case PreTransTree(tree, tpe) =>
-            PreTransTree(Block(stats :+ tree), tpe)
+            PreTransTree(finishTransformBindings(bindingsAndStats, tree), tpe)
         }
 
       case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
@@ -1034,6 +1035,42 @@ private[optimizer] abstract class OptimizerCore(
 
       case preTrans: PreTransGenTree =>
         preTrans
+    }
+  }
+
+  /** Resolves any [[RecordType]] in a [[PreTransform]].
+   *
+   *  If `preTrans` would resolve to a `PreTransRecordTree`, returns a `Some`
+   *  of its (lowered) [[RecordType]] and its `cancelFun`. Otherwise, returns
+   *  `None`.
+   *
+   *  Note that the record type is not the same as `preTrans.tpe.base`, which
+   *  is the *original* type of the tree (not lowered to a record type).
+   */
+  private def resolveRecordType(
+      preTrans: PreTransform): Option[(RecordType, CancelFun)] = {
+    preTrans match {
+      case PreTransBlock(_, result) =>
+        resolveRecordType(result)
+
+      case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
+        replacement match {
+          case ReplaceWithRecordVarRef(name, originalName,
+              recordType, used, cancelFun) =>
+            Some((recordType, cancelFun))
+
+          case InlineClassInstanceReplacement(recordType, fieldLocalDefs, cancelFun) =>
+            Some((recordType, cancelFun))
+
+          case _ =>
+            None
+        }
+
+      case PreTransRecordTree(tree, _, cancelFun) =>
+        Some((tree.tpe.asInstanceOf[RecordType], cancelFun))
+
+      case PreTransTree(_, _) =>
+        None
     }
   }
 
@@ -1064,8 +1101,8 @@ private[optimizer] abstract class OptimizerCore(
   private def finishTransformExpr(preTrans: PreTransform): Tree = {
     implicit val pos = preTrans.pos
     preTrans match {
-      case PreTransBlock(stats, result) =>
-        Block(stats :+ finishTransformExpr(result))
+      case PreTransBlock(bindingsAndStats, result) =>
+        finishTransformBindings(bindingsAndStats, finishTransformExpr(result))
       case PreTransLocalDef(localDef) =>
         localDef.newReplacement
       case PreTransRecordTree(_, _, cancelFun) =>
@@ -1083,14 +1120,61 @@ private[optimizer] abstract class OptimizerCore(
    *  it is once in the 'try' part and once in the 'fallback' part.
    */
   private def finishTransformStat(stat: PreTransform): Tree = stat match {
-    case PreTransBlock(stats, result) =>
-      Block(stats :+ finishTransformStat(result))(stat.pos)
+    case PreTransBlock(bindingsAndStats, result) =>
+      finishTransformBindings(bindingsAndStats, finishTransformStat(result))
     case PreTransLocalDef(_) =>
       Skip()(stat.pos)
     case PreTransRecordTree(tree, _, _) =>
       keepOnlySideEffects(tree)
     case PreTransTree(tree, _) =>
       keepOnlySideEffects(tree)
+  }
+
+  /** Finishes the bindings and statements followed by a result to get a
+   *  normal [[Tree]].
+   *  This method must not be called more than once per `BindingOrStat` and
+   *  per translation.
+   *  By "per translation", we mean in an alternative path through
+   *  `tryOrRollback`. It could still be called several times as long as
+   *  it is once in the 'try' part and once in the 'fallback' part.
+   */
+  private def finishTransformBindings(bindingsAndStats: List[BindingOrStat],
+      result: Tree): Tree = {
+    bindingsAndStats.foldRight(result) {
+      case (Left(PreTransBinding(localDef, value)), innerBody) =>
+        implicit val pos = value.pos
+
+        val LocalDef(tpe, mutable, replacement) = localDef
+
+        val (name, originalName, used) = (replacement: @unchecked) match {
+          case ReplaceWithVarRef(name, originalName, used, _) =>
+            (name, originalName, used)
+          case ReplaceWithRecordVarRef(name, originalName, _, used, _) =>
+            (name, originalName, used)
+        }
+
+        if (used.value) {
+          val ident = Ident(name, originalName)
+          val varDef = resolveLocalDef(value) match {
+            case PreTransRecordTree(valueTree, valueTpe, cancelFun) =>
+              val recordType = valueTree.tpe.asInstanceOf[RecordType]
+              if (!isImmutableType(recordType))
+                cancelFun()
+              VarDef(ident, recordType, mutable, valueTree)
+
+            case PreTransTree(valueTree, valueTpe) =>
+              VarDef(ident, tpe.base, mutable, valueTree)
+          }
+
+          Block(varDef, innerBody)
+        } else {
+          val valueSideEffects = finishTransformStat(value)
+          Block(valueSideEffects, innerBody)
+        }
+
+      case (Right(stat), innerBody) =>
+        Block(stat, innerBody)(innerBody.pos)
+    }
   }
 
   /** Keeps only the side effects of a Tree (overapproximation). */
@@ -1915,7 +1999,7 @@ private[optimizer] abstract class OptimizerCore(
               inlineClassConstructorBodyList(allocationSite,
                   thisLocalDef, inputFieldsLocalDefs,
                   cls, rest, cancelFun)(buildInner) { tinner =>
-                cont(PreTransBlock(transformedStat :: Nil, tinner))
+                cont(PreTransBlock(transformedStat, tinner))
               }
             }
         }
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

sjrd:
For this test program:
```scala
package helloworld

import scala.scalajs.js

  def main() {
    rangeForeach1(10)
    println("---")
    rangeForeach2(10)
    println("---")
    rangeForeach3(10)
    println("---")
    rangeForeach4(10)
    println("---")
    rangeSum3(10)
  }

  def rangeForeach1(n: Int): Unit = {
    for (i <- 0 until n)
      println(i)
  }

  def rangeForeach2(n: Int): Unit = {
    for (i <- -3 to n)
      println(i)
  }

  def rangeForeach3(n: Int): Unit = {
    for (i <- 0 until n by 3)
      println(i)
  }

  def rangeForeach4(n: Int): Unit = {
    for (i <- -3 to n by 3)
      println(i)
  }

  def rangeSum3(n: Int): Unit = {
    val r = 0 until n by 3
    println(r.sum)
  }

  @noinline def println(x: Any): Unit = Console.println(x)
}
```
the diff of fastopt.js between before and after this commit is the following:
```diff
diff --git "a/C:\\Users\\Sepi\\Documents\\Projets\\helloworld-fastopt.js" "b/C:\\Users\\Sepi\\Documents\\Projets\\scalajs\\examples\\helloworld\\target\\scala-2.11\\helloworld-fastopt.js"
index 7456d97..91da8a4 100644
--- "a/C:\\Users\\Sepi\\Documents\\Projets\\helloworld-fastopt.js"
+++ "b/C:\\Users\\Sepi\\Documents\\Projets\\scalajs\\examples\\helloworld\\target\\scala-2.11\\helloworld-fastopt.js"
@@ -1682,16 +1682,16 @@ $c_Lhelloworld_HelloWorld$.prototype.init___ = (function() {
 });
 $c_Lhelloworld_HelloWorld$.prototype.rangeSum3__I__V = (function(n) {
   var isEmpty$4 = (n <= 0);
-  var numRangeElements$4 = (isEmpty$4 ? 0 : n);
+  if (isEmpty$4) {
+    /*<skip>*/
+  };
   var lastElement$4 = (isEmpty$4 ? (-1) : (((-1) + n) | 0));
-  var terminalElement$4 = ((1 + lastElement$4) | 0);
   var isEmpty$4$1 = (n <= 0);
   if (isEmpty$4$1) {
     var numRangeElements$4$1 = 0
   } else {
-    var x = ((n / 3) | 0);
-    var y = ((((n % 3) | 0) !== 0) ? 1 : 0);
-    var numRangeElements$4$1 = ((((x > 0) && (y > 0)) && (((x + y) | 0) < 0)) ? (-1) : ((x + y) | 0))
+    var len$1 = new $c_sjsr_RuntimeLong().init___I(((n / 3) | 0)).$$plus__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I(((((n % 3) | 0) !== 0) ? 1 : 0)));
+    var numRangeElements$4$1 = (len$1.$$greater__sjsr_RuntimeLong__Z(new $c_sjsr_RuntimeLong().init___I__I(2147483647, 0)) ? (-1) : len$1.lo$2)
   };
   if (isEmpty$4$1) {
     var lastElement$4$1 = (-3)
@@ -1699,7 +1699,6 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeSum3__I__V = (function(n) {
     var remainder = ((n % 3) | 0);
     var lastElement$4$1 = ((remainder !== 0) ? ((n - remainder) | 0) : (((-3) + n) | 0))
   };
-  var terminalElement$4$1 = ((3 + lastElement$4$1) | 0);
   var num = $m_s_math_Numeric$IntIsIntegral$();
   inlinereturn$16: {
     if ((num === $m_s_math_Numeric$IntIsIntegral$())) {
@@ -1724,13 +1723,14 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeSum3__I__V = (function(n) {
       var acc = 0;
       var i = (isEmpty$4$1 ? $m_sci_Nil$().head__sr_Nothing$() : 0);
       while (true) {
-        var x$1 = acc;
-        var y$1 = i;
-        var x$2 = $uI(x$1);
-        acc = $s_s_math_Numeric$IntIsIntegral$class__plus__s_math_Numeric$IntIsIntegral__I__I__I(num, x$2, y$1);
+        var x = acc;
+        var y = i;
+        var x$1 = $uI(x);
+        acc = $s_s_math_Numeric$IntIsIntegral$class__plus__s_math_Numeric$IntIsIntegral__I__I__I(num, x$1, y);
         if ((i === lastElement$4$1)) {
-          var x$3 = acc;
-          var jsx$1 = $uI(x$3);
+          var x$2 = acc;
+          var x$3 = $uI(x$2);
+          var jsx$1 = x$3;
           break inlinereturn$16
         };
         i = ((3 + i) | 0)
@@ -1741,9 +1741,10 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeSum3__I__V = (function(n) {
 });
 $c_Lhelloworld_HelloWorld$.prototype.rangeForeach1__I__V = (function(n) {
   var isEmpty$4 = (n <= 0);
-  var numRangeElements$4 = (isEmpty$4 ? 0 : n);
+  if (isEmpty$4) {
+    /*<skip>*/
+  };
   var lastElement$4 = (isEmpty$4 ? (-1) : (((-1) + n) | 0));
-  var terminalElement$4 = ((1 + lastElement$4) | 0);
   inlinereturn$8: {
     if ((!isEmpty$4)) {
       var i = 0;
@@ -1760,15 +1761,13 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeForeach1__I__V = (function(n) {
 });
 $c_Lhelloworld_HelloWorld$.prototype.rangeForeach4__I__V = (function(n) {
   var isEmpty$4 = (n < (-3));
-  var numRangeElements$4 = (isEmpty$4 ? 0 : ((n > 2147483643) ? (-1) : ((4 + n) | 0)));
+  if (isEmpty$4) {
+    /*<skip>*/
+  };
   var lastElement$4 = (isEmpty$4 ? (-4) : n);
-  var terminalElement$4 = ((1 + lastElement$4) | 0);
   var isEmpty$4$1 = (n < (-3));
   if (isEmpty$4$1) {
-    var numRangeElements$4$1 = 0
-  } else {
-    var len$1 = new $c_sjsr_RuntimeLong().init___I__I(1, 0).$$plus__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I__I(3, 0).$$plus__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I(n)).$$div__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I__I(3, 0)));
-    var numRangeElements$4$1 = (len$1.$$greater__sjsr_RuntimeLong__Z(new $c_sjsr_RuntimeLong().init___I__I(2147483647, 0)) ? (-1) : len$1.lo$2)
+    /*<skip>*/
   };
   if (isEmpty$4$1) {
     var lastElement$4$1 = (-6)
@@ -1776,7 +1775,6 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeForeach4__I__V = (function(n) {
     var remainder = new $c_sjsr_RuntimeLong().init___I__I(3, 0).$$plus__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I(n)).$$percent__sjsr_RuntimeLong__sjsr_RuntimeLong(new $c_sjsr_RuntimeLong().init___I__I(3, 0)).lo$2;
     var lastElement$4$1 = ((remainder !== 0) ? ((n - remainder) | 0) : n)
   };
-  var terminalElement$4$1 = ((3 + lastElement$4$1) | 0);
   inlinereturn$12: {
     if ((!isEmpty$4$1)) {
       var i = (-3);
@@ -1809,16 +1807,13 @@ $c_Lhelloworld_HelloWorld$.prototype.main__V = (function() {
 });
 $c_Lhelloworld_HelloWorld$.prototype.rangeForeach3__I__V = (function(n) {
   var isEmpty$4 = (n <= 0);
-  var numRangeElements$4 = (isEmpty$4 ? 0 : n);
+  if (isEmpty$4) {
+    /*<skip>*/
+  };
   var lastElement$4 = (isEmpty$4 ? (-1) : (((-1) + n) | 0));
-  var terminalElement$4 = ((1 + lastElement$4) | 0);
   var isEmpty$4$1 = (n <= 0);
   if (isEmpty$4$1) {
-    var numRangeElements$4$1 = 0
-  } else {
-    var x = ((n / 3) | 0);
-    var y = ((((n % 3) | 0) !== 0) ? 1 : 0);
-    var numRangeElements$4$1 = ((((x > 0) && (y > 0)) && (((x + y) | 0) < 0)) ? (-1) : ((x + y) | 0))
+    /*<skip>*/
   };
   if (isEmpty$4$1) {
     var lastElement$4$1 = (-3)
@@ -1826,7 +1821,6 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeForeach3__I__V = (function(n) {
     var remainder = ((n % 3) | 0);
     var lastElement$4$1 = ((remainder !== 0) ? ((n - remainder) | 0) : (((-3) + n) | 0))
   };
-  var terminalElement$4$1 = ((3 + lastElement$4$1) | 0);
   inlinereturn$16: {
     if ((!isEmpty$4$1)) {
       var i = 0;
@@ -1843,9 +1837,10 @@ $c_Lhelloworld_HelloWorld$.prototype.rangeForeach3__I__V = (function(n) {
 });
 $c_Lhelloworld_HelloWorld$.prototype.rangeForeach2__I__V = (function(n) {
   var isEmpty$4 = (n < (-3));
-  var numRangeElements$4 = (isEmpty$4 ? 0 : ((n > 2147483643) ? (-1) : ((4 + n) | 0)));
+  if (isEmpty$4) {
+    /*<skip>*/
+  };
   var lastElement$4 = (isEmpty$4 ? (-4) : n);
-  var terminalElement$4 = ((1 + lastElement$4) | 0);
   inlinereturn$6: {
     if ((!isEmpty$4)) {
       var i = (-3);
```
We can clearly see that `numRangeElements` and `terminalElement` are dce'ed away in all `foreach` methods, significantly improving them, particularly 3 and 4. The no-op `if (isEmpty) { /* skip */ }` should eventually be removed as well, but that is an orthogonal concern.

The only true regression is in `sum`, which does use `numRangeElements`. In the variant 3, the previous version optimizer could optimize the `Long` operations as `Int` operations, which the new version cannot do. However, I think this is a small price to pay for the general benefits (`sum` is much more rarely used than `foreach`).

FTR, the test-suite fastopt.js lost 25 KB in the process.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3580,45 +3619,90 @@ private[optimizer] object OptimizerCore {
...
+      new PreTransBlock(bindingsAndStats ::: result.bindingsAndStats,
+          result.result)
+    }
+
+    def apply(binding: PreTransBinding,
+        result: PreTransLocalDef): PreTransform = {
+      if (result.contains(binding.localDef))
+        new PreTransBlock(Left(binding) :: Nil, result)
+      else
+        result // eager dce of the binding to keep the `PreTransLocalDef` nature
```
gzm0: What do you mean by "keep the `PreTransLocalDef` nature"?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1093,6 +1122,53 @@ private[optimizer] abstract class OptimizerCore(
...
+          case ReplaceWithRecordVarRef(name, originalName, _, used, _) =>
+            (name, originalName, used)
+        }
+
+        if (used.value) {
+          val ident = Ident(name, originalName)
+          val varDef = resolveLocalDef(value) match {
+            case PreTransRecordTree(valueTree, valueTpe, cancelFun) =>
+              val recordType = valueTree.tpe.asInstanceOf[RecordType]
+              if (!isImmutableType(recordType))
```
gzm0: Can you quickly remind me why we can't inline mutable record types?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: I mean that the result keeps being a `PreTransLocalDef`, as opposed to a `PreTransBlock(..., PreTransLocalDef(...))`. Simple `PreTransLocalDef`s are better optimized in some cases than `PreTransBlock(..., PreTransLocalDef(...))` (basically because they can be duplicated everywhere we need them without risking duplicating the evaluations of the `bindingsAndStats` of the `PreTransBlock`).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: We can inline them (as in, alias them through `LocalDef`s) but we can never *copy* them, as in an `Assign` or a `VarDef`, otherwise they lose their identity. So if you change the copy, the original is not modified, and conversely.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Maybe just say "[...] to avoid unnecessary nesting in PreTransBlock for better optimization." or something like that.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ok. And since we do not know whether the `VarDef` will be used multiple times or not in the body, we must abort. Right?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1037,6 +1038,34 @@ private[optimizer] abstract class OptimizerCore(
     }
   }
 
+  /** Resolves any [[RecordType]] in a [[PreTransform]]. */
+  private def resolveRecordType(
+      preTrans: PreTransform): Option[(RecordType, CancelFun)] = {
```
gzm0: Consider naming this differently. It seems to be confusing w.r.t. `resolveLocalDef`. IIUC, this really just grabs the `CancelFun` of the record type, right? The type itself would already be available through `preTrans.tpe`, no?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes and no. Even if you use the `VarDef` only once, maybe you're *also* still using the rhs it was copied from. Consider (pseudo-scala notation):
```scala
val rec1 = { val x: Int = 5; var y: Int = 5 }
println(rec1.y) // prints 5
val rec2 = rec1
rec2.y = 3
println(rec1.y) // still prints 5
```
That happens even though I used `rec2` only once.
Whereas, if `rec1` and `rec2` had remained reference types, the last line would print 3 as well. So copying a reference and copying a mutual record value do not have the same semantics. That's why we have to abort.

To not abort here, we would have to ensure/check some kind of ownership transfer, by which `rec1` gets invalidated from then on. But that depends on the evaluation order, and checking evaluation order-dependent properties is something our IR is not good at.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: OK

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: No. `preTrans.tpe` would be the refined *original* type, not the record type it's been lowered to. For example, given
```scala
@inline class Foo(val x: Int)
val foo = new Foo(5)
```
`resolveRecordType(<foo's LocalDef>) == RecordType("x" -> RecordField(...))`
whereas
`preTrans.tpe == RefinedType(ClassType("Foo"), isExact = true, isNullable = false)`

That's why we have to do the same kind of resolution as `resolveLocalDef`, but we short-cut it so that we only find what we're interested in, rather than actually go and force the `PreTransform`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3302,9 +3327,23 @@ private[optimizer] abstract class OptimizerCore(
...
+                PreTransBlock(bindingsAndStats, tresult)
+              case PreTransRecordTree(tree, tpe, cancelFun) =>
+                PreTransRecordTree(
+                    finishTransformBindings(bindingsAndStats, tree),
+                    tpe, cancelFun)
+              case PreTransTree(tree, tpe) =>
+                PreTransTree(
+                    finishTransformBindings(bindingsAndStats, tree),
+                    tpe)
+            }
```
gzm0: Consider factoring this out with the same match above.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Makes sense. Didn't think of the case where the rhs is not a "new" value.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ahh, ok. Seen like this, the name makes sense.

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: I have added more comments in the Scaladoc of that method to avoid that confusion.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: They are not actually factorable. One works with an only `PreTransBinding` and the other with a `List[BindingOrStat]`. However I have extracted those functionalities in two helper methods defined next to each other, and with more doc, so that it's obvious they do "the same thing" but not on the same data types.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3326,6 +3345,51 @@ private[optimizer] abstract class OptimizerCore(
...
+        PreTransBlock(preTransBinding, result)
+      case PreTransRecordTree(tree, tpe, cancelFun) =>
+        PreTransRecordTree(
+            finishTransformBindings(Left(preTransBinding) :: Nil, tree),
+            tpe, cancelFun)
+      case PreTransTree(tree, tpe) =>
+        PreTransTree(
+            finishTransformBindings(Left(preTransBinding) :: Nil, tree),
+            tpe)
+    }
```
gzm0: Am I missing something, or why can't you just replace this body by: `addPreTransBinding(Left(preTransBinding) :: Nil, result)`?

gzm0: Aahhhh... Got it. Because of the optimization that triggers in the first two cases... Maybe it is worth mentioning that as a comment inline.

gzm0:
LGTM modulo a clarification comment in `addPreTransBinding`.

gzm0:
Feel free to merge after you added that.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Done.

gzm0:
LGTM

PR #2421, opened 2016-05-28
DO NOT MERGE Upgrade to GCC v20160517.

sjrd:
In theory, this version of GCC does not support JDK6 anymore, but maybe it works in practice because our use of it does not use any file I/O nor charset-based encoding/decoding.

CI only for now. I just want to see what breaks.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -571,7 +571,7 @@ object Build {
       base = file("tools/jvm"),
       settings = commonToolsSettings ++ Seq(
           libraryDependencies ++= Seq(
-              "com.google.javascript" % "closure-compiler" % "v20130603",
+              "com.google.javascript" % "closure-compiler" % "v20160517",
               "com.googlecode.json-simple" % "json-simple" % "1.1.1" exclude("junit", "junit"),
               "com.novocode" % "junit-interface" % "0.9" % "test"
           )
@@ -1529,7 +1529,7 @@ object Build {
               Seq(
                 "org.scala-sbt" % "sbt" % sbtVersion.value,
                 "org.scala-lang.modules" %% "scala-partest" % "1.0.13",
-                "com.google.javascript" % "closure-compiler" % "v20130603",
+                "com.google.javascript" % "closure-compiler" % "v20160517",
                 "io.apigee" % "rhino" % "1.7R5pre4",
                 "com.googlecode.json-simple" % "json-simple" % "1.1.1" exclude("junit", "junit")
               )
--- a/project/build.sbt
+++ b/project/build.sbt
@@ -11,7 +11,7 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.8")
 
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "0.8.0")
 
-libraryDependencies += "com.google.javascript" % "closure-compiler" % "v20130603"
+libraryDependencies += "com.google.javascript" % "closure-compiler" % "v20160517"
 
 libraryDependencies += "io.apigee" % "rhino" % "1.7R5pre4"
 
--- a/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureAstBuilder.scala
+++ b/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureAstBuilder.scala
@@ -7,7 +7,6 @@ import org.scalajs.core.tools.javascript.Trees.Tree
 import org.scalajs.core.tools.javascript.JSTreeBuilder
 
 import com.google.javascript.rhino._
-import com.google.javascript.rhino.jstype.{StaticSourceFile, SimpleSourceFile}
 import com.google.javascript.jscomp._
 
 import scala.collection.mutable
--- a/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureAstTransformer.scala
+++ b/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureAstTransformer.scala
@@ -264,7 +264,7 @@ private[closure] class ClosureAstTransformer(relativizeBaseURI: Option[URI]) {
     @inline def ctorDoc(node: Node) = {
       val b = new JSDocInfoBuilder(false)
       b.recordConstructor()
-      b.build(node)
+      b.build()
     }
 
     val block = new Node(Token.BLOCK)


sjrd:
Well ... that happens:
```
[info] Full optimizing /localhome/jenkins/a/workspace/scalajs-task-worker_2/examples/helloworld/target/scala-2.11/helloworld-opt.js
java.lang.UnsupportedClassVersionError: com/google/javascript/jscomp/ErrorManager : Unsupported major.minor version 51.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClassCond(ClassLoader.java:631)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:615)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:141)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:283)
	at java.net.URLClassLoader.access$000(URLClassLoader.java:58)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:197)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:295)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
	at org.scalajs.core.tools.linker.LinkerPlatformExtensions$class.apply(LinkerPlatformExtensions.scala:43)
	at org.scalajs.core.tools.linker.Linker$.apply(Linker.scala:76)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$1$$anonfun$11.apply(ScalaJSPluginInternal.scala:195)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$1$$anonfun$11.apply(ScalaJSPluginInternal.scala:194)
	at org.scalajs.core.tools.linker.ClearableLinker.org$scalajs$core$tools$linker$ClearableLinker$$ensureLinker(ClearableLinker.scala:78)
	at org.scalajs.core.tools.linker.ClearableLinker.linkerOp(ClearableLinker.scala:59)
	at org.scalajs.core.tools.linker.ClearableLinker.link(ClearableLinker.scala:51)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$4$$anonfun$apply$6$$anonfun$apply$7.apply(ScalaJSPluginInternal.scala:241)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$4$$anonfun$apply$6$$anonfun$apply$7.apply(ScalaJSPluginInternal.scala:229)
	at sbt.FileFunction$$anonfun$cached$1.apply(Tracked.scala:235)
	at sbt.FileFunction$$anonfun$cached$1.apply(Tracked.scala:235)
	at sbt.FileFunction$$anonfun$cached$2$$anonfun$apply$3$$anonfun$apply$4.apply(Tracked.scala:249)
	at sbt.FileFunction$$anonfun$cached$2$$anonfun$apply$3$$anonfun$apply$4.apply(Tracked.scala:245)
	at sbt.Difference.apply(Tracked.scala:224)
	at sbt.Difference.apply(Tracked.scala:206)
	at sbt.FileFunction$$anonfun$cached$2$$anonfun$apply$3.apply(Tracked.scala:245)
	at sbt.FileFunction$$anonfun$cached$2$$anonfun$apply$3.apply(Tracked.scala:244)
	at sbt.Difference.apply(Tracked.scala:224)
	at sbt.Difference.apply(Tracked.scala:200)
	at sbt.FileFunction$$anonfun$cached$2.apply(Tracked.scala:244)
	at sbt.FileFunction$$anonfun$cached$2.apply(Tracked.scala:242)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$4$$anonfun$apply$6.apply(ScalaJSPluginInternal.scala:246)
	at org.scalajs.sbtplugin.ScalaJSPluginInternal$$anonfun$org$scalajs$sbtplugin$ScalaJSPluginInternal$$scalaJSStageSettings$4$$anonfun$apply$6.apply(ScalaJSPluginInternal.scala:227)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:303)
	at java.util.concurrent.FutureTask.run(FutureTask.java:138)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:441)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:303)
	at java.util.concurrent.FutureTask.run(FutureTask.java:138)
	at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:886)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:908)
	at java.lang.Thread.run(Thread.java:662)
[error] (helloworld/compile:fullOptJS) java.lang.UnsupportedClassVersionError: com/google/javascript/jscomp/ErrorManager : Unsupported major.minor version 51.0
[error] Total time: 3 s, completed May 28, 2016 10:13:20 PM
```
@gzm0 @nicolasstucki : I'm thinking we should consider maintaining our own fork of GCC, under a different namespace, with class files that can be read by Java 6. Seriously. WDYT?

gzm0:
> I'm thinking we should consider maintaining our own fork of GCC.

Probably necessary, indeed. Because being stuck at some version forever is not going to cut it.

Not sure about the different namespace though, it might cause considerable overhead.

I guess the ideal would be a striped down, but 100% compatible artifact in the same namespace but running on JDK6.

Also, do we have any idea on the cost of maintaining such a fork?

sjrd:
We could try keeping the same namespace, in fact, yes. I didn't consider it because I thought that'd be crazy, but in fact it could work, indeed. It's only a bit more awkward if someone depends on the "true" one through another way; then they need some `dependencyFilter` magic to remove our fork, but that's doable.

> Also, do we have any idea on the cost of maintaining such a fork?

Assuming same-namespace, my recommendation would be only to change the `javac` options so that it emits Java 6-compatible .class files. They could still refer to JDK 7 APIs: as long as they're reached at run-time, the lazy linking of the JVM will make it work.

The safer alternative would be to change the CI script to build on openjdk6. In that case, the overhead of maintaining the fork might be bigger, as it will involve merging or rebasing the actual code. Hopefully, only a very shallow part of the API is concerned (and I hope nothing in the actual compiler passes).

sjrd:
Tests results: on JDK 7 and 8, only checksizes fail by 1KB of additional -opt.js (but not -opt.js.gz). So at least we don't have any *other* issue besides JDK 6.

gzm0:
> Assuming same-namespace, my recommendation would be only to change the javac options so that it emits Java 6-compatible .class files.

I think this is indeed an acceptable tradeoff. Especially since we are mostly not doing this for the users that actually `fullOptJS` in the end, but for community build, etc.

> then they need some dependencyFilter magic to remove our fork, but that's doable.

Ah, I didn't think of that. Since we won't have the same org, sbt won't automatically evict our dependency, so an older one could be picked up instead...
This is an argument for a different namespace... But that will just make maintenance so much harder :(

siriux:
As the changes needed to use the new version are minimal (provided you use JDK 7+), could this be provided as an option under a experimental flag util we have a fork running on JDK6?

This would be an instant improvement to most people using JDK7+ without compromising the community builds.

sjrd:
No that's not possible. The changes are minimal but still incompatible. We cannot *compile* our tools in a way that is compatible with both versions of gcc.

siriux:
That's a pity, we will have to wait for the real solution.

siriux:
I tried an alternative way of getting GCC for JDK6.

I've downloaded the last version of GCC from the web in a single jar, and I've used proguard to retarget to java 6. I don't have java 6, so I didn't test it, but I think this should have the same chances to work as a fork of GCC.

We would still need to upload it to maven, but that's much less work than maintaining a fork.

What do you think?

The jar is: https://dl.google.com/closure-compiler/compiler-latest.zip

The proguard config file is:

```
-injars compiler.jar
-outjars compiler_jdk6.jar
-libraryjars /usr/lib/jvm/java-8-openjdk/jre/lib/rt.jar:/usr/share/java/apache-ant/ant.jar
-dontwarn com.google.j2objc.**
-dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
-dontshrink
-dontoptimize
-dontobfuscate
-target 6
```

sjrd:
That's worth exploring, indeed. Thanks.

sjrd:
Superseded by #2490.

PR #2423, opened 2016-05-30
Make HTML runner part of sbt plugin

gzm0:


ci/matrix.xml
```diff
@@ -245,7 +245,10 @@
     sbt ++2.10.6 ir/publishLocal tools/publishLocal jsEnvs/publishLocal \
                  testAdapter/publishLocal sbtPlugin/publishLocal &&
     cd sbt-plugin-test &&
-    sbt noDOM/run withDOM/run multiTestJS/test:run test \
+    sbt noDOM/run noDOM/testHtmlFastOpt noDOM/testHtmlFullOpt \
+        withDOM/run withDOM/testHtmlFastOpt withDOM/testHtmlFullOpt \
+        multiTestJS/test:run multiTestJS/testHtmlFastOpt multiTestJS/testHtmlFullOpt \
+	test \
```
sjrd: Indentation issue (tab)

project/Build.scala
```diff
@@ -582,12 +579,23 @@ object Build {
...
-            ProvidedJS / artifact.getName % "test"
-          }
+          resourceGenerators in Test <+= Def.task {
+            val base = (resourceManaged in Compile).value
+            IO.createDirectory(base)
+            val outFile = base / "js-test-definitions.js"
+
+            val writer = new BufferedWriter(new FileWriter(outFile))
+            try {
+              org.scalajs.sbtplugin.HTMLRunnerTemplate.renderTestDefinitions(
```
sjrd: I guess this is the internal reason for which `HTMLRunnerTemplate` is public?
If that's the only reason, we can avoid it by making it `private[sbtplugin]`, and have a forwarder in `project/HTMLRunnerTemplateAccess.scala` like this:
```scala
package org.scalajs.sbtplugin

object HTMLRunnerTemplateAccess {
  def renderTestDefinitions(...): Unit =
    HTMLRunnerTemplate.renderTestDefinitions(...)
}
```
and use that from `Build.scala`.

Ugly, but less ugly than having to publicly expose something that should not be public.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
```diff
@@ -0,0 +1,97 @@
...
+        </script>
+        <script type="text/javascript" src="""")
+    writeHtmlEscaped(jsdepsFile.toURI.toASCIIString, out)
+    // Close: <script>. Open: <script> for SJS.
+    out.write(""""></script>
+        <script type="text/javascript" src="""")
+    writeHtmlEscaped(sjsFile.toURI.toASCIIString, out)
+    // Close: <script>. Open: <script> for test definitions.
+    out.write(""""></script>
+        <script type="text/javascript">""")
```
sjrd: Consider building all of the above as one `s""" ... """` interpolator, splicing things like `${htmlEscape(title)}` and `${css.toURI.toASCIIString}`.

The `renderTestDefinitions` should probably stay outside, as it's building a rather big string (and uses `writeJSON` anyway). But the shortest things above can be safely spliced.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
```diff
@@ -0,0 +1,97 @@
...
+      </body>
+      </html>""");
+  }
+
+  def renderTestDefinitions(
+      loadedFrameworks: Map[sbt.TestFramework, Framework],
+      definedTests: Seq[sbt.TestDefinition], out: Writer): Unit = {
+    val frameworks = loadedFrameworks.map(_._1.implClassNames.toList).toList
+
+    val tests = definedTests.map(t =>
```
sjrd: Use `.map {}` for that multi-line lambda.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -147,6 +147,12 @@ object ScalaJSPlugin extends AutoPlugin {
     val fullOptJS = TaskKey[Attributed[File]]("fullOptJS",
         "Link all compiled JavaScript into a single file and fully optimize", APlusTask)
 
+    val testHtmlFastOpt = TaskKey[Attributed[File]]("testHtmlFastOpt",
+        "Create a html test runner for fastOptJS", AMinusTask)
```
sjrd: a html -> an HTML

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -147,6 +147,12 @@ object ScalaJSPlugin extends AutoPlugin {
...
+
+    val testHtmlFullOpt = TaskKey[Attributed[File]]("testHtmlFullOpt",
+        "Create a html test runner for fullOptJS", AMinusTask)
```
sjrd: Same: a html -> an HTML

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -753,10 +753,66 @@ object ScalaJSPluginInternal {
       }
   )
 
+  def scalaJSTestHtmlTaskSettings(testHtmlKey: TaskKey[Attributed[File]],
+      sjsKey: TaskKey[Attributed[File]], jsdepsKey: TaskKey[File]) = {
```
sjrd: Public method needs result type.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -753,10 +753,66 @@ object ScalaJSPluginInternal {
...
+    testHtmlKey := {
+      if ((skip in jsdepsKey).value) {
+        sys.error(s"(skip in $jsdepsKey) must be false for $testHtmlKey.")
+      }
+
+      val log = (streams in testHtmlKey).value.log
```
sjrd: `streams` should not scoped. It is automatically scoped by sbt's magic.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -753,10 +753,66 @@ object ScalaJSPluginInternal {
...
+  }
+
+  val scalaJSTestHtmlSettings = Seq(
+      artifactPath in testHtmlFastOpt :=
+        ((crossTarget in testHtmlFastOpt).value /
+            ((moduleName in testHtmlFastOpt).value + "-fastopt-test.html")),
+      artifactPath in testHtmlFullOpt :=
+        ((crossTarget in testHtmlFullOpt).value /
+            ((moduleName in testHtmlFullOpt).value + "-opt-test.html"))
+
```
sjrd: Spurious blank line.

sjrd:
That's all. Nice!

gzm0:
As @yawnt figured out, we need to force the `test-interface` dependency to be at the newest version. Otherwise, if an "old" test framework is linked in, the new shiny HTML stuff will simply not be around.

Current "workaround":
```
libraryDependencies += "org.scala-js" %% "scalajs-test-interface" % "0.6.10-SNAPSHOT" % "test"
```

project/Build.scala
gzm0: Yeah, probably. I'm actually still considering whether we should write the JS test definitions separately. Then we could also make the `TestDetector` a first-class citizen. But we should probably wait until somebody asks us to.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
gzm0: Done.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
gzm0: Done.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Done.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Done.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: This should probably not be public anyways..

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Ok...

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Done.

ci/matrix.xml
gzm0: Done.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
gzm0: I made everything spliced. Should probably be fine. We can still optimize if it becomes an issue.

gzm0:
Updated and rebased.

gzm0:
Still need to do the `libraryDependencies` workaround.

```
libraryDependencies += "org.scala-js" %% "scalajs-test-interface" % "0.6.10-SNAPSHOT" % "test"
```

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TaskDefSerializers.scala
```diff
@@ -16,7 +16,7 @@ import org.scalajs.core.tools.json._
 import FingerprintSerializers._
 import SelectorSerializers._
 
-object TaskDefSerializers {
+private[scalajs] object TaskDefSerializers {
```
gzm0: Note this change which is new.

sjrd: Isn't that a source incompatibility with previously released versions?

sjrd:
I don't have anything else to add. I'll wait for the `libraryDependencies` thing. It might prove to be slightly tricky, that one. It's hard to justify injecting `testinterface` in the `Test` configuration of any project (after all, a project could use the `Test` configuration in its own way, independently of sbt's testing infratructure). But at the same time we can't inject it only when we run that task (it wouldn't make sense anyway).

At least, the task itself is only relevant if the `Test` config is indeed used with sbt's testing interface, in one way or another. Maybe the task could simply do a sanity check. It would verify that (a recent enough) version of `scalajs-test-interface` is on the classpath, and would display a nice error message to the developer if it's not the case (with the solution to their problem, i.e., add the "workaround" to their `settings`).

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TaskDefSerializers.scala
gzm0: Ah... I thought we haven't released this. But indeed, [the reverse change][1] is in v0.6.9... :-/
So I guess I shouldn't do this.

[1]: https://github.com/scala-js/scala-js/commit/1d775830780750217f99c14ea73340d4fa06b540#diff-36e6887d1a47947dbb3a572909fc3a6d is

sjrd: You could deprecate it.

sjrd: Or we don't care and just say breaking change. After all, we don't have very strict rules about the compatibility of the test-adapter.

gzm0: Agreed.

gzm0:
Updated.

gzm0:
Running manual tests of the HTML runners.

sjrd:
LGTM, assuming the manual tests passed.

gzm0:
I'm unable to build `testSuite/test:fullOptJS` on my little laptop :(
All other tests pass. I'll build on a decent machine tomorrow.

gzm0:
I'm unable to build a `fullOptJS` on any machine I have. It always OOMs (with `-Xmx8G`).

However, CI seems to do just fine. Any idea what is going on?

sjrd:
I can build it without problem on my machine (although it takes 3 minutes). I do have some failures in `testHtmlFullOpt` that seem to indicate that the `fullopt-stage` property is not discovered correctly; and/or the `LinkingInfo.productionMode`:
```
Test org.scalajs.testsuite.jsinterop.MiscInteropTest.should_have_a_meaningful_name_property started
Test org.scalajs.testsuite.jsinterop.MiscInteropTest.should_have_a_meaningful_name_property failed: null, took 0.001544999 sec
Test org.scalajs.testsuite.jsinterop.MiscInteropTest.should_have_a_meaningful_name_property finished, took 0.001544999 sec

Test org.scalajs.testsuite.compiler.OptimizerTest.must_not_break_when_folding_double_with_decimal_and_stringLit started
Test org.scalajs.testsuite.compiler.OptimizerTest.must_not_break_when_folding_double_with_decimal_and_stringLit failed: expected:<1.2323919403474454[e]+21hello> but was:<1.2323919403474454[]+21hello>, took 0.002692999 sec
Test org.scalajs.testsuite.compiler.OptimizerTest.must_not_break_when_folding_double_with_decimal_and_stringLit finished, took 0.002692999 sec

Test org.scalajs.testsuite.jsinterop.RuntimeLongTest.toDouble started
Test org.scalajs.testsuite.jsinterop.RuntimeLongTest.toDouble failed: expected:<9223372036854775000> but was:<9223372036854776000>, took 0.003561 sec
Test org.scalajs.testsuite.jsinterop.RuntimeLongTest.toDouble finished, took 0.003561 sec

Test run started
Test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_development_mode started
Test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_development_mode failed: null, took 0.000938 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_development_mode finished, took 0.000938 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_production_mode started
Test assumption in test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_production_mode failed: org.junit.internal.AssumptionViolatedException: Assumed in production mode, took 0.000998001 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.developmentMode_when_in_production_mode finished, took 0.000998001 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_development_mode started
Test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_development_mode failed: null, took 0.000814999 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_development_mode finished, took 0.000814999 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_production_mode started
Test assumption in test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_production_mode failed: org.junit.internal.AssumptionViolatedException: Assumed in production mode, took 0.000856999 sec
Test org.scalajs.testsuite.library.LinkingInfoTest.productionMode_when_in_production_mode finished, took 0.000856999 sec
Test run finished: 2 failed, 0 ignored, 4 total, 0.011087s
```

gzm0:
The problem is that test tags are based on the `scalaJSStage` setting. However, `testHtmlFullOpt` does not change these of course. I need to think how we can solve this.
Luckily, this is just a problem in our build.

gzm0:
Ok. If I do:

```
set scalaJSStage in testSuite := FullOptStage
testSuite/testSuite/testHtmlFullOpt
```

the test suite passes.

sjrd:
So are you suggesting we should always do that in our build?

gzm0:
No, I'm not sure. Essentially this points at a fundamental flaw in the `testHTMLFastOpt` / `testHTMLFullOpt` tasks:

The linking pipeline is laid out in parallel in two flavors: fastOpt and fullOpt. However, from there we use an sbt setting to feed the one we want to dependent tasks (running and testing). Therefore, all setting relevant to run and tests tasks (notably `javaOptions` and derivatives) are not specific to a given stage (we did have that a long time ago :)).

Hence there seems to be no way we can get the right settings when executing `testHTMLFullOpt` if currently `scalaJSStage` is set to `fastOptStage`.

I think we have a couple of options here:

1. Leave it the way it is, document it, let users deal with it.
2. Fail if `scalaJSStage` does not correspond to the version of `testHTML*` we are running.
3. Only provide a single task `testHTML` that itself depends on `scalaJSStage`. To what file (and whether it should be a different one based on the stage) this task should write is TBD.

I don't like any of these suggestions. Ideas, comments?

sjrd:
I think 1+2 (they're not exclusive) is a reasonable option.

gzm0:
> I think 1+2 (they're not exclusive) is a reasonable option.

So you mean just add a check that `scalaJSStage` is set to the right stage. What I do not like about this, is that it is going to annoy the 99% (blatantly made up value) of our users that do not have `scalaJSStage` specific builds.

Or did you mean to make the check opt-in?

sjrd:
Ah, I had not thought about that. Then I guess 1 is the way to go.

gzm0:
> Then I guess 1 is the way to go.

Ok. So I guess I'll add a check just for our build. That seems better than simply documenting that one has to set `scalaJSStage = FullOptStage` before running `testHTMLFullOpt`.

gzm0:
Updated

gzm0:
Running HTML runner tests.

gzm0:
HTML runner tests pass.

sjrd:
LGTM :tada:

gzm0:
Woot

Issue #2437, opened 2016-06-03
Exports in traits are not picked up as of 2.12.0-M5


sjrd:
As of 2.12.0-M5, concrete trait methods are not always mixed in classes anymore, if the normal default method resolution would resolve to the correct target. This means that `@JSExport`ed trait methods do not always get a forwarder in classes, and therefore they are not picked up by the export method generator.

PR #2438, opened 2016-06-03
Fix #2434: Support reflective calls to default methods.

### LINKED ISSUE #2434: Reflective calls to default methods crash the linker
The following test crashes the linker:
```scala
  @Test def reflectiveCallDefaultMethod(): Unit = {
    import scala.language.reflectiveCalls

    class ReflectiveCallIntComparator extends ju.Comparator[Int] {
      def compare(o1: Int, o2: Int): Int =
        o1.compareTo(o2)
    }

    val c = new ReflectiveCallIntComparator
    val c2: { def reversed(): ju.Comparator[Int] } = c
    val reversed = c2.reversed()

    assertTrue(reversed.compare(5, 7) > 0)
  }
```

sjrd:
There were two problems.

First, the Analyzer did not make any effort regarding default methods when looking for targets of reflective calls. Depending on the order of lookups, the algorithm could find existing default bridges, or fail. Now, default bridges are ignored when looking for potential targets, but implemented interfaces are searched for default methods as potential targets. If a default method is found, the creation of the reflective proxy will (automatically) subsequently trigger a lookup for the target method, which will synthesize a default bridge too.

Second, the BaseLinker crashed if the lookup resolved (by chance) to a default bridge, as it tried to find an actual `MethodDef` for the bridge in the class' `Tree`. Now, if the target of a reflective proxy is a default bridge, the bridge is "dereferenced" to its original default method to find a `MethodDef`.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
@@ -416,47 +416,43 @@ private final class Analyzer(semantics: Semantics,
       if (!allowAddingSyntheticMethods) {
         tryLookupMethod(proxyName)
       } else {
-        /* The lookup for a target method in this code implements the
-         * algorithm defining `java.lang.Class.getMethod`. This mimics how
-         * reflective calls are implemented on the JVM, at link time.
-         *
-         * Caveat: protected methods are not ignored. This can only make an
-         * otherwise invalid reflective call suddenly able to call a protected
-         * method. It never breaks valid reflective calls. This could be fixed
-         * if the IR retained the information that a method is protected.
-         */
-
-        @tailrec
-        def loop(ancestorInfo: ClassInfo): Option[MethodInfo] = {
-          if (ancestorInfo ne null) {
-            ancestorInfo.methodInfos.get(proxyName) match {
-              case Some(m) =>
-                assert(m.isReflProxy && !m.isAbstract)
-                Some(m)
-
-              case _ =>
-                ancestorInfo.findProxyMatch(proxyName) match {
-                  case Some(target) =>
-                    val targetName = target.encodedName
-                    Some(ancestorInfo.createReflProxy(proxyName, targetName))
-
-                  case None =>
-                    loop(ancestorInfo.superClass)
-                }
-            }
-          } else {
-            None
+        methodInfos.get(proxyName).orElse {
+          findReflectiveTarget(proxyName).map { reflectiveTarget =>
+            createReflProxy(proxyName, reflectiveTarget.encodedName)
           }
         }
+      }
+    }
 
-        loop(this)
+    private def findReflectiveTarget(proxyName: String): Option[MethodInfo] = {
+      /* The lookup for a target method in this code implements the
+       * algorithm defining `java.lang.Class.getMethod`. This mimics how
+       * reflective calls are implemented on the JVM, at link time.
+       *
+       * We add a bit of guess-work for default methods, as the documentation
+       * is very vague about them. Basically, we just take the first match in
+       * `ancestors`, as it's easy, and we're in a gray area anyway. At least,
+       * this will work when there is no overload.
+       *
+       * Caveat: protected methods are not ignored. This can only make an
+       * otherwise invalid reflective call suddenly able to call a protected
+       * method. It never breaks valid reflective calls. This could be fixed
+       * if the IR retained the information that a method is protected.
+       */
+
+      val superClasses =
+        Iterator.iterate(this)(_.superClass).takeWhile(_ ne null)
+      val superClassesThenAncestors = superClasses ++ ancestors.iterator
+
+      superClassesThenAncestors.map(_.findProxyMatch(proxyName)).collectFirst {
+        case Some(m) => m
       }
     }
 
     private def findProxyMatch(proxyName: String): Option[MethodInfo] = {
       val candidates = methodInfos.valuesIterator.filter { m =>
         // TODO In theory we should filter out protected methods
-        !m.isReflProxy && !m.isExported && !m.isAbstract &&
+        !m.isReflProxy && !m.isDefaultBridge && !m.isExported && !m.isAbstract &&
         reflProxyMatches(m.encodedName, proxyName)
       }.toSeq
 
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
@@ -315,7 +315,7 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
       implicit pos: Position): MethodDef = {
     val encodedName = methodInfo.encodedName
 
-    val inheritedMDef = findInheritedMethodDef(classInfo.superClass,
+    val inheritedMDef = findInheritedMethodDef(analysis, classInfo.superClass,
         encodedName, getTree, _.syntheticKind == MethodSyntheticKind.None)
 
     val origName = inheritedMDef.name.asInstanceOf[Ident].originalName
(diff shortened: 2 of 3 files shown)


sjrd:
Review by @gzm0 

/cc @lrytz This fixes the `AssertionError`. There are still 4 tests failing due to #2437.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -416,47 +416,55 @@ private final class Analyzer(semantics: Semantics,
...
+          }
+        } else {
+          // Look in default methods
+          ancestors.iterator.map(_.findProxyMatch(proxyName)).collectFirst {
+            case Some(m) => m
+          }
+        }
       }
+
+      loopParentClasses(this)
```
gzm0: How about:

```scala
val superClasses = Iterator.iterate(this)(_.superClass).takeWhile(_ ne null)
val candidates = superClasses.flatMap(_.findProxyMatch(proxyName)) ++
    ancestors.iterator.flatMap(_.findProxyMatch(proxyName))

candidates.headOption
```

Does the same (probably at the same speed as well) but doesn't mix two styles (tailrec and iterator) and is much shorter.

sjrd: OK. Anyway I don't really care about speed when reflective proxies are involved.

sjrd: It turns out `Iterator` doesn't have `headOption`, so I used `.map().collectFirst` as before. But otherwise I used your suggestion.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -337,7 +337,8 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
       analysis: Analysis): MethodDef = {
     val encodedName = methodInfo.encodedName
 
-    val targetMDef = findInheritedMethodDef(classInfo, targetName, getTree)
+    val targetMDef = findInheritedMethodDef(analysis, classInfo, targetName,
+        getTree)
```
gzm0: I do not understand why we need this. Isn't the `targetName` resolved to something that actually implements the method (as in: not via synthetic method). So we should be able to just call it there.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -416,47 +416,43 @@ private final class Analyzer(semantics: Semantics,
...
-
-                  case None =>
-                    loop(ancestorInfo.superClass)
-                }
-            }
-          } else {
-            None
+        methodInfos.get(proxyName).orElse {
+          findReflectiveTarget(proxyName).map { reflectiveTarget =>
+            createReflProxy(proxyName, reflectiveTarget.encodedName)
```
gzm0: Do we really want to create the reflexive proxy on the leaf node in the hierarchy? IIUC, if we have a base class `Base` with a method `foo` and three subclasses `A`, `B`, `C, that do not override `foo`, a call to `foo` on `Base` would generate a reflexive proxy in every class that is instantiated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -388,19 +389,34 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
...
-        loop(ancestorInfo.superClass)
+      inherited.find(p) match {
+        case Some(m) =>
+          m.syntheticKind match {
+            case MethodSyntheticKind.None =>
+              findMethodDef(ancestorInfo, methodName, getTree)
+
+            case MethodSyntheticKind.DefaultBridge(targetInterface) =>
+              val targetInterfaceInfo = analysis.classInfos(targetInterface)
+              findMethodDef(targetInterfaceInfo, methodName, getTree)
```
gzm0: Can you make an example when we actually have to follow a default bridge? IIUC, the analyzer would resolve the reflective call to the default method directly to the interface. So the target of such a reflective proxy would be the default method implementation, not a bridge...

gzm0:
I think this needs a round of explanation. Because either I got this wrong completely and have to spend much more time understanding the analyzer and the linker, or something is seriously wrong here... :-/

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
sjrd: Not always. We could, in theory, try and move the declaration as far up as possible, i.e., at the last superclass before an ancestor where it's either not resolvable, or resolves to a different target name. However, this requires re-resolving at every level of the inheritance chain; and in case the declaration can be moved up, we'll need to *redo* the whole process every single time we call that proxy (because we won't have a trace in the current class that we resolved it here already).

Previously, we could get away with a simple scheme where we would define it at the level where we found the target. This was fine with class methods only. But it doesn't work when default methods are introduced.

Therefore, we keep it simple here, if not optimal, and always create the proxies at the leaves, which is safe. Note that this is actually also what we do for default bridges, so far. In the future, we could optimize this, in which case we should probably do it for both default bridges and reflective proxies.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
sjrd: It happens all the time. That's because the reflective proxy is generated based on the target *name*, which causes a default bridge to be emitted for the reflective proxy. Consider this code:
```scala
trait Foo {
  def foo(): Int = 5
}

class Bar extends Foo

def test(bar: { def foo(): Int }): Unit =
  println(bar.foo())
```
The reflective call in `test` will try to find a reflective proxy for `foo__` in `Bar`. Since there is none yet, we look for a target. The target that is found is `Foo.foo`, *so the target name* is `foo__I`. Once the target name is found, we can generate the reflective proxy in `Bar`, which is:
```scala
class Bar extends Foo {
  def foo__(): any = this.foo__I()
}
```
Note that we remembered the *name* `foo__I`, but we "forgot" where we found it.

Now, that, in turns, is made reachable, and so we have a dynamic call to `Bar.foo__I`. That is not found either! So we look for a default method, and we find `Foo.foo__I`. We therefore generate a default bridge for that method in `Bar`, which becomes:
```scala
class Bar extends Foo {
  def foo__(): any = this.foo__I()

  def foo__I(): int = this.Foo::foo__I()
}
```
The default bridge `foo__I` knows that it's a bridge to `Foo.foo__I`. However, the reflective proxy only knows that it's a proxy to a method named `foo__I`, but we do not know where it is defined.

Therefore, when we come in the `BaseLinker`, we try and find the target of `Bar.foo__`, and we find, through inheritance, `Bar.foo__I`. That, in turn, is a default bridge, which we have to follow (through a direct link) to `Foo.foo__I`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
sjrd: See my other comment. The reflective proxy only knows a method *name*. It does not know in which class/interface it is actually implemented. And that means, at this point, it can resolve to a synthetic default bridge.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
gzm0: Ok

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
gzm0: Ah, that's what I got wrong. An the only reason we traverse through the superclasses is to reconstruct the argument types. Not for actual dispatch.

We should consider (in a future PR), to change the contents of the subclasses of `MethodSyntheticKind` to contain references to the actual `MethodInfos`, rather than just strings. This would not add any overhead in the `Analyzer`, but would simplify the `BaseLinker`.

gzm0:
LGTM

Issue #2456, opened 2016-06-07
Rewrite multiplications by powers of 2 as shifts


sjrd:
Typical examples:
* `x * 4` -> `x << 2`
* ~~`x / 4` -> `x >> 2`~~ (division does not work, see below)

This might seem like an optimization that the JS VM could do on its own, but not quite. For multiplication of `Int`s, the JS code we have is `$imul(4, x)`, which, although `$imul` is a built-in, requires more work from the JIT generated code than a simple operator `x << 2`.

For multiplication and division of `Long`s, this is obviously a big win, as the implementations of shifts are much simpler than multiply and divide.

For division of `Int`s, it's less clear that it's a win, but it makes sense to apply the rewriting there too, if only for symmetry. It can't hurt, at least (and it requires less characters in the generated JS file anyway).

sjrd:
As was observed during the prep talk, using `>>` for division is not correct, as it does not work for negative left-hand-sides. `/` is supposed to perform *truncating* division, but `>>` performs *floor* division.

sjrd:
One thing we can implement in the process is folding `0 * n` to `0`, though. It will also provide a good explanation of why `PreTransform`s provide a framework for local dead code elimination.

PR #2458, opened 2016-06-07
Add PreTransUnaryOp and PreTransBinaryOp in the optimizer.

sjrd:
Previously, the optimizer would be stupid on the following code:
```scala
val x: int = foo()
val y: int = x +[int] 4
val z: int = y *[int] 2
// don't use any of x, y, z
```
When performing the `withNewLocalDef` for `y`, it would have to reify `x` to put it in a `BinaryOp(Int_+, x, 4)`. Then, for the `withNewLocalDef` of `z`, it would repeat the process with `y`. Only `z` would remain virtual as a `PreTransLocalDef`.

Later on, when we discover that `z` is not used in the body of the val, it is eliminated, and the optimizer keeps only the side effects of its right-hand side, which collapses to `Skip()`.

However, `y` has already been marked as *used* a long time ago, when the right hand size of `z` was first translated. It is therefore kept, although it is clearly dead code. The same story happens to `x`, obviously.

This commit fixes this shortcoming of the optimizer by creating new `PreTransform`s for unary and binary ops. The rhs of `y` is therefore kept as a virtual `PreTransBinaryOp(Int_+, PreTransLocalDef(x), PreTransTree(4))` which does not need to reify `x`. The same happens to `z` with a virtual reference to `y`. After the body is translated, all local variables can now be removed!

This only came up recently because of the sheer ability to keep `PreTransBinding`s, which was added in 9bfc69c2a39ec3d9cab219a57c9e230cd5daa278. Now, it becomes clear that we need `PreTransform`s for all the kinds of `Tree`s that can be side-effect-free. This commit is a first (or second) step towards that goal.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -455,13 +455,15 @@ private[optimizer] abstract class OptimizerCore(
               finishTransform(isStat))
         }
 
-      case tree @ UnaryOp(_, arg) =>
-        if (isStat) transformStat(arg)
-        else transformUnaryOp(tree)
+      case tree: UnaryOp =>
+        trampoline {
+          pretransformUnaryOp(tree)(finishTransform(isStat))
+        }
 
-      case tree @ BinaryOp(op, lhs, rhs) =>
-        if (isStat) Block(transformStat(lhs), transformStat(rhs))
-        else transformBinaryOp(tree)
+      case tree: BinaryOp =>
+        trampoline {
+          pretransformBinaryOp(tree)(finishTransform(isStat))
+        }
 
       case NewArray(tpe, lengths) =>
         NewArray(tpe, lengths map transformExpr)
@@ -714,14 +716,14 @@ private[optimizer] abstract class OptimizerCore(
             sys.error(s"Cannot find local def '$name' at $pos\n" +
                 s"While optimizing $myself\n" +
                 s"Env is ${scope.env}\nInlining ${scope.implsBeingInlined}"))
-        cont(PreTransLocalDef(localDef))
+        cont(localDef.toPreTransform)
 
       case This() =>
         val localDef = scope.env.localDefs.getOrElse("this",
             sys.error(s"Found invalid 'this' at $pos\n" +
                 s"While optimizing $myself\n" +
                 s"Env is ${scope.env}\nInlining ${scope.implsBeingInlined}"))
-        cont(PreTransLocalDef(localDef))
+        cont(localDef.toPreTransform)
 
       case If(cond, thenp, elsep) =>
         val newCond = transformExpr(cond)
@@ -766,8 +768,8 @@ private[optimizer] abstract class OptimizerCore(
                       val newElsep = finishTransformExpr(telsep)
                       val refinedType =
                         constrainedLub(newThenp.tpe, newElsep.tpe, tree.tpe)
-                      cont(PreTransTree(
-                          foldIf(newCond, newThenp, newElsep)(refinedType)))
+                      cont(foldIf(newCond, newThenp, newElsep)(
+                          refinedType).toPreTransform)
                   }
                 }
               }
@@ -776,8 +778,8 @@ private[optimizer] abstract class OptimizerCore(
               val newElsep = transformExpr(elsep)
               val refinedType =
                 constrainedLub(newThenp.tpe, newElsep.tpe, tree.tpe)
-              cont(PreTransTree(
-                  foldIf(newCond, newThenp, newElsep)(refinedType)))
+              cont(foldIf(newCond, newThenp, newElsep)(
+                  refinedType).toPreTransform)
             }
         }
 
@@ -790,9 +792,9 @@ private[optimizer] abstract class OptimizerCore(
             } getOrElse default
             pretransformExpr(body)(cont)
           case _ =>
-            cont(PreTransTree(Match(newSelector,
+            cont(Match(newSelector,
                 cases map (c => (c._1, transformExpr(c._2))),
-                transformExpr(default))(tree.tpe)))
+                transformExpr(default))(tree.tpe).toPreTransform)
         }
 
       case Labeled(ident @ Ident(label, _), tpe, body) =>
@@ -818,6 +820,12 @@ private[optimizer] abstract class OptimizerCore(
         pretransformApplyStatic(tree, isStat = false,
             usePreTransform = true)(cont)
 
+      case tree: UnaryOp =>
+        pretransformUnaryOp(tree)(cont)
+
+      case tree: BinaryOp =>
+        pretransformBinaryOp(tree)(cont)
+
       case tree: JSFunctionApply =>
         pretransformJSFunctionApply(tree, isStat = false,
             usePreTransform = true)(cont)
@@ -832,8 +840,7 @@ private[optimizer] abstract class OptimizerCore(
               if (isSubtype(texpr.tpe.base, tpe.asInstanceOf[Type])) {
                 cont(texpr)
               } else {
-                cont(PreTransTree(
-                    AsInstanceOf(finishTransformExpr(texpr), tpe)))
+                cont(AsInstanceOf(finishTransformExpr(texpr), tpe).toPreTransform)
               }
           }
         }
@@ -856,7 +863,7 @@ private[optimizer] abstract class OptimizerCore(
                   RefinedType(AnyType, isExact = false, isNullable = false),
                   mutable = false,
                   replacement)
-              cont1(PreTransLocalDef(localDef))
+              cont1(localDef.toPreTransform)
             } (cont)
           } { () =>
             val newClosure = transformClosureCommon(captureParams, params, body,
@@ -868,8 +875,7 @@ private[optimizer] abstract class OptimizerCore(
         }
 
       case _ =>
-        val result = transformExpr(tree)
-        cont(PreTransTree(result))
+        cont(transformExpr(tree).toPreTransform)
     }
   }
 
@@ -931,7 +937,7 @@ private[optimizer] abstract class OptimizerCore(
           InlineClassBeingConstructedReplacement(fieldLocalDefs, cancelFun))) =>
         val fieldLocalDef = fieldLocalDefs(item.name)
         if (!isLhsOfAssign || fieldLocalDef.mutable) {
-          cont(PreTransLocalDef(fieldLocalDef))
+          cont(fieldLocalDef.toPreTransform)
         } else {
           /* This is an assignment to an immutable field of a inlineable class
            * being constructed, but that does not appear at the "top-level" of
@@ -945,7 +951,7 @@ private[optimizer] abstract class OptimizerCore(
           InlineClassInstanceReplacement(_, fieldLocalDefs, cancelFun))) =>
         val fieldLocalDef = fieldLocalDefs(item.name)
         if (!isLhsOfAssign || fieldLocalDef.mutable) {
-          cont(PreTransLocalDef(fieldLocalDef))
+          cont(fieldLocalDef.toPreTransform)
         } else {
           /* In an ideal world, this should not happen (assigning to an
            * immutable field of an already constructed object). However, since
@@ -1013,6 +1019,9 @@ private[optimizer] abstract class OptimizerCore(
             PreTransTree(finishTransformBindings(bindingsAndStats, tree), tpe)
         }
 
+      case _:PreTransUnaryOp | _:PreTransBinaryOp =>
+        PreTransTree(finishTransformExpr(preTrans))
+
       case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
         replacement match {
           case ReplaceWithRecordVarRef(name, originalName,
@@ -1052,6 +1061,9 @@ private[optimizer] abstract class OptimizerCore(
     preTrans match {
       case PreTransBlock(_, result) =>
         resolveRecordType(result)
+
+      case _:PreTransUnaryOp | _:PreTransBinaryOp =>
+        None
 
       case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
         replacement match {
@@ -1103,6 +1115,10 @@ private[optimizer] abstract class OptimizerCore(
     preTrans match {
       case PreTransBlock(bindingsAndStats, result) =>
         finishTransformBindings(bindingsAndStats, finishTransformExpr(result))
+      case PreTransUnaryOp(op, lhs) =>
+        UnaryOp(op, finishTransformExpr(lhs))
+      case PreTransBinaryOp(op, lhs, rhs) =>
+        BinaryOp(op, finishTransformExpr(lhs), finishTransformExpr(rhs))
       case PreTransLocalDef(localDef) =>
         localDef.newReplacement
       case PreTransRecordTree(_, _, cancelFun) =>
@@ -1122,6 +1138,10 @@ private[optimizer] abstract class OptimizerCore(
   private def finishTransformStat(stat: PreTransform): Tree = stat match {
     case PreTransBlock(bindingsAndStats, result) =>
       finishTransformBindings(bindingsAndStats, finishTransformStat(result))
+    case PreTransUnaryOp(_, lhs) =>
+      finishTransformStat(lhs)
+    case PreTransBinaryOp(_, lhs, rhs) =>
+      Block(finishTransformStat(lhs), finishTransformStat(rhs))(stat.pos)
     case PreTransLocalDef(_) =>
       Skip()(stat.pos)
     case PreTransRecordTree(tree, _, _) =>
@@ -1223,10 +1243,10 @@ private[optimizer] abstract class OptimizerCore(
         case NothingType =>
           cont(treceiver)
         case NullType =>
-          cont(PreTransTree(Block(
+          cont(Block(
               finishTransformStat(treceiver),
               Throw(New(ClassType("jl_NullPointerException"),
-                  Ident("init___", Some("<init>")), Nil)))))
+                  Ident("init___", Some("<init>")), Nil))).toPreTransform)
         case _ =>
           if (isReflProxyName(methodName)) {
             // Never inline reflective proxies
@@ -1424,8 +1444,8 @@ private[optimizer] abstract class OptimizerCore(
     implicit val pos = tree.pos
 
     if (args.exists(_.isInstanceOf[JSSpread])) {
-      cont(PreTransTree(
-          JSFunctionApply(transformExpr(fun), args.map(transformExprOrSpread))))
+      cont(JSFunctionApply(transformExpr(fun),
+          args.map(transformExprOrSpread)).toPreTransform)
     } else {
       pretransformExpr(fun) { tfun =>
         tfun match {
@@ -1436,15 +1456,15 @@ private[optimizer] abstract class OptimizerCore(
             alreadyUsed.value = true
             pretransformExprs(args) { targs =>
               inlineBody(
-                  Some(PreTransTree(Undefined())), // `this` is `undefined`
+                  Some(PreTransLit(Undefined())), // `this` is `undefined`
                   captureParams ++ params, AnyType, body,
-                  captureLocalDefs.map(PreTransLocalDef(_)) ++ targs, isStat,
+                  captureLocalDefs.map(_.toPreTransform) ++ targs, isStat,
                   usePreTransform)(cont)
             }
 
           case _ =>
-            cont(PreTransTree(
-                JSFunctionApply(finishTransformExpr(tfun), args.map(transformExpr))))
+            cont(JSFunctionApply(finishTransformExpr(tfun),
+                args.map(transformExpr)).toPreTransform)
         }
       }
     }
@@ -1613,7 +1633,7 @@ private[optimizer] abstract class OptimizerCore(
     lazy val newReceiver = finishTransformExpr(optTReceiver.get)
     lazy val newArgs = targs.map(finishTransformExpr)
 
-    @inline def contTree(result: Tree) = cont(PreTransTree(result))
+    @inline def contTree(result: Tree) = cont(result.toPreTransform)
 
     @inline def StringClassType = ClassType(Definitions.StringClass)
 
@@ -1810,15 +1830,15 @@ private[optimizer] abstract class OptimizerCore(
       cont: PreTransCont)(
       implicit scope: Scope, pos: Position): TailRec[Tree] = {
     pretransformNew(value, ClassType(Definitions.BoxedCharacterClass),
-        Ident("init___C"), List(PreTransTree(value)))(cont)
+        Ident("init___C"), List(value.toPreTransform))(cont)
   }
 
   private def unboxChar(tvalue: PreTransform)(
       cont: PreTransCont)(
       implicit scope: Scope, pos: Position): TailRec[Tree] = {
     val BoxesRunTimeModuleClassName = "sr_BoxesRunTime$"
-    val treceiver = PreTransTree(LoadModule(
-        ClassType(BoxesRunTimeModuleClassName)))
+    val treceiver = LoadModule(
+        ClassType(BoxesRunTimeModuleClassName)).toPreTransform
     val target = staticCall(BoxesRunTimeModuleClassName, "unboxToChar__O__C").getOrElse {
       throw new AssertionError("Cannot find method sr_BoxesRunTime$.unboxToChar__O__C")
     }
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0

sjrd:
The [diff of the output for the test suite](https://gist.github.com/sjrd/7b687b7d84d81a825f72c3035df4216e) is unimpressive, but it does get rid of clearly useless things, notably some module loads.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2319,20 +2324,137 @@ private[optimizer] abstract class OptimizerCore(
...
+
+    (op: @switch) match {
+      case === | Num_== | Long_== | Boolean_== =>
+        BooleanLiteral(literal_===(lhs, rhs))
+      case !== | Num_!= | Long_!= | Boolean_!= =>
+        BooleanLiteral(!literal_===(lhs, rhs))
+
+      case String_+ =>
+        throw new IllegalArgumentException(
+            "constFoldBinaryOp_except_String_+ must be called for String_+")
```
gzm0: must not be called

gzm0:
Just to check my understanding: This removes and generalizes a couple of special cases we needed for the long constant folding, right?

gzm0:
Also, can you explain this in the diff:

```diff
-  this$5.assertEquals__T__O__O__V(null, "hello", "hello");
+  var actual = "hello";
+  this$5.assertEquals__T__O__O__V(null, "hello", actual);
```

Everything else makes sense.

sjrd:
> Just to check my understanding: This removes and generalizes a couple of special cases we needed for the long constant folding, right?

Nope. So far it only removes special cases. They were not used anymore, and they were difficult to translate in the new scheme. The generalization is yet to come, but will be based on the changes brought by this PR.

sjrd:
> Also, can you explain this in the diff:

I am not completely sure. It also attracted my attention during my review of the diff. What I think happened is the following:

1. We inline `cast`, defined at https://github.com/scala-js/scala-js/blob/v0.6.9/javalanglib/src/main/scala/java/lang/Class.scala#L82-L89. In the process, we declare a `LocalDef` for `obj` which is a `ReplaceWithConstant("hello")`.
2. We pretransform the block of the body of `inline`. While doing so, the result expression is eagerly replaced by a literal at pretransform time (because of the new `.toPreTransform`s everywhere). But then the block of statement + result is not a `PreTransBlock` anymore, because the result part is a `Tree` and not a `LocalDef`. At that point, the statement is not completely folded away for some reason--which I am not completely certain why, but I believe that it's because `PreTransform`s are less capable than `Tree`s at removing side-effect-free ifs (so far).
3. Therefore, the result of inlining returns a `PreTransTree(Block(stat, Literal("hello")))`, whereas in the old version, it would have been a `PreTransBlock(stat, PreTransLocalDef("obj"))`.
4. This block is assigned to the `LocalDef` for `actual`, whereas previously it would have flattened out the `PreTransBlock`.
5. Since `actual` is used, it is kept.
6. Some late folding happens at some point--but again, I'm not sure, probably during `finishTransformExpr`--which reduces the `PreTransTree(Block(stat, Literal("hello")))` into `Literal("hello")`.

That's basically my vague understanding of the whole story. I suspect we hit here a bad interaction between folding and pretransforms. I think this artifact will go away when we move more things to the pretransform world.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2251,50 +2245,61 @@ private[optimizer] abstract class OptimizerCore(
                  * will not go into infinite recursion.
                  */
                 implicit val scope = Scope.Empty
-                inline(Nil, Some(PreTransTree(x)), Nil, target, isStat = false,
+                inline(Nil, Some(x), Nil, target, isStat = false,
                     usePreTransform = false)(finishTransform(isStat = false))
-              }
+              })
```
gzm0: This makes me think: Why do the fold operations not take continuations like all methods on pretransforms?

sjrd: Because they never make optimistic choices (`tryOrRollback`) nor call any method that can make ones (except this weird pair of hacks, but well ...). Therefore, then don't need to. And since they don't need to, they'd better not do it, as it's expensive.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2319,20 +2324,137 @@ private[optimizer] abstract class OptimizerCore(
...
+      case IntLiteral(value) => value
+    }
+
+    @inline def long(lit: Literal): Long = (lit: @unchecked) match {
+      case LongLiteral(value) => value
+    }
+
+    @inline def float(lit: Literal): Float = (lit: @unchecked) match {
+      case FloatLiteral(value) => value
+      case IntLiteral(value)   => value.toFloat
```
gzm0: I do not understand this. Is our spec really that float ops can work on ints, but double ops can't???

sjrd: Doubles can, but they use the special `NumberLiteral` extract (not `DoubleLiteral`) which matches all of `IntLiteral`, `FloatLiteral` and `DoubleLiteral`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ah, ok. So here we are fine with constructing a `PreTransTree` directly, since we know we can't do anything else.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ah.... Maybe I should read more code :-/

sjrd: That said, I now realize there is another problem: float operations are not supposed to accept int values, because `int` is not a subtype of `float`. So in fact this `case` should not be there.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2352,216 +2474,259 @@ private[optimizer] abstract class OptimizerCore(
...
+    @inline def default =
+      PreTransBinaryOp(op, lhs, rhs)
+
     (op: @switch) match {
       case === | !== =>
         val positive = (op == ===)
         (lhs, rhs) match {
-          case (lhs: Literal, rhs: Literal) =>
-            BooleanLiteral(literal_===(lhs, rhs) == positive)
+          case (_, PreTransTree(Null(), _)) if !lhs.tpe.isNullable =>
```
gzm0: Isn't `Null()` a literal anyways?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2352,216 +2474,259 @@ private[optimizer] abstract class OptimizerCore(
...
+            PreTransBlock(
+                finishTransformStat(lhs),
+                PreTransLit(BooleanLiteral(!positive)))
 
-          case (_: Literal, _) => foldBinaryOp(op, rhs, lhs)
-          case _               => default
+          case (PreTransLit(_), _) => foldBinaryOp(op, rhs, lhs)
```
gzm0: If `Null()` is a literal, then this case makes sense. Otherwise I do not understand how you achieve the symmetric case of the folding just above.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes, it is, but ... how is that a problem here?

(I guess I could have used `PreTransLit(Null())` instead of `PreTransTree(Nul(), _)`, indeed. Probably a refactoring opportunity missed.)

gzm0: Ah... Just realized that `PreTransList` is just an extractor :-/

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2352,216 +2474,259 @@ private[optimizer] abstract class OptimizerCore(
...
-          case (_, IntLiteral(_))             => foldBinaryOp(Int_*, rhs, lhs)
+          case (_, PreTransLit(IntLiteral(_))) =>
+            foldBinaryOp(Int_*, rhs, lhs)
 
-          case (IntLiteral(1), _)  => rhs
-          case (IntLiteral(-1), _) => foldBinaryOp(Int_-, IntLiteral(0), rhs)
+          case (PreTransLit(IntLiteral(1)), _)  =>
+            rhs
+          case (PreTransLit(IntLiteral(-1)), _) =>
+            foldBinaryOp(Int_-, PreTransLit(IntLiteral(0)), rhs)
```
gzm0: Do we want to add multiplication by 0?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2352,216 +2474,259 @@ private[optimizer] abstract class OptimizerCore(
...
       case Int_/ =>
         (lhs, rhs) match {
-          case (IntLiteral(l), IntLiteral(r)) if r != 0 => IntLiteral(l / r)
-
-          case (_, IntLiteral(1))  => lhs
-          case (_, IntLiteral(-1)) => foldBinaryOp(Int_-, IntLiteral(0), lhs)
+          case (_, PreTransLit(IntLiteral(1)))  =>
+            lhs
+          case (_, PreTransLit(IntLiteral(-1))) =>
+            foldBinaryOp(Int_-, PreTransLit(IntLiteral(0)), lhs)
```
gzm0: Do we want to add division of 0?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2352,216 +2474,259 @@ private[optimizer] abstract class OptimizerCore(
...
         }
 
       case Int_% =>
         (lhs, rhs) match {
-          case (IntLiteral(l), IntLiteral(r)) if r != 0 => IntLiteral(l % r)
-          case (_, IntLiteral(1 | -1))                  =>
-            Block(keepOnlySideEffects(lhs), IntLiteral(0))
-          case _                                        => default
+          case (_, PreTransLit(IntLiteral(1 | -1))) =>
+            PreTransBlock(finishTransformStat(lhs), PreTransLit(IntLiteral(0)))
```
gzm0: Do we want to add modulus of 0?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Not in this PR, at least. That's a very unrelated change. I'll probably do that at the same time #2456 during my Scala Days talk.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: We probably should, indeed. But again, not in this PR.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2752,129 +2923,146 @@ private[optimizer] abstract class OptimizerCore(
...
+            foldBinaryOp(Float_-, PreTransLit(FloatLiteral(0)), lhs)
 
           case _ => default
         }
 
       case Float_% =>
         (lhs, rhs) match {
-          case (FloatLiteral(l), FloatLiteral(r)) => FloatLiteral(l % r)
-          case _                                  => default
+          case _ => default
```
gzm0: You have a lot of trust in scalac to get rid of this ;-)

sjrd: Oh it won't. At least not before the new optimizer of 2.12. But I don't care. It's a rare case anyway. I just want my cases to be consistent.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2752,129 +2923,146 @@ private[optimizer] abstract class OptimizerCore(
...
+            foldBinaryOp(Double_+, PreTransLit(DoubleLiteral(x - y)), z)
+
+          case (_, PreTransBinaryOp(BinaryOp.Double_-,
+              PreTransLit(NumberLiteral(0)), x)) =>
             foldBinaryOp(Double_+, lhs, x)
 
           case _ => default
         }
 
       case Double_* =>
```
gzm0: Done until here. I need to sleep now...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ok :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2752,129 +2923,146 @@ private[optimizer] abstract class OptimizerCore(
...
         val positive = (op == Boolean_==)
         (lhs, rhs) match {
-          case (BooleanLiteral(l), _) =>
+          case (PreTransLit(BooleanLiteral(l)), _) =>
             if (l == positive) rhs
             else foldUnaryOp(UnaryOp.Boolean_!, rhs)
-          case (_, BooleanLiteral(r)) =>
+          case (_, PreTransLit(BooleanLiteral(r))) =>
             if (r == positive) lhs
             else foldUnaryOp(UnaryOp.Boolean_!, lhs)
```
gzm0: Clean this up to switch the literals like everywhere else?

gzm0:
Note for later: These foldings should be added (or are at least missing for some ops):

- Multiplication / division of 0
- Multiplication by 0
- x & false  --> { x, false } et al.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3736,18 +3920,32 @@ private[optimizer] object OptimizerCore {
       }
     }
 
-    def unapply(preTrans: PreTransBlock): Some[(List[BindingOrStat], PreTransLocalDef)] =
+    def unapply(preTrans: PreTransBlock): Some[(List[BindingOrStat], PreTransResult)] =
       Some(preTrans.bindingsAndStats, preTrans.result)
   }
 
-  private sealed abstract class PreTransNoBlock extends PreTransform
+  private sealed abstract class PreTransResult extends PreTransform
```
gzm0: Could you start documenting the different subtypes of pretransforms? Especially the difference between `PreTransResult` and `PreTransGenTree` seems (from a mere naming perspective) unclear to me.

gzm0: So if my understanding is correct, `PreTransGenTree` is a result of a `resolveLocalDef`. `PreTransResult` is a tree for an expression.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3766,7 +3964,7 @@ private[optimizer] object OptimizerCore {
   }
 
   private object PreTransTree {
-    def apply(tree: Tree): PreTransTree = {
+    def make(tree: Tree): PreTransTree = {
```
gzm0: If this deserves to be a different operation (and not inlined into `apply`), it deserves a better name. (And maybe `apply` as well).
Notably: If some of the things done inside `apply` are not correct in some contexts, consider renaming `apply` instead. IMHO calling `A(x, y, z)` should "always work" (more formally: should work if `{x, y, z}` satisfies a *minimal* set of preconditions).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2965,15 +3144,18 @@ private[optimizer] abstract class OptimizerCore(
...
     import BinaryOp._
     if (canBeEqual) {
       if (canBeLessThan) {
-        if (canBeGreaterThan)
-          Block(keepOnlySideEffects(lhs), keepOnlySideEffects(rhs), BooleanLiteral(true))
-        else
+        if (canBeGreaterThan) {
+          PreTransBlock(
+              Block(finishTransformStat(lhs), finishTransformStat(rhs)),
+              PreTransLit(BooleanLiteral(true)))
```
gzm0: Consider:

```scala
PreTransBlock(
              Right(finishTransformStat(lhs)) :: Right(finishTransformStat(rhs)) :: Nil,
              PreTransLit(BooleanLiteral(true)))
```

or alternatively giving `PreTransBlock` an `apply` method that can flatten blocks. This would allow us to flatten skips earlier and this seems to be a rather common case here.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3705,8 +3889,8 @@ private[optimizer] object OptimizerCore {
     }
 
     def apply(binding: PreTransBinding,
-        result: PreTransLocalDef): PreTransform = {
-      if (result.contains(binding.localDef)) {
+        result: PreTransResult): PreTransform = {
+      if (result.contains(binding.localDef) || binding.isAlreadyUsed) {
```
gzm0: Can you quickly explain, why this is suddenly necessary?

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Might as well, yes.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: OK I'll add some doc. `PreTransGenTree` is indeed a result of a `resolveLocalDef`. `PreTransResult` is anything that can the result expression of a `PreTransBlock`. That's basically everything but `PreTransGenTree` (otherwise the `PreTransBlock` itself would become a `PreTransGenTree` and `PreTransBlock` (otherwise the two `PreTransBlock` would be flattened). So far, that leaves `PreTransLocalDef`, `-UnaryOp` and `-BinaryOp`. In the future, pretty much everything else (like `PreTransIf`) would be a `PreTransResult`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: This one was hard. Basically, I wanted `apply` to do *the right thing* for most cases. Basically, if I want to create a `PreTransform` from a `Tree`, I call `PreTransTree(tree)`, and it is smart (and sometimes it doesn't return a `PreTransTree` but something else). `make` is needed for the few cases where we do *not* want it to be smart, which are basically the rare cases where you really want a `PreTransTree` no matter what. And in fact there is only one such place: in `resolveLocalDef`, where you were specifically asked not to be smart anymore, and instead force things to return a `PreTransGenTree`.

An alternative, I guess, would be to give back the meaning of the current `make` to `apply`, and give the meaning of `apply` to a `.toPreTransform` decorator method for `Tree`. That would also be more consistent with `PreTransLocalDef.toPreTransform`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: I cannot give `PreTransBlock` an `apply` method that does that, because flattening blocks (`finishTransformExpr`) needs the instance of `OptimizerCore` (to have the knowledge queries for `isModuleAccessorElidable` for example)

sjrd: Your first suggestion would work (if we enhance the constructor of `PreTransBlock(List[PreTransBinding])` to get rid of `Right(Skip())`s eagerly) but I am not sure why it is better than the above?

That said, now I realize that what I wrote is stupid. Since `PreTransLit` gives a `PreTransTree`, that constructor of `PreTransBlock` will turn everything into a `PreTransTree`, and so I should simply write:
```scala
PreTransTree(Block(
    keepOnlySideEffects(lhs),
    keepOnlySideEffects(rhs),
    BooleanLiteral(true)))
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Ah, yes. I almost thought there was a pre-existing major bug when I had to add that one.

The thing is, before, the only possible `result` was `PreTransLocalDef`. A `PreTransLocalDef` can only ever contain virtual references to local variables, and so `contains` catches all possible references to local variables defined by `binding`.

But now, `result` can also be a `PreTransBinaryOp`, for example. And one of the arguments could be a `PreTransTree`, which itself contains somewhere a `VarRef` to the variable defined by `binding`. That would happen, for example, if an operand was an `Apply(..., List(VarRef(...)))`, since the `Apply` cannot be virtualized, its arguments had to be eagerly be forced, and a `VarRef` was produced (setting `used.value = true` in the process). After that, the top-level `PreTransBinaryOp` does `contains(binding.localDef)`, but it used the binding nevertheless.

sjrd:
Updated.

I have separated in a dedicated commit the proposed rewriting of `PreTransTree.apply(tree)` -> `tree.toPreTransform` and `PreTransTree.make(tree)` -> `PreTransTree.apply(tree)`, so that you can see what it looks like. If you like it, I'll squash that commit with the rest.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: > After that, the top-level PreTransBinaryOp does contains(binding.localDef), but it used the binding nevertheless.

If that was supposed to be "does *not* `contain(binding.localDef)`, I understand and this makes sense. :)

gzm0:
> I have separated in a dedicated commit the proposed rewriting of PreTransTree.apply(tree) -> tree.toPreTransform and PreTransTree.make(tree) -> PreTransTree.apply(tree)

Yeah, I like it. It fells like `_.toX` methods have much more of a "conversion" feel to themselves, so doing smart stuff in them is much more natural than in an `apply` method.

gzm0:
> 3. Therefore, the result of inlining returns a PreTransTree(Block(stat, Literal("hello"))), whereas in the old version, it would have been a PreTransBlock(stat, PreTransLocalDef("obj")).

Ah... Maybe we should (probably in a later PR), adapt `PreTransLit` to also find these (and have the matcher return a second side effect tree).

> 6. Some late folding happens at some point--but again, I'm not sure, probably during finishTransformExpr--which reduces the PreTransTree(Block(stat, Literal("hello"))) into Literal("hello").

IIUC, `stat` doesn't get removed (it is the actual typecheck, isn't it). It just was there before (and of course must be kept).

gzm0:
LGTM modulo:

- Squashing.
- Figuring something out to fold over literals in blocks (not in this PR).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes, that was supposed to be "does not".

sjrd:
> Ah... Maybe we should (probably in a later PR), adapt PreTransLit to also find these (and have the matcher return a second side effect tree).

Yes, I have sometimes thought about that. It would significantly complicate the `fold` methods, though. But we probably need this, eventually.

sjrd:
Updated (only squashed).

gzm0:
LGTM

PR #2459, opened 2016-06-08
Fix #2427: Deprecate sbtplugin.{Implicits, LoggerJSConsole}

### LINKED ISSUE #2427: The conversions in sbtplugin.Implicits should not be implicit
https://github.com/scala-js/scala-js/blob/master/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/Implicits.scala

Those conversions should be available as explicit conversion methods, not as implicit conversions.

nicolasstucki:


sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/LoggerJSConsole.scala
```diff
@@ -13,6 +13,7 @@ import sbt.Logger
 import org.scalajs.jsenv.JSConsole
 
 /** A proxy for a Logger that looks like a Mozilla console object */
+@deprecated("", "0.6.10")
```
nicolasstucki: Not sure what to put as message for this one.

nicolasstucki:
Review by @sjrd

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/LoggerJSConsole.scala
sjrd: Use org.scalajs.jsenv.ConsoleJSConsole or define your own instead.

?

sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/Loggers.scala
```diff
@@ -0,0 +1,32 @@
+package org.scalajs.sbtplugin
+
+import org.scalajs.core.tools.logging._
+import sbt.{Level => SbtLevel, Logger => SbtLogger}
+
+object Loggers {
+  private class SbtLoggerWrapper(underlying: SbtLogger) extends Logger {
+    def log(level: Level, message: => String): Unit =
+      sbtLogger2ToolsLogger(underlying).log(level, message)
```
sjrd: This goes into infinite recursion. It should be
```scala
underlying.log(toolsLevel2sbtLevel(level), message)
```

sjrd:
That's all.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/Implicits.scala
```diff
@@ -5,7 +5,9 @@ import scala.language.implicitConversions
 import org.scalajs.core.tools.logging._
 import sbt.{Logger => SbtLogger, Level => SbtLevel}
 
+@deprecated("Use sbtplugin.Loggers explicitly instead.", "0.6.10")
 object Implicits {
+  @deprecated("Use sbtplugin.Loggers.SbtLoggerWrapper instead.", "0.6.10")
   private class SbtLoggerWrapper(underlying: SbtLogger) extends Logger {
```
sjrd: Why do you keep `SbtLoggerWrapper` at all? It's private, so we can freely remove it.

sjrd:
That's all.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/Implicits.scala
nicolasstucki: To avoid binary incompatibilities.

sjrd: You can ignore that one. Since it's private, there's no way a third party can directly refer to it, so binary compat is not necessary.

nicolasstucki: Done

project/BinaryIncompatibilities.scala
```diff
@@ -37,6 +37,9 @@ object BinaryIncompatibilities {
   )
 
   val SbtPlugin = Seq(
+    // private, not an issue
+    ProblemFilters.exclude[MissingClassProblem](
```
sjrd: 4 spaces ;) (sorry)

sjrd:
LGTM

PR #2460, opened 2016-06-08
Fix #2428: Disallow @js.native in Enumeration classes/objects.

### LINKED ISSUE #2428: @js.native on a class/object extending Enumeration does not report an error
But it should. This works correctly for things that do not extend `Enumeration`.

nicolasstucki:


nicolasstucki:
Review by @sjrd

sjrd:
The error message should be the same as the one produced by https://github.com/scala-js/scala-js/blob/1b4d32f229839ad9371a35367ce822b4f33f320d/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala#L189-L192

And the tests should also be together with the existing tests for that error message. https://github.com/scala-js/scala-js/blob/6ff73088bf15ba1102558fbe9516f3f60e32aa9b/compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala#L71

The fact that we need separate code for `Enumeration` subclasses is an artifact of our implementation, which should not be visible to users. To them, it doesn't matter whether it's `Enumeration` or not; all that matters is that it's not extending `js.Any`.

sjrd:
That's all.

sjrd:
LGTM

PR #2468, opened 2016-06-14
Upgrade to Scala 2.12.0-M5

### LINKED ISSUE #2441: Our javalib tests use `@Test` methods in traits, which don't work in M5+JUnit 4
As of Scala 2.12.0-M5, forwarder methods to trait's default methods are not created anymore if they can be avoided. This is problem for `@Test` methods in traits, since JUnit 4 does not look for default methods. JUnit 5 will fix this issue, but it is not released yet.

We need to do something about it.

nicolasstucki:


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJUnitMixinPlugin.scala
```diff
@@ -0,0 +1,30 @@
...
+import sbt._
+
+object ScalaJUnitMixinPlugin extends AutoPlugin {
+
+  override def projectSettings: Seq[Setting[_]] = Seq(
+    libraryDependencies ++= {
+      if (enablePlugin(scalaVersion.value)) {
+        Seq("org.scala-js" %
+            ("junit-mixin-support-scala-for-scala-2-12-_" + scalaVersion.value) %
+            "0.0.1-SNAPSHOT" % "plugin->test")
```
nicolasstucki: @sjrd, this is working if I publish `2.12.0-SNAPSHOT` locally, then publish the JUnit mixin plugin  for that version. Then tests it on  `2.12.0-SNAPSHOT`. 
What would be the correct way to do this?

sjrd: I'm afraid the only correct way to do this is to wait for M5 to released on Maven Central.

sjrd: It's good to know that it works with this manual hack, though, as we have some confidence that our entire test suite passes with M5 (as opposed to: bugs are hidden because their tests are not run).

nicolasstucki: OK

nicolasstucki:
Still need the JUnit mixin plugin to be published.

sjrd:
> Still need the JUnit mixin plugin to be published.

I'll take care of that this morning.

project/Build.scala
```diff
@@ -370,9 +370,12 @@ object Build {
             Seq(s"-Xplugin:$jar")
           }
         }
-      )
+      ).withScalaJUnitMixinPlugin
     }
 
+    def withScalaJUnitMixinPlugin: Project =
+      project.settings(ScalaJUnitMixinPlugin.projectSettings)
```
sjrd: I suggest you inline that definition right inside `withScalaJSJUnitPlugin`, which is the only user of this method.
Moreover, I suggest you inline `ScalaJUnitMixinPlugin.projectSettings right there within `withScalaJSJUnitPlugin` as well, and get rid of `ScalaJUnitMixinPlugin`.

sjrd: Ah it's also used in `testSuiteJVM`. Then maybe just make it a `scalaJUnitMixinPluginSetting` of type `Setting[_]`, which can be included in `.settings()`.

sjrd:
Isn't there any change to be made in the Whitelist and Blacklist of partests?

junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
```diff
@@ -108,7 +108,7 @@ class ScalaJSJUnitPlugin(val global: Global) extends NscPlugin {
     import global._
 
     val phaseName: String = "junit-inject"
-    val runsAfter: List[String] = List("mixin")
+    val runsAfter: List[String] = List("mixin", "junit-make-mixin-bridges")
```
sjrd: This change is not necessary, since the mixin plugin takes care of that. And we've discovered that `runsAfter` warns for missing dependencies (and errors out with fatal-warnings).

sjrd:
That's all. IMO the two commits can be squashed.

junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
nicolasstucki: If I remove it it fails, it looks like the `runBefore` is ignored.

nicolasstucki: I mean the `runBefore` in the other plugin.

sjrd: That's really surprising, because it worked for me on master.
Did you clear your ivy cache of the old, buggy mixin plugin?

nicolasstucki:
Partest has still not released for `2.12.0-M5`

sjrd:
It seems there is scala-partest 1.0.17 for 2.12.0-M5: http://search.maven.org/#artifactdetails%7Corg.scala-lang.modules%7Cscala-partest_2.12.0-M5%7C1.0.17%7Cjar
However for 2.11 they only have 1.0.16 -_-' We can select the appropriate one with a little `if`, though.

sjrd:
```
--- Classified Tests
+++ Existing Tests
@@ -2886,0 +2887 @@
+run/repl-implicits.scala
@@ -3740 +3740,0 @@
-run/t7445.scala
```
Did you remove the wrong line?

project/Build.scala
```diff
@@ -1452,6 +1438,7 @@ object Build {
           name := "JavaLib Ex Test Suite",
           publishArtifact in Compile := false,
           testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a", "-s"),
+          scalacOptions += "-Ynowarn-default-junit-methods",
```
nicolasstucki: I had to add the `-Ynowarn-default-junit-methods` explicitly `javalibExTestSuite` because for this project the plugin fails to set the flag. The flag is set using a `tryToSet` in `global.settings.nowarnDefaultJunitMethods.tryToSet(List("true"))` so it is not so surprising that it failed.

nicolasstucki: Yes, I checked that all the methods that get the warning are executing. Additionally if the plugin does not run the test would just fail with linking errors.

sjrd: OK that's good. It's still concerning, though. Any insight on why the plugin does not manage to set the option on its own?

nicolasstucki: Unfortunately no insight.

partest/src/main/scala/scala/tools/partest/scalajs/ScalaJSPartest.scala
```diff
@@ -190,15 +192,12 @@ class ScalaJSSBTRunner(
   // Partests take at least 5h. We double, just to be sure. (default is 4 hours)
   sys.props("partest.timeout") = "10 hours"
 
-  // Set showDiff on global UI module
-  if (options.showDiff)
-    NestUI.setDiffOnFail()
```
sjrd: What happens to the `showDiff` option? It seems not be honored anymore.

project/ScalaJUnitMixinPlugin.scala
```diff
@@ -0,0 +1,27 @@
...
+**                          |/____/                                     **
+\*                                                                      */
+
+import sbt.Keys._
+import sbt._
+
+object ScalaJUnitMixinPlugin extends AutoPlugin {
+
+  override def projectSettings: Seq[Setting[_]] = Seq(
+    libraryDependencies in Test ++=
```
sjrd: `libraryDependencies in Test` is never used. See:
```
> set libraryDependencies in (compiler, Test) += "org.blop" % "sfoih" % "1.0"
[info] Defining compiler/test:libraryDependencies
[info] The new value will be used by no settings or tasks.
[info] Reapplying settings...
```
We need a configuration like the one in https://github.com/scala-js/scala-2.12-junit-mixin-plugin/pull/13

Btw, I still think there is no good reason for this config to be in a separate file. It can just be a `val` within `Build`. Note in particular that, currently, the `AutoPlugin` nature of `ScalaJUnitMixinPlugin` is not even used. So this is really just an `object` with two `def`s, which can and should be in `Build`.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/InstanceTestsHijackedBoxedClassesTest.scala
```diff
@@ -84,7 +84,10 @@ class InstanceTestsHijackedBoxedClassesTest {
 
   @Test def should_support_asInstanceOf_negative(): Unit = {
     assumeTrue("Assumed compliant asInstanceOf", hasCompliantAsInstanceOfs)
-    assertThrows(classOf[Exception], (12345: Any).asInstanceOf[Unit])
+    if (!scalaVersion.startsWith("2.12."))
+      assertThrows(classOf[Exception], (12345: Any).asInstanceOf[Unit])
+    else
+      (12345: Any).asInstanceOf[Unit]
```
sjrd: The `else` part should be in `should_support_asInstanceOf_positive()` (with an `if (scalaVersion.startsWith("2.12.")))`), because it must also work when we do not have compliant as-instance-ofs.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
```diff
@@ -40,12 +37,8 @@ class ClassTest {
     assertSame(classOf[Long], java.lang.Long.TYPE)
     assertSame(classOf[Float], java.lang.Float.TYPE)
     assertSame(classOf[Double], java.lang.Double.TYPE)
-
```
sjrd: This blank line should be kept.

partest/src/main/scala/scala/tools/partest/scalajs/ScalaJSPartest.scala
nicolasstucki: [It went here](https://github.com/scala-js/scala-js/pull/2468/files#diff-a6531f0ed7cf3feadd0280492d634f17R40) due to a change in design of the `NestUI`.

sjrd: Ah OK!

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -795,11 +795,11 @@ class JSExportTest extends DirectTest with TestHelpers {
...
     }
     """ hasWarns
-    """
-      |newSource1.scala:5: warning: class JSExportNamed in package annotation is deprecated: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
+    s"""
+      |newSource1.scala:5: warning: class JSExportNamed in package annotation is deprecated${since("0.6.11")}: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
       |      def foo(x: Int, y: Int) = 1
       |          ^
-      |newSource1.scala:4: warning: class JSExportNamed in package annotation is deprecated: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
+      |newSource1.scala:4: warning: class JSExportNamed in package annotation is deprecated${since("0.6.11")}: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
```
nicolasstucki: @sjrd apart from your comments I also added this in the last PR update.

ci/checksizes.sh
```diff
@@ -45,11 +45,11 @@ case $FULLVER in
...
-  2.12.0-M4)
-    REVERSI_PREOPT_EXPECTEDSIZE=518000
-    REVERSI_OPT_EXPECTEDSIZE=121000
-    REVERSI_PREOPT_GZ_EXPECTEDSIZE=70000
-    REVERSI_OPT_GZ_EXPECTEDSIZE=32000
+  2.12.0-M5)
+    REVERSI_PREOPT_EXPECTEDSIZE=526000
+    REVERSI_OPT_EXPECTEDSIZE=124000
+    REVERSI_PREOPT_GZ_EXPECTEDSIZE=67000
+    REVERSI_OPT_GZ_EXPECTEDSIZE=31000
```
sjrd: That's surprising. Significantly more code in non-gzipped versions but significantly less in gzipped versions :-s I'm especially worried that fastopt.js is 8 KB larger. Could you post a diff of the two in a gist, please?

sjrd:
Otherwise LGTM

nicolasstucki:
https://gist.github.com/nicolasstucki/92fec54a1dae076749789129806b30ba

sjrd:
OK I understand what happened. Since M5 does not generate explicit bridges to default methods in classes anymore, the linker has more responsibility in creating default method bridges. But the way it does this is suboptimal. Consider:
```scala
trait Foo {
  def foo(): Int = 5
}
class A extends Foo
class B extends A
class C extends A
```
An ideal solution (and what happened when scalac generated the bridges) is to create the bridge in `A`:
```scala
class A extends Foo {
  def foo(): Int = this.Foo::foo()
}
class B extends A
class C extends A
```
But the way the linker works for now is a bit stupid, and it will generate bridges in all the instantiated subclasses of `A`. So if `A`, `B` and `C` are all instantiated, we end up with
```scala
class A extends Foo {
  def foo(): Int = this.Foo::foo()
}
class B extends A {
  def foo(): Int = this.Foo::foo()
}
class C extends A {
  def foo(): Int = this.Foo::foo()
}
```
This explains while fastopt.js became significantly bigger. It also provides a reasonable explanation why the gzipped versions are smaller, because of lot of this stupid code is actually duplicated code, which will be effectively "gzipped away".

The good news is that it is also fixable. We just have to make the default bridge generation smarter.

sjrd:
```
Checksizes: Scala version: 2.12.0-M5
Reversi preopt size = 528522 (expected 526000)
Reversi opt size = 122900 (expected 124000)
Reversi preopt gzip size = 66850 (expected 67000)
Reversi opt gzip size = 30104 (expected 31000)
```

sjrd:
This should now be rebased since #2524 was merged.

sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/4287/

sjrd:
LGTM, nightly passed, all that ...
Let's merge this!

PR #2469, opened 2016-06-14
Fix #2467: Keep side effects of values of discarded bindings.

### LINKED ISSUE #2467: run/names-default.scala is broken on master
https://github.com/scala/scala/blob/2.12.x/test/files/run/names-defaults.scala
https://github.com/scala/scala/blob/2.12.x/test/files/run/names-defaults.check
```diff
!! 249 - run/names-defaults.scala                  [output differs]
% diff /localhome/jenkins/a/workspace/scalajs-task-worker/partest/fetchedSources/2.12.0-M4/test/files/run/names-defaults-run.log /localhome/jenkins/a/workspace/scalajs-task-worker/partest/fetchedSources/2.12.0-M4/test/files/run/names-defaults.check
@@ -65,6 +65,7 @@ get: 11
 11
 get: 1
 get: 2
+get: 2
 3
 0
 get: 20
```
I suspect an optimizer bug. Something like erroneously discarding side effects. If that's the case, this is a blocker for 0.6.10.

sjrd:
Test suite diff: https://gist.github.com/sjrd/e93b0ff26592233286db1698d96127e2
The first diff is the actual fix. The other diffs are also fixes, in theory, though in practice they were not affecting the results because the statements that were discarded were side-effect-free in practice (although not proved to be so by the optimizer).

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
@@ -82,6 +82,20 @@ class OptimizerTest {
     assertFalse(classOf[Array[Array[Object]]] == classOf[Array[Object]])
   }
 
+  @Test def side_effect_discard_in_eliminated_binding_issue_2467(): Unit = {
+    val b = Array.newBuilder[AnyRef]
+    def mockPrintln(x: Any): Unit =
+      b += ("" + x)
+
+    def get[T](x: T) = { mockPrintln("get: "+ x); x }
+
+    def bn2(a: Int, b: => Int)(c: Int = b) = a + b
+    mockPrintln(bn2(b = get(2), a = get(1))()) // should get: 1, 2, 2
+
+    assertArrayEquals(Array[AnyRef]("get: 1", "get: 2", "get: 2", "3"),
+        b.result())
+  }
+
   // +[string] constant folding
 
   @Test def must_not_break_when_folding_two_constant_strings(): Unit = {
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/OptimizerTest.scala
```diff
@@ -82,6 +82,20 @@ class OptimizerTest {
...
+    def mockPrintln(x: Any): Unit =
+      b += ("" + x)
+
+    def get[T](x: T) = { mockPrintln("get: "+ x); x }
+
+    def bn2(a: Int, b: => Int)(c: Int = b) = a + b
+    mockPrintln(bn2(b = get(2), a = get(1))()) // should get: 1, 2, 2
+
+    assertArrayEquals(Array[AnyRef]("get: 1", "get: 2", "get: 2", "3"),
+        b.result())
```
sjrd: I checked that this test fails without the fix to the optimizer.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3515,27 +3515,24 @@ private[optimizer] abstract class OptimizerCore(
...
    */
-  private def addPreTransBinding(preTransBinding: PreTransBinding,
+  private def addPreTransBinding(binding: PreTransBinding,
       result: PreTransform): PreTransform = {
     /* This is not the same as
      *   addPreTransBindings(Left(preTransBinding) :: Nil, result)
      * because this function is able to optimize the case
      *   result: PreTransLocalDef
-     * if `!result.contains(preTransBinding)`.
+     * if `!result.contains(preTransBinding) && !binding.isAlreadyUsed`.
```
gzm0: !result.contains(binding)

gzm0:
LGTM otherwise

sjrd:
Updated.

gzm0:
LGTM

PR #2479, opened 2016-06-20
Fix #2456: Optimize multiplications by 0 and by powers of 2.

### LINKED ISSUE #2456: Rewrite multiplications by powers of 2 as shifts

sjrd:
- `0 * n` -> `{ n; 0 }`
- `x * n` if `x == 2^k` -> `n << k`


sjrd:
Review by @nicolasstucki 
Test suite diff: https://gist.github.com/sjrd/53e8c486e992a95f2cc81caaecb4d569 (without the tests added by this PR)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2571,10 +2571,24 @@ private[optimizer] abstract class OptimizerCore(
           case (_, PreTransLit(IntLiteral(_))) =>
             foldBinaryOp(Int_*, rhs, lhs)
 
-          case (PreTransLit(IntLiteral(1)), _)  =>
-            rhs
-          case (PreTransLit(IntLiteral(-1)), _) =>
-            foldBinaryOp(Int_-, PreTransLit(IntLiteral(0)), rhs)
+          case (PreTransLit(IntLiteral(l)), _) =>
```
nicolasstucki: Consider not using lowercase L for the literal.

nicolasstucki:
That is all. Otherwise LGTM.

sjrd:
Updated.

nicolasstucki:
LGTM

PR #2482, opened 2016-06-25
Fix #2475: Support test filtering on testHtml

### LINKED ISSUE #2475: Support test filtering on testHtml
It would be really nice to be able to include only a subset of the tests in the Html, as some frameworks do. This way you can iterate on a few tests, without the overhead of the full suite.

gzm0:


gzm0:
Demo is here: https://gzm0.github.io/sjs-html-filter-demo/
Review by @sjrd
@siriux please also have a look if this is what you had in mind and whether it fits your purpose.

gzm0:
Spoke too soon about the demo... The request string is apparently too large for the server. Alternatively you can just checkout [my gh pages][1] and open `index.html` locally.

[1]: https://github.com/gzm0/gzm0.github.io

sjrd:
> The request string is apparently too large for the server (when filtering tests).

Would that cause problems if people need to serve their .html file from a local web server, e.g., to test their AJAX requests?

To mitigate this issue, maybe you could allow `?include=...` or `?exclude=...`, and choose whichever is shorter (usually that would be `include`, unless all tests are run. Of course it doesn't solve all cases (e.g., when half of the tests are selected), but I guess in practice it would cover typical use cases.

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
```diff
@@ -39,14 +40,26 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
   }
 
   def main(): Unit = {
+    // Note: Test filtering is currently done based on the fully qualified name
+    // of a test. While this is reasonable in most cases, there could be a test
+    // that is run by multiple test frameworks.
```
sjrd: `/* */` for multi-line comment.

sjrd:
Otherwise LGTM

gzm0:
> Would that cause problems if people need to serve their .html file from a local web server, e.g., to test their AJAX requests?

I do not know. [This SO post][1] suggests that 2000 characters is a reasonable limit on URI size. The URL from the runner are far beyond that (up to 30KB).

> To mitigate this issue, maybe you could allow ?include=... or ?exclude=..., and choose whichever is shorter (usually that would be include, unless all tests are run. Of course it doesn't solve all cases (e.g., when half of the tests are selected), but I guess in practice it would cover typical use cases.

Yes, that would probably solve most use-cases. However, there is a more fundamental thought behind creating an exclude list: The runner wont accidentally ignore newly added tests (on reload). I think this is quite an important property and I'd really like to find a solution that has this property.

[1]: http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

sjrd:
Actually a hash of the *set* of *excluded* test list would be better.

Basically if using the `?include=` variant when clicking on "run selected/failed only), the program also computes `excludedTests.toSet.##`, and adds `&excludedHash=$hash`. When loading an `?include=` query string, the program also computes `excludedTests.toSet.##`, and if that differs from `&excludedHash=`, then show the warning.

siriux:
@gzm0 Yes, that's what I was thinking. Thanks !

gzm0:
@sjrd I updated the multi-line comment thing. Do you think we can merge this and open an issue for the filtering? Or do you want it resolved in this PR?

sjrd:
LGTM. We can defer the filtering issue to another PR, indeed.

PR #2483, opened 2016-06-26
Fix #2450: Locally dce VarDefs in keepOnlySideEffects.

### LINKED ISSUE #2450: Some empty If node is not locally dce'ed in for-range
The following snippet:
```scala
  def rangeForeach1(n: Int): Unit = {
    for (i <- 0 until n)
      println(i)
  }
```
somehow produces the following JS code:
```js
$c_Lhelloworld_HelloWorld$.prototype.rangeForeach1__I__V = (function(n) {
  var isEmpty$4 = (n <= 0);
  if (isEmpty$4) {
    /*<skip>*/
  };
  var lastElement$4 = (isEmpty$4 ? (-1) : (((-1) + n) | 0));
  inlinereturn$8: {
    if ((!isEmpty$4)) {
      var i = 0;
      while (true) {
        var v1 = i;
        $m_Lhelloworld_HelloWorld$().println__O__V(v1);
        if ((i === lastElement$4)) {
          break inlinereturn$8
        };
        i = ((1 + i) | 0)
      }
    }
  }
});
```
Although the `while` loop itself is nice, there is a mysterious empty `if` lying around:
```js
  if (isEmpty$4) {
    /*<skip>*/
  };
```
Note that the above is the `JSPrinter`'s way of displaying an `If(VarRef("isEmpty$4"), Skip(), Skip())`, which we could also write as:
```js
  if (isEmpty$4) {
    /*<skip>*/
  } else {
    /*<skip>*/
  };
```
It's obviously dead-code in statement position, so it should be removed. We even have code in the optimizer that is supposed to properly dce an `if` node in statement position at https://github.com/scala-js/scala-js/blob/c30cfe5cc9ef4a8038594d71a6d98a29feed2d8a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala#L1197-L1201

We should figure out why this happens and fix it. Although that particular example is harmless (the JS VM can obviously dce the above code away), it probably hides a more important shortcoming of the optimizer.

sjrd:
When keeping only the side effects of a `Block`, we know that only the last expression can be side-effect-free, by construction. However, after removing the side effects of the expression, if it collapses to `Skip()`, we can retry removing the side effects of the next to last statement, in case it is a `VarDef`. At this point, we know that a `VarDef` would be used by no expression, and it can therefore be recursively removed.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -1210,8 +1210,13 @@ private[optimizer] abstract class OptimizerCore(
   private def keepOnlySideEffects(stat: Tree): Tree = stat match {
     case _:VarRef | _:This | _:Literal =>
       Skip()(stat.pos)
+    case VarDef(_, _, _, rhs) =>
+      keepOnlySideEffects(rhs)
     case Block(init :+ last) =>
-      Block(init :+ keepOnlySideEffects(last))(stat.pos)
+      keepOnlySideEffects(last) match {
+        case Skip()      => keepOnlySideEffects(Block(init)(stat.pos))
+        case lastEffects => Block(init :+ lastEffects)(stat.pos)
+      }
     case LoadModule(ClassType(moduleClassName)) =>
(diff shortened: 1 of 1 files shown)


sjrd:
`partestSuite/testOnly -- --fastOpt` succeeded locally.

gzm0:
LGTM

PR #2484, opened 2016-06-26
Fix #2457: Remove useless breaks and labels in JSDesugaring.

### LINKED ISSUE #2457: Remove useless labels during desugaring
If we look at the output of a typical for-range loop, we can see the following:
```js
  inlinereturn$8: {
    if ((!isEmpty$4)) {
      var i = 0;
      while (true) {
        var v1 = i;
        $m_Lhelloworld_HelloWorld$().println__O__V(v1);
        if ((i === lastElement$4)) {
          break inlinereturn$8
        };
        i = ((1 + i) | 0)
      }
    }
  }
```
This is a bit ugly because of the labeled block. It would be nice if we could emit the following instead:
```js
  if ((!isEmpty$4)) {
    var i = 0;
    while (true) {
      var v1 = i;
      $m_Lhelloworld_HelloWorld$().println__O__V(v1);
      if ((i === lastElement$4)) {
        break
      };
      i = ((1 + i) | 0)
    }
  }
```
This is actually valid (replacing the `break inlinereturn$8` by `break`) because the target of an unqualified `break` at this point is the same as that of `break inlinereturn$8`. Then, the label `inlinereturn$8` is not used anymore, and can be removed.

We can analyze during desugaring whether `break`s to labels can be replaced by unqualified `break`, by maintaining, during the transformation, a set of labels from which a simple `break` would return. If the current set contains the label of a `break`, we can replace that `break` by an unqualified one.

sjrd:
The desugaring phase maintains two sets of label names:

First, a set of labels for which we are in "tail position", i.e., breaking to that label is equivalent to no-op. `Break`s to labels in that set can be replaced by `Skip()`, allowing the removal of the label altogether. For example, in
```javascript
var y;
lbl: {
  var x = 5;
  if (b) {
    y = x + 3;
    break lbl;
  } else {
    y = x + 5;
  }
}
```
the `break lbl` is in tail position for the label `lbl`, and can therefore be removed. After removal of the labeled block itself, the snippet rewrites as:
```javascript
var y;
var x = 5;
if (b) {
  y = x + 3;
} else {
  y = x + 5;
}
```

Second, a set of "default" labels, for which `break` is equivalent to `break lbl`. This is the set of tail-position labels of the closest enclosing `while/do..while/switch` "loop". Breaks to labels in that set can be replaced by `break` without label, also allowing the removal of the labeled block. For example, in
```javascript
var y;
lbl: {
  var i = 0;
  while (true) {
    if (x == 5) {
      y = 3;
      break lbl;
    }
    i = i - 1;
  }
}
```
the `break lbl` is in "default" position of the label `lbl`, since the closest enclosing `while/do..while/switch` is in tail position of `lbl`. The snippet can therefore be rewritten as:
```javascript
var y;
var i = 0;
while (true) {
  if (x == 5) {
    y = 3;
    break;
  }
  i = i - 1;
}
```
This is particularly interesting for `Range.foreach`, because, when inlined, its `return` statement becomes a `break` to an `inlinereturn` label, which is a default label at the point of the `return`. We can therefore avoid the useless labeled block for this common case.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
@@ -13,6 +13,8 @@ import scala.language.implicitConversions
 
 import scala.annotation.{switch, tailrec}
 
+import scala.collection.mutable
+
 import org.scalajs.core.ir
 import ir._
 import ir.Position._
@@ -107,6 +109,90 @@ import java.io.StringWriter
  *    Scala.js class encoding into account.
  *  * And tiny details.
  *
+ *  --------------------------------------------------------------------------
+ *
+ *  About `Labeled` blocks, this phase maintains two sets of label names.
+ *
+ *  First, `tailPosLabels`, a set of labels for which we are in "tail
+ *  position", i.e., breaking to that label is equivalent to no-op. `Break`s to
+ *  labels in that set can be replaced by `Skip()`, allowing the removal of
+ *  the label altogether. For example, in
+ *
+ *  {{{
+ *  var y;
+ *  lbl: {
+ *    var x = 5;
+ *    if (b) {
+ *      y = x + 3;
+ *      break lbl;
+ *    } else {
+ *      y = x + 5;
+ *    }
+ *  }
+ *  }}}
+ *
+ *  the `break lbl` is in tail position for the label `lbl`, and can therefore
+ *  be removed. After removal of the labeled block itself, the snippet rewrites
+ *  as:
+ *
+ *  {{{
+ *  var y;
+ *  var x = 5;
+ *  if (b) {
+ *    y = x + 3;
+ *  } else {
+ *    y = x + 5;
+ *  }
+ *  }}}
+ *
+ *  `tailPosLabels` is a property of the *evaluation order*, and is therefore
+ *  passed explicitly in `transform` methods (as opposed to being stored in the
+ *  environment).
+ *
+ *  Second, `defaultBreakTargets`, a set of "default" labels, for which `break`
+ *  is equivalent to `break lbl`. This is the set of tail-position labels of the
+ *  closest enclosing `while/do..while/switch` "loop". `Break`s to labels in
+ *  that set can be replaced by `break` without label, also allowing the removal
+ *  of the labeled block. For example, in
+ *
+ *  {{{
+ *  var y;
+ *  lbl: {
+ *    var i = 0;
+ *    while (true) {
+ *      if (x == 5) {
+ *        y = 3;
+ *        break lbl;
+ *      }
+ *      i = i - 1;
+ *    }
+ *  }
+ *  }}}
+ *
+ *  the `break lbl` is in "default" position of the label `lbl`, since the
+ *  closest enclosing `while/do..while/switch` is in tail position of `lbl`.
+ *  The snippet can therefore be rewritten as:
+ *
+ *  {{{
+ *  var y;
+ *  var i = 0;
+ *  while (true) {
+ *    if (x == 5) {
+ *      y = 3;
+ *      break;
+ *    }
+ *    i = i - 1;
+ *  }
+ *  }}}
+ *
+ *  This is particularly interesting for `Range.foreach`, because, when inlined,
+ *  its `return` statement becomes a `break` to an `inlinereturn` label, which
+ *  is a default label at the point of the `return`. We can therefore avoid the
+ *  useless labeled block for this common case.
+ *
+ *  `defaultBreakTargets` is property of the *scope*, and is therefore stored
+ *  in the environment.
+ *
  *  @author Sébastien Doeraene
  */
 private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
@@ -140,7 +226,7 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
       tree: Tree, isStat: Boolean): js.Tree = {
     val desugar = new JSDesugar(classEmitter, enclosingClassName, None)
     if (isStat)
-      desugar.transformStat(tree)(Env.empty)
+      desugar.transformStat(tree, Set.empty)(Env.empty)
     else
       desugar.transformExpr(tree)(Env.empty)
   }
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -207,6 +209,7 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
     // LHS'es for labeled expressions
 
     var labeledExprLHSes: Map[Ident, Tree] = Map.empty
```
gzm0: I know this is unrelated, but shouldn't this simply be in the environment?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1077,18 +1107,24 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
+                  pushLhsInto(newLhs, thenp, tailPosLabels),
+                  pushLhsInto(newLhs, elsep, tailPosLabels))
             }
           }
 
         case Try(block, errVar, handler, finalizer) =>
           extractLet { newLhs =>
+            val blockAndHandlerTailPosLabels =
+              if (finalizer == EmptyTree) tailPosLabels
+              else Set.empty[String]
```
gzm0: Are you 100% positive you need this?

Consider:
```javascript
while (true) { try { break; } finally { console.log("finally"); } } // prints "finally"
```

Needing this would make JavaScript an even more broken language :P

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1129,14 +1165,16 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
+            implicit val env = env0.withDefaultBreakTargets(tailPosLabels)
             extractLet { newLhs =>
               val newCases = {
                 for {
                   (values, body) <- cases
                   newValues = (values map transformExpr)
                   // add the break statement
-                  newBody = js.Block(pushLhsInto(newLhs, body), js.Break())
+                  newBody = js.Block(
+                      pushLhsInto(newLhs, body, tailPosLabels),
```
gzm0: Just to verify my understanding: You are allowed to pass `tailPosLabels` down here, since we manually add `js.Break()` at the end of this block, right?

Further: In some sense, adding the `tailPosLabels` to the default break targets is correct, but redundant, right?

gzm0:
Also: Could you move the commit message to a comment in the header? I think it is very valuably (if not indispensable) in understanding what the label positions mean.
You might also want to consider adding a comment why the tailPosLabels are not part of the environment.

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Yes, it should. #2495.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Nope, you're absolutely right. We can always keep `tailPosLabels`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: > Just to verify my understanding: You are allowed to pass `tailPosLabels` down here, since we manually add `js.Break()` at the end of this block, right?

Indeed.

> Further: In some sense, adding the `tailPosLabels` to the default break targets is correct, but redundant, right?

No, it is not redundant, because you could have a break to the label not in tail position of the `case`. For example:
```js
x: {
  switch (y) {
    case 1: {
      if (b) {
        break x;
      }
      foo();
      break; // coming from the desugaring of Match
  }
}
```
The `break x` can be transformed into `break` because `x` is in the default break targets at this point, but it is definitely not in tail pos labels.

sjrd:
> Could you move the commit message to a comment in the header? I think it is very valuably (if not indispensable) in understanding what the label positions mean.

OK

> You might also want to consider adding a comment why the tailPosLabels are not part of the environment.

OK. Short answer: because it depends on the control flow, whereas stuff in the environment depends on scoping.

gzm0:
> Short answer: because it depends on the control flow, whereas stuff in the environment depends on scoping.

Yes, I know. But it takes quite some reading the code to figure that out :) Essentially before you read the code that transforms a `Block`, you don't understand :)

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -207,6 +293,7 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
     // LHS'es for labeled expressions
 
     var labeledExprLHSes: Map[Ident, Tree] = Map.empty
+    val usedLabels = mutable.Set.empty[String]
```
gzm0: Forgot about this, sorry: Just to confirm, this means that label names need to be unique inside a method body. If so, we should probably update the IR checker to check this.

gzm0:
LGTM if my assumption about labeled block scope is correct.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Yes. In fact, so must local variable names.

sjrd: #2498

PR #2485, opened 2016-06-26
Small cleanups in RuntimeLong

sjrd:


sjrd:
Review by @nicolasstucki

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -666,7 +659,7 @@ final class RuntimeLong(val lo: Int, val hi: Int)
 
 object RuntimeLong {
   private final val TwoPow32 = 4294967296.0
-  private final val TwoPow53 = 9223372036854775808.0
+  private final val TwoPow63 = 9223372036854775808.0
```
sjrd: See commit message:

> The constant `TwoPow53` was incorrectly named all along. Its value is actually `Math.pow(2, 63)`. It was correctly used in `fromDouble` for its actual value of 2^63, so there was no bug. This commit changes its name to `TwoPow63` to reflect its actual value.

nicolasstucki:
LGTM

PR #2488, opened 2016-06-28
Stack-allocate RuntimeLongs and inline their operations.

sjrd:
This is a major review of how the optimizer handles Longs. Longs have been notoriously slow in Scala.js, and this commit dramatically improves their performance (with a 15x improvement on the SHA-512 benchmark).

The main idea is two-fold:

* Mark `RuntimeLong` with `@inline`
* Have the optimizer eagerly expand all `Long` operations into the
  method calls that implement them.

This combination makes it so that `RuntimeLong`s are virtualized
and stack-allocated, and that all their operations are inlined and
work on the `Int` values `lo` and `hi`. We also accompany those
two core changes with a few tricks--such as allocating new
instances of `RuntimeLong` on escape, rather than canceling, which
we can do because the identity of `RuntimeLong`s is not meaningful--to push the amount of stack allocation as much as possible.

This has two advantages. The first and foremost is that it is not necessary to allocate a new instance of `RuntimeLong` on heap for every single operation. Instead, instances are only created when the value escapes the scope of the current method.

The second advantage is that the optimizer can micro-optimize the operations specifically for their call site. For example, if one operand is a constant, as in `1L & x`, the lo and hi parts of the constant are actually constant themselves, and will be treated as such while optimizing "the inside" of the `&` operation. Which gives something like:

    val result$lo = 1 & x$lo;
    val result$hi = 0 & x$hi;

the latter then reduces to `0`, which in turns causes `result$hi` to be constant for subsequent operations.

Here are some of my favorite rewritings.

* `(x >>> 32).toInt` becomes `x$hi` (this was previously special-cased in the optimizer, but is now a consequence of the new optimizations)
* `x >= 0L` becomes `x$hi >= 0`, and `x < 0L` becomes `x$hi < 0`
* `(xInt.toLong << 32) | (yInt.toLong & 0xffffffffL)` becomes `val result$lo = xInt; val result$hi = yInt;`
* Even more impressive, `(xInt.toLong << 32) + (yInt.toLong & 0xffffffffL)` (with a `+` instead of `|`) also reduces to `val result$lo = xInt; val result$hi = yInt;` (but if the operands are swapped, that doesn't work so well anymore).
* `(xInt.toLong & 0xffffffffL) > (yInt.toLong & 0xffffffffL)` rewrites as `(xInt ^ 0x80000000L) > (yInt ^ 0x80000000L)`, and similarly for the other comparison operators (note that this is the "optimized by hand" version of comparing `xInt` and `yInt` as if they were unsigned integers)


sjrd:
Review by @gzm0 and @nicolasstucki. This one deserves two reviewers, I think :p

The first commit belongs to PR #2484, so don't review that one here, and don't merge.

sjrd:
The full diff is here, if you're brave enough to look at it: https://gist.github.com/sjrd/190e3bb439cf668544738d4d69300262

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -71,255 +59,420 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
   }
 
-  def *(b: RuntimeLong): RuntimeLong = {
-    val alo = a.lo
-    val ahi = a.hi
-    val blo = b.lo
-    val bhi = b.hi
+  private def fromDoubleImpl(value: Double): Int = {
+    if (value.isNaN) {
+      hiReturn = 0
```
nicolasstucki: Nice ;)

nicolasstucki:
LGTM

gzm0:
I know I still need to review this :) Is it super urgent? I have some cycles for sure on Friday.

sjrd:
Oh no don't worry. I was just doing some triage ;-) Take your time.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1540,8 +1540,15 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
+
+          /* If the receiver is maybe a hijacked class instance, and there
+           * exists a hijacked method helper for the method, use it. Methods
+           * that do not have helpers are
+           * - Reflective proxies
+           * - Methods of RuntimeLong that are not also in java.lang.Long
+           */
           if (isMaybeHijackedClass(receiver.tpe) &&
-              !Definitions.isReflProxyName(method.name)) {
+              hijackedClassMethodToHelperName.contains(method.name)) {
```
gzm0: Just to check my understanding: This change is necessary, since the optimizer now emits `RuntimeLongs` where before it emitted primitive `Long`s, right?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -216,8 +240,23 @@ private[optimizer] abstract class OptimizerCore(
...
+
+        case (ClassType(LongImpl.RuntimeLongClass),
+            ClassType(Definitions.BoxedLongClass)) =>
+          true
+
+        case _ =>
+          false
+      }
+    }
+  }
```
gzm0: Is there a reason this change is not directly in `ir/Types.scala`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -524,9 +585,15 @@ private[optimizer] abstract class OptimizerCore(
           pretransformExpr(expr) { texpr =>
             texpr.tpe match {
               case RefinedType(base: ReferenceType, true, false) =>
+                val actualBase = base match {
+                  case ClassType(LongImpl.RuntimeLongClass) =>
+                    ClassType(Definitions.BoxedLongClass)
+                  case _ =>
+                    base
+                }
```
gzm0: Shouldn't this move to `JSDesugaring`? IIUC, the following will create different results depending on whether the optimizer is on:

```scala
val l = new RuntimeLong
println(l.getClass)
```

I'd feel more comfortable by saying that a `RuntimeLong` is seen as a Long consistently...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1130,6 +1211,31 @@ private[optimizer] abstract class OptimizerCore(
...
+            createNewLong(lo, hi)
+          case recordVarRef: VarRef =>
+            createNewLong(recordVarRef)
+          case _ =>
+            val varRefIdent = Ident(freshLocalName("x", mutable = false))
+            val recordVarDef =
+              VarDef(varRefIdent, tree.tpe, mutable = false, tree)
+            Block(recordVarDef, createNewLong(recordVarDef.ref))
+        }
+
```
gzm0: IIUC, we are currently living with the fact that the following code will be much worse off, right?

```scala
val l = 1L
f(l)
g(l)
h(l)
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1756,16 +1869,23 @@ private[optimizer] abstract class OptimizerCore(
       // java.lang.Long
 
       case LongToString =>
-        contTree(Apply(firstArgAsRTLong, "toString__T", Nil)(StringClassType))
```
gzm0: `firstArgAsRTLong` is unused now.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2976,6 +3176,13 @@ private[optimizer] abstract class OptimizerCore(
           case (PreTransBinaryOp(Long_+, LongFromInt(x), LongFromInt(y)),
               PreTransLit(LongLiteral(Int.MaxValue))) =>
             trampoline {
+              /* HACK: We use an empty scope here for `withNewLocalDefs`.
+               * It's OKish to do that because we're only defining Ints, and
+               * we know withNewLocalDefs is not going to try and inline things
+               * when defining ints, so we cannot go into infinite inlining.
+               */
+              val emptyScope = Scope.Empty
```
gzm0: Why is this suddenly required?

gzm0: Nevermind. Saw it below.

gzm0:
Just to harden my understanding: What happens if I write:

```scala
val x = new Long("1")
```

?

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Yes and no. What can happen now it's that we have a `j.l.Long` on which we call e.g. `compareTo`. That is inlined, and calls `j.l.Long.compare`, which we reroute as an intrinsic to `RuntimeLong.compareTo(RuntimeLong)`, but that is not inlined. We end up with `(x: j.l.Long).compareTo(RuntimeLong)`. That method does not exist in `j.l.Long` and does not have a helper. Instead, or must be desugared straightforwardly as a method call.

Before, this did not happen because the optimizer would introduce an `asInstanceOf[RuntimeLong]` on the lhs. It doesn't do that anymore because it now considers that `j.l.Long <: RuntimeLong` (which it should have all along, but not doing it was not a problem before).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes. `ir/Types.scala` doesn't know about `RuntimeLong`, not should it. That is a link-time implementation detail.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: It is already in desugar, for the run-time semantics. (Actually in `scalajsenv.js`) Here it is for constant-folding of `GetClass`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes. Well that exact snippet was already "bad", because `1L` is a constant and so `l` is constant-folded anyway, causing three allocations. 

But now this is true in more cases, for example if we replace `val l = x + 1L`.

sjrd:
> What happens if I write:
> `val x = new Long("1")`
> ?

The *compiler* rewrites that to `j.l.Long.valueOf("1")`, as it has always done since 0.5.0. From there, nothing weird happens. It's just a regular call to `j.l.Long$.valueOf__T__J`, which is not inlined and returns a `RuntimeLong` instance. Then at call site that is immediately extracted in a record with `x$lo` and `x$hi`.

gzm0:
In that case, LGTM modulo unused `firstArgAsRTLong`.

sjrd:
Updated. Rebased + removed `firstArgAsRTLong` as well as `asRTLong`.

PR #2499, opened 2016-07-04
Fix #2445: Look inside PolyType for NullaryMethodType.

### LINKED ISSUE #2445: Paren-free def's with type-parameters are handled as a function in js.native traits.
The two following facades should return the same results: 

```
@js.native
trait GoodFacade[T] extends js.Object {
  def test: js.Array[Edge[T]] = js.native
}
```
```
@js.native
trait BadFacade[+T] extends js.Object {
  def test[T2 >: T]: js.Array[Edge[T2]] = js.native
}
```

When iterating over the array return by the ```test``` method the first facade will behave as expected, but the second one will fail with an ```test is not a function``` error.

A simplified example can be found [here](https://github.com/xsistens/scalajs-facade-bug).

sjrd:
Methods with type parameters but without parentheses have a type of the shape `PolyType(_, _: NullaryMethodType)` instead of `_: NullaryMethodType` for monomorphic methods. This commit fixes `isJSGetter` to correctly identify those as getters.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
@@ -109,7 +109,11 @@ trait JSGlobalAddons extends JSDefinitions
     /** has this symbol to be translated into a JS getter (both directions)? */
     def isJSGetter(sym: Symbol): Boolean = {
       sym.tpe.params.isEmpty && enteringUncurryIfAtPhaseAfter {
-        sym.tpe.isInstanceOf[NullaryMethodType]
+        sym.tpe match {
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
I believe that you should also add the two examples from #2445 as tests.

nicolasstucki:
That is all

sjrd:
OK I have added the second example as an additional test case (with a slight modification not to depend on `Edge` which is not provided in the bug report).

The first example is the normal monomorphic behavior which already worked, and for which we have zillions of other tests. So it did not make sense to add that one in this context.

nicolasstucki:
LGTM

PR #2503, opened 2016-07-04
Fix #2502: Fix the compiler handling of Int shift Long.

### LINKED ISSUE #2502: Int shift Long is not correctly handled by the compiler back-end
Was: Code from `run/t9516.scala` does not compile in 2.12.0-M5

Adding the code to `helloworld`
```scala
def main() {
    intShiftLeftLongConstantFolded()
    intShiftLeftLongAtRuntime()
    intShiftLogicalRightLongConstantFolded()
    intShiftLogicalRightLongAtRuntime()
    intShiftArithmeticRightLongConstantFolded()
    intShiftArithmeticRightLongAtRuntime()
  }
  def intShiftLeftLongConstantFolded(): Unit = {
    assert(0x01030507 << 36L == 271601776)
    val r = 0x01030507 << 36L
    assert(r == 271601776)
  }

  def intShiftLeftLongAtRuntime(): Unit = {
    var x: Int = 0x01030507
    var y: Long = 36L
    assert(x << y == 271601776)
    val r = x << y
    assert(r == 271601776)
  }

  def intShiftLogicalRightLongConstantFolded(): Unit = {
    assert(0x90503010 >>> 36L == 151323393)
    val r = 0x90503010 >>> 36L
    assert(r == 151323393)
  }

  def intShiftLogicalRightLongAtRuntime(): Unit = {
    var x: Int = 0x90503010
    var y: Long = 36L
    assert(x >>> y == 151323393)
    val r = x >>> y
    assert(r == 151323393)
  }

  def intShiftArithmeticRightLongConstantFolded(): Unit = {
    assert(0x90503010 >> 36L == -117112063)
    val r = 0x90503010 >> 36L
    assert(r == -117112063)
  }

  def intShiftArithmeticRightLongAtRuntime(): Unit = {
    var x: Int = 0x90503010
    var y: Long = 36L
    assert(x >> y == -117112063)
    val r = x >> y
    assert(r == -117112063)
  }
```

we get

```scala
[info] Fast optimizing /Users/nastucki/GitHub/scala-js/examples/helloworld/target/scala-2.12.0-M5/helloworld-fastopt.js
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(29:19:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(29:19:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(30:15:VarDef): IntType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(43:20:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(43:20:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(44:15:VarDef): IntType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(57:19:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(57:19:BinaryOp): DoubleType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[error] file:/Users/nastucki/GitHub/scala-js/examples/helloworld/HelloWorld.scala(58:15:VarDef): IntType expected but LongType found for tree of type org.scalajs.core.ir.Trees$BinaryOp
[trace] Stack trace suppressed: run last helloworld/compile:fastOptJS for the full output.
[error] (helloworld/compile:fastOptJS) There were 9 IR checking errors.
```

sjrd:
Unlike other operations involving an Int and a Long, the specific case of Int shift Long is an Int operation, and must behave as such. In particular, the right hand side should be coerced to an Int (instead of the left hand side being coerced to a Long).

This change is similar to what was done upstream in https://github.com/scala/scala/commit/eb954cafa2737a6f06f8a802ceec133a88959f71

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2804,11 +2804,18 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       implicit val pos = tree.pos
 
-      def isLongOp(ltpe: Type, rtpe: Type) =
-        (isLongType(ltpe) || isLongType(rtpe)) &&
-        !(toTypeKind(ltpe).isInstanceOf[FLOAT] ||
-          toTypeKind(rtpe).isInstanceOf[FLOAT] ||
-          isStringType(ltpe) || isStringType(rtpe))
+      val isShift = isShiftOp(code)
+
+      def isLongOp(ltpe: Type, rtpe: Type) = {
+        if (isShift) {
+          isLongType(ltpe)
+        } else {
+          (isLongType(ltpe) || isLongType(rtpe)) &&
+          !(toTypeKind(ltpe).isInstanceOf[FLOAT] ||
+            toTypeKind(rtpe).isInstanceOf[FLOAT] ||
+            isStringType(ltpe) || isStringType(rtpe))
+        }
+      }
 
       val sources = args map genExpr
 
@@ -2897,10 +2904,15 @@ abstract class GenJSCode extends plugins.PluginComponent
           def convertArg(tree: js.Tree, kind: TypeKind) = {
             /* If we end up with a long, the op type must be float or double,
              * so we can first eliminate the Long case by converting to Double.
+             *
+             * Unless it is a shift operation, in which case the op type would
+             * be int.
              */
-            val notLong =
-              if (kind == LongKind) js.UnaryOp(js.UnaryOp.LongToDouble, tree)
-              else tree
+            val notLong = {
+              if (kind != LongKind) tree
+              else if (isShift) js.UnaryOp(js.UnaryOp.LongToInt, tree)
+              else js.UnaryOp(js.UnaryOp.LongToDouble, tree)
+            }
 
             if (opType != jstpe.FloatType) notLong
             else if (kind == FloatKind) notLong
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala
@@ -9,6 +9,9 @@ package org.scalajs.testsuite.compiler
 
 import org.junit.Test
 import org.junit.Assert._
+import org.junit.Assume._
+
+import org.scalajs.testsuite.utils.Platform
 
 /* General note on the way these tests are written:
  * We leverage the constant folding applied by the Scala compiler to write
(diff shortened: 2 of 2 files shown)


sjrd:
Review by @nicolasstucki 

There is one design decision, here: we choose to use the appropriate run-time behavior even for older versions of scalac. This means that, with Scala < 2.12.0-M5, the same code might not do the same thing on the JVM and JS. An alternative design could be to fix the behavior only in M5+, and to reproduce the behavior of Scala/JVM for earlier versions. Essentially being bug compatible.

@nicolasstucki @gzm0 WDYT?

gzm0:
> There is one design decision, here: we choose to use the appropriate run-time behavior even for older versions of scalac.

IIUC this is only relevant in cases where modulus truncation of the shift value happens, right?

So:

```scala
1 << 32L // old: 0, new: 1
```

If so, I'd say do whatever is easier to implement for us since it is really a corner case most people probably don't even know about :)

sjrd:
Yes, that is exact.

The easier to implement is definitely what I did in this PR. The alternative would add an explicit branching to compile differently according to the version of Scala. Something like
```
def scalacHasBuggyIntShiftLong = (1 << 32L) != 1

if (scalacHasBuggyIntShiftLong)
  compileBuggyIntShiftLong
else
  compileCorrectIntShiftLong
```
It's not really *harder*, but it's more *complex*, obviously.

gzm0:
I think in this case: Just do what is correct, if someone complains we fix it (i.e. re-introduce the bug conditionally).
Especially since relying on the compiler being compiled with its own version is probably also not a good idea (we don't do that for example), so we end up in an explosion of cases since constant folding and what not also needs to be done consistently...

sjrd:
Sounds reasonable.

> if someone complains we fix it (i.e. re-introduce the bug conditionally).

We will not be able to re-introduce the bug later on, as that would break correct Scala.js-only code.

gzm0:
> We will not be able to re-introduce the bug later on, as that would break correct Scala.js-only code.

We could release with a know bug. Better: We can tell people that need old Scala.js versions to write:

```scala
x << (y % 32)
```

and done.

Bottom line: Keep it as is. No bug compatibility with older Scala versions.

sjrd:
OK, sounds good.

sjrd:
> `x << (y % 32)`

The proper code is actually
```scala
x << y.toInt
```
or, equivalently, `x << (y & 31)`. `y % 32` does not do the right thing for `y < 0`.

nicolasstucki:
Implementation wise LGTM.
I also agree that we should keep it as it is.

PR #2504, opened 2016-07-05
Fix #2498: IR-check the scope of local identifiers.

### LINKED ISSUE #2498: IR-check that identifiers for labels and local variables are unique per method

sjrd:
* Identifiers for labels must be unique per method.
* Identifiers for local variables must have non-overlapping scopes.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
@@ -677,7 +677,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
             typecheckExpect(rhs, env, IntType)
           case Float_+ | Float_- | Float_* | Float_/ | Float_% =>
             typecheckExpect(lhs, env, FloatType)
-            typecheckExpect(lhs, env, FloatType)
+            typecheckExpect(rhs, env, FloatType)
           case Long_+ | Long_- | Long_* | Long_/ | Long_% |
               Long_| | Long_& | Long_^ |
               Long_== | Long_!= | Long_< | Long_<= | Long_> | Long_>= =>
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

sjrd:
Updated with disabling the check in exported methods.

gzm0:
So I'm assuming you will change this again. Or do you want to get this in and change it again?

sjrd:
I'll change this again, yes. So on standby for now.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -828,33 +850,41 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+          }
+
+          for (ParamDef(name, ptpe, mutable, rest) <- params) {
+            if (ptpe == NoType)
+              reportError(s"Parameter $name has type NoType")
+            else if (ptpe != AnyType)
+              reportError(s"Closure parameter $name has type $ptpe instead of any")
+            if (rest)
+              reportError(s"Closure parameter $name cannot be a rest parameter")
+          }
```
gzm0: Might make sense to revert this. It's not required anymore, is it?

gzm0:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: It's still required for label identifiers. Besides, it just makes sense.

gzm0: Ok, fair enough. But ParamDef can't really have a label in it, can it? :)

gzm0:
LGTM

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: `ParamDef`, no. But `captureValues`, definitely.

PR #2505, opened 2016-07-05
Fix #2495: Move labeledExprLHSes of Desugar into the environment.

### LINKED ISSUE #2495: Move labeledExprLHSes of Desugar into the environment
When it was first introduced, the desugar phase didn't have an environment, but now it should be there.

sjrd:
In the process, simplify the desugaring of `Labeled`. There was a special case for `Return` which was not necessary as of the introduction of `usedLabels` to track which labels are actually used.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2507, opened 2016-07-05
Implement java.util.Optional

andreaTP:
I hope this could be merged even if it is a java 8 api.

### DIFF
--- a/javalib/src/main/scala/java/util/Optional.scala
+++ b/javalib/src/main/scala/java/util/Optional.scala
@@ -1,0 +1,58 @@
+package java.util
+
+final class Optional[T] private (value: T) {
+
+  def get(): T = {
+    if (!isPresent())
+      throw new NoSuchElementException()
+    else
+      value
+  }
+
+  def isPresent(): Boolean = value != null
+
+  //def ifPresent(consumer: Consumer[_ >: T]): Unit
+  //def filter(predicate: Predicate[_ >: T]): Optional[U]
+  //def map[U](mapper: Function[_ >: T, _ <: U]): Optional[U]
+  //def flatMap[U](mapper: Function[_ >: T, Optional[U]]): Optional[U]
+
+  def orElse(other: T): T = {
+    if (isPresent) value
+    else other
+  }
+
+  //def orElseGet(other: Supplier[_ <: T]): T
+  //def orElseThrow[X](exceptionSupplier: Supplier[_ <: X]): T
+
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case opt: Optional[_] =>
+        (!isPresent() && !opt.isPresent()) ||
+        (isPresent() && opt.isPresent() && value.equals(opt.get()))
+      case _ => false
+    }
+  }
+
+  override def hashCode(): Int = {
+    if (!isPresent()) 0
+    else value.hashCode()
+  }
+
+  override def toString(): String = {
+    if (!isPresent()) "Optional.empty"
+    else s"Optional[$value]"
+  }
+}
+
+object Optional {
+  def empty[T](): Optional[T] = new Optional[T](null.asInstanceOf[T])
+
+  def of[T](value: T): Optional[T] = {
+    if (value == null)
+      throw new NullPointerException()
+    else
+      new Optional[T](value)
+  }
+
+  def ofNullable[T](value: T): Optional[T] = new Optional[T](value)
+}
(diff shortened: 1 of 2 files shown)


gzm0:
Jenkins, ok to test

nicolasstucki:
@andreaTP it is fine to merge anything from Java 8 API. The only thing is that you can't use in `scala-js/javalib` in to implement parts of the API that are older than Java 8.

javalib/src/main/scala/java/util/Optional.scala
```diff
@@ -0,0 +1,56 @@
+package java.util
+
+class Optional[T] private (value: T) {
```
sjrd: The class should be `final`, as in the JavaDoc.

javalib/src/main/scala/java/util/Optional.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  def get(): T = {
+    if (!isPresent())
+      throw new NoSuchElementException()
+    else value
```
sjrd: Both branches an `if` should agree on whether they're on new line or not. So this should be:
```scala
if (!isPresent())
  throw new NoSuchElementException()
else
  value
```
Alternatively, it could be:
```scala
if (!isPresent())
  throw new NoSuchElementException()
value
```
(without `else`)

javalib/src/main/scala/java/util/Optional.scala
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+object Optional {
+  def empty[T](): Optional[T] = new Optional[T](null.asInstanceOf[T])
+
+  def of[T](value: T): Optional[T] = {
+    if (value == null)
+      throw new NullPointerException()
+    else new Optional[T](value)
```
sjrd: Same thing here.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/OptionalTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    Optional.ofNullable[String](null)
+  }
+
+  @Test def should_properly_compare_optionals(): Unit = {
+    assertEquals(Optional.empty[String](), Optional.ofNullable[String](null))
+    assertEquals(Optional.of[String](""), Optional.ofNullable[String](""))
+    assertEquals(Optional.of[Int](1), Optional.ofNullable[Int](1))
+    case class Test(value: Long)
+    assertEquals(Optional.of(Test(1L)), Optional.ofNullable(Test(1L)))
+  }
```
sjrd: There should also be some `assertNotEquals`s here, to test that `equals` returns `false` when it is supposed to return `false`.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/OptionalTest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  @Test def should_check_presence_of_value(): Unit = {
```
sjrd: Consider calling this `isPresent()` or `testIsPresent()`. The `should_xyz` style is not very idiomatic JUnit; we have a bunch of those inherited from the Jasmine era, but that style should not be used for new JUnit tests.

The same comment applies to other methods. Basically the testing method should be named after the method it tests.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/OptionalTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    assertTrue(fullInt.isPresent())
+    val fullString = Optional.of[String]("")
+    assertTrue(fullString.isPresent())
+  }
+
+  @Test def should_get_value_if_present(): Unit = {
+    val emp = Optional.empty[String]()
+    assertThrows(classOf[NoSuchElementException], emp.get())
+    val fullInt = Optional.of[Int](1)
+    assertEquals(fullInt.get(), 1)
```
sjrd: The parameters should be swapped: `1, fullInt.get()`. In JUnit, the *expected* value comes first.
The same applies to other calls to `assertEquals` below.

sjrd:
That's all.

andreaTP:
I think everything is targeted now.

nicolasstucki:
@andreaTP you missed the `final` for the `class`.

andreaTP:
=_='' sorry...

thanks @nicolasstucki :-)

fixed

sjrd:
LGTM

PR #2509, opened 2016-07-06
Fix #2442: Deprecate @JSExportNamed.

### LINKED ISSUE #2442: Deprecate @JSExportNamed
`@JSExportNamed` has funny semantics, and I find it very fragile in terms of language design. Their signature significantly differ on the Scala and JavaScript sides. It is also very partial, as it only supports the case where *all* parameters are passed in one config object (so no way for one param to be positional and the others to be named, for example).

It also seems not to be used. Google reports 0 use of `@JSExportNamed`, while GitHub reports 2 uses:
* [here](https://github.com/read-write-web/rww-scala-js/blob/edc8115b3b229c37472463983e24373eaf9f8b97/src/main/scala/rww/ui/RWwApp.scala#L47) with a call [here](https://github.com/read-write-web/rww-scala-js/blob/edc8115b3b229c37472463983e24373eaf9f8b97/html/foaf.html#L13)
* [here](https://github.com/bleporini/thread-dump-analyzer/blob/2c2a1ae0a40e0a894d69ade868f854317c96dc47/src/main/scala/io/blep/tda/Controller.scala#L13) with a call [here](https://github.com/bleporini/thread-dump-analyzer/blob/2c2a1ae0a40e0a894d69ade868f854317c96dc47/test.html#L98)

The ratio benefit/cost seems to be extremely low for this feature.

I suggest we deprecate it, and remove it in 1.0.0. Any opposition?

sjrd:


compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -784,6 +787,26 @@ class JSExportTest extends DirectTest with TestHelpers {
...
+      def foo(x: Int, y: Int) = 1
+    }
+    """ hasWarns
+    """
+      |newSource1.scala:5: warning: class JSExportNamed in package annotation is deprecated: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
+      |      def foo(x: Int, y: Int) = 1
+      |          ^
+      |newSource1.scala:4: warning: class JSExportNamed in package annotation is deprecated: Use @JSExport with an explicit option bag instead. See the Scaladoc for more details.
+      |      @JSExportNamed
+      |       ^
```
sjrd: The duplicate warning is due to a weird interaction with `PreJSExports`, which creates an export bridge *with* the same annotation. Apparently deprecation warnings are handled later in the pipeline, and so the deprecation is reported twice.

TBH, I don't care ^^

sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

nicolasstucki:
You got some new fatal warnings

sjrd:
Updated.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -1009,14 +1047,15 @@ class JSExportTest extends DirectTest with TestHelpers {
...
     class A {
       @JSExportNamed("apply")
       @JSExport("foo")
       def apply(): Int = 1
     }
-    """ hasWarns
+    """ containsWarns
     """
       |newSource1.scala:7: warning: Member cannot be exported to function application. It is available under the name apply instead. Add @JSExport("apply") to silence this warning. This will be enforced in 1.0.
       |      def apply(): Int = 1
```
sjrd: This part changed since the last review.

nicolasstucki:
LGTM

PR #2490, opened 2016-06-29
Upgrade Google Closure Compiler version.

nicolasstucki:

PR #2511, opened 2016-07-09
RuntimeLong tweaks

sjrd:


sjrd:
Review by @nicolasstucki

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -182,38 +182,93 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     * constructors, we get a version with only one RuntimeLong output, which
+     * avoids reification as records by the optimizer, yielding shorter code.
+     * It is potentially slightly less efficient, except when `n` is constant,
+     * which is often the case anyway.
+     *
+     * Finally we have:
+     */
+    new RuntimeLong(
+        if ((n & 32) == 0) lo << n else 0,
+        if ((n & 32) == 0) (lo >>> 1 >>> (31-n)) | (hi << n) else lo << n)
```
nicolasstucki: Would it be better if we factor out the following?
```scala
val cond = (n & 32) == 0
val loN = lo << n
new RuntimeLong(
  if (cond) loN else 0,
  if (cond) (lo >>> 1 >>> (31-n)) | (hi << n) else loN)
}
```

nicolasstucki:
That is all.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: It's actually a good question. I've spent hours trying to decide how far to go. For example, there's a branchless version of the first `if` which is `(lo << n) & ~((n << 26) >> 31)`. I have stopped when reasoning was not enough to justify performance gains, and when changes would increase code size.

In this case, performance-wise, I have no idea what is better. I would have to look at the generated assembly (e.g., the JIT might already generate branchless code for the lo part). Code size-wise, your snippet will be longer.

sjrd:
I suggest we merge #2468 first.

nicolasstucki:
I agree that we should wait until #2468 is merged.

nicolasstucki:
Rebase and probable changes in `checksizes` needed.

sjrd:
I'll wait until #2490 is merged.

sjrd:
Rebased. Without Tighten checksizes for now so I can see what happens.

nicolasstucki:
LGTM

nicolasstucki:
Still LGTM

Issue #970, opened 2014-08-26
Allow @JSExport in constructor parameters


mysticfall:
When I code in Scala.js, I often find the need to define data model types to interoperate with JS codes. Ideally, I'd use a plain case class for that, but it seems I can't put a _@JSExport_ annotation to a construct variable like below:

``` scala
case class User(@JSExport name: String, @JSExport email: String)
```

Which leads to following exception : 
`Main.scala:9: error: You may not export a local definition`

And the following alternative doesn't work either:

``` scala
case class User(@(JSExport @field) name: String, @(JSExport @field) email: String)
```

It leads me to write something like the following, which I don't think is the ideal way to define models :

``` scala
case class User(_name: String, _email: String) {

  @JSExport
  val name = _name

  @JSExport
  val email = _email
}
```

Or alternatively,

``` scala
case class User(name: String, email: String) {
  @JSExport("name")
  def nameJS = name

  @JSExport("email")
  def emailJS = email
}
```

And finally, sometimes I wish there to be some annotation to tell all of public properties of the annotated class to be exposed as Javascript API, or an option to let _@JSExportDescendantObjects_ export all properties along with the subclasses themselves.

Original discussion thread : 
- https://groups.google.com/d/msg/scala-js/iL6r06TCIw8/IK_X3JPERh8J

PR #2514, opened 2016-07-10
Fix #2513: Annotate `@JSExport` with `@field @getter @setter`.

### LINKED ISSUE #2513: @JSExport does not work for abstract vals and vars
Similar to the issue #2197 which was about `@JSName`.

sjrd:
In addition, the compiler and the `js.use(x).as[T]` macro need to look for `@JSExport` on the getter/setter itself, if there is no accessed field. In theory, it should do so unconditionally, but backward binary compatibility demand that we still recognize exports built with a previous version of the library.

Incidentally, this provides a better fix to the old issue #970. It is not necessary to use `@(JSExport @meta.field)` for constructor parameter fields anymore.

### DIFF
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
@@ -291,20 +291,6 @@ class JSExportTest extends DirectTest with TestHelpers {
   }
 
   @Test
-  def infoExportLocal: Unit = {
-
-    """
-    class A(@JSExport val x: Int)
-    """ hasErrors
-    """
-      |newSource1.scala:3: error: You may not export a local definition. To export a (case) class field, use the meta-annotation scala.annotation.meta.field like this: @(JSExport @field).
-      |    class A(@JSExport val x: Int)
-      |             ^
-    """
-
-  }
-
-  @Test
   def noMiddleVarArg: Unit = {
 
     """
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
@@ -153,6 +153,46 @@ class ExportsTest {
     assertEquals(43, foo.y)
   }
 
+  @Test def exports_for_abstract_properties_in_class_issue_2513(): Unit = {
+    abstract class Foo {
+      @JSExport
+      val x: Int
+      @JSExport
+      var y: Int
+    }
+
+    class Bar extends Foo {
+      val x: Int = 5
+      var y: Int = 6
+    }
+
+    val bar = (new Bar).asInstanceOf[js.Dynamic]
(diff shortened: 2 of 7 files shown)


sjrd:
Review by @nicolasstucki

nicolasstucki:
The issue number of the fix seems to be wrong, #2515 is a PR.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -291,20 +291,6 @@ class JSExportTest extends DirectTest with TestHelpers {
...
   @Test
-  def infoExportLocal: Unit = {
-
-    """
-    class A(@JSExport val x: Int)
-    """ hasErrors
-    """
-      |newSource1.scala:3: error: You may not export a local definition. To export a (case) class field, use the meta-annotation scala.annotation.meta.field like this: @(JSExport @field).
-      |    class A(@JSExport val x: Int)
-      |             ^
```
nicolasstucki: Should we not keep this test with the new error message?

sjrd: There is no error anymore. It now compiles. There are new positive tests for this in `ExportsTest`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -153,6 +153,46 @@ class ExportsTest {
...
+
+    class Bar extends Foo {
+      val x: Int = 5
+      var y: Int = 6
+    }
+
+    val bar = (new Bar).asInstanceOf[js.Dynamic]
+    assertEquals(5, bar.x)
+    assertEquals(6, bar.y)
+    bar.y = 7
```
nicolasstucki: What do we expect when we do `bar.x = 8`? Does this call a setter that will throw?

nicolasstucki:
That is all.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
nicolasstucki: Ok

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
sjrd: It depends on strict mode. Basically for a `val` the JS property has a getter but no setter. Under strict mode, trying to assign a property that has no setter causes a `TypeError`. In sloppy mode this silently fails.

It's true that we don't really have tests for that, but that's orthogonal to this PR, IMO. I've filed #2519

sjrd:
> The issue number of the fix seems to be wrong, #2515 is a PR.

Indeed. It should be #2513.

sjrd:
Updated the commit message to reference #2513.

nicolasstucki:
LGTM

PR #2516, opened 2016-07-10
Fix #2497: Include or exclude list in HTML runner

### LINKED ISSUE #2497: Html runner request too large
With large test filters the request string on the HTML runner can get too large. If you are affected by this, please +1 this issue, so we know you are affected.

Solutions to the problem are discussed in #2482.

gzm0:
Listing all the excluded tests in the URL might cause a request size
that is too large if most tests are excluded. Therefore, we create an
include or an exclude list, depending on the size of each respective
set.
Further, when creating an include list, we hash the excluded tests to
avoid implicitly excluding a new test (a warning is shown if the hashes
of the implicitly excluded tests mismatch).

### DIFF
--- a/test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
+++ b/test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
@@ -44,10 +44,22 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
      * of a test. While this is reasonable in most cases, there could be a test
      * that is run by multiple test frameworks.
      */
-    val testFilter: TaskDef => Boolean = {
+    val (testFilter, optExcludedHash): (TaskDef => Boolean, Option[Int])  = {
       val search = dom.document.location.search.stripPrefix("?")
-      val excludeSet = search.split("&").map(decodeURIComponent).toSet
-      t => !excludeSet.contains(t.fullyQualifiedName)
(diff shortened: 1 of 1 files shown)


gzm0:
Review by @sjrd

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
```diff
@@ -147,6 +165,10 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
       updateCounts()
     }
 
+    def warnNewTestsAdded(): Unit = {
+      rootBox.log("Potentially excluding new tests.", "warn")
```
sjrd: This message should be a bit more user-friendly. Something like

> There are new excluded tests in the test suite. You may wish to "Run all" to rediscover all available tests.

?

sjrd:
That's all, again assuming you've tested this locally.

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
gzm0: Good point. I avoided the word "suite" though, since we use it in the interface to refer to a test "class".

gzm0:
Updated. Tested locally.

sjrd:
LGTM.

PR #2517, opened 2016-07-10
Fix #2481: Display failures on top in html runner

### LINKED ISSUE #2481: Add easy way to skip to failures in HTML runner
Currently, one sees that there are failures, but one has to scroll down manually.

gzm0:


gzm0:
Review by @sjrd

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
```diff
@@ -379,6 +396,7 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
       var className: String = js.native
       val style: Style = js.native
       var onclick: js.Function0[Boolean] = js.native
+      def insertAdjacentElement(location: String, element: Element): Unit = js.native
```
gzm0: Is it OK to use this?

sjrd: According to the [browser compat table of MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML), I hope so.

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
```diff
@@ -242,14 +250,19 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
       def log(msg: String, clss: String): dom.Element =
         body.newElement(clss = s"log $clss", text = msg, tpe = "pre")
 
+      def setNextSibling(that: TestBox): Unit = {
+        this.box.insertAdjacentElement("afterEnd", that.box)
```
sjrd: [On MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML) the doc says to use `"afterend"` all-lowercase. I suggest we do not tempt fate with alternative casing.

sjrd:
That's all.

gzm0:
Updated. Tested locally with more than one failing test, with and without test selection.

sjrd:
LGTM

PR #2526, opened 2016-07-13
Fix #2474: Check conflicting top-level exports

### LINKED ISSUE #2474: 2+ top-level exports with the same name cause trouble at run-time
What works, exporting case class
~~~ scala
// Scala
package foo
@JSExport
case class Point(
  @(JSExport @field) x: Double,
  @(JSExport @field) y: Double)
~~~
~~~ javascript
// javascript
var p = foo.Point(1,2);
~~~

What also works, export factory methods in acc. object
~~~ scala
// Scala
package foo
@JSExport
class Point(
  @(JSExport @field) val x: Double,
  @(JSExport @field) val y: Double)

@JSExport
object Point {
  @JSExport
  def create(x: Double, y: Double): Point =  new Point(x, y)
}
~~~
~~~ javascript
// javascript
var p = foo.Point().create(1,2);
~~~

What doens't works, combining both:
~~~ scala
// Scala
@JSExport
case class Point(
  @(JSExport @field) val x: Double,
  @(JSExport @field) val y: Double)

@JSExport
object Point {
  @JSExport
  def create(x: Double, y: Double): Point = new Point(x, y)
}
~~~
~~~ javascript
// javascript
var p = foo.Point(1,2);
var p = foo.Point().create(1,2);  // ERROR: uncaught exception: scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: undefined is not an instance of java.lang.Double
~~~
Calling `Point()` will iso creating the object `Point`, try to call the constructor (or factory method) of the case class `Point`, and thus miss the required arguments.

gzm0:


gzm0:
Review by @sjrd

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -137,6 +137,9 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
       assemble(infoInput, getTree, analysis)
     }
 
+    // Make sure we don't export to the same name twice.
+    checkConflictingExports(linkResult, logger, bypassLinkingErrors)
+
```
gzm0: A quick note in case you wonder: It would be nicest to do this in the analysis of course. However, we do not have the names we actually export something under in the Infos, so we can't.
This is maybe something we should consider for 1.0.0. Simply having an analyzer error would be much cleaner.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
```diff
@@ -61,6 +61,12 @@ final class LinkedClass(
   def encodedName: String = name.name
   def isExported: Boolean = classExports.nonEmpty
 
+  /** Names this class / module is exported under */
+  def exportNames: List[String] = classExports.collect {
+    case ConstructorExportDef(name, _, _) => name
+    case ModuleExportDef(name)            => name
+  }
```
sjrd: There's also `JSClassExportDef`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
sjrd: Yes, I agree.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
```diff
@@ -61,6 +61,12 @@ final class LinkedClass(
   def encodedName: String = name.name
   def isExported: Boolean = classExports.nonEmpty
 
+  /** Names this class / module is exported under */
+  def exportNames: List[String] = classExports.collect {
```
sjrd: Consider calling this `topLevelExportNames`. This makes it clear that instance methods exports are not included. It also suggests that we can reuse that when we support exporting top-level functions.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -434,4 +437,32 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
...
+      name -> classDef
+    }
+
+    val level = if (bypassLinkingErrors) Level.Warn else Level.Error
+    val errors = for {
+      (name, namesAndClasses) <- namesAndClasses.groupBy(_._1)
+      if namesAndClasses.size > 1
+    } yield {
+      logger.log(level, s"Conflicting top-level exports to $name:")
+      logger.log(level, s"The following classes/objects all export to $name:")
```
sjrd: Those two lines of error message that repeat the essential information `$name` might confuse users as two separate errors. Consider simply adding "from the following classes:" at the end of the first message.

Not sure about this.

sjrd:
That's all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
gzm0: Good point.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
gzm0: I though of it as well. Simplified.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
gzm0: #2528

gzm0:
Updated

gzm0:
I manually checked that this actually fails and reports the right classes.

sjrd:
LGTM

PR #2530, opened 2016-07-14
Fix #2512: Fix type of java.util.Arrays.asList

### LINKED ISSUE #2512: java.util.Arrays.asList cannot be called
It seems I can't call `java.util.Arrays.asList` despite it being there: https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/Arrays.scala#L590

```
[error] Referring to non-existent method java.util.Arrays$.asList([java.lang.Object)java.util.List
```

@sjrd says:
> The problem is that the signature is not binary compatible.
> Array[T] erases to Object when T is unconstrained, but the required binary signature takes an Object[].

nicolasstucki:


nicolasstucki:
Review by @sjrd

sjrd:
LGTM

PR #1669, opened 2015-05-16
Optimize away ClassTags and ArrayBuilders

sjrd:

PR #2531, opened 2016-07-14
Fix #2521: Update overrides of ClassTag and Manifest for 2.12.

### LINKED ISSUE #2521: Adapt our ClassTags to named ClassTag instances from 2.12.0-RC1
See https://github.com/scala/scala/pull/5264

sjrd:


sjrd:
Review by @nicolasstucki

scalalib/overrides-2.12/scala/reflect/ClassTag.scala
```diff
@@ -101,7 +101,7 @@ trait ClassTag[T] extends ClassManifestDeprecatedApis[T] with Equals with Serial
   // case class accessories
   override def canEqual(x: Any) = x.isInstanceOf[ClassTag[_]]
   override def equals(x: Any) = x.isInstanceOf[ClassTag[_]] && this.runtimeClass == x.asInstanceOf[ClassTag[_]].runtimeClass
-  override def hashCode = scala.runtime.ScalaRunTime.hash(runtimeClass)
+  override def hashCode = runtimeClass.##
```
nicolasstucki: Why was this one changed?

scalalib/overrides-2.12/scala/reflect/Manifest.scala
```diff
@@ -87,29 +86,14 @@ object ManifestFactory {
...
-  def Nothing: Manifest[scala.Nothing] = NothingManifest
-
   private object ByteManifest extends AnyValManifest[scala.Byte]("Byte") {
     def runtimeClass = java.lang.Byte.TYPE
     override def newArray(len: Int): Array[Byte] = new Array[Byte](len)
     override def newWrappedArray(len: Int): WrappedArray[Byte] = new WrappedArray.ofByte(new Array[Byte](len))
     override def newArrayBuilder(): ArrayBuilder[Byte] = new ArrayBuilder.ofByte()
     private def readResolve(): Any = Manifest.Byte
   }
+  def Byte: AnyValManifest[Byte] = ByteManifest
```
nicolasstucki: Why did you change the original (from scala/scala#5264) `val` into a `def`? Same for all the cases.

scalalib/overrides-2.12/scala/reflect/Manifest.scala
```diff
@@ -166,35 +156,45 @@ object ManifestFactory {
...
+  def Boolean: AnyValManifest[Boolean] = BooleanManifest
 
   private object UnitManifest extends AnyValManifest[scala.Unit]("Unit") {
     def runtimeClass = java.lang.Void.TYPE
     override def newArray(len: Int): Array[Unit] = new Array[Unit](len)
     override def newWrappedArray(len: Int): WrappedArray[Unit] = new WrappedArray.ofUnit(new Array[Unit](len))
     override def newArrayBuilder(): ArrayBuilder[Unit] = new ArrayBuilder.ofUnit()
+    override protected def arrayClass[T](tp: Class[_]): Class[Array[T]] =
+      if (tp eq runtimeClass) classOf[Array[scala.runtime.BoxedUnit]].asInstanceOf[Class[Array[T]]]
+      else super.arrayClass(tp)
```
nicolasstucki: Is this change related to scala/scala#5264 or is it fixing something else?

scalalib/overrides-2.12/scala/reflect/ClassTag.scala
sjrd: `ScalaRunTime.hash()` was deprecated in 2.12 (by me).

scalalib/overrides-2.12/scala/reflect/Manifest.scala
sjrd: That's because as a `def`, the optimizer can inline it, see the precise type `ByteManifest$`, and from there inline the methods such as `runtimeClass`. This is part of the Scala.js-specific optimizations we applied on `ClassTag` in https://github.com/scala-js/scala-js/pull/1669

nicolasstucki:
Do we have tests on the manifests?

scalalib/overrides-2.12/scala/reflect/Manifest.scala
sjrd: It comes from something else. I didn't really look at the individual changes. I just took the current state of `Manifest.scala` and reapplied Scala.js-specific optimizations on it.

nicolasstucki:
That is all.

scalalib/overrides-2.12/scala/reflect/Manifest.scala
sjrd: Blame says https://github.com/scala/scala/commit/a01d8399bbe9f3c20a7e2e9ea9c38aefc2e07ff6

sjrd:
We have tests for `ClassTag`s. The most direct ones are in https://github.com/scala-js/scala-js/blob/9bdf5abda975d1f7a17e303439e2f303a1d10e76/test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/ClassTagTest.scala, but they're also tested by so many other library tests, such as https://github.com/scala-js/scala-js/blob/9bdf5abda975d1f7a17e303439e2f303a1d10e76/test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/ArrayBuilderTest.scala

We don't test `Manifest`s per se because we do not really support them. But since `ClassTag`s and `Manifest`s are blurred at the implementation level, we have to change things in `Manifest.scala` (which are in fact tested via `ClassTag`s). For example, `ByteTag` and `ByteManifest` are the same object (`eq`).

scalalib/overrides-2.12/scala/reflect/Manifest.scala
nicolasstucki: The question is more on the lines of if it should be in the same commit.

scalalib/overrides-2.12/scala/reflect/Manifest.scala
nicolasstucki: Ok, suspected that.

nicolasstucki:
I believe that our tests do not cover `IntersectionTypeManifest`.

sjrd:
> I believe that our tests do not cover IntersectionTypeManifest.

No they don't, because that part of Manifests is Manifest-only. It's not used by `ClassTag`s. If we try to test those, the test suite won't link.

nicolasstucki:
LGTM

PR #2472, opened 2016-06-15
Fix #2471: java.io.BufferedReader should close its underlying Reader

alexandru:
Fix for issue #2471. See the JavaDoc for [BufferedReader.close](https://docs.oracle.com/javase/7/docs/api/java/io/BufferedReader.html#close()). The contract is this:

- it is idempotent (check)
- it should release any associated resources

The current implementation of `BufferedReader` doesn't release the underlying `Reader` on `close()`.

PR #2386, opened 2016-05-09
Add java.util.EventObject

Atry:
The class is part of Java Bean specification, and used by [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala)

PR #2533, opened 2016-07-14
Small javalib changes

sjrd:
Supersedes #2386 and #2472, which were lacking follow up from their authors.

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/EventObjectTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/EventObjectTest.scala
@@ -1,0 +1,49 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2016, LAMP/EPFL   **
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @nicolasstucki

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/EventObjectTest.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Assert._
+
+import java.util.EventObject
+
+class EventObjectTest {
+  @Test def getSource(): Unit = {
+    val src = new AnyRef
+    val e = new EventObject(src)
+    assertSame(src, e.getSource)
+  }
```
nicolasstucki: As `source` is a protected `var` I would check that we can modify it from a class that extends from `EventObject`. Mainly to check that the JVM behaves the same way.

nicolasstucki:
That is all.
Otherwise LGTM.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/EventObjectTest.scala
sjrd: See the commented test down below.

nicolasstucki: Oh, missed it.

PR #2535, opened 2016-07-18
[WIP] Add ScalaJSTestNGPlugin to support tests written with TestNG

soc:
Hi @sjrd, just wanted to ask whether this looks like it is going into the right direction!

Things are more or less working! Now I just need to figure out how to run TestNG on the JVM so that I can compare the outputs and the formatting, etc.


soc:
I just had a look, it seems to come from `Test` (the annotation?).
Does Scala.js have issues with default arguments in annotations?

sjrd:
I'm currently in Rome for Curry On, so I cannot a good look at the moment. 

@nicolasstucki Can you have a look at this.

project/Build.scala
```diff
@@ -399,6 +399,19 @@ object Build {
...
+      project.settings(
+        scalacOptions in Test ++= {
+          if (isGeneratingEclipse) {
+            Seq.empty
+          } else {
+            val jar = (packageBin in (testNGPlugin, Compile)).value
+            Seq(s"-Xplugin:$jar")
+          }
+        }
+      )
```
nicolasstucki: We should probably have method `private def withInternalScalaTestPlugin(project: Project)` to factor out code from `withScalaJSJUnitPlugin` and `withScalaJSTestNGPlugin`.

nicolasstucki:
I see that most of the code is shared with the junit plugin/runtime. I suggest creating a `scalajs-test-plugin` library where there would be an `abstract class ScalaJSAbstractTestPlugin` that implements all the common functionality of `ScalaJSJUnitPlugin` and `ScalaJSTestNGPlugin`. The same could apply for common functionality of the runtime.

nicolasstucki:
Actually you might be able to do something like `class ScalaJSTestNGPlugin  extends ScalaJSJUnitPlugin`.

sjrd:
Is there any reason we need this to be in core? It seems to me that it can live in its own repo.

nicolasstucki:
I do not see any requirement that forces it to be in the core. It should definitively be outside.

soc:
I had it in core because I wasn't sure whether it would make sense to factor out the common parts.

sjrd:
What common parts? In the current PR there doesn't seem to be any code sharing with other parts of the codebase, are there?

soc:
This is a rough sketch that I wanted to discuss as many parts where taken from the junit code and adapted. I wasn't sure whether you wanted to have another "common" artifact with shared code between JUnit and TestNG. If the opinion is that the TestNG stuff shouldn't go next to the JUnit stuff into the main repo, I probably won't bother with common parts. (I have enough nightmares of various Scala things having intricate dependencies, but living in different repos (looking at you, scalac and partest!).)

That leaves us with either:

- Main repo, no shared stuff.
- Main repo, shared stuff.
- Other repo, no shared stuff.

WDYT?

nicolasstucki:
I would say its own repo with no shared code. @sjrd what is your opinion?

sjrd:
The ultimate criterium that we've used so far wrt. core vs non-core is extremely easy: if we need it to make our own infrastructure and tests work, it goes in core. Otherwise it doesn't.

In this case, it seems that it doesn't fit the criterium, so it should be non-core.

soc:
Alright, thanks.

PR #2538, opened 2016-07-19
Privatize value `p` of ThenableOps

soc:
It's unlikely that `p` was exposed on purpose.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Thenable.scala
+++ b/library/src/main/scala/scala/scalajs/js/Thenable.scala
@@ -39,7 +39,7 @@ trait Thenable[+A] extends js.Object {
 }
 
 object Thenable {
-  implicit class ThenableOps[+A](val p: Thenable[A]) extends AnyVal {
+  implicit class ThenableOps[+A](private val p: Thenable[A]) extends AnyVal {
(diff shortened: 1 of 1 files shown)


soc:
@sjrd Please let me know if this is fine or if this requires some deprecation period.

sjrd:
Jenkins, test this please

sjrd:
As I feared, this is not valid in Scala 2.10.

soc:
Alternatives? deprecated + deprecatedName?

sjrd:
It can't be deprecated, since we need to use it within `ThenableOps` without deprecation warning. The only alternative I can see would be to use a particularly ugly name + `@deprecatedName`. Something like:
```scala
implicit class ThenableOps[+A](@deprecatedName('p) val __thenableOpsAnyValSelf: Thenable[A]) extends AnyVal {
  @deprecated("No replacement.", "0.6.11")
  def p: Thenable[A] = __thenableOpsAnyValSelf
}
```
and use `__thenableOpsAnyValSelf` within `ThenableOps`. It's not technically *source* compatible because we turn `val p` into a `def p`, but that's all right.

If we do this, we should search all `AnyVal`s in the codebase and apply the same treatment to all of them.

sjrd:
Closing for lack of follow up.

PR #2541, opened 2016-07-20
Consider module accessor a JSGetter

adriaanm:
Redundant on 2.11, but required for the upcoming 2.12.0-RC1,
where the field refactoring synthesizes module accessors
after uncurry, so that their type history does not even
go back as far as uncurry, where a NullaryMethodType might
have been observed otherwise.

I tried splicing NMTs in the info history of modules,
but they already are pretty hacking (a module's term symbol
is converted into a method pretty late in the game, and it
would be risky to have them start with NMTs from namers).

With this, scala-js's test suite passes on scala/scala#5141 
(well, I'm about to rewind that branch to a stable point where this holds)

# TODO
  - [x] add comment

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
@@ -108,13 +108,17 @@ trait JSGlobalAddons extends JSDefinitions
 
     /** has this symbol to be translated into a JS getter (both directions)? */
     def isJSGetter(sym: Symbol): Boolean = {
-      sym.tpe.params.isEmpty && enteringUncurryIfAtPhaseAfter {
+      /* We only get here when `sym.isMethod`, thus `sym.isModule` implies that
+       * `sym` is the module's accessor. In 2.12, module accessors are synthesized
+       * after uncurry, thus their first info is a MethodType at phase fields.
+       */
+      sym.isModule || (sym.tpe.params.isEmpty && enteringUncurryIfAtPhaseAfter {
         sym.tpe match {
           case _: NullaryMethodType              => true
           case PolyType(_, _: NullaryMethodType) => true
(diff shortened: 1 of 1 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -108,13 +108,13 @@ trait JSGlobalAddons extends JSDefinitions
 
     /** has this symbol to be translated into a JS getter (both directions)? */
     def isJSGetter(sym: Symbol): Boolean = {
-      sym.tpe.params.isEmpty && enteringUncurryIfAtPhaseAfter {
+      sym.isModule || (sym.tpe.params.isEmpty && enteringUncurryIfAtPhaseAfter {
```
adriaanm: we only get here when `sym.isMethod`, thus `sym.isModule` can only hold for the module's accessor

sjrd:
Jenkins, test this please

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
sjrd: Could you write that as a comment in the code?

See https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#non-scaladoc-comments for the style we use for comments.

sjrd:
That's all. Thanks!

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
adriaanm: Will do tomorrow!

adriaanm:
Amended my commit.

sjrd:
LGTM Jenkins, retest this please

PR #2542, opened 2016-07-21
Fix #2539: Fix java.util.ConcurrentHashMap#putIfAbsent.

### LINKED ISSUE #2539: java.util.ConcurrentHashMap#putIfAbsent is wrong
The documentation of `putIfAbsent` specifies that it returns `null` if the key doesn't exist, while the implementation uses `getOrElseUpdate` which specifies that either the existing or the new value is returned.

Other classes might also be affected.

nicolasstucki:


nicolasstucki:
Review by @sjrd

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -50,6 +50,13 @@ class ConcurrentHashMapTest extends MapTest {
     assertEquals("one", chm.get("ONE"))
   }
 
+  @Test def putIfAbcentShould_issue_2539(): Unit = {
```
sjrd: That test name looks weird. Consider `testPutIfAbsent_issue_2539`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -50,6 +50,13 @@ class ConcurrentHashMapTest extends MapTest {
...
+    val chm = factory.empty[String, String]
+    assertNull(chm.putIfAbsent("abc", "def"))
+    assertEquals("def", chm.putIfAbsent("abc", "def"))
+    assertEquals("def", chm.putIfAbsent("abc", "ghi"))
+    assertEquals("def", chm.putIfAbsent("abc", "jkl"))
```
sjrd: Maybe worth testing `putIfAbsent` with something else than `"abc"` and verify that it returns `null` again, for a different key.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -50,6 +50,13 @@ class ConcurrentHashMapTest extends MapTest {
     assertEquals("one", chm.get("ONE"))
   }
 
+  @Test def putIfAbcentShould_issue_2539(): Unit = {
+    val chm = factory.empty[String, String]
+    assertNull(chm.putIfAbsent("abc", "def"))
+    assertEquals("def", chm.putIfAbsent("abc", "def"))
```
sjrd: Maybe test `assertEquals("def", chm.get("abc"))` before that line.

sjrd:
That's all.

sjrd:
LGTM

PR #2543, opened 2016-07-22
Fix #2449: Avoid duplicating the default case in switch with guards.

### LINKED ISSUE #2449: @switch match with guards duplicate their default case body
Consider the following method:
```scala
  def switchWithGuardsExpr(x: Int, y: Int): Unit = {
    val message = x match {
      case 1            => "one"
      case 2 if y < 10  => "two special"
      case 2            => "two"
      case 3 if y < 10  => "three special"
      case 3 if y > 100 => "three big special"
      case z if y > 100 => "big " + z
      case _            => "None of those"
    }
    println(message)
  }
```
Its scalac tree looks like this:
```scala
    def switchWithGuardsExpr(x: Int, y: Int): Unit = {
      val message: String = {
        case <synthetic> val x1: Int = x;
        (x1: Int) match {
          case 1 => "one"
          case 2 => if (y.<(10))
            "two special"
          else
            "two"
          case 3 => if (y.<(10))
            "three special"
          else
            if (y.>(100))
              "three big special"
            else
              default7()
          case _ => default7(){
            if (y.>(100))
              "big ".+(scala.Int.box(x1))
            else
              "None of those"
          }
        }
      };
      HelloWorld.this.println(message)
    };
```
Note the label-jump to `default7()` in `case 3 =>`.

Consequently, its IR looks like this:
```scala
  def switchWithGuardsExpr__I__I__V(x: int, y: int) {
    val message: T = {
      val x1: int = x;
      match (x1) {
        case 1:
          "one";
        case 2:
          if ((y < 10)) {
            "two special"
          } else {
            "two"
          };
        case 3:
          if ((y < 10)) {
            "three special"
          } else if ((y > 100)) {
            "three big special"
          } else if ((y > 100)) {
            ("big " +[string] x1)
          } else {
            "None of those"
          };
        default:
          if ((y > 100)) {
            ("big " +[string] x1)
          } else {
            "None of those"
          };
      }
    };
    this.println__O__V(message)
  }
```
As you can see, the body of the default cases (`case z if y > 100 =>` and `case _ =>`) is *duplicated* in the last alternative of the `case 3 =>`, because all `case 3` have `if`s. Since the IR does not have arbitrary jumps, the compiler has to do that.

However, there is a better possible translation, using 2 `Labeled` blocks, which should look like this:
```scala
  def switchWithGuardsExpr__I__I__V(x: int, y: int) {
    val message: T = {
      val x1: int = x;
      x$2[T]: {
        $default: {
          match (x1) {
            case 1:
              return(x$2) "one";
            case 2:
              return(x$2) if ((y < 10)) {
                "two special"
              } else {
                "two"
              };
            case 3:
              return(x$2) if ((y < 10)) {
                "three special"
              } else if ((y > 100)) {
                "three big special"
              } else {
                return($default) (void 0)
              };
            default:
              /*<skip>*/;
          }
        };
        if ((y > 100)) {
          ("big " +[string] x1)
        } else {
          "None of those"
        }
      }
    };
    this.println__O__V(message)
  }
```
We should improve the compiler back-end to emit the latter IR when it finds a `Match` with guards like the above.

Something similar can be done for switch-matches in statement position.

sjrd:


sjrd:
Review by @nicolasstucki

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2508,31 +2548,31 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-      val (defaultBody, defaultLabelSym) = defaultBody0 match {
-        case LabelDef(_, Nil, rhs) if hasSynthCaseSymbol(defaultBody0) =>
-          (rhs, defaultBody0.symbol)
-        case _ =>
-          (defaultBody0, NoSymbol)
-      }
-
-      val genDefaultBody = genStatOrExpr(defaultBody, isStat)
+      val defaultLabelSym = cases.collectFirst {
+        case CaseDef(Ident(nme.WILDCARD), EmptyTree,
```
nicolasstucki: What about
```scala
x match {
  case 1 if y > 0 => a
  case 2          => b
  case n          => c 
}
```

sjrd: That would still be a `WILDCARD`, and in the body there would be a `val n: Int = x`. You can see that we already were only handling WILDCARD here: https://github.com/scala-js/scala-js/pull/2543/files/595ddc57fa831f88ae3a42b8916832417f8b79a6#diff-c9a195a739d51194e79f0655bc93bd61R2615

sjrd:
The test failure is spurious (Phantom not connected).

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
nicolasstucki: Ok

sjrd:
Jenkins, retest this please

PR #2544, opened 2016-07-22
Optimize BigInteger by making smarter use of Longs.

sjrd:
Here are a couple general ideas that have been used in several places:

* Using appropriately the unsigned division and remainder operations provided by `{Integer,Long}.{divide,remainder}Unsigned`.
* Carry variables in loops can be Ints, and reinterpreted as Longs inside the loop, which avoids having mutable records.
* When performing both unsigned division and remainder on Longs, it is much more efficient to compute the remainder from the quotient: subtraction + multiplication are way faster than one remainder.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2545, opened 2016-07-22
Fix #2491: Consider all exported properties as externs for GCC.

### LINKED ISSUE #2491: GCC name clash
reproduction project: https://github.com/fdietze/pigment/tree/fullopt-bug

```
GCC is renaming toString__T into y, and that clashes with the exported y.
IMO it shouldn't do that, because the string "y" is present in the source we give it
```

Workaround in build.sbt:
```scala
scalaJSOptimizerOptions in (Compile, fullOptJS) ~= { _.withUseClosureCompiler(false) }
```

sjrd:
The name clashes were caused by GCC renaming some Scala methods or fields to a name that was also exported. Hence, a Scala class could have, as instance member, both a (renamed) Scala method/field and a (non-renamed) exported method or property.

To prevent this, we list all exported methods and properties in the linking unit, and we feed them as externs to GCC. This forces GCC to avoid those names when renaming Scala methods and fields.

### DIFF
--- a/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
+++ b/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
@@ -75,19 +75,50 @@ final class ClosureLinkerBackend(
     module.add(new CompilerInput(ast, ast.getInputId(), false))
 
     // Compile the module
-    val closureExterns =
-      toClosureSource(ClosureLinkerBackend.ScalaJSExternsFile)
+    val closureExterns = List(
+        toClosureSource(ClosureLinkerBackend.ScalaJSExternsFile),
+        toClosureSource(makeExternsForExports(unit)))
     val options = closureOptions(output.name)
     val compiler = closureCompiler(logger)
 
     val result = logger.time("Closure: Compiler pass") {
       compiler.compileModules(
-          List(closureExterns).asJava, List(module).asJava, options)
+          closureExterns.asJava, List(module).asJava, options)
     }
 
     logger.time("Closure: Write result") {
       writeResult(result, compiler, output)
     }
+  }
+
+  /** Constructs an externs file listing all exported properties in a linking
+   *  unit.
+   *
+   *  This is necessary to avoid name clashes with renamed properties (#2491).
+   */
+  private def makeExternsForExports(linkingUnit: LinkingUnit): VirtualJSFile = {
+    import org.scalajs.core.ir.Trees._
+
+    def exportName(tree: Tree): String = (tree: @unchecked) match {
+      case MethodDef(_, StringLiteral(name), _, _, _) => name
+      case PropertyDef(StringLiteral(name), _, _, _)  => name
+    }
+
+    val exportedPropertyNames = for {
+      classDef <- linkingUnit.classDefs
+      member <- classDef.exportedMembers
+      name = exportName(member.tree)
+      if isValidIdentifier(name)
+    } yield {
+      name
+    }
+
+    val content = new java.lang.StringBuilder
+    for (exportedPropertyName <- exportedPropertyNames.distinct)
+      content.append(s"Object.prototype.$exportedPropertyName = 0;\n")
+
+    new MemVirtualJSFile("ScalaJSExportExterns.js")
+      .withContent(content.toString())
   }
 
   private def closureCompiler(logger: Logger) = {


sjrd:
Review by @gzm0.
Note that this is basically untestable, since the bug only manifests itself in very specific combinations of code that cause GCC to, unfortunately, cause a clash.
I have however locally tested that this fixes [the repo mentioned in the issue](https://github.com/fdietze/pigment/tree/fullopt-bug).

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -75,21 +75,51 @@ final class ClosureLinkerBackend(
...
+          case MethodDef(_, StringLiteral(name), _, _, _) => name
+          case PropertyDef(StringLiteral(name), _, _, _)  => name
+        }
+      }
+      allNames.filter(isValidIdentifier).distinct
+    }
+
+    val content = new java.lang.StringBuilder
+    for (exportedPropertyName <- collectAllExportedPropertyNames())
+      content.append(s"Object.prototype.$exportedPropertyName = 0;\n")
```
gzm0: Just to test my understanding:

We can't set this on the real prototype of the class, because:

- It is not visible in this scope.
- It is a total abuse of the externs feature.

gzm0: Also, is `= 0` intentional, or should this be `= void 0`?

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -75,21 +75,51 @@ final class ClosureLinkerBackend(
...
+        (exportedM.tree: @unchecked) match {
+          case MethodDef(_, StringLiteral(name), _, _, _) => name
+          case PropertyDef(StringLiteral(name), _, _, _)  => name
+        }
+      }
+      allNames.filter(isValidIdentifier).distinct
+    }
+
+    val content = new java.lang.StringBuilder
+    for (exportedPropertyName <- collectAllExportedPropertyNames())
```
gzm0: Consider:

```scala
def exportName(tree: Tree): String = (tree @unchecked) match {
  case MethodDef(_, StringLiteral(name), _, _, _) => name
  case PropertyDef(StringLiteral(name), _, _, _)  => name
}

for {
  classDef <- linkingUnit.classDefs
  member <- classDef.exportedMembers
  name = exportName(member)
  if isValidIdentifier(name)
} {
  content.append(/* snip */)
}
```

Seems much clearer to me.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: > It is a total abuse of the externs feature.

The externs feature is a total abuse of the JavaScript syntax, you mean :-p

It is a language of its own, that has the syntax of JavaScript, but completely crazy "semantics". All it cares is recognizing the names of 1) global variables and 2) properties. There is no *execution* of externs. Their semantics is only "static": they are a very obscure syntax to define 2 sets of names. It does not matter on which object you declare properties, or what values you give them. The only thing that matters is that they are defined somewhere.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -75,21 +75,51 @@ final class ClosureLinkerBackend(
...
 
     logger.time("Closure: Write result") {
       writeResult(result, compiler, output)
     }
   }
 
+  /** Constructs an externs file listing all exported properties in a linking
+   *  unit.
+   *
+   *  This is necessary to avoid #2491.
```
gzm0: Consider:

This is necessary to avoid name clashes with renamed properties (#2491).

It's not much more text, but will help a potential reader *much* more.

gzm0:
That's all.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: > The externs feature is a total abuse of the JavaScript syntax, you mean :-p

Ah, I see. I thought this thing is smarter than that... Fair enough :P

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: That fails to capture the `.distinct`. In other words, it potentially redefines many times the same property. It's not a big deal, but it's less elegant IMO.

It's still possible with an additional mutable `Set` to remember what we already emitted, like this:
```scala
val emittedProperties = mutable.Set.empty[String]

for {
  classDef <- linkingUnit.classDefs
  member <- classDef.exportedMembers
  name = exportName(member)
  if isValidIdentifier(name)
} {
  if (emittedProperties.add(name))
    content.append(/* snip */)
}
```

gzm0: Oh... I didn't even notice this one :P 

In that case, how about:

```scala
val exportedPropertyNames = linkingUnit.classDefs.flatMap(_.exportedMembers)
    .map(exportName).filter(isValidIdentifier).distinct

for (exportedPropertyName <- exportedPropertyNames)
  content.append(/* snip */)

Comment (sjrd) [inline tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala]: Meh ... I preferred the for comprehension to this long chain of methods. How about this:
```scala
val exportedPropertyNames = for {
  classDef <- linkingUnit.classDefs
  member <- classDef.exportedMembers
  name = exportName(member)
  if isValidIdentifier(name)
} yield {
  name
}

for (exportedPropertyName <- exportedPropertyNames.distinct)
  content.append(/* snip */)
```
?

gzm0: Works for me :)

sjrd:
Updated.

gzm0:
LGTM

PR #2546, opened 2016-07-22
Micro-improvements to {highest,lowest}OneBit.

sjrd:
`Integer.highestOneBit` is now branchless.

`Long.highestOneBit` (resp. `lowestOneBit`) now uses a branch only in the `lo` part (resp. `hi` part), the other being branchless, which generates less JavaScript code.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

PR #2550, opened 2016-08-01
Fix default overload

gzm0:


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -667,8 +665,14 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
 
             if (isRawJSType(trgSym.toTypeConstructor)) {
-              assert(isScalaJSDefinedJSClass(defaultGetter.owner))
-              genApplyJSClassMethod(trgTree, defaultGetter, defaultGetterArgs)
+              if (isScalaJSDefinedJSClass(defaultGetter.owner)) {
+                genApplyJSClassMethod(trgTree, defaultGetter, defaultGetterArgs)
+              } else {
+                reporter.error(param.pos, "You cannot override a method of a " +
+                    "js.native class/trait with a default argument without a " +
+                    "default argument.")
```
sjrd: The "with default argument without default argument part" is confusing. Consider a positive phrasing instead,  like

    When overriding a native method with default arguments, the overriding method must explicitly repeat the default arguments.

sjrd:
Do we already have a pos test that when the default parameter is indeed explicitly specified, then it works?

sjrd:
That's all.

gzm0:
Updated.

sjrd:
LGTM

PR #2555, opened 2016-08-08
Fix #2554: Don't ignore code inside matchEnds

### LINKED ISSUE #2554: String concatenation and Options interaction
The minimal test case I got to reproduce the bug is the following:

~~~ scala
val foo: Option[Int] = Some(42)
println("foo = " ++ (foo match { case Some(0) => "zero" case _ => "unknown" }))
~~~

It works on the JVM (it prints « foo = unknown »), but crashes on JS with the following “stacktrace”:

~~~
Uncaught TypeError: that.seq__sc_TraversableOnce is not a function
$s_sc_TraversableLike$class__$$plus$plus__sc_TraversableLike__sc_GenTraversableOnce__scg_CanBuildFrom__O @ client-fastopt.js:2653
$c_Lmet_Main$.main__V @ client-fastopt.js:12453
$c_Lmet_Main$.$$js$exported$meth$main__O @ client-fastopt.js:12470
$c_Lmet_Main$.main @ client-fastopt.js:12473
(anonymous function) @ (index):14
~~~

If I remove the string concatenation or the Option (e.g. if `foo` has type `Int` or `String`), it works well. But when I have both the concatenation (using `++`) and the pattern matching on the `Option`, then I get the error…

gzm0:


gzm0:
Running a [nightly build](https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/4349/)

gzm0:
Review by @sj

julienrf:
Thanks @gzm0!

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/MatchTest.scala
```diff
@@ -62,4 +62,17 @@ class MatchTest {
     assertEquals("None of those", test(5, 20))
   }
 
+  // #2554
+  @Test def matchWithNonIdentityEndMatch(): Unit = {
```
sjrd: ` EndMatch` -> `MatchEnd`

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/MatchTest.scala
```diff
@@ -62,4 +62,17 @@ class MatchTest {
...
+
+    /* This match generates a value class boxing operation in the matchEnd (in
+     * 2.10 and 2.11).
+     */
+    val result =
+      "foo = " ++ (foo match { case Some(0) => "zero" case _ => "unknown" })
+
+    assertEquals("foo = unknown", result)
+  }
+
```
sjrd: Could you also add your minimized version that does not depend on the Scala lib?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2767,8 +2767,28 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          innerResultType, translatedCases, returnCount)
+
+      matchEndBody match {
+        case Ident(_) if matchEndParam.symbol == matchEndBody.symbol =>
+          // matchEnd is identity.
+          optimized
+
+        case Literal(Constant(())) =>
+          // Unit return type.
+          optimized
```
sjrd: These two cases probably call for AST tests that check we're not emitting `VarDef`s when it is not necessary.

sjrd:
That's all.

gzm0:
Updated. I have manually verified that the tests do not pass when:

- We directly return `optimized` (pos test fails)
- Remove either of the first two cases (ast tests fail)

sjrd:
LGTM. 

If someone can check the nightly build, they can merge.

gzm0:
Nightly has passed.

PR #2566, opened 2016-08-24
Fix #2553: Allow to create String from large char/codepoint array.

### LINKED ISSUE #2553: Constructing a String from a large byte array throws "Maximum call stack size exceeded"
When constructing a string from a large buffer, the Javascript runtime will fail for large byte arrays. (125,056 bytes in V8/Chrome, 500,000 bytes in Firefox).

The culprit appears to be in RuntimeString.scala :: newString (https://github.com/scala-js/scala-js/blob/7441caa9727285f81428a6afab04d524df848b2c/library/src/main/scala/scala/scalajs/runtime/RuntimeString.scala#L309

In particular, line `309` 

    fromCharCode(charCodes: _*)

This translates to Javascript as 

    jsx$2.fromCharCode.apply(jsx$2, charCodes);

In Summary - It's calling `fromCharCode` with each byte in the array being passed as a separate parameter to the function. So constructing a string with `Array[Byte](1,2,3,4,5)` is essentially calling `String.fromCharCode(1,2,3,4,5)`

Unfortunately, Javascript has a limit on the number of arguments that can be passed to a function (See https://mathiasbynens.be/demo/javascript-argument-count), putting a hard limit on the number of bytes that can be decoded.

Example Scala Code to replicate.

    val buffer = Array.fill[Byte](1000000)('x')
    val str = new String(buffer, "UTF-8")

sjrd:
We avoid calling `String.fromCharCode.apply` with a very large JavaScript array. Otherwise, some engines fail with "Maximum call stack size exceeded".

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/RuntimeString.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/RuntimeString.scala
@@ -300,13 +300,13 @@ private[runtime] object RuntimeString {
     if (offset < 0 || end < offset || end > value.length)
       throw new StringIndexOutOfBoundsException
 
-    val charCodes = new js.Array[Int]
+    var result = ""
     var i = offset
     while (i != end) {
-      charCodes += value(i).toInt
+      result += value(i).toString
       i += 1
     }
-    fromCharCode(charCodes: _*)
+    result
   }
 
   def newString(bytes: Array[Byte]): String =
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @nicolasstucki

sjrd:
I verified that both test cases did actually fail before the fix.

sjrd:
Ping @nicolasstucki

nicolasstucki:
LGTM

PR #2568, opened 2016-08-26
Improve js.JavaScriptException.getMessage().

sjrd:
Instead of overriding `toString()` to forward to the inner exception's `toString()` message, we do this in `getMessage()`. This improves the display of `JavaScriptException`s in tools that bypass `toString()`, such as JUnit's error reporter.


sjrd:
Review by @nicolasstucki

nicolasstucki:
LGTM

Issue #2175, opened 2016-01-16
Integrating with JS module systems


ochrons:
# Goal

Seamlessly integrate Scala.js applications with existing JavaScript module systems.
# Background
### JS module systems

Until quite recently the use of _modules_ in JavaScript was not the common. Instead, all libraries polluted the global namespace with their entrypoints. Nowadays there are several competing module systems including AMD, CommonJS and ES6's own module syntax. Traditionally AMD has been a favorite amongs frontend devs due to its dynamic nature, whereas Node.js users went with CommonJS (as it's natively supported by Node).

Latest in the game is the ES6 module syntax, but as it's the only language level module system, it's bound to become the standard in the future.
### Bundling

JS modules tend to be small and numerous, so it makes sense to bundle them together to create one downloadable JS file (just like what Scala.js does) instead of browser downloading potentially dozens of individual files. The bundling process also allows removal of unused exports, reducing the final code size.

Most common tools for bundling these days are [browserify](http://browserify.org/) and [webpack](https://webpack.github.io/), plus a few others like [JSPM](http://jspm.io/) and [RollUp](http://rollupjs.org/).

Due to bundling, the build process of a modern JS web app is quite close to what happens in Scala.js, including dependency/version management, compilation (TS/ES6 to ES5) and linking (bundling).
### Current Scala.js situation

Scala.js is totally oblivious to JS modules and related bundling activities. It operates in its own sandbox, producing a single JS file that assumes dependencies will be available at runtime through the global namespace. The build process supports simplistic bundling (concatenation) of JS libs, to create a single JS dependencies file to be included as part of the application.

Luckily most frontend JS libs today are available in the "ready-to-use" format that you can directly include in the website (or in Scala.js jsdeps), but recently more and more libs have been published only in the module format (for example [Material UI](https://github.com/callemall/material-ui)). On Node.js CommonJS is ubiquitous, so most libs publish only in the module format. To use these libraries in a Scala.js project, developers need to package them separately using the JS toolchain.
# Solutions
## Simple ad-hoc

A quick-and-dirty way of using JS modules in a Scala.js project is to pre-bundle them using the regular JS tools like `webpack`. All you need to do is to define a short JavaScript file serving as an entry point for bundling (and have the dependencies available in the npm style). For example to use React, you would define following:

``` js
window.React = require('react');
```

Running `webpack` produces a `bundle.js` which can be included as part of `jsDependencies` and it will expose React in the global `window` object so it can be used in the Scala.js application.

[...]

PR #2572, opened 2016-08-27
Add `@JSImport` and the generation of CommonJS modules.

sjrd:
This is the first and biggest step towards #2175. It contains all the semantic changes to the language and the IR. The only thing that is left is to actually emit ES2015 modules rather than CommonJS modules in the emitter.

This commit contains two major changes, which are inherently linked:

* An `@JSImport` annotation which can be used on top-level native JS classes and objects to signify that they should be imported from a JavaScript module.
* A backend configuration `ModuleKind`, which instructs the emitter to generate a CommonJS module, rather than a script meant to be executed at the top-level of a JS engine.

A codebase containing a reachable `@JSImport` cannot be successfully emitted to a top-level script, as there is no universal way of importing things.

We could imagine emitting Node.js modules without `@JSImport` support at first, but it would be a bit nonsensical to have modules that cannot import from other modules.

When emitting a module, top-level `@JSExport`s become the module exports, obviously.

Semantically, `@JSImport` and `@JSExport` are ES2015 module imports and exports. They are therefore static, which means that a) things cannot be conditionally imported or exported and b) their "key" must be a compile-time string constant.

The `ModuleKind` can be specified in sbt with the new key `scalaJSModuleKind`, which is scoped per project and per configuration, but not per stage. Indeed, they specify the interface with the outside JavaScript world, and that should not change in fastOpt versus fullOpt. Obviously, setting `scalaJSModuleKind` to `CommonJSModule` requires Node.js to be used for `run` and `test` tasks.

The historical fact that the `scalaJSLauncher` is stage-independent unfortunately makes `packageJSLauncher` incompatible with a `scalaJSModuleKind` different from `NoModule`. This is because, with modules, the contents of the launcher must reference the module ID, which is different in fastOpt and fullOpt.

In the future, we will extend `ModuleKind` with a new kind for actual ECMAScript 2015 modules.

---

Main things that could warrant discussion:

* The semantics of `@JSImport`
* The Tools API changes
* The way to *run* and *test* a Scala.js application that uses Node.js modules from sbt

# Semantics of `@JSImport`

See the Scaladoc of `@JSImport`. I'm personally happy with them as is, but we need to get this right, so it's good to have other people have a look at it.

# The Tools API change

`ModuleKind` comes along `Semantics` and `OutputMode` to be one of three mandatory "options" of the linker. It makes sense as a mandatory thing because it impacts the behavior of the Scala.js output. `LinkerFrontend.Config` and `LinkerBackend.Config` only contain superficial things, like source map emission, which don't have impact on the behavior.

# sbt integration for running and testing

I am not sure at all about the current design. Basically the idea is to alter the launcher script to `require` the produced module, which gives something like that for the hello world:
```js
require("/path/to/helloworld-fastopt.js")["helloworld"]["HelloWorld"]().main();
```
Note that we need the fully qualified path, and that it is specific to FastOpt. There needs to be a *different* launcher for fullOpt. The existing design of the sbt plugin does not accommodate well to this requirement. Hence, in the current implementation, `persistLauncher := true` is forbidden when `moduleKind != NoModule`. Moreover, Rhino does not support `moduleKind != NoModule` either, obviously.

The error reporting when misconfiguring those things in sbt might subpar ...


sjrd:
Opening for discussion, ping @gzm0, @nicolasstucki

library/src/main/scala/scala/scalajs/js/annotation/JSImport.scala
```diff
@@ -0,0 +1,44 @@
...
+
+/** Marks the annotated class or object as imported from another JS module.
+ *
+ *  Intuitively, this corresponds to the following ECMAScript import
+ *  directive:
+ *  {{{
+ *  import { <name> as AnnotatedClassOrObject } from <module>
+ *  }}}
+ *
+ *  To import the default import of a module, use `JSImport.Default` as `name`.
```
julienrf: “To import the default **export** of a module”?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/ModuleKind.scala
```diff
@@ -0,0 +1,34 @@
...
+   *  `__ScalaJSEnv.exportsNamespace`.
+   *
+   *  Imports are not supported.
+   */
+  case object NoModule extends ModuleKind
+
+  /** A Node.js module.
+   *
+   *  Imported modules are fetched with `require`. Exports go to the `exports`
+   *  module-global variable.
```
julienrf: “module-**local**”?

library/src/main/scala/scala/scalajs/js/annotation/JSImport.scala
sjrd: Yes, that's better indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/ModuleKind.scala
sjrd: "module-global" is Node.js terminology. Used for example here: https://nodejs.org/docs/latest/api/modules.html#modules_the_module_object

julienrf:
Looks promising!

If I understand correctly, now if I want to integrate a npm module in an application that targets the browser, I have to introduce a npm module bundling step in my build, right? Do you think we should rely on the sbt-web ecosystem to do so or implement our own sbt support for this?

sjrd:
> If I understand correctly, now if I want to integrate a npm module in an application that targets the browser, I have to introduce a npm module bundling step in my build, right?

That, or use a browser-compatible (dist) version of the library, and continue to use Scala.js in "browser mode".

> Do you think we should rely on the sbt-web ecosystem to do so or implement our own sbt support for this?

It was more or less decided in #2175 to leave that to tools other than Scala.js core, in particular/for example sbt-web, indeed.

julienrf:
I noticed that the incremental compiler does not work well with this: if I change the `scalaJsModuleKind` setting the project is not automatically re-compiled, I have to run `clean` to achieve that. Also, if I add an `@JSImport` annotation to a definition, the project is not automatically re-compiled.

sjrd:
> if I change the scalaJsModuleKind setting the project is not automatically re-compiled

That's expected, if unfortunate. We have the same problem with other configuration settings like `scalaJSSemantics`. We should eventually fix that.

> Also, if I add an @JSImport annotation to a definition, the project is not automatically re-compiled.

That is *not* expected. What exactly *does* get recompiled? I guess you still get "Compiling 1 Scala source ...". You should probably also get "Fast optimizing ...". But then the result hasn't changed? If that's true, we have a problem ^^

julienrf:
Sorry I was wrong: the project is indeed correctly recompiled, but the output is like if I didn’t change the `@JSImport` annotation:

- if I add the annotation, references to the object are still compiled to `$g["foo"]` (unless I `sbt clean`) ;
- if I remove the annotation, references to the object are still compiled to `$i_foo` (even though this `$i_foo` is not defined anymore in the generated code) ;
- however, if I change the module name (as a parameter of the annotation) the output is correctly updated (even without `clean`ing).

sjrd:
Ah OK. Yes, that's expected in the current WiP. I haven't properly incrementalized `JSDesugaring` wrt. `@JSImport` (or, as they're called at this point of the pipeline, `NativeJSLoadingSpec`s)

sjrd:
For the incremental behavior, this will have to be rebased on top of #2600 and adapted to it.

julienrf:
Another question that comes to my mind: is it possible to write code explicitly calling `require`, rather than having to write a `@JSImport`ed facade?

sjrd:
No, that will not be possible. It still goes against the ES spec, and we're only implementing ES spec (desugared to Node.js).

sjrd:
Review by @gzm0. The only thing I'm still missing is neg tests. Otherwise the PR should be complete.

sjrd:
PR complete :)

js-envs/src/main/scala/org/scalajs/jsenv/rhino/ScalaJSCoreLib.scala
```diff
@@ -20,6 +20,7 @@ import org.scalajs.core.tools.linker.{LinkedClass, LinkingUnit}
 import org.scalajs.core.tools.javascript._
 import org.scalajs.core.tools.io._
 
+import org.scalajs.core.tools.linker.backend.ModuleKind.NoModule
```
sjrd: Useless import.

test-interface/src/main/scala/org/scalajs/testinterface/internal/FrameworkLoader.scala
```diff
@@ -7,8 +7,10 @@ import sbt.testing.Framework
 private[internal] object FrameworkLoader {
 
   def loadFramework(frameworkName: String): Framework = {
+    val exportsNamespace =
+      scala.scalajs.runtime.environmentInfo.exportsNamespace
```
gzm0: So just to double check: This has been around since 0.6.0, right? So binary compatibility is in any case not an issue.

sjrd:
> We probably want to drop a line on the doc of JSImport, that it is not compatible with jsDependencies at all. commonJSName etc. will not work as expected (which is OK, because we have to break this nightmare at some point).

Ah, right.

> You might want to split the interface change in the Linker (adding the Config class) into another commit, just to make the last one a bit smaller.

Good point.

ir/src/main/scala/org/scalajs/core/ir/ClassKind.scala
```diff
@@ -35,38 +43,45 @@ sealed abstract class ClassKind {
...
+    case _                   => this
   }
 }
 
 object ClassKind {
   case object Class extends ClassKind
   case object ModuleClass extends ClassKind
   case object Interface extends ClassKind
-  case object RawJSType extends ClassKind
+  case object AbstractJSType extends ClassKind
```
sjrd: :)

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -869,6 +869,37 @@ object Trees {
...
+     *
+     *  // import { Bar as y } from 'foo'
+     *  // y.Baz
+     *  Import("foo", List("Bar", "Baz"))
+     *
+     *  // import * as x from 'foo' (whole module import)
+     *  Import("foo", Nil)
+     *
+     *  // import x from 'foo' (default import)
+     *  Import("foo", List("default"))
```
sjrd: The situation here is extremely unclear, wrt. interactions between ES2015 modules and "legacy" modules. There is at least [one well-regarded source](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) (section Default exports) saying that a legacy module would appear to ES6 as a module with a default import. Which would mean that your two lines above (CommonJS and ES2015) are actually equivalent if `my_lib.js` is a legacy module.

The [example for "external imports" from Rollup](http://rollupjs.org/?version=0.36.1&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyJTJGJTJGJTIwWW91JTIwY2FuJTIwaW1wb3J0JTIwZXh0ZXJuYWwlMjBtb2R1bGVzJTIwaW50byUyMHlvdXIlMjBidW5kbGUlMjAlRTIlODAlOTMlNUNuJTJGJTJGJTIwaXQlMjBkb2Vzbid0JTIwbWF0dGVyJTIwaWYlMjB0aGV5J3JlJTIwRVMyMDE1JTIwb3IlMjBsZWdhY3klNUNuaW1wb3J0JTIwJTI0JTIwZnJvbSUyMCdqcXVlcnknJTNCJTVDbiU1Q24lMjQoJTIwJ2JvZHknJTIwKS5odG1sKCUyMCclM0NoMSUzRUhlbGxvJTIwd29ybGQhJTNDJTJGaDElM0UnJTIwKSUzQiUyMiU3RCU1RCU3RA==) seems to somewhat corroborate this view, as they translate the following ES2015 module:
```js
// You can import external modules into your bundle –
// it doesn't matter if they're ES2015 or legacy
import $ from 'jquery';

$( 'body' ).html( '<h1>Hello world!</h1>' );
```
into the following, with a magical `_interopDefault` to fallback to the module itself if it doesn't have a `"default"` field:
```js
function _interopDefault (ex) {
  return (ex && (typeof ex === 'object') && 'default' in ex) ? ex['default'] : ex;
}
var $ = _interopDefault(require('jquery'));
$( 'body' ).html( '<h1>Hello world!</h1>' );
```
Note that none of this is *specified* per se. It all depends on the module Loader that orchestrates imports. Loaders can be different on different platforms (Node.js vs browser, for example), and for neither does there exist a spec at this point.

So I think it is fair to say, with the current understanding of ES modules, that the current specification for `Import` is likely going to do what one wants in practice. It certainly would with the current implementation of Rollup.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
```diff
@@ -73,4 +73,77 @@ final class Linker(frontend: LinkerFrontend, backend: LinkerBackend)
...
+      /** Whether to generate source maps. */
+      val sourceMap: Boolean,
+      /** Whether to use the Scala.js optimizer. */
+      val optimizer: Boolean,
+      /** Whether things that can be parallelized should be parallelized.
+       *  On the JavaScript platform, this does not have any effect.
+       */
+      val parallel: Boolean,
+      /** Whether to use the Google Closure Compiler pass.
+       *  On the JavaScript platform, this does not have any effect.
```
sjrd: We're not giving up type-safety per se. The config simply now says "use GCC *if it is available*". Similarly, `parallel` means "use parallelism *if possible*". (The doc should say that though.)

I think this new meaning is actually more useful than the previous approach, as it simplifies writing cross-platform code *using* the Linker API.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/ModuleKind.scala
sjrd: Actually, no. The rationale in my head was that, if you were using this, it was because you want to use your module with Node.js. But that's not necessarily the case. One could generate CommonJS modules to be consumed by a later bundler like Webpack. I guess I implicitly discarded that scenario as, for such a use case, one would typically generate an ES6 module. But there could be people who already have an established CommonJS-based bundling pipeline, and don't want to migrate to something that understand ES6 modules.

test-interface/src/main/scala/org/scalajs/testinterface/internal/FrameworkLoader.scala
sjrd: Even if it wasn't, it wouldn't be an issue. Linking the test-interface compiled by 0.6.13 will require a 0.6.13 linker anyway, which will provide you with an `exportsNamespace`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -533,33 +535,49 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
-            !isJSGlobalScope(implDef)) {
-          if ((enclosingOwner is OwnerKind.ScalaMod) &&
-              !sym.owner.isPackageObjectClass) {
-            if (sym.isModuleClass) {
-              reporter.error(implDef.pos, "Native JS objects inside " +
-                  "non-native objects must have an @JSName annotation")
+          val loadSpec = {
+            if (enclosingOwner is OwnerKind.JSNativeMod) {
+              val ownerLoadSpec = jsInterop.jsNativeLoadSpecOf(sym.owner)
+              val jsName = jsInterop.jsNameOf(sym)
```
gzm0: Ah, good to know :) Maybe we should prohibit dots in these names then? (via warning and prohibition in 1.0)

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -645,6 +711,17 @@ class JSInteropTest extends DirectTest with TestHelpers {
 
     """
     object A {
+      @JSImport("InnerB")
+      @js.native
```
gzm0: I agree.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Yes, we'd need to do this. Otherwise we won't have the duality as in the article:

```javascript
// ES6 equivalent of `var colors = require("colors/safe");`
import colors from "colors/safe";
```

What I do not like about this, is that it relies on magic names. There will be **no** way to fully import a CommonJS module that has a member named `default`. So a later addition to the library might even silently break existing code...

The alternative is of course, that in Scala.js we do make a distinction between the default import and importing `default`. But that means more special casing and a terrible semantics based on legacy, once CommonJS modules are not around anymore.

OTOH it might be a good idea to have `@JSImport` without a second argument be the default import. I guess this mostly depends how modules end up exporting their stuff.
To be safe, we maybe should not allow a single argument for now, so we can define it's semantics later?

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: Maybe we should, but that's a job for a separate PR, IMO.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
sjrd: What about this:

1. Rename `closureCompiler` (resp. `withClosureCompiler`) to `closureCompilerIfAvailable` (resp. `withClosureCompilerIfAvailable`).
2. Add an implicit pimp only in `jvm/`:

```scala
implicit class ConfigPlatformOps(val self: Config) {
  def closureCompiler: Boolean =
    self.closureCompilerIfAvailable
  def withClosureCompiler(closureCompiler: Boolean): Config =
    self.withClosureCompilerIfAvailable(closureCompiler)
}
```
and possibly a read-only pimp in `js/`:
```scala
implicit class ConfigPlatformOps(val self: Config) {
  def closureCompiler: Boolean = false
}
```
This should give all of:
1. No code duplication
2. Default, short name is fully typechecked
3. There is longer, very explicit name for the ability to write cross-platform code.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: > Also, it is actually quite unlikely for any legacy JS module to export a member named default, as it is a keyword in JS.

Good point. So let's emit the `_interopDefault` fallback.

> What syntax would you suggest for the * import, in that case?

For example

```scala
@JSImport("module.js", All)
```

where `All` is an object.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
gzm0: Yeah, that works for me.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
gzm0: Filed as #2615

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
sjrd: Relevant information: how Node.js plans to deal with ES importing legacy: https://github.com/nodejs/node-eps/blob/master/002-es6-modules.md#54-es-consuming-commonjs

> `module.exports` is a single value. As such it does not have the dictionary like properties of ES module exports. In order to facilitate named imports for ES modules, all properties of `module.exports` will be hoisted to named exports after evaluation of CJS modules with the exception of `default` which will point to `module.exports` directly.

That interpretation *does* hide a property actually called `default`, as the first example in that section shows. This is not the same interpretation as Rollup. Note that this interpretation is virtually impossible to polyfill properly (when importing `default`, we don't know whether it should be the module itself (if it's legacy) or its default export (if it's ES)).

Irrelevant information: the current state of the Browser Loader spec does not seem to mention `default` anywhere: https://whatwg.github.io/loader/
To be fair, though, it *seems* that it does not *allow* to import non-ES modules at all, which bypasses the issue.

sjrd: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
sjrd: Done.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -578,6 +608,12 @@ abstract class PrepJSInterop extends plugins.PluginComponent
           // Mark module classes as having the new format
           if (sym.isModuleClass)
             sym.addAnnotation(HasJSNativeLoadSpecAnnotation)
+        } else {
+          assert(sym.isTrait) // just tested in the previous `if`
+          if (sym.hasAnnotation(JSImportAnnotation)) {
+            reporter.error(implDef.pos,
+                "Traits may not have an @JSImport annotation")
+          }
```
gzm0: Hmmm... Probably we should also emit a warning for `@JSName` as it does not do anything. Filed as #2617.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -918,6 +960,19 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
 
+  /** Checks that arguments to `@JSImport` on [[sym]] are literals.
+   *  Reports an error on each annotation where this is not the case.
+   */
+  private def checkJSImportLiteral(sym: Symbol): Unit = {
+    for {
+      annot <- sym.getAnnotation(JSImportAnnotation)
+      if (0 until annot.args.size).exists(i => annot.stringArg(i).isEmpty)
+    } {
+      reporter.error(annot.pos,
```
gzm0: Could you report the position of the offending argument?

project/Build.scala
```diff
@@ -1349,6 +1349,17 @@ object Build {
       testHtmlSettings(testHtmlFullOpt, FullOptStage) ++ Seq(
         name := "Scala.js test suite",
 
+        unmanagedSourceDirectories in Test ++= {
+          def includeIf(testDir: File, condition: Boolean): List[File] =
+            if (condition) List(testDir)
+            else Nil
```
gzm0: Do you repeat this on purpose? Maybe lift it so you don't have to define it twice?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -609,7 +629,8 @@ object ScalaJSPluginInternal {
         val libs =
           resolvedJSDependencies.value.data ++ scalaJSConfigurationLibs.value
         resolvedJSEnv.value match {
-          case env: LinkingUnitJSEnv =>
+          case env: LinkingUnitJSEnv
+              if scalaJSModuleKind.value == ModuleKind.NoModule =>
```
gzm0: You do not need this guard. `linker.linkUnit` is independent of the `ModuleKind`. It might be worth documenting this on `Linker`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -412,20 +418,34 @@ object ScalaJSPluginInternal {
...
+          Def.task {
+            Attributed.blank((artifactPath in packageScalaJSLauncher).value)
+          }
+        } else {
+          Def.task {
+            mainClass.value map { mainCl =>
+              val file = (artifactPath in packageScalaJSLauncher).value
+              val moduleKind = scalaJSModuleKind.value
+              IO.write(file,
+                  launcherContent(mainCl, moduleKind, None),
```
gzm0: Can you pass in `moduleKind.NoModule` (or assert it) to make it explicit that this does not work with modules?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -676,15 +731,22 @@ object ScalaJSPluginInternal {
...
+            (mainClass in scalaJSLauncher).value map { mainClass =>
+              val moduleKind = scalaJSModuleKind.value
+              val moduleIdentifier = scalaJSModuleIdentifier.value
+              val memLaunch =
+                memLauncher(mainClass, moduleKind, moduleIdentifier)
+              Attributed[VirtualJSFile](memLaunch)(
+                  AttributeMap.empty.put(name.key, mainClass))
+            } getOrElse {
+              sys.error("No main class detected.")
+            }
```
gzm0: While you are touching this, use a `fold`?

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSFramework.scala
```diff
@@ -61,10 +71,26 @@ final class ScalaJSFramework(
...
     val name = jsonToString(frameworkName.toJSON)
     val code = s"""
-      new org.scalajs.testinterface.internal.InfoSender($name).initAndSend();
+      new ${prefix}org.scalajs.testinterface.internal.InfoSender($name).initAndSend();
     """
     new MemVirtualJSFile(s"testFrameworkInfo.js").withContent(code)
   }
+
+  private[testadapter] def optionalExportsNamespacePrefix: String = {
+    // !!! DUPLICATE code with ScalaJSPlugin.makeExportsNamespaceExpr
```
gzm0: We'll probably have to do better than this, since this is not likely to go away anytime soon. Also, there is not really a big reason you cannot put this into `tools.javascript` or something, is there?
I'm fine if you leave this as it is for now and add a cleanup issue.

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -0,0 +1,113 @@
...
+    val decoder = new StringDecoder()
+    assertEquals("こんにちは", decoder.write(b))
+    assertEquals("", decoder.end())
+  }
+
+}
+
+object ModulesTest {
+  @js.native
+  @JSImport("querystring")
```
gzm0: .js is optional?

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -0,0 +1,113 @@
...
+\*                                                                      */
+package org.scalajs.testsuite.jsinterop
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+import org.junit.Assert._
+import org.junit.Test
+
+class ModulesTest {
```
gzm0: Can you please add a comment stating that this relies on Node.js modules, because we currently have no reasonable way of loading non-built in modules and that it should be made VM independent once we have this.

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -0,0 +1,113 @@
...
+    assertFalse(Buffer.isBuffer(5))
+  }
+
+  @Test def testImportClassInModule(): Unit = {
+    val b = new Buffer(5)
+    for (i <- 0 until 5)
+      b(i) = (i * i).toShort
+
+    for (i <- 0 until 5)
+      assertEquals(i * i, b(i).toInt)
```
gzm0: Check `isBuffer`?

tools/js/src/main/scala/org/scalajs/core/tools/linker/LinkerPlatformExtensions.scala
```diff
@@ -27,12 +27,18 @@ trait LinkerPlatformExtensions { this: Linker.type =>
...
 
     new Linker(frontend, backend)
   }
 
+  @deprecated("Use the overload with an explicit ModuleKind.", "0.6.13")
+  def apply(semantics: Semantics, outputMode: OutputMode,
+      config: Config): Linker = {
+    apply(semantics, outputMode, ModuleKind.NoModule, config)
+  }
+
```
gzm0: I think it is OK to remove this one :)

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/LinkerPlatformExtensions.scala
```diff
@@ -33,17 +33,23 @@ trait LinkerPlatformExtensions { this: Linker.type =>
...
 
     new Linker(frontend, backend)
   }
 
+  @deprecated("Use the overload with an explicit ModuleKind.", "0.6.13")
+  def apply(semantics: Semantics, outputMode: OutputMode,
+      config: Config): Linker = {
+    apply(semantics, outputMode, ModuleKind.NoModule, config)
+  }
+
```
gzm0: Same here :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -66,32 +79,72 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+      emitModuleImports(orderedClasses, builder)
+
       for (classInfo <- orderedClasses)
         emitLinkedClass(classInfo, builder)
     } finally {
       endRun(logger)
     }
   }
 
+  private def emitModuleImports(orderedClasses: List[LinkedClass],
```
gzm0: I'm having the feeling that this more belongs into the `ClassEmitter`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -66,32 +79,72 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
-        builder.addLine("}).call(this);")
+    moduleKind match {
+      case ModuleKind.NoModule =>
+        outputMode match {
+          case OutputMode.ECMAScript51Global =>
+          case OutputMode.ECMAScript51Isolated | OutputMode.ECMAScript6 =>
+            builder.addLine("}).call(this);")
+        }
+
+      case ModuleKind.CommonJSModule =>
```
gzm0: Consider moving the logic of whether you need an isolation guard to a separate method.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2279,14 +2279,74 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
       case JSNativeLoadSpec.Global(path) =>
-        path.foldLeft(envField("g")) {
-          (prev, part) => genBracketSelect(prev, js.StringLiteral(part))
+        pathSelection(envField("g"), path)
+
+      case JSNativeLoadSpec.Import(module, path) =>
+        val moduleValue = envModuleField(module)
+        path match {
+          case DefaultExportName :: rest =>
+            val defaultField = genCallHelper("moduleDefault", moduleValue)
```
gzm0: This is a bit bothersome, but frankly I also don't know how to do this better...

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -961,15 +961,33 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
       annot <- sym.getAnnotation(JSImportAnnotation)
-      if (0 until annot.args.size).exists(i => annot.stringArg(i).isEmpty)
     } {
-      reporter.error(annot.pos,
-        "The arguments to JSImport must be literal strings")
+      val firstArgIsValid = annot.stringArg(0).isDefined
+      val secondArgIsValid = {
+        annot.stringArg(1).isDefined ||
+        (annot.argAtIndex(1) match {
+          case Some(tree @ (_:Ident | _:Select))
```
gzm0: Do you really want to care about the tree type? Consider:

```scala
annot.argAtIndex(1).exists(_.symbol == JSImportNamespaceObject)
```

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -961,15 +961,33 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+            true
+          case _ =>
+            false
+        })
+      }
+
+      if (!firstArgIsValid || !secondArgIsValid) {
+        reporter.error(annot.pos,
+            "The arguments to JSImport must be literal strings or the " +
+            "JSImport.Namespace object.")
```
gzm0: Ah :) My previous comment about position still applies :)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -961,15 +961,33 @@ abstract class PrepJSInterop extends plugins.PluginComponent
   }
 
   /** Checks that arguments to `@JSImport` on [[sym]] are literals.
+   *
+   *  The second argument can also be the singleton `JSImport.Namespace`
+   *  object.
+   *
```
gzm0: I like this definition much better for now. Better play it safe :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -58,7 +58,8 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
       implicit val ctx = ErrorContext(classDef)
       checkStaticMembers(classDef)
       classDef.kind match {
-        case ClassKind.RawJSType =>
+        case ClassKind.AbstractJSType | ClassKind.NativeJSClass |
+            ClassKind.NativeJSModuleClass =>
```
gzm0: Don't forget this one :)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: That's actually precisely the issue. If I don't guard here, `env.loadLibs(libs).loadLinkingUnit(unit)` will have no idea that it is supposed to emit `unit` as a `ModuleKind`. So it will ignore it and desugar with `NoModule` anyway. The deeper problem is that the `LinkingUnitJSEnv` API was not designed to support `ModuleKind`s. I don't think it's worth evolving it to support them, though, since the only real use case is Rhino, and Rhino will disappear.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSFramework.scala
sjrd: It's the same argument I once gave about I don't know which duplication. The problem is that there is no private common place to put this; and at the same time it shouldn't be public. I don't want to pollute the public API of `tools.javascript` (or `tools`, for that matter) with a relatively ugly implementation detail like this, which is not well specifiable on its own.

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
sjrd: That would make `testImportClassInModule` dependent on what is supposed to be tested by `testImportObjectInModule`.

However I could definitely add that in `testImportIntegrated`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: I don't think so. `ClassEmitter` is about class-level things. This is linking-unit-level. So IMO it is similar to emitting the prelude/postlude (including the CoreJSLib).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: It's done in `checkJSNativeLoadSpec`

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSFramework.scala
gzm0: Ok. Let's leave it like that for now.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
gzm0: Ok.

sjrd:
I should have addressed all the comments.

PR #2574, opened 2016-08-29
Fix #2573: Fix `String.split()` yet again (leading 0-width match).

### LINKED ISSUE #2573: split("") is not identical to JVM split
This assert fails in JS but passes in JVM:

```scala
val line = "123"
val splits = line.split("").toList
assert (splits == List("1","2","3")) //JS returns List("", "1", "2", "3")
```

I think this ignores the rule from the javadoc that "A zero-width match at the beginning never produces an empty leading substring". 

Similarly `line.split("a?").toList` has an empty leading substring.

sjrd:
This brings `String.split()` up to date with the behavior of JDK 8. If there is a zero-width match of the separator at the beginning of the string, it is ignored, i.e., the empty string that precedes it is not included in the result array.


sjrd:
Review by @gzm0, since you have played with `String.split()` a number of times before :p

gzm0:
LGTM

PR #2575, opened 2016-08-30
Update the compiler back-end for specialized `synchronized`.

sjrd:
Starting from 2.12.0-RC1, `AnyRef.synchronized` is specialized by erasure in the same way as `isInstanceOf` and `asInstanceOf`, so that it retains its type parameter until `cleanup` (and hence, until our back-end).

This commit updates the back-end to accommodate both the old generic `synchronized` (from `genPrimitiveOp`) and the new specialized one (from `genApplyTypeApply`).

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2004,7 +2004,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       fun match {
         case TypeApply(_, _) =>
-          genApplyTypeApply(tree)
+          genApplyTypeApply(tree, isStat)
 
         case _ if isRawJSDefaultParam =>
           js.UndefinedParam()(toIRType(sym.tpe.resultType))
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @adriaanm 
I tested this locally with `++2.12.0-328b50c-SNAPSHOT`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3224,11 +3235,11 @@ abstract class GenJSCode extends plugins.PluginComponent
     }
 
     /** Gen JS code for a call to AnyRef.synchronized */
-    private def genSynchronized(tree: Apply, isStat: Boolean): js.Tree = {
+    private def genSynchronized(tree: Tree, receiver: Tree, arg: Tree,
+        isStat: Boolean): js.Tree = {
```
gzm0: Consider just passing in the position instead of `tree`. Passing in the tree and its destructured form just to get the position seems a bit strange.

sjrd: It's a common pattern in `GenJSCode`, though. We typically pass trees for context rather than positions, even when use the trees only for their `pos`.

sjrd: Fixed.

adriaanm:
LGTM. Thank you!

PR #2584, opened 2016-09-07
Fix #2579: Deprecate RhinoJSEnv, use NodeJSEnv by default instead

### LINKED ISSUE #2579: Make scalaJSUseRhino := false by default
The only reason we have `scalaJSUseRhino := true` by default is the out-of-the-box experience. We've always wanted to say: hey, you don't *need* to install anything but sbt, and Scala.js just works.

But, in practice, no one keeps using Rhino. Everyone installs Node.js and rolls with it. So the "default" build includes `scalaJSUseRhino in Global := false` anyway. Besides, sometimes beginners are baffled by how slow Scala.js is, and at some point veterans realize they're still using Rhino, and tell them not to.

I believe it would do everyone a favor to disable Rhino by default. Yes, our tutorial should then say that users need to install Node.js by default, but that is what every reasonable Scala.js user does anyway.

Note: this only affects users who actually use the `run` and/or `test` tasks from sbt.

sjrd:
~~Second commit to come to actually deprecate `RhinoJSEnv`. CI for now.~~ Done

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -15,11 +15,9 @@
 
   <task id="main"><![CDATA[
     setJavaVersion $java
+    sbtretry ++$scala 'set scalaJSUseRhino in Global := true' helloworld/run &&
     sbtretry ++$scala helloworld/run &&
-    sbtretry 'set scalaJSUseRhino in Global := false' \
-        ++$scala helloworld/run &&
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
-        'set scalaJSUseRhino in Global := false' \
         ++$scala helloworld/run \
         helloworld/clean &&
     sbtretry 'set inScope(ThisScope in helloworld)(jsEnv := JSDOMNodeJSEnv().value)' \
@@ -29,9 +27,9 @@
         ++$scala helloworld/run \
         helloworld/clean &&
     sbtretry 'set scalaJSOptimizerOptions in helloworld ~= (_.withDisableOptimizer(true))' \
+        'set scalaJSUseRhino in Global := true' \
         ++$scala helloworld/run &&
     sbtretry 'set scalaJSOptimizerOptions in helloworld ~= (_.withDisableOptimizer(true))' \
-        'set scalaJSUseRhino in Global := false' \
         ++$scala helloworld/run \
         helloworld/clean &&
     sbtretry 'set scalaJSSemantics in helloworld ~= (_.withAsInstanceOfs(org.scalajs.core.tools.sem.CheckedBehavior.Unchecked))' \
@@ -42,11 +40,10 @@
     sbtretry 'set inScope(ThisScope in helloworld)(jsEnv := new org.scalajs.jsenv.RetryingComJSEnv(PhantomJSEnv().value))' \
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala helloworld/run &&
-    sbtretry ++$scala testingExample/test:run testingExample/test &&
-    sbtretry 'set scalaJSUseRhino in Global := false' \
+    sbtretry 'set scalaJSUseRhino in Global := true' \
         ++$scala testingExample/test:run testingExample/test &&
-    sbtretry 'set scalaJSUseRhino in Global := false' \
-        'set scalaJSStage in Global := FullOptStage' \
+    sbtretry ++$scala testingExample/test:run testingExample/test &&
+    sbtretry 'set scalaJSStage in Global := FullOptStage' \
         ++$scala testingExample/test:run testingExample/test \
         testingExample/clean &&
     sbtretry 'set inScope(ThisScope in testingExample)(jsEnv := new org.scalajs.jsenv.RetryingComJSEnv(PhantomJSEnv().value))' \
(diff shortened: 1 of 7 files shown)


sjrd:
The second commit is there.

Review by @gzm0, please.

ci/matrix.xml
```diff
@@ -134,21 +124,17 @@
...
-        'set scalaJSUseRhino in Global := false' \
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala $testSuite/test
   ]]></task>
 
   <task id="test-suite-ecma-script6"><![CDATA[
     setJavaVersion $java
-    sbtretry ++$scala jUnitTestOutputsJVM/test jUnitTestOutputsJS/test \
-        'set scalaJSUseRhino in Global := false' jUnitTestOutputsJS/test \
-        'set scalaJSStage in Global := FullOptStage' jUnitTestOutputsJS/test &&
```
gzm0: You remove this, since it is already in the ES5 task, right?

ci/matrix.xml
```diff
@@ -257,14 +241,13 @@
     sbt noDOM/run noDOM/testHtmlFastOpt noDOM/testHtmlFullOpt \
         withDOM/run withDOM/testHtmlFastOpt withDOM/testHtmlFullOpt \
         multiTestJS/test:run multiTestJS/testHtmlFastOpt multiTestJS/testHtmlFullOpt \
```
gzm0: These now change how they run IIUC. Intentional?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -188,8 +217,17 @@ object ScalaJSPlugin extends AutoPlugin {
...
 
-    val scalaJSUseRhino = SettingKey[Boolean]("scalaJSUseRhino",
-        "Whether Rhino should be used", APlusSetting)
+    /** Non-deprecated alias of `scalaJSUseRhino` for internal use. */
+    private[sbtplugin] val scalaJSUseRhinoInternal = SettingKey[Boolean](
+        "scalaJSUseRhino", "Whether Rhino should be used", KeyRanks.Invisible)
+
+    @deprecated(
+        "Rhino is not used by default anymore. " +
+        "Setting `scalaJSUseRhino` to `false` is therefore redundant. " +
```
gzm0: Maybe say that it will be removed?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -61,6 +62,34 @@ object ScalaJSPlugin extends AutoPlugin {
...
+    def RhinoJSEnv(): Def.Initialize[Task[RhinoJSEnv]] = Def.task {
+      /* We take the semantics from the linker, since they depend on the stage.
+       * This way we are sure we agree on the semantics with the linker.
+       */
+      import ScalaJSPluginInternal.{scalaJSLinker, scalaJSRequestsDOM}
+      val semantics = scalaJSLinker.value.semantics
+      val withDOM = scalaJSRequestsDOM.value
+      new RhinoJSEnv(semantics, withDOM)
+    }
+
```
gzm0: I do not understand why we are providing all this, if we are fully removing Rhino anyways. Until 1.0.0, people should be able to use `scalaJSUseRhino := true`. Afterwards this thing will be removed as well, right?

In any case, you'd have to deprecate this immediately...

ci/matrix.xml
sjrd: Yes. It probably shouldn't have been in the ES6 task to begin with.

ci/matrix.xml
sjrd: Yes, it is intentional. We test the default a bit more heavily than the non-default, here.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: Good point.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: It does become deprecated in the second commit, indeed.

The initial idea of this method was to decompose the deprecations into:
1. We deprecated `scalaJSUseRhino`, but there is still an easy, non-deprecated way to enable it.
2. Then, we really deprecate `RhinoJSEnv` itself.

In case people scream about the deprecation of Rhino, we can easily revert the second commit. `scalaJSUseRhino` would still be deprecated.

Does that make sense?

gzm0: Yeah. That makes perfect sense. I didn't realize you did deprecate it in the second commit.

sjrd:
Updated.

gzm0:
LGTM

sjrd:
We need to wait for https://github.com/scala/community-builds/pull/276 to be resolved before merging this PR.

sjrd:
I have added a commit that replaces Rhino by Node.js in the partest runner.
I locally tested `partestSuite/testOnly -- --noOpt`.

partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
```diff
@@ -74,19 +73,9 @@ class MainGenericRunner {
...
-        val output = WritableMemVirtualJSFile("partest-fastOpt.js")
-        linker.link(ir, output, logger)
-        new NodeJSEnv().loadLibs(ResolvedJSDependency.minimal(output) :: Nil)
-      }
+      val output = WritableMemVirtualJSFile("partest-fastopt.js")
+      linker.link(ir, output, logger)
+      new NodeJSEnv().loadLibs(ResolvedJSDependency.minimal(output) :: Nil)
     }
 
     libJSEnv.jsRunner(jsRunner).run(logger, jsConsole)
```
gzm0: Replace this with:

```scala
val sjsCode = {
  val output = WritableMemVirtualJSFile("partest.js")
  linker.link(ir, output, logger)
  ResolvedJSDependency.minimal(output) :: Nil
}

new NodeJSEnv().jsRunner(sjsCode, jsRunner).run(logger, jsConsole)
```

Seems simpler to understand to not split loading of lib and runner code... Now that I realize that it is actually not less code, your call I guess.

sjrd: Makes sense. It does seem a bit more readable, indeed.

sjrd:
Updated.

gzm0:
LGTM

sjrd:
https://github.com/scala/community-builds/pull/276 was merged :)

PR #2588, opened 2016-09-09
Fix #2587: Fix integer multiplication truncation error.

### LINKED ISSUE #2587: BigDecimal is not distributive
It seems like there's a bug in the `BigDecimal` implementation, possibly related to the fact that `Int.MinValue` equals `-Int.MinValue`:

``` scala
object TestCase extends scala.scalajs.js.JSApp {
  def main(): Unit = {
    val x = BigDecimal(Int.MinValue)
    println(x * (x + x))
    println(x * x + x * x)
  }
}
```

Here's the output:

> -9223372036854775808
> 9223372036854775808

(The correct answer is `9223372036854775808`.)

This is using Scala 2.11.8 and Scala.js 0.6.11.

non:
The Int.MinValue \* Int.MinValue case was a situation where both values
fit in 32-bits, but the result of the multiplication cannot fit into
64-bits.

Fixes #2587

### DIFF
--- a/javalib/src/main/scala/java/math/BigDecimal.scala
+++ b/javalib/src/main/scala/java/math/BigDecimal.scala
@@ -672,13 +672,27 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
 
   def multiply(multiplicand: BigDecimal): BigDecimal = {
     val newScale = this._scale.toLong + multiplicand._scale
-    if (this.isZero || multiplicand.isZero) {
-      zeroScaledBy(newScale)
-    } else if (this._bitLength + multiplicand._bitLength < 64) {
-      valueOf(this._smallValue * multiplicand._smallValue, safeLongToInt(newScale))
-    } else {
+
+    @inline
+    def bigMultiply: BigDecimal = {
       val unscaled = this.getUnscaledValue.multiply(multiplicand.getUnscaledValue)
       new BigDecimal(unscaled, safeLongToInt(newScale))
+    }
+
+    import java.lang.{Long => JLong}
+    if (this.isZero || multiplicand.isZero) {
+      zeroScaledBy(newScale)
+    } else if (this._bitLength + multiplicand._bitLength >= 64) {
+      bigMultiply
+    } else {
+      val x = this._smallValue
+      val y = multiplicand._smallValue
+      val z = x * y
+      if (JLong.signum(z) == JLong.signum(x) * JLong.signum(y)) {
+        valueOf(z, safeLongToInt(newScale))
+      } else {
+        bigMultiply
+      }
     }
   }
 
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/math/BigDecimalArithmeticTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/math/BigDecimalArithmeticTest.scala
@@ -1405,4 +1405,11 @@ class  BigDecimalArithmeticTest {
     assertEquals(result.toString, c)
     assertEquals(result.scale(), cScale)
   }
+
+  // https://github.com/scala-js/scala-js/issues/2587
+  @Test def testMultiplicationOverflow(): Unit = {
+    val x = new BigDecimal(Int.MinValue)
+    val y = new BigDecimal("9223372036854775808")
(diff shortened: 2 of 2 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/math/BigDecimalArithmeticTest.scala
```diff
@@ -1405,4 +1405,11 @@ class  BigDecimalArithmeticTest {
     assertEquals(result.toString, c)
     assertEquals(result.scale(), cScale)
   }
+
+  // https://github.com/scala-js/scala-js/issues/2587
+  @Test def testZeroPow0(): Unit = {
```
sjrd: The method name does not seem to fit. It seems you copy-pasted the method just above but forgot to change the name ;)

non: Whoops! Sorry. You're right, I just added the test quickly. I'll fix it.

sjrd:
Could you also alter the commit message so that first line reads as

    Fix #2587: Fix integer multiplication truncation error.

(and remove the `Fixes #2587` at the bottom, which is then redundant). This is the format we use in this repo.

non:
@sjrd Sure! Should I rebase and fold the test case fix in too?

sjrd:
Yes, please rebase to squash everything in one commit.

non:
I wasn't able to get the tests running locally (I was in a coffee shop when I wrote it), so I may have to push additional corrections if Travis fails. I'll be sure to squash/rebase to keep it to a  single commit.

sjrd:

non:
There was an error in the test -- I just rebased and force-pushed to fix it.

sjrd:
Jenkins, retest this please

non:
Apologies for the failures. I'm back home, have successfully run the test locally, and have high confidence it will work if run from Jenkins.

javalib/src/main/scala/java/math/BigDecimal.scala
```diff
@@ -672,13 +672,27 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
...
-      valueOf(this._smallValue * multiplicand._smallValue, safeLongToInt(newScale))
+    } else if (this._bitLength + multiplicand._bitLength >= 64) {
+      bigMultiply
     } else {
-      val unscaled = this.getUnscaledValue.multiply(multiplicand.getUnscaledValue)
-      new BigDecimal(unscaled, safeLongToInt(newScale))
+      val x = this._smallValue
+      val y = multiplicand._smallValue
+      val z = x * y
+      if (JLong.signum(z) == JLong.signum(x) * JLong.signum(y)) {
```
sjrd: Is there any other case than `MinValue * MinValue` here?

If not, then explicitly testing `x != MinValue || y != MinValue` would make that clearer, and be faster (long multiplication is mildly expensive in Scala.js)

If yes, could you give another example that needs this test like this? Why does this test always work for other problematic situations? (unlike for addition, in 2's complement multiplication checking the sign of the result is not enough to detect overflow in general).

non: So my original description was wrong -- it really has to be something like `MinValue * MinValue * 2` to trigger the bug, which means that there are at least two cases to check instead of one.

If you would prefer to do it that way I can go ahead and save a `Long` constant and then do two tests to catch it. I think doing the `Long` multiplication rather than those two tests ends up being a bit more efficient, but either way works for me.

sjrd: OK, but then my concern is that, as I said above, in general checking the signs is not enough to guarantee that no overflow occurs in a 2's complement multiplication (it is possible to multiply 2 negative numbers and receive a *negative* overflowed result). Why is it OK to do it like this in this instance? In other words, will any overflow happening when computing `z` be caught by this test, and why?

non: I'll write up a proof -- I think the short answer is that checking the bit lengths guarantees you that the values aren't big enough to wrap back around. But I'll try to create a more formalized explanation for that.

sjrd:
Superseded by #2674.

PR #2589, opened 2016-09-12
Widen Settings[_] to SettingsDefinition in CrossProject.[js, jvm]Settings()

MasseGuillaume:
reference: https://github.com/sbt/sbt/blob/1.0.x/main/src/main/scala/sbt/Project.scala#L194

this allows to write

```scala
val defaultSettings = Seq(scalaVersion := "2.11.8")

lazy val app = crossProject.settings(defaultSettings)

// or
lazy val app = crossProject.jsSettings(defaultSettings)

// or
lazy val app = crossProject.jvmSettings(defaultSettings)
```

it's backward compatible with

```scala
lazy val app = crossProject.settings(defaultSettings: _*)

// or
lazy val app = crossProject.jsSettings(defaultSettings: _*)

// or
lazy val app = crossProject.jvmSettings(defaultSettings: _*)
```

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/cross/CrossProject.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/cross/CrossProject.scala
@@ -225,11 +225,11 @@ final class CrossProject private (
     copy(js = transformer(js))
 
   /** Add settings specific to the underlying JVM project */
-  def jvmSettings(ss: Def.Setting[_]*): CrossProject =
+  def jvmSettings(ss: Def.SettingsDefinition*): CrossProject =
     jvmConfigure(_.settings(ss: _*))
 
   /** Add settings specific to the underlying JS project */
-  def jsSettings(ss: Def.Setting[_]*): CrossProject =
+  def jsSettings(ss: Def.SettingsDefinition*): CrossProject =
     jsConfigure(_.settings(ss: _*))
 
   // Concrete alteration members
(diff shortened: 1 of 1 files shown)


sjrd:
This won't be backwards *binary* compatible, though (unless `Setting[_]` is a subtype of `SettingsDefinition`?). There should either be a hack to maintain binary compatibility, or a very compelling reason to change (in this case, I don't think it's very compelling because we can always just use `: _*`)

gzm0:
I think we can accept a binary incompatibility in the sbt plugin only, if it aligns how things are done in sbt (how did they fix the binary issue?). However, if this is not how the newest version of sbt works (yet), we probably should not merge it.

sjrd:
http://www.scala-sbt.org/0.13/api/#sbt.Init$Setting

> `sealed class Setting[T] extends SettingsDefinition`

It seems `Setting[_]` is a subtype of `SettingsDefinition` after all. That would be how they made it binary compatible.

The change comes from the commit https://github.com/sbt/sbt/commit/a058a38b7b7a4b3959be085b3718ff7afc58bbb3
It's supposed to be part of sbt 0.13.8, but [the latest Scaladoc for 0.13.x](http://www.scala-sbt.org/0.13/api/#sbt.Project) still says `def settings(ss: Def.Setting[_]*): Project)`. Perhaps they haven't updated the Scaladoc on the website?

sjrd:
@MasseGuillaume I suggest you elaborate a bit on the commit message. Also the first line should be phrased in imperative form, as an action performed on the codebase. What it allows should be in the message body. Something like this:

    Widen Settings[_] to SettingsDefinition in CrossProject.xyzSettings().

    This change aligns with `Project.settings()` in sbt 0.13.8+. It allows
    to call `.settings(someSeqOfSettings)` in addition to
    `.settings(someSeqOfSettings: _*)`.

MasseGuillaume:
@sjrd updated commit message

sjrd:

PR #2590, opened 2016-09-13
Fix #2582: Upgrade to Scala 2.12.0-RC1.

### LINKED ISSUE #2582: Upgrade to Scala 2.12.0-RC1

sjrd:
In the process, we get rid of scala-2.12-junit-mixin-plugin, since its functionality has been taken over by scalac itself.

Also, we need to avoid using `Either.right`, because it's been deprecated in 2.12.0-RC1.


sjrd:
Review by @nicolasstucki 

Also ping @gzm0, maybe you have a better idea for replacing `Either.right`?

gzm0:
> maybe you have a better idea for replacing `Either.right`?

The only idea I have is add a compat object that we pull in depending on the scala version. So we'd need scala version specific source folders and do:

scala 2.12
```scala
object Compat {}
```

scala 2.11 / 2.10
```scala
object Compat {
  implicit class RightBiasEither {
    /* snip */
  }
}
```

This should compile warning free and we'd get our code up to 2.12 level.

sjrd:
Meh ... I think I prefer the verbose version than a version-dependent compat object.

That said, that solution would also work with an implicit `RightBiasEither` that is available on all Scala versions. The true `Either.foreach` method of 2.12 would take precedence over the implicit pimp when compiling with 2.12.

Maybe we should do that. WDYT?

gzm0:
> implicit RightBiasEither that is available on all Scala versions

Right, I forgot that we are actually able to write that warning free :) Just need to pattern match.

Yeah, that's probably neater. I prefer having compat objects that hide the ugliness and have our code be on the newest Scala level where possible.

sjrd:
Updated with the `Either` pimp, and a few updates to the overridden `.check` files for partest.

sjrd:
Uh oh!
```
Checksizes: Scala version: 2.12.0-RC1
Reversi preopt size = 594344 (expected 528000)
Reversi opt size = 148876 (expected 124000)
Reversi preopt gzip size = 77217 (expected 67000)
Reversi opt gzip size = 35696 (expected 30000)
```
That's a very significant jump from M5 to RC1 :frowning: We'll have to investigate that.

sjrd:
Here is the culprit: https://github.com/scala/scala/commit/30876fe2dd8cbe657a6cad6b11bbc34f10c29b36
It introduces a call to `String.split(Char)`, which transitively brings in the `java.util.matching.*` implementation. Since `stringPrefix` is called by `MostCollection.toString()`, this means that every single Scala.js codebase is paying the tax of `java.util.matching.*`. We really need to fix this.

sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/4417/

ci/checksizes.sh
```diff
@@ -45,11 +45,11 @@ case $FULLVER in
     REVERSI_PREOPT_GZ_EXPECTEDSIZE=71000
     REVERSI_OPT_GZ_EXPECTEDSIZE=32000
     ;;
-  2.12.0-M5)
-    REVERSI_PREOPT_EXPECTEDSIZE=528000
-    REVERSI_OPT_EXPECTEDSIZE=124000
-    REVERSI_PREOPT_GZ_EXPECTEDSIZE=67000
-    REVERSI_OPT_GZ_EXPECTEDSIZE=30000
+  2.12.0-RC1)
+    REVERSI_PREOPT_EXPECTEDSIZE=595000
```
nicolasstucki: Is there a good reason for this increase?

nicolasstucki:
That is all. Otherwise LGTM.

ci/checksizes.sh
sjrd: Define "good" :P The reason is https://github.com/scala-js/scala-js/pull/2590#issuecomment-246685424
It's coming from scala/scala, there's nothing we can do about right now (in this PR), but I've recorded the problem as #2591 so that we can address it later.

sjrd:
Nightly passed :)

PR #2593, opened 2016-09-14
Fix #2592: Do shrink the result of Pattern.split() to an empty array.

### LINKED ISSUE #2592: Yet another String.split() inconcistency
Scala/JVM:
```scala
scala> "a".split("a")
res12: Array[String] = Array()
```
Scala.js:
```scala
Array("")
```
In the Scala.js implem we have:
```scala
    /* With `limit == 0`, remove trailing empty strings (but do not reduce
     * the array to be zero length).
     */
```
It appears the "but do not reduce the array to be zero length" is erroneous. The array can be shrunk to be empty.

However, splitting an empty string always returns an array of exactly one empty empty string.

sjrd:
But not if the input string is the empty string itself.

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Pattern.scala
+++ b/javalib/src/main/scala/java/util/regex/Pattern.scala
@@ -42,46 +42,48 @@ final class Pattern private (jsRegExp: js.RegExp, _pattern: String, _flags: Int)
     split(input, 0)
 
   def split(input: CharSequence, limit: Int): Array[String] = {
-    val lim = if (limit > 0) limit else Int.MaxValue
-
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0. Best reviewed [with ?w=1](https://github.com/scala-js/scala-js/pull/2593/files?w=1)

gzm0:
LGTM

gzm0:
Didn't know about `w=1`. Is it new?

sjrd:
It has existed for as long as I've been using GitHub. I sometimes recommend to use it on PRs. I'm surprised it's the first time you see me mention it.

PR #2596, opened 2016-09-15
Tone down the number of combinations in CI.

sjrd:
Our resource consumption of CI has grown much too intense, to a point where it is slowing us down, as PR testing can be delayed by several hours.

This commit "strategically" reduces the number of combinations that are tested on CI. Basically, we assume 2.10.2, 2.11.8 and 2.12.0-RC1 combined with JDK8 only as the "default", "base" combination. From that combination, we vary all the other parameters independently, but we do not combine several axes together. One exception is for sbt-plugin related things, whose "base" combination is 2.10.6 with all the JDKs.

To make sure that the `testSuite` (which contains JDK7-only and JDK8-only parts) runs on JDK6 and 7 in `pr`, one `testSuite` run (in the default config, i.e., Node.js + fastOpt) is added to the `main` task.

Partest fastOpt moves from `nightly` to `pr`. The tests in `pr` have become slow enough that two partest tasks (for 2.11.8 and 2.12.0-RC1) will not kill it. And this is the one thing from `nightly` that often breaks, so it is worth having in `pr`.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -58,7 +58,8 @@
         'set scalaJSOptimizerOptions in testingExample ~= (_.withDisableOptimizer(true))' \
         ++$scala testingExample/test:run testingExample/test &&
     sbtretry ++$scala testSuiteJVM/test testSuiteJVM/clean &&
-    sbtretry ++$scala 'testSuite/test:runMain org.scalajs.testsuite.junit.JUnitBootstrapTest' testSuite/clean &&
+    sbtretry ++$scala 'testSuite/test:runMain org.scalajs.testsuite.junit.JUnitBootstrapTest' &&
+    sbtretry ++$scala testSuite/test &&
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @gzm0 and @nicolasstucki. WDYT?

gzm0:
Sounds good in principle. Will review in detail tonight.

ci/matrix.xml
```diff
@@ -486,452 +415,118 @@
       <v n="scala">2.11.0</v>
       <v n="java">1.7</v>
     </run>
```
sjrd: Yes it's for 2.11.0 specifically.

gzm0: Ok :)

sjrd:
>  gzm0 approved these changes 3 hours ago 

Looks like GitHub has killed "LGTM" with their new feature ;-)

nicolasstucki:
LGTM

PR #2597, opened 2016-09-16
Give 1 KB more to the checksizes for 2.12.0-RC1.

sjrd:
Apparently, the previous limit was a little bit too tight, and on some machines the file is a little bit bigger.

No review.


gzm0:
Ahem... As I was going to comment on the other PR: Isn't something **seriously** wrong if our output size is not deterministic?

gzm0:
Just to be clear: We can of course merge this to make CI pass again, but I think we should investigate this. This might be a latent 2.12 bug (how much concurrency is in there?).

sjrd:
We've always had some wiggle room of 100 bytes or so of difference depending on ... things. On any single machine, the output size is completely deterministic, so I think this is environmental. We sure could try and find exactly what is causing this.

gzm0:
Hmm... Ok. Sounds very scary, but if you say it was like that, at least it does not indicate a 2.12 bug.

PR #2599, opened 2016-09-17
Fix #2583: Use JSDOMNodeJSEnv by default when requestsDOM is true.

### LINKED ISSUE #2583: Use JSDOMNodeJSEnv by default when requestsDOM is true (instead of PhantomJS)
Since we're breaking defaults around JS envs anyway (#2579), maybe we should consider switching the default JS env when `requestsDOM` is true to be `JSDOMNodeJSEnv` instead of `PhantomJSEnv`.

I don't know whether the jsdom env has been seriously adopted so far. But we definitely know that [PhantomJS is unnerving people](https://github.com/scala-js/scala-js/issues/1881) who actually try to use it. Switching to jsdom by default will probably have a positive net result. For people who still prefer PhantomJS, it's a one-line config to enable it back:
```scala
jsEnv := PhantomJSEnv().value
```
Hopefully, in 1.0.0 we can move `PhantomJSEnv` out of core, like `SeleniumJSEnv`. With the removal of `RhinoJSEnv` altogether, that would leave us with only Node.js-based envs (with and without jsdom) in core, which would be nice.

sjrd:
(instead of PhantomJS)

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -20,9 +20,9 @@
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
         ++$scala helloworld/run \
         helloworld/clean &&
-    sbtretry 'set inScope(ThisScope in helloworld)(jsEnv := JSDOMNodeJSEnv().value)' \
+    sbtretry 'set requiresDOM in helloworld := true' \
         ++$scala helloworld/run &&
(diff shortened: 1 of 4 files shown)


sjrd:
Review by @gzm0

sjrd:
Jenkins, retest this please

PR #2600, opened 2016-09-18
Properly incrementalize the Emitter using knowledge queries.

sjrd:
When a paper's terminology makes its way into a production codebase ...

The `ScalaJSClassEmitter` and `JSDesugaring` do not have access to the `linkingUnit` anymore. All their requests for global knowledge go through a `GlobalKnowledge` interface, which contains the knowledge queries. The `Emitter` contains a subsystem `KnowledgeGuardian`, which, as its name implies, is responsible for guarding global knowledge. It tracks the dependencies between arbitrary `Invalidatable`s and the knowledge queries they perform. The `Invalidatable`s are `ClassCache` and `MethodCache`, which also implement `GlobalKnowledge` by delegating queries to the `KnowledgeGuardian`.

This fixes some corner case bugs in the incremental behavior of the `Emitter`, the most significant one being changes in the `@JSName` of a native JS class or object. Previously, users of `LoadJSConstructor` (resp. `LoadJSModule`) would not be properly re-desugared when the `@JSName` of a target class (resp. module) was changed.

The new architecture can now easily be extended with new kinds of global knowledge without fear of breaking incrementality.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
@@ -33,7 +33,11 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
   def this(semantics: Semantics, outputMode: OutputMode) =
     this(semantics, outputMode, InternalOptions())
 
-  private var classEmitter: ScalaJSClassEmitter = _
+  private val knowledgeGuardian = new KnowledgeGuardian
+
+  private val classEmitter =
+    new ScalaJSClassEmitter(semantics, outputMode, internalOptions)
+
   private val classCaches = mutable.Map.empty[List[String], ClassCache]
 
   private[this] var statsClassesReused: Int = 0
@@ -73,15 +77,13 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
   }
 
   def emit(unit: LinkingUnit, builder: JSTreeBuilder, logger: Logger): Unit = {
-    classEmitter = new ScalaJSClassEmitter(outputMode, internalOptions, unit)
-    startRun()
+    startRun(unit)
     try {
       val orderedClasses = unit.classDefs.sortWith(compareClasses)
       for (classInfo <- orderedClasses)
         emitLinkedClass(classInfo, builder)
     } finally {
       endRun(logger)
-      classEmitter = null
     }
   }
 
@@ -103,11 +105,16 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
     else lhs.encodedName.compareTo(rhs.encodedName) < 0
   }
 
-  private def startRun(): Unit = {
+  private def startRun(unit: LinkingUnit): Unit = {
     statsClassesReused = 0
     statsClassesInvalidated = 0
     statsMethodsReused = 0
     statsMethodsInvalidated = 0
+
+    val invalidateAll = knowledgeGuardian.update(unit)
+    if (invalidateAll)
+      classCaches.clear()
+
     classCaches.valuesIterator.foreach(_.startRun())
   }
 
@@ -136,24 +143,24 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       val methodCache = classCache.getStaticCache(m.info.encodedName)
 
       addTree(methodCache.getOrElseUpdate(m.version,
-        classEmitter.genMethod(className, m.tree)))
+          classEmitter.genMethod(className, m.tree)(methodCache)))
     }
 
     if (linkedClass.hasInstances && kind.isAnyScalaJSDefinedClass) {
       val ctor = classTreeCache.constructor.getOrElseUpdate(
-          classEmitter.genConstructor(linkedClass))
+          classEmitter.genConstructor(linkedClass)(classCache))
 
       // Normal methods
       val memberMethods = for (m <- linkedClass.memberMethods) yield {
         val methodCache = classCache.getMethodCache(m.info.encodedName)
 
         methodCache.getOrElseUpdate(m.version,
-            classEmitter.genMethod(className, m.tree))
+            classEmitter.genMethod(className, m.tree)(methodCache))
       }
 
       // Exported Members
       val exportedMembers = classTreeCache.exportedMembers.getOrElseUpdate(
-          classEmitter.genExportedMembers(linkedClass))
+          classEmitter.genExportedMembers(linkedClass)(classCache))
 
       outputMode match {
         case OutputMode.ECMAScript51Global | OutputMode.ECMAScript51Isolated =>
@@ -169,14 +176,14 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
             case js.Skip()         => Nil
             case oneMember         => List(oneMember)
           }
-          addTree(classEmitter.genES6Class(linkedClass, allMembers))
+          addTree(classEmitter.genES6Class(linkedClass, allMembers)(classCache))
       }
     } else if (kind == ClassKind.Interface) {
       // Default methods
       for (m <- linkedClass.memberMethods) yield {
         val methodCache = classCache.getMethodCache(m.info.encodedName)
         addTree(methodCache.getOrElseUpdate(m.version,
-            classEmitter.genDefaultMethod(className, m.tree)))
+            classEmitter.genDefaultMethod(className, m.tree)(methodCache)))
       }
     }
 
@@ -189,7 +196,7 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
 
     if (linkedClass.hasRuntimeTypeInfo) {
       addTree(classTreeCache.typeData.getOrElseUpdate(
-          classEmitter.genTypeData(linkedClass)))
+          classEmitter.genTypeData(linkedClass)(classCache)))
     }
 
     if (linkedClass.hasInstances && kind.isClass && linkedClass.hasRuntimeTypeInfo)
@@ -201,7 +208,7 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
           classEmitter.genModuleAccessor(linkedClass)))
 
     addTree(classTreeCache.classExports.getOrElseUpdate(
-        classEmitter.genClassExports(linkedClass)))
+        classEmitter.genClassExports(linkedClass)(classCache)))
   }
 
   // Helpers
@@ -226,15 +233,44 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       emitNextLine(0)
   }
 
+  // Private API for Rhino
+
+  private[scalajs] object rhinoAPI { // scalastyle:ignore
+    /** A GlobalKnowledge that never tracks dependencies. This can be used in
+     *  cases where we do not use any cache, which is what `genClassDef()` in
+     *  this class does.
+     */
+    private val globalKnowledge: GlobalKnowledge =
+      new knowledgeGuardian.KnowledgeAccessor {}
+
+    def initialize(linkingUnit: LinkingUnit): Unit =
+      startRun(linkingUnit)
+
+    def getHeaderFile(): org.scalajs.core.tools.io.VirtualJSFile =
+      CoreJSLibs.lib(semantics, outputMode)
+
+    def genClassDef(linkedClass: LinkedClass): js.Tree =
+      classEmitter.genClassDef(linkedClass)(globalKnowledge)
+  }
+
   // Caching
 
-  private final class ClassCache {
+  private final class ClassCache extends knowledgeGuardian.KnowledgeAccessor {
     private[this] var _cache: DesugaredClassCache = null
     private[this] var _lastVersion: Option[String] = None
     private[this] var _cacheUsed = false
 
     private[this] val _staticCaches = mutable.Map.empty[String, MethodCache]
     private[this] val _methodCaches = mutable.Map.empty[String, MethodCache]
+
+    override def invalidate(): Unit = {
+      /* Do not invalidate contained methods, as they have their own
+       * invalidation logic.
+       */
+      super.invalidate()
+      _cache = null
+      _lastVersion = None
+    }
 
     def startRun(): Unit = {
       _cacheUsed = false
@@ -244,6 +280,7 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
 
     def getCache(version: Option[String]): DesugaredClassCache = {
       if (_cache == null || _lastVersion.isEmpty || _lastVersion != version) {
+        invalidate()
         statsClassesInvalidated += 1
         _lastVersion = version
         _cache = new DesugaredClassCache
@@ -265,21 +302,28 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       _methodCaches.retain((_, c) => c.cleanAfterRun())
 
       if (!_cacheUsed)
-        _cache = null
+        invalidate()
 
       _staticCaches.nonEmpty || _methodCaches.nonEmpty || _cacheUsed
     }
   }
 
-  private final class MethodCache {
+  private final class MethodCache extends knowledgeGuardian.KnowledgeAccessor {
     private[this] var _tree: js.Tree = null
     private[this] var _lastVersion: Option[String] = None
     private[this] var _cacheUsed = false
+
+    override def invalidate(): Unit = {
+      super.invalidate()
+      _tree = null
+      _lastVersion = None
+    }
 
     def startRun(): Unit = _cacheUsed = false
 
     def getOrElseUpdate(version: Option[String], v: => js.Tree): js.Tree = {
       if (_tree == null || _lastVersion.isEmpty || _lastVersion != version) {
+        invalidate()
         statsMethodsInvalidated += 1
         _tree = v
         _lastVersion = version
@@ -290,7 +334,12 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       _tree
     }
 
-    def cleanAfterRun(): Boolean = _cacheUsed
+    def cleanAfterRun(): Boolean = {
+      if (!_cacheUsed)
+        invalidate()
+
+      _cacheUsed
+    }
   }
 }
 
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
@@ -1,0 +1,250 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js tools             **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2016, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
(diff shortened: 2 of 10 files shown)


sjrd:
Review by @gzm0. I think you'll like this one :P

/cc @tOverney and @paullepoulpe, as this will also be the basis for making #2142 and #2411 properly incremental.

sjrd:

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -226,16 +232,277 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+    def getHeaderFile(): org.scalajs.core.tools.io.VirtualJSFile =
+      CoreJSLibs.lib(semantics, outputMode)
+
+    def genClassDef(linkedClass: LinkedClass): js.Tree =
+      classEmitter.genClassDef(linkedClass)
+  }
+
+  // Guardian of all global knowledge
+
+  private abstract class InvalidatableBag {
```
gzm0: Could we use the same terminology as in the optimizer? IIUC this is called `Unregisterable` in the optimizer (which I find easier to understand).

Further, the meaning of "register" (in `registerBag`) seems to be inversed. In the optimizer, an `Invalidatable`, i.e. a `MethodImpl` can register to a dependency and is itself `registeredTo` and can be `unregistered`.

Here we use the terminology `register` as `registeredTo` and `forget` for unregister.

TBH, I'm wondering if it is worth making a common base type `Unregisterable[T]` for both the optimizer and here (maybe also for `Invalidatable`)? (This makes me think that `invalidate` is called `tag` in the optimizer. `invalidate` is definitely the better name, so we should keep it and maybe change it in the optimizer).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -226,16 +232,277 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+    def forgetAbout(invalidatable: Invalidatable): Unit
+  }
+
+  private object KnowledgeGuardian { knowledgeGuardian =>
```
gzm0: Could we factor this out to a separate class?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -226,16 +232,277 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+      classes(className).askJSSuperClass(invalidatable)
+
+    def askJSClassFieldDefs(className: String, invalidatable: Invalidatable): List[FieldDef] =
+      classes(className).askJSClassFieldDefs(invalidatable)
+  }
+
   // Caching
 
-  private final class ClassCache {
+  private abstract class Invalidatable extends GlobalKnowledge {
```
gzm0: I would reserve `Invalidatable` for really just the basic invalidatable (with `invalidate()` and nothing more).

Maybe this is a `KnowledgeAccessor` or something?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -226,16 +232,277 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       emitNextLine(0)
   }
 
+  // Private API for Rhino
+
+  private[scalajs] object rhinoAPI { // scalastyle:ignore
+    private implicit val GlobalKnowledge: GlobalKnowledge = new Invalidatable {
```
gzm0: I think this deserves a comment that you just don't want to write another set of forwarders in the `KnowledgeGuardian` (which is absolutely fair).
OTOH, if we call `Invalidatable` `KnowledgeAccessor`, this is probably much clearer.

sjrd: It's probably clearer with `KnowledgeAccessor`, indeed.

It really isn't about creating other forwarders in `KnowledgeGuardian`. This should be seen as a sin, since we would have a guardian that is not actually guarding anything, since it offers an API that doesn't actually track anything. It is the logical thing to do in `rhinoAPI` to fake the tracking, because we know that Rhino doesn't reuse an `Emitter` twice. I should write that as a comment, in fact ^^

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: > Could we use the same terminology as in the optimizer? IIUC this is called Unregisterable in the optimizer (which I find easier to understand).
>
> Further, the meaning of "register" (in registerBag) seems to be inversed. In the optimizer, an Invalidatable, i.e. a MethodImpl can register to a dependency and is itself registeredTo and can be unregistered.
>
> Here we use the terminology register as registeredTo and forget for unregister.

Yes, you're right. I'll adapt to the terminology already used in the optimizer.

> TBH, I'm wondering if it is worth making a common base type Unregisterable[T] for both the optimizer and here (maybe also for Invalidatable)?

IMO that is overkill, and will introduce an internal dependency between the optimizer and the emitter that is not semantically nor practically justified.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: `KnowledgeAccessor` sounds good.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: I don't think this is meaningful/wise. The knowledge guardian is logically coupled to maintaining the caches. Most of `Emitter` is about maintaining caches. There is only a very limited part which deals with the mega-structure of the code emission, and that part is quite directly manipulating the caches, so in fact it is also about the caches.

We could move all of those things together to something like `IncEmitter`, but there wouldn't remain anything in `Emitter` but a call to `incEmitter.emit()` ;-)

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
gzm0: > The knowledge guardian is logically coupled to maintaining the caches.

Are you sure about that? It seems that the point of the knowledge guardian precisely is to decouple caching and invalidation logic. The knowledge guardian does not know about caches (neither it should). All it does is track what was using which assumptions and invalidate these elements as needed.

The Emitter on the other hand ties caches and Invalidatables together. Think of it that way: It would be absolutely sensible to use the knowledge guardian to implement a piece of code that replaces code on the fly in a VM every new linker run. There are no caches (at least not in the Emitters sense) in that scenario, nevertheless, an instance of the knowledge guardian can be used unchanged.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
gzm0: > IMO that is overkill.

Agreed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: I have added an additional commit with `KnowledgeGuardian` moved outside of `Emitter`. Is that what you have in mind?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+
+  trait Invalidatable {
+    private[KnowledgeGuardian] def registeredTo(
+        unregisterable: Unregisterable): Unit
+
+    def invalidate(): Unit
+  }
+
+  abstract class KnowledgeAccessor(knowledgeGuardian: KnowledgeGuardian)
+      extends GlobalKnowledge with Invalidatable {
```
sjrd: This class could also be inside the *class* `KnowledgeGuardian`, and we would use it as `extends knowledgeGuardian.KnowledgeAccessor` instead of `extends KnowledgeAccessor(knowledgeGuardian)`. WDYT?

gzm0: Yes, probably not a bad idea.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+
+  /** Returns `true` if *all* caches should be invalidated.
+   *
+   *  For global properties that are rarely changed and heavily used (such as
+   *  isParentDataAccessed), we do not want to pay the price of the
+   *  dependency graph, in terms of memory consumption and time spent
+   *  maintaining it. It is a better trade-off to invalidate everything in
+   *  the rare events where they do change.
+   */
+  def startRun(linkingUnit: LinkingUnit): Boolean = {
```
gzm0: Thinking of it, maybe we should call this `update`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+
+    isParentDataAccessed = newIsParentDataAccessed
+    hasNewRuntimeLong = newHasNewRuntimeLong
+
+    if (invalidateAll)
+      classes.valuesIterator.foreach(_.unregisterAll())
+    invalidateAll
+  }
+
+  private[KnowledgeGuardian] def askHasNewRuntimeLong(invalidatable: Invalidatable): Boolean =
```
gzm0: I like that this is private.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+
+    def getSuperClassOfJSClass(className: String): String =
+      knowledgeGuardian.askJSSuperClass(className, this)
+
+    def getJSClassFieldDefs(className: String): List[FieldDef] =
+      knowledgeGuardian.askJSClassFieldDefs(className, this)
+  }
+}
+
+private[emitter] final class KnowledgeGuardian { knowledgeGuardian =>
```
gzm0: I think the knowledgeGuardian this name is unused.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+      extends GlobalKnowledge with Invalidatable {
+
+    /* In theory, an KnowledgeAccessor should *contain* a GlobalKnowledge, not
+     * *be* a GlobalKnowledge. We organize it that way to reduce memory
+     * footprint and pointer indirections.
+     */
+
+    private val _registeredTo = mutable.Set.empty[Unregisterable]
+
+    def invalidate(): Unit = {
```
gzm0: I'm wondering if we should make this abstract and make a guardian private method that calls `invalidate` and `unregisterFromAll` which we call in this method's stead.
This essentially prevents any mistakes when implementing subclasses of `KnowledgeAccessor`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -0,0 +1,265 @@
...
+  private[KnowledgeGuardian] trait Unregisterable {
+    def unregister(invalidatable: Invalidatable): Unit
+  }
+
+  trait Invalidatable {
+    private[KnowledgeGuardian] def registeredTo(
+        unregisterable: Unregisterable): Unit
+
+    def invalidate(): Unit
+  }
```
gzm0: Do we even need this trait anymore?

sjrd:
> Can you live with the separate KnowledgeGuardian or do you still think it should live inside the Emitter?

I'm fine either way, so it can be separate.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: Yes. I had that thought at some point too, so I guess it should be `update`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: OK

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: Not necessarily, but then `Unregisterable.unregister` needs to take a `KnowledgeAccessor`. If `KnowledgeAccessor` goes inside the class `KnowledgeGuardian`, then `Unregisterable` must do too (which is fine), unless we type is as `KnowledgeGuardian#KnowledgeAccessor`.

Also, if `Invalidate` goes away, by the same argument we could get rid of `Unregisterable` since only `Class` implements it. But that has a weirder feel about it.

IMO keeping `Unregisterable` and `Invalidatable` separate is clearer.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: Indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: Makes sense.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -17,65 +17,33 @@ import org.scalajs.core.ir.Trees.FieldDef
...
 private[emitter] object KnowledgeGuardian {
-  private[KnowledgeGuardian] trait Unregisterable {
+  trait Unregisterable {
     def unregister(invalidatable: Invalidatable): Unit
   }
 
   trait Invalidatable {
-    private[KnowledgeGuardian] def registeredTo(
-        unregisterable: Unregisterable): Unit
+    def registeredTo(unregisterable: Unregisterable): Unit
```
sjrd: I had to make this public in the version where `KnowledgeAccessor` is in the class `KnowledgeGuardian`. Otherwise scalac would complain with abstract methods not being implemented although there was no reason. Apparently refchecks got confused with `private[KnowledgeGuardian]`, thinking it means something different in the context of the object or the class :(

gzm0: Hmmm... Sadness... Moving it into the class also doesn't help? IMO that would be acceptable and better than having public stuff because Scalac can't handle it.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
gzm0: Ok. That's fair. Having `Invalidatable` in the code definitely read easier than `KnowledgeAccessor`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: If it were truly public, I would agree. But here it's only "public" up to `emitter`, so it's not really a big deal. IMO it's better to keep them in the companion to avoid the extra outer pointers.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -244,6 +282,7 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
 
     def getCache(version: Option[String]): DesugaredClassCache = {
       if (_cache == null || _lastVersion.isEmpty || _lastVersion != version) {
+        invalidate()
```
gzm0: Ah, I didn't think about this. Probably not a performance problem if it doesn't deregister itself.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -265,21 +304,27 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       _methodCaches.retain((_, c) => c.cleanAfterRun())
 
       if (!_cacheUsed)
-        _cache = null
+        invalidate()
```
gzm0: Oh... But this one might actually cause memory leaks IIUC :(

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -290,7 +335,12 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
       _tree
     }
 
-    def cleanAfterRun(): Boolean = _cacheUsed
+    def cleanAfterRun(): Boolean = {
+      if (!_cacheUsed)
+        invalidate()
```
gzm0: Same here?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
sjrd: I managed to make them private again, simply by making `registeredTo()` and `invalidateAndUnregisterFromAll()` concrete in `Invalidatable`, which makes sense anyway.

gzm0: Nice!

sjrd:
Duh! Yes, those are leaks indeed! I think we need to make `invalidateAndUnregisterFromAll()` public (or at least protected final).

gzm0:
(of course after squashing)

sjrd:
Squashed.

PR #2604, opened 2016-09-22
Fix #2602: Linker gives bad error if used in invalid state

### LINKED ISSUE #2602: Linker thinks it's used concurrently when in fact it's been made invalid after an exception
See https://github.com/scalafiddle/scalafiddle-editor/issues/23
Source for context: https://github.com/scala-js/scala-js/blob/v0.6.12/tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala#L54

Sequence:

1. A Linker is used properly, but throws an exception -> enters the invalid state
2. The Linker is used a second time, but it's invalid. It throws `IllegalStateException("Invalid due to a previous exception")` as expected, *but* it stays in the state "I'm being used" (`_linking` is true)
3. The linker is used a third time. It should say that it's invalid, but instead it stays it is used concurrently.

In any case, there should be an exception. But the error message is not the appropriate one, which can hinder debuggability.

Trivial fix: move the Invalid check inside the try/finally.

gzm0:


gzm0:
Review by @sjrd

PR #2607, opened 2016-09-25
Fix #2603: Default values for non-exposed defs.

### LINKED ISSUE #2603: Inner def with default param in a Scala.js-defined JS class produces invalid IR
Reproduction:
```scala
import scala.scalajs.js
import js.annotation._

@ScalaJSDefined
class Foo extends js.Object {
  def bar(): Int = {
    def inner(start: Int = 0): Int = start + 1
    inner()
  }
}

  def main() {
    val foo = new Foo
    assert(foo.bar() == 1)
  }
}
```
IR checker error:
```
[error] file:/localhome/doeraene/projects/scalajs/examples/helloworld/HelloWorld.scala(15:5:UndefinedParam): Invalid expression tree
```

gzm0:


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1998,7 +1998,24 @@ abstract class GenJSCode extends plugins.PluginComponent
           isRawJSCtorDefaultParam(sym)
         } else {
           sym.hasFlag(reflect.internal.Flags.DEFAULTPARAM) &&
-          isRawJSType(sym.owner.tpe)
+          isRawJSType(sym.owner.tpe) && {
+            /* If this is a default parameter accessor on a
+             * ScalaJSDefinedJSClass, we need to know if the method for which we
+             * are the default parameter is exposed or not.
+             * We do this by looking up the de-mangled name in the owner. Note
```
sjrd: "de-mangled" makes me think about `unencodedName` much more than what's happening here. Consider something like "We do this by removing the `$default` suffix from the method name, and looking up a member with that name in the owner."

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1998,7 +1998,24 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+             * that this does not work for local methods. But local methods are
+             * never exposed.
+             * Further note that overloads are easy, because either all or none
+             * of them are exposed.
+             */
+            def isOwnerMethodExposed = {
```
sjrd: "owner" is not very accurate wrt what scalac calls an owner. Consider maybe `isAttachedMethodExposed`?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -1307,6 +1307,19 @@ class ScalaJSDefinedTest {
     assertEquals("hello3 4", dyn.foo("hello", 4))
   }
 
+  @Test def overload_with_default_parameter(): Unit = {
+    @ScalaJSDefined
+    class OverloadDefaultParameter extends js.Object {
+      def foo(x: Int) = x
+      def foo(x: String = "") = x
```
sjrd: Public methods need result type (I think Scalastyle will complain about this).

sjrd:
What made the issues you were describing in #2603 go away?

gzm0:
> What made the issues you were describing in #2603 go away?

Checking for the existence of the `Symbol` before feeding it to `isExposed`.

gzm0:
Updated.

PR #2614, opened 2016-10-01
Implement java.util.Timer.

axel22:
Here's a tentative implementation of the `Timer` and `TimerTask` classes. I'm not sure if the current test suite supports asynchronous testing, what do you suggest?

### DIFF
--- a/javalib/src/main/scala/java/util/Timer.scala
+++ b/javalib/src/main/scala/java/util/Timer.scala
@@ -1,0 +1,130 @@
+package java.util
+
+import scala.collection._
+import scala.concurrent.duration._
+
+class Timer() {
+  private[util] var canceled: Boolean = false
+
+  def this(isDaemon: Boolean) = this()
+
+  def this(name: String) = this()
+
+  def this(name: String, isDaemon: Boolean) = this()
+
+  private def acquire(task: TimerTask): Unit = {
+    if (canceled)
+      throw new IllegalStateException("Timer already cancelled.")
+    else if (task.owner != null || task.canceled) {
+      throw new IllegalStateException("TimerTask already scheduled or canceled.")
+    }
+    task.owner = this
+  }
+
+  private def checkDelay(delay: Long): Unit = {
+    if (delay < 0 || (delay + System.currentTimeMillis) < 0)
+      throw new IllegalArgumentException("Negative delay.")
+  }
+
+  private def checkTime(time: Date): Unit = {
+    if (time.getTime < 0)
+      throw new IllegalArgumentException(s"Negative time: $time.")
+  }
+
+  private def checkPeriod(period: Long): Unit = {
+    if (period <= 0)
+      throw new IllegalArgumentException("Non-positive period.")
+  }
+
+  private def scheduleOnce(task: TimerTask, delay: Long): Unit = {
+    acquire(task)
+    task.timeout(delay.millis) {
+      task.scheduledOnceAndStarted = true
+      task.doRun()
+    }
+  }
+
+  private def getMillisUntil(time: Date): Long =
+    math.max(0L, time.getTime - System.currentTimeMillis())
+
+  def schedule(task: TimerTask, delay: Long): Unit = {
+    checkDelay(delay)
+    scheduleOnce(task, delay)
+  }
+
+  def schedule(task: TimerTask, time: Date): Unit = {
+    checkTime(time)
+    val delay = getMillisUntil(time)
+    scheduleOnce(task, delay)
+  }
+
+  private def schedulePeriodically(
+      task: TimerTask, delay: Long, period: Long): Unit = {
+    acquire(task)
+    task.timeout(delay.millis) {
+      def loop(): Unit = {
+        val startTime = System.nanoTime()
+        task.doRun()
+        val endTime = System.nanoTime()
+        val duration = (endTime - startTime) / 1000000
+        task.timeout((period - duration).millis) {
+          loop()
+        }
+      }
+      loop()
+    }
+  }
+
+  def schedule(task: TimerTask, delay: Long, period: Long): Unit = {
+    checkDelay(delay)
+    checkPeriod(period)
+    schedulePeriodically(task, delay, period)
+  }
+
+  def schedule(task: TimerTask, firstTime: Date, period: Long): Unit = {
+    checkTime(firstTime)
+    checkPeriod(period)
+    val delay = getMillisUntil(firstTime)
+    schedulePeriodically(task, delay, period)
+  }
+
+  private def scheduleFixed(
+      task: TimerTask, delay: Long, period: Long): Unit = {
+    acquire(task)
+    task.timeout(delay.millis) {
+      def loop(scheduledTime: Long): Unit = {
+        task.doRun()
+        val nextScheduledTime = scheduledTime + period
+        val nowTime = System.nanoTime / 1000000L
+        if (nowTime >= nextScheduledTime) {
+          // Re-run immediately.
+          loop(nextScheduledTime)
+        } else {
+          // Re-run after a timeout.
+          task.timeout((nextScheduledTime - nowTime).millis) {
+            loop(nextScheduledTime)
+          }
+        }
+      }
+      loop(System.nanoTime / 1000000L + period)
+    }
+  }
+
+  def scheduleAtFixedRate(task: TimerTask, delay: Long, period: Long): Unit = {
+    checkDelay(delay)
+    checkPeriod(period)
+    scheduleFixed(task, delay, period)
+  }
+
+  def scheduleAtFixedRate(task: TimerTask, firstTime: Date, period: Long): Unit = {
+    checkTime(firstTime)
+    checkPeriod(period)
+    val delay = getMillisUntil(firstTime)
+    scheduleFixed(task, delay, period)
+  }
+
+  def cancel(): Unit = canceled = true
+
+  def purge(): Int = 0
+
+}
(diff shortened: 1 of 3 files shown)


sjrd:
For "asynchronous" testing, you could reuse the [`TimeoutMock`](https://github.com/scala-js/scala-js/blob/master/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/TimeoutMock.scala), which is used for example here: https://github.com/scala-js/scala-js/blob/97979a7f5f1460c90faf5e0806c2212baed338a2/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/AsyncTest.scala#L82

gzm0:
QQ, this is relying on `setTimeout`, which is a non-standard JS method. Do we want to accept the precedent of using them in javalib?

sjrd:
I think we already created that precedent a long time ago with `Future`s ;)

javalib/src/main/scala/java/util/Timer.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  private[util] def scheduleOnce(task: TimerTask, delay: Long): Unit = {
+    tasks += task
+    setTimeout(delay.millis) {
+      task.doRun()
+    }
+  }
+
+  private[util] def getMillisUntil(time: Date): Long = {
+    math.max(0, time.getTime - (new Date()).getTime)
```
sjrd: Consider using `System.currentTimeMillis()` instead of `(new Date()).getTime`, which avoids the allocation of the `java.util.Date` object.

javalib/src/main/scala/java/util/TimerTask.scala
```diff
@@ -0,0 +1,30 @@
+package java.util
+
+import scala.scalajs.js.timers
+
+abstract class TimerTask {
+  private[util] var cancelled: Boolean = false
```
sjrd: Same remark on spelling.

javalib/src/main/scala/java/util/Timer.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  def this(isDaemon: Boolean) = this("timer", isDaemon)
+
+  def this(name: String) = this(name, false)
+
+  private[util] def scheduleOnce(task: TimerTask, delay: Long): Unit = {
+    tasks += task
+    setTimeout(delay.millis) {
+      task.doRun()
+    }
```
sjrd: This method and the other private schedule helpers are missing a test for `canceled`, which should throw an `IllegalStateException`, as well as tests for the validity of `delay`/`time` which should throw `InvalidArgumentException`.

sjrd: Also I think the `task` should be removed from `tasks` upon completion, otherwise there's a memory leak in `tasks`.

javalib/src/main/scala/java/util/TimerTask.scala
```diff
@@ -0,0 +1,30 @@
...
+  private[util] var lastScheduled: Long = 0L
+
+  def run(): Unit
+
+  def cancel(): Unit = {
+    cancelled = true
+  }
```
sjrd: Since we have no weak reference mechanism in JavaScript to lazily clean up the `tasks` queue, `cancel()` should actually remove this task from the `Timer.tasks` queue eagerly.

javalib/src/main/scala/java/util/TimerTask.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  private[util] def doRun() {
+    if (!cancelled) {
+      lastScheduled = System.currentTimeMillis()
+      try {
+        run()
+      } catch {
+        case t: Throwable =>
+          cancelled = true
+          throw t
```
sjrd: Is that behavior on exception specified? In other words, should a repeated task actually be canceled if one of its runs throws an exception?

axel22:
About the `tasks` comment - what you say is what I initially wanted to do, one potential problem is that a `TimerTask` could be submitted to multiple timers (I think). Maybe I missed it, but I can't remember docs saying anything about that.

javalib/src/main/scala/java/util/TimerTask.scala
axel22: As far as I know, an exception cancels the timer task, but this is not specified in the docs (e.g. http://stackoverflow.com/questions/8743027/java-timer-class-timer-tasks-stop-to-execute-if-in-one-of-the-tasks-exception-i).

sjrd:
It seems the doc doesn't say anything about that, indeed. However, I think it is fair to say that it is clearly not the expected use. For example, if it is accepted, then `timer.cancel()` is underspecified as whether the tasks it has scheduled are also canceled from the other timers' perspectives or not? In general, the documentation of all the methods seems to be written with the assumption that a `TimerTask` is only scheduled in one `Timer`.

Also, there is no working around the fact that we *need* to eagerly remove tasks from the timer's queue when `task.cancel()` is called. So a `TimerTask` has no choice but to know in which `Timer` it has been scheduled. Granted, it could remember a *set* of timers it has been scheduled to, though.

javalib/src/main/scala/java/util/TimerTask.scala
axel22: Indeed, I had not tested this yet, it was meant just as a strawman. It indeed should remove itself from the `tasks` list, or even better, there should be no `tasks` (although it's unclear to me if `TimerTask`s can go to multiple `Timer`s).

axel22:
Indeed, it seems like reusing a `TimerTask` should throw an exception. That will simplify things.

Sorry that I made you review all of this prematurely - I intended it as a strawman, but was not explicit about it.

javalib/src/main/scala/java/util/Timer.scala
```diff
@@ -0,0 +1,115 @@
...
+      }
+      tasks.clear()
+    }
+  }
+
+  def purge(): Int = {
+    var count = 0
+    for (task <- tasks.toArray) {
+      if (task.cancelled) {
+        tasks.remove(task)
```
axel22: It's not a mutable set, it has been converted to an array - but, we're removing `tasks` anyway.

javalib/src/main/scala/java/util/Timer.scala
```diff
@@ -0,0 +1,115 @@
...
+  def schedule(task: TimerTask, firstTime: Date, period: Long): Unit = {
+    val delay = getMillisUntil(firstTime)
+    schedulePeriodically(task, delay, period)
+  }
+
+  private[util] def scheduleFixed(
+      task: TimerTask, delay: Long, period: Long): Unit = {
+    tasks += task
+    setTimeout(delay.millis) {
+      def loop(lastTime: Long) {
```
axel22: Done.

sjrd:
Jenkins, test this please

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/TimerTest.scala
```diff
@@ -0,0 +1,147 @@
...
+      var completed = false
+      val timer = new Timer
+      val task = new TimerTask {
+        def run(): Unit = {
+          completed = true
+        }
+      }
+      timer.schedule(task, 1000)
+      tick(500)
+      task.cancel()
```
sjrd: You could even `assertTrue(task.cancel())`

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/TimerTest.scala
```diff
@@ -0,0 +1,147 @@
...
+      assertFalse(task.cancel())
+
+      try {
+        timer.schedule(new TimerTask {
+          def run(): Unit = {}
+        }, 1000, 0)
+        assertTrue("Unreachable, negative delay.", false)
+      } catch {
+        case e: IllegalArgumentException =>
+      }
```
sjrd: `asserThrows`

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/TimerTest.scala
```diff
@@ -0,0 +1,147 @@
...
+      assertEquals(Seq(0, 1, 2, 3, 4, 5, 6, 7), seen)
+
+      try {
+        timer.schedule(new TimerTask {
+          def run(): Unit = {}
+        }, 1000, 100)
+        assertTrue("Unreachable, already canceled timer.", false)
+      } catch {
+        case e: IllegalStateException =>
+      }
```
sjrd: `assertThrows`

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/TimerTest.scala
axel22: Fixed.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/TimerTest.scala
axel22: Fixed.

javalib/src/main/scala/java/util/TimerTask.scala
```diff
@@ -0,0 +1,52 @@
...
+  def run(): Unit
+
+  def cancel(): Boolean = {
+    if (handle != null) {
+      clearTimeout(handle)
+      handle = null
+    }
+    if (canceled || owner == null || completed) {
+      canceled = true
+      false
```
axel22: Indeed, and good catch. Fixed, and added test cases.

axel22:
Comments should be addressed now - PTAL

sjrd:
Jenkins, retest this please

PR #2620, opened 2016-10-08
Small language cleanups related to JS native load specs

sjrd:


sjrd:
Review by @gzm0

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1090,25 +1119,73 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+            "(or extend js.GlobalScope).")
+      }
     }
-    if (sym.hasAnnotation(JSImportAnnotation)) {
-      reporter.error(pos,
-          "Non JS-native classes, traits and objects may not have an " +
-          "@JSImport annotation")
+  }
+
+  private def checkAndGetJSNativeLoadingSpecAnnotOf(pos: Position,
```
gzm0: pos is unused.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -694,6 +655,77 @@ abstract class PrepJSInterop extends plugins.PluginComponent
       enterOwner(kind) { super.transform(implDef) }
     }
 
+    private def checkAndComputeJSNativeLoadSpecOf(pos: Position,
```
gzm0: pos is (transitively) unused.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -694,6 +655,77 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+              "" // do not care because it does not compile anyway
+            }
+            annot.stringArg(1).fold {
+              JSNativeLoadSpec.Import(module, Nil)
+            } { pathName =>
+              val path = pathName.split('.').toList
+              JSNativeLoadSpec.Import(module, path)
+            }
+
+          case _ if loadSpecAnnot.forall(_.symbol == JSNameAnnotation) =>
```
gzm0: I'm not sure this is very clear and worth the effort just to not duplicate the creation of `JSNativeLoadSpec.Global(path)`. Consider:

```scala
def globalFromName = {
  val path = jsInterop.jsNameOf(sym).split('.').toList
  JSNativeLoadSpec.Global(path)
}

checkAndGetJSNativeLoadingSpecAnnotOf(sym) match {
  // snip

  case Some(annot) if annot.symbol == JSNameAnnotation =>
    globalFromName

  case None =>
    val needsExplicitJSName = ...
    // etc.

    globalFromName
```

Factor

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1090,25 +1119,73 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
-    if (sym.hasAnnotation(JSImportAnnotation)) {
-      reporter.error(pos,
-          "Non JS-native classes, traits and objects may not have an " +
-          "@JSImport annotation")
+  }
+
+  private def checkAndGetJSNativeLoadingSpecAnnotOf(pos: Position,
+      sym: Symbol): Option[Annotation] = {
+    val annots = sym.annotations.filter {
+      annot => JSNativeLoadingSpecAnnots.contains(annot.symbol)
```
gzm0: Is this new, that you write the argument on the new line? Examples in CODINGSTYLE are different.

sjrd: No, it's a mistake ;) `annot =>` should be on the previous line, indeed.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: It is used in the `if (needsExplicitJSName)` case.

sjrd:
Done for your other comments.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -539,90 +559,31 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+            annot <- sym.annotations
+            annotSym = annot.symbol
+            if JSNativeLoadingSpecAnnots.contains(annotSym)
+          } {
+            if (annotSym == JSNameAnnotation) {
+              reporter.warning(implDef.pos,
+                  "Traits should not have an @JSName annotation, as it does " +
+                  "not have any effect. This will be enforced in 1.0.")
+            } else {
+              reporter.error(implDef.pos,
```
gzm0: Actually, shouldn't you warn on the annotation here?

sjrd:
> Feel free to change the two positions at your own discretion.

I don't understand this. What are you talking about?

gzm0:
> I don't understand this. What are you talking about?

The error report positions I commented on (and I missed in my first review); you may change them to the annotations or leave them as they are and merge this PR to keep moving ahead as it is a minor thing.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: No, here we are left with the cases of `@JSGlobalScope` and `@JSImport`.

* `@JSGlobalScope` cannot be written in older versions at all, and its automatic additions for `extends js.GlobalScope` is also not a problem because a trait extending `js.GlobalScope` was already an error before.
* `@JSImport` was not checked in the previous *commit*, but there was no release where it existed and was not checked, so we're safe there as well.

sjrd: Oh! Duh! It just clicked what you're saying! I thought you were saying *warn* instead of *error*. But you're talking about the *position* primarily. Yes you are absolutely right, it should be on the annotation. And that also explains what you were saying about the "two positions" ^^

sjrd: Done.

sjrd:
retest this please

PR #2621, opened 2016-10-08
Remove Empty Tree

gzm0:


ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
```diff
@@ -154,11 +151,16 @@ object Hashers {
...
-          mixTag(TagTry)
+        case TryCatch(block, errVar, handler) =>
+          mixTag(TagTryCatch)
           mixTree(block)
           mixIdent(errVar)
           mixTree(handler)
+          mixType(tree.tpe)
+
+        case TryFinally(block, finalizer) =>
+          mixTag(TagTryCatch)
```
sjrd: `TagTryFinally`

ir/src/main/scala/org/scalajs/core/ir/Infos.scala
```diff
@@ -313,9 +312,10 @@ object Infos {
         .setEncodedName(propertyDef.name.name)
         .setIsExported(true)
 
-      // Any of getterBody and setterBody can be EmptyTree, but that's fine
-      traverse(propertyDef.getterBody)
-      traverse(propertyDef.setterBody)
+      propertyDef.getterBody.map(traverse)
+      propertyDef.setterArgAndBody map { case (_, body) =>
+        traverse(body)
+      }
```
sjrd: `map` -> `foreach` (twice)

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -154,9 +151,14 @@ object Serializers {
           writeByte(TagDoWhile)
           writeTree(body); writeTree(cond); writeOptIdent(label)
 
-        case Try(block, errVar, handler, finalizer) =>
-          writeByte(TagTry)
-          writeTree(block); writeIdent(errVar); writeTree(handler); writeTree(finalizer)
+        case TryCatch(block, errVar, handler) =>
+          writeByte(TagTryCatch)
+          writeTree(block); writeIdent(errVar); writeTree(handler);
```
sjrd: Trailing `;`

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -154,9 +151,14 @@ object Serializers {
...
+          writeType(tree.tpe)
+
+        case TryFinally(block, finalizer) =>
+          writeByte(TagTryFinally)
+          writeTree(block); writeTree(finalizer);
```
sjrd: Trailing `;`

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -634,11 +646,36 @@ object Serializers {
...
+        }
+      }
+    }
+
+    private def readTreeAllowEmpty(): Option[Tree] = {
+      require(useHacks068)
+      val pos = readPosition()
+      val tag = input.readByte()
+      if (tag == TagEmptyTree) None
+      else Some(readTreeFromTag(tag)(pos))
```
sjrd: I wonder whether it wouldn't be a good idea to keep this old encoding even for the new format. i.e., `TagEmptyTree` for `None` and other tags for `Some(_)`. This would save one byte every time.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -653,7 +690,29 @@ object Serializers {
...
+          val block = readTree()
+          val errVar = readIdent()
+          val handler = readTreeAllowEmpty()
+          val finalizer = readTreeAllowEmpty()
+          val tpe = readType()
+
+          val maybeCatch = handler.fold(block)(handler =>
+              TryCatch(block, errVar, handler)(tpe))
+          finalizer.fold(maybeCatch)(finalizer =>
+              TryFinally(maybeCatch, finalizer)(tpe))
```
sjrd: With `()`, the `handler =>` part should be on the next line, since it's one normal argument, that happens to be a lambda. (It's different from the `{}` case.)

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -799,8 +858,24 @@ object Serializers {
             result2
           }
         case TagPropertyDef =>
-          PropertyDef(readPropertyName(), readTree(),
-              readTree().asInstanceOf[ParamDef], readTree())
+          val name = readPropertyName()
+          val getterBody = readOptTree()
+          val setterArgAndBody = if (useHacks068) {
+            for {
+              setterArg <- readTreeAllowEmpty().map(_.asInstanceOf[ParamDef])
```
sjrd: `setterArg` could not be `EmptyTree` in the old format (as is clearly seen by the existing `.asInstanceOf[ParamDef]`) so you should use `readTree()` here.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -170,7 +165,10 @@ object Trees {
     val tpe = NoType // cannot be in expression position
   }
 
-  case class Try(block: Tree, errVar: Ident, handler: Tree, finalizer: Tree)(
+  case class TryCatch(block: Tree, errVar: Ident, handler: Tree)(
+      val tpe: Type)(implicit val pos: Position) extends Tree
+
+  case class TryFinally(block: Tree, finalizer: Tree)(
       val tpe: Type)(implicit val pos: Position) extends Tree
```
sjrd: A `TryFinally` does not need an explicit `val tpe`. It's `tpe` is always `block.tpe`.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -109,7 +109,7 @@ final class ClosureLinkerBackend(
 
     def exportName(tree: Tree): String = (tree: @unchecked) match {
       case MethodDef(_, StringLiteral(name), _, _, _) => name
-      case PropertyDef(StringLiteral(name), _, _, _)  => name
+      case PropertyDef(StringLiteral(name), _, _)  => name
```
sjrd: Keep the arrows aligned.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
```diff
@@ -161,10 +161,11 @@ object LinkedClass {
...
       // Normal methods
       case m: MethodDef if m.name.isInstanceOf[Ident] =>
-        if (m.body == EmptyTree)
-          abstractMethods += linkedMethod(m)
-        else
+        if (m.body.isDefined) {
           memberMethods += linkedMethod(m)
+        } else {
+          abstractMethods += linkedMethod(m)
+        }
```
sjrd: The braces are superfluous.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1180,34 +1177,26 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
-            }
+            val newHandler = pushLhsInto(newLhs, handler, tailPosLabels)
+            js.Try(newBlock, errVar, newHandler, js.EmptyTree)
+          }
+
+        case TryFinally(block, finalizer) =>
+          extractLet { newLhs =>
+            val newBlock = pushLhsInto(newLhs, block, tailPosLabels)
+            val newFinalizer = transformStat(finalizer, tailPosLabels)
+            js.Try(newBlock, js.Ident("dummy"), js.EmptyTree, newFinalizer)
```
sjrd: Should we file a follow up issue to try and get rid of `javascript.Trees.EmptyTree` as well?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -282,10 +282,12 @@ private[emitter] final class ScalaJSClassEmitter(
   /** Generates a method. */
   def genMethod(className: String, method: MethodDef)(
       implicit globalKnowledge: GlobalKnowledge): js.Tree = {
+    assert(method.body.isDefined, "Cannot generate an abstract method")
```
sjrd: Consider `val methodBody = method.body.getOrElse { throw ... }`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -374,31 +376,25 @@ private[emitter] final class ScalaJSClassEmitter(
...
+    val optSetter = property.setterArgAndBody map { case (arg, body) =>
+      val fun = desugarToFunction(this, className, arg :: Nil,
+          body, isStat = true)
+      js.StringLiteral("set") -> fun
     }
 
+    // Options passed to the defineProperty method
+    val descriptor = js.ObjectConstr(
+      List(js.StringLiteral("enumerable") -> js.BooleanLiteral(true)) ++
+      optGetter ++ optSetter
```
sjrd: Same as somewhere else. I suggest not to change the order of the fields.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -312,41 +316,47 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
           "but must be NoType")
     }
 
-    val bodyStats = body match {
-      case Block(stats) => stats
-      case _            => body :: Nil
-    }
+    body match {
+      case None =>
```
sjrd: Any reason this is not a `fold`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -230,12 +230,13 @@ final class BaseLinker(semantics: Semantics, esLevel: ESLevel, considerPositions
...
+          if (m.name.isInstanceOf[StringLiteral]) {
             exportedMembers += linkedMethod(m)
-          else if (m.body == EmptyTree)
-            abstractMethods += linkedMethod(m)
-          else
+          } else if (m.body.isDefined) {
             memberMethods += linkedMethod(m)
+          } else {
+            abstractMethods += linkedMethod(m)
+          }
```
sjrd: The braces are superfluous.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -438,12 +442,12 @@ private[optimizer] abstract class OptimizerCore(
           case _                     => DoWhile(newBody, newCond, None)
         }
 
-      case Try(block, errVar, EmptyTree, finalizer) =>
+      case TryFinally(block, finalizer) =>
         val newBlock = transform(block, isStat)
         val newFinalizer = transformStat(finalizer)
-        Try(newBlock, errVar, EmptyTree, newFinalizer)(newBlock.tpe)
+        TryFinally(newBlock, newFinalizer)(newBlock.tpe)
```
sjrd: Consider moving the `TryFinally` case after the `TryCatch` case, so that we maintain the same order everywhere.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1115,32 +1113,32 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
               js.Return(transformExpr(rhs))
-            case Return(_, Some(l)) =>
+            case Lhs.Return(Some(l)) =>
               doReturnToLabel(l)
           }
 
         // Almost base case with RecordValue
 
         case RecordValue(recTpe, elems) =>
           (lhs: @unchecked) match {
```
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2138,16 +2135,26 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
     }
   }
 
+  // A left hand side that can be pushed into a right hand side tree.
+  sealed trait Lhs
```
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2138,16 +2135,26 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
+
+  object Lhs {
+    case class Assign(lhs: Tree) extends Lhs
+    case class VarDef(name: Ident, tpe: Type, mutable: Boolean) extends Lhs
+    case class Return(label: Option[Ident]) extends Lhs
+    case object Empty extends Lhs
```
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4040,17 +4038,36 @@ private[optimizer] object OptimizerCore {
     val Nothing = RefinedType(NothingType)
   }
 
-  private class AllocationSite(private val node: Tree) {
-    override def equals(that: Any): Boolean = that match {
-      case that: AllocationSite => this.node eq that.node
-      case _                    => false
+  /** Originating tree of an operation (currently only inlining).
```
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4040,17 +4038,36 @@ private[optimizer] object OptimizerCore {
...
+   *
+   *  This is used to allow for recursive inlining of the same method multiple
+   *  times while preventing an infite inline loop.
+   *
+   *  Note: Using an anonymous allocation site is always OK, since it will over
+   *  approximate when inlining is not possible anymore: Two anonymous sites
```
gzm0: Done

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4040,17 +4038,36 @@ private[optimizer] object OptimizerCore {
...
+   *  could be the same but are not necessarily. However, they compare equal.
+   */
+  private sealed trait AllocationSite
```
gzm0: Done

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4040,17 +4038,36 @@ private[optimizer] object OptimizerCore {
...
+      override def toString(): String = "AllocationSite(<anonymous>)"
     }
 
-    override def hashCode(): Int =
-      System.identityHashCode(node)
+    def Tree(tree: Tree): AllocationSite = new TreeAllocationSite(tree)
 
-    override def toString(): String =
-      s"AllocationSite($node)"
+    private class TreeAllocationSite(val node: Tree) extends AllocationSite {
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2665,8 +2665,10 @@ abstract class GenJSCode extends plugins.PluginComponent
         }
       }
 
+      assert(elseClause.isDefined, "No elseClause in pattern match")
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -606,7 +606,8 @@ abstract class GenJSCode extends plugins.PluginComponent
         case mdef: js.MethodDef =>
           implicit val pos = mdef.pos
           val name = mdef.name.asInstanceOf[js.StringLiteral]
-          val impl = lambda(mdef.args, mdef.body)
+          assert(mdef.body.isDefined, "Got anon SJS class with abstract method")
+          val impl = lambda(mdef.args, mdef.body.get)
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -618,15 +619,19 @@ abstract class GenJSCode extends plugins.PluginComponent
...
             List(js.StringLiteral(name) -> value)
 
-          def accessor(tree: js.Tree, name: String)(genBody: js.Tree => js.Tree) =
-            if (tree == js.EmptyTree) Nil
-            else field(name, genBody(tree))
+          val optGetter = pdef.getterBody map { body =>
+            js.StringLiteral("get") -> lambda(params = Nil, body)
+          }
+
+          val optSetter = pdef.setterArgAndBody map { case (arg,body) =>
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -618,15 +619,19 @@ abstract class GenJSCode extends plugins.PluginComponent
...
 
           val descriptor = js.JSObjectConstr(
-              accessor(pdef.getterBody, "get")(lambda(params = Nil, _)) ++
-              accessor(pdef.setterBody, "set")(lambda(pdef.setterArg :: Nil, _)) ++
-              field("configurable", js.BooleanLiteral(true)) ++
-              field("enumerable", js.BooleanLiteral(true))
+              List(
+                  js.StringLiteral("configurable") -> js.BooleanLiteral(true),
+                  js.StringLiteral("enumerable") -> js.BooleanLiteral(true)
+              ) ++ optGetter ++ optSetter
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -662,7 +667,8 @@ abstract class GenJSCode extends plugins.PluginComponent
           case tree =>
             super.transform(tree, isStat)
         }
-      }.transform(ctorBody, isStat = true)
+      }.transform(ctorBody.getOrElse(
+          throw new AssertionError("Constructor is abstract")), isStat = true)
```
gzm0: Done

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1398,8 +1405,8 @@ abstract class GenJSCode extends plugins.PluginComponent
             super.transform(tree, isStat)
         }
       }
-      val newBody =
-        transformer.transform(body, isStat = resultType == jstpe.NoType)
+      val newBody = body.map(transformer.transform(_,
+          isStat = resultType == jstpe.NoType))
```
gzm0: Done

ir/src/main/scala/org/scalajs/core/ir/Hashers.scala
gzm0: Done

ir/src/main/scala/org/scalajs/core/ir/Infos.scala
gzm0: Done

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
gzm0: What about the position? The problem with reading a tree, is that we need to read a position first. We could (should?) inverse tree tags and positions, but I'm not sure this is a good thing to do in a minor IR version (although possible).

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
gzm0: Done

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
gzm0: Done

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Done

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: Done

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
gzm0: Hmmm... I thought ifs with side-effects must have braces. Can't find this rule in the style guide though...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
gzm0: #2622

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
gzm0: Done

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
gzm0: Done.
Also, why is this only enumerable, where in the compiler we emit configurable as well?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
gzm0: Not really. Short of side-effectness... Style guide actually says no use of fold for side effects... Although we do...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
gzm0: Same thing about side effects.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Done

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: Hum good point. I think the intention in the style guide is that it applies to folds on *collections*. `Option.fold` has been treated differently in our codebase.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
sjrd: I don't recall having applied such a rule. For example, it is very common for us to use a `throw` then part without braces.

Maybe you're confusing it with the rule about if/else in *statement* position, which demand that their branches always be on a dedicated line. In other words, no two-liner if/else in statement position.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportASTTest.scala
```diff
@@ -24,7 +24,7 @@ class JSExportASTTest extends JSASTTest {
       override def foo = 2
     }
     """.hasExactly(1, "definitions of property `foo`") {
-      case js.PropertyDef(js.StringLiteral("foo"), _, _, _) =>
+      case js.PropertyDef(js.StringLiteral("foo"), _, _) =>
```
sjrd: This change is in the wrong commit.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
sjrd: It is indeed unfortunate that the format has the position before the kind. I think we could just hard-code the encoding for `NoPosition` (i.e., `-1.toByte`) when serializing `None`. In 1.0, we can flip the position and kind and get rid of this. WDYT?

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -170,9 +165,14 @@ object Trees {
     val tpe = NoType // cannot be in expression position
   }
 
-  case class Try(block: Tree, errVar: Ident, handler: Tree, finalizer: Tree)(
+  case class TryCatch(block: Tree, errVar: Ident, handler: Tree)(
       val tpe: Type)(implicit val pos: Position) extends Tree
 
+  case class TryFinally(block: Tree, finalizer: Tree)(
+    implicit val pos: Position) extends Tree {
```
sjrd: 2 more spaces here.

sjrd:
You will most likely need to filter out a bunch of binary incompatibilities ;)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
sjrd: I believe this is a bug. ES6 `get()` and `set()` operators define configurable properties ([spec](http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions-runtime-semantics-propertydefinitionevaluation)), so that's what we should do too. Filed as #2624.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
gzm0: Done. What do you want to do about `PropertyDef`? Leave it like it is?

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedClass.scala
gzm0: Done.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
gzm0: Done.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
sjrd: Hum, good question. Let's leave it as is. There aren't that many `PropertyDef`s. We'll be able to revisit in 1.0 if we wish.

One thing we could do would be to write/read an `Option[ParamDef]` for the setter argument, and only write/read a `Tree` for the body if the option is `Some(_)`.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -461,6 +467,11 @@ object Serializers {
       trees.foreach(writeTree)
     }
 
+    def writeOptTree(optTree: Option[Tree]): Unit = {
+      writePosition(Position.NoPosition)
+      optTree.fold(buffer.writeByte(TagEmptyTree))(writeTree)
```
gzm0: Done

gzm0:
There is a strange, deterministic CI failure on 2.12 with `toolsJS/test`:
https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/146516/console

sjrd:
Can the failure be reproduced locally? It would be good to be able to decipher that stack trace in terms of .scala files.

sjrd:
This part:

> Emitter: Method tree cache stats: resued: 0 -- invalidated: 1

is actually pretty encouraging in terms of debugging, because it shows that this fails while emitting the very first method that is encountered. It means it is very likely that it fails in this way for *all* methods.

Or at least, for all *default* methods, which end up being the ones first emitted. And indeed, the stack trace shows `genDefaultMethod` at the bottom. Given that this only happens on 2.12, this is quite strong evidence that the problem is related to emitting default methods.

gzm0:
Investigating.

gzm0:
There seems to be a problem with outer pointer checks in 2.12. See #2625.

gzm0:
I have moved the nested classes outside of `JSDesugaring`. I don't know why they were in there in the first place.

gzm0:
@sjrd If you have a spare moment, this should be ready for final review.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -582,8 +582,8 @@ abstract class GenJSCode extends plugins.PluginComponent
       generatedClasses += ((sym, None, newClassDef))
 
       // Construct inline class definition
-      val js.MethodDef(_, _, ctorParams, _, ctorBody) = constructor.getOrElse(
-          throw new AssertionError("No ctor found"));
+      val js.MethodDef(_, _, ctorParams, _, Some(ctorBody)) =
+          constructor.getOrElse(throw new AssertionError("No ctor found"))
```
sjrd: 2 spaces

gzm0: Done.

sjrd:
The build went awry and we can't access the logs, but I guess there are binary incompat filters missing in the last commit.

sjrd:

gzm0:
Updated

PR #2628, opened 2016-10-12
Fix #2625: Fix clash of outer pointers in 2.12.0-RC2 onwards.

### LINKED ISSUE #2382: Name clash for $outer pointers of two different nesting levels
As minimized by @nicolasstucki:

``` scala
  def assertEquals(expect: Int, actual: Int, field: String): Unit =
    assert(expect == actual, s"Expected $field to be <$expect> but was <$actual>")

  def main() {
    val b = new B { }
    val y = new b.Y
    val z = new y.Z

    assertEquals(1, b.a, "b.a")
    assertEquals(1, z.z1, "z.z1")
    assertEquals(2, y.y, "y.y")

    assertEquals(2, z.z2, "z.z2")
    //  assertEquals(2, z.z3, "z.z3") // for variant

  }
}

trait A {
  val a: Int = 1
  class X {
    def x: Int = a
  }
}

trait B extends A {
  class Y {
    def y: Int = 2
    class Z extends X {
      def z1: Int = a
      def z2: Int = x
    }
  }
}

//// Variant that also fails
//trait B extends A {
//  class Y {
//    def y: Int = 2
//    class Z extends X {
//      def z1: Int = a
//      def z3: Int = y
//    }
//  }
//}
```

### LINKED ISSUE #2625: Outer pointer checks fail in 2.12
Take the following:

```scala
class A {
  def call() = f(Foo.B(1))

  private def f(x: Foo) = x match {
    case Foo.B(x) => println(x)
  }

  sealed abstract class Foo

  object Foo {
    case class B(x: Int) extends Foo
  }
}

  def main() {
    (new A).call()
  }
}
```

This will fail with a `MatchError` in 2.12.0-RC1, but works as expected (prints 1) in 2.11.8.

sjrd:
The root cause of the bug #2625 which surfaces in 2.12.0 is actually #2382. We cannot fix #2382 without breaking backwards binary compatibility in general. However, we can fix it only for 2.12.0-RC2 onwards, as RCs of Scala (and the upcoming final) have separate binary ecosystems anyway.

So this commit fixes #2382 but only for 2.12.0-RC2 onwards. The fix is trivial: name mangle all outer pointers in the same way as if they were private, i.e., with the number of class ancestors of its owner.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
@@ -86,6 +86,12 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
         refClass.values ++ volatileRefClass.values)
   }
 
+  /** See comment in `encodeFieldSym()`. */
+  private lazy val shouldMangleOuterPointerName = {
+    val v = scala.util.Properties.versionNumberString
+    !(v.startsWith("2.10.") || v.startsWith("2.11.") || v == "2.12.0-RC1")
+  }
+
   def encodeFieldSym(sym: Symbol)(implicit pos: Position): js.Ident = {
     require(sym.owner.isClass && sym.isTerm && !sym.isMethod && !sym.isModule,
         "encodeFieldSym called with non-field symbol: " + sym)
@@ -99,12 +105,29 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
      * because they are emitted as private by our .scala source files, but
      * they are considered public at use site since their symbols come from
      * Java-emitted .class files.
+     *
+     * Starting with 2.12.0-RC2, we also special case outer fields. This
+     * essentially fixes #2382, which is caused by a class having various $outer
+     * pointers in its hierarchy that points to different outer instances.
+     * Without this fix, they all collapse to the same field in the IR. We
+     * cannot fix this for all Scala versions at the moment, because that would
+     * break backwards binary compatibility. We *do* fix it for 2.12.0-RC2
+     * onwards because that also fixes #2625, which surfaced in 2.12 and is
+     * therefore a regression. We can do this because the 2.12 ecosystem is
+     * not binary compatible anyway (because of Scala) so we can break it on
+     * our side at the same time.
      */
-    val idSuffix =
-      if (sym.isPrivate || allRefClasses.contains(sym.owner))
+    val idSuffix: String = {
+      val usePerClassSuffix = {
+        sym.isPrivate ||
+        allRefClasses.contains(sym.owner) ||
+        (shouldMangleOuterPointerName && sym.isOuterField)
+      }
+      if (usePerClassSuffix)
         sym.owner.ancestors.count(!_.isTraitOrInterface).toString
       else
         "f"
+    }
 
     val encodedName = name + "$" + idSuffix
     js.Ident(mangleJSName(encodedName), Some(sym.unexpandedName.decoded))
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1409,6 +1409,31 @@ object Build {
           IO.write(outFile,
               replaced.replace("@Test def workTest(): Unit = sys.error(\"stubs\")", unitTests))
           Seq(outFile)
+        },
+
+        // Exclude tests based on version-dependent bugs
+        sources in Test := {
+          val sourceFiles = (sources in Test).value
+          val v = scalaVersion.value
+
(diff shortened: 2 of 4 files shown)


sjrd:
Review by @gzm0

sjrd:
I tested this locally with ++2.12.0-1e81a09-SNAPSHOT, built from today's tip of scala/scala's `2.12.0` branch. Commit hash: https://github.com/scala/scala/commit/1e81a09a896a0d7497687b5df1b8220172eaec92

compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
```diff
@@ -86,6 +86,12 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
         refClass.values ++ volatileRefClass.values)
   }
 
+  /** See comment in `encodeFieldSym()`. */
+  private lazy val applyFixFor2382 = {
```
gzm0: Could we give this a more descriptive name? As in "mangleOuterPointerName" or something?

compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
```diff
@@ -99,12 +105,26 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
...
+     * not binary compatible anyway (because of Scala) so we can break it on
+     * our side at the same time.
      */
-    val idSuffix =
-      if (sym.isPrivate || allRefClasses.contains(sym.owner))
+    val idSuffix: String = {
+      if (applyFixFor2382 && sym.isOuterField)
+        encodeClassFullName(sym.owner)
+      else if (sym.isPrivate || allRefClasses.contains(sym.owner))
         sym.owner.ancestors.count(!_.isTraitOrInterface).toString
```
gzm0: I do not even understand why we need to special case this? Is the Scalac backend special casing this? Why are field overrides with the same name allowed if they are outer fields?

gzm0: Oh, and also: Why are we not mangling with the superclass count here?

sjrd: Well ... the problem is that scalac generates *protected* fields for `$outer`. And that's basically the only way to get scalac to ever generate a protected field. We see that after mixin:
```scala
scala> class Foo { class Bar }
[[syntax trees at end of                     mixin]] // <console>
package $line3 {
  ...
  class iw$Foo$Bar extends Object {
    <synthetic> <paramaccessor> <artifact> protected val $outer: $line3.iw$Foo = _;
    <synthetic> <stable> <artifact> def $outer(): $line3.iw$Foo = iw$Foo$Bar.this.$outer;
    def <init>($outer: $line3.iw$Foo): $line3.iw$Foo$Bar = {
      if ($outer.eq(null))
        throw null
      else
        iw$Foo$Bar.this.$outer = $outer;
      iw$Foo$Bar.super.<init>();
      ()
    }
  }
}
```
And apparently, the JVM is fine with this. It seems to disambiguate fields by their enclosing class. Field selection opcodes reference an enclosing class in addition to a field name (and actually the field *type* as well, I've been told; so you can *overload* a field!!), which allows it to disambiguate outer fields correctly. According to a small test with `javac`, I figured that it also considers "inherited" fields (so if it says `B.x` but `x` is defined in `A` that `B` extends, it will find it).

Alternatively to `sym.isOuterField`, we could probably do `sym.isProtected`. I wager it would identify the exact same set of symbols.

We could be much more extreme, and do this for *all* symbols, including public fields. That would be valid if moving a field one class up the hierarchy is considered a binary *incompatible* change. But that's not true on the JVM according to my experiments. It would be OK for all Scala-generated code, though, since only synthetic fields are ever non-private. So maybe we *could* declare it binary incompatible for .sjsir.

gzm0: Ok, so let's keep `isOuterField` and the superclass count for now. Maybe we should revisit fields entirely in 1.0.0. It seems like they are accumulating a lot of things.

sjrd: Agreed. Filed the fields rethinking as #2629.

sjrd:
Updated.

sjrd:
The commit message was obsolete; it was still saying mangling with the class' full name. The new commit is only a fix to the commit message.

PR #2633, opened 2016-10-15
Fix #2622: Remove js.EmtpyTree

### LINKED ISSUE #2619: Emit TryFinally(TryCatch(...)) as single try..catch..finally
We used not to do this because of #563 / google/closure-compiler#186. This seems fixed, so we should emit it as joint block if we can.

### LINKED ISSUE #2622: Remove js.EmptyTree

gzm0:
- In VarDef and Let, replace by Optional[Tree]
- Split Try into TryCatch and TryFinally
- In Switch, replaced by Skip

As a side effect, this fixes #2619.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
@@ -1010,7 +1010,7 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
           })
 
         case _ =>
-          genLet(ident, mutable = true, js.EmptyTree)
+          genEmptyMutableLet(ident)
       }
     }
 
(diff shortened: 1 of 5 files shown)


gzm0:
CI only so far. Making sure try ... catch ... finally issue is really fixed.

gzm0:
Review by @sjrd

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2133,6 +2125,13 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
 
   private[emitter] def genLet(name: js.Ident, mutable: Boolean, rhs: js.Tree)(
       implicit outputMode: OutputMode, pos: Position): js.LocalDef = {
+    genLet(name, mutable, Some(rhs))
+  }
+
+  private[emitter] def genLet(name: js.Ident, mutable: Boolean,
+      rhs: Option[js.Tree])(implicit outputMode: OutputMode,
+      pos: Position): js.LocalDef = {
+    require(rhs.isDefined || mutable, "Unset constant is meaningless")
```
sjrd: I have a feeling this invariant can be encoded in the types. Instead of this signature, use instead:
```scala
private[emitter] genEmptyMutableLet(name: js.Ident)(implicit ...)
```
It seems that all call sites statically know whether their `rhs` is empty or not. For the empty case, it further seems that the only call site is https://github.com/gzm0/scala-js/blob/f58bc89165cc9a63dfe382abd0757325c6aea977/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala#L1013 where `mutable = true`.

sjrd:
Also, there's a typo in the commit message: `Optional[Tree]` -> `Option[Tree]`.

gzm0:
Updated

PR #2634, opened 2016-10-15
Fix PropertyDefs properties

### LINKED ISSUE #2624: PropertyDefs should be emitted with configurable: true in ES5
See https://github.com/scala-js/scala-js/pull/2621#discussion_r82519863

gzm0:
Also add tests for enumerability.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
@@ -198,6 +198,16 @@ class ExportsTest {
     assertEquals(7, bar.y)
   }
 
+  @Test def properties_are_not_enumerable(): Unit = {
+    class Foo {
+      @JSExport
+      def myProp: Int = 1
+    }
+
+    val x: js.Any = (new Foo()).asInstanceOf[js.Any]
+    assertFalse(js.Object.properties(x).contains("myProp"))
+  }
+
   @Test def overloaded_exports_for_methods(): Unit = {
     class Foo {
       @JSExport("foobar")
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
@@ -565,6 +565,24 @@ class ScalaJSDefinedTest {
     assertEquals(100, dyn.foo())
   }
 
+  @Test def properties_are_not_enumerable(): Unit = {
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      def myProp: Int = 1
+    }
+
+    val x: js.Any = (new Foo()).asInstanceOf[js.Any]
+    assertFalse(js.Object.properties(x).contains("myProp"))
+
+    // Anonymous classes
+    val y = new js.Object {
+      def myProp: Int = 1
+    }
+
+    assertFalse(js.Object.properties(y).contains("myProp"))
+  }
+
   @Test def properties_are_configurable(): Unit = {
     // Named classes
     @ScalaJSDefined
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
@@ -565,6 +565,32 @@ class ScalaJSDefinedTest {
     assertEquals(100, dyn.foo())
   }
 
+  @Test def properties_are_configurable(): Unit = {
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      def myProp: Int = 1
+    }
+
+    // Delete property from prototype.
+    val prototype = js.constructorOf[Foo].prototype
+    prototype.asInstanceOf[js.Dictionary[js.Any]].delete("myProp")
+
+    // Check it is actually gone.
+    assertTrue(js.isUndefined((new Foo()).asInstanceOf[js.Dynamic].myProp))
+
+    // Anonymous classes
+    val y = new js.Object {
+      def myProp: Int = 1
+    }
+
+    // The property should be on the instance itself.
+    assertTrue(y.hasOwnProperty("myProp"))
+    y.asInstanceOf[js.Dictionary[js.Any]].delete("myProp")
+    assertTrue(js.isUndefined(y.asInstanceOf[js.Dynamic].myProp))
+    assertFalse(y.hasOwnProperty("myProp"))
+  }
+
   @Test def properties_with_explicit_name(): Unit = {
     @ScalaJSDefined
     class PropertiesWithExplicitName extends js.Object {
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -629,10 +629,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           val descriptor = js.JSObjectConstr(
               optGetter.toList ++
               optSetter ++
-              List(
-                  js.StringLiteral("configurable") -> js.BooleanLiteral(true),
-                  js.StringLiteral("enumerable") -> js.BooleanLiteral(true)
-              )
+              List(js.StringLiteral("configurable") -> js.BooleanLiteral(true))
           )
 
           js.JSBracketMethodApply(jsObject, js.StringLiteral("defineProperty"),
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
@@ -381,10 +381,7 @@ private[emitter] final class ScalaJSClassEmitter(
     val descriptor = js.ObjectConstr(
       optGetter.toList ++
       optSetter ++
-      List(
-          js.StringLiteral("configurable") -> js.BooleanLiteral(true),
-          js.StringLiteral("enumerable") -> js.BooleanLiteral(true)
-      )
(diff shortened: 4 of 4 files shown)


gzm0:
Review by @sjrd

gzm0:
It seems like ES6 properties are not enumerable on Node.js v.5.3.0:

```javascript
class A {
  get x() { return 1; }
}

const a = new A();
console.log(a.x)  // --> 1
for (let i in a) {
  console.log(i);
}  // no output.
```

The same happens on Chrome 53.0.2785.143.

sjrd:
> It seems like ES6 properties are not enumerable on Node.js v.5.3.0:

Ah, yes. I think that's spec'ed. IIRC I saw a message about that in es-discuss not very long ago.

gzm0:
> Ah, yes. I think that's spec'ed. IIRC I saw a message about that in es-discuss not very long ago.

So this means we are emitting properties wrong in ES5? The property defs we generate should therefore not be enumerable?

sjrd:
Yes and no. They should be non-enumerable to mimic the ES6 behavior, yes. But we don't actually promise that our ES5 output is semantically equivalent to the ES6 output. Implementing that would kill our ES5 performances (e.g., we would have to add a lot of logic to classes to reproduce ES6 behavior, and even then we couldn't get the job done 100%, as no matter what we do, an ES5 "class" cannot extend an ES6 class).

So we are at liberty to *choose* what we want to do about this. The rationale should be to mimic the ES6 behavior, unless doing so has a performance cost. Making this configurable+enumerable doesn't have a performance cost, because VMs like configurable+enumerable properties better. But removing enumerability could have a performance cost. At least for normal fields and methods, VMs are known to prefer enumerable+configurable. On the other hand, getters and setters are not well optimized to begin with, so probably making them non-enumerable doesn't make them any worse.

Given the total absence of benchmarks to support any claim, and the fact that they were previously non configurable, I would be inclined to make them now non enumerable. At worst, configurable+non enumerable shouldn't be any worse than non configurable+enumerable.

gzm0:
Updated

gzm0:
Updated.

sjrd:
I just realized that `js.delete` doesn't test anything meaningful in the case where the property is in the prototype chain (so every case except anonymous JS class):
```js
C:\Users\Sepi>node --use-strict
> function F() {}
undefined
> Object.defineProperty(F.prototype, "x", {get: function() { return 42; }, enumerable: false, configurable: true});
F {}
> var f = new F()

> f.x
42
> f.x = 4
TypeError: Cannot set property x of #<F> which has only a getter
    at repl:1:5
    at sigintHandlersWrap (vm.js:32:31)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInContext (vm.js:31:12)
    at REPLServer.defaultEval (repl.js:308:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:477:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:188:7)
> delete f.x
true
> f.x
42
> delete F.prototype.x

> f.x

>
```
So in the named Scala.js-defined JS class, what we need to test is that we can remove the property on the prototype; not on the object itself.
For good measure, it would be a good idea to test that accessing the property does return `undefined` afterwards.

Also, I realized that the test of configurability for *exports* should not exist. It is undefined behavior to mess with the prototype chain of Scala classes. Read/write/call exports is fine. Deleting exports, adding fields/methods, etc. is UB. In fact, theoretically speaking, all Scala classes are frozen, and all their instances are sealed (although we don't actually enforce this due to performance concerns).

gzm0:
Updated (although not fully tested). Let's see :)

sjrd:
```
[error] Test org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest.properties_are_configurable failed: java.lang.AssertionError: null, took 0.001000192 sec
[error]     at org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest.properties$undare$undconfigurable(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/js/target/scala-2.10/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/junit-runtime/src/main/scala/org/junit/Assert.scala:18)
[error]     at org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest$scalajs$junit$bootstrapper$.invoke(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/js/target/scala-2.10/scalajs-test-suite-test-fastopt.js:199297)
[error]     ... 3 more
[info] Test org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest.properties_are_not_enumerable started
[error] Test org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest.properties_are_not_enumerable failed: java.lang.AssertionError: null, took 0.000999936 sec
[error]     at org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest.properties$undare$undnot$undenumerable(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/js/target/scala-2.10/file:/localhome/jenkins/a/workspace/scalajs-task-worker_2/junit-runtime/src/main/scala/org/junit/Assert.scala:24)
[error]     at org.scalajs.testsuite.jsinterop.ScalaJSDefinedTest$scalajs$junit$bootstrapper$.invoke(/localhome/jenkins/a/workspace/scalajs-task-worker_2/test-suite/js/target/scala-2.10/scalajs-test-suite-test-fastopt.js:199299)
[error]     ... 2 more
```

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,32 @@ class ScalaJSDefinedTest {
...
 
+  @Test def properties_are_configurable(): Unit = {
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      var myProp: Int = 1
+    }
+
+    // Delete property from prototype.
+    val prototype = js.constructorOf[Foo].asInstanceOf[js.Dynamic].prototype
```
sjrd: This cast is unnecessary. `js.constructorOf` returns a `js.Dynamic`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,32 @@ class ScalaJSDefinedTest {
     assertEquals(100, dyn.foo())
   }
 
+  @Test def properties_are_configurable(): Unit = {
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      var myProp: Int = 1
```
sjrd: This creates an actual *field*, not a property. You should have a `def` without `()` instead to create an actual getter.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,32 @@ class ScalaJSDefinedTest {
...
+    // Delete property from prototype.
+    val prototype = js.constructorOf[Foo].asInstanceOf[js.Dynamic].prototype
+    prototype.asInstanceOf[js.Dictionary[js.Any]].delete("myProp")
+
+    // Check it is actually gone.
+    assertTrue(js.isUndefined((new Foo()).asInstanceOf[js.Dynamic].myProp))
+
+    // Anonymous classes
+    val y = new js.Object {
+      var myProp: Int = 1
```
sjrd: Same here: `def`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,24 @@ class ScalaJSDefinedTest {
     assertEquals(100, dyn.foo())
   }
 
+  @Test def properties_are_not_enumerable(): Unit = {
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      var myProp: Int = 1
```
sjrd: `def`

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -198,6 +198,16 @@ class ExportsTest {
     assertEquals(7, bar.y)
   }
 
+  @Test def properties_are_not_enumerable(): Unit = {
+    class Foo {
+      @JSExport
+      var myProp: Int = 1
```
sjrd: Maybe for consistency with the Scala.js-defined JS tests, use a `def` here as well. Though for Exports even `val`s and `var`s are properties, so this does not change anything here.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,24 @@ class ScalaJSDefinedTest {
...
+    }
+
+    val x: js.Any = (new Foo()).asInstanceOf[js.Any]
+    assertFalse(js.Object.properties(x).contains("myProp"))
+
+    // Anonymous classes
+    val y = new js.Object {
+      var myProp: Int = 1
```
sjrd: `def`

gzm0:
Updated.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +565,24 @@ class ScalaJSDefinedTest {
...
+
+    val x: js.Any = (new Foo()).asInstanceOf[js.Any]
+    assertFalse(js.Object.properties(x).contains("myProp"))
+
+    // Anonymous classes
+    val y = new js.Object {
+      def myProp: Int = 1
+    }
+
+    assertFalse(js.Object.properties(x).contains("myProp"))
```
sjrd: Ouch. This should be `y`. I almost missed it.

gzm0:
Updated.

PR #2644, opened 2016-11-06
Fix #1381: Add top-level exports

### LINKED ISSUE #1381: Add a way to export a top-level method/value
Some APIs (at least the Google maps API) need the name of a callback function as a get parameter. They apparently do not allow calling a method to get the callback (see [SO Post](http://stackoverflow.com/questions/27440235/using-scala-js-method-as-callback)).

This is a problem for Scala.js interop, since it basically requires adding a tiny bit of JS wrapper code or abusing a module initializer.

gzm0:
The new annotation @JSExportTopLevel allows modules to export their
methods to the top-level of the export scope, so they can be called
directly. For example:

    object MyApp {
      @JSExportTopLevel("main")
      def main(): Unit = ???
    }

Assuming that the export namespace is the `window` object, `MyApp.main`
can now be invoked as follows in JavaScript:

    window.main()

As opposed to (if one assumes `MyApp` and `main` to be exported):

    window.MyApp().main()

Overloading is supported from within the same module. Exporting to the
same name (irrespective of the export type) from different modules will
cause a link time error.

Current restrictions of top-level exports:

- Properties (or fields) cannot be exported.
- The fully qualified name of the export needs to be given.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
@@ -25,8 +25,11 @@ trait PrepJSExports { this: PrepJSInterop =>
       jsName: String,
       pos: Position,
       isNamed: Boolean,
+      isTopLevel: Boolean,
       ignoreInvalid: Boolean
-  ) extends jsInterop.ExportInfo
+  ) extends jsInterop.ExportInfo {
+    assert(!isNamed || !isTopLevel)
+  }
 
   /** Generate the exporter for the given DefDef
    *  or ValDef (abstract val in class, val in trait or lazy val;
@@ -87,8 +90,15 @@ trait PrepJSExports { this: PrepJSInterop =>
       // Reset interface flag: Any trait will contain non-empty methods
       clsSym.resetFlag(Flags.INTERFACE)
 
+      val (topLevelExports, otherExports) = exports.partition(_.isTopLevel)
+
+      /* We can handle top level exports entirely in the backend. So just
+       * register them here.
+       */
+      jsInterop.registerForExport(baseSym, topLevelExports)
+
       // Actually generate exporter methods
-      exports.flatMap { exp =>
+      otherExports.flatMap { exp =>
         if (exp.isNamed)
           genNamedExport(baseSym, exp.jsName, exp.pos) :: Nil
         else
@@ -167,11 +177,13 @@ trait PrepJSExports { this: PrepJSInterop =>
   def exportsOf(sym: Symbol): List[ExportInfo] = {
     val exports = directExportsOf(sym) ++ inheritedExportsOf(sym)
 
-    // Calculate the distinct exports for this symbol (eliminate double
-    // occurrences of (name, isNamed) pairs).
-    val grouped = exports.groupBy(exp => (exp.jsName, exp.isNamed))
+    /* Calculate the distinct exports for this symbol (eliminate double
+     * occurrences of (name, isNamed, isTopLevel) tuples).
+     */
+    val grouped = exports.groupBy(
+        exp => (exp.jsName, exp.isNamed, exp.isTopLevel))
 
-    for (((jsName, isNamed), exps) <- grouped.toList)
+    for ((_, exps) <- grouped.toList)
       // Make sure that we are strict if necessary
       yield exps.find(!_.ignoreInvalid).getOrElse(exps.head)
   }
@@ -192,11 +204,8 @@ trait PrepJSExports { this: PrepJSInterop =>
     }
 
     // Annotations that are directly on the member
-    val directAnnots = for {
-      annot <- trgSym.annotations
-      if annot.symbol == JSExportAnnotation ||
-         annot.symbol == JSExportNamedAnnotation
-    } yield annot
+    val directAnnots = trgSym.annotations.filter(
+        annot => isDirectMemberAnnot(annot.symbol))
 
     // Is this a member export (i.e. not a class or module export)?
     val isMember = sym.isMethod && !sym.isConstructor
@@ -214,6 +223,7 @@ trait PrepJSExports { this: PrepJSInterop =>
     } yield {
       val isNamedExport = annot.symbol == JSExportNamedAnnotation
       val isExportAll = annot.symbol == JSExportAllAnnotation
+      val isTopLevelExport = annot.symbol == JSExportTopLevelAnnotation
       val hasExplicitName = annot.args.nonEmpty
 
       def explicitName = annot.stringArg(0).getOrElse {
@@ -244,7 +254,7 @@ trait PrepJSExports { this: PrepJSInterop =>
 
       // Make sure we do not override the default export of toString
       def isIllegalToString = {
-        isMember && !isNamedExport &&
+        isMember && !isNamedExport && !isTopLevelExport &&
         name == "toString" && sym.name != nme.toString_ &&
         sym.tpe.params.isEmpty && !jsInterop.isJSGetter(sym)
       }
@@ -255,7 +265,7 @@ trait PrepJSExports { this: PrepJSInterop =>
       }
 
       def isIllegalApplyExport = {
-        isMember && !hasExplicitName &&
+        isMember && !hasExplicitName && !isTopLevelExport &&
         sym.name == nme.apply &&
         !(isExportAll && directAnnots.exists(annot =>
             annot.symbol == JSExportAnnotation &&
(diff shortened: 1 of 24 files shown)


gzm0:
Review by @sjrd

sjrd:
Ci says it doesn't compile.

gzm0:
Didn't compile on 2.10. The reason was that non-trailing named parameters seem to be not allowed in 2.10:

```scala
foo(x = 1, 2) // compiles on >= 2.11, fails on 2.10
```

gzm0:
Ah and yes: Updated.

sjrd:
I have tons of things on my plate currently at EPFL, so I'll probably have to delay a proper review of this by several days. Worst case would be until next Wednesday, after the PLDI deadline.

gzm0:
> I have tons of things on my plate currently at EPFL, so I'll probably have to delay a proper review of this by several days. Worst case would be until next Wednesday, after the PLDI deadline.

Don't worry about it. It's not like I would have time to address comments until this weekend or even the next one :P

sjrd:
Jenkins, retest this please

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -87,8 +90,14 @@ trait PrepJSExports { this: PrepJSInterop =>
       // Reset interface flag: Any trait will contain non-empty methods
       clsSym.resetFlag(Flags.INTERFACE)
 
+      val (topLevelExports, otherExports) = exports.partition(_.isTopLevel)
+
+      // We can handle top level exports entirely in the backend. So just
+      // register them here.
```
sjrd: `/*  */`

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -168,10 +177,11 @@ trait PrepJSExports { this: PrepJSInterop =>
     val exports = directExportsOf(sym) ++ inheritedExportsOf(sym)
 
     // Calculate the distinct exports for this symbol (eliminate double
-    // occurrences of (name, isNamed) pairs).
-    val grouped = exports.groupBy(exp => (exp.jsName, exp.isNamed))
+    // occurrences of (name, isNamed, isTopLevel) tuples).
+    val grouped = exports.groupBy(exp =>
+        (exp.jsName, exp.isNamed, exp.isTopLevel))
```
sjrd: With `()`, the `exp =>` part should be on the same line as the body of the lambda.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -1073,4 +1073,117 @@ class JSExportTest extends DirectTest with TestHelpers {
     if (version.startsWith("2.10.") || version.startsWith("2.11.")) ""
     else s" (since $v)"
   }
+
+  @Test
+  def noExportTopLevelModule: Unit = {
+    """
+    @JSExportTopLevel("foo")
+    object A
```
sjrd: Maybe also a `noExportTopLevelTrait`?

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -1073,4 +1073,117 @@ class JSExportTest extends DirectTest with TestHelpers {
...
+      |       ^
+    """
+  }
+
+  @Test
+  def noExportTopLevelGetter: Unit = {
+    """
+    object A {
+      @JSExportTopLevel("foo")
+      def a: Int = 1
```
sjrd: Maybe also test with a `val` and with a `var`.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -1073,4 +1073,117 @@ class JSExportTest extends DirectTest with TestHelpers {
...
+      |       ^
+    """
+  }
+
+  @Test
+  def noExportTopLevelSetter: Unit = {
+    """
+    object A {
+      @JSExportTopLevel("foo")
+      def a_=(x: Int): Unit = ()
```
sjrd: Maybe also test with a `var`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -880,6 +885,27 @@ private[emitter] final class ScalaJSClassEmitter(
     )
   }
 
+  def genTopLevelExportDef(cd: LinkedClass, tree: TopLevelExportDef)(
+      implicit globalKnowledge: GlobalKnowledge): js.Tree = {
+    import TreeDSL._
+
+    val MethodDef(_, StringLiteral(fullName), args, resultType, Some(body)) =
+        tree.member
```
sjrd: Maybe use `true` instead of `_`. Trying to emit a non-static top-level method should not silently succeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -255,19 +268,25 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
   private def checkExportedMethodDef(methodDef: MethodDef,
-      classDef: LinkedClass): Unit = withPerMethodState {
+      classDef: LinkedClass, isTopLevel: Boolean): Unit = withPerMethodState {
     val MethodDef(static, StringLiteral(name), params, resultType, body) = methodDef
     implicit val ctx = ErrorContext(methodDef)
 
-    if (!classDef.kind.isAnyScalaJSDefinedClass) {
+    if (isTopLevel && classDef.kind != ClassKind.ModuleClass) {
+      reportError(s"TopLevel export can only appear in a module")
+      return
```
sjrd: While I understand why top-level methods can only appear in Scala `object`s at the source level, I don't think there is any reason to forbid them anywhere else at the IR level.

In fact, we are almost obliged to allow them in Scala `class`es, at the very least, because our linker can turn objects into classes if they're not instantiated. Consider this snippet of code:
```scala
object Foo {
  @ExportTopLevel("bar")
  def bar(): Int = 42
}
```
In the IR, we'll get something like:
```scala
module class Foo$ {
  def bar__I(): int = 42
  export top static def "bar"(): any = mod:Foo$.bar__I()
}
```
Now suppose the optimizer processes `"bar"` (which it doesn't, but it could). It would inline `Foo.bar`, and throw away the module load:
```scala
module class Foo$ {
  def bar__I(): int = 42
  export top static def "bar"(): any = 42
}
```
Susquently, the Refiner would turn `Foo$` into a (non-module) class, because it is not instantiated anymore:
```
class Foo$ {
  export top static def "bar"(): any = 42
}
```
That shouldn't freak out the IR checker nor the emitter.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -293,9 +312,11 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
             "but must be Any")
       }
 
-      val thisType =
-        if (classDef.kind.isJSClass) AnyType
+      val thisType = {
+        if (isTopLevel) NoType
```
sjrd: Consider `if (static)`, which is the actually reason `thisType` is `NoType`. The fact that `static == isTopLevel` is an accident, from the point of view of `thisType`.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -1603,3 +1628,26 @@ object ExportObjSetterNamed_= { // scalastyle:ignore
...
+  @JSExportTopLevel("org.scalajs.testsuite.jsinterop.toplevel.set")
+  def setMyVar(x: Int): Unit = myVar = x
+
+  object Nested {
+    var myVar: Int = _
+
+    @JSExportTopLevel("org.scalajs.testsuite.jsinterop.toplevel.setNested")
+    def setMyVar(x: Int): Unit = myVar = x
+  }
+}
```
sjrd: Could you also add a top-level object that is not reachable by any other way than through its top-level exports? I'd like to get confidence that the analyzer does the right thing here.

gzm0:
Also seems we are missing the stub for `@JSTopLevelExport`

gzm0:
Updated

gzm0:
Rebased on top of master.

gzm0:
Jenkins, test this please

PR #2647, opened 2016-11-07
Fix #2643: Do not replace `this` by `thisIdent` inside `Closure`s.

### LINKED ISSUE #2643: ThisFunction passing Scala-this instead of JS-this when using Scala 2.12.0
I have a code like this, and the ```foo``` function gets called from native JavaScript:
```scala
js.Dynamic.literal(
  "foo" -> js.ThisFunction.fromFunction1((that: js.Dynamic) => {
    foo(that)
  })
)
```

With Scala 2.11.8 and Scala JS 0.6.13, it compiles down to the following and JS-this is properly passed to the ```ThisFunction```.
```scala
"foo": (function(f$1) {
  return (function() {
    return f$1.apply__O__O(this)
  })
})
```

However, with Scala 2.12.0 and Scala JS 0.6.13, it compiles down to the following, and ```this``` is being Scala-this instead of JS-this:
```scala
"foo": (function(f$1) {
  return (function() {
    return f$1.apply__O__O($thiz)
  })
})
```

Only difference between the 2 builds are Scala version (2.11.8 vs. 2.12.0), so I am assuming that's the cause.

Let me know if you need some additional information.

sjrd:
`Closure`s reintroduce a new binding for `this`, which must not be replaced by `thisIdent`.

### DIFF
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -12,6 +12,12 @@ object BinaryIncompatibilities {
       ProblemFilters.exclude[MissingMethodProblem](
           "org.scalajs.core.tools.linker.backend.emitter.ScalaJSClassEmitter.genDeclareTypeData"),
 
+      // private, not an issue
+      ProblemFilters.exclude[MissingMethodProblem](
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0

PR #2650, opened 2016-11-08
java.util.NavigableMap interface

cquiroz:
This is the definition of the java interface `java.util.NavigableMap` as defined in [NavigableMap](https://docs.oracle.com/javase/7/docs/api/java/util/NavigableMap.html)

This interface is required to implement the `ZoneRulesProvider` class on the `java.time` API, in particular methods like [getVersions](https://docs.oracle.com/javase/8/docs/api/java/time/zone/ZoneRulesProvider.html#getVersions-java.lang.String-)

### DIFF
--- a/javalib/src/main/scala/java/util/NavigableMap.scala
+++ b/javalib/src/main/scala/java/util/NavigableMap.scala
@@ -1,0 +1,25 @@
+package java.util
+
+trait NavigableMap[K, V] extends SortedMap[K, V] {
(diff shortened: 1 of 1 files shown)


cquiroz:
I'm not providing an implementation as a full implementation of e.g. (`TreeMap`) is not needed on `java.time` only the signature is needed

sjrd:
Jenkins, test this please

PR #2652, opened 2016-11-12
Shorter and faster `Long` multiplication.

sjrd:
The key insight was that we do not need to compute the `hi` products by 16-bit components. The `hi` part is basically

    a.lo*b.hi + a.hi*b.hi + carry_from_lo_*

so only the carry from multiplying `a.lo*b.lo` needs to be computed by 16-bit components. The resulting algorithm is significantly shorter, and we can afford to completely inline it.

Subsequently, we can use the 16-bit products computed for the hi part to also compute the lo part, removing one multiplication.

The fact that we can now inline the whole algorithm allows it to benefit from call-site optimizations on the underlying `Int` arithmetics. This mostly happens when one side is constant and/or the `hi` part of operands is 0. The whole thing collapses particularly well when `a` is a constant in the range `0 <= a <= 0xffff`, which happens for mundane things like `10 * x`.

This improved implementation is significantly faster. A benchmark based on `BigInteger.multiply` exhibits > 2x improvement.

The benchmark code is the following:
```scala
  def benchmarkBigInteger(): Unit = {

    println("Bench BigInteger")
    import java.math.BigInteger

    // generate a prime number each time is too long, so here's one:
    val modHexa = "8600638d58155caded48cb45a1258593d775a03fe82679f62c62ec6c75e62e76fcb4e565f74675abe6cf6602939f6b4a159f56843f02f3f825e0f176a8df9f04f41eaa1cf60e598ca892670e30ea5b2d7002fb500b84e331d238708f600a14ec791dec30afc0fad2e5a0338847b0505f6d0a839516c892c06f4015362fbce145"

    val rand = new java.util.Random()
    val aHexa =  new BigInteger(1024, rand).toString(16)
    val bHexa =  new BigInteger(1024, rand).toString(16)

    println("Hexa String version")
    println("mod = " + modHexa)
    println("a = " + aHexa)
    println("b = " + bHexa)

    val mod =  new BigInteger(modHexa, 16)
    val a =  new BigInteger(aHexa, 16)
    val b =  new BigInteger(bHexa, 16)

    println("BigInteger version")
    println("mod = " + mod.toString(16))
    println("a = " + a.toString(16))
    println("b = " + b.toString(16))

    println("--------------------------")
    println("Do multiplication")

    var t0, t1:Long = 0
    val nbLoop = 100000
    var mutRes:BigInteger = null
    t0 = System.currentTimeMillis()
    for(i <- 0 until nbLoop) {
      mutRes = a multiply b
    }
    t1 = System.currentTimeMillis()
    val tBigInteger = (t1 - t0) * 1.0 / nbLoop
    println("--------------------------")
    println("BigInteger result = " + mutRes.toString(16))
    println("" + tBigInteger + " ms")
    println("--------------------------")
  }
```

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
@@ -302,9 +302,229 @@ final class RuntimeLong(val lo: Int, val hi: Int)
 
   @inline
   def *(b: RuntimeLong): RuntimeLong = {
+    /* The following algorithm is based on the decomposition in 32-bit and then
+     * 16-bit subproducts of the unsigned interpretation of operands.
+     *
+     * Since everything is interpreted as unsigned, all values are Natural
+     * numbers and are >= 0, by construction.
+     *
+     * We are looking to compute
+     * a *[64] b = (a * b) % 2^64
+     *
+     * We use the notation * and + for mathematical, non-overflowing
+     * operations, and *[64] and +[64] for overflowing operations. Eventually,
+     * we need to implement everything in terms of *[32] and +[32]. The symbol
+     * ^ is used for exponentiation (not bitwise xor).
+     *
+     * The decomposition in 32-bit components yields:
+     *
+     * a *[64] b
+     *   = ( (2^32*ahi + alo) * (2^32*bhi + blo) ) % 2^64
+     *   = ( 2^64*ahi*bhi + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *
+     * With Natural numbers, congruence theory tells us that we can "distribute"
+     * `% n` on + and *, as long as we also keep the outer `% n`. To be more
+     * precise:
+     *   (a + b) % n = (a%n + b) % n = (a + b%n) % n = (a%n + b%n) % n
+     *   (a * b) % n = (a%n * b) % n = (a * b%n) % n = (a%n * b%n) % n
+     *
+     * From the latter, we derive a corollary that we'll implicitly use several
+     * times later:
+     *   (n * x) % n = 0  for any n > 0 and any x
+     *
+     * We can use these equivalences to get rid of parts of our computation:
+     *
+     * ( 2^64*ahi*bhi + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *   = ( (2^64*ahi*bhi % 2^64) + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *        -------------------
+     *             = 0
+     *   = ( 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *
+     * Observe that we can rewrite any quantity x as
+     * x = n*(x/n) + x%n
+     * where n is > 0 and / denotes the floor division.
+     *
+     * We can rewrite the product ahi*blo as
+     *
+     * ahi*blo
+     *   = 2^32*(ahi*blo / 2^32) + (ahi*blo % 2^32)
+     *   = 2^32*(ahi*blo / 2^32) + (ahi *[32] blo)
+     *
+     * Similarly,
+     *
+     * bhi*alo = 2^32*(alo*bhi / 2^32) + (alo *[32] bhi)
+     *
+     * Taking back our complete computation:
+     *
+     * a *[64] b
+     *   = ( 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *   = ( 2^64*(ahi*blo / 2^32) + 2^32*(ahi *[32] blo)
+     *        + 2^64*(alo*bhi / 2^32) + 2^32*(alo *[32] bhi)
+     *        + alo*blo ) % 2^64
+     *
+     * where distributing % 2^64 allows to get rid of the most awful parts:
+     *
+     *   = ( 2^32*(ahi *[32] blo) + 2^32*(alo *[32] bhi) + alo*blo) % 2^64
+     *
+     * Now we focus on the `alo*blo` part. We decompose it in 16-bit components.
+     *
+     * alo * blo
+     *   = 2^32*a1*b1 + 2^16*a1*b0 + 2^16*a0*b1 + a0*b0
+     *
+     * Because a1, a0, b1 and b0 are all <= 2^16-1, their pair-wise products
+     * are all <= (2^16-1)^2 = 2^32 - 2*2^16 + 1 = 0xfffe0001 < 2^32. This
+     * means that, for example,
+     *   a1*b0 = (a1*b0) % 2^32 = a1 *[32] b0
+     * with the same applying to other subproducts.
+     *
+     * Let
+     *   a1b1 = a1 *[32] b1
+     *   a1b0 = a1 *[32] b0
+     *   a0b1 = a0 *[32] b1
+     *   a0b0 = a0 *[32] b0
+     *
+     * Each of those is <= 0xfffe0001.
+     *
+     * We now have:
+     *
+     * alo * blo
+     *   = 2^32*a1b1 + 2^16*a1b0 + 2^16*a0b1 + a0b0
+     *
+     * We can decompose it using / and % as follows:
+     * alo * blo
+     *   = 2^32*((alo * blo) / 2^32) + ((alo * blo) % 2^32)
+     *
+     * Let
+     *   aloblo = (alo * blo) % 2^32
+     *   carry_from_lo_* = (alo * blo) / 2^32
+     *
+     * Then
+     * alo * blo = 2^32 * carry_from_lo_* + aloblo
+     *
+     * aloblo = (alo * blo) % 2^32
+     *   = (2^32*a1b1 + 2^16*a1b0 + 2^16*a0b1 + a0b0) % 2^32
+     *   = (2^16*a1b0 + 2^16*a0b1 + a0b0) % 2^32
+     *   = (((2^16*a1b0 % 2^32 + 2^16*a0b1 % 2^32) % 2^32) + a0b0 % 2^32) % 2^32
+     *   = (2^16*a1b0 % 2^32) +[32] (2^16*a0b1 % 2^32) +[32] (a0b0 % 2^32)
+     *   = (a1b0 <<[32] 16) +[32] (a0b1 <<[32] 16) +[32] a0b0
+     *
+     * carry_from_lo_* is more difficult.
+     *
+     * carry_from_lo_* = (alo * blo) / 2^32
+     *   = (2^32*a1b1 + 2^16*a1b0 + 2^16*a0b1 + a0b0) / 2^32
+     *   = a1b1 + (2^16*a1b0 + 2^16*a0b1 + a0b0) / 2^32
+     *   = a1b1 + (2^16*a1b0 + 2^16*a0b1 + (2^16*(a0b0 / 2^16) + (a0b0 % 2^16))) / 2^32
+     *   = a1b1 + (2^16*(a1b0 + a0b1 + (a0b0 / 2^16)) + (a0 % 2^16)) / 2^32
+     *             ----------------------------------
+     *                  multiple of 2^16
+     *   = a1b1 + ( (2^16*(a1b0 + a0b1 + (a0b0 / 2^16))) / 2^16 + (a0 % 2^16) / 2^16 ) / 2^16
+     *                                                             ---------
+     *                                                               < 2^16
+     *   = a1b1 + (a1b0 + a0b1 + (a0b0 / 2^16)) / 2^16
+     *   = a1b1 + (a1b0 + (a0b1 + (a0b0 >>>[32] 16))) / 2^16
+     *                     ----    ---------------
+     *               <= 0xfffe0001    <= 0xffff
+     *                     ------------------------
+     *                       <= 0xffff0000, hence the + does not overflow
+     *   = a1b1 + (a1b0 + (a0b1 +[32] (a0b0 >>>[32] 16))) / 2^16
+     *
+     * Let
+     *   c1part = a0b1 +[32] (a0b0 >>>[32] 16)
+     *
+     * carry_from_lo_*
+     *   = a1b1 + (a1b0 + c1part) / 2^16
+     *   = a1b1 + (a1b0 + (2^16*(c1part / 2^16) + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part &[32] 0xffff))) / 2^16
+     *                                     ----    -------------------
+     *                               <= 0xfffe0001     <= 0xffff
+     *                                     ----------------------------
+     *                              <= 0xffff0000, hence the + does not overflow
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 +[32] (c1part &[32] 0xffff))) / 2^16
+     *             --------------------
+     *               multiple of 2^16
+     *   = a1b1 + ( 2^16*(c1part / 2^16) / 2^16 + (a1b0 +[32] (c1part &[32] 0xffff)) / 2^16 )
+     *   = a1b1 + (c1part / 2^16) + (a1b0 +[32] (c1part &[32] 0xffff)) / 2^16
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0 

I'm not sure I trust this code quite yet. Tests pass, and I thought about it. But since Georg Schmid is trying to *prove* it, let's wait and see what that gives before merging.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -304,7 +304,29 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
     val blo = b.lo
-    new RuntimeLong(alo * blo, RuntimeLong.timesHi(alo, a.hi, blo, b.hi))
+
+    val a0 = alo & 0xffff
+    val a1 = alo >>> 16
+    val b0 = blo & 0xffff
+    val b1 = blo >>> 16
+
+    val a0b0 = a0 * b0
+    val a1b0 = a1 * b0 // collapses to 0 when a is constant and 0 <= a <= 0xffff
```
gzm0: Consider formulating this comment in a way that makes it explicit that it applies equally to `a0b1` and `b`.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -304,7 +304,29 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+    val a0b1 = a0 * b1
+
+    /* lo = a.lo * b.lo, but we compute the above 3 subproducts for hi
+     * anyway, we reuse them to compute lo too, trading a * for 2 +'s and 1 <<.
+     */
+    val lo = a0b0 + ((a1b0 + a0b1) << 16)
+
+    // hi = a.lo*b.hi + a.hi*b.hi + carry_from_lo_*
```
gzm0: `a.hi*b.lo`

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -304,7 +304,29 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+    val c1part = (a0b0 >>> 16) + a0b1
+    val hi = {
+      alo*b.hi + a.hi*blo + a1 * b1 + (c1part >>> 16) +
+      (((c1part & 0xffff) + a1b0) >>> 16) // collapses to 0 when a1b0 = 0
```
gzm0: I think a comment *why* this way of calculating `carry_from_lo_*` is ok would be helpful.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -304,7 +304,29 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+
+    /* lo = a.lo * b.lo, but we compute the above 3 subproducts for hi
+     * anyway, we reuse them to compute lo too, trading a * for 2 +'s and 1 <<.
+     */
+    val lo = a0b0 + ((a1b0 + a0b1) << 16)
+
+    // hi = a.lo*b.hi + a.hi*b.hi + carry_from_lo_*
+    val c1part = (a0b0 >>> 16) + a0b1
+    val hi = {
+      alo*b.hi + a.hi*blo + a1 * b1 + (c1part >>> 16) +
```
gzm0: Will the folder be able to collapse `c1part >> 16` to 0 when `a0b1 = 0`?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Ah, but that is (almost) never going to happen. If `b` is constant, then either `a` is constant too, and the multiplication is constant-folded away; or `a` is not constant, and the optimizer would have normalized the multiplication as `b * a` so that the lhs is constant.

The only case where `a0b1` would be constant 0 is if `b`'s `lo` part is constant but not its `hi` part. That's not a likely scenario, though (not seen at all in our test suite).

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: No, it is not (yet) smart enough for that. But as explained above, `a0b1 == 0` is not a likely scenario.

sjrd:
Updated. Now you get to verify the proof you made me write :P

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+    /* The following algorithm is based on the decomposition in 32-bit and then
+     * 16-bit subproducts of the unsigned interpretation of operands.
+     *
+     * We are looking to compute
+     * a *[64] b = (a * b) % 2^64
+     *
+     * We use the notation * and + for mathematical, non-overflowing
+     * operations, and *[64] and +[64] for overflowing operations. Eventually,
+     * we need to implement everything in terms of *[32] and +[32]. The symbol
+     * ^ is used for exponentiation (nor bitwise xor).
```
gzm0: nor -> not

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *   = ( (2^32*ahi + alo) * (2^32*bhi + blo) ) % 2^64
+     *   = ( 2^64*ahi*bhi + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *
+     * With Natural numbers, congruence theory tells us that we can "distribute"
+     * `% n` on + and *, as long as we also keep the other `% n`. To be more
+     * precise:
+     *   (a + b) % n = (a%n + b) % n = (a + b%n) % n = (a%n + b%n) % n
+     *   (a * b) % n = (a%n * b) % n = (a * b%n) % n = (a%n * b%n) % n
+     *
+     * We can use these equivalence to get rid of parts of our computation:
```
gzm0: "these equivalences" or "this equivalence".

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *
+     * ( 2^64*ahi*bhi + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
+     *   = ( (2^64*ahi*bhi % 64) + 2^32*(ahi*blo + bhi*alo) + alo*blo ) % 2^64
```
gzm0: `% 2^64`

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *
+     * Then
+     * alo * blo = 2^32 * carry_from_lo_* + aloblo
+     *
+     * aloblo = (alo * blo) % 2^32
+     *   = (2^32*a1b1 + 2^16*a1b0 + 2^16*a0b1 + a0b0) % 2^32
+     *   = (2^16*a1b0 + 2^16*a0b1 + a0b0) % 2^32
+     *   = (((2^16*a1b0 % 2^32 + 2^16*a0b1 % 2^32) % 2^32) + a0b0 % 2^32) % 2^32
+     *   = (2^16*a1b0 % 2^32) +[32] (2^16*a0b1 % 2^32) +[32] (a0b0 % 2^32)
+     *   = (a1b0 << 16) +[32] (a0b1 << 16) +[32] a0b0
```
gzm0: I would say, use `<<[32]` instead of `<<` for 100% clarity.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
 
   @inline
   def *(b: RuntimeLong): RuntimeLong = {
+    /* The following algorithm is based on the decomposition in 32-bit and then
+     * 16-bit subproducts of the unsigned interpretation of operands.
+     *
+     * We are looking to compute
+     * a *[64] b = (a * b) % 2^64
```
gzm0: Are you assuming that `a >= 0` and `b >= 0` here?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *   = a1b1 + (2^16*a1b0 + 2^16*a0b1 + a0b0) / 2^32
+     *   = a1b1 + (2^16*a1b0 + 2^16*a0b1 + (2^16*(a0b0 / 2^16) + (a0b0 % 2^16))) / 2^32
+     *   = a1b1 + (2^16*(a1b0 + a0b1 + (a0b0 / 2^16)) + (a0 % 2^16)) / 2^32
+     *             ----------------------------------
+     *                  multiple of 2^16
+     *   = a1b1 + ( (2^16*(a1b0 + a0b1 + (a0b0 / 2^16))) / 2^16 + (a0 % 2^16) / 2^16 ) / 2^16
+     *                                                             ---------
+     *                                                               < 2^16
+     *   = a1b1 + (a1b0 + a0b1 + (a0b0 / 2^16)) / 2^16
+     *   = a1b1 + (a1b0 + (a0b1 + (a0b0 >>> 16))) / 2^16
```
gzm0: You are using `a0b0 >= 0` here. Can you justify this? (I'm assuming you are doing everything unsigned, but we should state this).

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *                     ----    -----------
+     *                <= fffe0001    <= ffff
```
gzm0: Can you please:
- prefix with `0x` for better clarity.
- proof that `a0b1 <= 0xfffe0001` (enough to enhance line 368 IIUC).

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *   = a1b1 + (a1b0 + (a0b1 +[32] (a0b0 >>> 16))) / 2^16
+     *
+     * Let
+     *   c1part = a0b1 +[32] (a0b0 >>> 16)
+     *
+     * carry_from_lo_*
+     *   = a1b1 + (a1b0 + c1part) / 2^16
+     *   = a1b1 + (a1b0 + (2^16*(c1part / 2^16) + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part & 0xffff))) / 2^16
```
gzm0: Same comment about `c1part >= 0`.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,212 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     * carry_from_lo_*
+     *   = a1b1 + (c1part >>> 16) + ((a1b0 +[32] (c1part & 0xffff)) >>> 16)
+     *
+     * Substituting,
+     *
+     * a *[64] b
+     *   = ( 2^32*(ahi *[32] blo) + 2^32*(alo *[32] bhi) + 2^32*carry_from_lo_* + aloblo ) % 2^64
+     *   = ( 2^32*((ahi *[32] blo) + (alo *[32] bhi) + carry_from_lo_*) + aloblo ) % 2^64
+     *   = ( 2^32*((ahi *[32] blo) + (alo *[32] bhi) + carry_from_lo_*) % 2^64 + aloblo ) % 2^64
+     *   = ( 2^32*(((ahi *[32] blo) + (alo *[32] bhi) + carry_from_lo_*) % 2^32) + aloblo ) % 2^64
```
gzm0: This step (`n*x % n^2 = n * (x % n)`) is non-obvious, can you make this more explicit?

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: > of the unsigned interpretation of operands

Yes

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: OK I'll come up with something.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: There are two of them just above. (although I'm not sure I actually use the second one; I'll check)

gzm0: Yes... It's just the grammar. "equivalence" is missing an "s".

sjrd:
Updated.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
sjrd: Turns out I had to prove a lemma to convince myself that this rewriting was indeed correct ^^

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,229 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *   = a1b1 + (2^16*a1b0 + 2^16*a0b1 + (2^16*(a0b0 / 2^16) + (a0b0 % 2^16))) / 2^32
+     *   = a1b1 + (2^16*(a1b0 + a0b1 + (a0b0 / 2^16)) + (a0 % 2^16)) / 2^32
+     *             ----------------------------------
+     *                  multiple of 2^16
+     *   = a1b1 + ( (2^16*(a1b0 + a0b1 + (a0b0 / 2^16))) / 2^16 + (a0 % 2^16) / 2^16 ) / 2^16
+     *                                                             ---------
+     *                                                               < 2^16
+     *   = a1b1 + (a1b0 + a0b1 + (a0b0 / 2^16)) / 2^16
+     *   = a1b1 + (a1b0 + (a0b1 + (a0b0 >>>[32] 16))) / 2^16
+     *                     ----    -----------
```
gzm0: The second underline is too short.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,229 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *               <= 0xfffe0001    <= 0xffff
+     *                     --------------------
```
gzm0: Also, this underline is not long enough.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -302,9 +302,229 @@ final class RuntimeLong(val lo: Int, val hi: Int)
...
+     *   c1part = a0b1 +[32] (a0b0 >>>[32] 16)
+     *
+     * carry_from_lo_*
+     *   = a1b1 + (a1b0 + c1part) / 2^16
+     *   = a1b1 + (a1b0 + (2^16*(c1part / 2^16) + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part % 2^16))) / 2^16
+     *   = a1b1 + (2^16*(c1part / 2^16) + (a1b0 + (c1part &[32] 0xffff))) / 2^16
+     *                                     ----    -------------------
+     *                               <= 0xfffe0001     <= 0xffff
+     *                                     ---------------------
```
gzm0: Too short underline.

PR #2654, opened 2016-11-17
25% faster Long.toString() for large Longs (>= 2^53).

sjrd:
For large `Long` values, `toString()` divides once by 10^9, then converts the quotient and remainder to strings and concatenates the results.

Previously, the quotient and remainder were returned in a tuple from `unsignedDivModHelper()`. Now, we let the latter perform the final step of conversion to string, which avoids the tuple allocation. According to micro-benchmarks, this optimization shaved off 25% of the run-time of `Long.toString()` for large `Long` values.

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
@@ -427,7 +427,7 @@ object RuntimeLong {
 
   private final val AskQuotient = 0
   private final val AskRemainder = 1
-  private final val AskBoth = 2
+  private final val AskToString = 2
 
   /** The hi part of a (lo, hi) return value. */
   private[this] var hiReturn: Int = _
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @gzm0

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -852,7 +846,10 @@ object RuntimeLong {
       hiReturn = remHi
       remLo
     } else {
-      js.Tuple4(quotLo, quotHi, remLo, remHi)
+      // AskToString
+      val quot = asUnsignedSafeDouble(quotLo, quotHi)
+      val remStr = remLo.toString
+      quot.toString + "000000000".jsSubstring(remStr.length) + remStr
```
gzm0: Can you please add a comment why `remHi` can be disregarded here? There currently is a non-explicit precondition that `AskToString` may only be used if `b` is large enough.

sjrd:
Updated.

PR #2657, opened 2016-11-19
Remove overrides for Range and RangeNumeric

### LINKED ISSUE #2655: Inconsistent behavior in min method between JVM and JS
# JVM

```scala
scala> val x = 1 to 10 by 1
x: scala.collection.immutable.Range = Range(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)

scala> println(x.min)
1

scala> val y = 1.0 to 10.0 by 1.0
y: scala.collection.immutable.NumericRange[Double] = NumericRange(1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0)

scala> println(y.min)
1.0
```

# JS

```scala
val x = 1 to 10 by 1
println(x.min) // 1

val y = 1.0 to 10.0 by 1.0
println(y.min) // error
```

https://gitter.im/scala-js/scala-js?at=582f309db563b5516c371ee4

gzm0:
Since we now have full implementations of BigInteger and BigDecimal, we
do not need to override Range anymore (and they fully work).

As a side effect, this fixes #2655 (which was caused by the overrides
going out of sync).


gzm0:
Review by @sjrd

gzm0:
Just as a general maintenance task, I'll check if there are other overrides we can get rid of.

gzm0:
As mentioned by @sjrd on #2655, this will make code size explode. So closing this.

Issue #1541, opened 2015-03-15
Hijack scala.Symbol as ES6 Symbols


sjrd:
The more I think about it, the more I am convinced that we should hijack `scala.Symbol` to represent them as the new primitive type `Symbol` of ES6.
Doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol

Unlike `scala.Symbol`, there are 2 variants of ES6 symbols: named and opaque (or shared and non-shared, to take MDN's terminology). Named symbols are created with `Symbol.for("foo")`, and correspond to `scala.Symbol`:

* Two shared symbols created for the same string have the same identity: `Symbol.for("foo") === Symbol.for("foo")`
* You can get the string that was used to created a shared symbol: `Symbol.keyFor(Symbol.for("foo")) === "foo")`

Unique symbols are created with `Symbol()` or `Symbol(desc)`. The description is only relevant for debugging purposes, and cannot be retrieved. Each symbol created this way is unique, even if the same description is given twice.

I believe all of these can fit in a `scala.Symbol`. `Symbol.apply` would use `Symbol.for` to create a shared symbol, so only shared symbols can be created with the Scala API. But unique symbols created with the JS API would be valid "instances" of `scala.Symbol`. The semantics of `name()` would be modified to throw for unique symbols.

One gotcha: `scala.Symbol` allows `Symbol.apply(null)`, and gives a perfectly working `Symbol` whose name is `null` (not `"null"` the string). In JS, `Symbol.for(null) === Symbol.for("null")`, so we would create a dedicated unique symbol for `Symbol.apply(null)`.

The semantics of `.toString()` would change. In JS, `Symbol.for("foo").toString() === "Symbol(foo)"`, but in Scala it should be `'foo`.

JS' `Symbol`s are the only hope to ever get memory-leak-free `scala.Symbol`s. Note that `WeakMap`s do not help, because their *keys* are weak, not their values, and they are not iterable.

**Alternative**

[...]

Issue #1111, opened 2014-09-25
Turn all warnings in PrepJSInterop into errors


gzm0:
See #130

gzm0:
I want to do this, once it is time :)

sjrd:
OK ^^

sjrd:
Generalizing this to include all warnings generated by `PrepJSInterop` into errors.

sjrd:
@gzm0 

> I want to do this, once it is time :)

It is time :)

PR #2659, opened 2016-11-21
Fix #1997: Add support for js.Symbols in @JSName annotations.

### LINKED ISSUE #1997: Need @JSSymName(..) annotation

gzm0:
(Old PR description below)

The `@JSName` annotation could previously be used to specify the JavaScript *string* name of a property/method, independently of its Scala name. This can be used both in native JS classes as well as Scala.js-defined JS classes.

This commit extends the capability of `@JSName` to admit a (static, stable) reference to a `js.Symbol`. This applies to both native and Scala.js-defined JS types.

At call site, the semantics and compilation scheme are straightforward: it is a `JSBracketSelect` whose `item` part is a call to the symbol accessor. Since it is static by construction, we can call it from anywhere.

At definition site for native JS types, there is nothing to do (besides validity checks).

At definition site for Scala.js-defined JS classes, things get tricky. We can now have "exported" class members whose name is not a constant string anymore. This requires to extend the IR with `ComputedName`s (a concept that exists in ECMAScript 2015). A computed name is an arbitrary expression `Tree` that is evaluated to obtain the dynamic name of a property. This complicates the Emitter somewhat, but not that much because Scala.js-defined JS classes are created lazily, so we can execute arbitrary user-defined code at class-definition time.

For overloading resolution, `ComputedName`s have a `logicalName`, derived from the fully qualified name of the symbol accessors. All `ComputedName`'d properties/methods with the same logical name (i.e., referencing the same static stable `js.Symbol`) are considered to be overloaded. This has an unavoidable caveat: if two `ComputedName`s in the same class have a different logical name, but actually evaluate to the same symbol, things will go wrong at run-time.

`JSObjectConstr` is enhanced to support `ComputedName`s as well, for consistency. It is also consistent with Object Initializers in ECMAScript 2015. This actually complicates their treatment, mostly because of evaluation order concerns.

Exports (in Scala classes and at the top-level) cannot use symbolic names. This is not expected to change in the future.

There is one implementation restriction: `@JSName` with a symbol is not supported on native JS classes and objects that are nested within a native JS object. There is no language reason to disallow this, but the implementation is surprisingly difficult, due to the way scalac flattens such entities.

---

# Old PR description

We introduce @JSSymbol as a pendant to @JSName to define member access through ES6 symbols. On a high level, the following happens in the compiler:

## PrepJSInterop / Definition Site
Take the content of @JSSymbol annotations and put it into a special method, the "symbol forwarder". This is necessary, since otherwise the trees won't get compiled. As a nice side effect, it allows changing the symbol without breaking binary compatibility.

Further, we replace the tree inside @JSSymbol with a simple tree calling the symbol forwarder. This tree will not get compiled, but we will only use its symbol.

## PrepJSInterop / Call Site
Nothing.

## GenJSCode / Definition Site
We emit symbol forwarders as static methods on the native JS class. This does not need an IR change, since this is already supported (but was unused so far).

For traits pre 2.12.0, this is a bit more involved, since the symbol forwarder gets moved to the implementation class. We retrieve the implementation class and move the forwarder.

## GenJSCode / Call Site
Instead of generating a string literal inside the bracket select, we generate a call to the symbol forwarder. The symbol forwarder's symbol is conveniently available inside the tree of the @JSSymbol annotation.

## Caveat
Since we generate number suffixes to disambiguate symbol forwarder overloads, adding an overload to a method with @JSSymbol is a potentially binary breaking change.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -1100,7 +1100,7 @@ abstract class GenJSCode extends plugins.PluginComponent
         else (subConstructors.head.overrideNumBounds._1, overrideNum)
 
       def get(methodName: String): Option[ConstructorTree] = {
-        if (methodName == this.method.name.name) {
+        if (methodName == this.method.name.encodedName) {
           Some(this)
         } else {
           subConstructors.iterator.map(_.get(methodName)).collectFirst {
@@ -1370,7 +1370,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       var overrideNum = -1
       def mkConstructorTree(method: js.MethodDef): ConstructorTree = {
-        val methodName = method.name.name
+        val methodName = method.name.encodedName
         val subCtrTrees = ctorToChildren(methodName).map(mkConstructorTree)
         overrideNum += 1
         new ConstructorTree(overrideNum, method, subCtrTrees)
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -3908,9 +3908,10 @@ abstract class GenJSCode extends plugins.PluginComponent
          *   {name1: arg1, name2: arg2, ... }
          */
 
-        def warnIfDuplicatedKey(pairs: List[(js.StringLiteral, js.Tree)]): Unit = {
-          val allKeys = pairs.collect { case (js.StringLiteral(keyName), _) => keyName }
-          val keyCounts = allKeys.distinct.map(key => key -> allKeys.count(_ == key))
+        def warnIfDuplicatedKey(keys: List[js.StringLiteral]): Unit = {
+          val keyNames = keys.map(_.value)
+          val keyCounts =
+            keyNames.distinct.map(key => key -> keyNames.count(_ == key))
           val duplicateKeyCounts = keyCounts.filter(1 < _._2)
           if (duplicateKeyCounts.nonEmpty) {
             reporter.warning(pos,
@@ -3923,12 +3924,21 @@ abstract class GenJSCode extends plugins.PluginComponent
           }
         }
 
+        def keyToPropName(key: js.Tree, index: Int): js.PropertyName = key match {
+          case key: js.StringLiteral => key
+          case _                     => js.ComputedName(key, "local" + index)
+        }
+
         // Extract first arg to future proof against varargs
         extractFirstArg(genArgs) match {
-          // case js.Dynamic.literal("name1" -> ..., "name2" -> ...)
-          case (js.StringLiteral("apply"), jse.LitNamed(pairs)) =>
-            warnIfDuplicatedKey(pairs)
-            js.JSObjectConstr(pairs)
+          // case js.Dynamic.literal("name1" -> ..., nameExpr2 -> ...)
+          case (js.StringLiteral("apply"), jse.Tuple2List(pairs)) =>
+            warnIfDuplicatedKey(pairs.collect {
+              case (key: js.StringLiteral, _) => key
+            })
+            js.JSObjectConstr(pairs.zipWithIndex.map {
+              case ((key, value), index) => (keyToPropName(key, index), value)
+            })
 
           /* case js.Dynamic.literal(x: _*)
            * Even though scalac does not support this notation, it is still
@@ -3950,31 +3960,27 @@ abstract class GenJSCode extends plugins.PluginComponent
           // case js.Dynamic.literal(x, y)
           case (js.StringLiteral("apply"), tups) =>
             // Check for duplicated explicit keys
-            val pairs = jse.LitNamedExtractor.extractFrom(tups)
-            warnIfDuplicatedKey(pairs)
+            warnIfDuplicatedKey(jse.extractLiteralKeysFrom(tups))
 
-            // Create tmp variable
-            val resIdent = freshLocalIdent("obj")
-            val resVarDef = js.VarDef(resIdent, jstpe.AnyType, mutable = false,
-                js.JSObjectConstr(Nil))
-            val res = resVarDef.ref
-
-            // Assign fields
+            // Evaluate all tuples first
             val tuple2Type = encodeClassType(TupleClass(2))
-            val assigns = tups flatMap {
-              // special case for literals
-              case jse.Tuple2(name, value) =>
-                js.Assign(js.JSBracketSelect(res, name), value) :: Nil
-              case tupExpr =>
-                val tupIdent = freshLocalIdent("tup")
-                val tup = js.VarRef(tupIdent)(tuple2Type)
-                js.VarDef(tupIdent, tuple2Type, mutable = false, tupExpr) ::
-                js.Assign(js.JSBracketSelect(res,
-                    genApplyMethod(tup, js.Ident("$$und1__O"), Nil, jstpe.AnyType)),
-                    genApplyMethod(tup, js.Ident("$$und2__O"), Nil, jstpe.AnyType)) :: Nil
+            val evalTuples = tups.map { tup =>
+              js.VarDef(freshLocalIdent("tup"), tuple2Type, mutable = false,
+                  tup)(tup.pos)
             }
 
-            js.Block(resVarDef +: assigns :+ res: _*)
+            // Build the resulting object
+            val result = js.JSObjectConstr(evalTuples.zipWithIndex.map {
+              case (evalTuple, index) =>
+                val tupRef = evalTuple.ref
+                val key = genApplyMethod(tupRef, js.Ident("$$und1__O"), Nil,
+                    jstpe.AnyType)
+                val value = genApplyMethod(tupRef, js.Ident("$$und2__O"), Nil,
+                    jstpe.AnyType)
+                keyToPropName(key, index) -> value
+            })
+
+            js.Block(evalTuples :+ result)
 
           // case where another method is called
           case (js.StringLiteral(name), _) if name != "apply" =>
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -46,7 +46,7 @@ abstract class GenJSCode extends plugins.PluginComponent
   import rootMirror._
   import definitions._
   import jsDefinitions._
-  import jsInterop.{jsNameOf, compat068FullJSNameOf, jsNativeLoadSpecOf}
+  import jsInterop.{jsNameOf, compat068FullJSNameOf, jsNativeLoadSpecOf, JSName}
   import JSTreeExtractors._
 
   import treeInfo.hasSynthCaseSymbol
@@ -491,7 +491,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       val constructorTrees = new ListBuffer[DefDef]
       val generatedMethods = new ListBuffer[js.MethodDef]
-      val dispatchMethodNames = new ListBuffer[String]
+      val dispatchMethodNames = new ListBuffer[JSName]
 
       def gen(tree: Tree): Unit = {
         tree match {
@@ -621,15 +621,14 @@ abstract class GenJSCode extends plugins.PluginComponent
               assert(mdef.static, "Non-static method in SJS defined JS class")
               staticMembers += mdef
 
-            case js.StringLiteral(name) =>
+            case js.StringLiteral("constructor") =>
               assert(!mdef.static, "Exported static method")
+              assert(constructor.isEmpty, "two ctors in class")
+              constructor = Some(mdef)
 
-              if (name == "constructor") {
-                assert(constructor.isEmpty, "two ctors in class")
-                constructor = Some(mdef)
-              } else {
-                classMembers += mdef
-              }
+            case _ =>
+              assert(!mdef.static, "Exported static method")
+              classMembers += mdef
           }
 
         case property: js.PropertyDef =>
@@ -667,8 +666,9 @@ abstract class GenJSCode extends plugins.PluginComponent
         case fdef: js.FieldDef =>
           implicit val pos = fdef.pos
           val select = fdef.name match {
-            case lit: js.StringLiteral => js.JSBracketSelect(selfRef, lit)
-            case ident: js.Ident       => js.JSDotSelect(selfRef, ident)
+            case ident: js.Ident          => js.JSDotSelect(selfRef, ident)
+            case lit: js.StringLiteral    => js.JSBracketSelect(selfRef, lit)
+            case js.ComputedName(tree, _) => js.JSBracketSelect(selfRef, tree)
           }
           js.Assign(select, jstpe.zeroOf(fdef.tpe))
 
@@ -896,7 +896,7 @@ abstract class GenJSCode extends plugins.PluginComponent
         }
 
         val name =
-          if (isExposed(f)) js.StringLiteral(jsNameOf(f))
+          if (isExposed(f)) genPropertyName(jsNameOf(f))
           else encodeFieldSym(f)
 
         val irTpe = {
@@ -1844,7 +1844,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           } else if (isScalaJSDefinedJSClass(sym.owner)) {
             val genQual = genExpr(qualifier)
             val boxed = if (isExposed(sym))
-              js.JSBracketSelect(genQual, js.StringLiteral(jsNameOf(sym)))
+              js.JSBracketSelect(genQual, genExpr(jsNameOf(sym)))
             else
               js.JSDotSelect(genQual, encodeFieldSym(sym))
             unboxFieldValue(boxed)
@@ -1929,7 +1929,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
               if (isScalaJSDefinedJSClass(sym.owner)) {
                 val genLhs = if (isExposed(sym))
-                  js.JSBracketSelect(genQual, js.StringLiteral(jsNameOf(sym)))
+                  js.JSBracketSelect(genQual, genExpr(jsNameOf(sym)))
                 else
                   js.JSDotSelect(genQual, encodeFieldSym(sym))
                 js.Assign(genLhs, genBoxedRhs)
@@ -4199,7 +4199,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           js.JSFunctionApply(receiver, args)
 
         case _ =>
-          def jsFunName = js.StringLiteral(jsNameOf(sym))
+          def jsFunName: js.Tree = genExpr(jsNameOf(sym))
 
           def genSuperReference(propName: js.Tree): js.Tree = {
             superIn.fold[js.Tree] {
@@ -5142,6 +5142,40 @@ abstract class GenJSCode extends plugins.PluginComponent
           paramsLocal :+ ensureBoxed(body, methodType.resultType))
 
       (patchedParams, patchedBody)
+    }
+
+    // Methods to deal with JSName ---------------------------------------------
+
+    def genPropertyName(name: JSName)(implicit pos: Position): js.PropertyName = {
+      name match {
+        case JSName.Literal(name) => js.StringLiteral(name)
+
+        case JSName.Computed(sym) =>
+          js.ComputedName(genComputedJSName(sym), encodeComputedNameIdentity(sym))
+      }
+    }
+
+    def genExpr(name: JSName)(implicit pos: Position): js.Tree = name match {
+      case JSName.Literal(name) => js.StringLiteral(name)
+      case JSName.Computed(sym) => genComputedJSName(sym)
+    }
+
+    private def genComputedJSName(sym: Symbol)(implicit pos: Position): js.Tree = {
+      /* By construction (i.e. restriction in PrepJSInterop), we know that sym
+       * must be a static method.
+       * Therefore, at this point, we can invoke it by loading its owner and
+       * calling it.
+       */
+      val module = genLoadModule(sym.owner)
+
+      if (isRawJSType(sym.owner.tpe)) {
+        if (!isScalaJSDefinedJSClass(sym.owner) || isExposed(sym))
+          genJSCallGeneric(sym, module, args = Nil, isStat = false)
+        else
+          genApplyJSClassMethod(module, sym, arguments = Nil)
+      } else {
+        genApplyMethod(module, sym, arguments = Nil)
+      }
     }
 
     // Utilities ---------------------------------------------------------------
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -27,7 +27,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
   import jsAddons._
   import definitions._
   import jsDefinitions._
-  import jsInterop.jsNameOf
+  import jsInterop.{jsNameOf, JSName}
 
   trait JSExportsPhase { this: JSCodePhase =>
 
@@ -54,7 +54,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     }
 
     def genJSClassDispatchers(classSym: Symbol,
-        dispatchMethodsNames: List[String]): List[js.Tree] = {
+        dispatchMethodsNames: List[JSName]): List[js.Tree] = {
       dispatchMethodsNames
         .map(genJSClassDispatcher(classSym, _))
     }
@@ -91,7 +91,8 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
           implicit val pos = ctors.head.pos
 
           val js.MethodDef(_, _, args, _, body) =
-            withNewLocalNameScope(genExportMethod(ctors, jsName, static = false))
+            withNewLocalNameScope(genExportMethod(ctors, JSName.Literal(jsName),
+                static = false))
 
           js.ConstructorExportDef(jsName, args, body.get)
         }
@@ -225,8 +226,8 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
 
         // Generate the export
 
-        val exportedMember = genMemberExportOrDispatcher(classSym, jsName,
-            isProp, alts, static = true)
+        val exportedMember = genMemberExportOrDispatcher(classSym,
+            JSName.Literal(jsName), isProp, alts, static = true)
 
         val exportDef =
           if (destination == ExportDestination.Static) exportedMember
@@ -317,11 +318,11 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
             s"Exported $kind $jsName conflicts with ${alts.head.fullName}")
       }
 
-      genMemberExportOrDispatcher(classSym, jsName, isProp, alts,
-          static = false)
+      genMemberExportOrDispatcher(classSym, JSName.Literal(jsName), isProp,
+          alts, static = false)
     }
 
-    private def genJSClassDispatcher(classSym: Symbol, name: String): js.Tree = {
+    private def genJSClassDispatcher(classSym: Symbol, name: JSName): js.Tree = {
       var alts: List[Symbol] = Nil
       for {
         sym <- classSym.info.members
@@ -348,7 +349,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
       }
     }
 
-    def genMemberExportOrDispatcher(classSym: Symbol, jsName: String,
+    def genMemberExportOrDispatcher(classSym: Symbol, jsName: JSName,
         isProp: Boolean, alts: List[Symbol], static: Boolean): js.Tree = {
       withNewLocalNameScope {
         if (isProp)
@@ -358,10 +359,12 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
       }
     }
 
-    def genJSConstructorExport(alts: List[Symbol]): js.MethodDef =
-      genExportMethod(alts.map(ExportedSymbol), "constructor", static = false)
+    def genJSConstructorExport(alts: List[Symbol]): js.MethodDef = {
+      genExportMethod(alts.map(ExportedSymbol), JSName.Literal("constructor"),
+          static = false)
+    }
 
-    private def genExportProperty(alts: List[Symbol], jsName: String,
+    private def genExportProperty(alts: List[Symbol], jsName: JSName,
         static: Boolean): js.PropertyDef = {
       assert(!alts.isEmpty)
       implicit val pos = alts.head.pos
@@ -379,7 +382,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
             s"Found more than one instance getter to export for name $jsName.")
         for (duplicate <- getter.tail) {
           reporter.error(duplicate.pos,
-              s"Duplicate static getter export with name '$jsName'")
+              s"Duplicate static getter export with name '${jsName.displayName}'")
         }
       }
 
@@ -398,13 +401,13 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         }
       }
 
-      js.PropertyDef(static, js.StringLiteral(jsName), getterBody,
+      js.PropertyDef(static, genPropertyName(jsName), getterBody,
           setterArgAndBody)
     }
 
     /** generates the exporter function (i.e. exporter for non-properties) for
      *  a given name */
-    private def genExportMethod(alts0: List[Exported], jsName: String,
+    private def genExportMethod(alts0: List[Exported], jsName: JSName,
         static: Boolean): js.MethodDef = {
       assert(alts0.nonEmpty,
           "need at least one alternative to generate exporter method")
@@ -414,7 +417,10 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
       val alts = {
         // toString() is always exported. We might need to add it here
         // to get correct overloading.
-        if (jsName == "toString" && alts0.forall(_.params.nonEmpty))
+        val needsToString =
+          jsName == JSName.Literal("toString") && alts0.forall(_.params.nonEmpty)
+
+        if (needsToString)
           ExportedSymbol(Object_toString) :: alts0
         else
           alts0
@@ -526,7 +532,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         }
       }
 
-      js.MethodDef(static, js.StringLiteral(jsName),
+      js.MethodDef(static, genPropertyName(jsName),
           formalArgs, jstpe.AnyType, Some(body))(OptimizerHints.empty, None)
     }
 
@@ -698,7 +704,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
 
       val cls = jstpe.ClassType(encodeClassFullName(currentClassSym))
       val receiver = js.This()(jstpe.AnyType)
-      val nameString = js.StringLiteral(jsNameOf(sym))
+      val nameString = genExpr(jsNameOf(sym))
 
       if (jsInterop.isJSGetter(sym)) {
         assert(allArgs.isEmpty)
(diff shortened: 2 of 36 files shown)


gzm0:
Review by @sjrd

library/src/main/scala/scala/scalajs/js/Symbol.scala
```diff
@@ -0,0 +1,43 @@
...
+    def apply(description: String): Symbol = js.native
+
+    def `for`(key: String): Symbol = js.native
+    def keyFor(sym: Symbol): String = js.native
+
+    val iterator: Symbol = js.native
+  }
+
+  @inline def apply(): Symbol = NativeSymbol()
+  @inline def apply(description: String): Symbol = NativeSymbol()
```
sjrd: `NativeSymbol(description)`?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -260,33 +260,36 @@ abstract class GenJSCode extends plugins.PluginComponent
          * - lambdas for scala.FunctionN. This is only an optimization and may
          *   fail. In the case of failure, we fall back to generating a
          *   fully-fledged Scala class.
+         * - Implementation classes for native JS classes. The only useful
+         *   thing they contain are symbol forwarders for @JSSymbol access.
+         *   We want to emit the symbol forwarders as static methods on the
+         *   native JS class so we do not need to generate another class file.
```
sjrd: another *sjsir* file?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -704,8 +706,56 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      // Generates symbol forwarders
+      def gen(tree: Tree): List[js.MethodDef] = tree match {
+        case Template(_, _, body) => body.flatMap(gen)
+
+        case dd: DefDef if jsInterop.isSymbolForwarder(dd.symbol) =>
+          val sym = dd.symbol
+          val patchedDef = {
+            if (scalaUsesImplClasses && sym.owner.isTraitOrInterface) {
+              assert(sym.isDeferred, "Found non-abstract method in trait at " +
+                s"${dd.pos}: ${sym.fullName}")
```
sjrd: 4 spaces

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -704,8 +706,56 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+
+              /* We grab the body from the implementation class. This does not
+               * work so directly in general, since the parameter symbols and
```
sjrd: "so directly" does not sound right.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -704,8 +706,56 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+               * the `this` reference would be wrong in the body. Here it works,
+               * because we have a static, parameterless method.
+               */
+              val nrhs = implMethodsByName(sym.unexpandedName.encoded).rhs
+              treeCopy.DefDef(dd, dd.mods, dd.name, dd.tparams,
+                  dd.vparamss, dd.tpt, nrhs)
+            } else {
+              assert(!sym.isDeferred, "Found an abstract symbol forwarder at " +
+                  s"${dd.pos}: ${sym.fullName}")
+              dd  // No patching necessary.
```
sjrd: There are 2 spaces between `dd` and `//`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1255,7 +1305,8 @@ abstract class GenJSCode extends plugins.PluginComponent
         if (scalaPrimitives.isPrimitive(sym) &&
             !jsPrimitives.shouldEmitPrimitiveBody(sym)) {
           None
-        } else if (isAbstractMethod(dd)) {
+        } else if (isAbstractMethod(dd) && !(scalaUsesImplClasses &&
+            jsInterop.isSymbolForwarder(sym))) {
```
sjrd: Is `scalaUsesImplClasses &&` necessary here?
If yes, can you put it on the next line, so that the line break goes with the outer-most `&&` rather than the inner-most one?

sjrd:
In general, I'm concerned that we add this for native JS classes without knowing what to do about Scala.js-defined JS classes. I would like both things to be addressed at the same time.

For Scala.js-defined JS classes, it means we need a definition-time access to calling the mentioned symbol. This in turn means that we probably need to create Scala.js-defined JS classes lazily, when they are first required. (this might mean immediately if they are exported).

gzm0:
> In general, I'm concerned that we add this for native JS classes without knowing what to do about Scala.js-defined JS classes. I would like both things to be addressed at the same time.

That is a very good point :-/ I already had the feeling that we should impose some restrictions on the contents of `@JSSymbol`, but I wasn't able to come up with anything sensible...

sjrd:
Yeah, managed to make the CI pass again. Now on to write more tests!

sjrd:
Since we decided to restrict `@JSName` not to admit symbols for top-level classes and objects, the changes to the format of `JSNativeLoadSpec` were not necessary anymore. I have reverted them. I'm keeping a backup commit with the changes in my branch https://github.com/sjrd/scala-js/tree/save-jsnativeloadspec-with-propertynames

gzm0:
Euh, how do you handle nested classes / objects that have a `@JSName(sym)`? I thought we wanted to allow these...

sjrd:
Nested classes are not supported at all so far, even with string names. Or am I getting confused? 

We need to test objects, though, indeed. They probably work by virtue of being handled uniformly with vals already for stringly named, though.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -693,7 +694,7 @@ abstract class PrepJSInterop extends plugins.PluginComponent
         }
 
         val ownerLoadSpec = jsInterop.jsNativeLoadSpecOf(sym.owner)
-        val jsName = jsInterop.jsNameOf(sym)
+        val JSName.Literal(jsName) = jsInterop.jsNameOf(sym)
```
gzm0: In the current state, I expect this line to throw a `MatchError` for the following:

```scala
@js.native
object Foo extends js.Any {
  @JSName(sym)
  @js.native
  class A extends js.Any
}
```

sjrd: Ah. That may well be the case, indeed. I hadn't though about that :-s

I suspect we don't need to change the IR for that, though. `js.constructor[Foo.A]` should translate to `JSBracketSelect(JSLoadModule("Foo$"), <refTo sym>)`

gzm0: Hmm... But then you just need to attach the location of the nested class with an annotation, no? Not sure that is much cleaner. Also it is a step backwards for binary compatibility, no? (Can't change the jsname of a nested class, top-level class is fine though)...

sjrd: It is not cleaner within the compiler, no. But it avoids complicating the IR, which I think is worth doing.

It is a step back for bincompat, yes. But as long as we don't have a complete solution for perfect backward compat of all `@JSName`, including term members, I'm not sure it's worth fighting for. Also, such a change should only happen if the underlying JS library changes its public API, which is a seriously breaking change to begin with. We can hardly promise more binary compat than the underlying source compat.

gzm0: Fair enough. I'm fine with simplifying this in the IR.

sjrd: It seems this is going to be much harder than I thought, because of the comment here: https://github.com/scala-js/scala-js/blob/51b537ae4c64783db016ca3e90e675a895d50992/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala#L566-L569

I'm not sure yet how to make scalac do what we want around here ...

gzm0: Well.. Yet another internal annotation, no? But this annotation needs to support an arbitrary mix of symbols and strings :S

sjrd: Yes but that would be terribly difficult to jungle with wrt. backward binary compat of the scalac Symbols at this point.

I'm looking in the direction of using a scalac `InfoTransform`. It might save the situation.

gzm0: Ok. Just to make sure we don't need a ton of git juggling afterwards: Can you make sure you implement that change in a separate commit (I mean the "emit subselects at callsite change").

gzm0: Other question: How are you going to implement the desugaring of `LoadJSConstructor` for a class like this? Or do we not want to support it anymore for these classes (but how to remain backwards compatible)?

sjrd: In fact I realized that that it *needs* to be in the IR. Otherwise we cannot implement `java.lang.Class.isInstance` for these classes. Since there's no language reason not to support it, we need to.

So I'll revert to the IR changes you had made.

sjrd: Ah! But those IR changes also require to change the `Analyzer` in a non-obvious way. Because reaching a `classOf[Foo.Bar]` requires to reach the contents of the `JSName.Computed`s in the load spec of `Foo.Bar`. And that requires more infos, and that's non trivial.

At this point I'm inclined to disallow `@JSName(sym)` on native `object`s and `class`es in native `object`s, as an implementation restriction. It's less critical than the rest, and obviously we need more time to get a good grasp of what to do here.

gzm0: SGTM. It seems like just allowing `@JSName(sym)` on term members is a good initial step and indeed the most critical thing right now. We probably rethink the global import and naming story in general.

sjrd:
I guess it's complete, then :)

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -359,9 +360,10 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     }
 
     def genJSConstructorExport(alts: List[Symbol]): js.MethodDef =
-      genExportMethod(alts.map(ExportedSymbol), "constructor", static = false)
+      genExportMethod(alts.map(ExportedSymbol), JSName.Literal("constructor"),
+          static = false)
```
sjrd: This method body now needs braces.

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -69,6 +69,20 @@ trait JSGlobalAddons extends JSDefinitions
       val destination: ExportDestination
     }
 
+    sealed abstract class JSName {
+      def displayName: String
+    }
+
+    object JSName {
+      final case class Literal(name: String) extends JSName {
```
sjrd: These non-static `final case class`es cause compilation warnings. It's a petty thing that scalac likes to do. I don't like it but they should non-final.

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -69,6 +69,20 @@ trait JSGlobalAddons extends JSDefinitions
...
+        def displayName: String = name
+      }
+
+      final case class Computed(sym: Symbol) extends JSName {
```
sjrd: Same comment about `final` here.

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -188,7 +209,7 @@ trait JSGlobalAddons extends JSDefinitions
       assert(sym.isModuleClass,
           s"compat068FullJSNameOf called for non-module-class symbol $sym")
       sym.getAnnotation(JSFullNameAnnotation).flatMap(_.stringArg(0)) getOrElse {
-        jsNameOf(sym)
+        jsNameOf(sym).asInstanceOf[JSName.Literal].name
```
sjrd: Maybe a comment explaining why we can cast to `JSName.Literal` here? Basically because symbols did not exist in 0.6.8, so obviously if we're reading something from 0.6.8 it has to be a string.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1060,13 +1087,25 @@ abstract class PrepJSInterop extends plugins.PluginComponent
   /** Checks that argument to @JSName on [[sym]] is a literal.
    *  Reports an error on each annotation where this is not the case.
    */
-  private def checkJSNameLiteral(sym: Symbol): Unit = {
+  private def checkJSNameArgument(sym: Symbol): Unit = {
     for {
       annot <- sym.getAnnotation(JSNameAnnotation)
-      if annot.stringArg(0).isEmpty
+      argTree <- annot.args.headOption
```
sjrd: Any reason not to use `annot.args.head` here? If it doesn't have any argument, it should have failed to typecheck prior to getting to this phase.

examples/helloworld/HelloWorld.scala
```diff
@@ -8,87 +8,26 @@ package helloworld
...
-}
-
-@js.native
-trait DOMDocument extends js.Object {
-  def getElementById(id: String): DOMElement = js.native
-  def createElement(tag: String): DOMElement = js.native
+object A {
+  val sym1 = js.Symbol()
+  val sym2 = js.Symbol()
+  val sym3 = js.Symbol()
```
sjrd: We need to revert the changes in the hello world.

tools/scalajsenv.js
```diff
@@ -143,6 +143,7 @@ ScalaJS.t = {};         // Static fields
 ScalaJS.f = {};         // Default methods
 ScalaJS.n = {};         // Module instances
 ScalaJS.m = {};         // Module accessors
+ScalaJS.y = {};         // Symbol accessors
```
sjrd: This isn't needed anymore, is it?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -1498,8 +1571,10 @@ private[emitter] class JSDesugaring(semantics: Semantics,
           } else {
             val names = fields map (_._1)
             val items = fields map (_._2)
-            unnest(items) { (newItems, env) =>
-              redo(JSObjectConstr(names.zip(newItems)))(env)
+            unnestPropertyNames(names) { (newNames, env) =>
+              unnest(items) { (newItems, env) =>
```
sjrd: Hum ... we need to check what is the proper evaluation order of keys and values in ES6, in the presence of computed names. Hopefully it interleaves: key0, name0, key1, name1, etc. If that's the case the above way of unnesting is not correct, as it might evaluate some keys earlier than required.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -12,7 +12,7 @@ package org.scalajs.core.tools.linker.backend.emitter
 import scala.collection.mutable
 
 import org.scalajs.core.ir.ClassKind
-import org.scalajs.core.ir.Trees.{FieldDef, JSNativeLoadSpec}
+import org.scalajs.core.ir.Trees.{FieldDef, JSNativeLoadSpec, Ident}
```
sjrd: Useless import, it seems.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2262,7 +2344,7 @@ private[emitter] class JSDesugaring(semantics: Semantics,
 
   private[emitter] def genRawJSClassConstructor(className: String,
       spec: Option[JSNativeLoadSpec])(
-      implicit pos: Position): js.Tree = {
+      implicit globalKnowledge: GlobalKnowledge, pos: Position): js.Tree = {
```
sjrd: This appears not to be used. Similarly, callers of `genRawJSClassConstructor` should be checked whether their new `implicit globalKnowledge` was added only to call `genRawJSClassConstructor`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -319,7 +320,12 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
       }
     }
 
-    if (classDef.kind.isJSClass && name == "constructor" && !static) {
+    def isJSConstructor = pName match {
+      case StringLiteral("constructor") => true
+      case _                            => false
+    }
+
+    if (classDef.kind.isJSClass && isJSConstructor && !static) {
```
sjrd: `&& !static` should be moved as part of the definition of `isJSConstructor`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Experimentation in Firefox and reading the spec confirm that the evaluation order is interleaved between keys and values. So this is not correct.

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
gzm0: Ahhh... the `final` causes this... :P I was wondering...

sjrd:
Updated with a fixed treatment of `ComputedName`s in `JSObjectConstr`. It turns out they were completely broken. But since nothing could possibly *emit* them before, it was not detected. I have added a commit that makes sure we generate some, so that they are tested.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3923,12 +3924,27 @@ abstract class GenJSCode extends plugins.PluginComponent
...
 
+        def keyToPropName(key: js.Tree): js.PropertyName = key match {
+          case key: js.StringLiteral =>
+            key
+          case _ =>
+            /* ComputedName wants a logicalName, but it is not used for
+             * JSObjectConstr.
+             * TODO We should improve this.
+             */
+            js.ComputedName(key, "dummy")
```
sjrd: I had to do this ugly thing, here. @gzm0 Do you have a better idea?

gzm0: Calling the logical name "dummy" is OK-ish (I would suggest "local"). However, the `ComputeName` names should be unique, so we should probably add a counter or something.

sjrd: OK

sjrd:
We appear to have an IR checker situation in the bootstrap test:
```
[error] OptimizerCore.scala(662:32:If):
  ClassType(Lorg_scalajs_core_ir_Trees$PropertyName) expected but
  ClassType(s_Product) found
  for tree of type org.scalajs.core.ir.Trees$Block
```
It refers to this line: https://github.com/gzm0/scala-js/blob/7b7c3896c472e39d6978ca52092a87af604a998e/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala#L662

I don't think it's related to this PR per se. I think the bug existed before, but that the changes in this portion of the code brought triggered the bug. Probably scalac picking the "wrong" lub in the nested `if`. I guess scalac's back-end makes up for it with a cast (scalac loves to insert casts whenever things don't fit), but we don't. I think I'll side-step the issue right now with a slightly rearranged code, and file the issue for later.

sjrd:
Aaannd ... rebased and cleaned up the history!

Sending back the review by @gzm0 now ;)

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -1180,6 +1255,78 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+        @JSName(Sym.sym)
+        @js.native
+        $kind A extends js.Object
+      }
+      """ hasErrors
+      """
+        |newSource1.scala:11: error: Implementation restriction: @JSName with a js.Symbol is not supported on nested native classes and objects
+        |        @JSName(Sym.sym)
+        |         ^
+      """
```
gzm0: I'm not sure I would make the distinction between this error message and the one above (at this point).

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -1541,6 +1688,467 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+    """
+      |newSource1.scala:19: warning: A member of a JS class is overriding another member with a different JS name.
+      |
+      |def foo: Int in trait B with JSName 'Syms.sym1'
+      |    is conflicting with
+      |def foo: Int in trait A with JSName 'foo'
+      |
+      |    abstract class C extends A with B
+      |                   ^
+    """
```
gzm0: I never realized that this is only a warning. Will this become an error? If no, we should:

- Have a valid use case for it.
- A way to silence the warning.

Both of these are of course completely orthogonal to this PR.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSSymbolTest.scala
```diff
@@ -0,0 +1,345 @@
...
+
+  @Test def sjsdefined_with_overloaded_runtime_dispatch_methods(): Unit = {
+    val obj = new SJSDefinedOverloadedRuntimeDispatchMethod
+    assertEquals(5, obj.foo(2))
+    assertEquals("Hello World", obj.foo("World"))
+
+    val asTrait: OverloadedRuntimeDispatchMethodTrait = obj
+    assertEquals(6, asTrait.foo(3))
+    assertEquals("Hello Moon", asTrait.foo("Moon"))
+  }
```
gzm0: For all of the sjs defined classes, I think we should also check that accessing them via the dynamic interface works. (meaning for now, a handwritten dynamic interface).

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3950,31 +3960,27 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            // Build the resulting object
+            val result = js.JSObjectConstr(evalTuples.zipWithIndex.map {
+              case (evalTuple, index) =>
+                val tupRef = evalTuple.ref
+                val key = genApplyMethod(tupRef, js.Ident("$$und1__O"), Nil,
+                    jstpe.AnyType)
+                val value = genApplyMethod(tupRef, js.Ident("$$und2__O"), Nil,
+                    jstpe.AnyType)
+                keyToPropName(key, index) -> value
+            })
```
gzm0: Wouldn't you want to retain the literals where you can here?

compiler/src/main/scala/org/scalajs/core/compiler/JSTreeExtractors.scala
```diff
@@ -22,45 +22,36 @@ object JSTreeExtractors {
...
-        LitNamedExtractor.genNamedLitExtract(exprs, Nil, true)
+    object Tuple2List {
+      def unapply(exprs: List[Tree]): Option[List[(Tree, Tree)]] = {
+        val tuples = exprs.collect {
+          case Tuple2(key, value) => (key, value)
+        }
+        if (tuples.size == exprs.size)
+          Some(tuples)
+        else
+          None
```
gzm0: OMG, I'm turning to the dark side... I just thought that this would be a prime use-case for 

```scala
exps.sequence(Tuple2.unapply _)
```

except that Scala's List doesn't have a sequencing operator...

library/src/main/scala/scala/scalajs/js/Symbol.scala
```diff
@@ -0,0 +1,41 @@
...
+  @js.native
+  private object NativeSymbol extends js.Object {
+    def apply(): Symbol = js.native
+    def apply(description: String): Symbol = js.native
+
+    def `for`(key: String): Symbol = js.native
+    def keyFor(sym: Symbol): js.UndefOr[String] = js.native
+
+    val iterator: Symbol = js.native
+  }
```
sjrd: Because at some point we realized we would have want `js.Array` to be non-native, to include custom things in there. I'm mostly anticipating the need to evolve `Symbol` into more than just its facade.

Also, for 1.0 I think we should turn the type `js.Symbol` into a hijacked class. We have hijacked classes to represent all other JS primitive types, but not `symbol`. I think it would make sense. In that case, it would also be weird, possibly problematic, for the companion object to be a native.

But I don't have very strong feelings about this.

library/src/main/scala/scala/scalajs/js/Symbol.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  @inline def apply(): Symbol = NativeSymbol()
+  @inline def apply(description: String): Symbol = NativeSymbol(description)
+
+  @inline def forKey(key: String): Symbol = NativeSymbol.`for`(key)
+  @inline def keyFor(sym: Symbol): js.UndefOr[String] = NativeSymbol.keyFor(sym)
+
+  @inline def iterator: Symbol = NativeSymbol.iterator
```
sjrd: Ouch ^^

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5144,6 +5144,40 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+      val module = genLoadModule(sym.owner)
+
+      if (isRawJSType(sym.owner.tpe)) {
+        if (!isScalaJSDefinedJSClass(sym.owner) || isExposed(sym))
+          genJSCallGeneric(sym, module, args = Nil, isStat = false)
+        else
+          genApplyJSClassMethod(module, sym, arguments = Nil)
+      } else {
+        genApplyMethod(module, sym, arguments = Nil)
+      }
```
sjrd: I didn't really try that, because things are decomposed early into JS/Scala calls in the `genApply` giant dispatch. So seeing them all together like this never came up before. I think it's OK like that, if a bit ugly.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: The optimizer will take care of that, if necessary.

compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
```diff
@@ -239,6 +239,11 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
   def needsModuleClassSuffix(sym: Symbol): Boolean =
     sym.isModuleClass && !foreignIsImplClass(sym)
 
+  def encodeComputedNameIdentity(sym: Symbol): String = {
```
sjrd: `PropertyName#encodedName` seems a reasonable place for that.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
sjrd: The error message above does not work here, because `A` really is a member of a JS type (`Enclosing`).

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
sjrd: Basically all warnings of PrepJSInterop will become errors in 1.0. #1111

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -931,8 +931,42 @@ object Printers {
       printEscapeJS(ident.name, out)
 
     private final def print(propName: PropertyName): Unit = propName match {
-      case lit: StringLiteral => print(lit: Tree)
-      case ident: Ident       => print(ident)
+      case lit: StringLiteral        => print(lit: Tree)
+      case ident: Ident              => print(ident)
```
sjrd: We don't. This is a mistake.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -931,8 +931,42 @@ object Printers {
...
+        case JSNativeLoadSpec.Global(path) =>
+          print("<global>")
+          printPath(path)
+
+        case JSNativeLoadSpec.Import(module, path) =>
+          print("import(")
+          print(module)
+          print(')')
+          printPath(path)
+      }
```
sjrd: Yes, it was at some point. But it's indeed not relevant anymore. I'll make a separate commit.

ir/src/main/scala/org/scalajs/core/ir/Serializers.scala
```diff
@@ -980,8 +974,21 @@ object Serializers {
...
+          case TagPropertyNameIdent =>
+            readIdent()
+
+          case TagPropertyNameStringLiteral =>
+            readTree().asInstanceOf[StringLiteral]
+
+          case TagPropertyNameComputedName =>
+            ComputedName(readTree(), readString())
+        }
+      }
```
sjrd: I did review it quite thoroughly, but yes we should definitely try reading older IR.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSSymbolTest.scala
sjrd: Hum, yes. It's kind of a consequence of the fact that reading via the trait works + the previous test passes. But it's not very self-contained. I'll add a "dynamic"-based test as well.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -508,36 +510,48 @@ private[emitter] final class ScalaJSClassEmitter(semantics: Semantics,
...
-    js.Assign(select, value)
+    js.Assign(genPropSelect(obj, name), value)
   }
 
   /** Generate `obj.name = value` */
-  def genAddToObject(obj: js.Tree, name: PropertyName,
-      value: js.Tree)(implicit pos: Position): js.Tree = {
+  def genAddToObject(className: String, obj: js.Tree, name: PropertyName,
+      value: js.Tree)(
+      implicit globalKnowledge: GlobalKnowledge, pos: Position): js.Tree = {
```
sjrd: Hum yes, that's intentional. What's weird about it?

library/src/main/scala/scala/scalajs/js/Symbol.scala
gzm0: Don't you mean turning `scala.Symbol` into a hijacked class? I do get the story for `js.Array`, but `js.Array` is also much closer to the runtime. And I think between 0.6 and 1.0 its ok to have the theoretically source incompatible change of making `js.Symbol` not extend `js.Any` anymore.

library/src/main/scala/scala/scalajs/js/Symbol.scala
gzm0: And now that I think of it, we should probably add all well known symbols here.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Ok. But what we should probably do, is add tests for all code paths here. Currently we only read symbols for definitions from Scala modules.

We probably should also have this nasty test case here:

```scala
@ScalaJSDefined
object Foo extends js.Any {
  val sym = js.Symbol()

  @JSName(sym)
  def yolo(): Int = 1
}
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
gzm0: It seems like this would make more sense:

```scala
def genAddToObject(className: String, obj: js.Tree, name: PropertyName,
    value: js.Tree)(implicit globalKnowledge: GlobalKnowledge,
    pos: Position): js.Tree =
```

Though looking at it, it looks weird as well :P and there is definitely value in having the implicit on the new line :)

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
gzm0: Yeah. Also since this works for objects, I'm fine with this.

library/src/main/scala/scala/scalajs/js/Symbol.scala
sjrd: No I really mean `js.Symbol`. `scala.Symbol` we already ruled out, because their semantics are too different from `js.Symbol` (#1541). However we should be able to do things like
```scala
x match {
  case x: js.Symbol =>
}
```
and we can only do that if `x.isInstanceOf[js.Symbol]` has a meaningful spec, i.e., `type of x === 'symbol'`. And that, we can only do if `js.Symbol` is a hijacked class.

> I do get the story for js.Array, but js.Array is also much closer to the runtime.

IMO `js.Symbol` is *much* closer to the run-time than `js.Array`. For `js.Array` we're only optimizing its `apply` method a bit. `js.Symbol`s creep into the very definition of the language, now that they're supported in `@JSName`.

> And I think between 0.6 and 1.0 its ok to have the theoretically source incompatible change of making js.Symbol not extend js.Any anymore.

OK. Let's have an `@js.native object Symbol` for now, then.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: > We probably should also have this nasty test case here:

Well, that one doesn't work. It stack overflows. It makes sense.

sjrd:
Updated.

sjrd:
I locally tested that we can depend on uTest and ScalaTest, and use them to test things with this branch. I believe this should give enough confidence regarding the ability to read old IR.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Should we warn? Just like bad initialization order? (Not in this PR for sure).

gzm0:
LGTM (I can't approve the PR since it is mine).

gzm0:
I'll let you approve and merge (unless you have concerns :)).

gzm0:
Wooot :tada:

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: #2778

PR #2660, opened 2016-11-21
Fix #1480: Mark js.Dictionary.empty @inline

### LINKED ISSUE #1480: js.Dictionary.empty should be marked @inline

gzm0:


gzm0:
Review by @sjrd

PR #2661, opened 2016-11-22
Fix #1655: Allow export of classes inside objects

### LINKED ISSUE #1655: Allow export of classes inside object
In Akka classes are, most of the time, encapsulated in an outer object when they aren't exposed to the user:

``` scala
object Outer {
  class Inner {}
}
```

It's a fairly standard pattern, but unfortunately the current implementation of `@JSExport` prevents from exporting nested classes.
If we could improve the annotation system to allow for such a behaviour, it would help tremendously!
Thanks :)

gzm0:
This is a trivial extension to allow exports of static classes / modules. The only non-trivial issues is name choice, which we effectively circumvent by forcing the user to choose a name.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
@@ -75,7 +75,7 @@ trait PrepJSExports { this: PrepJSInterop =>
         err("You may not export an abstract class")
       else if (clsSym.isLocalToBlock)
         err("You may not export a local class")
-      else if (clsSym.isNestedClass)
+      else if (!clsSym.isStatic)
         err(s"You may not export a nested class. $createFactoryInOuterClassHint")
       else {
         jsInterop.registerForExport(baseSym, exports)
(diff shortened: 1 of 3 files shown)


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -274,6 +274,17 @@ trait PrepJSExports { this: PrepJSInterop =>
...
 
+      // Don't allow nested class / module exports without explicit name.
+      def isStaticNested = {
+        val clsSym = if (sym.isClass) sym else sym.owner
+        clsSym.isNestedClass && clsSym.isStatic && !clsSym.isLocalToBlock
+      }
+
+      if (!isMember && !hasExplicitName && isStaticNested) {
+        reporter.error(annot.pos,
+            "Must set an explicit name for exports of nested classes.")
```
sjrd: This sentence needs a subject ("You").

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -274,6 +274,17 @@ trait PrepJSExports { this: PrepJSInterop =>
             "This will be enforced in 1.0.")
       }
 
+      // Don't allow nested class / module exports without explicit name.
+      def isStaticNested = {
+        val clsSym = if (sym.isClass) sym else sym.owner
```
sjrd: When is this not a class?

sjrd: Ah, when it's a constructor. Maybe a comment about that?

gzm0:
Updated.

PR #2662, opened 2016-11-25
Add an annotation `@JSOptional` for optional members in JS types.

sjrd:
In a JS type, and particularly in a Scala.js-defined JS type, vals,
vars and defs without parentheses can be annotated with
`@JSOptional`. In that case, they must be concrete, and their rhs
must be `js.undefined`. This is also valid in a Scala.js-defined JS
trait, and is the only way to define a concrete member in such a
trait.

An `@JSOptional` member does not create a field nor getter in the
object or class. Instead, it gets a value of `undefined` by virtue
of JavaScript using `undefined` for missing fields. In case of a
`var`, the field can be created later by explicit assignment. For
`val`s and `def`s, it can be created in subclasses with an
`override val` or `override def`.

This is particularly useful to model "configuration objects" as
used in many JavaScript APIs. A Scala.js-defined JS trait with
`JSOptional` fields can define a typed API for the configuration
object. It is then easy to create a configuration object with only
a few fields set by defining `override val`s. Other fields stay
optional (by inheritance) and are not created on the resulting
JavaScript object.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -281,12 +281,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           val isPrimitive =
             isPrimitiveValueClass(sym) || (sym == ArrayClass)
 
-          /* Similarly, do not emit code for impl classes of raw JS traits. */
-          val isRawJSImplClass =
-            sym.isImplClass && isRawJSType(
-                sym.owner.info.decl(sym.name.dropRight(nme.IMPL_CLASS_SUFFIX.length)).tpe)
-
-          if (!isPrimitive && !isRawJSImplClass) {
+          if (!isPrimitive && !isRawJSImplClass(sym)) {
             withScopedVars(
                 currentClassSym          := sym,
                 unexpectedMutatedFields  := mutable.Set.empty,
@@ -468,6 +463,8 @@ abstract class GenJSCode extends plugins.PluginComponent
               /* Exposed accessors must not be emitted, since the field they
                * access is enough.
                */
+            } else if (sym.hasAnnotation(JSOptionalAnnotation)) {
+              // Optional methods must not be emitted
             } else {
               generatedMethods ++= genMethod(dd)
 
@@ -811,6 +808,7 @@ abstract class GenJSCode extends plugins.PluginComponent
       (for {
         f <- classSym.info.decls
         if !f.isMethod && f.isTerm && !f.isModule
+        if !f.hasAnnotation(JSOptionalAnnotation)
       } yield {
         implicit val pos = f.pos
 
@@ -1722,38 +1720,43 @@ abstract class GenJSCode extends plugins.PluginComponent
           val sym = lhs.symbol
           if (sym.isStaticMember)
             abort(s"Assignment to static member ${sym.fullName} not supported")
-          val genRhs = genExpr(rhs)
-          lhs match {
-            case Select(qualifier, _) =>
-              val ctorAssignment = (
-                  currentMethodSym.isClassConstructor &&
-                  currentMethodSym.owner == qualifier.symbol &&
-                  qualifier.isInstanceOf[This]
-              )
-              if (!ctorAssignment && !suspectFieldMutable(sym))
-                unexpectedMutatedFields += sym
+          if (rhs.symbol == Runtime_optionalUndefined) {
+            // Discard initialization assignment to @JSOptional field
+            js.Skip()
+          } else {
+            val genRhs = genExpr(rhs)
+            lhs match {
+              case Select(qualifier, _) =>
+                val ctorAssignment = (
+                    currentMethodSym.isClassConstructor &&
+                    currentMethodSym.owner == qualifier.symbol &&
+                    qualifier.isInstanceOf[This]
+                )
+                if (!ctorAssignment && !suspectFieldMutable(sym))
+                  unexpectedMutatedFields += sym
 
-              val genQual = genExpr(qualifier)
+                val genQual = genExpr(qualifier)
 
-              if (isScalaJSDefinedJSClass(sym.owner)) {
-                val genLhs = if (isExposed(sym))
-                  js.JSBracketSelect(genQual, js.StringLiteral(jsNameOf(sym)))
-                else
-                  js.JSDotSelect(genQual, encodeFieldSym(sym))
-                val boxedRhs =
-                  ensureBoxed(genRhs,
-                      enteringPhase(currentRun.posterasurePhase)(rhs.tpe))
-                js.Assign(genLhs, boxedRhs)
-              } else {
+                if (isScalaJSDefinedJSClass(sym.owner)) {
+                  val genLhs = if (isExposed(sym))
+                    js.JSBracketSelect(genQual, js.StringLiteral(jsNameOf(sym)))
+                  else
+                    js.JSDotSelect(genQual, encodeFieldSym(sym))
+                  val boxedRhs =
+                    ensureBoxed(genRhs,
+                        enteringPhase(currentRun.posterasurePhase)(rhs.tpe))
+                  js.Assign(genLhs, boxedRhs)
+                } else {
+                  js.Assign(
+                      js.Select(genQual, encodeFieldSym(sym))(toIRType(sym.tpe)),
+                      genRhs)
+                }
+              case _ =>
+                mutatedLocalVars += sym
                 js.Assign(
-                    js.Select(genQual, encodeFieldSym(sym))(toIRType(sym.tpe)),
+                    js.VarRef(encodeLocalSym(sym))(toIRType(sym.tpe)),
                     genRhs)
-              }
-            case _ =>
-              mutatedLocalVars += sym
-              js.Assign(
-                  js.VarRef(encodeLocalSym(sym))(toIRType(sym.tpe)),
-                  genRhs)
+            }
           }
 
         /** Array constructor */
@@ -2133,7 +2136,15 @@ abstract class GenJSCode extends plugins.PluginComponent
       val sym = fun.symbol
 
       if (isScalaJSDefinedJSClass(currentClassSym)) {
-        genJSSuperCall(tree, isStat)
+        if (sym.isMixinConstructor) {
+          /* Do not emit a call to the $init$ method of JS traits.
+           * This exception is necessary because @JSOptional fields cause the
+           * creation of a $init$ method, which we must not call.
+           */
+          js.Skip()
+        } else {
+          genJSSuperCall(tree, isStat)
+        }
       } else {
         val superCall = genApplyMethodStatically(
             genThis()(sup.pos), sym, genActualArgs(sym, args))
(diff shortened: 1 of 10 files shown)


sjrd:
Review by @gzm0 

I've had this on my mind for months, but I never actually implemented it because the cost of having to wrote the explicit type when overriding vals caused this feature to be inconvenient in practice. However, with the new type inference for overriding vals in Scala 2.12, the explicit type is no longer necessary, which finally makes it all worthwhile.

/cc @jducoeur This is supposed to address the last use case for which `JSOptionsBuilder` is necessary.

jducoeur:
Okay, cool.  Once this is released, I'll put an item on my to-do list to evaluate whether JSOptionBuilder is now entirely redundant and should be deprecated...

gzm0:
So just to clarify my understanding: This is essentially to allow this pattern:

```scala
@ScalaJSDefined
trait MyOptions extends js.Object {
  @JSOptional val a: UndefOr[Int] = js.undefined
  @JSOptional val b: UndefOr[Int] = js.undefined
  @JSOptional val c: UndefOr[Int] = js.undefined
}

foo(new MyOptions { override val a = 1 })
```

In this case, why do we require an annotation to trigger this behavior (and not just trigger it implicitly on traits)? I'll discuss a couple of points below that might be controversial with respect to this proposal:

### Readability

IMO the version without `@JSOptional` is easier to understand to Scala programmers, since they need not learn about the `@JSOptional` annotation. The fact that we do not emit anything instead of `js.undefined` is an implementation detail for most use cases.
A JS only programmer will need to learn the concept of trait first anyways (and frankly would not be helped by `@JSOptional`s either).

```scala
@ScalaJSDefined
trait MyOptions extends js.Object {
  val a: UndefOr[Int] = js.undefined
  val b: UndefOr[Int] = js.undefined
  val c: UndefOr[Int] = js.undefined
}
```

### Classes
Dropping `@JSOptional` will not allow triggering this behavior in Scala.js defined classes anymore. IIUC, this will only make a difference if something verifies the existence of the field rather than just its value (note that field enumeration is broken-ish by the precedent of private fields already).

I do not know how much of a use case this is for JS libraries. In any case the usability on the Scala.js side seems terrible, since now the options for creating a field are:

```scala
@ScalaJSDefined
class A extends js.Object {
  val a: UndefOr[Int] = 1 // Set it
  val b: UndefOr[Int] = js.undefined // "Set it" to undefined
  @JSOptional val c: UndefOr[Int] = js.undefined // "Really" undefined
}
```

If it is an actual use case to distinguish the latter two cases at object creation site, I'm not sure `@JSOptional` is the best name for the annotation. I'd rather have something like `@JSNoField`; a name that implies that no field should be created at all.

### Trait mixins
If we ever decide to support some sort of trait mixin, this would pose similar challenges for traits.

WDYT?

gzm0:
@jducoeur 

> Once this is released, I'll put an item on my to-do list to evaluate whether JSOptionBuilder is now entirely redundant and should be deprecated...

I'm having a hunch that @sjrd would like you to do this before we commit to the feature ;)

Essentially, once the feature is out, we'll need to support it. If it doesn't fully fulfil its purpose, it's much much easier to change it now than once it is out. If you do not have time, that is of course fine. Then maybe one of us should do this?

jducoeur:
Hmm.  It's not a great time for me to commit sight-unseen -- I'm very distracted with a major Querki release coming up, and prepping for the trip to England (and subsequent Christmas vacation), so I'm not certain that I'm thinking it through properly.  So this is all kind of off-the-top-of-my-head.

From your comment above, I'm not sure that you've come across the key use case, and there may be some misunderstanding on one of our parts.  If I'm understanding the proposal correctly, the point of `@JSOptional` is that it would allow you to *omit* that field in concrete instantiations, which isn't currently possible.  That's a huge deal with building "options" objects to pass into JS libraries -- these objects often involve anywhere from three to 40 fields, *all* of them optional, so creating one involves either tons of boilerplate at the callsite, or fragile builder functions.

As of where things currently stand, `JSOptionBuilder` is the lowest-boilerplate way to create one of these things.  I belive @sjrd is saying that the combination of `@ScalaJSDefined` and `@JSOptional` will be at least as good, likely better, and I think quite that's plausible (since there is considerable per-trait boilerplate for `JSOptionBuilder`), but I'm going to have to work through some examples to see if there are any gotchas.  That's hard to do with confidence in my head, without being able to try it on real code -- hence my reluctance to firmly commit.  I think it's *likely* that he's correct, but I am so deeply dependent on this (I created `JSOptionBuilder` because I needed it desperately) that it's hard to me to say for sure until I've taken a chunk of Querki code and actually converted it...

jducoeur:
Also, keep in mind that my opinion is of limited importance -- `JSOptionBuilder` is rather idiosyncratic opt-in code, so people are very likely to prefer the built-in `@JSOptional` if it suits their needs.  I suspect that even if I disagreed, `JSOptionBuilder` would gradually fall into disuse.

Really, I think that the only reason I'm really relevant at all is because of the couple of places that `jquery-facade` uses `JSOptionBuilder`, which will need to get deprecated and switched over somehow.  (I *think* this should be a reasonably straightforward deprecation cycle, since `JSOptionBuilder` depends on an object and `@JSOptional` on a trait, but I won't be sure until I try it...)

sjrd:
@gzm0 

> So just to clarify my understanding: This is essentially to allow this pattern

That is indeed the main intended use case. I tried to generalize it a bit.

The alternative of automatically applying this behavior in JS traits bothers me mostly because whether a field is declared in a trait or in a class/object changes its semantics. There isn't really a precedent for anything like that in Scala.js nor Scala/JVM. Maybe that isn't as bad as it seems to me, but it's worth taking into account. That's why I had chosen an annotation for that. I agree that there isn't many use cases for optional fields in classes and objects (if at all). Once I had the annotation, it felt better to make it work for classes as well as traits.

@jducoeur

I think @gzm0 understands your case, as highlighted in his first example.

It's fine if you do not have time to look into this. Maybe you could give us a pointer or two to specific (challenging) declaration and use sites of `JSOptionsBuilder` that we could try to "port".

sjrd:
> Really, I think that the only reason I'm really relevant at all is because of the couple of places that jquery-facade uses JSOptionBuilder

It's more than that. IMO your experience with a lot of such options objects would be very welcome in identifying whether the language feature indeed solves the problem it is intended to solve.

julienrf:
I understand the concerns about performance, but from a usability point of view, IMHO, if we compare this facade proposed by @gzm0:

~~~ scala
@ScalaJSDefined
trait MyOptions extends js.Object {
  @JSOptional val a: UndefOr[Int] = js.undefined
  @JSOptional val b: UndefOr[Int] = js.undefined
  @JSOptional val c: UndefOr[Int] = js.undefined
}

foo(new MyOptions { override val a = 1 })
~~~

With this one:

~~~ scala
case class MyOptions(
  @(JSExport @field) a: UndefOr[Int] = js.undefined,
  @(JSExport @field) b: UndefOr[Int] = js.undefined,
  @(JSExport @field) c: UndefOr[Int] = js.undefined
)

foo(MyOptions(a = 1))
~~~

I prefer the second one (again, I’m comparing the *use* site). The downside is that we have to annotate the annotation, to precise that we want to annotate the field rather than the constructor parameter.

sjrd:
@julienrf case classes cannot extend `js.Any`, and so they are out of the equation for APIs that they are given POJOs (Plain Old JavaScript Object). This is a semantic concern, not a performance one.

julienrf:
Can you elaborate on what can go wrong in case we pass a case class instance to an API that expects a POJO?

sjrd:
It will simply throw a `TypeError` ...

julienrf:
Oh, OK, I got it, I didn’t notice that in your previous comment you were talking about *checking* that the passed argument is a POJO (this happens for instance when the same API accepts a `String` or a POJO).

jducoeur:
Hmm.  The most *important* use case is probably [JQueryAjaxSettings](https://github.com/jducoeur/jquery-facade/blob/master/src/main/scala/org/querki/jquery/JQueryAjaxSettings.scala), but since I don't use that directly (I do all of my AJAX work through Play's JS machinery), I don't have an example of usage.  (Indeed, since I only finished it a few months ago, I'm not certain whether it's being used yet.)

Within Querki, the chewiest examples are probably the JQueryUI stuff -- those were the facades that broke the camel's back in the first place and forced me to create `JSOptionBuilder`.  A good example is [the JQueryUISortableFacade](https://github.com/jducoeur/Querki/blob/master/querki/scalajs/src/main/scala/org/querki/facades/jqueryui/JQueryUISortableFacade.scala), which corresponds to [this library](https://jqueryui.com/sortable/), and is actually used in the code [here](https://github.com/jducoeur/Querki/blob/master/querki/scalajs/src/main/scala/querki/display/input/SortableList.scala#L162-L178) and [here](https://github.com/jducoeur/Querki/blob/master/querki/scalajs/src/main/scala/querki/editing/PropertySection.scala#L48-L62).  (This facade really ought to get released as part of querki.org, but I haven't had the cycles to deal yet.)

My intuition is that you're right -- that the defining site involves rather less boilerplate with `@JSOptional` (indeed, quite a bit less in some cases), and the call site only slightly more.  But they do need to get worked through to check that belief...

sjrd:
Alternative implementing @gzm0's suggestion: #2663

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1722,38 +1720,43 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-                val genLhs = if (isExposed(sym))
-                  js.JSBracketSelect(genQual, js.StringLiteral(jsNameOf(sym)))
-                else
-                  js.JSDotSelect(genQual, encodeFieldSym(sym))
-                val boxedRhs =
-                  ensureBoxed(genRhs,
-                      enteringPhase(currentRun.posterasurePhase)(rhs.tpe))
-                js.Assign(genLhs, boxedRhs)
-              } else {
+          if (rhs.symbol == Runtime_optionalUndefined) {
```
sjrd: No, because we cannot distinguish from the *initial* assignment caused by the declaration of the field, versus other assignments elsewhere in the constructor (and methods). Consider:
```scala
@ScalaJSDefined
class Foo extends js.Object {
  @JSOptional var x: js.UndefOr[Int] = js.undefined
  x = 5
  println(x)
  x = js.undefined
  println(x)
}
```
would result in the following constructor definition:
```scala
def this() {
  this.x = js.undefined // needs to be dropped
  this.x = 5
  println(this.x)
  this.x = js.undefined // needs to be kept!
  println(this.x)
}
```

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -847,10 +872,10 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
         // Traits must be pure interfaces
         if (sym.owner.isTrait && sym.isTerm && !sym.isConstructor) {
-          if (!sym.isDeferred) {
+          if (!sym.isDeferred && !sym.hasAnnotation(JSOptionalAnnotation)) {
             reporter.error(tree.pos,
-                "A Scala.js-defined JS trait can only contain abstract members")
-          } else if (isPrivateMaybeWithin(sym)) {
+                "A Scala.js-defined JS trait can only contain abstract members and/or @JSOptional members")
+          } else if (sym.isMethod && isPrivateMaybeWithin(sym)) {
```
sjrd: Because now you can have a private field created by a concrete `val` or `var`. Previously we only had abstract vals and vars, for which `typer` does not create a private field.

sjrd:
> IMO one can consider what happens to fields in traits a semantic difference w.r.t. to what happens in abstract classes notably with respect to binary compatibility

Meh, I'm not sure binary compat "semantic" differences can be used to justify actual semantic differences happening even with full recompilation.

> (IIUC, this will still be the case in 2.12, since interfaces won't allow fields, right)?

It still *is* the case in 2.12. 2.12 is out, now ;)

> So yes, not requiring the annotation would introduce semantic differences between fields in traits and classes, but IMO they are subtle enough so that the average programmer need not bother with them too often, therefore not warranting the visual overhead of an annotation.

I think that convinces me most. Let's settle on #2663 then.

sjrd:
Superseded by #2663.

PR #2663, opened 2016-11-27
Allow concrete val/var/defs in JS traits with `= js.undefined`.

sjrd:
Normally, term members in Scala.js-defined JS traits must be abstract. This commit relaxes the restriction to allow val, vars and defs without parentheses to be concrete, if their right-hand-side is `js.undefined`.

Such fields and methods are not actually created on Scala.js-defined JS classes and objects that extend those traits. Instead, they get a value of `undefined` by virtue of JavaScript using `undefined` for missing fields. In case of a `var`, the field can be created later by explicit assignment. For `val`s and `def`s, it can be created in subclasses with an `override val` or `override def`.

This is particularly useful to model "configuration objects" as used in many JavaScript APIs. A Scala.js-defined JS trait with concrete `js.undefined` fields can define a typed API for the configuration object. It is then easy to create a configuration object with only a few fields set by defining `override val`s. Other fields stay optional (by inheritance) and are not created on the resulting JavaScript object.

This is an alternative to #2662.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -281,12 +281,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           val isPrimitive =
             isPrimitiveValueClass(sym) || (sym == ArrayClass)
 
-          /* Similarly, do not emit code for impl classes of raw JS traits. */
-          val isRawJSImplClass =
-            sym.isImplClass && isRawJSType(
-                sym.owner.info.decl(sym.name.dropRight(nme.IMPL_CLASS_SUFFIX.length)).tpe)
-
-          if (!isPrimitive && !isRawJSImplClass) {
+          if (!isPrimitive && !isRawJSImplClass(sym)) {
             withScopedVars(
                 currentClassSym          := sym,
                 unexpectedMutatedFields  := mutable.Set.empty,
@@ -468,6 +463,8 @@ abstract class GenJSCode extends plugins.PluginComponent
               /* Exposed accessors must not be emitted, since the field they
                * access is enough.
                */
+            } else if (sym.hasAnnotation(JSOptionalAnnotation)) {
+              // Optional methods must not be emitted
             } else {
               generatedMethods ++= genMethod(dd)
 
@@ -811,6 +808,7 @@ abstract class GenJSCode extends plugins.PluginComponent
       (for {
         f <- classSym.info.decls
         if !f.isMethod && f.isTerm && !f.isModule
+        if !f.hasAnnotation(JSOptionalAnnotation)
       } yield {
         implicit val pos = f.pos
 
@@ -2133,7 +2131,15 @@ abstract class GenJSCode extends plugins.PluginComponent
       val sym = fun.symbol
 
       if (isScalaJSDefinedJSClass(currentClassSym)) {
-        genJSSuperCall(tree, isStat)
+        if (sym.isMixinConstructor) {
+          /* Do not emit a call to the $init$ method of JS traits.
+           * This exception is necessary because @JSOptional fields cause the
+           * creation of a $init$ method, which we must not call.
+           */
+          js.Skip()
+        } else {
+          genJSSuperCall(tree, isStat)
+        }
       } else {
         val superCall = genApplyMethodStatically(
             genThis()(sup.pos), sym, genActualArgs(sym, args))
@@ -2346,7 +2352,15 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     def genTraitImplApply(method: Symbol, arguments: List[js.Tree])(
         implicit pos: Position): js.Tree = {
-      genApplyStatic(method, arguments)
+      if (method.isMixinConstructor && isRawJSImplClass(method.owner)) {
+        /* Do not emit a call to the $init$ method of JS traits.
+         * This exception is necessary because @JSOptional fields cause the
+         * creation of a $init$ method, which we must not call.
+         */
+        js.Skip()
+      } else {
+        genApplyStatic(method, arguments)
+      }
     }
 
     def genApplyJSClassMethod(receiver: js.Tree, method: Symbol,
@@ -5011,6 +5025,12 @@ abstract class GenJSCode extends plugins.PluginComponent
   /** Tests whether the given class is a JS native class. */
   private def isJSNativeClass(sym: Symbol): Boolean =
     isRawJSType(sym.tpe) && !isScalaJSDefinedJSClass(sym)
+
+  /** Tests whether the given class is the impl class of a raw JS trait. */
+  private def isRawJSImplClass(sym: Symbol): Boolean = {
+    sym.isImplClass && isRawJSType(
+        sym.owner.info.decl(sym.name.dropRight(nme.IMPL_CLASS_SUFFIX.length)).tpe)
+  }
 
   /** Tests whether the given member is exposed, i.e., whether it was
    *  originally a public or protected member of a Scala.js-defined JS class.
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -610,30 +610,52 @@ abstract class PrepJSInterop extends plugins.PluginComponent
         }
       }
 
-      // Check for overrides with different JS names - issue #1983
+      // Check for consistency of JS semantics across overriding
       for (overridingPair <- new overridingPairs.Cursor(sym).iterator) {
         val low = overridingPair.low
         val high = overridingPair.high
-        if (jsInterop.jsNameOf(low) != jsInterop.jsNameOf(high)) {
-          val pos = {
-            if (sym == low.owner) low.pos
-            else if (sym == high.owner) high.pos
-            else sym.pos
-          }
 
+        def errorPos = {
+          if (sym == low.owner) low.pos
+          else if (sym == high.owner) high.pos
+          else sym.pos
+        }
+
+        def memberDefString(membSym: Symbol): String =
+          membSym.defStringSeenAs(sym.thisType.memberType(membSym))
+
+        // Check for overrides with different JS names - issue #1983
+        if (jsInterop.jsNameOf(low) != jsInterop.jsNameOf(high)) {
           val msg = {
-            def memberDefString(membSym: Symbol) = {
-              membSym.defStringSeenAs(sym.thisType.memberType(membSym)) +
+            def memberDefStringWithJSName(membSym: Symbol) = {
+              memberDefString(membSym) +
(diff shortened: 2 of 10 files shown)


sjrd:
Review by @gzm0. Commits are meant to be squashed, obviously, so review the PR diff.

sjrd:
Updated with a proper commit and commit message.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -811,6 +808,7 @@ abstract class GenJSCode extends plugins.PluginComponent
       (for {
         f <- classSym.info.decls
         if !f.isMethod && f.isTerm && !f.isModule
+        if !f.hasAnnotation(JSOptionalAnnotation)
```
sjrd: Here they only happen in `$init$` methods of traits, and those methods are dropped altogether (as well as calls to them).

compiler/src/test/scala/org/scalajs/core/compiler/test/JSOptionalTest.scala
```diff
@@ -0,0 +1,141 @@
...
+    @ScalaJSDefined
+    trait A extends js.Object {
+      val a: js.UndefOr[Int]
+      def b: js.UndefOr[Int]
+    }
+
+    @ScalaJSDefined
+    trait B extends A {
+      override val a: js.UndefOr[Int] = js.undefined
+      override def b: js.UndefOr[Int] = js.undefined
```
sjrd: I do not think there is a good reason. I was planning to try and remove that restriction.

compiler/src/test/scala/org/scalajs/core/compiler/test/ScalaJSDefinedTest.scala
```diff
@@ -599,9 +599,9 @@ class ScalaJSDefinedTest extends DirectTest with TestHelpers {
     }
     """ hasErrors
     """
-      |newSource1.scala:7: error: A Scala.js-defined JS trait can only contain abstract members
+      |newSource1.scala:7: error: Members of Scala.js-defined JS traits must either be abstract, or their right-hand-side must be `js.undefined`.
```
sjrd: Ah good point! Yes, I think we should improve that message for methods with `()`.

library/src/main/scala/scala/scalajs/js/annotation/internal/JSOptional.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import scala.annotation.meta._
+
+/** IMPLEMENTATION DETAIL: Marks a member of a JavaScript trait as an optional
+ *  field, that must be emitted.
+ *
+ *  Do not use this annotation yourself.
+ */
+@field @getter @setter
+class JSOptional extends scala.annotation.StaticAnnotation
```
sjrd: The annotation is very important. In fact in this version we only truly test the annotation on *mixed in* fields and defs. We are using the fact that `mixin` copies annotations to the fields and methods it creates to be able to identify those created from JSOptional fields/defs in traits.

sjrd:
Updated.

library/src/main/scala/scala/scalajs/js/annotation/internal/JSOptional.scala
gzm0: Two things:

- I think we are missing tests (and code) that this annotation cannot be used in source code (like we have for other internal annotations).
- Now that this is internal, I'm wondering if this is the best name (the other extreme of the name spectrum "granularity" would be @JSIgnore). I have been thinking about this for a while, but haven't been able to come up with a fully satisfying name.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSOptionalTest.scala
```diff
@@ -0,0 +1,291 @@
...
+
+    assertEquals(js.undefined, obj.y)
+    assertFalse(js.Object.hasProperty(obj, "y"))
+
+    assertEquals(js.undefined, obj.y2)
+    assertFalse(js.Object.hasProperty(obj, "y2"))
+
+    assertEquals(js.undefined, obj.z)
+    assertFalse(obj.hasOwnProperty("z"))
+    obj.z = Some(3)
```
gzm0: Hummm... In this very case, how do we protect ourselves from GCC renaming something on `ClassImplementsTraitWithOptional`  to `z`?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSOptionalTest.scala
```diff
@@ -0,0 +1,291 @@
...
+    override val x: js.UndefOr[Int] = js.undefined
+    override val y: js.UndefOr[String] = js.undefined
+    override def y2: js.UndefOr[String] = js.undefined
+  }
+
+  @ScalaJSDefined
+  trait OverrideOptionalWithOptionalImplicitType extends TraitWithOptional {
+    override val x = js.undefined
+    override val y = js.undefined
+    override def y2 = js.undefined // scalastyle:ignore
```
gzm0: I do not understand why this example works (in pre 2.12), but all the other overrides that omit result types don't...

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -610,30 +610,53 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+          def isJSOptional(sym: Symbol): Boolean = {
+            sym.owner.isTrait && !sym.isDeferred && !sym.isConstructor &&
+            sym.owner.hasAnnotation(ScalaJSDefinedAnnotation)
+          }
+
+          if (isJSOptional(low) && !(high.isDeferred || isJSOptional(high))) {
+            reporter.error(errorPos,
+                s"Cannot override concrete ${memberDefString(high)} with " +
+                s"optional ${memberDefString(low)} (an optional member is " +
+                "one defined in a non-native JS trait with `= js.undefined`).")
```
sjrd: Yes, that's better indeed.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -865,14 +888,53 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+          if (sym.isMethod && isPrivateMaybeWithin(sym)) {
             reporter.error(tree.pos,
                 "A Scala.js-defined JS trait cannot contain private members")
+          } else if (!sym.isDeferred) {
+            /* Tell the back-end not emit this thing. In fact, this only
+             * matters for mixed-in members created from this member.
+             */
+            sym.addAnnotation(JSOptionalAnnotation)
+
+            def rhsPos = if (tree.rhs != EmptyTree) tree.rhs.pos else tree.pos
```
sjrd: When the method is abstract. But of course that can't happen here because we're in `if (!sym.isDeferred)`. I must have copy-pasted that from somewhere else.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSOptionalTest.scala
```diff
@@ -0,0 +1,160 @@
...
+    }
+    """ hasErrors
+    s"""
+      |newSource1.scala:16: error: Cannot override concrete val a: scala.scalajs.js.UndefOr[Int] with optional override val a: scala.scalajs.js.UndefOr[Int] (an optional member is one defined in a non-native JS trait with `= js.undefined`).
+      |      override val a: js.UndefOr[Int] = js.undefined
+      |                   ^
+      |newSource1.scala:17: error: Cannot override concrete def b: scala.scalajs.js.UndefOr[Int] with optional override def b: scala.scalajs.js.UndefOr[Int] (an optional member is one defined in a non-native JS trait with `= js.undefined`).
+      |      override def b: js.UndefOr[Int] = js.undefined
+      |                   ^
+    """
```
sjrd: That's prohibited altogether. A Scala.js-defined thing can never directly extend a native JS trait.

library/src/main/scala/scala/scalajs/js/annotation/internal/JSOptional.scala
sjrd: > I think we are missing tests (and code) that this annotation cannot be used in source code (like we have for other internal annotations).

Indeed.

> Now that this is internal, I'm wondering if this is the best name (the other extreme of the name spectrum "granularity" would be @JSIgnore). I have been thinking about this for a while, but haven't been able to come up with a fully satisfying name.

Yeah, I wasn't very happy with the name either, but couldn't come with better. It's quite technical :s

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSOptionalTest.scala
sjrd: Oh. That is a very good observation. But in fact the problem is much larger than with optional fields. JS types have JS semantics, so any JS code can add any arbitrary field or method to our JS classes. That means all private fields in JS classes should be protected from being renamed. Ideally they would have `symbol` names, but that wouldn't work in ES5 and would probably have performance issues in ES6. We'll need to think about this, but the bug already exists without this PR so it is a separate issue.

Filed as #2680.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSOptionalTest.scala
sjrd: Because `js.undefined` has type `js.UndefOr[Nothing]`, which is actually a subtype of `js.UndefOr[String]`, so it works because `y2` is redefined with a more specific type. Other examples do not work because `String` is not actually a subtype of `js.UndefOr[String]` (it needs an implicit conversion).

sjrd:
Rebased.

sjrd:
Updated. All comments should have been addressed, except the naming of `@JSOptional`. I still haven't found a better name, but at least it's an internal concept that does not surface to the user.

sjrd:
Jenkins, retest this please

library/src/main/scala/scala/scalajs/js/annotation/internal/JSOptional.scala
```diff
@@ -0,0 +1,22 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2016, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package scala.scalajs.js.annotation.internal
```
gzm0: Why did we not think of this before?!? :-/

Issue #2642, opened 2016-11-05
Cannot find module 'jsdom'


japgolly:
Hi. Using 0.6.13 I decided to try out nodejs & jsdom.

The only Scala.JS test setting I set is:
```
requiresDOM := true
```

I installed jsdom globally, see here:
```
> npm list -g jsdom
/usr/lib
└── jsdom@9.8.3 
```

However when I run tests I get:
```
> sbt 
[info] Loading global plugins from /home/golly/.sbt/0.13/plugins
[info] Loading project definition from /home/golly/projects/scalajs-react/project
[info] Set current project to root (in build file:/home/golly/projects/scalajs-react/)
> test
[info] Fast optimizing /home/golly/projects/scalajs-react/test/target/scala-2.11/test-test-fastopt.js
[error] module.js:474
[error]     throw err;
[error]     ^
[error] 
[error] Error: Cannot find module 'jsdom'
[error]     at Function.Module._resolveFilename (module.js:472:15)
[error]     at Function.Module._load (module.js:420:25)
[error]     at Module.require (module.js:500:17)
[error]     at require (internal/module.js:20:19)
[error]     at [stdin]:35:17
[error]     at [stdin]:93:3
[error]     at ContextifyScript.Script.runInThisContext (vm.js:25:33)
[error]     at Object.exports.runInThisContext (vm.js:77:17)
[error]     at Object.<anonymous> ([stdin]-wrapper:6:22)
[error]     at Module._compile (module.js:573:32)
[trace] Stack trace suppressed: run last test/test:loadedTestFrameworks for the full output.
[error] (test/test:loadedTestFrameworks) org.scalajs.jsenv.ExternalJSEnv$NonZeroExitException: Node.js with JSDOM exited with code 1
[error] Total time: 13 s, completed 05/11/2016 7:13:44 PM
```

Versions are:
```
> node --version
v7.0.0
> npm --version 
3.10.9
```

Do you see something I'm missing?

ochrons:
It doesn't work when installed globally (for some reason), so you need to install it per project. Remember to add `node_modules` to .gitignore :)

Also in Travis the default version of Node is too old so you need to add to your `.travis.yml` this:
```
install:
  - . $HOME/.nvm/nvm.sh
  - nvm install stable
  - nvm use stable
  - npm install
  - npm install jsdom
```

japgolly:
[...]

PR #2665, opened 2016-11-29
#2642 Locally install jsdom on-the-fly

julienrf:
The PR is a work in progress.

I would like to get your feedback on this before going further. It automatically installs jsdom on the fly, locally, if it’s needed by the tests.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
@@ -18,6 +18,7 @@ import org.scalajs.jsenv._
 import org.scalajs.core.ir.Utils.escapeJS
 
 class JSDOMNodeJSEnv(
+  jsdomDirectory: File,
   nodejsPath: String = "node",
   addArgs: Seq[String] = Seq.empty,
   addEnv: Map[String, String] = Map.empty
@@ -88,7 +89,14 @@ class JSDOMNodeJSEnv(
            |})();
            |""".stripMargin
       }
-      Seq(new MemVirtualJSFile("codeWithJSDOMContext.js").withContent(jsDOMCode))
+      val codeFile = new File(jsdomDirectory, "codeWithJSDOMContext.js")
+      val writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(codeFile), "UTF-8"))
+      try {
+        writer.write(jsDOMCode)
+        Seq(FileVirtualJSFile(codeFile))
+      } finally {
+        writer.close()
+      }
     }
 
     override protected def getJSFiles(): Seq[VirtualJSFile] =
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
@@ -24,6 +24,12 @@ class JSDOMNodeJSEnv(
   addEnv: Map[String, String] = Map.empty
 ) extends AbstractNodeJSEnv(nodejsPath, addArgs, addEnv, sourceMap = false) {
 
+  def this(
+    nodejsPath: String = "node",
+    addArgs: Seq[String] = Seq.empty,
+    addEnv: Map[String, String] = Map.empty
+  ) = this(new File(System.getProperty("user.dir")), nodejsPath, addArgs, addEnv)
+
   protected def vmName: String = "Node.js with JSDOM"
 
   override def jsRunner(libs: Seq[ResolvedJSDependency],
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
@@ -14,14 +14,17 @@ import java.io.{Console => _, _}
 import org.scalajs.core.tools.io._
 import org.scalajs.core.tools.jsdep.ResolvedJSDependency
 import org.scalajs.jsenv._
-
 import org.scalajs.core.ir.Utils.escapeJS
 
+import scala.sys.process.{Process, ProcessLogger}
+
 /**
-  * @param jsdomDirectory Directory where jsdom is installed (there must be a node_modules/jsdom/ subdirectory)
+  * @param jsdomDirectory Directory where jsdom will be installed
+  * @param jsdomVersion Version of jsdom to use
   */
 class JSDOMNodeJSEnv(
   jsdomDirectory: File,
+  jsdomVersion: String,
   nodejsPath: String = "node",
   addArgs: Seq[String] = Seq.empty,
   addEnv: Map[String, String] = Map.empty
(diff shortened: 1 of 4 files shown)


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -590,13 +596,15 @@ object ScalaJSPluginInternal {
...
 
       resolvedJSEnv := jsEnv.?.value.getOrElse {
-        if (scalaJSUseRhinoInternal.value) {
-          RhinoJSEnvInternal().value
-        } else if (scalaJSRequestsDOM.value) {
-          JSDOMNodeJSEnv().value
-        } else {
-          NodeJSEnv().value
-        }
+        Def.taskDyn {
```
julienrf: I turned this into a `taskDyn` so that jsdom is only installed if required.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -961,7 +969,24 @@ object ScalaJSPluginInternal {
         new PhantomJettyClassLoader(jettyLoader, getClass.getClassLoader)
       },
       scalaJSJavaSystemProperties := Map.empty,
-      scalaJSConfigurationLibs := Nil
+      scalaJSConfigurationLibs := Nil,
+      scalaJSInstallJsdom := {
+        val jsdomDirectory = target.value / "scalajs-jsdom"
```
julienrf: This is the place where jsdom will be installed.

julienrf:
This would fix #2642. But it requires users to have `npm` installed on their environment.

sjrd:
Jenkins, test this please

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
```diff
@@ -88,7 +89,14 @@ class JSDOMNodeJSEnv(
            |})();
            |""".stripMargin
       }
-      Seq(new MemVirtualJSFile("codeWithJSDOMContext.js").withContent(jsDOMCode))
+      val codeFile = new File(jsdomDirectory, "codeWithJSDOMContext.js")
```
sjrd: This seems like a hack, and wouldn't generalize to having custom modules to be able to reference somehow. 

IMO the natural way of addressing this issue would be to give an extra list of paths where it can look for modules. Would assume there exists a command-line option for that which we can use.

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
```diff
@@ -18,6 +18,7 @@ import org.scalajs.jsenv._
 import org.scalajs.core.ir.Utils.escapeJS
 
 class JSDOMNodeJSEnv(
+  jsdomDirectory: File,
```
sjrd: Consequently, this could be generalized to an arbitrary sequence of directories, and not be tied to `jsdom` installation per se. 

Or without changing the API, simply use `addArgs` to pass command-line arguments.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -961,7 +969,24 @@ object ScalaJSPluginInternal {
...
+        val logger = streams.value.log
+        if (!jsdomDirectory.exists()) {
+          val npm = sys.props("os.name").toLowerCase match {
+            case os if os.contains("win") ⇒ "cmd /c npm"
+            case _ ⇒ "npm"
+          }
+          IO.createDirectory(jsdomDirectory)
+          val process = Process(s"$npm install jsdom", jsdomDirectory)
```
sjrd: We should not install a completely unknown version like that. That undermines the default stability of a Scala.js build. 

We should have a setting for the version of `jsdom` that should be installed instead.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -95,6 +95,12 @@ object ScalaJSPluginInternal {
       "Files used to compute this value (can be used in FileFunctions later).",
       KeyRanks.Invisible)
 
+  val scalaJSInstallJsdom = TaskKey[File](
```
sjrd: To follow our conventions, `JSDOM` should be all-caps (acronyms in camelCase are all caps unless they're in the first position)

julienrf:
Also, the changes in the `JSDOMNodeEnv` constructor break binary compatibility. What should we do about that?

sjrd:
If you use `addArgs`, you won't need that change. Otherwise, a secondary constructor with the old signature must be added for compatibility.

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
julienrf: > IMO the natural way of addressing this issue would be to give an extra list of paths where it can look for modules. Would assume there exists a command-line option for that which we can use.

If I understand correctly, this way is (unfortunately) deprecated: https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders

sjrd:
Jenkins, add to whitelist

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
julienrf: For reference here is an extract of the above linked page (emphasis is *not* mine):

> **You are highly encouraged to place your dependencies locally in `node_modules` folders.**

gzm0:
Have you considered moving this logic inside the `NodeDomSJEnv`? It seems like a better place, since it will keep the sbt plugin independent of the env (and allow non-sbt users to use the install functionality as well).

julienrf:
@gzm0 Good point. I added another commit that does that.

julienrf:
I’m considering delaying this to 1.0 because keeping binary compatibility leads to the addition of noisy code.

sjrd:
> keeping binary compatibility leads to the addition of noisy code

We do that all the time. Grep for "TODO" in the repo and you'll find as much noisy code as you want. This is stuff we have to deal with all the time.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -95,6 +95,18 @@ object ScalaJSPluginInternal {
...
+    "scalaJSJSDOMDirectory",
+    "Directory in which jsdom should be installed.",
+    KeyRanks.Invisible
+  )
+
+  val scalaJSJSDOMVersion = SettingKey[String](
+    "scalaJSJSDOMVersion",
+    "Version of jsdom to install",
+    KeyRanks.Invisible
+  )
```
gzm0: Can't we hide these within the `jsEnv` setting (especially if they are internal keys) and just use the default values inside `JSDOMNodeJSEnv`?

(And maybe add arguments for them, but if you didn't want to expose these in the first place, no need).

julienrf:
@gzm0 Great suggestion again! I added another commit that does that.

sjrd:
Since this has a solution in scalajs-bundler, and we're going to move jsdom support in a different repo anyway, I'm going to close this PR.

erikvanoosten:
What is the currently recommend way to get jsdom?

sjrd:
Hello @erikvanoosten,

For questions, please ask [on StackOverflow](https://stackoverflow.com/questions/tagged/scala.js) or [on Gitter](https://gitter.im/scala-js/scala-js), where more than just the core developers can see and answer.

Thank you for your understanding.

PR #2670, opened 2016-12-03
Various fixes for @JSBracketAccess

gzm0:


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -804,6 +804,24 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+            reporter.error(param.pos,
+                "@JSBracketAccess methods may not have repeated parameters")
+          }
+
+          for {
+            firstParam <- flatParams.headOption
+            if firstParam.isParamWithDefault
+          } {
+            reporter.error(firstParam.pos, "The first parameter of a " +
+                "@JSBracketAccess method may not be a default parameter")
```
sjrd: What would it mean for the second parameter to have a default value? If we allow that, we should have a positive test for it. But I believe it should be forbidden.

gzm0: It will toggle between assignment and select. It is our own little support for jQuery style APIs :P

I'm fine with forbidding it (can always use an overload, which is probably better anyways). This is just one piece that worked before.

gzm0:
LOL:
```
[error] /localhome/jenkins/a/workspace/scalajs-task-worker_2/library/src/main/scala/scala/scalajs/js/Dynamic.scala:39: error: @JSBracketAccess methods with two parameters must return Unit
[error]   def updateDynamic(name: String)(value: Any): Unit = native
[error]       ^
[error] one error found
```

sjrd:
> LOL:

You need to use `finalResultType` instead of `resultType`, IIRC.

gzm0:
Was just looking for this thing. Thanks.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
gzm0: > Seriously, that works?

Yes... It's all post JS argument extraction magic. So even this works:

```scala
@JSBracketAccess
def foo(x: Int*): Unit = js.native

x.foo()          // crash
x.foo(1)         // x[1]
x.foo(1, 2)      // x[1] = 2
x.foo(1, 2, 3)   // crash
x.foo(mySeq: _*) // crash
```

So if we make the default argument thingy just a warning, we also need to make this a warning and proper errors at the call sites (so we don't crash).

Do you think we could consider breaking backwards compatibility just a bit here?

gzm0:
Updated.

gzm0:
Updated.

PR #2672, opened 2016-12-04
Optimize fromFunction1 for AnonFunction1

Atry:


sjrd:
What motivates this change? What is your use case, and what change does it bring to your generated code that makes this worth it?

Atry:
I saw that [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala) creates some `js.Function` wrapper for `AnonFunction1` in `<input onclick={_: Event => ???}/>`
``` javascript
    element$macro$59$1.onclick = (function(f) {
      return (function(arg1) {
        return f.apply__O__O(arg1)
      })
    })(newValue$macro$61$1)
```

Atry:
I hope the wrapper can be eliminated by this PR

Atry:
There were similar approach in the implicit converter from `js.WrappedArray` to `js.Array`

sjrd:
> I hope the wrapper can be eliminated by this PR

"hope"? Didn't you actually try? Does this PR have any impact on the observed performance?

sjrd:
Actually, this change can't possibly make any difference, because the body of `js.Any.fromFunctionN` is not even taken into account. The compiler rewrites calls to `fromFunctionN` as a primitive. Closing this.

PR #2674, opened 2016-12-04
Fix #2587: Corner case in BigDecimal.multiply().

### LINKED ISSUE #2587: BigDecimal is not distributive

sjrd:
When multiplying two negative numbers for which the product should be `-2^63` (which is `-Long.MinValue`), the sum of the bit lengths is 63, but the result is still not representable. This is now detected as a special case.

This PR based on #2588 with some tweaks to avoid 3 relatively costly calls to `JLong.signum`. Also, in the corner case, we precisely know the one possible result, so we return it directly instead of going through the big multiplication.

Supersedes #2588.

### DIFF
--- a/javalib/src/main/scala/java/math/BigDecimal.scala
+++ b/javalib/src/main/scala/java/math/BigDecimal.scala
@@ -675,7 +675,14 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
     if (this.isZero || multiplicand.isZero) {
       zeroScaledBy(newScale)
     } else if (this._bitLength + multiplicand._bitLength < 64) {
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0 
/cc @non

PR #2683, opened 2016-12-12
Do not rely on `import` being relative to the `scala` package.

sjrd:
Because Eclipse sometimes does not like it, and it's ugly anyway.


sjrd:

PR #2684, opened 2016-12-13
fix a couple out-of-date comments

SethTisue:


sjrd:
Jenkins, test this please (only because I like green marks)

Issue #1902, opened 2015-09-14
Support for static fields in @ScalaJSDefined classes


chandu0101:
Tracking issues for  https://groups.google.com/forum/#!topic/scala-js/v3gue_-Tms4

xsistens:
:+1:

chandu0101:
@sjrd 

I working on mutation's in Relay (https://facebook.github.io/relay/docs/api-reference-relay-mutation.html#fragments-static-property), 

Relay.Mutation class has static field `fragments` in our mutation class we must define fragments. 

```scala
@ScalaJSDefined
class AddTodoMutation(props : js.Dynamic) extends RelayMutation(props) {
... 
}

val addTodoCtor = js.constructorOf[AddTodoMutation]
val addTodoFragFunction: js.Function = ..
addTodoCtor.fragments = js.Dictionary("viewer" -> addTodoFragFunction)
 
``` 

this indeed solved one of two problems but for other problem  react-relay looking for `fragments` field in base class(RelayMutation) constructor https://github.com/facebook/relay/blob/fbf8c50825ff609a5ab101a934f0b9e165ab5d9b/src/mutation/RelayMutation.js#L230

can we work around this problem .. ?

sjrd:
I don't understand what *is* the problem? Can you elaborate?

chandu0101:
sure , 
https://github.com/facebook/relay/blob/fbf8c50825ff609a5ab101a934f0b9e165ab5d9b/src/mutation/RelayMutation.js#L230

```js

 var fragments = RelayDeprecated.getMutationFragments(this.constructor);

  getMutationFragments: function getMutationFragments(Mutation) {
	    var queries = Mutation.queries;
	    if (queries && !Mutation.fragments) {
	      RelayDeprecated.warn({
	        was: Mutation.name + '.queries',
	        now: Mutation.name + '.fragments'
	      });
	      Mutation.fragments = queries;
	    }
	    return Mutation.fragments;  // its looking for static fragments field
	  }
```
in runtime fragments is undefined for `AddTodoMutation`  , what i really want is my fragments should be available in constructor. 

 `addTodoCtor.fragments = js.Dictionary("viewer" -> addTodoFragFunction)`   this line is not useful in above scenario.

sjrd:
Isn't `this.constructor` exactly `eq` to `js.constructorOf[AddTodoMutation]` at that point?

chandu0101:
no

sjrd:
[...]

PR #2685, opened 2016-12-13
DO NOT MERGE Create Scala.js-defined JS class values lazily.

sjrd:
This will be necessary to implement features which require user-defined code to be executed at class-definition-time. Since such code can have arbitrary cyclic dependencies on anything else, it is absolutely necessary that everything else is defined or lazily loaded.

The two features that will need that are:

* `@JSSymbol` (#1997)
* Static fields, with their static initializers (#1902)

### DIFF
--- a/tools/scalajsenv.js
+++ b/tools/scalajsenv.js
@@ -125,6 +125,8 @@ ScalaJS.clz32 = ScalaJS.g["Math"]["clz32"] || (function(i) {
 // Other fields
 //!if outputMode == ECMAScript51Global
 ScalaJS.d = {};         // Data for types
+ScalaJS.a = {};         // Scala.js-defined JS class value accessors
+ScalaJS.b = {};         // Scala.js-defined JS class value fields
 ScalaJS.c = {};         // Scala.js constructors
 ScalaJS.h = {};         // Inheritable constructors (without initialization code)
 ScalaJS.s = {};         // Static methods
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
@@ -2217,8 +2217,8 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
       implicit outputMode: OutputMode, pos: Position): js.Tree = {
     spec match {
       case None =>
-        // this is a Scala.js-defined JS class
-        encodeClassVar(className)
+        // This is a Scala.js-defined JS class, call its class value accessor
+        js.Apply(envField("a", className), Nil)
 
       case Some(spec) =>
         genLoadJSFromSpec(spec)
@@ -2343,6 +2343,14 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
   private[emitter] def envFieldDef(field: String, subField: String,
       origName: Option[String], value: js.Tree, mutable: Boolean)(
       implicit outputMode: OutputMode, pos: Position): js.Tree = {
+    envFieldDef(field, subField, origName, value, mutable,
+        keepFunctionExpression = false)
+  }
+
+  private[emitter] def envFieldDef(field: String, subField: String,
+      origName: Option[String], value: js.Tree, mutable: Boolean,
+      keepFunctionExpression: Boolean)(
+      implicit outputMode: OutputMode, pos: Position): js.Tree = {
     val globalVar = envField(field, subField, origName)
     def globalVarIdent = globalVar.asInstanceOf[js.VarRef].ident
 
@@ -2354,7 +2362,11 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
         value match {
           case js.Function(args, body) =>
             // Make sure the function has a meaningful `name` property
-            js.FunctionDef(globalVarIdent, args, body)
+            val functionExpr = js.FunctionDef(globalVarIdent, args, body)
+            if (keepFunctionExpression)
+              js.VarDef(globalVarIdent, Some(functionExpr))
+            else
+              functionExpr
           case _ =>
             js.VarDef(globalVarIdent, Some(value))
         }
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
@@ -38,13 +38,11 @@ private[emitter] final class ScalaJSClassEmitter(
 
   private implicit def implicitOutputMode: OutputMode = outputMode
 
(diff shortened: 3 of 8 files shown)


sjrd:
Review by @gzm0, but I suggest we do not merge this as part of this PR. Rather, we should build either `@JSSymbol` or Static fields on top of this commit to gain confidence that it addresses everything we need.

sjrd:
Well, this causes a `SyntaxError` in PhantomJS 1.9.7 and 1.9.8, but it won't tell me where. Grrr. And it works with PhantomJS 2.1.1 (the latest).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2204,21 +2212,37 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
...
+        if (enclosingClassName.exists(_ == className)) {
+          encodeClassVar(className)
+        } else {
+          /* In ECMAScript51Global mode, the accessor in under `ScalaJS.a`
+           * rather than `ScalaJS.c` to avoid clashes.
+           */
+          val accessor = outputMode match {
+            case OutputMode.ECMAScript51Global => envField("a", className)
+            case _                             => encodeClassVar(className)
+          }
```
gzm0: Is there a reason we need to do this (and not just always use `a`)?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2354,7 +2386,11 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
         value match {
           case js.Function(args, body) =>
             // Make sure the function has a meaningful `name` property
-            js.FunctionDef(globalVarIdent, args, body)
+            val functionExpr = js.FunctionDef(globalVarIdent, args, body)
+            if (keepFunctionExpression)
+              js.VarDef(globalVarIdent, Some(functionExpr))
```
gzm0: I do not understand why this is necessary. I understand that this prevents hoisting of the function (probably, although the function keeps its name....). What does that change? It's not like the function body is evaluated...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -113,6 +117,39 @@ private[emitter] final class ScalaJSClassEmitter(
...
+      val createClassValueVar =
+        envFieldDef("b", className, js.Undefined(), mutable = true)
+
+      val createAccessor = {
+        val classValueVar = envField("b", className)
+
+        val assignClassValueVar = js.Block(
+            entireClassDef,
+            classValueVar := envField("c", className)
+        )
```
gzm0: Could you move this inside the `js.If` or make it somehow clear that this is executed conditionally? It's kind of hard to read top-to-bottom otherwise.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -113,6 +117,39 @@ private[emitter] final class ScalaJSClassEmitter(
...
+        val initBlock =
+          js.If(!classValueVar, assignClassValueVar, js.Skip())
+
+        val body = js.Block(initBlock, js.Return(classValueVar))
+
+        val envFieldName = outputMode match {
+          case OutputMode.ECMAScript51Global => "a"
+          case _                             => "c"
+        }
+        envFieldDef(envFieldName, className, js.Function(Nil, body))
```
gzm0: Same here: Is the "override `c` stuff" really necessary?

tools/scalajsenv.js
```diff
@@ -125,6 +125,8 @@ ScalaJS.clz32 = ScalaJS.g["Math"]["clz32"] || (function(i) {
 // Other fields
 //!if outputMode == ECMAScript51Global
 ScalaJS.d = {};         // Data for types
+ScalaJS.a = {};         // Scala.js-defined JS class value accessors
+ScalaJS.b = {};         // Scala.js-defined JS class value fields
 ScalaJS.c = {};         // Scala.js constructors
```
gzm0: Just do double check: There are three of these, such that during initialization of a SJS defined class, the class itself is not yet available externally.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -203,7 +203,7 @@ private[emitter] class JSDesugaring(internalOptions: InternalOptions) {
   /** Desugars parameters and body to a JS function.
    */
   private[emitter] def desugarToFunction(
-      classEmitter: ScalaJSClassEmitter, enclosingClassName: String,
+      classEmitter: ScalaJSClassEmitter, enclosingClassName: Option[String],
```
gzm0: Just to double check. `enclosingClassName` is becoming optional because:

- It makes more sense.
- To prevent static methods of SJS defined JS classes from being treated specially w.r.t. their classes.

If it is not too much of a hassle, consider moving this change into a separate commit.

sjrd: That's right. I'll try to extract that in a separate commit. It shouldn't be too hard.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Not really. It's probably better to consistently use `a` indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: It caused a `ParseError` in PhantomJS if I did that differently. I'm not sure there's something against that in the ES spec (5.1, 5.1 strict and/or 2015), but I seem to remember I saw something about that someday. The current implementation is safe and assured to always work.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
sjrd: Doing so actually significantly improves the situation, as `enclosingClassName` is not necessary anymore, so I can revert all the changes I made about `enclosingClassName`.

sjrd:
Updated.

sjrd:
Rebased. Solved conflict in `project/BinaryIncompatibilities.scala`.

sjrd:
This has been integrated as part of #2705.

PR #2693, opened 2016-12-18
Add a helper `js.defined(x)` to drive type inference in some cases.

sjrd:
When the expected type of an expression `x` is of the form `js.UndefOr[A]`, but an implicit conversion or a SAM conversion is required to type `x` as an `A`, the normal implicit conversion to `js.UndefOr[A]` does not work.

We can use the helper `js.defined(x)` to drive Scala's type inference to apply the implicit conversion or the SAM conversion. This is particularly relevant for SAM conversion, for which a short-cut implicit conversion with an evidence would not be enough to allow parameter type inference anyway.

This is particularly important for configuration objects with function fields.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/defined.scala
+++ b/library/src/main/scala/scala/scalajs/js/defined.scala
@@ -1,0 +1,31 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2016, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+
+object defined { // scalastyle:ignore
+  /** Explicitly upcasts an `A` to a `js.UndefOr[A]`.
+   *
+   *  This method is useful in some cases to drive Scala's type inference.
+   *  For example, when calling a method expecting a `js.UndefOr[js.FunctionN]`
+   *  as shown below:
+   *
+   *  {{{
+   *  def foo(f: js.UndefOr[js.Function1[Int, Int]] = js.undefined): Int = ???
+   *
+   *  foo((x: Int) => x + 1) // compile error (requires 2 chained implicits)
+   *  foo(js.defined((x: Int) => x + 1)) // compiles
+   *
+   *  // 2.12+ only:
+   *  foo(js.defined(x => x + 1)) // compiles as well
+   *  }}}
+   */
+  def apply[A](a: A): js.UndefOr[A] = a
+}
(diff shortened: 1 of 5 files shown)


sjrd:
Review by @gzm0 

I realized the necessity for this helper when prepare release notes for 0.6.14. It was incredibly awkward to write configuration objects with function fields without this helper.

library/src/main/scala/scala/scalajs/js/defined.scala
```diff
@@ -0,0 +1,31 @@
...
+   *  def foo(f: js.UndefOr[js.Function1[Int, Int]] = js.undefined): Int = ???
+   *
+   *  foo((x: Int) => x + 1) // compile error (requires 2 chained implicits)
+   *  foo(js.defined((x: Int) => x + 1)) // compiles
+   *
+   *  // 2.12+ only:
+   *  foo(js.defined(x => x + 1)) // compiles as well
+   *  }}}
+   */
+  def apply[A](a: A): js.UndefOr[A] = a
```
gzm0: What is the reason you are not defining this on the `js` package object? You should probably document that reason :)

gzm0:
> js.defined is an object with an apply method (rather than a def in the package object) to allow for a possible evolution with an unapply method to allow pattern matching on js.UndefOr[A].

So much for my ability to read...

Are you sure we would want to have a lower case pattern? Main concerns:

- There is no precedent in the standard library for it (and the pattern matcher is biased against it).
- We cannot make it work for `js.undefined` without some hack for it to become stable (but still inlineable). Also, it will not do what someone expects if they `import js._`.

OTOH, the following does not look bad at all:

```scala
myValue match {
  case js.defined(x) => println("yes " + x)
  case js.undefined  => println("no")
}
```

Then again, who pattern matches on option-like types... :P

TBH, I'm quite split about this one, notably because having this as a separate object will significantly hurt discoverability in the API doc as well...

sjrd:
> By this, you mean that adding a view bound to UndefOr.any2undefOrA would not work for SAMs?

Yes, exactly. I should improve the wording.

sjrd:
> OTOH, the following does not look bad at all:
>
>     myValue match {
>       case js.defined(x) => println("yes " + x)
>       case js.undefined  => println("no")
>     }

That's indeed what I had in mind.

> Then again, who pattern matches on option-like types... :P

Everyone but us, I'm afraid.

> We cannot make it work for js.undefined without some hack for it to become stable (but still inlineable).

Indeed. I have a wip in a branch: https://github.com/scala-js/scala-js/compare/master...sjrd:undefor-patmat

> Also, it will not do what someone expects if they `import js._`

This is a bad idea anyway. It imports a different `Any`, `Array`, `Object` in scope already. `import js._` should not be done.

Note that it would still actually do what one expects for `case defined(x)`, because with `()`, `defined` is not a capture, it's really a call to `defined.unapply`. It would do the unexpected thing for `case undefined` though.

> TBH, I'm quite split about this one, notably because having this as a separate object will significantly hurt discoverability in the API doc as well...

All term members (vals, defs, objects and even packages) are shown together in APIs (see e.g., [around here](https://www.scala-js.org/api/scalajs-library/latest/index.html#scala.scalajs.js.package@typedarray)). Discoverability in the API docs should be exactly the same for an `object` versus a `def`.

gzm0:
Ok, fair enough. Let's go ahead with this one then.

sjrd:
Updated with `// scalastyle:ignore` and more explicit commit message.

sjrd:
@gzm0 Anything else here? This is the last PR for 0.6.14.

PR #2697, opened 2016-12-22
Fix #2694: Blacklist partest pos/t9181.scala.

### LINKED ISSUE #2694: pos/t9181.scala tends to spuriously fail ~1/10 of the time
Source: https://github.com/scala/scala/blob/a9802c317a72925838bcd5c53a5919f75b577caa/test/files/pos/t9181.scala
It's clearly a test designed to stress-test some part of the scalac compiler. I think it's stressing it a bit too much given our CI specs. We should consider blacklisting it.

sjrd:
It's an artificially giant `if/else if/.../else` chain, which causes nesting of `If` nodes in our IR. This kills our IR serializer with a `StackOverflowException`.

Blacklisting as won't fix, until and unless a Scala.js user hits that limitation in real life.


sjrd:
Review by @gzm0.

PR #2698, opened 2016-12-22
Fix #2519: Test the behavior of setting readonly properties.

### LINKED ISSUE #2519: Test the behavior of assignments to exported properties without setter
https://github.com/scala-js/scala-js/pull/2514/files/c3f8ac01539ffa4534ecddd976722f4d0b689bac#r70224249

sjrd:
* Exports of `val`s and `def`s without parentheses
* `def`s without parentheses in Scala.js-defined JS classes

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
@@ -11,9 +11,12 @@ import scala.scalajs.js
 import scala.scalajs.js.annotation._
 
 import org.junit.Assert._
+import org.junit.Assume._
 import org.junit.Test
 
 import org.scalajs.testsuite.utils.JSAssert._
+import org.scalajs.testsuite.utils.Platform
+import org.scalajs.testsuite.utils.AssertThrows.assertThrows
 
 class ScalaJSDefinedTest {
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -565,6 +568,32 @@ class ScalaJSDefinedTest {
...
 
+  @Test def readonly_properties(): Unit = {
+    assumeFalse(
+        "Assuming strict mode semantics, which are not honored by Rhino",
+        Platform.executingInRhino)
+
+    // Named classes
+    @ScalaJSDefined
+    class Foo extends js.Object {
+      def bar: Int = 1
```
gzm0: Just to check my understanding: You are not testing for vals, because we cannot make them immutable.

sjrd: We *could* make them immutable (using the `writable = false` flag to `defineProperty`), but we don't. Our "spec" doesn't mention making vals actually immutable on the js side. If we did, it could have bad performance impacts. We can reconsider this, but that would be a spec change.

gzm0: OK. Wenn probably should consider making this UB at least. Otherwise it seems wrong to allow `val`s at all...

Issue #2702, opened 2016-12-28
Proposal: @JSExportStatic: static fields and methods for JS classes


sjrd:
This is a proposal to address #1902. Besides, it also encompasses the definition of static methods on Scala.js-defined JS classes, as well as exporting top-level vals and vars. In fact, it generalizes `@JSExportTopLevel` in two directions: a) as `@JSExportStatic` to export as static member rather than top-level member and b) to apply to val/vars as well as methods.

## `@JSExportStatic` for methods

The annotation `@JSExportStatic` can be used on a `def` of a Scala top-level `object` (similarly to `@JSExportTopLevel` as in 0.6.14) but only if the companion class of that object is a Scala.js-defined JS class. For example:
```scala
@ScalaJSDefined
class Foo extends js.Object

object Foo { // a Scala object, not a JS object
  @JSExportStatic
  def bar(x: Int): Int = x + 1
}
```
The semantics of such an export is that a static method `"bar"` will be defined on the JS class `Foo`, which is a typical export-like forwarder to `Foo.bar`. Overloading is handled as usual for exports. With those definitions, the following snippet would work:
```scala
val fooClass: js.Dynamic = js.constructorOf[Foo]
val y = fooClass.bar(5)
println(y) // 6

fooClass.bar = (x: Int) => x * 2 // patch it, because we can!
println(fooClass.bar(5)) // 10
println(Foo.bar(5)) // still 6! We only patched the forwarder, not the Scala def
```
So far, just like `@JSExportTopLevel`, only defs with parentheses can be exported as static.

## Exporting getters and setters as static

Unlike `@JSExportTopLevel`, we can easily generalize `@JSExportStatic` to support getters (`def` without parentheses) and setters (`def` with a name ending with `_=`), as we can define static properties on the companion class.

This is an extension that does not necessarily need to be supported in a first implementation of the proposal.

## Exporting static fields (vals and vars)

`@JSExportStatic` can also be used on `val`s and `var`s in top-level Scala objects whose companion class is a Scala.js-defined JS class. In that case, an actual *field* is created on the JS class (*not* a getter/setter as is the case for `@JSExport`). The synthetic Scala getters and setters for that field are rewritten to access the static field (and box/unbox it as needed).

This can be used as follows:
```scala
@ScalaJSDefined
class Foo extends js.Object

object Foo { // a Scala object, not a JS object
  @JSExportStatic
  val alpha: Int = 1
  @JSExportStatic
  var beta: String = "hello"
}

[...]

PR #2705, opened 2016-12-30
Implementation of `@JSExportStatic` from Proposal #2702

sjrd:
CI only.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -125,15 +125,59 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         yield js.TopLevelExportDef(m)(m.pos)
     }
 
-    def genStaticExports(classSym: Symbol): List[js.MethodDef] =
+    def genStaticExports(classSym: Symbol): List[js.Tree] =
       genTopLevelOrStaticExports(classSym, ExportDestination.Static)
 
     private def genTopLevelOrStaticExports(classSym: Symbol,
-        destination: ExportDestination): List[js.MethodDef] = {
+        destination: ExportDestination): List[js.Tree] = {
       require(
           destination == ExportDestination.TopLevel ||
           destination == ExportDestination.Static)
 
+      val exportsNamesAndPositions = {
+        genTopLevelOrStaticFieldExports(classSym, destination) ++
+        genTopLevelOrStaticMethodExports(classSym, destination)
+      }
+
+      for {
+        exportsWithSameName <- exportsNamesAndPositions.groupBy(_._2).values
+        duplicate <- exportsWithSameName.tail
+      } {
+        val strKind =
+          if (destination == ExportDestination.TopLevel) "top-level"
+          else "static"
+        reporter.error(duplicate._3,
+            s"Duplicate $strKind export with name '${duplicate._2}': " +
+            "a field may not share its exported name with another field or " +
+            "method")
+      }
+
+      exportsNamesAndPositions.map(_._1)
+    }
+
+    private def genTopLevelOrStaticFieldExports(classSym: Symbol,
+        destination: ExportDestination): List[(js.FieldDef, String, Position)] = {
+      (for {
+        fieldSym <- classSym.info.members
+        if !fieldSym.isMethod && fieldSym.isTerm && !fieldSym.isModule
+        export <- jsInterop.registeredExportsOf(fieldSym)
+        if export.destination == destination
+      } yield {
+        // in fact, top-level field exports do not exist yet
+        assert(export.destination == ExportDestination.Static)
+
+        implicit val pos = fieldSym.pos
+
+        val mutable = true // static fields must always be mutable
+        val name = js.StringLiteral(export.jsName)
+        val irTpe = genExposedFieldIRType(fieldSym)
+
+        (js.FieldDef(static = true, name, irTpe, mutable), export.jsName, pos)
+      }).toList
+    }
+
+    private def genTopLevelOrStaticMethodExports(classSym: Symbol,
+        destination: ExportDestination): List[(js.MethodDef, String, Position)] = {
       val allRelevantExports = for {
         methodSym <- classSym.info.members
         if methodSym.isMethod && !methodSym.isConstructor
@@ -149,7 +193,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         implicit val pos = tups.head._1.pos
 
         val alts = tups.map(t => ExportedSymbol(t._2)).toList
-        genExportMethod(alts, jsName, static = true)
+        (genExportMethod(alts, jsName, static = true), jsName, pos)
       }
     }
 
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -121,26 +121,36 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     }
 
     def genTopLevelExports(classSym: Symbol): List[js.TopLevelExportDef] = {
-      val allTopLevelExports = for {
+      for (m <- genTopLevelOrStaticExports(classSym, ExportDestination.TopLevel))
+        yield js.TopLevelExportDef(m)(m.pos)
+    }
+
+    def genStaticExports(classSym: Symbol): List[js.MethodDef] =
+      genTopLevelOrStaticExports(classSym, ExportDestination.Static)
+
+    private def genTopLevelOrStaticExports(classSym: Symbol,
+        destination: ExportDestination): List[js.MethodDef] = {
+      require(
+          destination == ExportDestination.TopLevel ||
+          destination == ExportDestination.Static)
+
+      val allRelevantExports = for {
         methodSym <- classSym.info.members
         if methodSym.isMethod && !methodSym.isConstructor
         export <- jsInterop.registeredExportsOf(methodSym)
+        if export.destination == destination
       } yield {
-        assert(export.isTopLevel)
         (export, methodSym)
       }
 
-      val result = for {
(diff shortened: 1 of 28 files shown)


sjrd:
`@JSExportTopLevel` for fields will require more work than I thought, and less similar to `@JSExportStatic` than I thought, so I think it is best to keep that one out from this PR, and submit it later.

Therefore this PR is ready for review. Review by @gzm0. The second commit has been cherry-picked verbatim from #2685 and need not be reviewed again.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSExportStaticTest.scala
```diff
@@ -0,0 +1,185 @@
...
+
+import org.scalajs.testsuite.utils.AssertThrows._
+import org.scalajs.testsuite.utils.Platform._
+
+import org.junit.Assert._
+import org.junit.Assume._
+import org.junit.Test
+
+class JSExportStaticTest {
+  import JSExportStaticTest._
```
gzm0: Consider not importing this here, for better clarity in the nested tests (not entirely sure it's more readable though).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -301,7 +301,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
       }
     }
 
-    if (classDef.kind.isJSClass && name == "constructor") {
+    if (classDef.kind.isJSClass && name == "constructor" && !static) {
```
gzm0: This makes me think: What happens if I export a static under the name "constructor"?

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -128,6 +129,18 @@ trait JSGlobalAddons extends JSDefinitions
     def isJSSetter(sym: Symbol): Boolean =
       nme.isSetterName(sym.name) && sym.isMethod && !sym.isConstructor
 
+    /** Is this symbol a static field at the IR level? */
+    def isStaticField(sym: Symbol): Boolean = {
```
gzm0: Can we call this `isFieldStatic` to make it extra clear, that this does not verify whether `sym` is even a field?

library/src/main/scala/scala/scalajs/js/annotation/JSExportStatic.scala
```diff
@@ -23,7 +23,6 @@ import scala.annotation.meta._
  *
  *  @see [[https://www.scala-js.org/doc/interoperability/sjs-defined-js-classes.html Write JavaScript classes in Scala.js]]
  */
-@field @getter @setter
```
gzm0: Intentional? IIRC, this is only relevant for fields that are defined in the ctor of a class, so not really for statics. But in that case, shouldn't have been there in the first place...

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSExportStaticTest.scala
```diff
@@ -182,4 +256,32 @@ object JSExportStaticTest {
     @JSExportStatic
     def alsoExistsAsMember(x: Int): Int = x * 5
   }
+
+  @ScalaJSDefined
+  class StaticExportFields extends js.Object {
+    val alsoExistsAsMember: Int = 5
```
gzm0: Did you want to introduce a static field with this name as well? (and maybe a test?)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -303,6 +305,34 @@ private[emitter] final class ScalaJSClassEmitter(
...
+      implicit val pos = field.pos
+      val classVar = envField("c", className)
+      val select = genPropSelect(classVar, genPropertyName(name))
+      js.Assign(select, genZeroOf(ftpe))
+    }
+    js.Block(stats)(tree.pos)
+  }
+
+  /** Generates the static initializer invocation of a JavaScript class. */
+  private def genStaticInitializerOfJSClass(tree: LinkedClass)(
```
gzm0: Call this `genStaticInitializationOfJSClass` or `genStaticInitializerCallForJSClass`? It doesn't generate the static initializer itself.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -26,6 +26,7 @@ trait PrepJSExports { this: PrepJSInterop =>
       pos: Position,
       isNamed: Boolean,
       isTopLevel: Boolean,
+      isStatic: Boolean,
```
sjrd: Done. There is an ADT/enum `ExportDestination` with 3 values: `Normal`, `TopLevel` and `Static`.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -1251,4 +1251,334 @@ class JSExportTest extends DirectTest with TestHelpers {
...
+    object StaticContainer {
+      @JSExportStatic
+      val a: Int = 1
+
+      @JSExportStatic("a")
+      var b: Int = 1
+    }
+    """ hasErrors
+    """
+      |newSource1.scala:8: error: Duplicate static export with name 'a'
```
sjrd: Indeed. I have added that precision.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -121,26 +121,72 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+        genTopLevelOrStaticFieldExports(classSym, topLevels) ++
+        genTopLevelOrStaticMethodExports(classSym, topLevels)
+      }
+
+      for {
+        (head :: tail) <- exportsNamesAndPositions.tails
+        duplicate <- tail if duplicate._2 == head._2
+      } {
+        reporter.error(duplicate._3,
+            s"Duplicate ${if (topLevels) "top-level" else "static"} export " +
```
sjrd: Indeed.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -121,26 +121,72 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+      } yield {
+        assert(export.isStatic) // in fact, top-level field exports do not exist yet
+
+        implicit val pos = fieldSym.pos
+
+        val mutable = true // static fields must always be mutable
+        val name = js.StringLiteral(export.jsName)
+        val irTpe = genExposedFieldIRType(fieldSym)
+
+        (js.FieldDef(static = true, name, irTpe, mutable), export.jsName, pos)
```
sjrd: I'm not sure about this one. This place doesn't seem to be the right place to introduce this comment (after all, for instance fields we also do not mention that). But also I did not find an appropriate place.

ir/src/main/scala/org/scalajs/core/ir/Definitions.scala
```diff
@@ -66,6 +66,9 @@ object Definitions {
   val AncestorsOfPseudoArrayClass = Set(
       ObjectClass, SerializableClass, CloneableClass)
 
+  /** Name of the static initializer method. */
+  final val StaticInitializerName = "clinit___"
```
sjrd: Yes. Just like the instance constructor is named `<init>` on the JVM, the static initializer is named `<clinit>` (for class init)

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSExportStaticTest.scala
sjrd: A static field and a test, indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: It works out just fine, there's nothing special about it. I have added a test.

library/src/main/scala/scala/scalajs/js/annotation/JSExportStatic.scala
sjrd: You're right, it shouldn't have been there in the first place.

It's important *not* to have these meta-annotations so that an annotation on a `val` or `var` is applied *only* on the field, and not on its getter and setter.

sjrd:
Updated. I should have addressed all your comments, except mentioning that static fields are initialized in the companion module's constructor.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
gzm0: That's a fair point. I always forget that this is normal :P

compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
```diff
@@ -142,6 +143,18 @@ trait JSGlobalAddons extends JSDefinitions
...
       nme.isSetterName(sym.name) && sym.isMethod && !sym.isConstructor
 
+    /** Is this field symbol a static field at the IR level? */
+    def isFieldStatic(sym: Symbol): Boolean = {
+      sym.owner.isModuleClass && // usually false, avoids a lookup in the map
+      registeredExportsOf(sym).nonEmpty
+    }
+
+    /** The only export info of a static field.
+     *  Requires `isStaticField(sym)`.
```
gzm0: isFieldStatic

sjrd: Fixed.

PR #2706, opened 2017-01-01
Fix #2703: Hide private and boilerplate docs

### LINKED ISSUE #2703: Hide stuff in library doc
Internal namespaces:

- `scala.scalajs.runtime`
- `scala.scalajs.js.annotation.internal`

"Obvious" extensions of arity. Just like in the standard libarary:
- `js.Function3+`
- `js.ThisFunction3+`
- `js.Tuple3+`

gzm0:


gzm0:
In action:

screen shot 2017-01-01 at 4 16 12 pm screen shot 2017-01-01 at 4 20 46 pm

gzm0:
Review by @sjrd

project/Build.scala
```diff
@@ -1005,6 +1004,34 @@ object Build {
...
+            )
+
+            (sources in doc).value.filter(filter.accept)
+          },
+
+          /* Add compiled .class files to doc dependencyClasspath, so we can
+           * still compile even with only part of the files being present.
+           */
+          dependencyClasspath in doc +=
+            Def.task(classDirectory.value).dependsOn(compile).value,
```
sjrd: You should use `++= Attributed.data((exportedProducts in Compile).value)` instead. `exportedProducts` is the normal task to use to get the list of `classes` directories after compilation has been done.

project/Build.scala
```diff
@@ -1005,6 +1004,34 @@ object Build {
...
+                val path = f.getAbsolutePath.replace('\\', '/')
+                pat.matcher(path).matches()
+              }
+            }
+
+            val filter: FileFilter = (
+                AllPassFilter
+                  -- regexFileFilter(".*/scala/scalajs/runtime/.*")
+                  -- regexFileFilter(".*/scala/scalajs/js/annotation/internal/.*")
+                  -- "*.nodoc.scala"
```
sjrd: Given the two uses of `regexFileFilter`, a regex is a bit overkill. You could have a `containsFileFilter("/scala/scalajs/runtime/")` that simply performs `path.contains(s)`.

library/src/main/scala/scala/scalajs/js/ThisFunction.nodoc.scala
```diff
@@ -0,0 +1,110 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+/** Definitions for js.ThisFunction3 to js.ThisFunction22 that do not show in doc */
```
sjrd: This should not be a Scaladoc comment, but a normal comment.

library/src/main/scala/scala/scalajs/js/Tuple.nodoc.scala
```diff
@@ -0,0 +1,702 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+/** Definitions for js.Tuple4 to js.Tuple22 that do not show in doc */
```
sjrd: This should not be a Scaladoc comment, but a normal comment.

library/src/main/scala/scala/scalajs/js/Function.nodoc.scala
```diff
@@ -0,0 +1,114 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+/** Definitions for js.Function3 to js.Function22 that do not show in doc */
```
sjrd: This should not be a Scaladoc comment, but a normal comment.

project/Build.scala
gzm0: Note that this pulls in all the other library projects as well (scalalib, etc.). This is not an issue, but makes building the doc a bit slower than necessary.

gzm0:
Updated.

Issue #2627, opened 2016-10-12
IR Checker does not check for member name collisions


gzm0:
We currently can't enable this, due to #2382.

PR #2716, opened 2017-01-09
Export fields at the top-level with `@JSExportTopLevel`.

sjrd:
This commit implements the third part of the proposal #2702, i.e., exporting fields at the top-level.

The `@JSExportTopLevel` annotation can now be used on `val`s and `var`s in addition to methods. This causes the specified field to be stored as a static field in the IR (which becomes a top-level var in JS) and exported at the top-level. Field exports are read-only views, as is the case in ECMAScript 2015 modules, so JS code cannot modify exported vars, but can observe changes to it done by Scala code.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Infos.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Infos.scala
@@ -44,6 +44,8 @@ object Infos {
       val isStatic: Boolean,
       val isAbstract: Boolean,
       val isExported: Boolean,
+      val staticFieldsRead: Map[String, List[String]],
+      val staticFieldsWritten: Map[String, List[String]],
       val methodsCalled: Map[String, List[String]],
       val methodsCalledStatically: Map[String, List[String]],
       val staticMethodsCalled: Map[String, List[String]],
@@ -57,6 +59,8 @@ object Infos {
   )
 
   object MethodInfo {
+    @deprecated("Use the overload with all the fields and no defaults",
+        "0.6.15")
     def apply(
         encodedName: String,
         isStatic: Boolean = false,
@@ -69,7 +73,29 @@ object Infos {
         accessedModules: List[String] = Nil,
         usedInstanceTests: List[String] = Nil,
         accessedClassData: List[String] = Nil): MethodInfo = {
+      apply(encodedName, isStatic, isAbstract, isExported,
+          staticFieldsRead = Map.empty, staticFieldsWritten = Map.empty,
+          methodsCalled, methodsCalledStatically, staticMethodsCalled,
+          instantiatedClasses, accessedModules, usedInstanceTests,
+          accessedClassData)
+    }
+
+    def apply(
+        encodedName: String,
+        isStatic: Boolean,
+        isAbstract: Boolean,
+        isExported: Boolean,
+        staticFieldsRead: Map[String, List[String]],
+        staticFieldsWritten: Map[String, List[String]],
+        methodsCalled: Map[String, List[String]],
+        methodsCalledStatically: Map[String, List[String]],
+        staticMethodsCalled: Map[String, List[String]],
+        instantiatedClasses: List[String],
+        accessedModules: List[String],
+        usedInstanceTests: List[String],
+        accessedClassData: List[String]): MethodInfo = {
       new MethodInfo(encodedName, isStatic, isAbstract, isExported,
+          staticFieldsRead, staticFieldsWritten,
           methodsCalled, methodsCalledStatically, staticMethodsCalled,
           instantiatedClasses, accessedModules, usedInstanceTests,
           accessedClassData)
@@ -131,6 +157,8 @@ object Infos {
     private var isAbstract: Boolean = false
     private var isExported: Boolean = false
 
+    private val staticFieldsRead = mutable.Map.empty[String, mutable.Set[String]]
+    private val staticFieldsWritten = mutable.Map.empty[String, mutable.Set[String]]
     private val methodsCalled = mutable.Map.empty[String, mutable.Set[String]]
     private val methodsCalledStatically = mutable.Map.empty[String, mutable.Set[String]]
     private val staticMethodsCalled = mutable.Map.empty[String, mutable.Set[String]]
@@ -156,6 +184,16 @@ object Infos {
 
     def setIsExported(isExported: Boolean): this.type = {
       this.isExported = isExported
+      this
+    }
+
+    def addStaticFieldRead(cls: String, field: String): this.type = {
+      staticFieldsRead.getOrElseUpdate(cls, mutable.Set.empty) += field
+      this
+    }
+
+    def addStaticFieldWritten(cls: String, field: String): this.type = {
+      staticFieldsWritten.getOrElseUpdate(cls, mutable.Set.empty) += field
       this
     }
 
@@ -233,14 +271,21 @@ object Infos {
     }
 
     def result(): MethodInfo = {
+      def toMapOfLists(
+          m: mutable.Map[String, mutable.Set[String]]): Map[String, List[String]] = {
+        m.mapValues(_.toList).toMap
+      }
+
       MethodInfo(
           encodedName = encodedName,
           isStatic = isStatic,
           isAbstract = isAbstract,
           isExported = isExported,
-          methodsCalled = methodsCalled.toMap.mapValues(_.toList),
-          methodsCalledStatically = methodsCalledStatically.toMap.mapValues(_.toList),
-          staticMethodsCalled = staticMethodsCalled.toMap.mapValues(_.toList),
+          staticFieldsRead = toMapOfLists(staticFieldsRead),
+          staticFieldsWritten = toMapOfLists(staticFieldsWritten),
+          methodsCalled = toMapOfLists(methodsCalled),
+          methodsCalledStatically = toMapOfLists(methodsCalledStatically),
+          staticMethodsCalled = toMapOfLists(staticMethodsCalled),
           instantiatedClasses = instantiatedClasses.toList,
           accessedModules = accessedModules.toList,
           usedInstanceTests = usedInstanceTests.toList,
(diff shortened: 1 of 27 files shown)


sjrd:
Review by @gzm0. Only the last commit belongs to this PR and needs to be reviewed here. The previous commits belong to #2705.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -849,10 +875,35 @@ private final class Analyzer(semantics: Semantics,
...
+      encodedName: String,
+      isStatic: Boolean = false,
+      isAbstract: Boolean = false,
+      methodsCalled: Map[String, List[String]] = Map.empty,
+      methodsCalledStatically: Map[String, List[String]] = Map.empty,
+      instantiatedClasses: List[String] = Nil
+  ): Infos.MethodInfo = {
+    Infos.MethodInfo(
+        encodedName,
+        isStatic = false,
```
gzm0: isStatic = isStatic (or remove the parameter)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -849,10 +875,35 @@ private final class Analyzer(semantics: Semantics,
...
+        isAbstract = false,
```
gzm0: isAbstract = isAbstract (or remove the parameter)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -270,6 +276,40 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
 
+  /** Emits the static fields of a linked class.
+   *
+   *  This is done after emitting the classes themselves, because defining
+   *  static fields of type `Long` requires to instantiate `RuntimeLong`, which
+   *  needs to be defined prior to that.
+   *
+   *  We also need to define all static fields before emitting all static
+   *  initializers, otherwise some static initializers could observe undefined
+   *  static fields from other classes.
```
gzm0: I think the lower part of this comment better belongs into `emit`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -270,6 +276,40 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+   */
+  private def emitLinkedClassStaticFields(
+      linkedClass: LinkedClass, builder: JSTreeBuilder): Unit = {
+
+    def addTree(tree: js.Tree): Unit = builder.addJSTree(tree)
```
gzm0: Is it really worth defining this here?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -270,6 +276,40 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+      addTree(classTreeCache.staticFields.getOrElseUpdate(
+          classEmitter.genCreateStaticFieldsOfScalaClass(linkedClass)(classCache)))
+    }
+  }
+
+  /** Emits the static initializer of a linked class, if any. */
+  private def emitLinkedClassStaticInitializer(
+      linkedClass: LinkedClass, builder: JSTreeBuilder): Unit = {
+
+    def addTree(tree: js.Tree): Unit = builder.addJSTree(tree)
```
gzm0: Same.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -203,7 +212,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
     val FieldDef(static, name, tpe, mutable) = fieldDef
     implicit val ctx = ErrorContext(fieldDef)
 
-    if (static && !classDef.kind.isJSClass)
+    if (static && classDef.kind.isJSType && !classDef.kind.isJSClass)
       reportError(s"FieldDef '$name' cannot be static")
```
gzm0: IIUC, this check is bogus. The only things it triggers on are `AbstractJSType | NativeJSClass | NativeJSModuleClass` and these do not allow fields on a much higher level.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -1143,17 +1175,20 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
-      new CheckedField(name, tpe, mutable)
+      val FieldDef(static, Ident(name, _), tpe, mutable) = fieldDef
+      new CheckedField(static, name, tpe, mutable)
     }
   }
 
-  private class CheckedField(val name: String, val tpe: Type,
-      val mutable: Boolean)
+  private class CheckedField(val static: Boolean, val name: String,
+      val tpe: Type, val mutable: Boolean)
```
gzm0: You should check for static field collisions (I should probably have said this when you introduced static fields). See #2627.

ir/src/main/scala/org/scalajs/core/ir/Infos.scala
```diff
@@ -233,14 +271,21 @@ object Infos {
     }
 
     def result(): MethodInfo = {
+      def toMapOfLists(
+          m: mutable.Map[String, mutable.Set[String]]): Map[String, List[String]] = {
+        m.mapValues(_.toList).toMap
```
sjrd: That's why I now call `.mapValues` before calling `.toMap`. The latter will force, so the laziness is not an issue.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -816,6 +819,11 @@ object Trees {
     val tpe = NoType
   }
 
+  case class TopLevelFieldExportDef(fullName: String, field: Ident)(
+      implicit val pos: Position) extends Tree {
+    val tpe = NoType
+  }
+
```
sjrd: Yes, I realized `TopLevelExportDef` didn't fit the bill for fields. Fields are a bit special because they need to be defined as proper Scala entities, and then they are exported. Where as top-level methods there is no Scala def to be aliased as an export. I think they're bound to be treated very differently, and so deserve separate IR nodes. I don't think we can consolidate the two `TopLevelExport` trees in one unique concept. You can't define a method eligible for export as a Scala entity first, because Scala entities cannot have `...rest` parameters, and for good reasons.

I guess in the 1.0.0 cleanup of `Trees.scala`, we'll simply have to rename `TopLevelExportDef` to `TopLevelMethodExportDef` and have it take a `MethodDef` rather than an arbitrary `Tree`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
sjrd: Uh oh! That's actually quite dramatic! It should definitely be `isStatic`!

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: No. Copy-paste ;)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: Indeed.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Right. But in that case, we can rename the tree in a binary compatible fashion. The rename does not change the binary format and the `MethodDef` only restriction is already in place, it's just not statically checked. Filed as #2725.

sjrd:
Updated.

ir/src/main/scala/org/scalajs/core/ir/Infos.scala
gzm0: Right :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -109,10 +109,30 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
 
       emitModuleImports(orderedClasses, builder, logger)
 
+      /* Emit all the classes, in the appropriate order:
+       *
+       * - First, all class definitions, which depend on nothing but their
+       *   superclasses.
+       * - Second, all static field definitions, which depend on nothing,
+       *   except those of type Long which need to instantiate RuntimeLong.
+       * - Third, all static initializer, which in the worst case can observe
```
gzm0: static initializer**s**

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -122,6 +122,19 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
         checkFieldDef(field, classDef)
       }
 
+      /* Check for static field collisions.
+       * TODO #2627 We currently cannot check instance field collisions because
+       * of #2382.
+       */
+      val staticFieldDefs = classDef.fields.filter(_.static)
```
gzm0: Just to double check: Fields and methods have a different namespace, right?

sjrd: For `Ident` statics, yes.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: Fixed.

PR #2718, opened 2017-01-09
Fix #2715: Emit object literals for new anonymous JS class.

### LINKED ISSUE #2715: Create literals from new @ScalaJSDefined trait instead of new Object
```scala
@ScalaJSDefined
trait Point extends js.Object {
  var x: js.UndefOr[Double] = js.undefined
  var y: js.UndefOr[Double] = js.undefined
}
 val p = new Point {x = 3; y = 5}

//fullOptCode
a=new f.Object;a.x=3;a.y=5;
```

from this SO answer : http://stackoverflow.com/a/21436082/986387 , looks like literal creation is faster! , can we get this in future versions of scala.js /GCC  ?

sjrd:
When instantiating an anonymous JS class, we actually instantiate the parent class. If that parent class happens to be `js.Object`, and there is no formal argument to the constructor, we can use an empty object literal `{}` instead.

This makes the compiler behavior consistent with a direct `new js.Object()`, for which we already applied this optimization. This allows Closure to fuse it with subsequent field assignments, eventually producing a direct object literal with fields.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -646,10 +646,16 @@ abstract class GenJSCode extends plugins.PluginComponent
           case js.JSSuperConstructorCall(args) =>
             implicit val pos = tree.pos
 
-            val ident =
-              origJsClass.superClass.getOrElse(sys.error("No superclass"))
-            val superTpe = jstpe.ClassType(ident.name)
-            val newTree = js.JSNew(js.LoadJSConstructor(superTpe), args)
+            val newTree = {
+              val ident =
+                origJsClass.superClass.getOrElse(sys.error("No superclass"))
+              if (args.isEmpty && ident.name == "sjs_js_Object") {
+                js.JSObjectConstr(Nil)
+              } else {
+                val superTpe = jstpe.ClassType(ident.name)
+                js.JSNew(js.LoadJSConstructor(superTpe), args)
+              }
+            }
 
             js.Block(
                 js.VarDef(selfName, jstpe.AnyType, mutable = false, newTree) ::
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
@@ -172,4 +172,55 @@ class OptimizationTest extends JSASTTest {
     }
   }
 
+  @Test
+  def newSJSDefinedTraitProducesObjectConstr: Unit = {
+    """
+    import scala.scalajs.js
+    import scala.scalajs.js.annotation._
+
+    @ScalaJSDefined
+    trait Point extends js.Object {
+      val x: Double
+      val y: Double
+    }
+
+    class Test {
+      def newSJSDefinedTraitProducesObjectConstr(): Any = {
+        new Point {
+          val x = 5.0
+          val y = 6.5
+        }
+      }
+    }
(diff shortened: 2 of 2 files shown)


sjrd:
Review by @gzm0

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -646,10 +646,16 @@ abstract class GenJSCode extends plugins.PluginComponent
...
             implicit val pos = tree.pos
 
-            val ident =
-              origJsClass.superClass.getOrElse(sys.error("No superclass"))
-            val superTpe = jstpe.ClassType(ident.name)
-            val newTree = js.JSNew(js.LoadJSConstructor(superTpe), args)
+            val newTree = {
+              val ident =
+                origJsClass.superClass.getOrElse(sys.error("No superclass"))
+              if (args.isEmpty && ident.name == "sjs_js_Object") {
```
gzm0: Put the string constant into `Definitions`?

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
```diff
@@ -172,4 +172,55 @@ class OptimizationTest extends JSASTTest {
...
+        new Point {
+          val x = 5.0
+          val y = 6.5
+        }
+      }
+    }
+    """.hasNot("`new Object`") {
+      case js.JSNew(_, _) =>
+    }.has("object literal") {
+      case js.JSObjectConstr(Nil) =>
```
gzm0: Are you sure you want to assert the `Nil` inside? Having the actual fields would be ok, too, wouldn't it?

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
```diff
@@ -172,4 +172,55 @@ class OptimizationTest extends JSASTTest {
...
+        new Point {
+          x = 5.0
+          y = 6.5
+        }
+      }
+    }
+    """.hasNot("`new Object`") {
+      case js.JSNew(_, _) =>
+    }.has("object literal") {
+      case js.JSObjectConstr(Nil) =>
```
gzm0: Same.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: No, `Definitions` is for things that are special at the IR level. `js.Object` is only special at the compiler level.

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
sjrd: Reasonably sure. It's not like we're going to the fields there from the compiler any time soon. And when it happens, I'd rather update my optimization test to assert that it *does* contain the fields, so I won't mind being reminded that this test exists by it failing.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: OMG. This will teach me to review things hastily. I thought this was in Desugaring :-/

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
gzm0: Ok, fair enough.

PR #2723, opened 2017-01-11
Fix #2709: Support reflective call of eq, ne, synchronized on AnyVal.

### LINKED ISSUE #2709: Reflective call `eq` and `ne` on value class
`eq` and `ne` are defined on `AnyRef` and not `Any`, which means that they can be implemented in value classes.

This leads to a wrong behaviour when doing a reflective call on one of these methods.

```scala
class Foo(val f: Int) extends AnyVal {
  def eq(other: AnyRef) = true
}

def callEq(obj: Any { def eq(other: AnyRef): Boolean }, other: AnyRef) = obj.eq(other)

callEq(new Foo(0), List())
``` 

This should return `true`, but returns `false`.

sjrd:
To implement, we drop support for reflective call of those methods on AnyRef instances, though. This is bug-compatible with the JVM.


sjrd:
Review by @gzm0

PR #2724, opened 2017-01-11
Fix #2708: Remove a years-old workaround for a Closure bug.

### LINKED ISSUE #518: Example with optimizeJS broken
Used to work in 0.4.2, broken in 0.4.3. For some reason Closure does not consider the assignments to `ScalaJS.e` reachable anymore in that setting (but it does for HelloWorld and Reversi, hum...)

### LINKED ISSUE #2708: fastOptJS code not working in JavaScriptCore(JSC)!
I am updating my code base to use new `@JSImport..` , all of sudden my mobile app(react-native) not working in `fastOptJS` mode .

screen shot 2017-01-04 at 3 43 47 pm

Environment  : 

Scalajs 0.16.14 

[JSC - No JIT](http://trac.webkit.org/wiki/JavaScriptCore)

>When using React Native, you're going to be running your JavaScript code in two environments:
On iOS simulators and devices, Android emulators and devices React Native uses JavaScriptCore which is the JavaScript engine that powers Safari. On iOS JSC doesn't use JIT due to the absence of writable executable memory in iOS apps.

sjrd:
In the original implementation of a custom exports namespace, 6680f72e8167af954d58797d26e943cf3326c4af, we introduced a workaround for an apparent bug of Closure: we forced some unelidable read of `ScalaJS.e` so that Closure would not get rid of it, and all our exports alongside it.

The workaround was refined later in 8793d47c9f88def176e927a2296fb82db893859a to fix #518.

It appears this workaround is not needed anymore. It might be because of the update of Closure, or because we have another read of `ScalaJS.e` at the beginning of the file, or anything else really. In any case, I cannot reproduce the bug anymore without the workaround.

Since that line was causing #2708, removing it fixes that issue.

### DIFF
--- a/tools/scalajsenv.js
+++ b/tools/scalajsenv.js
@@ -709,12 +709,6 @@ ScalaJS.typedArray2DoubleArray = function(value) {
   return new arrayClassData.constr(new ScalaJS.g["Float64Array"](value));
 };
 
-/* We have to force a non-elidable *read* of ScalaJS.e, otherwise Closure will
- * eliminate it altogether, along with all the exports, which is ... er ...
- * plain wrong.
- */
-this["__ScalaJSExportsNamespace"] = ScalaJS.e;
-
 // TypeData class
 
 //!if outputMode != ECMAScript6


sjrd:
Review by @gzm0 

/cc @chandu0101 Could you confirm that removing this line fixes #2708? An easy way to test is to simply manually remove that line in the produced `fastopt.js` file.

chandu0101:
> Could you confirm that removing this line fixes #2708? An easy way to test is to simply manually remove that line in the produced fastopt.js file.

yes , i can confirm that after removing that line from `fastOptJS` output file , no errors while loading in JSC and apps working as expected.

Issue #2727, opened 2017-01-12
Why different bound checks in ThreadLocalRandom next*(min, max) methods


fhalde:
In https://github.com/scala-js/scala-s/blob/master/javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala

for `nextInt(min, max)`, we do

`if (difference > 0)`

but for `nextLong(min, max)`, we do

`if (different >= 0)`

Am I missing a corner case ? Or is this how it should be ?

sjrd:
It doesn't matter. The difference cannot be 0 because we check before that `max > min`.

Please refrain from using GitHub issues to ask questions. They are reserved for issues and feature requests. Use the Gitter channel or, in this case, you could simply have mentioned me in https://github.com/scala/scala/pull/5611

fhalde:
I remember drafting this. Not sure when I posted this!

PR #2728, opened 2017-01-12
Consistent condition checking in nextInt and nextLong

fhalde:
On calling `nextInt(0, IntMax)` there is an unnecessary switch to the recursive approach. This change tries to fix that corner case

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
@@ -78,7 +78,7 @@ class ThreadLocalRandom extends Random {
       throw new IllegalArgumentException()
 
     val difference = bound - least
-    if (difference >= 0) {
+    if (difference > 0) {
       nextLong(difference) + least
     } else {
       /* The interval size here is greater than Long.MaxValue,


gzm0:
Jenkins, ok to test

javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala
```diff
@@ -28,7 +28,7 @@ class ThreadLocalRandom extends Random {
       throw new IllegalArgumentException()
 
     val difference = bound - least
-    if (difference > 0) {
+    if (difference >= 0) {
```
gzm0: The equals condition can never occur, since it would require `least == bound`, which is illegal.

gzm0: The case you describe (`nextInt(0, MaxInt)`) would cause, `difference == MaxInt`.

sjrd:
As I said in #2727, the case where `difference == 0` cannot happen by construction.

I welcome the change to make `nextInt` and `next Long` consistent, though. But the commit message should then explain that this is only for consistency of the code and doesn't actually change anything.

fhalde:
Yes correct. Rather to confuse users why the `equality` is there, let's make `nextLong` same as `nextInt`

sjrd:
Your commit is somehow not attributed to your GitHub account. Did you use the correct e-mail address when creating the commit? You can fix it as explained [here](http://stackoverflow.com/a/3042512/1829647).

fhalde:
Ahh. I simultaneously work for my company and personally. so it keeps getting mixed, fixing it. thanks for spotting

sjrd:

PR #2730, opened 2017-01-13
Fix #2719: Throw StringIndexOOBE in String{Buffer,Builder}.

### LINKED ISSUE #2719: Minor inconsistency in StringBuilder/Buffer error handling
The default semantics of out of bounds failure for StringBuilder and StringBuffer is to throw StringIndexOutOfBounds, not just IndexOutOfBounds:

```
scala> val buf = new java.lang.StringBuilder
buf: StringBuilder =

scala> buf.append("foobar")
res0: StringBuilder = foobar

scala> buf.setCharAt(2, 'x')

scala> buf.setCharAt(5, 'h')

scala> buf.setCharAt(-1, 'h')
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
  at java.lang.AbstractStringBuilder.setCharAt(AbstractStringBuilder.java:407)
  at java.lang.StringBuilder.setCharAt(StringBuilder.java:76)
  ... 29 elided
```

Currently this is not the case on Scala.js. In particular at https://github.com/scala-js/scala-js/blob/master/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/StringBufferTest.scala#L101 and https://github.com/scala-js/scala-js/blob/master/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/StringBufferTest.scala#L114.

Discovered while porting StringBuilder/Buffer tests to Scala Native.

sjrd:


sjrd:
Review by @gzm0, for the sake of it :p

PR #2736, opened 2017-01-15
Implement checked behavior of ArrayIndexOutOfBoundsExceptions.

sjrd:
This brings support for `ArrayIndexOutOfBoundsExceptions` on par with `ClassCastException`s, i.e., they are controlled by a `CheckedBehavior` setting in `scalaJSSemantics`.

The checked behavior is Fatal by default in fastOpt and Unchecked in fullOpt, as `asInstanceOfs`.


sjrd:
Review by @gzm0 
I focused on `ArrayIndexOutOfBounds` because it's the one tripping people most often. It's particularly bad because it silently returns `undefined` when getting out-of-bounds elements, which cause very obscure error messages down the line.

sjrd:
Uh oh: https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/RedBlackTree.scala#L488-L506

gzm0:
I take it that review of this is on hold until you figured out how to deal with the Scalalib?

sjrd:
I'll basically need an override to fix the bug they're covering with the try/catch. The implementation of the checked behavior will not change, though, so you can start reviewing that if you want.

sjrd:
https://github.com/scala/scala/pull/5645

sjrd:
The only reason really is that I preserved the message check when previous tests had the message check. But for new tests I did not test messages, mostly because we typically don't. I can add message checks everywhere if you want.

gzm0:
> I can add message checks everywhere if you want.

Nah, it's fine. Was just wondering.

sjrd:
Updated with the appropriate overrides. The last commit is unchanged compared the last review.

sjrd:
> The last commit is unchanged compared the last review.

... except that there's a change in checksizes, now.

PR #2739, opened 2017-01-16
Fix #2738: Implement Float and Double methods from JDK8.

### LINKED ISSUE #2738: Some methods from JDK 8 are missing
For example calling `java.lang.Double.isFinite(1.0)` results in

```
Referring to non-existent method java.lang.Double$.isFinite(scala.Double)scala.Boolean
```

Another ones are static `hashCode`, `min`, `max`, `sum` and probably others annotated with `@since 1.8`. The same applies for `java.lang.Float`.

Relates to #1981

sjrd:
* `hashCode`
* `isFinite`
* `max`
* `min`
* `sum`
* and `toHexString` (actually this one is from JDK5)

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Double.scala
+++ b/javalanglib/src/main/scala/java/lang/Double.scala
@@ -51,10 +51,12 @@ object Double {
   final val NEGATIVE_INFINITY = 1.0 / -0.0
   final val NaN = 0.0 / 0.0
   final val MAX_VALUE = scala.Double.MaxValue
+  final val MIN_NORMAL = 2.2250738585072014e-308
   final val MIN_VALUE = scala.Double.MinPositiveValue
   final val MAX_EXPONENT = 1023
   final val MIN_EXPONENT = -1022
   final val SIZE = 64
+  final val BYTES = 8
 
   @inline def valueOf(doubleValue: scala.Double): Double =
     new Double(doubleValue)
--- a/javalanglib/src/main/scala/java/lang/Double.scala
+++ b/javalanglib/src/main/scala/java/lang/Double.scala
@@ -29,7 +29,7 @@ final class Double private () extends Number with Comparable[Double] {
   }
 
   @inline override def hashCode(): Int =
-    scala.scalajs.runtime.Bits.numberHashCode(doubleValue)
+    Double.hashCode(doubleValue)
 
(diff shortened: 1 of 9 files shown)


sjrd:
Review by @gzm0.

javalanglib/src/main/scala/java/lang/Double.scala
```diff
@@ -82,6 +84,61 @@ object Double {
   @inline def toString(d: scala.Double): String =
     "" + d
 
+  def toHexString(d: scala.Double): String = {
```
sjrd: [JavaDoc of `Double.toHexString`](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#toHexString-double-)

javalanglib/src/main/scala/java/lang/Float.scala
```diff
@@ -65,21 +66,81 @@ object Float {
   @inline def toString(f: scala.Float): String =
     "" + f
 
+  def toHexString(f: scala.Float): String = {
```
sjrd: [JavaDoc of `Float.toHexString`](https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#toHexString-float-)

javalanglib/src/main/scala/java/lang/Double.scala
```diff
@@ -51,10 +51,12 @@ object Double {
   final val NEGATIVE_INFINITY = 1.0 / -0.0
   final val NaN = 0.0 / 0.0
   final val MAX_VALUE = scala.Double.MaxValue
+  final val MIN_NORMAL = 2.2250738585072014e-308
```
sjrd: I got it out of `java.lang.Double.MIN_NORMAL` in Scala REPL ;)

javalanglib/src/main/scala/java/lang/Float.scala
```diff
@@ -50,10 +50,12 @@ object Float {
   final val NEGATIVE_INFINITY = 1.0f / -0.0f
   final val NaN = 0.0f / 0.0f
   final val MAX_VALUE = scala.Float.MaxValue
+  final val MIN_NORMAL = 1.17549435e-38f
```
sjrd: Same here with `Float.MIN_NORMAL`.

PR #2740, opened 2017-01-17
Fix #2733: Implement on-demand reflective instantiation.

### LINKED ISSUE #2733: Proposal: Principled replacement for "reflective" instantiation
The discussion in Proposal #2720 raised an interesting point: `@JSExport`ing Scala classes and objects is weird if we deprecate instance member exports. However, testing frameworks (ab)use `@JSExportDescendantClasses` and `@JSExportDescendantObjects` to "reflectively" instantiate test classes.

I have checked the following testing frameworks:

- ScalaTest
- ScalaCheck
- catalysts/SpecLite
- uTest
- MiniTest
- GreenLight
- ScalaProps

All of them use our [`TestUtils`](https://github.com/scala-js/scala-js/blob/master/test-interface/src/main/scala/org/scalajs/testinterface/TestUtils.scala) to instantiate test classes and loading test modules. Moreover, for those that use `newInstance()` (as opposed to `loadModule()`), they always use 0 arguments to the constructor.

Besides testing frameworks, it seems that `@JSExportDescendantClasses` is basically unused, except by Akka.js. I will come back to the topic of Akka.js later.

## The proposal

I propose we introduce a new core API specifically addressing the needs of reflective class instantiation/module loading. Its public API would look like the following. Names and signatures are tentative; I'd love to receive suggestions for better names.
```scala
package scala.scalajs.reflect

/** Enable `Reflect.lookupInstantiatableClass()` for all descendants of the annotated class
 *  that have a public constructor with 0 argument or that are singleton objects.
 */
class EnableReflectiveInstantiation extends StaticAnnotation

final class InstantiatableClass[T <: AnyRef] private (
    val runtimeClass: Class[T],
    val isModule: Boolean,
    newInstanceFun: js.Function0[T]
) {
  // Intended to be used with non-module classes
  def newInstance(): T = newInstanceFun()
  // Intended to be used with module classes
  def loadModule(): T = newInstanceFun()
}

object Reflect {
  def lookupInstantiatableClass(fqcn: String): Option[InstantiatableClass[_ <: AnyRef]] =
    ???
}
```
Alternatively, we could separate classes from module classes in the API.

## Implementation

Implementation-wise, `Reflect` maintains a lookup dictionary, and has an internal method to register classes:
```scala
object Reflect {
  private val instantiatableClasses = js.Dictionary.empty[InstantiableClass[_]]

  // `protected[reflect]` makes it public in the IR
  protected[reflect] def registerInstantiatableClass[T <: AnyRef](
      fqcn: String, runtimeClass: Class[T], isModule: Boolean, newInstanceFun: js.Function0[T]): Unit = {
    instantiatableClasses(fqcn) = new InstantiatableClass(runtimeClass, isModule, newInstanceFun)
  }

  def lookupInstantiatableClass(fqcn: String): Option[InstantiatableClass[_ <: AnyRef]] =
    instantiatableClasses.get(fqcn)
}
```
The missing part is ... who calls `Reflect.registerInstantiatableClass` and when? Well, now that our IR has static initializers anyway, we can build on top of those in our back-end. For every class `pack.C` extending a type annotated with `@EnableReflectiveInstantiation` with a public no-arg constructor, we synthesize a static initializer with (pseudo-IR with pretty names):
```scala
def clinit___() {
  scala.scalajs.reflect.Reflect.registerInstantiatableClass(
      "pack.C", classOf[pack.C], false, lambda<>() = { new pack.C() })
}
```
And for every module class `pack.O$`, we generate:
```scala
def clinit___() {
  scala.scalajs.reflect.Reflect.registerInstantiatableClass(
      "pack.O$", classOf[pack.O$], true, lambda<>() = { mod:pack.O$ })
}
```
The nice thing about this strategy is that only the `compiler/` and `library/` need to be changed. In other words, the IR spec is unchanged, and consequently the analyzer, optimizer and emitter do not need to be aware of all of this.

## Migration path

Since testing frameworks all use `TestUtils`, we can provide a pretty easy migration path. We simply have to update `TestUtils.loadModule()` and `TestUtils.newInstance()` to try `Reflect.lookupInstantiatableClass` before the `exportsNamespace`.

`@JSExportDescendantObjects` and `@JSExportDescendantClasses` would be deprecated with a message telling to use `@EnableReflectiveInstantiation` instead.

In 1.0.0, we can remove `@JSExportDescendantObjects` and `@JSExportDescendantClasses` as well as the backup path in `TestUtils`. We will also be able to remove `exportsNamespace`, which will be a relief when emitting ES6 modules.

## Akka.js

Akka.js, as always, uses more reflection than everything else. The crucial part is concentrated in [`ReflectiveDynamicAccess.scala`](https://github.com/akka-js/akka.js/blob/354902581e0a7068a09f9f4f2017f5631689794a/akka-js-actor/js/src/main/scala/akka/actor/ReflectiveDynamicAccess.scala). Contrary to testing frameworks (and other rare uses of `@JSExportDescendantClasses`), Akka.js does use constructors with 1 ore more arguments.

Such a use case is not supported by the replacement `@EnableReflectiveInstantiation`. However, it is supported by [`scalajs-reflect`](https://github.com/sjrd/scalajs-reflect), which was specifically developed for the use cases of Akka.js quite a while ago.

## Alternative with all public constructors

Alternatively to the above proposal which restricts itself to the no-arg constructors, we could bring in the full feature of [`scalajs-reflect`](https://github.com/sjrd/scalajs-reflect) and support all declared public constructors. The API and implementation would look like the following. Here we really need to separate normal classes from module classes for sanity.
```scala
final class LoadableModuleClass[T <: AnyRef] private (
    val runtimeClass: Class[T],
    loadModuleFun: js.Function[T]
) {
  def loadModule(): T = loadModuleFun()
}

final class InstantiatableClass[T <: AnyRef] private (
    val runtimeClass: Class[T],
    val declaredConstructors: immutable.Seq[InvokableConstructor[T]]
)

final class InvokableConstructor[T <: AnyRef] private (
    val parameterTypes: immutable.Seq[Class[_]],
    newInstanceFun: js.Function
) {
  def newInstance(args: Any*): T =
    newInstanceFun.asInstanceOf[js.Dynamic].apply(args.asInstanceOf[Seq[js.Any]]: _*)
}

object Reflect {
  private val loadableModuleClasses = js.Dictionary.empty[LoadableModuleClas[_]]
  private val instantiatableClasses = js.Dictionary.empty[InstantiableClass[_]]

  // `protected[reflect]` makes it public in the IR
  protected[reflect] def registerLoadableModuleClass[T <: AnyRef](
      fqcn: String, runtimeClass: Class[T], loadModuleFun: js.Function0[T]): Unit = {
    loadModuleClasses(fqcn) = new LoadableModuleClass(runtimeClass, newInstanceFun)
  }

  protected[reflect] def registerInstantiatableClass[T <: AnyRef](
      fqcn: String, runtimeClass: Class[T],
      constructors: js.Array[js.Tuple2[js.Array[Class[_]], js.Function]]): Unit = {
    instantiatableClasses(fqcn) = new InstantiatableClass(runtimeClass,
        constructors.map(c => new InvokableConstructor(c._1.toList, c._2)).toList)
  }

  def lookupLoadableModuleClass(fqcn: String): Option[LoadableModuleClass[_ <: AnyRef]] =
    loadableModuleClasses.get(fqcn)

  def lookupInstantiatableClass(fqcn: String): Option[InstantiatableClass[_ <: AnyRef]] =
    instantiatableClasses.get(fqcn)
}
```
This would be slightly more complicated to implement, but would provide Akka.js with everything it needs out of the box.

sjrd:
This commit implements the variant of Proposal #2733 that supports all public constructors.

We introduce a new API in `scala.scalajs.reflect` which allows to do some amount of run-time reflection, namely, instantiate classes and load module instances given their fully qualified names.

Not all classes are eligible. Only objects and classes that have an ancestor annotated with `@EnableReflectiveInstantiation` can be found and instantiated that way. Moreover, traits, abstract classes, and classes without any public constructors cannot be found either.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -420,19 +420,42 @@ abstract class GenJSCode extends plugins.PluginComponent
 
         val topLevelExports = genTopLevelExports(sym)
 
-        val needsStaticInitializer =
-          topLevelExports.exists(_.isInstanceOf[js.TopLevelFieldExportDef])
-        val optStaticInitializer =
-          if (!needsStaticInitializer) Nil
-          else genStaticInitializerLoadingModule(sym) :: Nil
+        memberExports ++ exportedConstructorsOrAccessors ++ topLevelExports
+      }
 
-        (memberExports ++ exportedConstructorsOrAccessors ++ topLevelExports ++
-            optStaticInitializer)
+      // Static initializer
+      val optStaticInitializer = {
+        // Initialization of reflection data, if required
+        val reflectInit = {
+          val enableReflectiveInstantiation = {
+            (sym :: sym.ancestors).exists { ancestor =>
+              ancestor.hasAnnotation(EnableReflectiveInstantiationAnnotation)
+            }
+          }
+          if (enableReflectiveInstantiation)
+            genRegisterReflectiveInstantiation(sym)
+          else
+            None
+        }
+
+        // Initialization of the module because of field exports
+        val needsStaticModuleInit =
+          exports.exists(_.isInstanceOf[js.TopLevelFieldExportDef])
+        val staticModuleInit =
+          if (!needsStaticModuleInit) None
+          else Some(genLoadModule(sym))
+
+        val staticInitializerStats =
+          reflectInit.toList ::: staticModuleInit.toList
+        if (staticInitializerStats.nonEmpty)
+          Some(genStaticInitializerWithStats(js.Block(staticInitializerStats)))
+        else
+          None
       }
 
       // Hashed definitions of the class
       val hashedDefs =
-        Hashers.hashDefs(generatedMembers ++ exports)
+        Hashers.hashDefs(generatedMembers ++ exports ++ optStaticInitializer)
 
       // The complete class definition
       val kind =
@@ -521,7 +544,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           }
           if (exports.exists(_.isInstanceOf[js.FieldDef])) {
             val staticInitializer =
-              genStaticInitializerLoadingModule(companionModuleClass)
+              genStaticInitializerWithStats(genLoadModule(companionModuleClass))
             exports :+ staticInitializer
           } else {
             exports
@@ -920,15 +943,99 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     // Static initializers -----------------------------------------------------
 
-    private def genStaticInitializerLoadingModule(sym: Symbol)(
+    private def genStaticInitializerWithStats(stats: js.Tree)(
         implicit pos: Position): js.MethodDef = {
       js.MethodDef(
           static = true,
           js.Ident(ir.Definitions.StaticInitializerName),
           Nil,
           jstpe.NoType,
-          Some(genLoadModule(sym)))(
+          Some(stats))(
           OptimizerHints.empty, None)
+    }
+
+    private def genRegisterReflectiveInstantiation(sym: Symbol)(
+        implicit pos: Position): Option[js.Tree] = {
+      if (sym.isModuleClass)
+        genRegisterReflectiveInstantiationForModuleClass(sym)
+      else
+        genRegisterReflectiveInstantiationForNormalClass(sym)
+    }
+
+    private def genRegisterReflectiveInstantiationForModuleClass(sym: Symbol)(
+        implicit pos: Position): Option[js.Tree] = {
+      val fqcnArg = js.StringLiteral(sym.fullName + "$")
+      val runtimeClassArg = js.ClassOf(toReferenceType(sym.info))
+      val loadModuleFunArg = js.Closure(Nil, Nil, genLoadModule(sym), Nil)
+
+      val stat = genApplyMethod(
+          genLoadModule(ReflectModule),
+          Reflect_registerLoadableModuleClass,
+          List(fqcnArg, runtimeClassArg, loadModuleFunArg))
+
+      Some(stat)
+    }
+
+    private def genRegisterReflectiveInstantiationForNormalClass(sym: Symbol)(
+        implicit pos: Position): Option[js.Tree] = {
+      val ctors =
+        if (sym.isAbstractClass) Nil
+        else sym.info.member(nme.CONSTRUCTOR).alternatives.filter(_.isPublic)
+
+      if (ctors.isEmpty) {
+        None
+      } else {
+        val constructorsInfos = for {
+          ctor <- ctors
+        } yield {
+          withNewLocalNameScope {
+            val (parameterTypes, formalParams, actualParams) = (for {
+              param <- ctor.tpe.params
+            } yield {
+              /* Note that we do *not* use `param.tpe` entering posterasure
+               * (neither to compute `paramType` nor to give to `fromAny`).
+               * Logic would tell us that we should do so, but we intentionally
+               * do not to preserve the behavior on the JVM regarding value
+               * classes. If a constructor takes a value class as parameter, as
+               * in:
+               *
+               *   class ValueClass(val underlying: Int) extends AnyVal
+               *   class Foo(val vc: ValueClass)
+               *
+               * then, from a reflection point of view, on the JVM, the
+               * constructor of `Foo` takes an `Int`, not a `ValueClas`. It
+               * must therefore be identified as the constructor whose
+               * parameter types is `List(classOf[Int])`, and when invoked
+               * reflectively, it must be given an `Int` (or `Integer`).
+               */
+              val paramType = js.ClassOf(toReferenceType(param.tpe))
+              val paramDef = js.ParamDef(encodeLocalSym(param), jstpe.AnyType,
+                  mutable = false, rest = false)
+              val actualParam = fromAny(paramDef.ref, param.tpe)
+              (paramType, paramDef, actualParam)
+            }).unzip3
+
+            val paramTypesArray = js.JSArrayConstr(parameterTypes)
+
+            val newInstanceFun = js.Closure(Nil, formalParams, {
+              genNew(sym, ctor, actualParams)
+            }, Nil)
+
+            js.JSArrayConstr(List(paramTypesArray, newInstanceFun))
+          }
+        }
+
+        val fqcnArg = js.StringLiteral(sym.fullName)
+        val runtimeClassArg = js.ClassOf(toReferenceType(sym.info))
+        val ctorsInfosArg = js.JSArrayConstr(constructorsInfos)
+
+        val stat = genApplyMethod(
+            genLoadModule(ReflectModule),
+            Reflect_registerInstantiatableClass,
+            List(fqcnArg, runtimeClassArg, ctorsInfosArg))
+
+        Some(stat)
+      }
     }
 
     // Constructor of a Scala.js-defined JS class ------------------------------
--- a/library/src/main/scala/scala/scalajs/reflect/Reflect.scala
+++ b/library/src/main/scala/scala/scalajs/reflect/Reflect.scala
@@ -1,0 +1,102 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2017, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package scala.scalajs.reflect
+
+import scala.collection.mutable
+
+import scala.scalajs.js
+
+final class LoadableModuleClass private[reflect] (
+    val runtimeClass: Class[_],
+    loadModuleFun: js.Function0[Any]
+) {
+  /** Loads the module instance and returns it. */
+  def loadModule(): Any = loadModuleFun()
+}
+
+final class InstantiatableClass private[reflect] (
+    val runtimeClass: Class[_],
+    val declaredConstructors: List[InvokableConstructor]
+) {
+  /** Instantiates a new instance of this class using the zero-argument
+   *  constructor.
+   *
+   *  @throws java.lang.InstantiationException (caused by a
+   *    `NoSuchMethodException`)
+   *    If this class does not have a public zero-argument constructor.
+   */
+  def newInstance(): Any = {
+    getConstructor().fold[Any] {
+      throw new InstantiationException(runtimeClass.getName).initCause(
+          new NoSuchMethodException(runtimeClass.getName + ".<init>()"))
+    } { ctor =>
+      ctor.newInstance()
+    }
+  }
+
+  /** Looks up a public constructor identified by the types of its formal
+   *  parameters.
+   *
+   *  If no such public constructor exists, returns `None`.
+   */
+  def getConstructor(parameterTypes: Class[_]*): Option[InvokableConstructor] =
+    declaredConstructors.find(_.parameterTypes.sameElements(parameterTypes))
+}
+
+final class InvokableConstructor private[reflect]  (
+    val parameterTypes: List[Class[_]],
+    newInstanceFun: js.Function
+) {
+  def newInstance(args: Any*): Any = {
+    /* Check the number of actual arguments. We let the casts and unbox
+     * operations inside `newInstanceFun` take care of the rest.
+     */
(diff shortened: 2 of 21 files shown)


sjrd:
I'm not 100% happy with the `InvokableConstructor` API because it tends to produce warnings about existential types, e.g.,
```scala
val clazz: InstantiatableClass[_] = ???
val ctors = clazz.declaredConstructors
```
yields:

> inferred existential type `List[scala.scalajs.reflect.InvokableConstructor[_$6]] forSome { type _$6 }`, which cannot be expressed by wildcards, should be enabled by making the implicit value `scala.language.existentials` visible. [...]

but
```scala
val ctors: List[InvokableConstructor[_]] = clazz.declaredConstructors
```
does not have that issue.

Maybe I'll try with a type member in `InvokableConstructor` instead.

sjrd:
Aargh! GCC is "dead-code"-eliminating calls to `registerLoaduleModuleClass` :sob:

andreaTP:
This proposal looks really promising IMO :+1:

library/src/main/scala/scala/scalajs/reflect/Reflect.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+object Reflect {
+  /* I would like those val's to be `js.Dictionary`'es instead of full-blown
+   * Scala Maps. But if I do that, GCC "dead-code"-eliminates away statements
+   * that fill `loadableModuleClasses`! We fool it into not making a fool of
+   * itself by using more complicated data structures that it does not
+   * understand -_-'.
+   */
```
sjrd: *This*, is the ugliest thing I have ever done in this codebase. But honestly I don't know what else to do :(

andreaTP: no other ways to do it? i.e. exporting to JS something closely related

sjrd: Exporting to JS could be annihilated by a bundler that performs sub-module dce, because that export would effectively be unused.

andreaTP: this could be a problem even in end user code...
probably you should find a way to effectively avoid GCC in this case, correct?

sjrd: It won't be a problem in user code with the `mutable.Map`s, because nothing changes between our own test suite and user code.

sjrd:
Review by @gzm0 

I'm considering simply removing the type parameters to `InstantiatableClass`, `InvokableConstructor` and `LoadableModuleClass`. They're more trouble than they're worth. WDYT?

sjrd:
retest this please

sjrd:
retest this please

test-interface/src/main/scala/org/scalajs/testinterface/TestUtils.scala
```diff
@@ -1,24 +1,145 @@
...
+      "Use the overload with explicit formal constructor parameter types.",
+      "0.6.15")
+  def newInstance(name: String, loader: ClassLoader)(
+      args: Seq[AnyRef]): AnyRef = {
+    Reflect.lookupInstantiatableClass(name).fold[AnyRef] {
+      val ctor = deepSelect(namespace(loader), name)
+      js.Dynamic.newInstance(ctor)(args.asInstanceOf[Seq[js.Any]]: _*)
+    } { clazz =>
+      val ctor = guessAppropriateCtor(clazz.declaredConstructors, args)
+      ctor.newInstance(args: _*).asInstanceOf[AnyRef]
```
sjrd: In fact, [as we can see here](https://github.com/sjrd/scala-js/blob/v0.6.14/stubs/src/main/scala/org/scalajs/testinterface/TestUtils.scala#L28-L31), the JVM version of `newInstance` in master doesn't even try to resolve overloaded constructors. If there is more than one public constructor, I think it would be fair to bail here as well, instead of this `guessAppropriateCtor` thing which is super fragile.

sjrd: Maybe a better way to avoid that scenario would be not to support `Reflect`-based instantiation at all in this overload, and instead always throw an error (with an informative message). That way in the above scenario, *any* test will immediately crash, so the testing framework is forced to switch to the new overload early.

gzm0: Yes, I think just failing is better. When a test framework changes it's annotations on its tests, it can also change the overload.

test-interface/src/main/scala/org/scalajs/testinterface/internal/FrameworkDetector.scala
```diff
@@ -0,0 +1,37 @@
+package org.scalajs.testinterface.internal
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+import scala.scalajs.js.JSConverters._
+import scala.scalajs.reflect.Reflect
+
+private[internal] object FrameworkDetector {
+  @JSExportTopLevel("org.scalajs.testinterface.internal.detectFrameworks")
+  def detectFrameworks(exportsNamespace: js.Dynamic,
```
gzm0: Do we really need to inject the `exportsNamespace`? Can't we just use `scala.scalajs.runtime.environmentInfo.exportsNamespace`?

sbt-plugin-test/multiTest/js/src/test/scala/sbttest/multitest/JUnitUtil.scala
```diff
@@ -3,17 +3,16 @@ package sbttest.multitest
...
   private final val bootstrapperSuffix = "$scalajs$junit$bootstrapper"
 
   def loadBootstrapper(classFullName: String): JUnitTestBootstrapper = {
     val fullName = s"$classFullName$bootstrapperSuffix"
     try {
-      fullName.split('.').foldLeft(js.Dynamic.global) { (obj, n) =>
-        obj.selectDynamic(n)
-      }.apply().asInstanceOf[JUnitTestBootstrapper]
+      val modClass = Reflect.lookupLoadableModuleClass(fullName + "$").get
+      modClass.loadModule().asInstanceOf[JUnitTestBootstrapper]
```
gzm0: Why don't we use `TestUtil` here?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -420,19 +420,42 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          exports.exists(_.isInstanceOf[js.TopLevelFieldExportDef])
+        val staticModuleInit =
+          if (!needsStaticModuleInit) Nil
+          else genLoadModule(sym) :: Nil
 
-        (memberExports ++ exportedConstructorsOrAccessors ++ topLevelExports ++
-            optStaticInitializer)
+        if (reflectInit.nonEmpty || staticModuleInit.nonEmpty) {
+          genStaticInitializerWithStats(
+              js.Block(reflectInit ::: staticModuleInit)) :: Nil
```
sjrd: It was an efficiency concern, but probably cargo-cult based. Concatenating first should be better overall.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -420,19 +420,42 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        } else {
+          Nil
+        }
```
sjrd: Indeed. Same cargo-cult performance thingy. I was in the mood, it seems :p

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -920,17 +943,90 @@ abstract class GenJSCode extends plugins.PluginComponent
...
           js.Ident(ir.Definitions.StaticInitializerName),
           Nil,
           jstpe.NoType,
-          Some(genLoadModule(sym)))(
+          Some(stats))(
           OptimizerHints.empty, None)
     }
 
+    private def genRegisterReflectiveInstantiation(sym: Symbol)(
+        implicit pos: Position): List[js.Tree] = {
```
sjrd: Historical. At some point this returned several statements. It doesn't anymore, so it should definitely be `Option`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -920,17 +943,90 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+          ctor <- ctors
+        } yield {
+          withNewLocalNameScope {
+            val (parameterTypes, formalParams, actualParams) = (for {
+              param <- ctor.tpe.params
+            } yield {
+              /* Note that we do *not* use `param.tpe` entering posterasure.
+               * Logic would tell us that we should do so, but we intentionally
+               * do not to preserve "bug"-compatibility with the JVM regarding
+               * value classes.
```
sjrd: It applies to all of them. I'll make that clearer.

The "bug" is that if you have:
```scala
class ValueClass(val underlying: Int) extends AnyVal

class Foo(val vc: ValueClass)
```
and you try to instantiate it with reflection, you need to ask for a constructor with a parameter of type `Int` (the underlying type of `ValueClass`), and of provide an actual argument of type `Int(eger)`. This breaks the abstraction of value classes.

This is a recurring theme with value classes, where "reflection" is involved. For example, reflective proxies work with the underlying types of value classes, even though they have a boxed context.

library/src/main/scala/scala/scalajs/reflect/Reflect.scala
```diff
@@ -0,0 +1,95 @@
...
+final class InstantiatableClass private[reflect] (
+    val runtimeClass: Class[_],
+    val declaredConstructors: List[InvokableConstructor]
+) {
+  /** Instantiates a new instance of this class using the zero-argument
+   *  constructor.
+   *
+   *  @throws java.lang.InstantiationException (caused by a
+   *    `NoSuchMethodException`)
+   *    If this class is abstract or does not have a public zero-argument
```
sjrd: Indeed. The comment is obsolete. At some point I did return an `InstantiatableClass` with no `declaredConstructors`, but then I thought better of it (for the reason that you give).

library/src/main/scala/scala/scalajs/reflect/Reflect.scala
sjrd: Well ... there's obviously a bug somewhere in GCC. So in theory users could trip against it too, for unrelated reasons. But unless I spend I-don't-know-how-much-time trying to minimize the bug for GCC, there's not much we can do about it but working around it in this very specific case.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/FrameworkDetector.scala
```diff
@@ -52,18 +55,33 @@ private[sbtplugin] final class FrameworkDetector(jsEnv: JSEnv,
...
+            console.log("$ConsoleFrameworkPrefix" + (results[i] || ""));
           }
-          if (!gotOne) {
-            // print an empty line with prefix to zip afterwards
-            console.log("$ConsoleFrameworkPrefix");
+        } else {
+          /* Backwards compatibility. This fallback is necessary if there is
+           * no testing framework using the new test-interface from 0.6.15.
+           * We still need to discover frameworks in that case.
+           */
```
sjrd: Oh! I didn't even know about that. That simplifies things a bit :)

test-interface/src/main/scala/org/scalajs/testinterface/internal/FrameworkDetector.scala
sjrd: Indeed. I blindly ported the parameter from the JS code in `FrameworkDetector`. We definitely can use `envInfo.exportsNamespace` instead :)

sbt-plugin-test/multiTest/js/src/test/scala/sbttest/multitest/JUnitUtil.scala
sjrd: This one is intentional. I want to test that JUnit indeed works through `Reflect`, and not through the fall-back of `exportsNamespace`. This is also the reason this specific change was applied in the same commit that changes JUnit to use `@EnableReflectiveInstantiation`, whereas all changes to `TestUtils` are done in the commit "Prepare the testing interface ...".

sjrd:
Updated.

library/src/main/scala/scala/scalajs/reflect/Reflect.scala
gzm0: Ok, fair enough. Filed #2749 so we don't forget to try and update this.

Issue #2720, opened 2017-01-10
Proposal: Deprecate exports (except @JSExportTopLevel and @JSExportStatic)


sjrd:
**Edit:** The original proposal was to only deprecate `@JSExport` on instance members on Scala classes. The discussion showed that, once those are deprecated, there are little uses left for `@JSExport` on top-level classes and objects either. The use cases for those are real, but can be replaced with better alternatives, which were spun off as Proposals #2733 and #2734. The current text of the proposal includes those changes, and deprecates all uses of `@JSExport`, `@JSExportDescendantClasses`/`Objects`, and `@JSExportAll`.

Here is a potentially crazy idea: what if we deprecated (and removed) all exports except `@JSExportTopLevel` and `@JSExportStatic`?

Let us immediately see what exactly falls in that category:

## What does this proposal concern?

```scala
// Export members of Scala classes
object Foo {
  @JSExport def bar() = ???
  @JSExport val foobar = ???
  @JSExport var baz = ???
}

class Babar(@JSExport val x: Int)

// Export all members of a Scala class with @JSExportAll
@JSExportAll
case class CaseClass(x: Int, y: Int)

// Export a Scala class or object
@JSExport
class Foo
@JSExport
object Bar

// Non-top-level export of a JS class or object
@JSExport
@ScalaJSDefined
class Foo extends js.Object
@JSExport
@ScalaJSDefined
object Bar extends js.Object

// Export all descendants of a type (Scala or JS)
@JSExportDescendantClasses
trait SomeScalaTrait1
@JSExportDescendantObjects
trait SomeScalaTrait2

@JSExportDescendantClasses
@ScalaJSDefined
trait SomeJSTrait1 extends js.Object
@JSExportDescendantObjects
@ScalaJSDefined
trait SomeJSTrait2 extends js.Object
```

## What does this proposal *not* concern

```scala
// Export a JS class or object at the top-level
@JSExportTopLevel("Foo")
@ScalaJSDefined
class Foo extends js.Object

@JSExportTopLevel("Bar")
@ScalaJSDefined
object Bar extends js.Object

// Export object members at the top-level
object Container {
  @JSExportTopLevel("foo")
  def foo(): Int = 42
}

// Export object members as static in the companion class
@ScalaJSDefined
class Foo extends js.Object

object Foo {
  @JSExportStatic
  def foo(): Int = 42
}
```
[...]

PR #2743, opened 2017-01-18
Fix #2734: Allow @JSExportTopLevel on classes and objects.

### LINKED ISSUE #2734: Proposal: Allow @JSExportTopLevel on JS classes and objects
In the let's-rething-`@JSExport` series, here is another one: allow `@JSExportTopLevel` on top-level JS classes and objects. For classes, that would be the same as `@JSExport`. On objects, they would be exported as *variables* rather than as a 0-arg function.

In other words: this becomes valid:
```scala
@JSExportTopLevel("Foo")
@ScalaJSDefined
class Foo(val x: Int) extends js.Object

@JSExportTopLevel("Bar")
@ScalaJSDefined
object Bar extends js.Object {
  def baz(x: Int): Int = x + 1
}
```
and these can be used from JS as:
```js
var foo = new Foo(5);
console.log(foo.x);

console.log(Bar.baz(5)); // note `Bar` and not `Bar()`
```
Note that, given that we now export fields to the top-level, this does not add any expressive power to Scala.js (only convenience), as we can already do:
```scala
object ExportsFromAboveAsFields {
  @JSExportTopLevel("Foo")
  val foo = js.constructorOf[Foo]

  @JSExportTopLevel("Bar")
  val bar = Bar
}
```
At the same time, we deprecate `@JSExport` on classes and objects. For JS classes/objects, it should be replaced by `@JSExportTopLevel`. For Scala classes/objects, it should become basically useless because of #2720.

Together with #2720 and #2733, *all* uses of `@JSExport` will be deprecated. `@JSExportTopLevel` replaces `@JSExport` for all cases where it provides interoperability semantics for the `export` directives of ES 2015 modules. When `@JSExport` is definitely dead, we can re-alias the name `@JSExport` to the semantics of `@JSExportTopLevel`, which are closer to the JS concept of `export`.

sjrd:
This commit implements the Proposal #2734. The annotation `@JSExportTopLevel` can now be used on static classes and objects (both Scala classes/objects and Scala.js-defined JS classes/objects) to export them at the top-level.

For objects, the difference wrt. `@JSExport` is that the module instance itself is exported, rather than the module accessor.

For classes, this is equivalent to exporting them via `@JSExport`. We include the possibility to be consistent with objects, and with the prospect of deprecating and removing `@JSExport` on classes and objects altogether.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
@@ -809,7 +809,25 @@ object Trees {
     val tpe = NoType
   }
 
+  /** Traditional `@JSExport` for top-level objects, as a 0-arg function.
+   *
+   *  This exports a module as a 0-arg function that returns the module
+   *  instance. It is initialized lazily in that case.
+   *
+   *  This alternative should eventually disappear.
+   */
   case class ModuleExportDef(fullName: String)(
+      implicit val pos: Position) extends Tree {
+    val tpe = NoType
+  }
+
+  /** New-style `@JSExportTopLevel` for top-level objects, directly as the
+   *  object.
+   *
+   *  This exports a module directly as a variable holding the module instance.
+   *  The instance is initialized during ES module instantiation.
+   */
+  case class TopLevelModuleExportDef(fullName: String)(
       implicit val pos: Position) extends Tree {
     val tpe = NoType
   }
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
@@ -926,6 +926,8 @@ private[emitter] final class ScalaJSClassEmitter(semantics: Semantics,
         genJSClassExportDef(tree, e)
       case e: ModuleExportDef =>
         genModuleExportDef(tree, e)
+      case e: TopLevelModuleExportDef =>
+        genTopLevelModuleExportDef(tree, e)
       case e: TopLevelMethodExportDef =>
         genTopLevelMethodExportDef(tree, e)
       case e: TopLevelFieldExportDef =>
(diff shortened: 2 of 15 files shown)


sjrd:
Review by @gzm0

sjrd:
Since #2720 was rejected, this should be updated to also allow `@JSExportTopLevel` on Scala classes and objects.

sjrd:
Updated.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -814,6 +814,11 @@ object Trees {
     val tpe = NoType
   }
 
+  case class TopLevelModuleExportDef(fullName: String)(
+      implicit val pos: Position) extends Tree {
+    val tpe = NoType
+  }
+
```
gzm0: Have you considered giving the `ModuleExportDef` a `lazy` flag instead? IIUC, we eventually want to get rid of lazy module exports (i.e. module exports being called by function) and only allow those (or export an accessor function). We might make this more explicit by giving `ModuleExportDef` a `lazy` flag and hopefully making the parallel to constructor export def more clear.

In any case, please document the difference between these, since as it stands, it is impossible to understand the difference between them when only looking at `Trees.scala`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ScalaJSClassEmitter.scala
```diff
@@ -988,6 +990,16 @@ private[emitter] final class ScalaJSClassEmitter(semantics: Semantics,
     genClassOrModuleExportDef(cd, tree.fullName, baseAccessor)
   }
 
+  def genTopLevelModuleExportDef(cd: LinkedClass,
```
gzm0: Same here about the "lazyness". It is not clear from the names what should be different between `genTopLevelModuleExportDef` and `genModuleExportDef`.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
sjrd: I actually considered that, and took the exact opposite decision precisely because I anticipate the lazy variant to disappear. If it was going to stay, I would have taken the effort to merge the IR nodes. But given that one is going to disappear, it was much easier to create a separate IR node.

I will document the differences.

sjrd:
Updated with a few more details in Scaladoc.

sjrd:
retest this please

sjrd:

gzm0:

sjrd:
I think we have a deterministic JVM OOM on the bootstrap test + Node.js 6 that was installed on CI breaks other stuff.

The OOM seems to be simply that the test suite passed a critical size that makes optimizing it and the tools in fullOpt require too much memory. We should try giving more memory to that process.

The Node.js 6 stuff I'll investigate next week, when I'm back from vacation.

sjrd:
Testing with 4G given to sbt on the CI config.

Issue #2741, opened 2017-01-17
The varargs optimization does not work with multiple parameter lists


chandu0101:
### Multi ParamList

```scala
@inline
def View(props:js.Object)(children:ReactNode*):ReactElement = React.createElement("View",props,children :_*)

val v = View(js.Dynamic.literal(x= 10))("hello","hello1")
```

***fastOptJS***

```scala
  var props = {
    "x": 10
  };
  var children = $m_s_Predef$().wrapRefArray__AO__scm_WrappedArray($makeNativeArrayWrapper($d_sjs_js_$bar.getArrayOf(), ["hello", "hello1"]));
  var jsx$1 = $i_react;
  var jsx$4 = jsx$1.createElement;
  if ($is_sjs_js_ArrayOps(children)) {
    var x2 = $as_sjs_js_ArrayOps(children);
    var jsx$3 = x2.scala$scalajs$js$ArrayOps$$array$f
  } else if ($is_sjs_js_WrappedArray(children)) {
    var x3 = $as_sjs_js_WrappedArray(children);
    var jsx$3 = x3.array$6
  } else {
    var result = [];
    var i = 0;
    var len = children.length__I();
    while ((i < len)) {
      var arg1 = children.apply__I__O(i);
      $uI(result.push(arg1));
      i = ((1 + i) | 0)
    };
    var jsx$3 = result
  };
  var jsx$2 = ["View", props].concat(jsx$3);
  var v = jsx$4.apply(jsx$1, jsx$2);
```
***fullOptJS***  

```scala
var a;cd||(cd=(new dd).a());var b=new (Fa(Wa).ra)(["hello","hello1"]);
null===b?b=null:0===b.R.length?(Nc||(Nc=(new Mc).a()),b=Nc.Ha):(a=new H,a.T=b,b=a);a=b;b=Oa.createElement;
if(a&&a.b&&a.b.j.ee)a=a.Nd;else if(a&&a.b&&a.b.j.ge)a=a.T;
else{for(var c=[],e=0,l=a.ja();e<l;){var k=a.S(e);c.push(k)|0;e=1+e|0}a=c}
a=["View",{x:10}].concat(a);b=b.apply(Oa,a);
```

### Single Param List 

```scala
@inline
def View(props:js.Object,children:ReactNode*):ReactElement = React.createElement("View",props,children :_*)
val v = View(js.Dynamic.literal(x= 10),"hello","hello1")
```

***fastOptJS*** : 

```scala
 var props = {
[...]

PR #2745, opened 2017-01-21
Fix #2265 and #2741: Look for repeated args in all parameter lists.

### LINKED ISSUE #2265: The varargs optimization of the compiler does not work for AnyVal methods
``` scala
object MyJSArrayOps {
  implicit class MyJSArray(val self: js.Array.type) extends AnyVal {
    @inline
    def make[A](elems: A*): js.Array[A] = elems.toJSArray
  }
}

  def main(): Unit = {
    import MyJSArrayOps._

    val a = js.Array.make("hello", 7, 12)
  }
}
```

gives

``` scala
  def main__V() {
    val a: any = mod:Lhelloworld_MyJSArrayOps$MyJSArray$.make$extension__sjs_js_Array$__sc_Seq__sjs_js_Array(mod:Lhellow
orld_MyJSArrayOps$.MyJSArray__sjs_js_Array$__sjs_js_Array$(<envinfo>["global"]["Array"]), mod:s_Predef$.genericWrapArray
__O__scm_WrappedArray(O[]("hello", 7, 12)));
    ...
  }
```

but if it is not an `AnyVal`, it gives, as expected:

``` scala
  def main__V() {
    val a: any = mod:Lhelloworld_MyJSArrayOps$.MyJSArray__sjs_js_Array$__Lhelloworld_MyJSArrayOps$MyJSArray(<envinfo>["g
lobal"]["Array"]).make__sc_Seq__sjs_js_Array(new sjs_js_WrappedArray().init___sjs_js_Array(["hello", 7, 12]));
    ...
```

sjrd:
Usually, in the compiler back-end, we use `methodTpe.params` to read the list of all parameters. This works because multiple parameter lists are flattened. The code that optimizes varargs as `new js.WrappedArray(js.Array(...))` was reading parameters right after `typer` in the type history. It has to use `.paramss.flatten` in that case, because `.params` represents only the first parameter list.

### DIFF
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
@@ -10,7 +10,6 @@ class OptimizationTest extends JSASTTest {
 
   @Test
   def unwrapScalaFunWrapper: Unit = {
-
     // Make sure we do not wrap and unwrap right away
     """
     import scala.scalajs.js
@@ -35,7 +34,10 @@ class OptimizationTest extends JSASTTest {
     has("runtime.AnonFunction ctor") {
       case js.New(jstpe.ClassType("sjsr_AnonFunction1"), _, _) =>
     }
+  }
 
+  @Test
+  def testJSArrayApplyOptimization: Unit = {
     /* Make sure js.Array(...) is optimized away completely for several kinds
      * of data types.
      */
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0 
/cc @chandu0101

chandu0101:
@sjrd  thanks alot :) , i just tested in my local working perfect 👍

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
```diff
@@ -56,21 +58,58 @@ class OptimizationTest extends JSASTTest {
...
       val e = List(new VC(151189))
+
+      def single(x: Int, ys: Int*): Int = x + ys.size
+      def multiple(x: Int)(ys: Int*): Int = x + ys.size
+    }
+    """.
+    hasNot("any of the wrapArray methods") {
+      case js.Apply(_, js.Ident(name, _), _)
+          if name.startsWith("wrap") && name.endsWith("__scm_WrappedArray") =>
+    }
```
gzm0: It seems like this is duplicate

sjrd: Oops, indeed.

sjrd:
Updated.

PR #2747, opened 2017-01-21
Fix #2532: Emit "private" field access for Java-defined fields.

### LINKED ISSUE #2532: Cannot declare Java fields in JDK classes
If a class from the JDK has a public or protected _field_, we have no way in our implementation to declare it. No matter what, the field will stay private at the IR level, and attempts to access the field from user code will not link.

### LINKED ISSUE #2737: Failure when accessing protected buf in ByteArrayOutputStream
I've got some Scala code that extends `ByteArrayOutputStream` to add access to the buffer without copying. In my real code I'm returning a `ByteBuffer`, but this simplified version shows the issue:

```scala
import java.io.ByteArrayOutputStream
import org.scalatest.FlatSpec
import org.scalatest.prop.GeneratorDrivenPropertyChecks

class EnhancedByteArrayOutputStream extends ByteArrayOutputStream {
  def getBuf(): Array[Byte] = buf
}

class EnhancedByteArrayOutputStreamSpec extends FlatSpec
  with GeneratorDrivenPropertyChecks {
  "getBuf" should "provide access to the buffer" in forAll { b: Byte =>
    val bytes = new EnhancedByteArrayOutputStream()
    bytes.write(b.toInt)
    bytes.close()

    assert(bytes.getBuf()(0) === b)
  }
}
```

This works fine on the JVM, but fails on Scala.js:

```
[info] EnhancedByteArrayOutputStreamSpec:
[info] getBuf
[info] - should provide access to the buffer *** FAILED ***
[info]   JavaScriptException was thrown during property evaluation.
[info]     Message: TypeError: Cannot read property 'u' of undefined
[info]     Occurred when passed generated values (
[info]       arg0 = 0 // 1 shrink
[info]     )
```

This is easy enough to work around with a Scala.js-specific implementation that uses `toByteArray` instead of `buf`, and I haven't investigated why it's happening, but I thought it might be worth reporting.

sjrd:
Java-defined fields are now always accessed as if they were private. This is necessary because they are defined as private by our .scala source files, but they are considered `!isPrivate` at use site, since their symbols come from Java-emitted .class files. Fortunately, we can easily detect those as `isJavaDefined`.

This includes fields of Ref types (IntRef, ObjectRef, etc.) which were previously special-cased at use-site. This commit generalizes the idea to all fields that are Java-defined (as per the .class files).

This also fixes #2737, which was directly caused by #2532.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
@@ -80,12 +80,6 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
     js.Ident(localSymbolName(sym), Some(sym.unexpandedName.decoded))
   }
 
-  private lazy val allRefClasses: Set[Symbol] = {
-    import definitions._
-    (Set(ObjectRefClass, VolatileObjectRefClass) ++
-        refClass.values ++ volatileRefClass.values)
-  }
-
   /** See comment in `encodeFieldSym()`. */
   private lazy val shouldMangleOuterPointerName = {
     val v = scala.util.Properties.versionNumberString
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0.

I manually tested that the test suite compiled with the new scheme correctly links against the standard library compiled with the old scheme, and tests pass in that situation.

sjrd:
Yes, it does make that an incompatible change. But then again, before nothing would work anyway.

I have seen classes getting new interfaces in different JDK versions, but I do not recall any class gaining a new parent class anywhere.

compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
```diff
@@ -101,10 +101,13 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
...
-     * because they are emitted as private by our .scala source files, but
-     * they are considered public at use site since their symbols come from
-     * Java-emitted .class files.
+    /* Java-defined fields are always accessed as if they were private. This
+     * is necessary because they are defined as private by our .scala source
+     * files, but they are considered `!isPrivate` at use site, since their
+     * symbols come from Java-emitted .class files. Fortunately, we can
+     * easily detect those as `isJavaDefined`. This includes fields of Ref
+     * types (IntRef, ObjectRef, etc.) which were special-cased at use-site
+     * in Scala.js < 0.6.15.
```
gzm0: Please add a comment here that this makes changing the length of the superclass chain of a .java defined Scala class a binary incompatible change.

sjrd:
Updated.

PR #2753, opened 2017-01-25
Fix #2752: Survive in `test` if there is no testing framework.

### LINKED ISSUE #2752: `test` task crashes if there is no testing framework (instead of succeeding with 0 test)
Discovered by the community build, see https://github.com/scala/community-builds/issues/465
Reproducible with `library/test`.

> The thing is: `library/test` is empty; it doesn't have any test. Most importantly, it doesn't have any testing *framework*, and that means it doesn't depend on our testing interface. The changes in https://github.com/scala-js/scala-js/pull/2740 cause the `test` task to crash if the testing interface is not in scope, whereas before it would silently (and correctly) return an empty list of testing frameworks, and hence execute 0 test.

sjrd:
The `test` task should trivially succeed if there is no testing framework on the classpath. However, f46050f245bed6c318e2f2855a811ea6afb5d0cd broke this scenario, by requiring an export from the testing interface to be present.

This commit fixes the scenario to tolerate the absence of that export. In that case, no candidate testing framework is reported to be available.

We add `library/test` to the CI matrix as a test for this scenario, as the `library` project does not and will not contain any tests.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -57,6 +57,7 @@
     sbtretry 'set inScope(ThisScope in testingExample)(jsEnv := new org.scalajs.jsenv.RetryingComJSEnv(PhantomJSEnv().value))' \
         'set scalaJSOptimizerOptions in testingExample ~= (_.withDisableOptimizer(true))' \
         ++$scala testingExample/test:run testingExample/test &&
+    sbtretry ++$scala library/test &&
     sbtretry ++$scala testSuiteJVM/test testSuiteJVM/clean &&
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @SethTisue, especially if you could run a community build against that PR ;)

SethTisue:
tested locally with

```
-  scala-js-uri:                 "https://github.com/scala-js/scala-js.git"
+  scala-js-uri:                 "https://github.com/sjrd/scala-js.git#fix-test-when-no-testingframework"
```

and `version=2.12.1 ./run.sh scala-js`

yielding `[info] Project scala-js----------------: SUCCESS (project rebuilt ok)`

👍

sjrd:
Thanks.

PR #2765, opened 2017-02-06
Fix #2758: Add `UndefOr.contains` to mimic `Option.contains`.

### LINKED ISSUE #2758: `UndefOr` could have `contains` implemented like `Option`
Currently `UndefOr` has many members similar to `Option`, but I do not see `contains`. The workaroud is easy, either I can use `.toOption.contains(x)` or `.exists(_ == x)`, however as direct implementation should be straightforward, I think it could be added (unless there is some reason not to do so).

sjrd:
`Option.contains` was added in Scala 2.11.0. This commit adds an equivalent method to `UndefOr`.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/UndefOr.scala
+++ b/library/src/main/scala/scala/scalajs/js/UndefOr.scala
@@ -158,6 +158,18 @@ final class UndefOrOps[A](val self: UndefOr[A]) extends AnyVal {
   @inline final def withFilter(p: A => Boolean): WithFilter[A] =
     new WithFilter(self, p)
 
+  /** Tests whether the $option contains a given value as an element.
+   *
+   *  `x.contains(y)` differs from `x == y` only in the fact that it will
+   *  return `false` when `x` and `y` are both `undefined`.
+   *
+   *  @param elem the element to test.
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

library/src/main/scala/scala/scalajs/js/UndefOr.scala
```diff
@@ -158,6 +158,19 @@ final class UndefOrOps[A](val self: UndefOr[A]) extends AnyVal {
   @inline final def withFilter(p: A => Boolean): WithFilter[A] =
     new WithFilter(self, p)
 
+  /** Tests whether the $option contains a given value as an element.
+   *
+   *  `x.contains(y)` differs from `x == y` only in the fact that it will
+   *  return `false` when `x` and `y` are both `undefined` (which is also
+   *  `()`).
```
gzm0: I'm not sure the comment in parens actually adds something here. If the comment is needed, it should probably be on the class level. Here it is confusing, as it raises the question: "Why is () special in `contains`?"

sjrd: Yes, you're right. I have removed the part in the parens.

PR #2766, opened 2017-02-06
Fix #2755: Fix array out of bounds in BigDecimal.divide(scale).

### LINKED ISSUE #2755: java.math.BigDecimal.divide fails when used with a scale and rounding mode
Running this code:

```scala
import java.math.{BigDecimal, RoundingMode}

val a = new BigDecimal(2L)
val b = new BigDecimal(1L)

a.divide(b, 1, RoundingMode.UNNECESSARY)
```

results in this exception:

```
target/scala-2.12/bdbug-fastopt.js:5755
    var lo$2 = t.lo$2;
                ^
TypeError: Cannot read property 'lo$2' of undefined
    at $c_Ljava_math_BigDecimal.divide__Ljava_math_BigDecimal__I__Ljava_math_RoundingMode__Ljava_math_BigDecimal (target/scala-2.12/bdbug-fastopt.js:5755:17)
    at $c_LBug$.main__V (target/scala-2.12/bdbug-fastopt.js:3712:5)
    at $c_LBug$.$$js$exported$meth$main__O (target/scala-2.12/bdbug-fastopt.js:3715:8)
    at $c_LBug$.main (target/scala-2.12/bdbug-fastopt.js:3718:15)
    at [stdin]:40:65
    at ContextifyScript.Script.runInThisContext (vm.js:26:33)
    at Object.exports.runInThisContext (vm.js:79:17)
    at Object.<anonymous> ([stdin]-wrapper:6:22)
    at Module._compile (module.js:571:32)
    at Immediate.<anonymous> (bootstrap_node.js:383:29)
```

Looks like the reason may be that `diffScale` on [line 707](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/math/BigDecimal.scala#L707) gets a negative value and causes a lookup with negative index [into `LongTenPows`](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/math/BigDecimal.scala#L732).

sjrd:
There were computations based on `LongTenPows` that would access indices before testing whether they are valid or not. The fix is to delay those accesses after we've made sure that indices are valid.

In the process, I have rearranged the computation of `diffScale` to convert it to an `Int` earlier, avoid some useless `Long` operations.

### DIFF
--- a/javalib/src/main/scala/java/math/BigDecimal.scala
+++ b/javalib/src/main/scala/java/math/BigDecimal.scala
@@ -704,52 +704,56 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
     else if (divisor.isZero)
       throw new ArithmeticException("Division by zero")
 
-    val diffScale: Long = (this._scale.toLong - divisor._scale) - scale
+    val diffScale = {
+      val diffScaleLong = (this._scale.toLong - divisor._scale) - scale
+      val diffScale = diffScaleLong.toInt
 
-    // Check whether the diffScale will fit into an int.
-    if (bitLength(diffScale) > 32) {
-      val msg = s"Unable to scale as difference is too big ($diffScale)"
-      throw new ArithmeticException(msg)
+      // Check whether the diffScale will fit into an Int
+      if (diffScale.toLong != diffScaleLong) {
+        throw new ArithmeticException(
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

PR #2767, opened 2017-02-06
Opt: Virtualize JSArrayConstr's in the optimizer.

sjrd:
This commit introduces a new kind of virtualized value (a new `LocalDefReplacement`) for JavaScript arrays created as a `JSArrayConstr` IR node. They can stay virtualized as long as the only operations performed on them are:

* Reading the `length` property (replaced by a constant)
* Reading an indexed property (replaced by the `LocalDef` at the appropriate index in the virtualized array, or `undefined` if the index is out of bounds)
* Using the array in a `JSSpread` (the spread is expanded and replaced by a fixed number of arguments)

This is particularly effective in two scenarios:

* When writing Scala wrappers taking varargs that forward them as to a JS method taking varargs: if inlined, at call site the optimizer can expand the arguments as fixed arguments to the JS method.
* When manipulating js.Tuples, which now have the same stack allocation capabilities as Scala tuples.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -287,16 +287,6 @@ private[optimizer] abstract class OptimizerCore(
   private def transformExpr(tree: Tree)(implicit scope: Scope): Tree =
     transform(tree, isStat = false)
 
-  /** Transforms an expression or a JSSpread. */
-  private def transformExprOrSpread(tree: Tree)(implicit scope: Scope): Tree = {
-    tree match {
-      case JSSpread(items) =>
-        JSSpread(transformExpr(items))(tree.pos)
-      case _ =>
-        transformExpr(tree)
-    }
-  }
-
   /** Transforms a tree. */
   private def transform(tree: Tree, isStat: Boolean)(
       implicit scope: Scope): Tree = {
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -840,10 +834,44 @@ private[optimizer] abstract class OptimizerCore(
...
+      case JSArrayConstr(items) =>
+        if (items.exists(_.isInstanceOf[JSSpread])) {
+          /* TODO This means spread in array constr does not compose under
+           * this optimization. We could improve this with a
+           * pretransformExprsOrSpreads() or something like that.
+           */
+          cont(JSArrayConstr(transformExprsOrSpreads(items)).toPreTransform)
+        } else {
+          pretransformExprs(items) { titems =>
+            tryOrRollback { cancelFun =>
```
gzm0: Do we have any empirical data about how often this will happen in a codebase closer to JS? I'm a bit concerned that this causes too many rollbacks.

sjrd: In a codebase "closer to JS", we wouldn't get here in `pretransformExpr(JSArrayConstr)`. The `JSArrayConstr` would probably be a direct argument to a JS method call, in which case it would be transformed by `transformExpr(JSArrayConstr)`, which does not try to optimize it with a `tryOrRollback`.

It would still happen if the developer stores the `js.Array()` in a `val` first, though.
```scala
val a = js.Array(1, 2)
jsObj.foo(a)
```
would have a rollback, indeed. But
```scala
jsObj.foo(js.Array(1, 2))
```
wouldn't.

sjrd: That said, no, I do not have any empirical data.

gzm0: Meh, OK. Let's see what happens :P

sjrd:
Updated.

PR #2768, opened 2017-02-06
Make the union type understand variance for 1-arg type constructors.

sjrd:
The subtyping rules of the union type, materialized by implicit `Evidence`s, are enhanced with support for variance in type constructors of 1 argument.

Concretely, this means that `List[Int]` can now be assigned to a `List[Int | String]`, and that a `Consumer[Int | String]` can be assigned to a `Consumer[Int]` (for some `Consumer[-A]`).

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Union.scala
+++ b/library/src/main/scala/scala/scalajs/js/Union.scala
@@ -10,6 +10,7 @@
 package scala.scalajs.js
 
 import scala.language.implicitConversions
+import scala.language.higherKinds
 
 /** Value of type A or B (union type).
  *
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0

library/src/main/scala/scala/scalajs/js/Union.scala
```diff
@@ -66,6 +79,14 @@ object | { // scalastyle:ignore
   implicit def from[A, B1, B2](a: A)(implicit ev: Evidence[A, B1 | B2]): B1 | B2 =
     a.asInstanceOf[B1 | B2]
 
+  /** Upcast `F[A]` to `F[B]`.
+   *
+   *  This needs evidence that `F[A] <: F[B]`.
+   */
+  implicit def fromTypeConstructor[F[_], A, B](a: F[A])(
+      implicit ev: Evidence[F[A], F[B]]): F[B] =
```
gzm0: So just to double check: You allow conversion to *any* type in the constructor so that, for example, the following apparent subtyping relationship holds: `List[List[Int]] <: List[List[Int | String]]`.

sjrd: Yep. In fact, ideally, there would be only one entry point conversion
```scala
implicit def from[A, B](a: A)(implicit ev: Evidence[A, B]): B
```
Both the actual `from` and `fromTypeConstructor` are workaround for the "diverging implicit" limitations of scalac, which happen worth the ideal universal `from`.

PR #2769, opened 2017-02-07
Implement System.lineSeparator().

sjrd:
This is a JDK7 API. In Scala.js, it trivially returns `"\n"`.

scala/scala is deprecating `scala.compat.Platform.lineSeparator`, and encourages people to use `System.lineSeparator()` instead. So we need to support it :p See https://github.com/scala/scala/pull/5677

### DIFF
--- a/javalanglib/src/main/scala/java/lang/System.scala
+++ b/javalanglib/src/main/scala/java/lang/System.scala
@@ -245,6 +245,8 @@ object System {
   def getProperties(): ju.Properties =
     SystemProperties.value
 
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0, because we love blindly following Processes :P

gzm0:
> because we love blindly following Processes :P

HAHAHAHAHA, I didn't even see this :P

sjrd:
Ah ah! That makes it so much funnier :)

PR #2775, opened 2017-02-16
Fix #2199: Add the notion of module initializers to the linker.

### LINKED ISSUE #2199: Add first-class support for running with Node
Currently (from http://stackoverflow.com/q/34763039/463761) the way to run an application with Node requires this added to the build:

``` scala
scalaJSOutputWrapper := ("", "example.ScalaJSExample().main();")
```

However this causes `run` to run the script twice.

@sjrd Recommends this workaround:

``` scala
inConfig(Compile)(Seq(
  packageScalaJSLauncher := {
    val file = (artifactPath in packageScalaJSLauncher).value
    sbt.IO.write(file, "", Charset.forName("UTF-8"))
    Attributed.blank(file)
  }
))
```

It would be nice to have first class support here for both output files (`fastOptJS`/`fullOptJS`) and `run` (and similar..).

sjrd:
**Edit**: the original version of this PR used the term "entry point" for what is now "module initializer".

There are two main parts in this PR. The first one is focused on the linker, and adds the core mechanism of module initializers to it. The second part focuses on the sbt plugin, to automatically derive an module initializer from `mainClass`. Subsequently, we deprecate `persistLauncher` and `scalaJSOutputWrapper`.

## Add user-defined module initializers to the linker.

A module initializer is a piece of code that is executed when the module produced for a linked Scala.js application is loaded.

This commit adds an sbt task of type `Seq[ModuleInitializer]`, which allows the user to define custom module initializers to their module.
Example:
```scala
scalaJSModuleInitializers += {
  org.scalajs.core.tools.linker.ModuleInitializer.mainMethod(
      "fully.qualified.ObjectName", "main")
}
```

Module initializers can be specified `in Compile`, in which case they are not executed during tests.

This fills a gap in our interoperability criterium, as previously there was no way that user-defined code could be executed at module load time (without hacking around it by using an exported static field that is not actually used).

## Add scalaJSUseMainModuleInitializer, to use an ModuleInitializer for the mainClass.

When the setting `scalaJSUseMainModuleInitializer` is set to true, a module initializer will be derived from `mainClass` (which is itself derived from discovered `js.JSApp`s in the program). This causes a call to the main method of a main class to be built in the fastOptJS/fullOptJS output.

Consequently, `run` is adapted to know about `scalaJSUseMainModuleInitializer`, so that it doesn't execute the main method twice.

The setting is mutually exclusive with `persistLauncher`, as it does not make sense to write a launcher file when the main .js file already executes the main method.

In fact, `scalaJSUseMainModuleInitializer` is intended to supersede launcher files and `persistLauncher`.

### `persistLauncher` and `packageScalaJSLauncher`

The functionality offered by `persistLauncher` and `packageScalaJSLauncher` has been superseded by
`scalaJSUseMainModuleInitializer`. This commit deprecates settings and tasks related to the generation of a launcher file, and recommends to use `scalaJSUseMainModuleInitializer` instead.

### `scalaJSOutputWrapper`

`scalaJSOutputWrapper` was used in a variety of ways to work around limitations of Scala.js. All the use cases found in the wild now have better replacements:

* Acquire the `require` function from the lexical scope and store it in `global`: emit CommonJS modules instead.
* Set the `exportsNamespace` to `module.exports`: also emit CommonJS modules.
* Execute a main method at startup of the .js file: use module initializers instead, in particular `scalaJSUseMainModuleInitializer`.
* Export a top-level function or similar: use `@JSExportTopLevel` instead.


sjrd:
Completed. Review by @gzm0

sjrd:
It seems we have a deterministic PhantomJS crash in 2.11.8 in FullOpt. I suspect an OOM.

sjrd:
I cannot reproduce the problem locally, nor can I find any option of PhantomJS that would increase its allocated memory. I'm not sure what to do now ...

gzm0:
I'm having one high-level concern with this: It is inconsistent with how we handle module exports.

I fully agree that linking time is when you should define what and how goes into your module. However, this is currently not true for (top-level) exports: A module defines what it exports in it's code.

So I'm a bit split, because it seems like this is the most reasonable thing to do for now, but I see that it is quite inconsistent (and it's not like we do not have precedents for linking options actually changing what code does).

Another alternative would be a more C-like approach (`@JSMain` annotation or something) where just a single such a symbol may be in the linking bundle. The obvious downside of course being that this is very inconsistent with Java / Scala.

All in all, I guess I'm just asking for alternatives that you considered :)

sjrd:
Well, the only alternative I have really come up with is, as you say, an `@JSMain`-like annotation. However, this has at least two significant problems:
* In case there are several `@JSMain`-annotated methods, how do we specify ordering?
* More importantly, an annotation-based solution forces the `Test` configuration to include all the entry points available in the `Compile` configuration. But that's almost never what we want, as we only want the `main()` to be executed in the Compile config, but not when we run tests.

sjrd:
It seems reducing the size of the test suite on PhantomJS allowed it to pass, so I'm going to use that trick. Anyway in 1.0.0 PhantomJS support won't be in the same repo anymore, so it won't matter very long.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/EntryPoint.scala
```diff
@@ -0,0 +1,57 @@
...
+ *  in the specified order, after having initialized everything else (notably
+ *  static initializers).
+ *
+ *  Instances of `EntryPoint` can be created with methods of
+ *  [[EntryPoint$ the EntryPoint companion object]].
+ */
+sealed abstract class EntryPoint
+
+/** Factory for [[EntryPoint]]s. */
+object EntryPoint {
```
gzm0: I'm not 100% sure about the name `EntryPoint` TBH. It is more some sort of "loading code". It is specifically *not* a thing where one can "enter" the module. Maybe one of:
- "loading hooks"
- "module initializers" (attention: evil term overloading)
- "setup point"

None of these are really satisfying. TBH I secretly hope that `EntryPoint` is ES terminology and hence a no-brainer :P

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -200,6 +200,15 @@ object ScalaJSPlugin extends AutoPlugin {
...
         AMinusTask)
 
+    val scalaJSUseMainEntryPoint = SettingKey[Boolean]("scalaJSUseMainEntryPoint",
+        "If true, adds the `mainClass` as an entry point of the Scala.js module",
+        APlusSetting)
+
+    val scalaJSMainEntryPoint = TaskKey[Option[EntryPoint]](
+        "scalaJSMainEntryPoint",
+        "The main entry point, derived from `mainClass` if `scalaJSUseMainEntryPoint` is true",
+        BPlusTask)
```
gzm0: When would somebody set this task (rather than `mainClass in scalaJS<something>`)?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -742,12 +750,40 @@ object ScalaJSPluginInternal {
...
       mainClass in scalaJSLauncher := (mainClass in run).value,
       scalaJSLauncher := Def.taskDyn[Attributed[VirtualJSFile]] {
         if (persistLauncher.value) {
           Def.task {
             packageScalaJSLauncher.value.map(FileVirtualJSFile)
           }
+        } else if (scalaJSUseMainEntryPoint.value) {
+          Def.task {
+            val base = Attributed.blank[VirtualJSFile](
+                new MemVirtualJSFile("Generated launcher file"))
```
gzm0: Maybe make it explicit in the name of the memfile that it is intentionally a noop.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/EntryPoint.scala
sjrd: It is not ES terminology, no. In ES that's just code at the top-level of the module. It doesn't make sense from a Scala point of view.

I think "module initializer" is the best one. It's only overloaded internally. Static initializers are not available to user code, and objects being known as "modules" is also pretty internal.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: If their main class is not a `js.JSApp`, or in general if their `EntryPoint` doesn't use `"main"` as the main method name.

cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
```diff
@@ -68,6 +79,11 @@ object Scalajsld {
         .unbounded()
         .action { (x, c) => c.copy(cp = c.cp :+ x) }
         .text("Entries of Scala.js classpath to link")
+      opt[EntryPoint]('e', "execute")
+        .valueName("<full.name.Object.main>")
+        .unbounded()
+        .action { (x, c) => c.copy(entryPoints = c.entryPoints :+ x) }
+        .text("Execute the specified main method on startup")
```
sjrd: `-m` was taken unfortunately.

I get your point, though. I suggest `--main-method` with an `.abbr("mm")`, then.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkingUnit.scala
```diff
@@ -11,8 +11,8 @@ final class LinkingUnit private[linker] (
     val esLevel: ESLevel,
     val classDefs: List[LinkedClass],
     private[linker] val infosInternal: Map[String, Infos.ClassInfo],
-    val isComplete: Boolean,
-    dummyInternal: Unit
+    val entryPoints: List[EntryPoint],
```
sjrd: Currently you get a linking error saying:
```
[error] Referring to non-existent class org.scalajs.testsuite.compiler.EntryPointInNoConfigurationFoo$
[error]   called from core module entry points
[error] Referring to non-existent method org.scalajs.testsuite.compiler.EntryPointInNoConfigurationFoo$.main()scala.Unit
[error]   called from core module entry points
```
IMO this is good enough, don't you think?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -742,12 +750,40 @@ object ScalaJSPluginInternal {
...
+                "was set to true. If you do not want to derive a main entry " +
+                "point from discovered main classes, add " +
+                "`scalaJSMainEntryPoint := None`")
+          } { mainClass =>
+            Some(EntryPoint.mainMethod(mainClass, "main"))
+          }
+        } else {
+          None
+        }
+      },
```
gzm0: What happens if we make this non-optional, but do not pull it in unless `scalaJSUseMainEntryPoint` is true?

Currently, we have the implicit assumption that `scalaJSUseMainEntryPoint == scalaJSMainEntryPoint.isDefined`. It would be nicer if we could get rid of the unneeded degree of freedom. In the worst case, we should just assert that.

cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
gzm0: Ok, main method is fine with me. But can we make it `mainMethod` to be consistent with the other names?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkingUnit.scala
gzm0: Yes, of course it is. Forgot about that one.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: Let's just remove it altogether, and inline its value in `scalaJSEntryPoints ++=` just below.

cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
sjrd: Ah yes, sure.

sjrd:
Updated.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: So in fact I kept it, because it can be useful if the developer has a non-`JSApp` module initializer. They want to `run` their app (so `scalaJSUseMainModuleInitializer` is true) but they do not want to have an automatically derived module initializer (so `scalaJSMainModuleInitializer` is customized).

Moreover, we don't have an assumption `scalaJSUseMainModuleInitializer == scalaJSMainModuleInitializer.isDefined`. It's perfectly OK to have `None` when `scalaJSUseMainModuleInitializer` is true.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkingUnit.scala
```diff
@@ -6,17 +6,46 @@ import org.scalajs.core.tools.javascript.ESLevel
...
 import ir.{Definitions, Infos}
 
-final class LinkingUnit(
+final class LinkingUnit private[linker] (
     val semantics: Semantics,
     val esLevel: ESLevel,
     val classDefs: List[LinkedClass],
-    val infos: Map[String, Infos.ClassInfo],
-    val isComplete: Boolean
+    private[linker] val infosInternal: Map[String, Infos.ClassInfo],
```
sjrd: Yes. It's going to fall naturally when removing the deprecated fields.

sjrd:
Rebased. The command-line git was able to do an automatic merge, despite the renaming `ScalaJSClassEmitter` -> `ClassEmitter`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/GenLinker.scala
```diff
@@ -18,15 +18,29 @@ import org.scalajs.core.tools.linker.analyzer.SymbolRequirement
...
 /** Common supertrait of [[Linker]] and [[ClearableLinker]].
  *
- *  Essentially anything that has the [[link]] and [[linkUnit]] methods.
+ *  Essentially anything that has the `link` and `linkUnit` methods.
  */
 trait GenLinker {
   def semantics: Semantics
   def esLevel: ESLevel
 
+  @deprecated("Use the overload with explicit entry points.", "0.6.15")
```
gzm0: module initializers

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -742,12 +750,41 @@ object ScalaJSPluginInternal {
 
   // These settings will be filtered by the stage dummy tasks
   val scalaJSRunSettings = Seq(
+      scalaJSMainModuleInitializer := {
+        if (scalaJSUseMainModuleInitializer.value) {
+          mainClass.value.fold {
```
gzm0: `mainClass in scalaJSMainModuleInitializer`?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -742,12 +750,41 @@ object ScalaJSPluginInternal {
...
+                "scalaJSUseMainModuleInitializer was set to true. If you do " +
+                "not want to derive a main module initializer from " +
+                "discovered main classes, add " +
+                "`scalaJSMainModuleInitializer := None`")
+          } { mainClass =>
+            Some(ModuleInitializer.mainMethod(mainClass, "main"))
+          }
+        } else {
+          None
+        }
```
gzm0: Now I know what I found weird here :) This should detect the main initializer regardless of `scalaJSUseMainModuleInitializer`. It can then become non-optional.

This makes much more sense IMO. The key returns the main module initializer. Regardless of whether we want to use it or not (just like `mainClass` returns the main class, even if you `runMain` or not use it at all).

Maybe to better align with sbt, we should map on `mainClass` instead and put the error message in the updated `scalaJSEntryPoints ++=`?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -742,12 +750,41 @@ object ScalaJSPluginInternal {
...
+      },
+
+      scalaJSModuleInitializers ++= scalaJSMainModuleInitializer.value.toList,
```
gzm0: This instead should check whether we want to *use* the main module initializer and conditionally add it.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkingUnit.scala
```diff
@@ -11,8 +11,8 @@ final class LinkingUnit private[linker] (
     val esLevel: ESLevel,
     val classDefs: List[LinkedClass],
     private[linker] val infosInternal: Map[String, Infos.ClassInfo],
-    val isComplete: Boolean,
-    dummyInternal: Unit
+    val moduleInitializers: List[ModuleInitializer],
```
gzm0: I'm ok with this right now, but it does not make a lot of sense: Why would we want to expose a read-only data structure?

We probably need a full API visibility rehaul before 1.0.0. Probably even a way to separate the linker API into "usage public" and "plugin public". The fields of LinkingUnit for example would only be plugin public, and have lower compatibility guarantees. Unclear though, how we apply SemVer to that (#2812).

project/Build.scala
```diff
@@ -1499,6 +1499,37 @@ object Build {
...
+         */
+        sources in Test := {
+          def isPhantomJS(env: JSEnv): Boolean = env match {
+            case _: PhantomJSEnv       => true
+            case env: RetryingComJSEnv => isPhantomJS(env.baseEnv)
+            case _                     => false
+          }
+
+          val sourceFiles = (sources in Test).value
+          if ((jsEnv in Test).?.value.exists(isPhantomJS)) {
```
sjrd: I don't think I understand what you mean by "in JVM land", now. From your previous comment I had added the above sentence in the comments:

> Note that `jsEnv` is never set from this Build, but it is set via the command-line in the CI matrix.

gzm0: Ah, sorry. It doesn't explain why you need to query `jsEnv.?`. I mean that if we run the tests in Scala JVM, `jsEnv` is not set. That is not necessarily clear in this JS focused build.

sjrd: Ah, that's what you meant! Well we're in `testSuite.settings(...)`. These settings are not even applied to `testSuiteJVM` ;)

The reason we need `jsEnv.?` is like everywhere else: because `jsEnv` doesn't have a default value in the sbt plugin. `resolvedJSEnv` also reads `jsEnv.?`. IIRC that was part of some clever tricks we did to mostly maintain compatibility when we removed `preLinkJSEnv` and `postLinkJSEnv`. I think in 1.0 we won't need that hack anymore.

gzm0: Yeah, I just realized that. So I guess it's fine as is :P

partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
```diff
@@ -75,10 +75,11 @@ class MainGenericRunner {
 
     val sjsCode = {
       val output = WritableMemVirtualJSFile("partest.js")
-      linker.link(ir, output, logger)
+      linker.link(ir, moduleInitializers = Nil, output, logger)
       ResolvedJSDependency.minimal(output) :: Nil
     }
 
+    // TODO We should probably replace this by an ModuleInitializer, eventually.
```
sjrd: It wasn't harder. I think at the time I didn't want to mess with the partest runner. I'll change it.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: No, I don't think so. It seems weird to me that one would want a different `mainClass` to compute `scalaJSMainModuleInitializer` versus the names displayed by `Running helloworld.Main` and other things like that.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: I almost agreed. But then I realized that your proposal does not support the use case where someone has a custom, non `mainClass`-based module initializer (e.g., their main *method* is not called `main`), but still wants to `run` (hence requires `scalaJSUseMainModuleInitializer`).

sjrd:
Updated. I should have addressed all comments, except those related to `scalaJSMainModuleInitializer` (see my answer about that).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Huh? I meant this:

```scala
scalaJSMainModuleInitializer := mainClass.value.fold {
  throw new MessageOnlyException(...)
} { mainClass => ModuleInitializer.mainMethod(mainClass, "main")

scalaJSModuleInitializers ++= Task.Dyn {
  if (scalaJSUseMainModuleInitializer.value)
    List(scalaJSMainModuleInitializer.value)
  else
    Nil
}
```

Now if you want to run something else, you can just do:

```scala
scalaJSMainModuleInitializer := /* my awesome initializer */
scalaJSUseMainModuleInitializer := true
```

Arguably the term "Main" is a bit overloaded in this case, but it still seems cleaner than:

```scala
scalaJSMainModuleInitializer := None
scalaJSUseMainModuleInitializer := true
scalaJSModuleInitializers += /* my awesome initializer */
```

One thing that particularly bothers me in the current state is that the value of `scalaJSMainModuleInitializer` (which I interpret as "the main initializer of this project") depends on `scalaJSUseMainModuleInitializer` (which I interpret as "should we use the main initializer of this project").

Maybe it's more the naming that is confusing?

sjrd: Ah yes you're right. I'll change it for your scheme.

sjrd: Done

> Maybe to better align with sbt, we should map on mainClass instead and put the error message in the updated `scalaJSEntryPoints ++=`?

That's what I ended up doing, yes. In addition to being more aligned with sbt, it avoids the use of a dynTask, which is a nice win (dynTasks don't show up well in `inspect tree`).

sjrd:
Woot :tada:

PR #2779, opened 2017-02-20
Deprecate top-level @JSExport, and @JSExportDescendent{Classes,Objects}.

sjrd:
Top-level `@JSExport`s (on classes and objects) should be replaced by `@JSExportTopLevel`. For objects, this is a change of semantics, as the object itself is exported, rather than a 0-arg function returning the object. It is still possible to achieve the same behavior as before by explicitly exporting a 0-arg method returning the object with `@JSExportTopLevel`, so we do not lose any expressiveness.

`@JSExportDescendentClasses` and `@JSExportDescendentObjects` should not have any use cases left, now that reflective instantiation and module loading is properly supported through `@EnableReflectiveInstantiation` and `scala.scalajs.reflect.Reflect`.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
@@ -31,6 +31,11 @@ trait PrepJSExports { this: PrepJSInterop =>
     assert(!isNamed || destination == ExportDestination.Normal)
   }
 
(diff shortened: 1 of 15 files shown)


sjrd:
Review by @gzm0

gzm0:
Good to merge this one, or do you want to wait for module initializers?

sjrd:
Let's merge this one. I'll have a non-trivial rebase of module initializers anyway because of the renaming, so this won't make it any more difficult.

PR #2784, opened 2017-02-25
Add facades for ES6 Iterator and Iterable

gzm0:
Also add conversions to and from relevant Scala types. Note that these facades do not take return values into account (which are needed for full generator typing).

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Iterable.scala
+++ b/library/src/main/scala/scala/scalajs/js/Iterable.scala
@@ -1,0 +1,23 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2017, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+import js.annotation._
+
+/** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+ *  JavaScript Iterable.
+ */
+@ScalaJSDefined
+trait Iterable[+A] extends js.Object {
+  /** JavaScript Iterator for this Iterable. */
+  @JSName(Symbol.iterator)
+  def jsIterator(): Iterator[A]
+}
--- a/library/src/main/scala/scala/scalajs/js/IterableOps.scala
+++ b/library/src/main/scala/scala/scalajs/js/IterableOps.scala
@@ -1,0 +1,17 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package scala.scalajs.js
+
+/** Adapts a JavaScript Iterable to a Scala Iterable */
+@inline
+final class IterableOps[+A](self: Iterable[A])
+    extends scala.collection.Iterable[A] {
+  @inline
+  def iterator: scala.collection.Iterator[A] = self.jsIterator().toIterator
+}
--- a/library/src/main/scala/scala/scalajs/js/Iterator.scala
+++ b/library/src/main/scala/scala/scalajs/js/Iterator.scala
@@ -1,0 +1,54 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+import js.annotation._
+
+/** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+ *  JavaScript Iterator.
+ */
+@ScalaJSDefined
+trait Iterator[+A] extends js.Object {
+  def next(): Iterator.Entry[A]
+}
+
+object Iterator {
+  /** Return value of [[Iterator.next]]. */
+  @ScalaJSDefined
+  trait Entry[+A] extends js.Object {
+    /** Whether the iterator has completed. */
+    def done: Boolean
+
+    /** The current value. Reading this value is only valid if done is false. */
+    def value: A
+  }
+
+  @inline
+  private final class WrappedIterator[+A](self: Iterator[A])
+      extends scala.collection.Iterator[A] {
+    private[this] var lastEntry = self.next()
+
+    @inline
+    def hasNext: Boolean = !lastEntry.done
+
+    @inline
+    def next(): A = {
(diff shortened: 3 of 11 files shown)


sjrd:
Yes, that looks reasonable. I think I had not made `js.Array` extend `js.Iterable` because collection operations on `js.Array` should use `WrappedArray`, but the lower priority implicit is better. 

Btw I think we have precedent for naming that one Low*est*ProrityImplicits.

gzm0:
Updated. Review by @sjrd

library/src/main/scala/scala/scalajs/js/Iterable.scala
```diff
@@ -0,0 +1,23 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2015, LAMP/EPFL   **
```
sjrd: 2017.

Seems we should have some global pass over these header comments, one of these days.

library/src/main/scala/scala/scalajs/js/IterableOps.scala
```diff
@@ -0,0 +1,17 @@
...
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-lang.org/     **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package scala.scalajs.js
+
+/** Adapts a JavaScript Iterable to a Scala Iterable */
+@inline
+final class IterableOps[+A](private[this] val self: Iterable[A])
```
sjrd: Remove `private[this] val`?

library/src/main/scala/scala/scalajs/js/Iterator.scala
```diff
@@ -0,0 +1,55 @@
...
+    @inline
+    def next(): A = {
+      val value = lastEntry.value
+      lastEntry = self.next()
+      value
+    }
+  }
+
+  @inline
+  final implicit class IteratorOps[A](private[this] val self: Iterator[A]) {
```
sjrd: This should be a value class, instead of an `@inline` reference class.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -41,6 +42,46 @@ object JSConverters extends JSConvertersLowPrioImplicits {
     @inline final def toJSArray: Array[T] = genTraversableOnce2jsArray(col)
   }
 
+  implicit class JSRichGenIterable[T](val col: GenIterable[T]) extends AnyVal {
```
sjrd: Use obscure names like `__self` for the underlying field of an `implicit` value class, so that it's obvious it should not be used directly on a `GenIterable`.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -41,6 +42,46 @@ object JSConverters extends JSConvertersLowPrioImplicits {
...
+    @inline final def toJSIterable: Iterable[T] = new IterableAdapter(col)
+  }
+
+  implicit class JSRichIterator[T](
+      val it: scala.collection.Iterator[T]) extends AnyVal {
+    @inline final def toJSIterator: Iterator[T] = new IteratorAdapter(it)
+  }
+
+  @inline
```
sjrd: `@inline` on a JS class doesn't do anything. I suggest you remove it so that we do not pretend it does.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -41,6 +42,46 @@ object JSConverters extends JSConvertersLowPrioImplicits {
...
+  @ScalaJSDefined
+  private class IterableAdapter[+T](private[this] val col: GenIterable[T])
+      extends Iterable[T] {
+    @inline
+    @JSName(Symbol.iterator)
+    final def jsIterator(): Iterator[T] = col.iterator.toJSIterator
+  }
+
+  @inline
```
sjrd: Same here.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -41,6 +42,46 @@ object JSConverters extends JSConvertersLowPrioImplicits {
...
+  implicit class JSRichIterator[T](
+      val it: scala.collection.Iterator[T]) extends AnyVal {
+    @inline final def toJSIterator: Iterator[T] = new IteratorAdapter(it)
+  }
+
+  @inline
+  @ScalaJSDefined
+  private class IterableAdapter[+T](private[this] val col: GenIterable[T])
+      extends Iterable[T] {
+    @inline
```
sjrd: Same here.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -41,6 +42,46 @@ object JSConverters extends JSConvertersLowPrioImplicits {
...
+    @JSName(Symbol.iterator)
+    final def jsIterator(): Iterator[T] = col.iterator.toJSIterator
+  }
+
+  @inline
+  @ScalaJSDefined
+  private class IteratorAdapter[+T](
+      private[this] val it: scala.collection.Iterator[T]) extends Iterator[T] {
+    @inline
```
sjrd: And same here.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/IterableTest.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.Assume._
+import org.junit.{BeforeClass, Test}
+
+object IterableTest {
+  @BeforeClass def assumeSymbolsAreSupported(): Unit = {
+    assumeTrue("Assuming JavaScript symbols are supported",
+        org.scalajs.testsuite.utils.Platform.areJSSymbolsSupported)
+  }
+
+  @inline
```
sjrd: Same: `@inline` is not doing anything here.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/IterableTest.scala
```diff
@@ -0,0 +1,116 @@
...
+  @ScalaJSDefined
+  private class CounterIterable(val max: Int) extends js.Iterable[Int] {
+    @JSName(js.Symbol.iterator)
+    def jsIterator(): js.Iterator[Int] = new CounterIterator(max)
+  }
+
+  @inline
```
sjrd: One more useless `@inline`

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/IterableTest.scala
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+    }
+  }
+}
+
+class IterableTest {
+  import IterableTest._
+
+  def assertValue[T](expected: T, entry: js.Iterator.Entry[T]) = {
```
sjrd: Explicit result type missing.

gzm0:
Updated.

gzm0:
Jenkins, retest this please

gzm0:
Note: The conversion from `js.Iterable` to `sc.Iterable` is potentially unsafe. MSN states "A generator object is both, iterator and iterable”. The `[Symbol.iterator]` method of a generator just returns itself, implying that it is OK (required?) for a `js.Iterable` to not return a fresh iterator. I am concerned that this might break some methods of `sc.Iterator`.

Maybe we should only allow the `toIterator` conversion and maybe add a utility to get a Scala iterator from a `js.Iterable`?

WDYT?

sjrd:
The main goal of the conversion to `sc.Iterable` is to be able to use for comprehensions on such things, which is necessary to be on par with JavaScript and its for..of loop. Generators return themselves because they're essentially `Iterator`s to begin with, but JavaScript needs them to be `Iterable` to be usable in for..of. We don't need to present them as iterables, because we can use our for comprehensions directly on iterators too. 

So basically IMO what you describe is a non-issue for Scala.js, and we should not sacrifice the conversion to `sc.Iterable` for that.

gzm0:
Ok, fair enough. So we'll just not introduce something like `Generatable`. Do we need to clarify the contract on `js.Iterable`? It might be just confusing...

PR #2789, opened 2017-03-03
Fix #2778: Warn on self-referencing @JSName

### LINKED ISSUE #2778: Warn if a @JSName refers to a symbol in the enclosing object.
The following code:
```scala
@ScalaJSDefined
object Foo extends js.Any {
  val sym = js.Symbol()

  @JSName(sym)
  def yolo(): Int = 1
}
```
crashes with a stack overflow, because `sym` is needed to create the class `Foo$`, but `Foo$` is needed to instantiate `Foo` to get `sym`.

There's nothing wrong with the fact that it crashes, but we could probably warn in simple cases like that.

gzm0:


gzm0:
Review by @sjrd

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1100,10 +1100,15 @@ abstract class PrepJSInterop extends plugins.PluginComponent
         }
       } else {
         // We have a js.Symbol
-        val sym = argTree.symbol
-        if (!sym.isStatic || !sym.isStable) {
+        val symSym = argTree.symbol
+        if (!symSym.isStatic || !symSym.isStable) {
           reporter.error(argTree.pos,
               "A js.Symbol argument to JSName must be a static, stable identifier")
+        } else if (sym.owner.hasAnnotation(ScalaJSDefinedAnnotation) &&
```
sjrd: Can't you use `enclosingOwner` here?

gzm0:
Updated. Note the new type-pedantry commit.

PR #2790, opened 2017-03-03
Fix #2772: Allow @JSName on objects inside SJS defined JS class

### LINKED ISSUE #2772: Erroneous warning on @JSName on an object inside a Scala.js-defined JS class
The following works:
```scala
@ScalaJSDefined
trait InnerObjectTrait extends js.Any {
  @JSName("foo")
  val innerObject: AnyRef
}

@ScalaJSDefined
class SJSDefinedInnerObject extends js.Object with InnerObjectTrait {
  @JSName("foo")
  object innerObject { // scalastyle:ignore
    override def toString(): String = "SJSDefinedInnerObject.innerObject"
  }
}

val obj0 = new SJSDefinedInnerObject
assertEquals("object", js.typeOf(obj0.innerObject.asInstanceOf[js.Any]))
assertEquals("SJSDefinedInnerObject.innerObject", obj0.innerObject.toString())

val obj1: InnerObjectTrait = new SJSDefinedInnerObject
assertEquals("object", js.typeOf(obj1.innerObject.asInstanceOf[js.Any]))
assertEquals("SJSDefinedInnerObject.innerObject", obj1.innerObject.toString())
```
but it erroneously emits the following warning:
```
[warn] /localhome/doeraene/projects/scalajs/examples/helloworld/HelloWorld.scala:37: Non JS-native classes, traits and objects should not have an @JSName annotation, as it does not have any effect. This will be enforced in 1.0.
[warn]   @JSName("foo")
[warn]    ^
```

gzm0:


compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -129,6 +129,43 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+    }
+    """ hasWarns
+    """
+      |newSource1.scala:7: warning: Non JS-native classes, traits and objects should not have an @JSName annotation, as it does not have any effect. This will be enforced in 1.0.
+      |      @JSName("foo")
+      |       ^
+      |newSource1.scala:10: warning: Non JS-native classes, traits and objects should not have an @JSName annotation, as it does not have any effect. This will be enforced in 1.0.
+      |      @JSName("bar")
+      |       ^
+    """
```
gzm0: Do we need better messages here?

sjrd:
Looks good, but I'd like to examine on an actual computer the interactions with my ongoing branch https://github.com/scala-js/scala-js/compare/master...sjrd:js-import-global (which I realize now you had no idea I was working on :-s)

PR #2792, opened 2017-03-05
TypedArrays are js.Iterable

gzm0:


project/BinaryIncompatibilities.scala
```diff
@@ -152,6 +152,75 @@ object BinaryIncompatibilities {
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
           "scala.scalajs.js.Promise.all"),
 
+      // js.Iterable support in TypedArray.
+      // Not a compatibility issue (due to JS land).
+      ProblemFilters.exclude[IncompatibleMethTypeProblem](
+          "scala.scalajs.js.typedarray.Int8Array.set"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem](
+          "scala.scalajs.js.typedarray.Int8Array.set"),
```
sjrd: These are duplicates. There are quite a number of other duplicates here.

test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
```diff
@@ -192,25 +192,48 @@ trait TypedArrayTest[V, T <: TypedArray[V, T]] {
...
+    import js.JSConverters._
+
+    val x = itCtor(Iterable(1, 2, 3, 4).toJSIterable)
+
+    assertEquals(4, x.length)
+    assertEquals(1, x(0))
+    assertEquals(2, x(1))
+    assertEquals(3, x(2))
+    assertEquals(4, x(3))
+  }
```
sjrd: These two new tests must assume symbols are supported.

library/src/main/scala/scala/scalajs/js/typedarray/Float32Array.scala
```diff
@@ -18,7 +18,7 @@ class Float32Array private extends TypedArray[Float, Float32Array] {
   def this(typedArray: TypedArray[_, _]) = this()
 
   /** Creates a new Float32Array with the elements in the given array */
-  def this(array: js.Array[_]) = this()
+  def this(array: js.Iterable[_]) = this()
```
sjrd: With this new constructor, is the constructor taking a `TypedArray` still necessary?

Shouldn't we simply add an overload with `js.Iterable` rather than modifying this one? How is the spec formulated?

gzm0: Spec says there is one constructor taking `object` that forwards to `TypedArray.from`, which in turn dynamically checks whether `@@iterator` is defined. If it is, elements are extracted using an iterator. Otherwise, it is assumed `object` is an array-like object and extraction is performed using indexed access.

Given that `js.Array[T] <: js.Iterable[T]`, this seemed the most reasonable typing, although indeed, the constructor taking a `TypedArray` also should go away. This seems like the most reasonable option to me.

sjrd: Yes, I reread the spec and it seems what you did now is the right thing to do. The spec has a special-case for TypedArrays, and a generic case for other iterables. It does not mention Arrays specifically (those fall into the iterable case).

gzm0:
Updated.

PR #2793, opened 2017-03-05
Fix #2234: Clean up emitter package

### LINKED ISSUE #2234: Clean up ScalaJSClassEmitter and JSDesugaring
Seriously, this is such a mess ...

gzm0:


gzm0:
@sjrd I have made an attempt at bringing a bit of order to the chaos. WDYT?
If you are ok with these changes, I'll update the binary compat stuff.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -195,15 +195,14 @@ import java.io.StringWriter
  *
  *  @author Sébastien Doeraene
  */
-private[emitter] class JSDesugaring(semantics: Semantics,
-    outputMode: OutputMode, internalOptions: InternalOptions) {
+private[emitter] class JSDesugaring(semantics: Semantics, outputMode: OutputMode,
+    jsGen: JSGen) {
```
sjrd: If we give `semantics` and `outputMode` to `JSGen` and make them public `val`s thereof, then `JSDesugaring` and `ClassEmitter` only need to take a `JSGen` as parameter. It also guarantees some more consistency in the semantics and output mode applied between a `JSGen`'s methods and other parts using the `JSGen`

gzm0:
Updated. A note about the names.

> `ClassEmitter` to `ClassSkeletonEmitter`

Not a fan TBH. `ClassEmitter` *does* emit classes. The fact that it uses other components to achieve this is IMO an implementation detail (and should influence the name only in rare cases).

> `JSDesugaring` to `MethodBodyEmitter`.

Good point. I went with `FunctionEmitter` due to following advantages:

- Shorter.
- Also understandable in a JS context (where a method is not so well-defined).
- Closer to what it does: Every public method returns `js.Function` tree (except for `transformIdent` which should ideally not be there).

Are you OK with these names?

sjrd:
> Are you OK with these names?

Yes, they look reasonable.

gzm0:
Also, this is ready for final review.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -247,10 +245,10 @@ private[emitter] class JSDesugaring(semantics: Semantics,
       desugar.transformExpr(tree)(env)
   }
 
-  private[emitter] implicit def transformIdent(ident: Ident): js.Ident =
+  implicit def transformIdent(ident: Ident): js.Ident =
     js.Ident(ident.name, ident.originalName)(ident.pos)
```
sjrd: Why is this not private? If it needs to be public, I feel like it should be defined in `JSGen`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSDesugaring.scala
```diff
@@ -2261,276 +2259,6 @@ private[emitter] class JSDesugaring(semantics: Semantics,
...
-      implicit pos: Position): js.Tree = {
-    require(item != "eval")
-    if (internalOptions.optimizeBracketSelects)
-      js.DotSelect(qual, js.Ident(item))
-    else
-      js.BracketSelect(qual, js.StringLiteral(item))
-  }
-
   class DesugarException(tree: Tree,
       cause: Throwable) extends Exception(exceptionMsg(tree), cause)
```
sjrd: I think this one is completely unused. Candidate for a removal in the "Remove unused stuff" commit?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -0,0 +1,298 @@
...
+          case _ =>
+            pathSelection(moduleValue, path)
+        }
+    }
+  }
+
+  private final val DefaultExportName = "default"
+
+  def envModuleField(module: String)(
+      implicit pos: Position): VarRef = {
```
sjrd: Looks like this method header would fit on one line.

In general, a bunch of methods in `JSGen` have shorter headers than they used to, and their formatting should be revisited.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -0,0 +1,298 @@
...
+
+/** Collection of tree generators that are used accross the board.
+ *  This class is fully stateless.
+ *
+ *  Also carries around config (semantics and outputMode).
+ */
+private[emitter] final class JSGen(val semantics: Semantics,
+    val outputMode: OutputMode, internalOptions: InternalOptions) {
+
+  private final val ScalaJSEnvironmentName = "ScalaJS"
```
sjrd: This should go in the companion object.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -0,0 +1,298 @@
...
+          case DefaultExportName :: rest =>
+            val defaultField = genCallHelper("moduleDefault", moduleValue)
+            pathSelection(defaultField, rest)
+          case _ =>
+            pathSelection(moduleValue, path)
+        }
+    }
+  }
+
+  private final val DefaultExportName = "default"
```
sjrd: This should go in the companion object.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -23,11 +23,11 @@ import org.scalajs.core.tools.linker.backend.OutputMode
 import org.scalajs.core.tools.linker.{LinkedClass, LinkingUnit}
 
 /** Emitter for the skeleton of classes. */
-private[emitter] final class ScalaJSClassEmitter(jsGen: JSGen) {
+private[emitter] final class ClassEmitter(jsGen: JSGen) {
 
-  private val jsDesugaring = new JSDesugaring(jsGen)
+  private val jsDesugaring = new FunctionEmitter(jsGen)
```
sjrd: `val functionEmitter` ;)

gzm0:
Updated.

sjrd:

sjrd:

PR #2794, opened 2017-03-06
Demand `@JSGlobal` for non-`@JSImport` JS native classes and objects.

### LINKED ISSUE #2615: Disallow @JSName containing dots inside nested facades
This should be disallowed:

``` scala
@js.native
object Foo extends js.Object {
  @JSName("blah.foo")  // <- dot here is strange.
  @js.native
  class Bar
}
```

sjrd:
Previously, a top-level `@js.native` class/object was loaded from the global scope by default, either with an explicit name through `@JSName` or with an implicit name derived from its Scala name.

With this commit, such a native class/object should be declared with an `@JSGlobal` annotation instead (either with or without an explicit name).

For example,
```scala
  @js.native class Foo extends js.Object
  @js.native @JSName("Foobaz") object Bar extends js.Object
```
would now be written as
```scala
  @js.native @JSGlobal class Foo extends js.Object
  @js.native @JSGlobal("Foobaz") object Bar extends js.Object
```

There are two main reasons for such a change.

First, it brings loading from the global scope syntactically on par with importing from a module. As the JS eco-system moves towards modules, the syntactic bias towards globally scoped things is not relevant anymore.

Second, it removes the overloaded meaning of `@JSName`. Indeed, this annotation has significantly different meanings whether it is attached to a top-level native class/object versus a member of a JS type. For example, a `.` has a special meaning for top-level native class/object, but not for a member of a JS type. The changes in this commit close #2615 by leaving only the meaning where `.` is not special, and therefore need not be forbidden.

Since this is a fairly disruptive change, the deprecation warnings can be suppressed with the option `-P:scalajs:suppressMissingJSGlobalDeprecations` during the 0.6.x cycle, to ease migration.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/annotation/JSGlobal.scala
+++ b/library/src/main/scala/scala/scalajs/js/annotation/JSGlobal.scala
@@ -1,0 +1,46 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js API               **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2017, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package scala.scalajs.js.annotation
+
+/** Marks the annotated class or object as being a member of the JavaScript
+ *  global scope.
+ *
+ *  The annotated class/object must also be annotated with `@js.native`, and
+ *  therefore extend [[scala.scalajs.js.Any js.Any]].
+ *
+ *  Given:
+ *  {{{
+ *  @js.native
+ *  @JSGlobal
+ *  class Foo extends js.Object
+ *
+ *  @js.native
+ *  @JSGlobal("Foobar")
+ *  object Bar extends js.Object
+ *
+ *  @js.native
+ *  @JSGlobal("Lib.Babar")
+ *  class Babar extends js.Object
+ *  }}}
+ *
+ *  The following mappings apply (`global` denotes the global scope):
+ *
+ *  {{{
+ *  Scala.js                | JavaScript
+ *  ------------------------+------------------
+ *  new Foo()               | new global.Foo()
+ *  Bar                     | global.Foobar
+ *  js.constructorOf[Babar] | global.Lib.Babar
+ *  }}}
+ *
+ *  @see [[http://www.scala-js.org/doc/calling-javascript.html Calling JavaScript from Scala.js]]
+ */
+class JSGlobal extends scala.annotation.StaticAnnotation {
+  def this(name: String) = this()
+}
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
@@ -67,6 +67,7 @@ trait JSDefinitions { self: JSGlobalAddons =>
     lazy val JSExportStaticAnnotation  = getRequiredClass("scala.scalajs.js.annotation.JSExportStatic")
     lazy val JSExportTopLevelAnnotation = getRequiredClass("scala.scalajs.js.annotation.JSExportTopLevel")
     lazy val JSImportAnnotation        = getRequiredClass("scala.scalajs.js.annotation.JSImport")
+    lazy val JSGlobalAnnotation        = getRequiredClass("scala.scalajs.js.annotation.JSGlobal")
     lazy val JSGlobalScopeAnnotation   = getRequiredClass("scala.scalajs.js.annotation.JSGlobalScope")
     lazy val ScalaJSDefinedAnnotation  = getRequiredClass("scala.scalajs.js.annotation.ScalaJSDefined")
     lazy val SJSDefinedAnonymousClassAnnotation = getRequiredClass("scala.scalajs.js.annotation.SJSDefinedAnonymousClass")
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
@@ -191,17 +191,19 @@ trait JSGlobalAddons extends JSDefinitions
      */
     def jsNameOf(sym: Symbol): JSName = {
       sym.getAnnotation(JSNameAnnotation).fold[JSName] {
-        val base = sym.unexpandedName.decoded.stripSuffix("_=")
-        val name =
-          if (!sym.isMethod) base.stripSuffix(" ")
-          else base
-        JSName.Literal(name)
+        JSName.Literal(defaultJSNameOf(sym))
       } { annotation =>
         annotation.args.head match {
           case Literal(Constant(name: String)) => JSName.Literal(name)
           case tree                            => JSName.Computed(tree.symbol)
         }
       }
+    }
+
+    def defaultJSNameOf(sym: Symbol): String = {
+      val base = sym.unexpandedName.decoded.stripSuffix("_=")
+      if (!sym.isMethod) base.stripSuffix(" ")
+      else base
     }
 
     /** Gets the fully qualified JS name of a static module Symbol compiled
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -75,6 +75,11 @@ abstract class PrepJSInterop extends plugins.PluginComponent
     val scala_ = newTypeName("scala") // not defined in 2.10's tpnme
   }
 
+  private final val SuppressMissingJSGlobalDeprecationsMsg = {
+    "\n  (you can suppress this warning in 0.6.x by passing the option " +
+    "`-P:scalajs:suppressMissingJSGlobalDeprecations` to scalac)"
+  }
+
   class JSInteropTransformer(unit: CompilationUnit) extends Transformer {
 
     // Force evaluation of JSDynamicLiteral: Strangely, we are unable to find
@@ -546,6 +551,7 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
       if (shouldCheckLiterals) {
         checkJSNameArgument(implDef)
+        checkJSGlobalLiteral(sym)
         checkJSImportLiteral(sym)
       }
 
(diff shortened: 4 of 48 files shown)


sjrd:
Preliminary review by @gzm0. I still need to add "run" tests in test-suite/, hence the DO NOT MERGE flag for now.

sjrd:
I've discussed with @olafurpg to have a [scalafix](https://scalacenter.github.io/scalafix/) automated rewrite for this, so that maintainers of facade libraries are not plagued with going through dozens of manual changes.

gzm0:
From a very high-level point of view, this looks like a huge improvement to me.

One thing I'm unsure of is whether to keep the dot import separation.

- In earlier discussions you mentioned that you think disallowing "nested imports" for both `JSImport` and `JSGlobal` is the right way to go.
- I'm not sure that implicitly using dot separation is a good thing. Have you considered using a variable argument list?

I'm probably just missing the bigger picture on how this will lead us to the final state (including `@LoadVia`).

> I've discussed with @olafurpg to have a scalafix automated rewrite for this, so that maintainers of facade libraries are not plagued with going through dozens of manual changes.

Is anything more needed than renaming `@JSName` on static classes not inside a native object? (I know this is complex enough to warrant a rewrite, just want to make sure I'm not missing something).

sjrd:
> Is anything more needed than renaming @JSName on static classes not inside a native object? (I know this is complex enough to warrant a rewrite, just want to make sure I'm not missing something).

It needs to *add* the `@JSGlobal` if there is no `@JSName`, but otherwise, it's as simple as that.

Regarding dots. Well, for better or worse, we already decided to keep implicit dot-separation when we added `@JSImport`. IMO `@JSGlobal` should be consistent with `@JSImport`. We could move to a varargs-based version, and deprecated `@JSImport("foo.js", "Foo.Bar")` in favor of `@JSImport("foo.js", "Foo", "Bar")` in 0.6.x, and remove the special meaning of `.` in 1.0.0 (same for `@JSGlobal`). It's a bit orthogonal to the current PR though.

Currently, my vision for the final state is that an `@js.native` class or object must have exactly 1 annotation among `@JSGlobal`, `@JSImport` and `@JSLoadVia`. Whether the first two are `.`-separated or varargs-separated, I do not have strong feelings about. The need to use `.` as a normal character should be extremely rare, and in those cases we can fall back on `@JSLoadVia`.

olafurpg:
I took a quick stab at this rewrite, see https://github.com/scalacenter/scalafix/compare/master...olafurpg:js.native?expand=1#diff-7248893d3c1b81c7e0bf521edc075497

Some remaining challenges:

- what to do with imports? The current implementation leaves imports alone if the user configures `imports.organize = false` (default) in their `.scalafix.conf`. If `imports.organize = true`, then the rewrite inserts an import on `JSGlobal` if it's missing.
- how to ship the rewrite to users? This is still an unsolved problem. The options are
  1. take a shortcut and publish the rewrite with scalafix-core, but that's not a sustainable solution
  2. we merge this rewrite into some Scala.js repo and you guys publish the rewrite to maven. Then users add a dependency to that module and put `rewrites = [_root.org.scalajs.rewrites.DemandJSGlobal]` into `.scalafix.conf`. 

Feedback/suggestions are welcome. I can try and get option 2 working tomorrow.

sjrd:
@olafurpg For imports, I would suggest the following when `imports.organize = false`:
1. If there is an `import scala.scalajs.js.annotation._`, do nothing
2. If there is an `import scala.scalajs.js.annotation.JSName`, then:
  a) If `JSName` is otherwise unused in the compilation unit (not sure you can check that easily), replace it with `import scala.scalajs.js.annotation.JSGlobal`
  b) Otherwise, replace it with `import scala.scalajs.js.annotation.{JSName, JSGlobal}`
3. Otherwise, add an `import scala.scalajs.js.annotation.JSGlobal` after all other compilation-unit-level imports.

Note that there's nothing `JSName`- or `JSGlobal`-specific to the above. Ignoring the improvement of 2.a), it could be a generic patch `AddImport(fullName)` that tries to find an existing import statement for things at the same level.

On a different topic, @gzm0's comment made me realize that, to be completely correct, the rewrite should not apply to an `@js.native` class/object whose direct owner is itself an `@js.native`. Can this be implemented easily? For a test case:
```scala
@js.native
object Foo extends js.Object {
  @js.native
  class Bar extends js.Object

  @js.native
  @JSName("Foobaz")
  object Baz extends js.Object
}
```
should only affect `Foo`, and hence yield:
```scala
@js.native
@JSGlobal
object Foo extends js.Object {
  @js.native
  class Bar extends js.Object

  @js.native
  @JSName("Foobaz")
  object Baz extends js.Object
}
```

sjrd:
> I updated the implementation to handle nested @js.native classes, this was not so hard with a custom traverser.

Nice :)

sjrd:
@olafurpg Two more cases where the rewrite shouldn't be applied:

* The class/object is annotated with `@JSImport(_, _)`
* The class/object is annotated with `@JSGlobalScope`
* (ideally, but this syntax was already deprecated:) the class/object `extends js.GlobalScope`

The fact that the rewrite is not applied to those does not impact the fact that they are `@js.native` owners, and therefore their inner class/objects should not be affected by the rewrite.

Test case:
```scala
@js.native
@JSImport("foo.js", "Foo")
class Foo1 extends js.Object {
  @js.native class Inner1 extends js.Object
  @js.native @JSName("Baz") object Inner2 extends js.Object
}

@js.native
@JSImport("foo.js", "Foo")
object Foo1 extends js.Object {
  @js.native class Inner1 extends js.Object
  @js.native @JSName("Baz") object Inner2 extends js.Object
}

@js.native
@JSGlobalScope
object Foo2 extends js.Object {
  @js.native class Inner1 extends js.Object
  @js.native @JSName("Baz") object Inner2 extends js.Object
}

@js.native
object Foo3 extends js.GlobalScope {
  @js.native class Inner1 extends js.Object
  @js.native @JSName("Baz") object Inner2 extends js.Object
}
```
is untouched.

sjrd:
Updated with `test-suite` changes. It's complete now.

japgolly:
> my vision for the final state is that an @js.native class or object must have exactly 1 annotation among @JSGlobal, @JSImport and @JSLoadVia.

This is becoming a nightmare for me as a library maintainer. Given this direction I'm faced with 3 unhappy choices:

1. Put `@JSGlobal` on my facades and don't allow users to use `CommonJSModule`.
2. Put `@JSImport` on my facades and force users to use `CommonJSModule` and probably force them to use scalajs-bundler which will in turn force them to only import deps a certain way and, only output one huge single JS file.
3. Double build and double publish all of my library artifacts so that users can choose whether they want to use modules / script.

Wouldn't it be better to allow us to annotate our facades with both and then downstream at link time, choose one or the other depending on users' `ModuleKind` setting?

sjrd:
@japgolly Hum, that's an entirely different issue. Would you mind opening a separate issue for this so that we can discuss it without disrupting the PR comments.

japgolly:
Ah yeah, good point. Sorry I just saw that comment here and immediately pooped my pants. I'll poop them somewhere new.

library/src/main/scala/scala/scalajs/js/annotation/JSGlobal.scala
```diff
@@ -0,0 +1,46 @@
...
+ *  ------------------------+------------------
+ *  new Foo()               | new global.Foo()
+ *  Bar                     | global.Foobar
+ *  js.constructorOf[Babar] | global.Lib.Babar
+ *  }}}
+ *
+ *  @see [[http://www.scala-js.org/doc/calling-javascript.html Calling JavaScript from Scala.js]]
+ */
+class JSGlobal extends scala.annotation.StaticAnnotation {
+  def this(name: String) = this()
```
gzm0: Just to double check your opinion on this. Currently there is still an inconsistency with implied `@JSGlobal` names and implied `@JSExport` names: `@JSExport` exports to the fully qualified path, whereas `@JSGlobal` imports from the direct name only.

I'm assuming the story here is that with the advent of `@JSExportTopLevel` that will (hopefully) eventually be called `@JSExport` again, we can then change the default to direct name again (since `@JSExportTopLevel` does currently not allow for a default).

Sound about right?

sjrd:
Updated.

> Which of the following are allowed (if any). And what do they mean?

None of them are allowed. I have enhanced the tests `noJSGlobalOnMembers` and `noJSImportOnMembers` to assert this.

library/src/main/scala/scala/scalajs/js/annotation/JSGlobal.scala
sjrd: Yes, the idea is that `@JSExportTopLevel` takes over here, and since it doesn't have a default, it's all right.

I do not expect to rename `@JSExportTopLevel` to `@JSExport`. Currently `@JSExportTopLevel` is consistent between top-level classes/objects and static methods and fields. If we rename `@JSExportTopLevel` to `@JSExport`, we break this consistency, plus we reintroduce the *in*consistency between `@JSExport` on a top-level class/object versus on a member of a Scala class. If exports of instance members had been disallowed, the story would have been different.

japgolly:
There are no tests in the PR that cover the case of `@JSGlobal` & `@JSImport` coexisting on the same target. I'd really like to use such an approach in scalajs-react 1.0. Can I assume it works and is valid (at least for now)?

library/src/main/scala/scala/scalajs/js/annotation/JSGlobal.scala
gzm0: Ah, ok. I was expecting that we want to rename `@JSExport` to something else (as in `@JSBridge` :P). But maybe that is too much renaming for little gain. Anyways, this is OT on this PR.

gzm0:
> There are no tests in the PR that cover the case of @JSGlobal & @JSImport coexisting on the same target. 

The `noTwoJSNativeLoadSpecAnnots` test auto generates tests for all pairs of load spec annotations (`JSGlobal`, `JSGlobalScope`, `JSImport`, `JSName`) and makes sure no two of them are used at the same time. So it is currently not valid and that is tested :)

sjrd:
@japgolly There is test [here](https://github.com/sjrd/scala-js/blob/08589e5fd5f6312cff192005fb1644f8b0dd0510/compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala#L335) that asserts that a compile error is raised if two annotations among the set `@JSName`, `@JSGlobal` `@JSImport` and `@JSGlobalScope` are used. It doesn't show up in the PR diff because it's generically defined in terms of [`JSNativeLoadSpecAnnots`](https://github.com/scala-js/scala-js/pull/2794/files#diff-b5595f9f3973da793d674a8dd94d8a2dR18)

olafurpg:
Here's the output from running the rewrite on scala-js-dom: https://github.com/scala-js/scala-js-dom/compare/master...olafurpg:demand-js-global?expand=1

Please take a look at the diff and see if there's something wrong. When I try to compile the code I get a ton of "not found: type JSGlobal", is that expected?

Yesterday, I made it possible to load rewrites from source. This avoids the need to publish compiled rewrites to Maven. Instead, it's enough to link to a raw github url in the `.scalafix.conf` file.

I tried to run the rewrite on scalajs-react but it failed because scalajs-react doesn't compile with `-Yrangepos` enabled. The scalac option `-Yrangepos` is required to use the scala.meta semantic api.

sjrd:
> When I try to compile the code I get a ton of "not found: type JSGlobal", is that expected?

Yes, that's expected. You need 0.6.15-SNAPSHOT built from master to use `@JSGlobal`. You can build it from source using the instructions at https://github.com/scala-js/scala-js/blob/master/DEVELOPING.md#publish-locally

I'll have a look at the diff.

sjrd:
The diff looks good :+1: 

Here is a repo with `@JSImport`s: https://github.com/chandu0101/sri

PR #2796, opened 2017-03-08
Deprecate private members of native JS classes/traits/objects.

sjrd:
Historically, our compiler has tolerated `private`, `private[this]` and `private[Scope]` members in native JS classes, traits, and objects. Once compiled, it would consider that they are in fact public, with non-mangled names.

This is at odds with how private members of Scala.js-defined JS types are handled at use site, which means there are different semantics at use site between native versus Scala.js-defined types, which is bad.

This commit deprecates definitions of private members in native JS types. We still accept `private[this]` constructors, because they are sometimes necessary to have a primary constructor that other constructors delegate to. `private[this]` constructors are fine, because they cannot be called from anywhere but the secondary constructors. Note that even a `private` constructor can be called from the companion object, which can be non-native, so this is not allowed.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -962,6 +962,35 @@ abstract class PrepJSInterop extends plugins.PluginComponent
             "This will become an error in 1.0.0.")
       }
 
+      /* In native JS types, there should not be any private member, except
+       * private[this] constructors.
+       */
+      if ((enclosingOwner is OwnerKind.JSNative) && isPrivateMaybeWithin(sym)) {
+        // Necessary for `private[this] val/var
+        def isFieldPrivateThis: Boolean = {
+          sym.isPrivateThis &&
+          !sym.isParamAccessor &&
+          !sym.owner.info.decls.exists(s => s.isGetter && s.accessed == sym)
+        }
+
+        if (sym.isClassConstructor) {
+          if (!sym.isPrivateThis) {
+            reporter.warning(tree.pos,
+                "Declaring private constructors in native JS classes is " +
+                "deprecated, because they do not behave the same way as in " +
+                "Scala.js-defined JS classes. Use `private[this]` instead. " +
+                "This will become an error in 1.0.0.")
+          }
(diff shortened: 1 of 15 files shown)


sjrd:
Review by @gzm0. Only the second commit is part of this PR. The first one is part of #2794. Rebasing after the fact would have been annoying, so I already built on top of it.

sjrd:
Rebased.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -940,6 +940,35 @@ abstract class PrepJSInterop extends plugins.PluginComponent
...
+
+        if (sym.isClassConstructor) {
+          if (!sym.isPrivateThis) {
+            reporter.warning(tree.pos,
+                "Declaring private constructors in native JS classes is " +
+                "deprecated, because they do not behave the same way as in " +
+                "Scala.js-defined JS classes. Use `private[this]` instead. " +
+                "This will become an error in 1.0.0.")
+          }
+        } else if (sym.isMethod || isFieldPrivateThis) {
```
gzm0: Just to double check: You are just warning on methods to not have double warnings on fields.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -1002,6 +1002,76 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
   @Test
+  def warnPrivateMemberInNative: Unit = {
+
+    """
+    @js.native
+    @JSGlobal
+    class A extends js.Object {
+      private[this] val a: Int = js.native
+      private val b: Int = js.native
+      private[A] val c: Int = js.native
```
gzm0: Could you add a var as well, just to make sure we have the right number of warnings?

sjrd: Done.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
sjrd: Yes, that's right.

sjrd:
Updated.

PR #2802, opened 2017-03-13
Fix #2799: Don't use `equals` for comparing java.lang.Double/Float.

### LINKED ISSUE #2799: `==` between `j.l.{Float,Double}` gives incorrect results for zeros and NaNs.
Scala/JVM had the same issue. It was fixed in https://github.com/scala/scala/pull/5761 for 2.12.2+.

sjrd:
This is essentially a port of the fix for scala-dev#329, which was implemented in scala/scala#5761.

To remain bug-compatible with the JVM, we explicitly look at the version of the Scala compiler that we run on to determine whether or not the fix should be applied.

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
@@ -9,10 +9,9 @@ package org.scalajs.testsuite.compiler
 
 import scala.annotation.tailrec
 
-import scala.scalajs.js
-
 import org.junit.Test
 import org.junit.Assert._
+import org.junit.Assume._
 
 import org.scalajs.testsuite.utils.AssertThrows._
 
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @gzm0. Locally tested on 2.12.2-bin-3fadf69 (both JVM and JS).

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -435,6 +435,98 @@ class RegressionTest {
     assertEquals((Nil, 10), result)
   }
 
+  @Test def eqEqJLDouble(): Unit = {
+    // Taken from run/sd329.scala in scala/scala
+
+    val v = Platform.scalaVersion
+    assumeFalse("Bug scala-dev#329, #2799 on this version of Scala",
+        v.startsWith("2.10.") || v.startsWith("2.11.") ||
+        v == "2.12.0" || v == "2.12.1")
```
gzm0: Shouldn't we also add tests for the other case? I.e. when this assume is true? Seems simple enough and guards us against not being really bug compatible.

sjrd: Ah ah, indeed.

sjrd:
Updated. Re-tested locally with 2.12.2-bin-3fadf69.

sjrd:

PR #2807, opened 2017-03-15
Fix #2744: Restrict @JSExportStatic fields to be defined first.

### LINKED ISSUE #2744: Restrict @JSStatic fields to be defined *before* other fields
We should do this to be more in line with the [`@static` SIP](http://docs.scala-lang.org/sips/pending/static-members.html) for Scala.

sjrd:
More precisely, they must be defined before non-static fields, as well as before any constructor statement. This makes sure that there is no initialization code that can run before their own.

This makes `@JSExportStatic` consistent with the `@static` SIP (SIP-25).

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -401,6 +401,28 @@ abstract class PrepJSInterop extends plugins.PluginComponent
 
       case Template(parents, self, body) =>
         val clsSym = tree.symbol.owner
+
+        // Check that @JSExportStatic fields come first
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @gzm0.

sjrd:
Good question. @DarkDimius, is the following supposed to be valid under SIP-25 or not?
```scala
object Foo {
  lazy val a: Int = 1
  @staitc val b: Int = 2
}
```

DarkDimius:
@gzm0 @sjrd, nice example indeed.
It is valid. Lazy vals are treated like defs.
I guess I should clear it in the writeup.

sjrd:
OK thanks. I updated so that `lazy val`s are accepted before static fields.

Issue #2039, opened 2015-11-19
Integrate java.time implementation into Scala.js


soc:
Implementation here: https://github.com/soc/threetenbp

@ummels and @nicolasstucki, what do you think?

nicolasstucki:
@ummels @alonsodomin @japgolly @soc @sjrd 

We should consider cross compiling this implementation of `java.time` and drop the implementation from the core.

#### Pros
1. It might be ready faster.
2. Releases would not depend on the Scala.js releases and hence fixes/improvements would be more agile.
3. It would add support for `java.time` on JDK6, JDK7 and JDK8 on both the JVM and JS.

#### Cons
1. On Scala.js `java.time` would not be available out of the box. It would just need a library dependency.
2. If we want to cross compile we would need to port the existing test from that repo to JUnit. We could reuse the tests that we already have.

What are your thoughts about this? Any other pros or cons?

alonsodomin:
Thought this idea had been discarded since there was no activity on this ticket. I personally don't see a big issue having to add an additional dependency and working in @soc's repo we could focus more on the actual things that need to be rewritten from scratch, like the date & time formatting/parsing and the timezone data "compiler".

sjrd:
After some IRL discussion with @nicolasstucki, we've decided the following course of action in the immediate future.

We'll move the existing `java.time` implementation (the one already merged in this repo) to a *separate* repo, distinct from @soc's. That implementation will be published as 0.1 very soon after the release of 0.6.6. Developers will be required to add an explicit dependency to get this implementation, and it will only work when building on JDK8 (just like that in this repo would have).

As an immediate measure, it has the following advantages:
* It does not commit to `scalajs-library.jar` have an implementation of `java.time`, which will allow to move to another implementation (such as @soc's) without breaking backward compatibility.
* If we decide that it wants to live on on its own anyway, it can evolve more rapidly than scala-js/scala-js itself.
* We do not throw away the work that has been done far, and it will be readily available to 0.6.6 users.

soc:
Sounds reasonable! @alonsodomin I think the timezone data compiler is not that important, I wonder why it wasn't separated into a different library right from the beginning. It's more of a stand-alone utility, and no real part of the library. Agree on formatting/parsing! I'll revisit the current state of the code soon to figure out what the missing things are.

japgolly:
Sounds like a good idea. In the future when it's mature, we'll have the option and reintegrating it back into SJS if desirable. :+1:

[...]

PR #2809, opened 2017-03-16
Implement java.util.Base64.

mvh77:
Ok this should be gold now! :stuck_out_tongue:

### DIFF
--- a/javalib/src/main/scala/java/util/Base64.scala
+++ b/javalib/src/main/scala/java/util/Base64.scala
@@ -1,0 +1,587 @@
+package java.util
+
+import scala.annotation.tailrec
+
+import java.io._
+import java.nio.charset.StandardCharsets
+import java.nio.ByteBuffer
+
+object Base64 {
+
+  private val chars = ('A' to 'Z') ++ ('a' to 'z') ++ ('0' to '9')
+
+  private val basicEncodeTable: Array[Byte] =
+    (chars ++ Seq('+', '/')).map(_.toByte).toArray
+
+  private val urlSafeEncodeTable: Array[Byte] =
+    (chars ++ Seq('-', '_')).map(_.toByte).toArray
+
+  private def decodeTable(encode: Array[Byte]): Array[Int] = {
+    val decode = Array.fill[Int](256)(-1)
+    for ((b, i) <- encode.zipWithIndex)
+      decode(b) = i
+    decode('=') = -2
+    decode
+  }
+
+  private val basicDecodeTable = decodeTable(basicEncodeTable)
+  private val urlSafeDecodeTable = decodeTable(urlSafeEncodeTable)
+
+  private val mimeLineSeparators = Array[Byte]('\r', '\n')
+  private final val mimeLineLength = 76
+
+  private val basicEncoder =
+    new Encoder(basicEncodeTable)
+
+  private val basicDecoder =
+    new Decoder(basicDecodeTable, ignoreInvalid = false)
+
+  private val mimeEncoder =
+    new Encoder(basicEncodeTable, mimeLineLength, mimeLineSeparators)
+
+  private val mimeDecoder =
+    new Decoder(basicDecodeTable, ignoreInvalid = true)
+
+  private val urlSafeEncoder =
+    new Encoder(urlSafeEncodeTable)
+
+  private val urlSafeDecoder =
+    new Decoder(urlSafeDecodeTable, ignoreInvalid = false)
+
+  // --------------------------------------------------------------------------
+
+  def getEncoder(): Encoder = basicEncoder
+
+  def getUrlEncoder(): Encoder = urlSafeEncoder
+
+  def getMimeEncoder(): Encoder = mimeEncoder
+
+  def getMimeEncoder(lineLength: Int, lineSeparator: Array[Byte]): Encoder = {
+    for (b <- lineSeparator) {
+      if (basicDecodeTable(b & 0xff) != -1) {
+        throw new IllegalArgumentException(
+            "Illegal base64 line separator character 0x" + (b & 0xff).toHexString)
+      }
+    }
+    new Encoder(basicEncodeTable, lineLength / 4 * 4, lineSeparator)
+  }
+
+  def getDecoder(): Decoder = basicDecoder
+
+  def getUrlDecoder(): Decoder = urlSafeDecoder
+
+  def getMimeDecoder(): Decoder = mimeDecoder
+
+  // --------------------------------------------------------------------------
+
+  class Decoder private[Base64] (table: Array[Int], ignoreInvalid: Boolean) {
+
+    def decode(src: Array[Byte]): Array[Byte] = {
+      val dst = new Array[Byte](dstRequiredLength(src))
+      doDecode(new Wrapper(src), new Wrapper(dst))
+      dst
+    }
+
+    def decode(src: String): Array[Byte] =
+      decode(src.getBytes(StandardCharsets.ISO_8859_1))
+
+    def decode(src: Array[Byte], dst: Array[Byte]): Int = {
+      if (dst.length < dstMaxLength(src.length) && // dst is possibly too small
+          dst.length < dstRequiredLength(src)) {   // dst is actually too small
+        throw new IllegalArgumentException(
+            "Output byte array is too small for decoding all input bytes")
+      }
+
+      doDecode(new Wrapper(src), new Wrapper(dst))
+    }
+
+    def decode(buffer: ByteBuffer): ByteBuffer = {
+      val start = buffer.position()
+      try {
+        val src = new Array[Byte](buffer.remaining())
+        buffer.get(src)
+        val dst = new Array[Byte](dstRequiredLength(src))
+        val written = doDecode(new Wrapper(src), new Wrapper(dst))
+        ByteBuffer.wrap(dst, 0, written)
+      } catch {
+        case e: IllegalArgumentException =>
+          buffer.position(start)
+          throw e
+      }
+    }
+
+    def wrap(is: InputStream): InputStream =
+      new DecodingInputStream(is, table, ignoreInvalid)
+
+    // ------------------------------------------------------------------------
+    // PRIVATE
+    // ------------------------------------------------------------------------
+
+    private def doDecode(src: Wrapper, dst: Wrapper): Int = {
+      val srcBuffer = new Wrapper(new Array[Byte](4))
+
+      @inline
+      def inputData(): Unit = {
+        srcBuffer.position = 0
+        var shift = 18
+        var i = 0
+        while (srcBuffer.hasRemaining) {
+          i |= ((srcBuffer.get() & 0xff) << shift)
+          shift -= 6
+        }
+
+        if (shift == 12) {
+          throw new IllegalArgumentException(
+              "Last unit does not have enough valid bits")
+        }
+
+        if (shift <= 6)
+          dst.put((i >> 16).toByte)
+        if (shift <= 0)
+          dst.put((i >> 8).toByte)
+        if (shift <= -6)
+          dst.put(i.toByte)
+        srcBuffer.clear()
+      }
+
+      @tailrec
+      def iterate(): Unit = {
+        if (src.hasRemaining) {
+          if (srcBuffer.hasRemaining) {
+            val int = src.get() & 0xff
+            table(int) match {
+              case -2 =>
+
+              case -1 =>
+                if (!ignoreInvalid) {
+                  throw new IllegalArgumentException(
+                      "Illegal base64 character " + int.toHexString)
+                }
+                iterate()
+
+              case i =>
+                srcBuffer.put(i.toByte)
+                iterate()
+            }
+          } else {
+            inputData()
+            iterate()
+          }
+        }
+      }
+
+      iterate()
+
+      // end or padding
+      srcBuffer.flip()
+      inputData()
+      while (src.hasRemaining) {
+        val int = src.get() & 0xff
+        val value = table(int)
+        if (value != -2 && (!ignoreInvalid || value > 0)) {
+          throw new IllegalArgumentException(
+              s"Input byte array has incorrect ending byte at $int")
+        }
+      }
+
+      dst.position
+    }
+
+    private def dstRequiredLength(src: Array[Byte]): Int = {
+      var validBytes = 0
+
+      if (ignoreInvalid) {
+        for (i <- src.indices) {
+          if (table(src(i) & 0xff) >= 0)
+            validBytes += 1
+        }
+      } else {
+        /* We check the end for padding and compute the length from there.
+         * This is ok, if the rest contains garbage we'll have written
+         * something before throwing but the spec says "If the input byte array
+         * is not in valid Base64 encoding scheme then some bytes may have been
+         * written to the output byte array before IllegalArgumentException is
+         * thrown."
+         */
+        validBytes = src.length
+        if (src.length >= 1 && src.last == '=') {
+          validBytes -= 1
+          if (src.length >= 2 && src(src.length - 2) == '=')
+            validBytes -= 1
+        }
+
+        if (src.length >= 1 && validBytes == 0) {
+          throw new IllegalArgumentException(
+              "Input byte array has wrong 4-byte ending unit")
+        }
+      }
+
+      dstMaxLength(validBytes)
+    }
+
+    /** Computes the destination length solely based on the source length,
+     *  without knowing about padding.
+     */
+    private def dstMaxLength(srcLength: Int): Int =
+      (srcLength + 3) / 4 * 3 - (if (srcLength % 4 == 0) 0 else 4 - (srcLength % 4))
+
+  }
+
+  private object DecodingInputStream {
+    private final val DecodeState18 = 0
+    private final val DecodeState12 = 1
+    private final val DecodeState14 = 2
+    private final val DecodeState16 = 3
+  }
+
+  private class DecodingInputStream(in: InputStream, table: Array[Int],
+      ignoreInvalid: Boolean)
+      extends FilterInputStream(in) {
+
+    import DecodingInputStream._
+
+    private val oneBuf = new Array[Byte](1)
+
+    private var closed = false
+    private var eof = false
+    private var out = 0
+    private var shift = DecodeState18
+
+    override def read(): Int =
+      if (read(oneBuf, 0, 1) == -1) -1
+      else oneBuf(0) & 0xff
+
+    override def read(b: Array[Byte], off: Int, len: Int): Int = {
+      var written = 0
+
+      @inline
+      def writeValue(i: Int): Int = {
+        /* Max value means we're writing remaining bytes after EOF, no table
+         * lookup.
+         */
+        if (i == Int.MaxValue) {
+          0
+        } else {
+          table(i) match {
+            case -1 =>
+              if (!ignoreInvalid) {
+                throw new IOException(
+                    "Illegal base64 character " + i.toHexString)
+              }
+              0
+
+            case v =>
+              shift match {
+                case DecodeState18 =>
+                  out |= v << 18
+                  shift = DecodeState12
+                  0
+
+                case DecodeState12 =>
+                  out |= v << 12
+                  b(off + written) = (out >> 16).toByte
+                  out <<= 8
+                  shift = DecodeState14
+                  1
+
+                case DecodeState14 =>
+                  out |= v << 14
+                  b(off + written) = (out >> 16).toByte
+                  out <<= 8
+                  shift = DecodeState16
+                  1
+
+                case DecodeState16 =>
+                  out |= v << 16
+                  b(off + written) = (out >> 16).toByte
+                  out = 0
+                  shift = DecodeState18
+                  1
+              }
+          }
+        }
+      }
+
+      @inline
+      def endOfFile(): Int = {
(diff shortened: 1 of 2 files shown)


sjrd:

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+
+  private class EncodingOutputStream(out: OutputStream,
+                                     table: ByteTable,
+                                     ll: Int,
+                                     ls: Array[Byte],
+                                     withPadding: Boolean) extends OutputStream {
+
+    override def write(b: Int): Unit = write(Array((b & 0xff).toByte), 0, 1)
+
+    private val inputBuf = ByteBuffer.allocate(3)
```
sjrd: As elsewhere, use an `Array[Byte]` here (with an `inputBufPosition: Int` to remember where we are in the input buffer).

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+
+  // --------------------------------------------------------------------------
+
+  private class EncodingOutputStream(out: OutputStream,
+                                     table: ByteTable,
+                                     ll: Int,
+                                     ls: Array[Byte],
+                                     withPadding: Boolean) extends OutputStream {
+
+    override def write(b: Int): Unit = write(Array((b & 0xff).toByte), 0, 1)
```
sjrd: `(b & 0xff).toByte` -> `b.toByte`.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+      withoutEndLines + (if (lineLength > 0) ((withoutEndLines-1) / lineLength) * lineSeparator.length else 0)
+    }
+  }
+
+  // --------------------------------------------------------------------------
+
+  private class EncodingOutputStream(out: OutputStream,
+                                     table: ByteTable,
+                                     ll: Int,
+                                     ls: Array[Byte],
```
sjrd: Give descriptive names to `ll` and `ls`, e.g., `lineLength` and `lineSeparator`.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+                                     withPadding: Boolean) extends OutputStream {
```
sjrd: It's probably easier if you extend `FilterOutputStream`. It gives you some correct default implementations, such as `flush()`.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+          throw new IllegalArgumentException(s"Input byte array has incorrect ending byte at $int")
+      }
+      dst.position()
+    }
+
+    private def dstMaxLength(srcLength: Int): Int = (srcLength+3) / 4 * 3
+
+  }
+
+  private class DecodingInputStream(in: InputStream, table: IntTable, ignoreInvalid: Boolean) extends InputStream {
```
sjrd: I recommend you extend `FilterInputStream`. It gives you some correct default implementations.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/Base64Test.scala
```diff
@@ -0,0 +1,532 @@
...
+      out.write(in(0))
+      out.write(in, 1, in.length - 1)
+      out.close()
+      val rlt = new String(baos.toByteArray, StandardCharsets.ISO_8859_1)
+      assertEquals(s"output stream result doesn't match for: $name", exp, rlt)
+    }
+  }
+
+  // @Test // Fails on the JVM (bug JDK-8176379)
+  def encodeOutputStreamFailsOnJVM(): Unit = {
```
sjrd: If a test fails on the JVM, you should not disable the test. Instead, you should add the following at the beginning of the test:
```scala
assumeFalse("JDK bug JDK-8176379", Platform.executingInJVM)
```
with the following relevant imports at the top of the file:
```scala
import org.junit.Assume._

import org.scalajs.testsuite.utils.Platform
```

mvh77:
I actually read most of the coding style and thought "Gee, I'll need a degree in fashion before I can contribute". :wink:

sjrd:
We are looking into introducing [scalafmt](http://scalameta.org/scalafmt/) to our codebase to alleviate that issue, but we're not there yet.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,390 @@
...
+  def getUrlDecoder: Decoder = urlSafeDecoder
+
+  def getUrlEncoder: Encoder = urlSafeEncoder
+
+  // --------------------------------------------------------------------------
+
+  class Decoder private[Base64](table: IntTable, ignoreInvalid: Boolean) {
+
+    def decode(src: Array[Byte]): Array[Byte] = {
+      val dst = ByteBuffer.allocate(dstMaxLength(src.length))
```
mvh77: Why does manipulating a byte buffer have significant costs, isn't it just a wrapper around an array with a few ints to keep track of position, remaining etc..?

I agree that for the method that takes both the src and the dst array as arguments the allocation is unnecessary, but then we'll have to do tricks to find out if the dst array is large enough to accomodate for everything, we'll have to either traverse the src twice, once for the check (as the src can contain all sorts of garbage that is legal in the case of mime), another time for the decoding. Or we can just start decoding and throw if we run out of space in the dst. Both are inefficient in a way.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/Base64Test.scala
mvh77: cool :+1:

javalib/src/main/scala/java/util/Base64.scala
sjrd: `ByteBuffer` is inefficient because, on every `get()`, it has to check whether `position()` has got out of range and throw a `BufferUnderflowException` (or `BufferOverflowException` for `put()`). Using an `Array` and an explicit `index` avoids all of that.

> but then we'll have to do tricks to find out if the dst array is large enough to accomodate for everything

There's no reason you'd have to do anything different from what you do with the `ByteBuffer`. For decoding, it's true that the doc says

> No bytes will be be written to the output byte array if the output byte array is not big enough.

which means you have to either allocate some temporary storage, or traverse the source twice. I would recommend to traverse the source twice (the first pass doesn't have to actually *decode* anything; it only needs to count the number of valid input bytes) so that no allocation is necessary. It will still be faster than using `ByteBuffer`s everywhere.

sjrd: It's also possible to bypass the first checking pass if the output array is *definitely* big enough. You can test this by assuming that no input characters will be ignored. This will likely be a common case, since a caller would dimension their input arrays so that decoding always succeeds, which means so that it is definitely big enough.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -387,4 +453,34 @@ object Base64 {
...
+    def clear(): Unit = {
+      position = 0
+      length = array.length
+    }
+
+    def flip(): Unit = {
+      length = position
+      position = 0
+    }
+  }
```
mvh77: Ok so maybe we'll have to agree to disagree on what's readable, but I find it better if the index/position manipulation operations on arrays are hidden in a wrapper like this, especially since a syntax like array(index++) is not possible. I hope the @specialized stuff makes sense, let me know if that's pointless.

sjrd:
Could you rebase your PR on top *of the 0.6.x branch*, please?

mvh77:
Done

sjrd:
I think you confused "merge" and "rebase" ;)

mvh77:
Sorry I'm a Hg user by day, not that great with git. Isn't the end result the same in this case?

mvh77:
How about now? This is getting messy... :stuck_out_tongue:

mvh77:
Ok I talked to some GIT people, apparently this is serious stuff! 😄 I'll rework this, do you want the commits to be squashed as well?

sjrd:
Yes, eventually we'll ask that you squash your commits.

If you struggle too much with git, I can take care of the git magic and push to your branch (assuming you've checked the box "Allow edits from maintainers").

mvh77:
I'll try to figure it out later today, I have to learn this eventually anyway.

mvh77:
Ok how about now? I had to use the --force, I don't know if it can be done without. :smile:

sjrd:
The git part worked out :) I don't know yet when I'll have the time to get second look at your PR, so please be a bit patient for a few days.

sjrd:
Jenkins, retest this please

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+      } else {
+        validBytes = length-offset
+        /* We check the end for padding and compute the length from there. This is ok, if the rest contains garbage
+        we'll have written something before throwing but the spec says "If the input byte array is not in valid Base64
+        encoding scheme then some bytes may have been written to the output byte array before IllegalArgumentException
+        is thrown."*/
+        if (src(length-1) == '=') {
+          padding += 1
+          if (src(length-2) == '=')
+          padding += 1
```
sjrd: Consider doing `validBytes -= 1` instead of `padding += 1`, and get rid of the `padding` variable altogether as well as the ` - (if (validBytes % 4 == 0) padding else 4 - (validBytes % 4))` part of the final result.

sjrd: Also, I think this can cause an `IndexOutOfBoundsException` if the input buffer contains exactly `"="`. If you eliminate the `if (length == 0)` at the beginning of the method, it can also crash when the input buffer is empty. You should guard against those scenarios, e.g.,
```scala
if (length >= 1 && src(length - 1) == '=')
```
(there should be tests for those scenarios)

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+    private var out = 0
+    private var shift = DecodeState18
+
+    override def read(): Int = if (read(oneBuf, 0, 1) == -1) -1 else oneBuf(0) & 0xff
+
+    override def read(b: Array[Byte], off: Int, len: Int): Int = {
+      // scalastyle:off return
+      if (closed) throw new IOException("Stream is closed")
+      if (off < 0 || len < 0 || len > b.length - off) throw new IndexOutOfBoundsException
+      if (eof) return -1
```
sjrd: Consider moving this after the inner `def`s, as this will allow you to get rid of the `return`:
```scala
if (eof) {
  -1
} else {
  iterate()
  written
}
```
`return`s are *exceedingly* rare in this codebase. We have only 17 `return`s out of 120 kLoC. We only use them when we have no other choice.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+          writeValue(Integer.MAX_VALUE)
+      }
+
+      def iterate(): Unit = {
+        // scalastyle:off return
+        while (written < len)
+          in.read() match {
+            case -1 =>
+              written += endOfFile()
+              return
```
sjrd: You can get rid of this `return` by turning `iterate()` into a tail-recursive method:
```scala
@tailrec
def iterate(): Unit = {
  if (written < len) {
    in.read() match {
      case -1 =>
        written += endOfFile()
      case '=' =>
        written += padding()
        iterate()
      case int =>
        written += writeValue(int)
        iterate()
    }
  }
}
```

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+
+    def encodeToString(src: Array[Byte]): String = new String(encode(src), StandardCharsets.ISO_8859_1)
+
+    def wrap(os: OutputStream): OutputStream =
+      new EncodingOutputStream(os, table, lineLength, lineSeparator, withPadding)
+
+    def encode(buffer: ByteBuffer): ByteBuffer = {
+      val result = new Array[Byte](dstLength(buffer.remaining()))
+      val written =
+        doEncode(new Wrapper(buffer.array(), buffer.position, buffer.limit()), new Wrapper(result))
```
sjrd: Same comment about using `array()`, obviously. tl;dr, I recommend not to. Read into a temporary `Array[Byte]` instead.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    override def write(bytes: Array[Byte], off: Int, len: Int): Unit = {
+      // scalastyle:off return
+      if (closed)
+        throw new IOException("Stream is closed")
+      else if ((off < 0) || (len < 0) || (off > bytes.length) || ((off + len) < 0) || ((off + len) > bytes.length))
+        throw new IndexOutOfBoundsException
+      else if (len == 0)
+        return
```
sjrd: You can easily get rid of that `return` by doing `if (len != 0)` and putting the rest of the method in that if.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+      }
+
+      @inline
+      def iterate(): Unit = {
+        // scalastyle:off return
+        while (src.hasRemaining) {
+          if (srcBuffer.hasRemaining) {
+            val int = src.get & 0xff
+            table(int) match {
+              case -2 => return
```
sjrd: You can get rid of this `return` by turning `iterate()` into a tail-recursive method.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    def wrap(is: InputStream): InputStream = new DecodingInputStream(is, table, ignoreInvalid)
+
+    // ------------------------------------------------------------------------
+    // PRIVATE
+    // ------------------------------------------------------------------------
+
+    private def doDecode(src: Wrapper[Byte], dst: Wrapper[Byte]): Int = {
+      val srcBuffer = new Wrapper[Int](new Array[Int](4))
```
sjrd: This is the only `Wrapper[Int]` that you have. All the other ones are `Wrapper[Byte]`. You should also make this one a `Wrapper[Byte]`, and then you can actually hard-code `Wrapper` for `Byte`, eliminating the need for `@specialized`.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+
+      @inline
+      def inputData(): Unit = {
+        srcBuffer.position = 0
+        var shift = 18
+        var i = 0
+        while(srcBuffer.hasRemaining) {
+          i |= (srcBuffer.get << shift)
```
sjrd: If `srcBuffer` is a `Wrapper[Byte]`, this needs to be `i |= ((srcBuffer.get() & 0xff) << shift)`.

sjrd:
About coding style, would you rather me to

a. Point out in review comments one occurrence of each deviation wrt. our coding style, and you apply them everywhere
b. Point out in review comments *every* deviation (that can be overwhelming)
c. Fix myself the coding style myself, once everything else is good, and force-push on your branch

?

Usually people really don't like my comments on coding style, because it feels like I am being unreasonably strict. That's why I offer to fix them myself. It is also what takes the least time for me, actually.

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,486 @@
...
+      else if (dst.length < dstRequiredLength(src, 0, src.length))
+        throw new IllegalArgumentException("Output byte array is too small for decoding all input bytes")
+      else
+        doDecode(new Wrapper(src), new Wrapper(dst))
+    }
+
+    def decode(buffer: ByteBuffer): ByteBuffer = {
+      val start = buffer.position()
+      try {
+        val src = buffer.array()
```
mvh77: Ok, I'll do the temporary array. I thought unnecessary copying wouldn't get past you but if you say so. :) Enlighten me please: there are no "direct" buffers in JS right? So aren't all ByteBuffers backed by an array (buffer.hasArray() == true) in JS? Looking at the code I'm somewhat confused, I can't find an implementation for Buffer._array that's used in hasArray and IntelliJ is all red in all buffer impls.

sjrd: `TypedArrayByteBuffer` is an implementation of `ByteBuffer` that is direct, and is not backed by an array, so `hasArray` returns false.

The implementation of `nio` buffers in Scala.js is highly typeclass-based. I'm not surprised that IntelliJ has trouble with it.

javalib/src/main/scala/java/util/Base64.scala
mvh77: Ok I have a problem here, what do you think should be the result for decoding "=" with a Mime decoder? I think it should be an empty buffer, not an IllegalArgumentException like the JDK throws. I've filed a bug about it [here](https://bugs.openjdk.java.net/browse/JDK-8176043) that was rejected on flimsy grounds, that last comment there is mine. Should we follow the RFCs or replicate the buggyish implementation on the JDK?

sjrd: I don't know. I think they're right that this is not a bug. The docs at [`Base64`](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) for MIME say:

> Uses the "The Base64 Alphabet" **as specified in Table 1 of RFC 2045** for encoding and decoding operation. The encoded output must be represented in lines of no more than 76 characters each and uses a carriage return '\r' followed immediately by a linefeed '\n' as the line separator. No line separator is added to the end of the encoded output. All line separators or **other characters not found in the base64 alphabet table** are ignored in decoding operation.

Note that the character `=` *is* present in Table 1 of [RFC 2045](https://www.ietf.org/rfc/rfc2045.txt). As such, it is my understanding that it should *not* be ignored. But then, it's also true that `"TQ===*"` should be rejected as much as `"TQ=*"`.

Under such inconsistent behavior on the JDK's part, and unless and until the OpenJDK folk acknowledge some bug, I would suggest that

* we follow the JDK behavior if the implementation does not become too convoluted, or
* we accept more inputs than the JDK otherwise.

mvh77: Well, it's definitely inconsistent behavior and against the spirit of the RFC. I suggest we accept more input in these cases if it's ok with you. [RFC4648](https://www.ietf.org/rfc/rfc4648.txt) §3.3 says:

> [...] Such specifications may instead state, as MIME does, that characters outside the base encoding alphabet should   simply be ignored when interpreting data ("be liberal in what you accept").  Note that this means that any adjacent carriage return/line feed (CRLF) characters constitute "non-alphabet characters" and are ignored.  **Furthermore, such specifications MAY ignore the pad character, "=", treating it as non-alphabet data, if it is present before the end of the encoded data. If more than the allowed number of pad characters is found at the end of the string (e.g., a base 64 string terminated with "==="), the excess pad characters MAY also be ignored.**

javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -145,35 +156,34 @@ object Base64 {
...
       }
-      dstMaxLength(validBytes) - (if (validBytes % 4 == 0) padding else 4 - (validBytes % 4))
+      dstMaxLength(validBytes)
     }
 
-    /** compute the destination length solely based on the source length length without knowing about padding */
-    private def dstMaxLength(srcLength: Int): Int = (srcLength+3) / 4 * 3
+    /** compute the destination length solely based on the source length without knowing about padding */
+    private def dstMaxLength(srcLength: Int): Int =
+      (srcLength + 3) / 4 * 3 - (if (srcLength % 4 == 0) 0 else 4 - (srcLength % 4))
```
mvh77: I moved the `- (if (srcLength ....` stuff here but I see no way of getting rid of that since it won't work for lengths that are not multiples of 4.

mvh77:
And when it comes to coding style, you are unreasonably strict :wink:, so let's go with option C!

javalib/src/main/scala/java/util/Base64.scala
sjrd: Ah, you are right, of course.

javalib/src/main/scala/java/util/Base64.scala
sjrd: OK let's do that.

sjrd:
Jenkins, retest this please

sjrd:
Jenkins, retest this please

sjrd:
Jenkins, retest this please

mvh77:
Cool! Hey let me ask you one thing: what's the situation with java.time? I'd like to work on that but it seems the threeten backport is fully ported to Scala [here](https://github.com/cquiroz/scala-java-time) (I'm sure you know this) and there's even a port of java.util.Locale [here](https://github.com/cquiroz/scala-java-locales). Is it just a lack of manpower or what's preventing these from making it to the "standard" library?

sjrd:
I believe the following issues address your question: https://github.com/scala-js/scala-js/issues/1618 and https://github.com/scala-js/scala-js/issues/2039. And if they don't, I suggest you ask further questions on this topic on the relevant issues.

PR #2810, opened 2017-03-16
Disallow exports of lazy vals to the top-level or as static.

sjrd:
Exporting a lazy val to the top level does not make sense, because they must be evaluated at module loading time. In fact, the generated code was buggy.

Exporting a lazy val as static probably makes sense, but as SIP-25 on `@static` considers static lazy vals as an open question, it is safer to disallow them for now. This can be reconsidered in the future.


sjrd:
Review by @gzm0. It turns out this also needs to go into 0.6.15. The comment in #2807 about `lazy val`s made me realize they were not tested for `@JSExportTopLevel` nor `@JSExportStatic`. It turned out both must be disallowed.

Issue #2795, opened 2017-03-07
Debugging problems


Gralfim:
Even with a very basic source, we can clearly see problems with debugging (mapping lines in Scala sources):
1. during the function tracing, the cursor is switching to the beginning of the function
2. _Step Over_ often moves the cursor into the called function

This behavior makes tracing of complicated functions very difficult.

**Repro:**
In the following source, set the breakpoint to the first line of function DebugSample,main, launch it and try to use _Step Over_ to trace the function. See where the cursor will appear.

```
import scala.scalajs.js.JSApp
import scala.util.Random

object DebugSample extends JSApp {
  def main(): Unit = {
    val angle = Random.nextDouble() * 90.0
    val rad = angle.toRadians
    println(s"Random angle $rad radians")

    val angle2 = Random.nextDouble() * 45.0
    val rad2 = angle2.toRadians
    println(s"Random angle $rad2 radians")
  }
}

```

sjrd:
When using step-by-step execution of Scala.js, it is recommended to disable the optimizer with:
```scala
scalaJSOptimizerOptions ~= { _.withDisableOptimizer(true) }
```

OndrejSpanel:
I have tried this option. While it prevents stepping into the functions, still there is the issue of the current execution point jumping between real current location and the beginning of the `main` function on almost each step.

![image](https://cloud.githubusercontent.com/assets/6927223/23673857/1b779946-0374-11e7-9f64-25a6865a12a2.png)
![image](https://cloud.githubusercontent.com/assets/6927223/23673873/284ba130-0374-11e7-9634-58ee4ad63220.png)
![image](https://cloud.githubusercontent.com/assets/6927223/23673883/3154829c-0374-11e7-864c-5dc2a50790ad.png)

The screenshots are with IntelliJ IDE, when debugging with the Chrome Dev tools the results are even worse, it is not possible even to place breakpoints on most lines and stepping through is very erratic.

Are there any tools for inspecting or visualizing or otherwise debugging source maps? The behaviour seems surprisingly annoying for such small code sample.

PR #2819, opened 2017-03-20
Fix #2821: Adapt column pos in source maps for printEscapeJS

### LINKED ISSUE #2821: Source Maps bug - wrong column in the generated file
There seems to be a bug in the class `JSTreePrinterWithSourceMap` causing that the source maps does not exactly match the generated file.
The class has two purposes:
1. it generates the output javascript code using its base class `JSTreePrinter`
2. it maintains the position in the javascript (in `private var column`) and uses it to generate source map (using `sourceMap: SourceMapWriter`)
To serve well for both these functions, there is necessary to keep `var column` synchronized with the position in `out` stream (see `override protected def print(s: String): Unit`). However, in the function `override protected def print(ident: Ident): Unit` (and one similar place) there is some writing to `out` without change of `var column`.

Gralfim:
When investigating issue #2795, we found that there are situations when javascript code was written, but column field used by the source map writer to point where in javascript you are was not updated.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Utils.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Utils.scala
@@ -75,12 +75,13 @@ object Utils {
     sb.toString
   }
 
-  def printEscapeJS(str: String, out: java.lang.Appendable): Unit = {
+  def printEscapeJS(str: String, out: java.lang.Appendable): Int = {
     /* Note that Java and JavaScript happen to use the same encoding for
      * Unicode, namely UTF-16, which means that 1 char from Java always equals
      * 1 char in JavaScript. */
     val end = str.length()
     var i = 0
+    var writtenChars = 0
     /* Loop prints all consecutive ASCII printable characters starting
      * from current i and one non ASCII printable character (if it exists).
      * The new i is set at the end of the appended characters.
@@ -95,8 +96,10 @@ object Utils {
           c = str.charAt(i)
       }
       // Print ASCII printable characters from `start`
-      if (start != i)
+      if (start != i) {
         out.append(str, start, i)
+        writtenChars += i
+      }
 
       // Print next non ASCII printable character
       if (i != end) {
@@ -104,18 +107,23 @@ object Utils {
           if (6 < c && c < 14) {
             val i = 2 * (c - 7)
             out.append(EscapeJSChars, i, i + 2)
+            writtenChars += 2
           } else if (c == 34) {
             out.append(EscapeJSChars, 14, 16)
+            writtenChars += 2
           } else if (c == 92) {
             out.append(EscapeJSChars, 16, 18)
+            writtenChars += 2
           } else {
             out.append(f"\\u$c%04x")
+            writtenChars += 6
           }
         }
         escapeJSEncoded(c)
         i += 1
       }
     }
+    writtenChars
   }
 
   /** A ByteArrayOutput stream that allows to jump back to a given
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -3,6 +3,9 @@ import com.typesafe.tools.mima.core.ProblemFilters._
 
 object BinaryIncompatibilities {
   val IR = Seq(
+      // Breaking: Utils.printEscapeJS now returns Int
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.core.ir.Utils.printEscapeJS")
   )
 
   val Tools = Seq(
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
@@ -22,7 +22,6 @@ import org.scalajs.core.ir
 import ir.Position
 import ir.Position.NoPosition
 import ir.Printers.IndentationManager
-import ir.Utils.printEscapeJS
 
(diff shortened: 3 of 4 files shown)


gzm0:
@Gralfim thanks for the contribution.
Could you please [sign the CLA](http://www.lightbend.com/contribute/cla/scala)? We cannot look at your code before that.

Gralfim:
Signed.

sjrd:
Thanks. Could you also open an issue that describes the problem, so that you can reference it in your commit message? This is useful to refer to closed issue when, e.g., duplicates come up.

Also, administrative detail: since we're in the middle of the release process, I'll delay running the CI on your PR until we have a Towards 0.6.16 commit. Otherwise our CI gets confused.

Gralfim:
We have added commit message referencing to the original issue (#2795). Is it sufficient?

gzm0:
> We have added commit message referencing to the original issue (#2795). Is it sufficient?

It is probably neater to open a new issue describing the exact problem with the source map writer. #2795 is more of a catch-all "source maps are not super easy to use" issue.

Also, note the [guidelines on commit messages][1] (TL;DR: Start with "Fix #1234: ..." if you are fixing something).

[1]: https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#first-line-of-the-commit-message

Gralfim:
Commit message format fixed, pointing to the new issue #2821.

sjrd:
@Gralfim We're done with our release process for 0.6.15. Since now we have 2 main branches (master for 1.x, and 0.6.x for 0.6.x), could you rebase the branch of this PR on top of the 0.6.x branch, please? I have already updated the PR itself so that it aims to be merged in the 0.6.x branch.

sjrd:
The [introduction of `printEscapeJS`](https://github.com/scala-js/scala-js/commit/b5200df4861ff761c4ffd1cba6e2ec1329e9d6a4) was part of [a series of PRs](https://github.com/scala-js/scala-js/issues/1926) by @nicolasstucki aimed at reducing allocations and improving speed in the emitter (a performance-sensitive part). It would be good not to annihilate those gains.

I suggest declaring a `protected def printEscapeJS` in `JSTreePrinter`, called by the two locations that you modified. In `JSTreePrinter` it can simply call `Utils.printEscapeJS`. Overridden in `JSTreePrinterWithSourceMap`, it won't be that easy. Maybe we need to duplicate some of the code from `Utils` :s

Gralfim:
I tried to re-implement the fix based on your suggestion, but tried to keep it DRY. Please check if it is acceptable for you now.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala
```diff
@@ -555,8 +554,11 @@ object Printers {
       }
     }
 
+    protected def printEscapeJS(s: String): Int =
```
sjrd: It seems that this one can return `Unit`

ir/src/main/scala/org/scalajs/core/ir/Utils.scala
```diff
@@ -75,12 +75,13 @@ object Utils {
     sb.toString
   }
 
-  def printEscapeJS(str: String, out: java.lang.Appendable): Unit = {
+  def printEscapeJS(str: String, out: java.lang.Appendable): Int = {
```
sjrd: I like your solution a lot, it's a lot DRYer than I had thought possible.

However you'll have some difficulty with our binary compat checker, because this method is not binary-compatible with its previous incarnation. You can run it locally with
```
> ir/mimaReportBinaryIssues
```
I think we'll just have to break bin-compat of the tools, here. You can whitelist this method in `project/BinaryIncompatibilities.scala`. See for example [here](https://github.com/scala-js/scala-js/blob/v0.6.15/project/BinaryIncompatibilities.scala#L5-L18) how we do it, and the formatting/comments we use.

sjrd:
Jenkins, test this please

Gralfim:
Requested changes (binary incompatibilities, return type in Printers.scala) committed. Please, check it once more.

project/BinaryIncompatibilities.scala
```diff
@@ -3,6 +3,9 @@ import com.typesafe.tools.mima.core.ProblemFilters._
 
 object BinaryIncompatibilities {
   val IR = Seq(
+    // Breaking: Utils.printEscapeJS now returns Int
+    ProblemFilters.exclude[IncompatibleResultTypeProblem](
+      "org.scalajs.core.ir.Utils.printEscapeJS")
   )
```
sjrd: According to our coding style, the indentation inside `()` is 4 spaces, not 2. So to be pedantic (because we are), this should be:
```scala
  val IR = Seq(
      // Breaking: Utils.printEscapeJS now returns Int
      ProblemFilters.exclude[IncompatibleResultTypeProblem](
          "org.scalajs.core.ir.Utils.printEscapeJS")
  )
```

Gralfim:
Indentation fixed, commit message updated.

sjrd:
Jenkins, retest this please

sjrd:
Sorry for the long delay in replying here.
It seems this PR has tipped the bootstrap test just above its allocated memory. Could you increase the available memory to 3072 [in `ci/matrix.xml`](https://github.com/scala-js/scala-js/blob/385487abda38fc8d17447e7b13a399341b2f2a0b/ci/matrix.xml#L206-L210)? (in the same commit)

sjrd:
Jenkins, retest this please

PR #2825, opened 2017-03-22
Fix #2823: Remove support for Rhino (and consequences)

### LINKED ISSUE #2823: Remove support for Rhino
Rhino was deprecated in Scala.js 0.6.13 because it has an insane amount of influence on our internal design, and is taxing our maintenance costs.

sjrd:
* Remove `RhinoJSEnv` itself, and the hooks in the linker that supported it
* Remove `javascript.Printers.ReverseSourceMapPrinter`, which was only used by `RhinoJSEnv`
* Remove the `ECMAScript51Global` output mode, which was only supported for Rhino
* Rewrite `scalajsenv.js` with the `$`-prefixed style of variables, instead of the `ScalaJS.xyz` style which was required to support `ECMAscript51Global`


sjrd:
Let's see what Jenkins has to say about this.

sjrd:
Only 1h36 of wall clock time for the CI. That's almost a 2x improvement over the average PR build lately :P

sjrd:
Review by @gzm0 

After this PR, `git grep -i rhino` says that there are only two places we still reference Rhino:
* [`StackTrace.isRhino`](https://github.com/sjrd/scala-js/blob/9145768d74a0615a3d45a4030db127d6e1126c98/library/src/main/scala/scala/scalajs/runtime/StackTrace.scala#L65-L79), where we still detect Rhino/Nashorn just in case someone pulls up some support for them
* The Closure linker backend, because it uses the ASTs in the package `com.google.javascript.rhino._`, but that's only because it's the IR used by GCC

And `git grep -i ECMAScript51Global` does not find anything anymore.

library/src/main/scala/scala/scalajs/js/timers/package.scala
```diff
@@ -14,10 +14,11 @@ import scala.concurrent.duration.FiniteDuration
 /**
  *  <span class="badge badge-non-std" style="float: right;">Non-Standard</span>
  *  Non-standard, but in general well supported methods to schedule asynchronous
- *  exeuction.
+ *  execution.
  *
  *  The methods in this package work in all JavaScript virtual machines
- *  supported by Scala.js (currently Rhino, Node.js and PhantomJS).
+ *  supported by Scala.js (currently Node.js and PhantomJS in the core, plus
+ *  Selenium in an external repository), as well as all browsers.
```
gzm0: Selenium is not really a JavaScript virtual machine. The underlying browser is (which arguably *must* support this, because it's part of the DOM).

I would suggest we reword this to say something like:
"The methods in this package work in all JavaScript virtual machines supporting `setTimeout` and `setInterval`."

partest-suite/src/test/resources/scala/tools/partest/scalajs/2.11.6/BlacklistedTests.txt
```diff
@@ -31,6 +31,9 @@ neg/t6289
 # RUN
 #
 
+# Too slow, really too slow
+run/bridges.scala
+
```
gzm0: There is a section below "using too much time". Put it there?

library/src/main/scala/scala/scalajs/js/timers/package.scala
sjrd: Ah yes, much better!

partest-suite/src/test/resources/scala/tools/partest/scalajs/2.11.6/BlacklistedTests.txt
sjrd: Oh there is? I was looking for "slow" and things like that, didn't find anything ^^ I'll move it.

sjrd:
Updated

PR #2831, opened 2017-03-23
Fix #2829: Make Scala.js defined default

### LINKED ISSUE #2829: Deprecate @ScalaJSDefined, make it default

gzm0:


gzm0:
CI only for now.

sjrd:
Congratulations! You managed to have only a single obscure configuration fail, while everything else succeeds: 2.11.x with `-Xexperimental` :P

Hint: that's definitely related to the experimental SAM support of 2.11.x.

gzm0:
> Hint: that's definitely related to the experimental SAM support of 2.11.x.

Yeah. I'm pretty sure the problem is that SAMs for `js.FunctionN` now are regarded as Scala.js defined classes.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5305,14 +5305,14 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-    !sym.isTrait && sym.hasAnnotation(ScalaJSDefinedAnnotation)
+    !sym.isTrait && isRawJSType(sym.tpe) && !sym.hasAnnotation(JSNativeAnnotation)
 
   def isScalaJSDefinedAnonJSClass(sym: Symbol): Boolean =
     sym.hasAnnotation(SJSDefinedAnonymousClassAnnotation)
 
   /** Tests whether the given class is a JS native class. */
   private def isJSNativeClass(sym: Symbol): Boolean =
-    isRawJSType(sym.tpe) && !isScalaJSDefinedJSClass(sym)
+    isRawJSType(sym.tpe) && sym.hasAnnotation(JSNativeAnnotation)
```
sjrd: I think that `@js.native` implies `@RawJSType`, so this could be simplified as only `sym.hasAnnotation(JSNativeAnnotation)`.

library/src/main/scala/scala/scalajs/js/annotation/ScalaJSDefined.scala
```diff
@@ -14,4 +14,5 @@ package scala.scalajs.js.annotation
  *  This annotation may only be used on a class extending
  *  [[scala.scalajs.js.Any js.Any]].
  */
+@deprecated("Scala.js defined is now the default. This annotation has no effect.", "1.0.0")
 class ScalaJSDefined extends scala.annotation.StaticAnnotation
```
sjrd: Consider adding "and can be safely removed".

sjrd:
After this, it will become increasingly awkward to talk about "Scala.js-defined JS types" in user-visible messages/docs/etc. I believe we should find a better name for those. My best bet so far is to simply talk about "non-native JS types", or simply "JS types" when there is no ambiguity.

gzm0:
> After this, it will become increasingly awkward to talk about "Scala.js-defined JS types" in user-visible messages/docs/etc. I believe we should find a better name for those. My best bet so far is to simply talk about "non-native JS types", or simply "JS types" when there is no ambiguity.

Absolutely agree. We probably should go with the obvious choices: JS type, native JS type and non-native JS type where necessary to solve ambiguity.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1696,7 +1696,8 @@ abstract class GenJSCode extends plugins.PluginComponent
           else            genExpr(tree)
       }
 
-      if (!isScalaJSDefinedJSClass(currentClassSym)) {
+      if (!isScalaJSDefinedJSClass(currentClassSym) ||
+          isRawJSFunctionDef(currentClassSym)) {
```
gzm0: This is really ugly IMO. Not sure how to do it better though...

gzm0:
Updated.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: There are uglier things surrounding anonymous functions in the scalac backend. It doesn't bother me that much.

PR #2832, opened 2017-03-24
Remove LinkingUnitJSEnv.

sjrd:
This API was only useful for Rhino's lazy loading.


sjrd:
That's an easy one :) Another consequence of #2825, actually.

sjrd:
No I think `linkUnit` is the only way tools like https://github.com/lionelfleury/scala-js-call-graph can exist.

gzm0:
Not sure I take this as an argument :P The tool already has to instantiate a basic backend to get the symbol requirements from it. So it might as well wrap a backend for example.

The main reason I don't like `linkUnit`, is that it is in a public trait, but should be plugin private. We need to find a way to separate the linker API (and hence the guarantees we give) better:

- Usage should remain backwards compatible at all time (in a major branch).
- Plugins will need to accept breaking changes more often, otherwise we cannot develop anything any more.

So in practical terms, that means, the sbt plugin, or a gradle, maven, ant whatever plugin for Scala.js support should remain backwards compatible. However, linker plugins (like the call graph plugin or the reflect plugin) will need to deal with changes. This is coherent with how the Scala compiler itself handles its compatibility.

sjrd:
I agree with your general notion of compatibility levels, but I'm not sure how it impacts any decision on `linkUnit`. We won't escape having public methods in the Tools API. And it's normal. Actually every method used by a plugin should be public in the Tools. I mean, this is obvious. I really don't get how this makes an argument against `linkUnit` being public!?

> So it might as well wrap a backend for example.

You mean it should *pretend* to be a `LinkerBackend` (that wraps another one), just so that it can receive the `LinkingUnit` as an argument to its `emit` method, but actually it doesn't emit any .js file, but rather side-effects somewhere to store the `LinkingUnit` to perform the call graph visualization? This seems uglier than what we have now. (although, granted, it doesn't need any API to do that)

gzm0:
> I really don't get how this makes an argument against `linkUnit` being public!?

It's less about it being public than it being directly next to `link` which IMO should have very different compatibility guarantees.

> This seems uglier than what we have now.

Not sure it is uglier. For example now, if I configure my linker to have different backend (with different symbol requirements), the call graph visualization is plain wrong. Agreed that intercepting things in a compilation pipeline (and showing a UI or writing data somewhere) is never very nice, but I haven't seen anything nicer in practice.

In the end, what you want is "run everything just like normal, but give me this debug info".

PR #2838, opened 2017-03-25
Turn most warnings in PrepJSInterop into errors (#1111)

gzm0:
Some warnings remain, they will be addressed in separate issues (#2829, #2834, #2836).

### DIFF
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
@@ -847,7 +847,7 @@ class JSInteropTest extends DirectTest with TestHelpers {
   }
 
   @Test
-  def warnPrivateMemberInNative: Unit = {
+  def noPrivateMemberInNative: Unit = {
 
     """
     @js.native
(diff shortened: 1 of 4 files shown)


sjrd:
Unsurprisingly, this needs a rebase as well.

gzm0:
Rebased.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -1823,9 +1823,9 @@ class JSInteropTest extends DirectTest with TestHelpers {
         @JSName("foo")
         def a: Int = js.native
       }
-      """ hasWarns
+      """ hasErrors
       """
-        |newSource1.scala:13: warning: A duplicate @JSName annotation is ignored. This will become an error in 1.0.0.
+        |newSource1.scala:13: error: Only a single @JSName annotation per member is allowed.
```
sjrd: "a single [...] per member" is a weird formulation. It's just a feeling, mind you, but I would suggest a formulation similar to that for native load specs, e.g.,

    A member can only have one @JSName annotation.

sjrd:
```
sbt 'set scalaJSOutputMode in testSuite := org.scalajs.core.tools.linker.backend.OutputMode.ECMAScript6' 'set jsEnv in testSuite := NodeJSEnv().value.withSourceMap(false)' 'set scalaJSModuleKind in testSuite := ModuleKind.CommonJSModule' testSuite/test
[...]
[error] .../test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala:110: Native JS classes may not have private constructors. Use `private[this]` to declare an internal constructor.
[error]   class Buffer private () extends js.typedarray.Uint8Array(0) {
[error]                ^
```

gzm0:
Updated. Hopefully CI passes :P

Issue #2797, opened 2017-03-09
Facades should be able to support all values of ModuleKind


japgolly:
Library authors should be able to write facades in such a way that downstream users can use any `scalaJSModuleKind` value they want/need.

> my vision for the final state is that an @js.native class or object must have exactly 1 annotation among @JSGlobal, @JSImport and @JSLoadVia.

Not realising the above vision wasn't being realised just yet, I became very fearful because such a direction would leave me with only 3 unhappy choices for scalajs-react:

1. Put `@JSGlobal` on my facades and don't allow users to use `CommonJSModule`.
2. Put `@JSImport` on my facades and force users to use `CommonJSModule` and probably force them to use scalajs-bundler which will in turn force them to only import deps a certain way and, only output one huge single JS file.
3. Double build and double publish all of my library artifacts so that users can choose whether they want to use modules / script.

I think it'd be better to allow us to annotate our facades with both `@JsGlobal` and `@JsImport`, then downstream at link-time have Scala.JS choose one or the other depending on users' `ModuleKind` setting.

julienrf:
Hi, I’m not sure it would be easy to adapt to scalajs-react but a general pattern to suport both is the following:

1. define the facade *trait*

~~~ scala
package myfacade

@js.native
trait MyFacade extends js.Object {
  …
}
~~~

2. define a global facade object:

~~~ scala
package myfacade
package global

@JSGlobal
object MyFacade extends MyFacade
~~~

3. define a CommonJS module facade

~~~ scala
package myfacade
package module

@JSImport("my-facade", …)
object MyFacade extends MyFacade
~~~

Then, at use site, users can pick either `myfacade.global.MyFacade` or `myfacade.module.MyFacade`.

japgolly:
@julienrf Unfortunately that approach only works when downstream users are using raw facades directly. In order to provide a nicer dev experience and/or build Scala-based features on top of the facades, the *library* interfaces with the facades directly and so, has to make a choice about which to use (I mean either `myfacade.global.MyFacade` or `myfacade.module.MyFacade` in your example).

This below much more accommodating. What we you think about a scheme like this? 
```scala
// Great! This will work for both ModuleKind.NoModule and ModuleKind.CommonJSModule!
@JSGlobal
@JSImport("my-facade", …)
@js.native
object MyFacade


// This will only work for the script style.
// Maybe compiler could warn that ModuleKind.CommonJSModule won't be supported if a user uses it.
@JSGlobal
@js.native
[...]

PR #2846, opened 2017-03-28
Remove the dependency on ResolvedJSDependency in the JSEnv API

sjrd:
This removes the annoying dependency (sic!) of the `JSEnv` API on `ResolvedJSDependency`. This was the only part of `tools.jsdep` that was used in the `JSEnv` API, so it decouples those.

This will be very important as we separate `jsDependencies` in a separate plugin, as the `tools.jsdep` package should go with that, out of the core.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -667,7 +667,7 @@ object ScalaJSPluginInternal {
 
         val file = scalaJSLinkedFile.value
         log.debug(s"Loading JSEnv with linked file ${file.path}")
-        env.loadLibs((libs :+ file).map(ResolvedJSDependency.minimal(_)))
+        env.loadLibs(libs :+ file)
       },
 
       scalaJSModuleIdentifier := Def.taskDyn[Option[String]] {
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -638,11 +638,36 @@ object ScalaJSPluginInternal {
       loadedJSEnv := {
         val log = streams.value.log
         val env = resolvedJSEnv.value
-        val libs =
+        val deps =
           resolvedJSDependencies.value.data ++ scalaJSConfigurationLibs.value
+
+        /* Implement the behavior of commonJSName without having to burn it
+         * inside NodeJSEnv, and hence in the JSEnv API.
+         * Since this matches against NodeJSEnv specifically, it obviously
+         * breaks the OO approach, but oh well ...
+         */
+        val libs = env match {
+          case _: org.scalajs.jsenv.nodejs.NodeJSEnv =>
+            val libCache = new VirtualFileMaterializer(false)
+
+            for (dep <- deps) yield {
+              dep.info.commonJSName.fold {
+                dep.lib
+              } { commonJSName =>
+                val fname = libCache.materialize(dep.lib).getAbsolutePath
+                new MemVirtualJSFile(s"require-$fname").withContent(
+                  s"""$commonJSName = require("${escapeJS(fname)}");"""
+                )
+              }
+            }
+
+          case _ =>
+            deps.map(_.lib)
+        }
+
         val file = scalaJSLinkedFile.value
         log.debug(s"Loading JSEnv with linked file ${file.path}")
-        env.loadLibs(libs :+ ResolvedJSDependency.minimal(file))
+        env.loadLibs((libs :+ file).map(ResolvedJSDependency.minimal(_)))
       },
 
       scalaJSModuleIdentifier := Def.taskDyn[Option[String]] {
--- a/js-envs/src/main/scala/org/scalajs/jsenv/AsyncJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/AsyncJSEnv.scala
@@ -10,19 +10,18 @@
 package org.scalajs.jsenv
 
 import org.scalajs.core.tools.io.VirtualJSFile
-import org.scalajs.core.tools.jsdep.ResolvedJSDependency
 
 trait AsyncJSEnv extends JSEnv {
-  def asyncRunner(libs: Seq[ResolvedJSDependency], code: VirtualJSFile): AsyncJSRunner
+  def asyncRunner(libs: Seq[VirtualJSFile], code: VirtualJSFile): AsyncJSRunner
 
   final def asyncRunner(code: VirtualJSFile): AsyncJSRunner =
     asyncRunner(Nil, code)
 
-  override def loadLibs(libs: Seq[ResolvedJSDependency]): AsyncJSEnv =
+  override def loadLibs(libs: Seq[VirtualJSFile]): AsyncJSEnv =
     new AsyncLoadedLibs { val loadedLibs = libs }
 
   private[jsenv] trait AsyncLoadedLibs extends LoadedLibs with AsyncJSEnv {
-    def asyncRunner(libs: Seq[ResolvedJSDependency],
+    def asyncRunner(libs: Seq[VirtualJSFile],
         code: VirtualJSFile): AsyncJSRunner = {
       AsyncJSEnv.this.asyncRunner(loadedLibs ++ libs, code)
     }
--- a/js-envs/src/main/scala/org/scalajs/jsenv/ComJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/ComJSEnv.scala
@@ -10,7 +10,6 @@
 package org.scalajs.jsenv
 
 import org.scalajs.core.tools.io.VirtualJSFile
-import org.scalajs.core.tools.jsdep.ResolvedJSDependency
 
 /** An [[AsyncJSEnv]] that provides communication to and from the JS VM.
  *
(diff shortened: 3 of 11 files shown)


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -640,11 +640,36 @@ object ScalaJSPluginInternal {
...
         val env = resolvedJSEnv.value
-        val libs =
+        val deps =
           resolvedJSDependencies.value.data ++ scalaJSConfigurationLibs.value
+
+        /* Implement the behavior of commonJSName without having to burn it
+         * inside NodeJSEnv, and hence in the JSEnv API.
+         * Since this matches against NodeJSEnv specifically, it obviously
+         * breaks the OO approach, but oh well ...
+         */
```
gzm0: I'm fine with this as a transitionary measure, but going forward we definitely need a better solution than this (even inside core). What is the story going forward for this?

Personally, I believe the solution in #2797 (allow multiple loading specs) is the way to go forward. But I'm not sure you like it that much. In practical terms, it would simply move `CommonJSName` into the `.sjsir` and everything would be gone (up to the build author to make sure their JSEnv and module mode are in sync).

js-envs-test-suite/src/test/scala/org/scalajs/jsenv/test/RetryingComJSEnvTest.scala
```diff
@@ -27,17 +26,17 @@ class RetryingComJSEnvTest extends JSEnvTest with ComTests {
     private[this] var fails = 0
     private[this] var failedReceive = false
 
-    def jsRunner(libs: Seq[ResolvedJSDependency],
+    def jsRunner(libs: Seq[VirtualJSFile],
```
gzm0: Not directly related to this PR, but now that `LinkingUnitJSEnv` is gone, we should probably rethink whether `loadLibs` still makes sense.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: I'm not entirely sold on the multiple load specs solution yet, true. But I'm not sure how that relates to this PR. A double loading spec would be quite different than `commonJSName`, and in a good direction anyway. I don't think `commonJSName` is the way forward, and neither are `jsDependencies` the way they exist now. `jsDependencies` have serious issues with CommonJSModule support, so much so that they are incompatible. I don't think trying to find an elegant way to deal with `commonJSName` is a priority at this point.

js-envs-test-suite/src/test/scala/org/scalajs/jsenv/test/RetryingComJSEnvTest.scala
sjrd: Absolutely. I have this in my internal todo list. But one or two more simplifications are required before getting there.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: > But I'm not sure how that relates to this PR.

It relates to this PR in the sense that this PR is un-building a (bad) solution to a real problem we do not have (or have agreed on) a replacement yet.

IMHO requiring that the dependency resolver needs to know which JSEnvs need CommonJS resolution is an absolute no-go (but perfectly acceptable as a transition).

I would feel better about this PR if we had an idea of the end solution we envision and ideally some sort of agreement from the community.

sjrd: Ah, I see what you mean, now.

However, I'm pretty sure that the problem addressed by `commonJSName` is completely different than the problem in #2797. As such, I don't think we're removing a solution to #2797. Let me elaborate.

### The problem addressed by `commonJSName`

I design my codebase (and potentially facades) exclusively for a "browser" environment. Therefore, I depend only on libraries that are meant to be used directly in a browser environment, without any processing. However, I'm unlucky because one such library is "too smart", and it behaves differently if it detects a Node.js environment.

`commonJSName` solves this problem by "restoring" the browser behavior of the library when it is run in a Node.js environment.

Can multiple loading specs solve this problem too? Possibly, but I don't think so. Even if I prepare my library for `@JSGlobal`+`@JSImport`, only one of those is selected to perform `fastOptJS`, as determined by `scalaJSModuleKind`. The same `-fastopt.js` is run both in the browser and in Node.js, so it cannot adapt to the "too smart" behavior of the library anyway.

What else can solve this problem? IMO any bundling tool for JS libraries, or, put another way, JS dependency manager. Depending on where libraries are fetched from, and how they are processed, it might be necessary or not to process them so that they all conform to one model (either all conform to CommonJS, or all conform to browser). `jsDependencies` is one such JS dependency manager, and `commonJSName` causes it to process some JS libraries so that they all follow the browser environment. It therefore makes sense that the `commonJSName` "processor" stays with the dependency manager that uses it, i.e., `jsDependencies`.

### The problem in #2797

Here the problem is that I want to write a library in a way that is agnostic to whether the final application uses the browser environment (`NoModule`) or a CommonJS environment (possibly with an associated bundler).

Can `commonJSName` solve this problem? Absolutely not, at least not in its current incarnation: `commonJSName` is tight to `jsDependencies`, and we know `jsDependencies` to be completely incompatible with the `CommonJSModule` module kind. So there's no way they can provide a solution to a problem that only manifests itself when we double-target `NoModule` and `CommonJSModule`.

### Conclusion

The problem described in #2797 and the problem addressed by `commonJSName` are completely different issues. And the solutions to either cannot solve the other problem. Hence, I don't think that removing `commonJSName` from core is a step backwards wrt. #2797. It is completely orthogonal.

gzm0: Ah OK, I see. Indeed `commonJSName` allowed the same `-fastopt.js` to be run in Node.js and in the browser. I do not think this is a use case we need to support.

However, `commonJSName` also allowed facade authors to make the facade agnostic to whether it is run in a module or non-module environment. Examples:

- [Phaser Scala JS](https://github.com/dpitkevics/Phaser-Scala-JS/blob/master/build.sbt)
- [AWS Facade](https://github.com/balshor/aws-sdk-scalajs-facade/blob/master/build.sbt)
- [Raphael Scala.js](https://github.com/yoeluk/raphael-scala-js/blob/master/build.sbt)
- [Scala.js Moment.js](https://github.com/vpavkin/scala-js-momentjs/blob/master/build.sbt)

However, the most prevalent use on GitHub is `React` and `ReactDOM`.

We can easily replace these uses of `commonJSName` with #2797 as follows (at the example of Phaser):
- Library adds `@JSImport("phaser.js", JSImport.Namespace)` on whatever has `@JSGlobal("Phaser")`.
- Users chose appropriate module loading based on JS environment they are using.

I acknowledge that this solves the problem on a different level and is therefore indeed somewhat orthogonal to this PR. However, I still do not understand what your desired end-state with respect to `commonJSName` is (once all this is moved to a separate plugin).

Do you want to:
- Remove it entirely? (Fine with that per-se, community maybe not so much)
- Leave the type test on the JS Env? (Not fine with that :))
- Add an option to the new bundler plugin to chose the behavior? (Probably the best option at this point).

gzm0: Just re-reading this again: I think we are making different implicit assumptions about how much the module kind you are using is tied to the execution environment. It seems like:

- You think they are mostly independent.
- I think they are almost inseparable.

It feels like this causes the different perception of the two solutions to emerge.

sjrd: Yes, indeed, I think they are mostly independent. Or rather, I think that `CommonJSModule` does not preclude a browser environment, as long as it is used with an appropriate bundler (e.g., webpack in scalajs-bundler). Using `NoModule` but actually running on Node.js (in prod, I mean) should not happen, although there probably still are occurrences of that in codebases created before the introduction of `CommonJSModule`. Your list of facades actually provides one such example: the AWS facade.

The Phaser facade is also interesting. It is actually a very good example of why `commonJSName` and more generally `jsDependencies` is dangerously misdesigned: some classes such as [`pixi.Sprite`](https://github.com/dpitkevics/Phaser-Scala-JS/blob/master/src/main/scala/org/scalajs/phaser/pixi/Sprite.scala) actually have an `@JSName("PIXI.Sprite")`, referring to a different top-level JS object than the `commonJSName` they use. That class will work in the browser, but not in Node.js.

> However, I still do not understand what your desired end-state with respect to commonJSName is

In my mind, `commonJSName` as a surface feature is tight to `jsDependencies`, as it was added as a patch because `jsDependencies` did not work for "too smart" libraries. I would therefore move it together with `jsDependencies` in a separate plugin. `jsDependencies` itself, I basically consider to be on life support in 1.x. We keep it running as before for people who were using it, but we do not invest in it anymore.

As such, I don't consider the hack of the instance test on `NodeJSEnv` particularly annoying. It's a life-support hack. That said, if people complain, it would be easy to add a configurable predicate instead, so that people have an additional escape hatch if they use some other obscure CommonJS-based runtime.

> Add an option to the new bundler plugin to chose the behavior? (Probably the best option at this point).

Not sure what you are referring to with "the new bundler plugin". If you are talking about scalajs-bundler, it can already do that. It's even in its cookbook: https://scalacenter.github.io/scalajs-bundler/cookbook.html#global-namespace. That recipe basically forces an (npm) library to be exposed as a global object, instead of having to be `require`d.

gzm0: Ok. I guess we have a good enough understanding of what our options going forward are and the course of action you propose seems reasonable.

So just to resume what we'll do going forward:

1. Merge this PR as it is.
2. Separate all the `jsDependencies` logic into a separate plugin, making it clear that it is on life support only (that's what I meant with the "bundler plugin").
3. If the need arises, add a setting in the plugin to trigger the common JS behavior.

js-envs-test-suite/src/test/scala/org/scalajs/jsenv/test/RetryingComJSEnvTest.scala
gzm0: Also, I think the distinction `libs` / `code` doesn't make sense anymore. Also for another PR.

sjrd:
OK I'll just rebase and test the individual commits.

Issue #2811, opened 2017-03-18
Adjust to Scala 2.13 moving parallel collections into a module


SethTisue:
we are interested in keeping Scala.js in the Scala 2.13 community build continuously throughout the 2.13 cycle, if at all possible, because Scala.js is so important in itself, and because so many other projects in the community build depend on it

after the merge of https://github.com/scala/scala/pull/5603, Scala.js no longer compiles on 2.13 because of the references to parallel collections in scalalib/overrides-2.13/scala/collection/immutable/Range.scala (see [example failure log](https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/283/consoleFull))

I thought I might try to patch it myself, but I don't really understand what the `overrides` directories are for, and when I diff your Range.scala against the one in Scala 2.13 HEAD, I see other differences besides just ones related to parallel collections.

so... is this a change y'all could do yourselves...? 🙏

SethTisue:
you can reproduce the failure outside of dbuild with:

```
set resolvers += "scala-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration/"
++2.13.0-pre-e2a2cba
scalalib/compile
```

gzm0:
The overrides are replacements for scalalib implementations when we can't take them as they are (for JDK dependencies or efficiency reasons). So it's probably as simple as just backporting the changes made in the 2.13 branch. I'll look into it.

sjrd:
The fix for this issue has now been made its way to master as well.

PR #2849, opened 2017-03-30
Forward-Port 0.6.x to master

gzm0:
This is necessary to make the community build actually pass (see #2811).


project/BinaryIncompatibilities.scala
```diff
@@ -6,6 +6,19 @@ object BinaryIncompatibilities {
...
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.ConcurrencyUtils#AtomicAccOps.removeAll$extension"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.ConcurrencyUtils#AtomicAccOps.removeAll"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.ParIncOptimizer#CollOps.prepAdd"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.ParIncOptimizer#CollOps.finishAdd"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.core.tools.linker.frontend.optimizer.ParIncOptimizer#CollOps.emptyParIterable")
```
gzm0: Note sure what to do about this. It probably makes our life easier if we just don't touch this in the master branch... WDYT?

sjrd: Meh. It's probably less trouble, but it's also kind of absurd in a "static" view of the codebase. Maybe we should add a comment at the beginning of the file, on `master`, saying that the contents are irrelevant during the 1.0.0-SNAPSHOT cycle, and that they are simply accidental ports from the 0.6.x branch?

sjrd:
Yeah, I really don't like this absurd synchronous view of the repo. If we were sure that this will always ever be the only kind of potential conflict that we could have, then we could continue simply opening PRs from 0.6.x into master, and the practical benefit would probably outweigh the absurdity.

But there will be cases where merging 0.6.x causes (non-trivial) conflicts, which will have to be resolved in a merge commit anyway. In that case, we cannot simply PR 0.6.x into master. A separate local branch will have to be made. With a separate local branch, it's easy to alter the merge commit not to modify `BinaryIncompatibilities.scala` ever.

The scala/scala has adopted the following merging procedure, which I like: (see https://github.com/scala/scala/pull/5794 for an example PR)

```
% update local branches
$ git checkout master
$ git checkout -b merge-0.6.x-into-master-march-30
$ export mb=$(git merge-base scalajs/0.6.x scalajs/master)
$ git log --graph --oneline --decorate $mb..scalajs/0.6.x | cat
[...]  a nice graph of what is about to be merged
$ git merge scalajs/0.6.x
$ [...] resolve conflicts
$ git push -u origin merge-0.6.x-into-master-march-30
```
They even amend the commit message of the merge commit with a verbatim copy of the above  sequence of commands, so that the conflicts are well documented. After that, they create a PR to merge `merge-0.6.x-into-master-march-30` into `master`.

I think we should do precisely the same thing. This procedure works well, caters for merge conflicts that need to be resolved locally, and property documents everything. As part of this procedure, it would be a very small burden to clear out changes to `BinaryIncompatibilities.scala` in the merge commit.

WDYT?

gzm0:
I see. Making a local branch definitely makes sense. What I don't fully get though: Doesn't this create a "double" merge commit in `master`?

```
* 26504fe7f5 Merge pull request #5794 from SethTisue/merge-2.12-to-2.13-mar-21
* 17633b2db3 Merge branch '2.12.x' into 2.13.x
```

I think making a separate merge branch makes total sense (also to allow 0.6.x to move on), but I would suggest we the fast-forward the merge commit (or use GitHub's rebase feature?).

I have investigated more and what I didn't fully understand, is that from git's point of view, there is actually no merge. It's just a huge cherry-pick. scala/scala@7633b2db358fa82581c797aab5bfc2d6936a048 only has a single parent which is in 2.13.x IIUC.

What I do not understand then, is how does merge-base determination work next time?

sjrd:
Indeed, we can improve the procedure with a fast-forward :+1: I think scala/scala doesn't do fast-forwards, so I guess that's why they don't do that.

> I have investigated more and what I didn't fully understand, is that from git's point of view, there is actually no merge. It's just a huge cherry-pick.

The commit you mention leads to a 404, but I guess you're talking about https://github.com/scala/scala/commit/17633b2db358fa82581c797aab5bfc2d6936a048 (the "cherry-pick") and https://github.com/scala/scala/commit/26504fe7f5fd275eb0fa78ebe3788c4d969ce7cb (its GitHub-generated merge into master).

Hum, that doesn't seem right :s Did they make a mistake here?

Anyway, with a fast-forward we wouldn't have this issue.

gzm0:
Closing in favor of #2851.

PR #2850, opened 2017-03-30
Switch to Project#settings in the build

dwijnand:


dwijnand:
Best reviewed ignoring whitespace changes :)

dwijnand:
I'm happy to change any details. Review away with furious passion.

sjrd:
Jenkins, test this please

sjrd:
There's an infinite recursion in the build, with the following loop in the stack trace:
```
	at Build$ProjectOps$.withScalaJSCompiler$extension(Build.scala:371)
	at Build$.libraryAux$lzycompute(Build.scala:996)
	at Build$.libraryAux(Build.scala:993)
	at Build$.library$lzycompute(Build.scala:1077)
	at Build$.library(Build.scala:1007)
	at Build$.compiler$lzycompute(Build.scala:536)
	at Build$.compiler(Build.scala:516)
```

sjrd:
In [my previous port](https://github.com/sjrd/scala-js/commit/359152da10a0ccaf0a5bd5aa70a25d30740a1d9c), there were 3 places where I needed to use `.settingsLazy()` instead of `.settings()`.

dwijnand:
Indeed. Just came to report the same :)

dwijnand:
Sorry for the long delay. As you said your build has some cycles. `.settingsLazy` isn't available in 0.13.13, so I can't just hotfix this branch.

We might or might not ship settingsLazy in 0.13.x, we'll see.

I don't have any more time to spend on seeing if I can fix this branch, I need to get back to porting your build to sbt 1.0.0-SNAPSHOT. Sorry.

dwijnand:
@sjrd This is passing for me locally now. WDYT?

sjrd:
Jenkins, retest this please

project/Build.scala
```diff
@@ -1091,73 +1091,79 @@ object Build {
...
-          libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-          previousArtifactSetting
-      )
+      base = file("stubs")
+  ).settings(
+      commonSettings,
+      publishSettings,
+      name := "Scala.js Stubs",
+      libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value//,
+      //previousArtifactSetting
```
sjrd: `previousArtifactSetting` shouldn't be commented out, here.

project/Build.scala
```diff
@@ -1091,73 +1091,79 @@ object Build {
...
-    settings = commonSettings ++ publishSettings ++ myScalaJSSettings ++
-      fatalWarningsSettings ++ Seq(name := "Scala.js JUnit test runtime")
+    base = file("junit-runtime")
+  ).settings(
+    commonSettings,
+    publishSettings,
+    myScalaJSSettings,
+    fatalWarningsSettings,
+    name := "Scala.js JUnit test runtime"
   ).withScalaJSCompiler.dependsOn(testInterface)
```
sjrd: Since you're messing with these lines anyway, could you be a boy scout and fix the indentation inside the `()` to be 4 spaces instead of 2 (inside `Project( ... )` and inside `.settings( ... )`?

project/Build.scala
```diff
@@ -1173,36 +1179,39 @@ object Build {
...
-    )
+    base = file("junit-plugin")
+  ).settings(
+    commonSettings,
+    publishSettings,
+    fatalWarningsSettings,
+    name := "Scala.js JUnit test plugin",
+    crossVersion := CrossVersion.full,
+    libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value,
+    exportJars := true
```
sjrd: Same thing with the indentation.

project/Build.scala
```diff
@@ -1463,402 +1472,409 @@ object Build {
...
-          val hasBug2382 = v.startsWith("2.10.") || v.startsWith("2.11.")
-          val sourceFiles1 = {
-            if (hasBug2382)
-              sourceFiles.filterNot(_.getName == "OuterClassTest.scala")
-            else
-              sourceFiles
-          }
+    base = file("test-suite/js")
+  ).settings(
+    commonSettings,
```
sjrd: One more indentation fix.

sjrd:
@gzm0 Did we simply overlook `LocalProject` as a solution to our cyclic dependencies, or was there a deeper reason? I'm pretty sure we just overlooked it.

Also, this targets the master branch. I think it's reasonable. Targeting 0.6.x would simplify merges if there are future changes to `Build.scala` in 0.6.x, but that seems unlikely (by the very nature of 0.6.x being maintenance-only now).

sjrd:
@dwijnand Looks like there's a potential for a follow-up cleanup using the `project` macro rather than the `Project` constructor :)

Regarding commit history, ideally I'd like to see one commit introducing `LocalProject` first, then followed by one commit that uses `.settings(...)`. So basically reversing the two commits that you have now.

gzm0:
In think we just shied away from it, because it is stringly typed. IIRC, in some contexts at least it was not even necessary to say `LocalProject`. Just the string would be enough (might have changed though...).

dwijnand:
> IIRC, in some contexts at least it was not even necessary to say `LocalProject`. Just the string would be enough (might have changed though...).

Yeah, that does ring a bell - there some obscure API somewhere that accepts strings..

@sjrd Sounds all good. I'll update in the morning.

project/Build.scala
dwijnand: Shit. Sorry. Good eye, man.

dwijnand:
I did the additional changes in a 3rd commit. I'm happy to rewrite the history as preferred.

dwijnand:
@gzm0 found 'em: https://github.com/sbt/sbt/blob/v0.13.13/main/settings/src/main/scala/sbt/Reference.scala#L96-L103

If someone gets a chance please could you instruct Jenkins to test this again. Thanks.

sjrd:
Jekins, retest this please

dwijnand:
Ok. So tests passed 🎉 , but looks like there are merge conflicts 💩. Let me know what you want the history to look like, so I'll do one last rewrite (hopefully).

gzm0:
> Let me know what you want the history to look like.

Please rebase onto master and (probably) squash the new "More Changes" commit into the previous one.

sjrd:
Yes I think @gzm0's suggestion for the history is a good one. Keep the `LocalProject` change in its own commit, and then put everything else in a second commit.

The merge conflict is because one project was renamed (`javalibExTestSuite` -> `testSuiteEx`) and one was removed (`javalibEx`).

dwijnand:
Done.

project/Build.scala
```diff
@@ -1159,123 +1110,103 @@ object Build {
...
-  lazy val helloworld: Project = Project(
-      id = "helloworld",
-      base = file("examples") / "helloworld",
-      settings = exampleSettings ++ Seq(
-          name := "Hello World - Scala.js example",
-          moduleName := "helloworld",
-          scalaJSUseMainModuleInitializer := true
-      )
+  lazy val helloworld: Project = (project in file("examples") / "helloworld"
+  ).settings(
```
sjrd: That one is particularly confusing. Could you add `()` around `file("examples") / "helloworld"`, because the precedence is not clear, here.

project/Build.scala
```diff
@@ -1159,123 +1110,103 @@ object Build {
...
   ).withScalaJSCompiler.dependsOn(library)
 
-  lazy val reversi = Project(
-      id = "reversi",
-      base = file("examples") / "reversi",
-      settings = exampleSettings ++ Seq(
-          name := "Reversi - Scala.js example",
-          moduleName := "reversi"
-      )
+  lazy val reversi = (project in file("examples") / "reversi").settings(
```
sjrd: Same here.

project/Build.scala
```diff
@@ -1159,123 +1110,103 @@ object Build {
...
+      moduleName := "reversi"
   ).withScalaJSCompiler.dependsOn(library)
 
-  lazy val testingExample = Project(
-      id = "testingExample",
-      base = file("examples") / "testing",
-      settings = exampleSettings ++ Seq(
-          name := "Testing - Scala.js example",
-          moduleName := "testing",
+  lazy val testingExample = (project in file("examples") / "testing").settings(
```
sjrd: And here.

project/Build.scala
```diff
@@ -495,79 +489,74 @@ object Build {
...
-      base = file("ir/.js"),
-      settings = commonIrProjectSettings ++ myScalaJSSettings ++ Seq(
-          crossVersion := ScalaJSCrossVersion.binary,
-          unmanagedSourceDirectories in Compile +=
-            (scalaSource in Compile in irProject).value,
-          unmanagedSourceDirectories in Test +=
-            (scalaSource in Test in irProject).value
-      )
+  lazy val irProjectJS: Project = Project(id = "irJS", base = file("ir/.js")
+  ).settings(
```
sjrd: There are a bunch of other occurrences of this pattern. I'll spare you the line comments everywhere ;)

dwijnand:
Done. (for convenience, the pre-squashed commit is https://github.com/dwijnand/scala-js/commit/b2b3b5bd453a4de5a40c7a307c7f6173b3538b2f)

sjrd:
Jenkins, retest this please

PR #2853, opened 2017-03-31
Fix #2840: Remove the javalib-ex.

### LINKED ISSUE #2840: Move javalib-ex out of the core (or just plain remove it?)
Historically, `javalib-ex` has contained parts of the JDK that were implemented on top of non-standard JS features, or even depending on some JS libraries. Today, though, the only remaining thing is `ZipInputStream`, and it's not even complete.

The reason this thing exists in Core is that we need for the bootstrap test. Indeed, we need to unzip the .jars containing the .sjsir files in the bootstrapped linker, and we do that through `ZipInputStream`. However, for the purposes of the bootstrap tests, I think we should simply integrate zip reading in the bootstrap test classpath (which is not published), rather than depending on `javalib-ex` for that purpose.

This would then allow us to move `javalib-ex` out of the core. Since it's so small and rather incomplete, maybe it should just be removed. [Scaladex suggests that no one depends on it](https://index.scala-lang.org/scala-js/scala-js/scalajs-javalib-ex?target=_2.11). That said, to keep our promise on backward source compatibility, we should probably still publish it somewhere in a separate repo (possibly even a personal repo rather than one under `org.scala-js`).

Action items:

- [ ] Remove the dependency on `javalib-ex` in the bootstrap test
- [ ] Move javalib-ex out of the core

sjrd:


tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -173,56 +173,96 @@ trait VirtualSerializedScalaJSIRFile extends VirtualBinaryFile with VirtualScala
...
     } finally {
       stream.close()
     }
   }
+}
 
-  private class JarEntryIRFile(outerPath: String, val relativePath: String)
-      extends MemVirtualSerializedScalaJSIRFile(s"$outerPath:$relativePath")
-      with RelativeVirtualFile
+object VirtualJarFile {
```
gzm0: private?

sjrd: Indeed.

sjrd: Done.

PR #2854, opened 2017-04-03
update Jetty dependency

golem131:


sjrd:
Thanks @golem131.

Is this only for the sake of it, or does it address an observable issue? Given that jetty is there only for PhantomJS, and the latter is on its way out of the core repository (moved in a separate repo), I'd rather not touch this area (which has worked for 2.5 years) without a good reason.

golem131:
@sjrd Only for the sake of it

sjrd:
OK then I will close this PR, as I'd rather not take any chance with this area at the moment.

Issue #2842, opened 2017-03-27
Move JSDOMNodeJSEnv out of the core


sjrd:
With `jsDependencies` gone from the core #2841, there will be no further dependency on `JSDOMNodeJSEnv` in the rest of the sbt plugin. Therefore, `JSDOMNodeJSEnv` can itself be moved outside of the core, just like all the other JS envs (except the basic Node.js one).

Issue #2841, opened 2017-03-27
Move the jsDependencies out of the core


sjrd:
`jsDependencies` is a feature that never reached the level of stability and well-designed-ness of the rest of core. Moreover, there now exists a better alternative with `npmDependencies` of [scalajs-bundler](https://scalacenter.github.io/scalajs-bundler/). We can't just drop support for `jsDependencies`, though, as too many people rely on it still, and it's not broken enough to justify the drop.

We can however extract the entire `jsDependencies` in a separate sbt plugin. scalajs-bundler proves that it is possible.

Historically, `jsDependencies` was added in the core because we needed it to resolve Jasmine, back when our internal testing framework was Jasmine-based. Now that our testing is JUnit-based, and does not rely on any JS library, we can move `jsDependencies` out. There is still one place where we actually use `jsDependencies` in the core, though: `javalib-ex` depends on JSZip. However, `javalib-ex` should itself go out of the core #2840.

The disappearance of the internal setting `scalaJSRequestsDOM` from the core means that we won't support anymore the fact that `jsdom` support is automatically added as part of `resolvedJSEnv`. But IMO that's OK; it will even allow to move `JSDOMNodeJSEnv` outside of the core as well.

PR #2857, opened 2017-04-05
Untangle jsDependencies from JSDOMNodeJSEnv

sjrd:
This will be required so that we can move those two segments of the sbt plugin in separate repositories and sbt plugins (#2841, #2842).

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -19,9 +19,9 @@
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
         ++$scala helloworld/run \
         helloworld/clean &&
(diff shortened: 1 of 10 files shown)


sjrd:
Clean rebase, necessary for CI to pass after #2853 was merged.

sjrd:
I'll wait until #2850 is merged before rebasing this one, not to inflict yet another rebase on Dale.

dwijnand:
Thanks! :) Actually for me rebasing and reworking commits isn't a problem at all. It's the fact that I can't get Jenkins rolling on testing new changes that's frustrating. Makes everything take twice as long as (my) usual.

PR #2868, opened 2017-04-07
Fix #2780: Reach all methods of j.l.Integer before the optimizer.

### LINKED ISSUE #2780: Linking warnings after optimizer when calling Short.shortValue()
The following snippet:
```scala
    val x = new java.lang.Short(5.toShort)
    println(x.shortValue())
```
causes *warnings* in the refiner (after optimization), saying:
```
[warn] Referring to non-existent method java.lang.Integer.shortValue()scala.Short
[warn]   called from helloworld.HelloWorld$.main()scala.Unit
[warn]   called from core module entry points
[warn] involving instantiated classes:
[warn]   helloworld.HelloWorld$
```
This is because the first linker pass discards `java.lang.Integer.shortValue()`, as it is unused. But then the optimizer transforms the above into:
```scala
val x: int = 5
println(x.shortValue())
```
thanks to box/unbox elimination. And then it tries to invoke `shortValue()` on a primitive `int`, which makes it look into `Integer` instead of `Short`.

This does not prevent the program to be well-defined and work, so it's not super urgent to fix. Besides, it will be fixed by construction if we add primitive `byte`s and `short`s in the IR, which I intend to do in 1.0.0.

sjrd:
This was already partially fixed in 10997e9b04eb1a82225f85975073a64c09982448, which was a fix for `compareTo(Short)` were affected. #2780 shows that all methods of `java.lang.Integer` are affected, in fact, so this commit extends the existing fix to all methods.

Note that the methods still disappear after the refiner if they are not needed, and anyway they are all hijacked as helpers, so the quality of DCE is not affected by this fix.

There is no additional test, because the warnings could already be observed before, but as warnings, they did not fail the build.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
@@ -52,8 +52,25 @@ abstract class GenIncOptimizer private[optimizer] (semantics: Semantics,
 
     callMethods(LongImpl.RuntimeLongClass, LongImpl.AllIntrinsicMethods) ++
     optional(callMethods(LongImpl.RuntimeLongClass, LongImpl.OptionalIntrinsicMethods)) ++
-    callMethods(Definitions.BoxedIntegerClass,
-        Seq("compareTo__jl_Byte__I", "compareTo__jl_Short__I")) ++ // #2184
+    /* #2184 + #2780: we need to keep all methods of j.l.Integer, in case
+     * the corresponding methods are called on j.l.Byte or j.l.Short, and
+     * through optimizations become calls on j.l.Integer.
+     */
+    callMethods(Definitions.BoxedIntegerClass, Seq(
+        "byteValue__B",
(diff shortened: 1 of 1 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -52,8 +52,25 @@ abstract class GenIncOptimizer private[optimizer] (semantics: Semantics,
...
+        "longValue__J",
+        "floatValue__F",
+        "doubleValue__D",
+        "equals__O__Z",
+        "hashCode__I",
+        "compareTo__jl_Integer__I",
+        "toString__T",
+        "compareTo__jl_Byte__I",
+        "compareTo__jl_Short__I",
+        "compareTo__O__I"
```
gzm0: Just to double check: This is the bridge generated for `Comparable[Integer]`.

sjrd: Yes, that's right.

PR #2872, opened 2017-04-07
Deprecate NodeJSEnv().value and JSDOMNodeJSEnv().value.

sjrd:
The constructors `new NodeJSEnv()` and `new JSDOMNodeJSEnv()` should be used instead.

Those `Initialize` constructors were added to be consistent with `RhinoJSEnv().value` (which needs `scalaJSLinker.value` and `scalaJSRequestsDOM.value`) and `PhantomJSEnv().value` (which needs `scalaJSPhantomJSClassLoader.value`). However, as Rhino is deprecated, and PhantomJS will move in its own repo, it doesn't make any sense to keep being consistent. Moreover, `JSDOMNodeJSEnv` will also move to its own repo, and won't even provide a companion sbt plugin (because there is no need to), so `JSDOMNodeJSEnv().value` would have nowhere to go, breaking the consistency anyway.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -156,46 +156,46 @@
   <task id="test-suite-ecma-script6"><![CDATA[
     setJavaVersion $java
     sbtretry 'set scalaJSOutputMode in $testSuite := org.scalajs.core.tools.linker.backend.OutputMode.ECMAScript6' \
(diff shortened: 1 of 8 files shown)


sjrd:
Updated. There's an entire new commit.

PR #2885, opened 2017-04-09
Fix #2862: Disallow @JSExport on non-members

### LINKED ISSUE #2862: Disallow @JSExport on non-members

gzm0:


gzm0:
Rebased and updated.

gzm0:
Rebased

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -100,48 +100,63 @@ trait PrepJSExports { this: PrepJSInterop =>
...
+    val exports = exportsOf(sym)
+    def isScalaClass = !sym.isModuleClass && !isJSAny(sym)
 
-  /** Checks and registers class exports on the symbol. */
-  def registerClassExports(sym: Symbol): Unit = {
-    assert(!sym.isModuleClass && !sym.hasAnnotation(JSNativeAnnotation),
-        "Expected a Scala.js-defined JS class")
-    registerClassOrModuleExportsInternal(sym)
+    if (exports.nonEmpty && checkClassOrModuleExports(sym, exports.head.pos) &&
+        !isScalaClass) {
```
sjrd: Since `checkClassOrModuleExports` is called through `genExportMember` for constructors of Scala classes, shouldn't the test `!isScalaClass` come before the test `checkClassOrModuleExports`? Currently it seems to me that there is a duplicate check.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -281,6 +281,10 @@ trait PrepJSExports { this: PrepJSInterop =>
         }
       }
 
+      lazy val symOwner =
+        if (sym.isConstructor) sym.owner.owner
+        else sym.owner
```
sjrd: Consider make this a normal `val`, even if it is only used in 2/3 of the cases below. The overhead of the `lazy val` in the cases where we need it will probably outweigh the benefit of avoiding two or three point dereferences. Especially since the "Normal" case will likely be less common than the "Top-level" and "Static" ones.

In fact we should probably rename `ExportDestination.Normal` to `ExportDestination.Member`, now that it's only used for member exports.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -433,127 +471,75 @@ class JSExportTest extends DirectTest with TestHelpers {
   }
 
   @Test
-  def noImplicitNameNestedExportClass: Unit = {
```
sjrd: Why is this test removed? I would assume the error messages would change, but we should still make sure that the snippets in this test are rejected. Or are they tested by some other test already?

project/Build.scala
```diff
@@ -982,8 +982,6 @@ object Build {
       libraryDependencies +=
         "org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided",
 
-      // js.JSApp is annotated with @JSExportDescendentObjects
-      scalacOptions += "-P:scalajs:suppressExportDeprecations",
```
sjrd: Looks like this change is not in the right commit. It should be in the same commit where we remove `@JSExportDescendentObjects` from `JSApp`.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
gzm0: :P Indeed, this actually doesn't work because of the special casing of `@JSExport` in nested things... Maybe we should just drop it :S

sjrd: Hum ... It might be easier to just drop it, indeed.

gzm0:
Updated. I think I found a reasonable solution with the error messages.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
sjrd: What about this comment? Otherwise looks good.

sjrd:
There are a bunch of `@JSExport`s in `jsinterop/ExportsTest.scala` that want to become `@JSExportTopLevel`s ;)

gzm0:
Updated (I really hope I didn't remove too many tests). I admit I did this a bit in a hurry to finish it :S

gzm0:
🎉

PR #2886, opened 2017-04-10
Remove support for bypassing linking errors.

sjrd:


tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
```diff
@@ -31,38 +31,30 @@ final class Refiner {
...
+    /* There must not be linking errors at this point. If there are, it is a
+     * bug in the optimizer.
      */
-    analysis.errors.foreach(Analysis.logError(_, logger, Level.Warn))
+    if (analysis.errors.nonEmpty) {
+      analysis.errors.foreach(Analysis.logError(_, logger, Level.Error))
+      throw new AssertionError(
+          "There were linking errors after the optimizer has run. " +
+          "This is a bug, please report it. " +
+          "You can work around the bug by disabling the optimizer.")
```
gzm0: Maybe say how?

sjrd: Done.

sjrd:
Updated with the "say how", and also removal of the option in `Scalajsld.scala`.

sjrd:
Rebased to satisfy Jenkins.

sjrd:
Updated with an additional commit that blacklists two partest using .java files.

PR #2887, opened 2017-04-10
Fix #2874: Remove the distinction libs/code in the JSEnv API.

### LINKED ISSUE #2874: Remove the distinction libs/code in the JSEnv API

sjrd:
`JSEnv#jsRunner` and friends `asyncRunner` and `comRunner` now only take a `Seq[VirtualJSFiles]`, containing all JS files to be given to the JS environment.

The JS environments therefore make no distinction between "library" files and the "code" file. The disinction was arbitrary anyway, and unifying them makes sure that all the JS files are first-class. In particular, it forced to solve shortcomings in terms of error handling in `JSDOMNodeJSEnv`.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
@@ -27,66 +27,66 @@ class JSDOMNodeJSEnv(
 
   protected def vmName: String = "Node.js with JSDOM"
 
-  override def jsRunner(libs: Seq[VirtualJSFile],
-      code: VirtualJSFile): JSRunner = {
-    new DOMNodeRunner(libs, code)
-  }
+  override def jsRunner(files: Seq[VirtualJSFile]): JSRunner =
+    new DOMNodeRunner(files)
 
-  override def asyncRunner(libs: Seq[VirtualJSFile],
-      code: VirtualJSFile): AsyncJSRunner = {
-    new AsyncDOMNodeRunner(libs, code)
-  }
+  override def asyncRunner(files: Seq[VirtualJSFile]): AsyncJSRunner =
+    new AsyncDOMNodeRunner(files)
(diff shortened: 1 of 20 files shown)


js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
```diff
@@ -27,79 +27,95 @@ class JSDOMNodeJSEnv(
...
-           |    onload: function (window) {
-           |      jsdom.changeURL(window, "http://localhost");
-           |      for (var k in window) {
-           |        if (windowKeys.indexOf(k) == -1)
-           |          global[k] = window[k];
-           |      }
-           |
-           |      ${code.content}
-           |    }
+           |    scripts: [${scriptsURIsAsJSStrings.mkString(", ")}]
```
gzm0: The fixes to the error handling here seem unrelated to the other changes. Make a separate commit? Also, is there a way to add a test for this?

sjrd: It's more or less related. Previously error handling worked well for `code`, but not for `libs`. While unifying `code` and `libs`, I had to treat `code` the same as `libs` (not the other way around because namespaces would clash), and I lost the good error handling for `code`. So I had to figure out how to deal with error handling for `libs`, which caused all the other changes.

It's tested because we have `BasicJSEnvTests.{failureTest,syntaxErrorTest}`. They're only testing the "code" part, but since now it's all unified, everything is tested. Fixing the error handling without unifying would be hard to test (and useless in a following commit).

Making a separate commit would be hard at this point, and especially hard to test, so I wouldn't even know if I did the commit extraction correctly. TBH I'd rather not do it.

gzm0: Ok, works for me. Thanks for the explanation.

PR #2890, opened 2017-04-11
Preparations before we can extract jsDependencies.

sjrd:


sjrd:
Review by @gzm0 in case you're not already rushed down for your vacations.
Otherwise I'll probably leave it open until you're back, because I don't feel comfortable doing these changes without a review.

sjrd:
Oh and, obviously I tested the HTML test runners locally.

project/Build.scala
```diff
@@ -1384,14 +1384,36 @@ object Build {
...
+        val patchedSystemPropertiesFile =
+          new MemVirtualJSFile("setJavaSystemProperties.js").withContent(code)
+
+        // Replace the normal `setJavaSystemProperties.js` file with the patch
+        for (file <- previousFiles) yield {
+          if (file.path == "setJavaSystemProperties.js")
+            patchedSystemPropertiesFile
+          else
+            file
+        }
```
gzm0: Hmm... I was going to say: Can't you just override the system properties in JS land, but then I realized that you cannot (easily) determine where to insert that file in the inclusion order. I'm wondering how many projects will have that problem...

Here, could you do two things:

- Leave the system properties changes in the separate setting (and pull it in right after). Seems more idiomatic?
- Find a more principled way to get the file name? Maybe a `private[scalajs]` constant?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -577,45 +607,14 @@ object ScalaJSPluginInternal {
...
         }
       },
 
-      loadedJSEnv := {
-        val log = streams.value.log
-        val env = resolvedJSEnv.value
-        val deps =
-          resolvedJSDependencies.value.data ++ scalaJSConfigurationLibs.value
+      // Crucially, add the Scala.js linked file to the JS files
+      scalaJSJSFiles += scalaJSLinkedFile.value,
```
sjrd: I have actually considered that for a good while, but it comes with its own set of issues.

* It makes it tricky to customize that list. You would want to customize it unscoped most of the time. But here we would have to use a stage-dispatch in the unscoped task, to be used by running environments, which means you can't customize it unscoped. You have to repeat the customization in the two tasks. 
* It creates a coupling from the jsEnv set of keys to the notion of stage task. That will force us to keep those things tightly coupled forever, but I'd like to investigate whether we can make JS envs less dependent on the Scala.js linker. Ideally they wouldn't care about Scala.js tools and anything related.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -251,9 +251,10 @@ object ScalaJSPlugin extends AutoPlugin {
         "Prints the content of a .sjsir file in human readable form.",
         CTask)
 
-    val scalaJSConfigurationLibs = TaskKey[Seq[ResolvedJSDependency]](
-        "scalaJSConfigurationLibs",
-        "List of JS libraries used as project configuration.", CTask)
+    val scalaJSJSFiles = TaskKey[Seq[VirtualJSFile]](
```
sjrd: It's definitely not an ideal name. I like `jsExecutionFiles`. Or maybe even `jsEnvFiles` as they are mainly supposed to be used with a JS env.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
```diff
@@ -20,10 +20,9 @@ import org.scalajs.testadapter.TaskDefSerializers._
 /** Template for the HTML runner. */
 private[scalajs] object HTMLRunnerTemplate {
 
-  def render(baseURI: URI, title: String, sjsFile: URI, jsdepsFile: URI,
+  def render(baseURI: URI, title: String, jsFiles: Seq[URI],
```
sjrd: I haven't considered changing URIs here in this PR, because that can be done later without changing the public API. I'd rather limit changes in this PR to things that are necessary to the get the public API we need, and refine internals later.

project/Build.scala
sjrd: We're in our build. We don't have access to `private[scalajs]`.

Leaving the generation of system properties in their own setting doesn't solve the issue, as we must only adapt the system properties for the purpose of `testHtmlFastOpt`, and not in `test` for example. I had considered that. For the same reason we can't just override the `javaOptions` either.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/HTMLRunnerTemplate.scala
gzm0: OK :)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Fair enough. So a downstream plugin could read `scalaJSStage` if it wants to.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: I prefer `jsExecutionFiles`, but would be fine with `jsEnvFiles`.

project/Build.scala
gzm0: Yeah, I just meant change it and read it again when you patch `scalaJSJSFiles`. Simply because it's more sbt-like.

sjrd: Ah I see. Yes, that should be possible, and would be better.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: OK let's go with `jsExecutionFiles` then.

sjrd:
Thanks! And enjoy your vacation :D

Issue #2805, opened 2017-03-14
Ability to create new JavaScript classes dynamically


japgolly:
How can I create a function that creates and return new ES6 classes?

This approach (obviously) doesn't work:
```scala
  def newClass(nameArg: String): js.Dynamic = {
    @ScalaJSDefined
    class X extends js.Object {
      def name = nameArg
    }
    js.constructorOf[X]
  }

newClass("A") eq newClass("B") // returns true
```

One solution is to use macros. Any other ideas or suggestions?

sjrd:
Ah ah! @gzm0 I've told you someone would end up needing this ^^

@japgolly We've discussed this internally. It's not yet possible. It's a gap in our interoperability features. It is scheduled for 1.0, with precisely the syntax you suggest above.

If you desugar down to ECMAScript 5.1 for now, a workaround is to create it manually as a `js.ThisFunction`, and then patching its `prototype`:
```scala
val X = (((thiz: js.Dynamic) => ()): js.ThisFunction).asInstanceOf[js.Dynamic]
X.prototype.name = nameArg
X
```
If however you're actually emitting ECMAScript 6, you're out of luck until Scala.js supports it in 1.0.

japgolly:
Cool, thanks for the workaround. I think I'll just leave it for now. Do you have a rough idea of when 1.0.0 will be out?

sjrd:
A rough idea would be 3-4 months. As soon as 0.6.15 is published, the master branch of the repo will target 1.0.0.

PR #2903, opened 2017-04-24
Constructor methods (js interop)

oyvindberg:
I wanted to give hacking on the compiler a shot, and I have to say it was surprisingly easy (at least this far). Very cool!

This is something I have been wanting for JS interop for a while, given that it's a prevalent pattern in the JS world, and the transformation to classes is not always trivial.

I'm sure there is a lot that is not optimal here, I just wanted to hear if you think this is worthwhile before spending more time. WDYT?


sjrd:

sjrd:
@oyvindberg Given that this is a language change, it must go through a Proposal before we discuss the code, in the form of an Issue. The proposal should clearly state the problem it addresses, the proposed solution, and ideally evaluate pros and cons with alternatives.

Changing the language is not something we do lightly, even ourselves, because we keep supporting everything forever.

For some good examples of Proposals that we had before, see:

* Accepted: #1795, #2009, #2702, #2734 (the latter is shorter)
* Rejected: #2720

sjrd:
Also, for us to be able to look at your PR, we need you to sign [the Scala CLA](http://www.lightbend.com/contribute/cla/scala).

oyvindberg:
Ok, good, thanks. I'll have a look at that tomorrow :)

sjrd:
I realized that this feature is trying to address #2398, in a way. It's providing some Scala syntax for the JavaScript syntax `new x.Foo(a, b, c)`. The Scala syntax it offers is `x.Foo(a, b, c)`. It would be so much better to offer the syntax `new x.Foo(a, b, c)` in Scala too!

That would be provided by allowing (and supporting through the pipeline) nested JavaScript classes. The snippet from the test case in the PR would be written as:
```scala
  @js.native
  @JSGlobal("ConstructorMethodTest")
  object ConstructorMethodTest extends js.Any {
    @js.native
    class Class(s: String) extends js.Object {
      val field: Int = js.native
      def method(): String = js.native
      def getConstructorParam(): String = js.native
    }
  }
```
without any ceremony.

Supporting nested JavaScript classes is technically much more involved than your `@JSConstructorMethod`, but it is much better from a language design point of view, and provides an even more intuitive and boilerplate-free syntax to users. It is something we vaguely have in mind for 1.x, together with #2805 (they don't look similar on the surface, but in fact they are underneath).

Therefore, I don't think your `@JSConstructorMethod` proposal would make it, when contrasted to the larger scheme.

oyvindberg:
Alright, sorry for the late answer, I should have explained myself right away.

It is #2398 , yes. I had a look through some issues, but failed to notice that one.

I modelled this after Typescript's constructor methods. In Typescript, it [seems] facade `class`es are sugar for a `var` with constructor methods (the constructor type) plus an `interface` with the class members, so constructor methods are more general than classes there. 

Given that, it is sometimes useful to talk about the constructor type, when Javascript people use the fact that constructor type instances are first-class values, and are returned from functions and so on. For example from the stdlib:

```typescript
interface RegExpConstructor {
    new (pattern: RegExp, flags?: string): RegExp;
    //...
    [Symbol.species](): RegExpConstructor;
}
```

Anyhow, I'm sure inner classes cover most important use-cases. It was fun to have a go at the compiler anyway :)

[seems]: https://www.sitepen.com/blog/2014/08/22/advanced-typescript-concepts-classes-types/

sjrd:
I realize I forgot to answer here, sorry.

I agree that it it would sometimes be useful to talk about the constructor type. So far our best type for those are `js.Dynamic` that can be given to `js.Dynamic.newInstance(ctor)(...)`. That's for example the result type of `js.constructorOf`. In theory, better types should exist, but that would require a different approach, I think (e.g., well-known types `js.Constructor[Ts..., C]` or something like that, like we have `js.FunctionN`s).

In practice, when we really do want something more precise, we can use a user-defined value class wrapping the `js.Dynamic` constructor. The solution in this PR does not really provide a less verbose or safer solution than such a wrapper.

In the present situation, I would at least wait until we have addressed the inner class issue before proceeding in this direction. Better support for inner classes should be uncontroversial, and then we can see whether more is still needed.

I'm going to close this PR for now.

PR #2925, opened 2017-05-03
Disallow @JSName on non js-native classes

gzm0:
This (hopefully) is the last step for #1111.


sjrd:
> This (hopefully) is the last step for #1111.

There's still https://github.com/scala-js/scala-js/blob/0fc3654b315fad0d7d22c02fe9ad50a0dac126ef/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala#L604

PR #2931, opened 2017-05-04
Fix #2785: No IIFE in fullOpt CommonJSModule

### LINKED ISSUE #2785: FastOpt is not consistent with fullOpt in Module enclosing function, when emitting CommonJSModule
When `scalaJSModuleKind := ModuleKind.CommonJSModule`,
`fullOptJS` produces module with an enclosing function, while `fastOptJS` does not.

BTW: 

Suggest adding enclosing functions rather than not. It provides better consistency when `-opt.js` acts as a chunk file containing many modules. As it provides one more ways to emit multiple modules within one scala/sbt project without sub-project. See [MDN strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)

gzm0:


tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -54,6 +54,11 @@ final class ClosureLinkerBackend(
 
   val symbolRequirements: SymbolRequirement = emitter.symbolRequirements
 
+  private val needsIIFEWrapper = moduleKind match {
+    case ModuleKind.NoModule       => true
+    case ModuleKind.CommonJSModule => false
+  }
```
sjrd: Hum ... any reason you're not using the same test as [`Emitter.needsIIFEWrapper`](https://github.com/scala-js/scala-js/blob/cc6117a5ccf226bc4dfde00f12d64183446a93c1/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala#L63-L76)? Or even *reuse* `Emitter.needsIIFEWrapper`?

sjrd: Ah! Silly me. Of course because `ClosureLinkerBackend` always uses `ECMAScript51Isolated`.

Issue #2800, opened 2017-03-13
Cannot access `class`es, `let`s and `const`s in the global scope


sjrd:
See http://stackoverflow.com/q/42745118/1829647

To fix this, we really need to access the global *scope*, not the global *object*.

For some relevant theory, read: http://www.2ality.com/2015/02/es6-scoping.html#the_global_object

Basically, since ES 2015, there are things that are in the global *scope* but not in the global *object*: top-level `let`s, `const`s and `class`es. This means that our internal scheme and surface semantics of only allowing access to the global object is not enough.

This is a major, severe, critical and whatever-you-wanna-call-it limitation, which we absolutely need to fix by 1.0.

sjrd:
So ... this is going to be hard. Language-wise, I would propose the following (spoiler: it's a breaking change):

First, we change our "spec" so that we do not refer to the global *object* anymore. Instead, everything that was referring to the global object now represents the global scope. In particular:
* `js.Dynamic.global`
* `@JSGlobal`-annotated things
* `@JSGlobalScope`-annotated things (fitting, isn't it?)

Concretely, this incurs the following breaking change, language wise: it is not possible to grab the *value* of a global scope thing anymore. Any reference to a global scope must be immediately followed by a `Select` (i.e., `foo.bar`) that selects a member of the global scope. That thing must, in addition, have a JS name which is a constant, valid JS identifier (a fortiori, not a symbol). The whole selection becomes a reference to that identifier in the global scope.

For example:
```scala
val x = js.Dynamic.global // illegal: cannot load the value of `global` without a `.`
val y = js.Dynamic.global.selectDynamic("[") // illegal: "[" is not a valid JS identifier
val ybis = js.Dynamic.global.selectDynamic(someVal) // illegal: not constant
val z = js.Dynamic.global.foo // legal, and represents `foo` in the global scope

@js.native @JSGlobal
object Bar extends js.Object

val bar = Bar // legal, and it means `Bar` in the global scope (as opposed to `<global>.Bar` in the current spec)

@js.native @JSGlobalScope
object Glob extends js.Object {
  val foobar: String = js.native
}

val glob = Glob // illegal: Glob is like `js.Dynamic.global`
val foobar = Glob.foobar // legal, represents `foobar` in the global scope
```
In the IR, we could have a new IR node `GlobalRef`, similar to `VarRef`, but that guarantees that it accesses a global variable.

Now the *really* tricky part is, how to *implement* it in `FunctionEmitter`?

[...]

PR #2932, opened 2017-05-05
Remove the `namespace` argument of `ScalaJSClassLoader`.

sjrd:
That argument was only used by `TestUtils` to support loading from the exports namespace. However, since `@JSExportDescendent{Classes,Objects}` has been removed, this support is useless and can be removed.

These changes are a prerequisite to implementing #2800 in the master branch.


test-interface/src/main/scala/org/scalajs/testinterface/ScalaJSClassLoader.scala
```diff
@@ -6,12 +6,19 @@ import java.net.URL
...
  */
-final class ScalaJSClassLoader(
-    val namespace: js.Dynamic) extends ClassLoader(null) {
+final class ScalaJSClassLoader extends ClassLoader(null) {
+
+  @deprecated(
+      "The `namespace` argument is ignored. " +
+      "Use the constructor with 0 argument.",
+      "1.0.0")
+  def this(namespace: js.Dynamic) = this()
```
gzm0: Don't you want to just remove this? I would deprecate it in the 0.6.x branch if you really are concerned that people actually use this.

sjrd: We can't deprecate it in 0.6.x because it's *necessary* in 0.6.x.

My idea would be to keep it deprecated during 1 or 2 milestones, and then actually remove it before the RCs. More generally, we'll need to make a large swipe of all deprecated methods in the codebase before the RCs.

gzm0: Ok. In that case it's fine. It's just that the `"1.0.0"` argument is a bit confusing.

sjrd:
Jenkins, retest this please

Issue #2638, opened 2016-10-25
Interop feature for new.target


sjrd:
I realized we haven't filed this anywhere. We need some interop feature for ES 6's `new.target`.

IMO the only reasonable thing to do is to provide a new primitive `js.newTarget` which returns exactly what `new.target` would return. The front-end will have to ensure that `js.newTarget` is only used lexically where a `new.target` would be allowed, i.e., in the constructor of a JavaScript class.

sjrd:
Well, in fact it can also appear in `function` bodies, in which case it binds in that function. *But* in an `=>` function, it is not rebound. Just like `this`, it looks up its lexical environment (to the enclosing `class` or `function`).

For the case of binding in `function` bodies, using a lexical `js.newTarget` is just as bad as if we had invented `js.this` to access the `this` parameter of a function. It seems that, for this use case, we actually need a set of `js.NewTargetThisFunctionN` :s

sjrd:
Once #3247 is in, we'll be able to add this feature in a backward compatible way, so I'm going to delay this to post-1.0.0.

PR #2934, opened 2017-05-06
Fix #2800: Refer to the global scope instead of the global object.

### LINKED ISSUE #2800: Cannot access `class`es, `let`s and `const`s in the global scope

sjrd:
This is a major redesign of accesses to "global" variables, both
from a spec point of view and in the implementation.

Previously, Scala.js specified accesses to global variables in
terms of the *global object*. It assumed that it was capable of
accurately detecting the global object of any JS environment, and
worked from there. For example, `js.Dynamic.global.foo` referred
to `$g.foo`, where `$g` was supposed to be the global object.

From a spec point of view, this was therefore different than how
a typical JavaScript code would be written. Indeed, in "normal" JS
we would just write `foo`, and let it be evaluated as a binding in
the *global scope*. This was however fine as, in ECMAScript 5.1,
the global scope and the global object were basically the same
thing.

However, ECMAScript 2015 changed the situation. There are now
bindings in the global scope that are *not* properties of the
global object. Those are `let`, `const` and `class` bindings at
the top-level of scripts. This means that the previous spec of
Scala.js was utterly incapable of referencing those. They were
totally invisible to Scala.js.

Hence, in this commit we completely change the spec of Scala.js,
and consequently the implementation.

Language changes
================

A reference such as `js.Dynamic.global.foo` is therefore specified
as referring to the binding `foo` in the *global scope* of JS. Note
that it is safer than just saying "it is equivalent to writing
`foo` verbatim in JS", as the latter could collide with local
bindings in the scope chain. The spec protects references to the
global scope from any local definition. Objects annotated with
`@JSGlobalScope` are specified similarly to `js.Dynamic.global`.

Although this trivially solves the problem of accessing `let`,
`const` and `class` bindings, this spec change comes with some
backward source incompatible changes.

Compile-time validation
-----------------------

First, only static bindings (known at compile-time) to variables
whose names are valid JS identifiers are allowed. For example, the
following are legal:

    js.Dynamic.global.foo
    js.Dynamic.global.selectDynamic("bar")

but those are illegal:

    js.Dynamic.global.`not-a-JS-identifier`
    js.Dynamic.global.selectDynamic(computedName)

This restriction directly comes from the fact that only statically
known bindings with valid identifier names can be looked up in the
global scope, even in JavaScript.

The particular identifier `arguments` is similarly forbidden, as
it is a magic variable that automatically lexically binds to the
most nested enclosing function. It cannot be looked up in the
global scope in JavaScript either.

Note that `eval` is allowed, but will systematically be protected
as `(0, eval)` when it is *called*, so as to enforce that the
execution of the script is done at the global level, and does not
see (compiler-generated) local bindings.

The second restriction is that references to the global scope,
either as `js.Dynamic.global` or as `GlobalFoo` where `GlobalFoo`
is annotated with `@JSGlobalScope`, cannot be loaded as a *value*.
For example, the following is illegal:

    val x = js.Dynamic.global

This follows from the previous restriction, as references to the
members of such a reference would have to be looked up dynamically,
and that is not possible with the global scope. More generally, it
makes sense since the "global scope" is not a *value* either in
JavaScript, contrary to the global object.

Run-time semantic changes
-------------------------

In addition to the above compile-time restrictions, the spec change
incurs some differences in run-time semantics, when accessing a
global variable that was not declared. For example:

    val x = js.Dynamic.global.doesNotExist

would previously return `undefined`, but now throws a
`ReferenceError`. Similarly, assignments to non-existent global
variables throw a `ReferenceError`, rather than creating the
variable.

This run-time semantic change is the largest source of actual
breakages in user code, as it invalidates the previous idiom used
to perform feature detection in Scala.js. For example:

    if (js.isUndefined(js.Dynamic.global.Promise))

would previously return `true` when `Promise` was not declared,
but will now throw a `ReferenceError`. The fix is to use
`js.typeOf` instead:

    if (js.typeOf(js.Dynamic.global.Promise) == "undefined")

This brings us to the last semantic change, which is that
`js.typeOf` cannot be understood as a *method* anymore, as it does
not follow the evaluation model of Scala, which guarantees that
`method(expr)` is always equivalent to
`{ val x = expr; method(x) }`. With `js.typeOf(expr)`, if `expr`
is a reference to a non-existent global variable, `"undefined"` is
returned, whereas the rewriting throws a `ReferenceError`.

To allow user code to perform their own global object detection,
should they need to do so, we make the global `this` value
available through `LinkingInfo.globalThis`.

---

Implementation changes
======================

The above changes mostly impact the compiler and the emitter. The
other steps of the pipeline are basically kept intact.

The IR is augmented with an IR node `JSGlobalRef(ident)` which
denotes a reference to a global variable, in the global scope. It
is a valid lhs for the `Assign` node.

The compiler implements the compile-time validation described
above, and emits `JSGlobalRef`s instead of references to
`<linkingInfo>.envInfo.global`. Other than that, nothing changes.

The emitter had to be completely overhauled. The presence of
`JSGlobalRef`s caused significant issues with scoping and naming.

As a naive approach, the translation of `JSGlobalRef(Ident("Foo"))`
is straightforwardly `js.VarRef(js.Ident("Foo"))`. However, this
can cause collisions with two sources of compiler-generated
identifiers:

* Local bindings, such as `VarDef`s and `ParamDef`s.
* Global encoding bindings, such as `$c_sci_List` or
  `$doubleToInt`.

The first group can be dealt with locally in `JSDesugar`. If there
is a reference to a `JSGlobalRef(Ident("Foo"))` in the method, then
a local identifier `Foo` must be renamed. `JSDesugar` proceeds in
two passes to do so, with a mechanism that stops after the first
pass if possible. In the entire test-suite, the second pass is
necessary for only a single method.

The second group is significantly more complicated to address. If
*anywhere in the program*, someone references `$c_sci_List`, then
*everywhere in the program* we need to rename our internal encoding
of `$c_sci_List` into something else (we use `$$c_sci_List`, adding
as many `$`s as necessary to get a non-colliding name).

To do so, `JSDesugar` and `ClassEmitter` keep track of all global
variables that are referenced anywhere, wrapped in `WithGlobals`,
a monadic data structure that eases that propagation. The `Emitter`
invalidates all its caches and re-emits the entire program if the
set of global variables changes in an incremental run, wrt. to the
previous run. In batch-mode, there can therefore be two passes.

To reduce the likelihood of this happening, we eagerly filter out
non "dangerous global refs". A dangerous global ref is one that
starts with a `$` (and is not the single-char string `"$"`).
Indeed, only "dangerous global refs" can collide with our internal
global encodings. In the test-suite, there is not a single
reference to a dangerous global ref, which means that the emitter
can stop after a single pass. We expect most reasonable codebases
to exhibit the same property.

The test for dangerous global refs is therefore put in the
test-suite-ex, so that we do not unnecessarily slow down the
emission of the normal test-suite.


sjrd:
Aannd, it's complete! Review by @gzm0, please.
Note it sits on top of the PR #2942, which should ideally be merged before this one and be rebased on.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/InteroperabilityTest.scala
```diff
@@ -523,6 +523,74 @@ class InteroperabilityTest {
...
+  /** Runs the specified piece of code in the global context.
+   *
+   *  This only works on Node.js. It needs functionality from the `vm` module.
+   *
+   *  This method can be used to declare global let/const/classes. Any other
+   *  attempt to do so (e.g., using a `require`d source file or a `js.eval`)
+   *  would not expose the bindings in the global scope.
+   */
+  private def nodejs_runInThisContext(code: String): Unit = {
+    val vm = js.Dynamic.global.require("vm")
```
sjrd: An interesting side-effect of this new spec is that dynamic references to `require` now actually work :P And also for other magic variables of Node.js such as `module` and `__filename`.

sjrd:
Rebased.

gzm0:
Why do we only need to protect applications to `eval` and not all function application? For example:

```scala
class A {
  def foo() = {
    ja.Dynamic.global.bar()
  }
}
```

Wouldn't now the `this` of `A` leak into `bar`?

sjrd:
> Wouldn't now the `this` of `A` leak into `bar`?

No, because the call would be `bar()` in JS, which, in strict mode (which we always use), binds `this` to `undefined` inside `bar`. (in sloppy mode, `this` is bound to the global object)

> Why do we only need to protect applications to `eval`?

Because `eval` is special. In strict mode,
```js
function foo() {
  "use strict";
  var bar = 42;
  eval("console.log(bar);");
}
```
will execute the eval'ed code *in the local scope* where it is called, resulting in displaying 42. However, the slight variations
```scala
(0, eval)("...")
// or
var anyIdentBut_eval = eval;
anyIdentBut_eval("...")
```
will execute the code *in the global scope*. See http://www.ecma-international.org/ecma-262/6.0/#sec-performeval about the `direct` flag.

In our case, we must always evaluate `eval`ed code in the global scope, because we cannot afford our compiler-generated lexical scope to be visible from the eval'ed code.

sjrd:
Due to the potential conflicts with #2957, I think the present PR should wait until #2957 is merged and reaches `master` before rebasing and merging here. It's easier to fix merge conflicts in a rebase than in a merge of '0.6.x' into 'master'.

sjrd:
Rebased. Ready again.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -199,41 +199,100 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
+          })
+        }
+        slowPath()
+      }
+    }
 
     var syntheticVarCounter: Int = 0
 
     def newSyntheticVar()(implicit pos: Position): Ident = {
+      // TODO Integrate this with proper name management
```
gzm0: Oops :)

sjrd: That one is sort of intended. It's a reminder for later. It's already potentially broken in master/0.6.x. I came across the potential issue while working in that area, and put a comment to revisit this later. But this PR as such does not make it more or less broken than before.

The problem is that this `newSyntheticVar()` kind of assumes that it can generate local identifiers of the form `jsx$i` and that they won't clash with anything else. But in fact they can, in theory, if one uses textually a variable named `jsx$2`, for example, in the .scala source. The solution would be to integrate with the name management introduced in this PR, but then it has to return a `js.Ident` instead of an `Ident`. And that would have significant impacts at its call sites. I'd prefer not to introduce more significant changes in `FunctionEmitter` in this PR.

library/src/main/scala/scala/scalajs/LinkingInfo.scala
```diff
@@ -108,4 +108,35 @@ object LinkingInfo {
...
+   *  }
+   *  }}}
+   *  Note that the above code is not comprehensive, as there can be JavaScript
+   *  environments where the global object cannot be fetched neither through
+   *  `global` nor `this`. If your code needs to run in such an environment, it
+   *  is up to you to use an appropriate detection procedure.
+   */
+  @inline
+  def globalThis: js.Dynamic = linkingInfo.globalThis
+
```
sjrd: Do you object to it being defined in `scalajs.runtime.linkingInfo.globalThis` too, or only in `scalajs.LinkingInfo`? If it can't be in `scalajs.runtime.linkingInfo.globalThis`, then it needs to be yet another primitive with its own IR node. That's only the private API though, so I don't think it's a significant issue.

For the public API:

* `js.Dynamic.globalThis`: that was my first go-to, but I don't like it because it gives too much "visibility" to that method. It should almost never be used. Putting it next to high-profile methods such as `js.Dynamic.global` and `js.Dynamic.literal` feels wrong.
* `js.globalThis`: similar argument. I'm not a big fan of showing it in such a visible part of the API.
* `js.EnvInfo`: I have secret plans to kill `envInfo` altogether. We won't need it now that a) `global` is not there anymore, b) `exportsNamespace` is dying and c) the other things in there can be accessed with `global.__ScalaJSEnv`.

If I had to choose between one of those three, I'd take `js.globalThis`.

gzm0: > Do you object to it being defined in `scalajs.runtime.linkingInfo.globalThis` too?

No, that's fine. Implementation-wise, `globalThis` can be argued to satisfy the conditions for being inside `LinkingInfo` :P

> I have secret plans to kill envInfo altogether. 

I don't fully understand this. AFAICS, there is currently no public `EnvInfo`, only a facade type for documentation (that does not even appear in the doc anymore).

>  but I don't like it because it gives too much "visibility" to that method.

I thought this as well for quite a while, but is it really that bad?

As I am thinking of this, I start wondering: Should `global` itself not be on `js` itself? I would argue that no, because it makes it clearer that code is dynamically typed. Compare:

```scala
js.Dynamic.global.foo(a, b, c)  // oh, foo is dynamic!
js.global.foo(a, b, c)          // it's probably fine if it compiles (!)
js.Promise.all(a, b, c)         // it **is** fine if it compiles.
```

The same argument can be made for `globalThis` of course (and for `constructorOf`, but I recall that there were plans to type it better, though didn't find anything).

There is another option I just thought of: We could start an package `js.completeness` (or something) where we put features we think should rarely / never be used but we provide for completeness. Upside: It is *very* obvious that we don't want people to use it. Downside: I don't think there is anything else upcoming we'll put in there.

sjrd: > I don't fully understand this. AFAICS, there is currently no public EnvInfo, only a facade type for documentation (that does not even appear in the doc anymore).

Right, I meant the private facade type one. But in a sense, if that one disappears, it seems weird we would add now a public `EnvInfo`, which by all expectations will only ever contain `globalThis`.

> I thought this as well for quite a while, but is it really that bad?

I guess not, given that we also have `js.debugger()` there anyway 🙄 

> As I am thinking of this, I start wondering: Should global itself not be on js itself? I would argue that no, because it makes it clearer that code is dynamically typed.

I agree that `global` should not be in `js.global`. `global` is the entry point to the dynamically typed world. It can and should be avoided if the code is completely statically typed.

`globalThis` is slightly different in that respect. Even though it is typed as `js.Dynamic` (and now I'm wondering whether it should be `js.Object` instead, or even `scala.Any`), using it or not is orthogonal to living in a statically or dynamically world. If you need it, you need it, regardless of your typing strategy. That's why I would not put it in `js.Dynamic`.

> There is another option I just thought of: We could start an package js.completeness

I don't like the name but I see the point. Some things that could go there:

* `globalThis`
* `debugger()`
* `delete` (currently in `js.Dictionary`, but that's wrong)
* The primitive for `new.target` #2638.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -260,82 +371,77 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+
+    // Build the result
+
+    new GeneratedClass(
+        main.reverse,
+        staticFields,
+        staticInitialization,
+        classExports,
+        mentionedDangerousGlobalRefs
+    )
```
gzm0: Any reason you are not using the `WithGlobals` monad in this method?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -199,41 +199,100 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
+            while (globalVarNames.contains(result) ||
+                localVarNames.contains(result)) {
+              suffix += 1
+              result = name + "$" + suffix
+            }
+            localVarNames += result
+            result
+          })
+        }
+        slowPath()
```
gzm0: Why the "IIFE"?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -199,41 +199,100 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
+          localVarAllocs.getOrElseUpdate(name, {
+            var suffix = 0
+            var result = name
+            while (globalVarNames.contains(result) ||
+                localVarNames.contains(result)) {
+              suffix += 1
+              result = name + "$" + suffix
+            }
+            localVarNames += result
+            result
```
gzm0: Doesn't this expose the same quadratic behavior you just fixed for `newSyntheticVar`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
gzm0: Filed as #2971.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -278,23 +362,35 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
-      val thisIdent = js.Ident("$thiz", Some("this"))
-      val env = env0.withThisIdent(Some(thisIdent))
-      val js.Function(jsParams, jsBody) =
-        desugarToFunction(params, body, isStat, env)
-      js.Function(js.ParamDef(thisIdent, rest = false) :: jsParams, jsBody)
+      performOptimisticThenPessimisticRuns {
+        /* TODO The identifier `$thiz` cannot be produced by 0.6.x compilers due
+         * to their name mangling, which guarantees that it is unique. We should
+         * find a better way to do this in the future, though.
+         */
```
gzm0: Filed as #2972. IIUC, this is required for correctness. Otherwise someone could try to refer to the global name `$thiz` and boom!

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -237,16 +263,36 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+    @tailrec
+    def slowPath(lastNameTried: String): String = {
+      val nextNameToTry = "$" + lastNameTried
+      if (mentionedDangerousGlobalRefs.contains(nextNameToTry))
+        slowPath(nextNameToTry)
+      else
+        nextNameToTry
+    }
+
+    /* Hopefully this is JIT'ed away as `false` because `envFieldNamesToAvoid`
```
gzm0: Outdated var name in comment.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -147,27 +151,42 @@ private[emitter] final class JSGen(val semantics: Semantics,
   }
 
   def genRawJSClassConstructor(className: String)(
-      implicit globalKnowledge: GlobalKnowledge, pos: Position): Tree = {
+      implicit globalKnowledge: GlobalKnowledge,
+      pos: Position): WithGlobals[Tree] = {
```
gzm0: I would drop this helper.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -147,27 +151,42 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+    genRawJSClassConstructor(className, keepAllGlobalVarNames = true)
+  }
+
+  def genRawJSClassConstructor(className: String,
+      keepAllGlobalVarNames: Boolean)(
```
gzm0: Consider calling the param `keepOnlyDangerousVarNames` for improved readability.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
```diff
@@ -237,16 +263,36 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+
+  def avoidClashWithGlobalRef(envFieldName: String): String = {
+    @tailrec
+    def slowPath(lastNameTried: String): String = {
+      val nextNameToTry = "$" + lastNameTried
+      if (mentionedDangerousGlobalRefs.contains(nextNameToTry))
+        slowPath(nextNameToTry)
+      else
+        nextNameToTry
+    }
```
gzm0: Are you purposefully not caching this?

library/src/main/scala/scala/scalajs/LinkingInfo.scala
sjrd: Yes, I guess we should investigate that. Given the set we see now, perhaps `js.meta._`, since these things are a bit, well, meta. `new.target` is even called a meta-property in the ES spec. I'm afraid `js.meta._` would be confused with `scala.meta`, though, so strategically speaking it's probably not a good idea.

`new.target` I'm afraid will need *both* a primitive (to be called from the constructor of JS classes) and a set of `js.NewTargetFunctionN` (to be called from JS `function`s). The former is the most "critical", though. Using it in `function`s should be pretty niche even compared to using it in classes, which itself should be pretty rare. So we're not a time pressure to come up with a good design for that one.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -769,9 +768,11 @@ abstract class GenJSCode extends plugins.PluginComponent
       val superClass =
         if (sym.isTraitOrInterface) None
         else Some(encodeClassFullNameIdent(sym.superClass))
-      val jsNativeLoadSpec =
+      val jsNativeLoadSpec = {
         if (sym.isTraitOrInterface) None
+        else if (sym.hasAnnotation(JSGlobalScopeAnnotation)) None
         else Some(jsNativeLoadSpecOf(sym))
+      }
```
sjrd: Yes, it is a binary incompatible change. That's expected since it's also a *source* incompatible change now. For example, if I have
```scala
@js.native
@JSGlobal
object Foo extends js.Object
```
and
```scala
val foo = Foo
```
When I change `Foo` to `@JSGlobalScope` the statement `val foo = Foo` becomes illegal.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSGlobalScopeTest.scala
```diff
@@ -0,0 +1,255 @@
...
+      def bracketSelect(name: String): Int = js.native
+      @JSBracketAccess
+      def bracketUpdate(name: String, v: Int): Unit = js.native
+
+      @JSBracketCall
+      def bracketCall(name: String)(arg: Int): Int = js.native
+
+      var arguments: js.Array[Any] = js.native
+      @JSName("arguments") def arguments2(x: Int): Int = js.native
+    }
```
sjrd: It should naturally report
```
error: Selecting a field of the global scope with a dynamic name is not allowed.
```
but indeed, it should be a test case.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSGlobalScopeTest.scala
```diff
@@ -0,0 +1,255 @@
...
+    s"""
+    object Main {
+      def main(): Unit = {
+        val a = js.Dynamic.global.`not-a-valid-identifier-var`
+        js.Dynamic.global.`not-a-valid-identifier-var` = 3
+        val b = js.Dynamic.global.`not-a-valid-identifier-def`()
+
+        val c = SomeGlobalScope.`not-a-valid-identifier-var`
+        SomeGlobalScope.`not-a-valid-identifier-var` = 3
+        val d = SomeGlobalScope.`not-a-valid-identifier-def`()
```
sjrd: Hum ... possibly, but it's only a convenience. We still need to protect against something like
```scala
@js.native
@JSGlobal
class Foo extends js.Object {
  val `not-a-valid-identifier-def`: Int = js.native
}

@js.native
@JSGlobalScope
object FooGlobal extends Foo

val c = FooGlobal.`not-a-valid-identifier-def`
```
At definition site, `not-a-valid-identifier-def` is perfectly valid. But at call site, it cannot be called.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -880,11 +884,11 @@ class PrintersTest {
   @Test def printClassDefJSNativeLoadSpec(): Unit = {
     assertPrintEquals(
         """
-          |native js class LTest extends O loadfrom <global>.Foo {
+          |native js class LTest extends O loadfrom global:Foo.Bar {
```
sjrd: Bof ... if you want. I'm not attached to either.

test-suite-ex/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
```diff
@@ -10,6 +10,7 @@ package org.scalajs.testsuite.utils
 /** This is a partial copy of the implementation in the testSuite */
 object Platform {
 
+  def executingInNodeJS: Boolean = sysProp("nodejs")
   def typedArrays: Boolean = sysProp("typedarray")
 
   private def sysProp(key: String): Boolean =
```
sjrd: That would defeat the purpose of all the other tests in `testSuiteEx`, which fail to fail if they are part of the larger codebase of the test suite. So no, can't do that.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -74,15 +74,14 @@ final class ClosureLinkerBackend(
     val builder = new ClosureAstBuilder(config.relativizeSourceMapBase)
 
     // Build Closure IR
-    logger.time("Emitter (create Closure trees)") {
-      emitter.emit(unit, builder, logger)
+    val coreJSLib = logger.time("Emitter (create Closure trees)") {
+      emitter.emitForClosure(unit, builder, logger)
```
sjrd: I have considered doing that many times over the years. It would be a pretty huge endeavor.  And it would really make the content of `scalajsenv` significantly harder to read and write. I am not sure it would be a win overall.

At least now we're honest with ourselves that the `ClosureLinkerBackend` is a "friend" of `Emitter`. Before it was already the case, but we tried to pretend it was fine with `emitAll`/`emitPrelude`/etc.

We can reconsider later. If at all, IMO it should be done after this PR. IMO our API is cleaner now than before already. I wouldn't call what we have now an "escape hatch" any more than what we had before. And at least now it's `private[backend]`. Before it was public.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -987,13 +1058,15 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
...
-    js.Block(
-        createNamespace,
-        expAccessorVar := methodDef
-    )
+    for (methodDef <- methodDefWithGlobals) yield {
+      js.Block(
+          createNamespace,
+          expAccessorVar := methodDef
+      )
+    }
```
sjrd: So, to be clear, you want a commit with a `WithGlobals` that doesn't actually hold a set of globals? Just a monadic "shell"?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -43,9 +46,45 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
 
-  private val jsGen =
-    new JSGen(semantics, outputMode, moduleKind, internalOptions)
-  private val classEmitter = new ClassEmitter(jsGen)
+  private val baseCoreJSLib = CoreJSLibs.lib(semantics, outputMode, moduleKind)
+
+  private var lastMentionedDangerousGlobalRefs: Set[String] = Set.empty
+  private var jsGen: JSGen = recreateJSGen()
+  private var classEmitter: ClassEmitter = recreateClassEmitter()
+  private var coreJSLib: VirtualJSFile = recreateCoreJSLib()
```
sjrd: Good point.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: Because here nothing is cached. If we use `WithGlobals`, we pay the price of its cost every single time. That's why, in general, I avoid `WithGlobals` in `Emitter.scala`. It also brings much less advantages here. Worse, I expect the code to become *less* readable in this method with `WithGlobals` than without.

Overall, the ration benefit/cost is just not worth it here, like it is in `ClassEmitter`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -226,24 +282,79 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+  private def genAllClasses(orderedClasses: List[LinkedClass], logger: Logger,
+      secondAttempt: Boolean): List[GeneratedClass] = {
+    var mentionedDangerousGlobalRefs: Set[String] = Set.empty
+    val generatedClasses = for (linkedClass <- orderedClasses) yield {
+      val generatedClass = genClass(linkedClass)
+      mentionedDangerousGlobalRefs = unionPreserveEmpty(
+          mentionedDangerousGlobalRefs,
+          generatedClass.mentionedDangerousGlobalRefs)
+      generatedClass
+    }
```
sjrd: Actually, there's much better and shorter:
```scala
val generatedClasses = orderedClasses.map(genClass)
val mentionedDangerousGlobalRefs = generatedClasses.foldLeft(Set.empty[String]) {
  (prev, generatedClass) =>
    unionPreserveEmpty(prev, generatedClass.mentionedDangerousGlobalRefs
}
```

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Indeed it does ^^

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: So that the fast path is small enough for the JVM to inline. Definitely deserve a comment ^^

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Good point.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
sjrd: Yes. It should happen virtually never anyway. And the likelihood that it takes more than 1 iteration is so close to 0 that a cache really doesn't make any sense. It would slow things down.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/WithGlobals.scala
```diff
@@ -0,0 +1,95 @@
...
+ *          -----------------------------------------------
+ *  ... =?= WithGlobals((x => f(x).flatMap(g))(v).value, w ∪ (x => f(x).flatMap(g))(v).globalVarNames)
+ *                      -------------------------------      ----------------------------------------
+ *
+ *  Aside:
+ *    (x => f(x).flatMap(g))(v)
+ *    === f(v).flatMap(g)
+ *    === WithGlobals(g(f(v).value).value, f(v).globalVarNames ∪ g(f(v).value).globalVarNames)
+ *
+ *  ... =?= WithGlobals(g(f(v).value).value, w ∪ (f(v).globalVarNames ∪ g(f(v).value).globalVarNames))
```
sjrd: No. The story went like:
1. I'm writing these `val (tree1, globalVarNames1) = genSomething(); val (tree2, globalVarNames2) = genSomethingElse(); (combineSomehow(tree1, tree2), union(globalVarNames1, globalVarNames2))` everywhere. I need to fix this.
2. Hey, Scala has this cool syntax with for comprehensions. I can use it if I write a data structure with a `map` and a `flatMap`
3. Come up with the implementation of `WithGlobals`. Code becomes clean, me happy.
4. Oh, `WithGlobals` really looks like a monad. Let me document it as such.
5. But wait, if I *do* document it as a monad, but it's not, I'm going to be attacked by an army of cats.
6. The logical reaction would be: don't document it as a monad.
7. My reaction was: let's prove it's a monad.
8. That happened.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
sjrd: IIRC it made the code in `FunctionEmitter` really tedious. I'll try again. And at least put the helper in `FunctionEmitter`.

compiler/src/test/scala/org/scalajs/core/compiler/test/JSGlobalScopeTest.scala
```diff
@@ -0,0 +1,255 @@
...
+      |        SomeGlobalScope.arguments = null
+      |                                  ^
+      |newSource1.scala:36: error: Calling a method of the global scope whose name is `arguments` is not allowed.
+      |  See https://www.scala-js.org/doc/interoperability/global-scope.html for further information.
+      |        val d = SomeGlobalScope.arguments2(5)
+      |                                          ^
+    """
+  }
+
+}
```
sjrd: Good catch. I think it compiles, indeed!

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Although, I think here it's not a big deal. It's quadratic in theory, but in practice you need to try very hard to expose a quadratic behavior.

First, `showPath()` is only called in the pessimistic run. So like 1 in 40,000 methods. OK, that's not a valid explanation, since the `freshName()` problem in the optimizer happened for only 1 method too.

The real explanation is that we're not creating an arbitrary large number of fresh names for the same `base` (here called `name`). We only create one "fresh" name for any given base `name`. The optimizer problem was quadratic because we created thousands of fresh names from the same `base`. Here that cannot happen. Here, if we loop more than once, it's because there is *another* name *in the input* that happens to have `name$1` has its original name. And if that's the case, then creating a fresh name for `name$1` will only loop if there is a `name$1$1`. Which means that, for each iteration of the loop for any given input name `nameFoo`, there must also be an input `nameFoo$1`. I.e., the iterations for `nameFoo` are actually already amortized by the existence of `nameFoo$1` in the input.

So, after all, I believe the complexity of `slowPath()` is amortized O(1).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
gzm0: Oh, sorry. I just meant the one without the default value for `keepAllGlobalVarNames`. `genRawJSClassConstructor` can stay.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
gzm0: Yes. IIUC, `ClassEmitter` "doesn't actually change" in this commit. So if we can put the wiring in another one, we get:

- Guarantee that "nothing changes".
- Less load for someone actually looking at this commit.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/WithGlobals.scala
gzm0: Hahahaha, ok. Consider adding a clarifying comment that it is not really necessary to understand all the m-word stuff to understand the code (I'm absolutely fine with leaving the proof, personally I hate when people just call "everything that has `map` and `flatMap`" monad :P).

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: Ok. So let's leave it like this in this PR. But at the very least, I'd like to have a single abstraction for return values. Currently we have a builder and a return value, that's not very clean. But this can be cleaned up later indeed.

test-suite-ex/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
gzm0: Really? I didn't mean `testSuiteEx.dependsOn(testSuite % "test")`. That would not be a good idea. But `Platfrom` should probably be in `main` anyways, no?

compiler/src/test/scala/org/scalajs/core/compiler/test/JSGlobalScopeTest.scala
gzm0: Arrrgh... Didn't think of this. I almost feel like we should disallow subclassing like this. In fact, it is totally wrong that a `@JSGlobalScope` is a `js.Object`. Ideally it wouldn't have "any type". Maybe we should introduce a detached type, maybe call it `js.GlobalScope` :P

But fair enough, for now, let's add definition site errors for direct definitions and figure this out later.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Ah, right. Fair enough.

test-suite-ex/src/test/scala/org/scalajs/testsuite/utils/Platform.scala
sjrd: Ah I see. Yes that should work.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/JSGen.scala
sjrd: I had understood. ;) Not having the overload with `keepAllGlovalVarNames` defaulting to true is what was tedious.

Not having `genRawJSClassConstructor` at all would be insane.

library/src/main/scala/scala/scalajs/LinkingInfo.scala
sjrd: I have thought about that for the past two days, and also brainstormed a bit with @nicolasstucki this morning.

My proposal would be `js.special._`. It reflects the fact that the tools in there are quite special and should not be used on a regular basis.

It also does not have the connotation of `js.completeness._` that's it's there "for theoretical completeness" only (they are sometimes required in practice). `js.support._` was not bad, but it implies that the tools there are used to build other tools only, which is not the case of `debugger` or `newTarget`. I had discarded `BuildingBlocks` for the same reason (not to mention it's ugly).

compiler/src/test/scala/org/scalajs/core/compiler/test/JSGlobalScopeTest.scala
sjrd: > I almost feel like we should disallow subclassing like this.

That would be extremely inconvenient for something like the DOM global scope, which really *inherits* from `Window`. You would have to duplicate the definition of all members in `Window` and in an `@JSGlobalScope` object.

The global scope does *inherit* from a class, and that includes `js.Object`. However, we cannot load it as a *value*. These things are not contradictory. Since we prevent from loading it as a value, in practice it indeed does not have "any type", since the *subtyping* relationship is always overruled. However, *inheritance* should be allowed.

> let's add definition site errors for direct definitions

I'm not sure. It means strictly more code in `PrepJSInterop` (there is no existing similar code that we can generalize for this) + it would be incomplete anyway (because of inheritance) + it is caught at use site anyway. Doesn't seem like something worth paying the maintenance cost for.

gzm0: Ok, fair enough. I didn't think of it that way. Let's not do anything for now and if too many users hit this issue, we can add a warning that they are creating a definition which is unreachable.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: No in fact it's much more stupid than that. It's obviously not quadratic because this piece of code is guarded by `localVarAllocs.getOrElseUpdate(name, { ... })`. So clearly the loop there is only executed once per input `name`, and not _n_ times as was the case in the optimizer.

The core reason is still what I said: in the optimizer, we create *fresh* names. We invent many *different* fresh names based on the same `base`. Here we only need to allocate one name per input `name`.

It also becomes clear if you consider that the *fix* to the optimizer was to basically add such a `getOrElseUpdate(base, ...)`. There's nothing *more* that we could do here.

sjrd:
Updated.

Given the size of the PR, I have temporarily *added* commits to address review comments, so that you can more easily review the changes. I'll squash them afterwards.

I should have addressed all comments except the separate commit for the introduction of `WithGlobals`. I can hardly do that while there are many small commits for the review comments. So I'll do that after those small commits have been squashed with the main one.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/WithGlobals.scala
```diff
@@ -12,6 +12,14 @@ import GlobalRefUtils.unionPreserveEmpty
 
 /** A monad that associates a set of global variable names to a value.
  *
+ *  This is used to track the set of (dangerous) global variable names used in
+ *  a tree (or a list of trees), and easily perform operations that compose
+ *  such trees while accumulating all the mentioned global variable names.
+ *
+ *  Below follows a proof that `WithGlobals` is indeed a monad. We do not
+ *  actually use that property anywhere. The proof is there only so that I can
+ *  call it a monad and not be attacked by an army of cats.
```
gzm0: :P I approve of this comment.

gzm0:
LGTM for squashing and factoring the monad lifting.

sjrd:
Done. Phew, that wasn't easy ^^

sjrd:
:tada:

PR #2940, opened 2017-05-07
Fix #2939: Fix `%g` formatting for exact powers of 10.

### LINKED ISSUE #2939: `%g` formatting is broken for (some) exact powers of 10
Discovered while investigating https://github.com/scala-js/scala-js/pull/2919#issuecomment-299134553

```scala
println("%g".format(30.0))
println("%g".format(10.0))
```
dispalys
```
30.0000
10.00000
```
See https://scalafiddle.io/sf/ybsj18k/0

The algorithm to compute `sig` seems to be broken when `m` is an exact power of 10. Rereading the algorithm, and the specifications of both Java's `%g` conversion and JS' `toFixed` conversion, I think we want `sig` to be the smallest `k` such that `10^k` is *strictly* greater than `m`. This can be achieved with the following algorithm:
```scala
/* First approximation of the smallest power of 10 that is >= m
 * Due to rounding errors in the event of an imprecise `log10` function,
 * sig0 could actually be the smallest power of 10 that is > m.
 */
val sig0 = Math.ceil(Math.log10(m)).toInt
// Increment sig0 so that it is always the first power of 10 that is > m
val sig = if (Math.pow(10, sig0) <= m) sig0 + 1 else sig0
```
The current algorithm seems like it tries to compute the smallest `k` that is `>= m`. But it fails to do so for some values of `m` because of the imprecision of `Math.log10`, which caused our tests to pass. Improving the precision of `log10` in #2919 uncovered the issue.

sjrd:
The temporary variable `sig` needs to be the smallest value such 10^sig is *strictly* greater than `m`. It used to, sometimes (depending on rounding errors) be such that `10^sig = m`.

### DIFF
--- a/javalib/src/main/scala/java/util/Formatter.scala
+++ b/javalib/src/main/scala/java/util/Formatter.scala
@@ -220,7 +220,16 @@ final class Formatter(private val dest: Appendable) extends Closeable with Flush
                 else precision
               // between 1e-4 and 10e(p): display as fixed
               if (m >= 1e-4 && m < Math.pow(10, p)) {
-                val sig = Math.ceil(Math.log10(m)).toInt
+                /* First approximation of the smallest power of 10 that is >= m.
+                 * Due to rounding errors in the event of an imprecise `log10`
+                 * function, sig0 could actually be the smallest power of 10
+                 * that is > m.
+                 */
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @mkotsbak?

sjrd:
Thanks.

PR #2946, opened 2017-05-12
Pass `remoteArgs` to `Slave` in test interface

Duhemm:
I guess this was a typo.

### DIFF
--- a/test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSRunner.scala
+++ b/test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSRunner.scala
@@ -193,7 +193,7 @@ final class ScalaJSRunner private[testadapter] (
     val prefix = framework.optionalExportsNamespacePrefix
     val frameworkJS = jsonToString(framework.frameworkName.toJSON)
     val argsJS = jsonToString(args.toList.toJSON)
-    val remoteArgsJS = jsonToString(args.toList.toJSON)
+    val remoteArgsJS = jsonToString(remoteArgs.toList.toJSON)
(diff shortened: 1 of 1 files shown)


sjrd:

sjrd:
I'll have a look sometime during the week. Probably a typo indeed, but I'd like to understand this `remoteArgs` in the process.

gzm0:
> Probably a typo indeed, but I'd like to understand this `remoteArgs` in the process.

I'd like to understand `remoteArgs` as well :P Feel free to investigate more. From my side, this LG.

gzm0:
Also: This should probably target 0.6.x?

sjrd:
Ah yes, it should definitely target 0.6.x. @Duhemm could you rebase on top of the `0.6.x` branch and change the target branch of this PR please?

Duhemm:
@sjrd Done.

sjrd:
Thanks. Jenkins, retest this please

sjrd:
The best lead I found was in ScalaTest's codebase: https://github.com/scalatest/scalatest/blob/d52e2a2b7c49f988c8ed3a5728faec7c641be03e/scalatest/src/main/scala/org/scalatest/tools/Framework.scala#L904
Which suggests that this is used when testing is done in a forked JVM. Another interesting place to look at is https://github.com/scalatest/scalatest/blob/d52e2a2b7c49f988c8ed3a5728faec7c641be03e/scalatest/src/main/scala/org/scalatest/tools/Framework.scala#L770

All of that is probably irrelevant in Scala.js, since we don't support the fork mode (or rather, we only have a fork mode but with our own serialization mechanisms under the hood).

PR #2955, opened 2017-05-19
Fix 2 bugs in the codegen for pattern matches

sjrd:


compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2955,9 +2960,10 @@ abstract class GenJSCode extends plugins.PluginComponent
           translateMatch(expr)
 
         // Sometimes the pattern matcher casts its final result
-        case Apply(TypeApply(Select(expr: LabelDef, nme.asInstanceOf_Ob), _), _)
+        case Apply(TypeApply(Select(expr: LabelDef, nme.asInstanceOf_Ob), targs), _)
```
gzm0: Could you replace `targs` by `List(targ)` just to be extra prudent? (maybe also the last `_` by `Nil`)?

sjrd: Done.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -384,6 +384,33 @@ class RegressionTest {
     assertThrows(classOf[MatchError], bug.bug(2, false))
   }
 
+  @Test def return_x_match_issue_2928(): Unit = {
+    def testNonUnit(x: String): Boolean = {
+      return x match {
+        case "True" => true
```
SethTisue, line 390: @sjrd note that on Scala 2.13.8, matching on a `String` like this doesn't result in the weird `Return(LabelDef(...))` tree — instead you get a normal `Return(Match(...)` tree, because matching on string constants is now handled differently. I think the PR when this changed is https://github.com/scala/scala/pull/8451 ? So if you want this to actually be testing the change on Scala 2.13, I think you need to e.g. change `x: String` to `x: AnyRef`.

I hit this when trying to fix the same bug in my Fortify plugin.

sjrd: Ah, good catch! We should change the rest to use some other data type.

The good news is that this problem will be entirely gone in 2.13.9+, thanks to https://github.com/scala/scala/pull/10028

SethTisue: Oh, that's good to know 👍

sjrd: Eh eh, looks like we already added a test for that when we added support for `Match` nodes with strings in
https://github.com/scala-js/scala-js/commit/24a22d1bb43378eb1f6312e2e1a8be2adcfb8c52
I tried to unfix the fix and it does crash on that test case in 2.13.8.

No idea how I thought about that when I made that commit. 😅

PR #2956, opened 2017-05-19
Fix #2953: Avoid calling toString() in primitive isInstanceOf's.

### LINKED ISSUE #2953: Type-matching arbitrary things against Byte/Short/Int ends up calling toString
```scala
object Walker{
  def treeify(x: Any): Unit = {
    println("Treeify!")
    x match{
      case x: Short => println("short")
      case x: Int => println("int")
      case x: Iterable[_] => println("iterable ")
    }
  }
}

class C(){
  var counter = 0
  override def toString = {
    counter += 1
    println("increment! " + counter)
    "C"
  }
}

Walker.treeify(List.fill(4)(new C))
// Treeify!
// increment! 1
// increment! 1
// increment! 1
// increment! 1
// increment! 2
// increment! 2
// increment! 2
// increment! 2
// iterable
```
https://scalafiddle.io/sf/OkgMiGx/0

It seems that the `x: Short =>` and `x: Int =>` cases are calling `.toString` on my Cs
`x: Byte =>` is doing it too (adding that makes the increments go up to 3) but Float, Double and Char don't seem to do it
on the JVM the increment messages don't appear at all

> Using arithmetic operators causes to `ToPrimitive(x, "number")` internal function of the JS spec to be called, which, in the absence of a valueOf method, will fall back on calling toString().
> We must assert that `typeof v === "number"` before trying any operator.
> - sjrd

sjrd:
Even though `(v | 0) === v` will indeed return `false` when `v` is not even a `number`, the part `v | 0` will cause the internal spec function `ToPrimitive(v, hint Number)` to be called. That function (section 7.1.1 of the ES 2015 specification) will first try to call `v.valueOf()` if it exists, and otherwise will try `v.toString()`.

We must avoid this scenario by preemptively ruling out values that are not `number`s, which we do with a simple `typeof` test.

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
@@ -535,6 +535,28 @@ class RegressionTest {
     assertTrue(f3A != f4B)
   }
 
+  @Test def isInstanceOf_must_not_call_toString_issue_2953(): Unit = {
+    class C {
+      override def toString(): String =
+        throw new AssertionError("C.toString must not be called by isInstanceOf")
+    }
+
+    @noinline def makeC(): Any = new C
+
+    val c = makeC()
+
+    assertFalse("Boolean", c.isInstanceOf[Boolean])
+    assertFalse("Char", c.isInstanceOf[Char])
+    assertFalse("Byte", c.isInstanceOf[Byte])
+    assertFalse("Short", c.isInstanceOf[Short])
+    assertFalse("Int", c.isInstanceOf[Int])
+    assertFalse("Long", c.isInstanceOf[Long])
+    assertFalse("Float", c.isInstanceOf[Float])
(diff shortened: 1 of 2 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -535,6 +535,31 @@ class RegressionTest {
...
+      override def toString(): String =
+        throw new AssertionError("C.toString must not be called by isInstanceOf")
+    }
+
+    @noinline def makeC(): Any = new C
+
+    val c = makeC()
+
+    def failMatch(): Nothing =
+      throw new AssertionError("should not have matched")
```
gzm0: I think this is unused

sjrd: Duh! Indeed.

sjrd:
Updated.

PR #2957, opened 2017-05-20
Fix #2797: Add an optional global fallback to `@JSImport`.

### LINKED ISSUE #2797: Facades should be able to support all values of ModuleKind

sjrd:
If the codebase refers to an `@JSImport` entity, such as

```scala
@js.native
@JSImport("foo.js", "Bar")
class Bar extends js.Object
```

the codebase must be linked with `CommonJSModule` (or, in the future, other `ModuleKind`s supporting modules). Linking without module support causes a linking error.

This causes facades for JS libraries supporting both module-style and script-style to be faced with an early choice:

* either support linking with modules and use `@JSImport`, or
* support linking without module support, and use `@JSGlobal` with whatever global names the JS library uses.

It is however impossible to write a facade library that supports both use cases for their end users.

This commit addresses this issue by adding an optional "global fallback" to `@JSImport`. We can now define a facade as follows:

```scala
@js.native
@JSImport("foo.js", "Bar", globalFallback = "Foo.Bar")
class Bar extends js.Object
```

That facade will successfully link both with and without module support. With module support, it links as if declared like in the first snippet. Without module support, it links as if declared as

```scala
@js.native
@JSGlobal("Foo.Bar")
class Bar extends js.Object
```

Using this global fallback, a facade library can cater both for users who want to take advantages of modules, and users who want to stick to the script style.

---

Implementation-wise, this is quite easy to implement. We simply add a new `JSNativeLoadSpec.ImportWithGlobalFallback`. It wraps both a `JSNativeLoadSpec.Import` and `Global`. That JS native load spec goes through the entire pipeline and reaches the emitter as is. The emitter decides which one to use depending on the `moduleKind`.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ModulesWithGlobalFallbackTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ModulesWithGlobalFallbackTest.scala
@@ -1,6 +1,6 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
-**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2016, LAMP/EPFL   **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2017, LAMP/EPFL   **
 **  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
 ** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
 **                          |/____/                                     **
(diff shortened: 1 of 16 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ModulesWithGlobalFallbackTest.scala
```diff
@@ -0,0 +1,205 @@
...
+        areTypedArraysSupported)
+
+    js.Dynamic.global.ModulesWithGlobalFallbackTest_QueryString =
+      QueryStringFallbackImpl
+    js.Dynamic.global.ModulesWithGlobalFallbackTest_StringDecoder =
+      js.constructorOf[StringDecoderFallbackImpl]
+    js.Dynamic.global.ModulesWithGlobalFallbackTest_Buffer =
+      js.constructorOf[Uint8Array]
+    js.Dynamic.global.ModulesWithGlobalFallbackTest_BufferStatic =
+      BufferStaticFallbackImpl
```
sjrd: Because that crashes too early (eagerly, before the `assume` can kick in) if typed arrays are not supported.

gzm0: I assume you have tested this, but I can't see why (except for `ModulesWithGlobalFallbackTest_Buffer`, but that one is also not defined in this file).

gzm0: Actually, could you only assign these when we are not linking in a module? Currently we're not testing that module import has precedence over global.

sjrd: Ah yes that would be good indeed. I'll have to add a new flag set from the build, but yes, that can be done.

sjrd: Done.

PR #2958, opened 2017-05-20
Fix #2927: Fail on assertEquals on floats/doubles

### LINKED ISSUE #2927: JUnit assertEquals(_: Double, _:Double) not failing for JS, only for JVM
For JVM it gives:

`Use assertEquals(expected, actual, delta) to compare floating-point numbers`

but it does not fail for the Scala.Js version.

gzm0:
#2960 needs to go in first and the issues discussed in https://github.com/scala-js/scala-js/issues/2927#issuecomment-302926675 need to be solved.

### DIFF
--- a/junit-runtime/src/main/scala/org/junit/Assert.scala
+++ b/junit-runtime/src/main/scala/org/junit/Assert.scala
@@ -91,6 +91,26 @@ object Assert {
   def assertNotEquals(unexpected: Float, actual: Float, delta: Float): Unit =
     assertNotEquals(null, unexpected, actual, delta)
 
+  @deprecated("Use assertEquals(double expected, double actual, double " +
+      "epsilon) instead", "")
+  def assertEquals(expected: Double, actual: Double): Unit = {
+    fail("Use assertEquals(expected, actual, delta) to compare " +
+        "floating-point numbers")
+  }
+
+  @deprecated("Use assertEquals(String message, double expected, double " +
+      "actual, double epsilon) instead", "")
+  def assertEquals(message: String, expected: Double, actual: Double): Unit = {
+    fail("Use assertEquals(expected, actual, delta) to compare " +
+        "floating-point numbers")
+  }
+
+  def assertEquals(expected: Long, actual: Long): Unit =
+    assertEquals(null, expected, actual)
+
+  def assertEquals(message: String, expected: Long, actual: Long): Unit =
+    assertEquals(message, expected: Any, actual: Any)
+
   def assertArrayEquals(message: String, expecteds: Array[AnyRef],
       actuals: Array[AnyRef]): Unit = {
     internalArrayEquals(message, expecteds, actuals)
--- a/junit-test/shared/src/test/scala/org/scalajs/junit/AssertEqualsDoubleTest.scala
+++ b/junit-test/shared/src/test/scala/org/scalajs/junit/AssertEqualsDoubleTest.scala
@@ -1,0 +1,50 @@
+package org.scalajs.junit
+
+import org.junit.Assert._
+import org.junit.Test
+
+import org.scalajs.junit.utils._
+
+class AssertEqualsDoubleTest {
+  @Test def failsWithDouble(): Unit = {
+    assertEquals(1.0, 1.0)
+  }
+
+  @Test def failsWithDoubleMessage(): Unit = {
+    assertEquals("Message", 1.0, 1.0)
+  }
+
+  @Test def worksWithEpsilon(): Unit = {
+    assertEquals(1.0, 1.0, 0.1)
+    assertEquals("Message", 1.0, 1.0, 0.1)
+  }
+
+  @Test def worksWithByte(): Unit = {
+    // This is supposed to take the (long, long) overload.
+    assertEquals(1.toByte, 1.toByte)
+  }
+
+  @Test def worksWithShort(): Unit = {
+    // This is supposed to take the (long, long) overload.
+    assertEquals(2.toShort, 2.toShort)
+  }
+
+  @Test def worksWithInt(): Unit = {
+    // This is supposed to take the (long, long) overload.
+    assertEquals(1, 1)
+  }
+}
+
+class AssertEqualsDoubleTestAssertions extends JUnitTest {
+  protected def expectedOutput(builder: OutputBuilder): OutputBuilder = {
+    builder
+      .success("worksWithEpsilon")
+      .assertion("failsWithDouble",
+          "Use assertEquals(expected, actual, delta) to compare floating-point numbers")
+      .assertion("failsWithDoubleMessage",
+          "Use assertEquals(expected, actual, delta) to compare floating-point numbers")
+      .success("worksWithByte")
+      .success("worksWithShort")
+      .success("worksWithInt")
+  }
+}
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1167,7 +1167,7 @@ object Build {
       fatalWarningsSettings ++ Seq(name := "Scala.js JUnit test runtime")
   ).withScalaJSCompiler.dependsOn(testInterface)
 
-  val commonJUnitTestOutputsSettings = commonSettings ++ fatalWarningsSettings ++ Seq(
+  val commonJUnitTestOutputsSettings = commonSettings ++ Seq(
       publishArtifact in Compile := false,
       parallelExecution in Test := false,
       unmanagedSourceDirectories in Test +=
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/FloatJSTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/FloatJSTest.scala
@@ -21,25 +21,25 @@ class FloatJSTest {
 
   @Test def fround_for_special_values(): Unit = {
     assertTrue(froundNotInlined(Double.NaN).isNaN)
-    assertEquals(Double.PositiveInfinity, 1 / froundNotInlined(0.0).toDouble)
-    assertEquals(Double.NegativeInfinity, 1 / froundNotInlined(-0.0).toDouble)
-    assertEquals(Float.PositiveInfinity, froundNotInlined(Double.PositiveInfinity))
-    assertEquals(Float.NegativeInfinity, froundNotInlined(Double.NegativeInfinity))
+    assertEquals(Double.PositiveInfinity, 1 / froundNotInlined(0.0).toDouble, 0.0)
+    assertEquals(Double.NegativeInfinity, 1 / froundNotInlined(-0.0).toDouble, 0.0)
+    assertEquals(Float.PositiveInfinity, froundNotInlined(Double.PositiveInfinity), 0.0)
+    assertEquals(Float.NegativeInfinity, froundNotInlined(Double.NegativeInfinity), 0.0)
   }
 
   @Test def fround_overflows(): Unit = {
-    assertEquals(Double.PositiveInfinity, froundNotInlined(1e200))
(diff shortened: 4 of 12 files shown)


junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -47,35 +55,131 @@ abstract class JUnitTest {
...
+    lazy val logExceptionClass: Boolean = !args.contains("-c")
+    lazy val logAssert: Boolean = args.contains("-a")
 
-    lazy val verbose: Boolean = frameworkArgs.contains("-v")
-    lazy val noColor: Boolean = frameworkArgs.contains("-n")
-    lazy val quiet: Boolean = frameworkArgs.contains("-q")
-    lazy val decodeScalaNames: Boolean = frameworkArgs.contains("-s")
-    lazy val notLogExceptionClass: Boolean = frameworkArgs.contains("-c")
-    lazy val logAssert: Boolean = frameworkArgs.contains("-a")
+    override def toString: String = args.mkString("[", ", ", "]")
```
sjrd: `toString()` with `()`, as per https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#method-definition

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -47,35 +55,131 @@ abstract class JUnitTest {
...
+        testIgnoredOutput(testName),
+        skippedEvent
+    )
+
+    def ignoredClass(): OutputBuilder = append(0, 1, 0)(
+        testIgnoredClassOutput,
+        skippedEvent
+    )
+
+    def assumptionViolated(testName: String) = append(1, 0, 0)(
```
sjrd: Public method needs explicit result type.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -90,27 +194,26 @@ abstract class JUnitTest {
     def testIgnoredClassOutput: Output =
       Info(s"Test $formattedTestClass ignored")
 
-    def testExceptionMsgOutput(method: String, msg: String, exPack: String,
+    def testExceptionMsgOutput(method: String, msg: String,
         exClass: String): Output = {
-      val exClassStr =
-        if (notLogExceptionClass) ""
-        else ' ' + exPack + '.' + red(exClass) + ':'
+      val exClassStr = exceptionClassInfo(logExceptionClass,  exClass)
```
sjrd: Double space after `,`

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -137,12 +241,22 @@ abstract class JUnitTest {
...
+    def formattedTestClass: String = formatClass(suiteUnderTestName, yellow)
+
+    private def exceptionClassInfo(show: Boolean, fullName: String) = {
+      if (show) formatClass(fullName, red) + ": "
+      else ""
+    }
+
+    private def formatClass(fullName: String, color: String => String) = {
+      val (pack, cls) = fullName.splitAt(fullName.lastIndexOf('.'))
+      pack + '.' + color(cls.tail)
```
sjrd: If you split right after the dot rather than right before it, it's a bit easier:
```scala
val (packAndDot, cls) = fullName.splitAt(fullName.lastIndexOf('.') + 1)
packAndDot + color(cls)
```

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -47,35 +55,131 @@ abstract class JUnitTest {
...
+    private def append(t: Int, i: Int, f: Int)(out: Output*) = {
+      new OutputBuilder(argInfo, total + t, ignored + i, failed + f, output ++ out)
+    }
+
+    // Test method.
+
+    private[JUnitTest] def checkOutput(actual: List[Output]): Unit = {
+      val expected = (
+          testRunStartedOutput :: output :::
+          List(testRunFinishedOutput(total, ignored, failed), done)
```
sjrd: Consider using `::`s all the way, instead of using the `List()` constructor for part of it. Maybe format it as follows?
```scala
      val expected = (
          testRunStartedOutput ::
          output :::
          testRunFinishedOutput(total, ignored, failed) ::
          done ::
          Nil
      )
```

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -47,35 +55,131 @@ abstract class JUnitTest {
...
+        appendElems(actual, actualMinimized)
+        msg.append(")")
+  
+        throw new Exception(msg.result())
+      }
+    }
+
+    // Text builders.
 
     def done: Output = Done("")
```
sjrd: Looks like the methods returning `Output`s could become private, couldn't they? Maybe even `Output` itself can become private?

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -90,27 +194,26 @@ abstract class JUnitTest {
...
-          s"failed:$exClassStr $msg, took $TIME_TAG sec")
+          s"failed: $exClassStr$msg, took $TIME_TAG sec")
     }
 
     def testAssertionErrorMsgOutput(method: String, msg: String): Output = {
-      val assertClass =
-        if (notLogExceptionClass || !logAssert) ""
-        else  " java.lang." + red("AssertionError") + ':'
-      Error(s"Test $formattedTestClass.${red(method)} failed:$assertClass " +
+      val assertClass = exceptionClassInfo(logExceptionClass && logAssert,
```
sjrd: Consider explicitly writing `show =`.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -90,27 +194,26 @@ abstract class JUnitTest {
...
+          "java.lang.AssertionError")
+      Error(s"Test $formattedTestClass.${red(method)} failed: $assertClass" +
           s"$msg, took $TIME_TAG sec")
     }
 
     def testAssumptionViolatedOutput(method: String): Output = {
+      val exceptionStr = exceptionClassInfo(logExceptionClass,
```
sjrd: Consider explicitly writing `show =`.

junit-test/shared/src/test/scala/org/scalajs/junit/TestTest.scala
```diff
@@ -1,25 +0,0 @@
...
-    import context._
-    List(
-        testRunStartedOutput,
-        testStartedOutput("onlyTest"),
-        testFinishedOutput("onlyTest"),
-        successEvent,
-        testRunFinishedOutput,
-        done
-    )
-  }
```
sjrd: Why is this test removed?

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -130,55 +130,94 @@ abstract class JUnitTest {
...
+        }
+
+        msg.append("but got: List(\n")
+
+        val maxLen = expecteds.map(_.size).max
+
+        matched.takeRight(3).foreach(out =>
+          msg.append(GREY + "  " + withoutColor(out.toString) + NORMAL + "\n"))
+        remaining.take(maxLen).foreach(appendOut)
         msg.append(")")
```
sjrd: What does this look like, once rendered?

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -130,55 +130,94 @@ abstract class JUnitTest {
...
+          expected.foreach(appendOut)
+          msg.append(")\n\n")
+        }
+
+        msg.append("but got: List(\n")
+
+        val maxLen = expecteds.map(_.size).max
+
+        matched.takeRight(3).foreach(out =>
+          msg.append(GREY + "  " + withoutColor(out.toString) + NORMAL + "\n"))
```
sjrd: Consider a for:
```scala
for (out <- matched.takeRight(3))
  msg.append(GREY + "  " + withoutColor(out.toString) + NORMAL + "\n")
```
According to https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#method-call-style, the `()`+`=>` association only works if it fits entirely on the same line.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
sjrd: Well, since that disappears in a subsequent commit anyway, never mind that comment.

junit-test/shared/src/test/scala/org/scalajs/junit/TestTest.scala
gzm0: It is useless because:

- Testing a tests is something that is rarely done (when do you stop?)
- It doesn't even do this. Testing a test would mean to verify that a test fails if the code under test doesn't satisfy the conditions. This test definitely doesn't do that.
- Other tests test this as well.

I'm having the impression that this is just a relic from when the JUnit testing harness got bootstrapped.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
gzm0: Hum... This swallows the first letter of `cls`...

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
gzm0: This is what it looks when I change an `assertTrue(false)` to `assertTrue(true)`.

screen shot 2017-05-21 at 1 55 55 am

gzm0:
Updated. I still need to fix all the places in our test suite where we use the `assertEquals` that now fails.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
sjrd: Huh!? Did you forget to remove the `.tail` of `cls.tail`? If not I don't know why my suggestion is wrong :s

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
sjrd: Ah nice :)

junit-test/shared/src/test/scala/org/scalajs/junit/TestTest.scala
sjrd: OK makes sense.

junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
gzm0: Duh... of course I did :P

gzm0:
This is ready for review now.

gzm0:
Updated (adjusted failing tests).

PR #2959, opened 2017-05-20
Fix #2948: Remove js.use(x).as[T]

### LINKED ISSUE #2948: Remove js.use(x).as[T]
There are 3 good reasons for that:

* It definitely *can* be implemented and tested outside of the core, so it should.
* With Scala.js-defined JS classes, this feature has lost a lot of its usefulness (GitHub reports *2* occurrences of `js.use` in .scala files in the wild, and one of them should really be a Scala.js-defined JS class).
* It is the only macro in the core library, so getting rid of it means getting rid of all macros in the core lib.

gzm0:


sjrd:
+0 -1,325 .... lol

gzm0:
Well, let's see if it compiles :P

PR #2947, opened 2017-05-12
Separate all jsDependencies-related things in separate projects.

sjrd:
This is part of #2841.

* The contents of `org.scalajs.core.tools.jsdep._` are extracted in a separate project `jsdependencies-core`.
* All the settings related to `jsDependencies` are extracted in a separate sbt plugin `JSDependenciesPlugin` in `jsdependencies-plugin`.

PR #2961, opened 2017-05-21
Fix #2839: Remove PhantomJS-related stuff in the core repo.

### LINKED ISSUE #2839: Move PhantomJS out of the core
`PhantomJSEnv` should go in a separate library, just like Selenium. Unlike Selenium, though, it requires special help from the sbt plugin to resolve jetty8, so we need to also publish an accompanying sbt plugin to ease the use site of `PhantomJSEnv`.

Basically the same API should be supported, but would require an additional `addSbtPlugin(...)` in builds that want to use PhantomJS.

We can do this in two steps:

- [x] Logically extract the parts of the sbt plugin that deal with PhantomJS in a separate `AutoPlugin`, but still in the same repo
- [ ] Physically move `PhantomJSEnv` and that new `AutoPlugin` to a separate repo

sjrd:
PhantomJS support moved in the separate repository https://github.com/scala-js/scala-js-env-phantomjs.

It seems #2947 really wants this to happen first. It has such a high percentage of spurious failures related to PhantomJS.


gzm0:
Woot!!

Issue #2827, opened 2017-03-23
Slow incremental linking times for fastOptJS


olafurpg:
Steps to reproduce

- clone https://github.com/olafurpg/scalameta/tree/scala.js-slow-fastopjs
- make sure you are on commit https://github.com/olafurpg/scalameta/commit/f1d0149c4cd061ca13854ed5475e7bbb3d8a255a
- `sbt scalametaJS/test:fastOptJS` (takes a long time, as expected)
- Add blank line **inside a `test` block** in PublicSuite, for example line 18
- `sbt scalametaJS/test:fastOptJS` (takes ~16 seconds to optimize 9 methods, unexpectedly long)

```
[scala.js-slow-fastopjs] scalameta: last scalametaJS/test:fastOptJS
[info] Fast optimizing /Users/ollie/dev/scalameta/scalameta/scalameta/js/target/scala-2.11/scalameta-test-fastopt.js
[debug] Linker: Compute reachability: 2365588 us
[debug] Linker: Assemble LinkedClasses: 481434 us
[debug] Basic Linking: 2868037 us
[debug] Inc. optimizer: Batch mode: false
[debug] Inc. optimizer: Incremental part: 303286 us
[debug] Inc. optimizer: Optimizing 9 methods.
[debug] Inc. optimizer: Optimizer part: 16756267 us
[debug] Inc. optimizer: 17319874 us
[debug] Refiner: Compute reachability: 1456534 us
[debug] Refiner: Assemble LinkedClasses: 65163 us
[debug] Refiner: 1521887 us
[debug] Emitter: Class tree cache stats: reused: 6059 -- invalidated: 3
[debug] Emitter: Method tree cache stats: resued: 15877 -- invalidated: 7
[debug] Emitter (write output): 1220303 us
[debug] Global IR cache stats: reused: 3575 -- invalidated: 3 -- trees read: 3
```

Hardware: 
- MacBook Pro (Retina, 15-inch, Mid 2014)
- Processor: 2,5 GHz Intel Core i7
- Memory: 16 GB 1600 MHz DDR3

olafurpg:
Here's the results from a cleaner fastOptJS (26k methods). Note that the "Inc. optimizer: Optimizer part" takes close to 5 minutes.

```
[scala.js] scalameta: last scalametaJS/test:fastOptJS
[info] Fast optimizing /Users/ollie/dev/scalameta/scalameta/scalameta/js/target/scala-2.11/scalameta-test-fastopt.js
[...]

PR #2964, opened 2017-05-23
Fix #2943: Fix quadratic behaviors in the optimizer and cap JS array virtualization

### LINKED ISSUE #2943: Regression in optimiser
I believe there is a regression in the optimiser of Scala.js v0.6.16. When upgrading to this version and running my test cases, I encountered an OOM on Travis-CI: https://travis-ci.org/sparsetech/pine/jobs/230421157

Locally, it does not compile the test cases, getting stuck at "Fast optimizing". It doesn't finish even after a couple of minutes. `publishLocal` works fine and I did not notice any regressions in the artifact's functionality.

I investigated this issue as per @ochrons' suggestion by looking at the activity in JVisualVM. Here are the results:

jvisualvm

Please let me know how I can further assist in tracking down this issue!

sjrd:
The first two commits do not change the emitted .js file for the test suite (not even by different names being used). The last commit slightly changes the generated code for [the test data array of `Base64Test`](https://github.com/scala-js/scala-js/blob/195452a76760584338949229a85c0e0f9301058b/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/Base64Test.scala#L366).

In addition to the improvements seen on the pathological case of #2943, the first commit cuts the time it takes to optimize the test suite from 14 seconds down to ~3.7 s. The other two commits do not bring any measurable difference on the test suite.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -80,20 +80,22 @@ private[optimizer] abstract class OptimizerCore(
 
   private val labelNameAllocator: FreshNameAllocator = new FreshNameAllocator
 
-  /** A list of the States that have been allocated so far, and must be saved.
+  /** A list of backups for all updates done to States so far (excluding
+   *  those done in rolled back optimistic branches).
    *
-   *  This list only ever grows, even though, in theory, it will keep
-   *  references to states that are not used anymore.
-   *  This creates a "temporary memory leak", but the list is discarded when
-   *  `optimize` terminates anyway because the whole OptimizerCore is discarded.
-   *  It also means that RollbackException will save more state than strictly
-   *  necessary, but it is not incorrect to do so.
+   *  This list grows (from the head) every time the value of a `State` changes.
+   *  Each time, a `StateBackup` is prepended with the previous value.
    *
-   *  Manual "memory management" of this list has caused issues such as #1515
-   *  and #1843 in the past. So now we just let it grow in a "region-allocated"
-   *  style of memory management.
+   *  When starting an optimistic branch in `tryOrRollback`, we take a snapshot
+   *  of the current chain of backups. When doing a rollback, we restore all
+   *  the backups that have been added to the chain since the snapshot. We can
+   *  do this by comparing the nodes of the chain with `eq`.
+   *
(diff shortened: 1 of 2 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -80,20 +80,22 @@ private[optimizer] abstract class OptimizerCore(
 
   private val labelNameAllocator: FreshNameAllocator = new FreshNameAllocator
 
-  /** A list of the States that have been allocated so far, and must be saved.
+  /** A list of backups for all updates done to States so far (excluding
+   *  those done is rolled back optimistic branches).
```
gzm0: is -> in

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -5002,4 +4991,45 @@ private[optimizer] object OptimizerCore {
...
+        base
+      } else {
+        var i = usedNamesToNextCounter(base)
+        var result = base + "$" + i
+        while (usedNamesToNextCounter.contains(result)) {
+          i += 1
+          result = base + "$" + i
+        }
+        usedNamesToNextCounter =
+          usedNamesToNextCounter.updated(base, i + 1).updated(result, 1)
```
sjrd: `freshName` is very often called with a non-constant base, coming from the input IR. The archetypal example is [in `withNewLocalDef`](https://github.com/sjrd/scala-js/blob/84fc283c005e5fcfaf774b90f466edf7b4f75739/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala#L3992). So we cannot rule out `$`s here, unfortunately.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Done. (also a couple typo fixes in the commit messages)

sjrd:
I'm so glad I took this Advanced Algorithms course at EPFL 4 years ago! I would never have thought of this amortized solution, otherwise.

Btw, this also seems to fix #2827 to a large extent.

PR #2966, opened 2017-05-26
Fix #2921: Forbid SAMs of JS types (except js.{This,}Function).

### LINKED ISSUE #2921: SAM for a native JS type (other than js.{,This}FunctionN) incorrectly compiles
The following should not compile in 2.12:
```scala
object Main extends js.JSApp {
  def main(): Unit = {
    val varArgsFun: DependencyArgsToResult[Any, Any] = { (args) =>
      ()
    }
  }
}

@js.native
trait DependencyArgsToResult[-VarArgType, +ResultType] extends js.Object {
  def apply(args: VarArgType*): ResultType
}
```
It currently produces invalid IR that crashes the linker. It should fail to compile, just like the following snippet:
```scala
object Main extends js.JSApp {
  def main(): Unit = {
    val varArgsFun = new DependencyArgsToResult[Any, Any] {
      def apply(args: Any*): Any = ()
    }
  }
}
```
```
[error] /localhome/doeraene/projects/scalajs/examples/helloworld/HelloWorld.scala:16: A Scala.js-defined JS class cannot directly extend a native JS trait.
[error]     val varArgsFun = new DependencyArgsToResult[Any, Any] {
[error]                          ^
[error] /localhome/doeraene/projects/scalajs/examples/helloworld/HelloWorld.scala:17: A Scala.js-defined JavaScript class cannot declare a method named `apply` without `@JSName`
[error]       def apply(args: Any*): Any = ()
[error]           ^
[error] two errors found
```

sjrd:
Since using such a SAM before would produce invalid IR that sent the optimizer crashing, we directly make this into an *error*, rather than a warning.

In theory, we could support SAMs of Scala.js-defined JS types, if they do not have an `apply` method. But that would require more work in the compiler backend. This commit does not attempt to do so.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -246,6 +246,20 @@ abstract class PrepJSInterop extends plugins.PluginComponent
           }
           super.transform(tree)
 
+        /* Anonymous function, need to check that it is not used as a SAM for a
+         * JS type, unless it is js.FunctionN or js.ThisFunctionN.
+         * See #2921.
+         */
+        case tree: Function =>
+          val tpeSym = tree.tpe.typeSymbol
+          if (isJSAny(tpeSym) && !AllJSFunctionClasses.contains(tpeSym)) {
+            reporter.error(tree.pos,
+                "Using an anonymous function as a SAM for the JavaScript " +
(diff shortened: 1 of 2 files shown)


gzm0:
This targets the wrong branch.

sjrd:
Jenkins, retest this please

sjrd:
> This targets the wrong branch.

Oops, nice catch!

gzm0:
Still LGTM

PR #2967, opened 2017-05-26
[no-master] Add an option -P:scalajs:sjsDefinedByDefault.

sjrd:
When enabled, this option switches the default for `js.Any` things from `@js.native` (0.6.x style) to `@ScalaJSDefined` (1.x style). The option only applies to source files in the current run, not to things loaded from the classpath.

This option is intended as a nicer transition between 0.6.x and 1.x. For a time, users will want to cross-compile their codebases. But there is no way to have Scala.js-defined JS classes that compile without deprecation warnings under the two versions. This option fixes that. The codebase can be written in 1.x style, and the option activated when `scalaJSVersion.startsWith("0.6.")`.

The content of this commit should not reach the `master` branch.

### DIFF
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/SJSDefinedByDefaultTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/SJSDefinedByDefaultTest.scala
@@ -1,0 +1,312 @@
+package org.scalajs.core.compiler.test
+
+import org.scalajs.core.compiler.test.util._
+
+import org.junit.Test
+import org.junit.Ignore
+
+// scalastyle:off line.size.limit
+
+class SJSDefinedByDefaultTest extends DirectTest with TestHelpers {
+
+  /* We add the compiler's output path itself to the classpath, for tests
+   * involving separate compilation.
+   */
+  override def classpath: List[String] =
+    super.classpath ++ List(testOutputPath)
+
+  override def extraArgs: List[String] =
+    super.extraArgs :+ "-P:scalajs:sjsDefinedByDefault"
+
+  override def preamble: String =
+    """
+    import scala.scalajs.js
+    import scala.scalajs.js.annotation._
+    """
+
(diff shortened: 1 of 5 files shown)


sjrd:
Does that make sense, @gzm0?

Testing is hard. I added compiler tests, but no run tests in `testSuite`. I tested locally with:
```bash
$ cd test-suite/js/src/test/
$ find . -name '*.scala' | xargs sed -i 's/@ScalaJSDefined/@scala.annotation.strictfp/g'
$ cd -
$ sbt 'set scalacOptions in (testSuite, Test) += "-P:scalajs:sjsDefinedByDefault"' testSuite/test
```

compiler/src/test/scala/org/scalajs/core/compiler/test/SJSDefinedByDefaultTest.scala
```diff
@@ -0,0 +1,246 @@
...
+      |newSource1.scala:6: error: Cannot override concrete val a: scala.scalajs.js.UndefOr[Int] from A in a Scala.js-defined JS trait.
+      |      override val a: js.UndefOr[Int] = js.undefined
+      |                   ^
+      |newSource1.scala:7: error: Cannot override concrete def b: scala.scalajs.js.UndefOr[Int] from A in a Scala.js-defined JS trait.
+      |      override def b: js.UndefOr[Int] = js.undefined
+      |                   ^
+    """
+  }
+
+}
```
gzm0: Can you also add a test that a Scala.js defined class cannot directly extend a native JS trait? Just for extra safety :)

sjrd: Ah yes, that's a good one to test, indeed.

sjrd: Done.

PR #2968, opened 2017-05-26
Fix #2876: Cut dependencies on code that will move to modules

### LINKED ISSUE #2876: Cut dependencies on future modules from upstream
See https://github.com/scala/scala/pull/5830. We'll need to perform similar adjustments. In particular:
* Do not use `sys.error`
* Do not use `scala.compat._`

sjrd:
Ref: #2876

This is "ideologically" a port from https://github.com/scala/scala/pull/5830, except of course it is applied on our codebase.

After these changes, we have:
```
$ git grep 'sys\.'
scalalib/overrides/scala/util/control/NoStackTrace.scala: *  [[scala.sys.SystemProperties]].
scalalib/overrides/scala/util/control/NoStackTrace.scala:  // two-stage init to make checkinit happy, since sys.SystemProperties.noTraceSupression.value calls back into NoStackTrace.noSuppression
```
i.e., only in comments of overrides, and
```
$ git grep 'scala\.compat'
scalalib/overrides-2.10/scala/compat/Platform.scala:package scala.compat
scalalib/overrides-2.10/scala/compat/Platform.scala:   *  For example, the following code `scala.compat.Platform.createArray(classOf[Int], 4)` returns an array of the java primitive type `int`.
scalalib/overrides-2.10/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.createArray(classOf[Int], 4).asInstanceOf[Array[Int]] // returns Array[Int](0, 0, 0, 0)
scalalib/overrides-2.10/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.getClassForName("java.lang.Integer")  // returns the Class[_] for java.lang.Integer
scalalib/overrides-2.11/scala/compat/Platform.scala:   *  For example, the following code `scala.compat.Platform.createArray(classOf[Int], 4)` returns an array of the java primitive type `int`.
scalalib/overrides-2.11/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.createArray(classOf[Int], 4).asInstanceOf[Array[Int]] // returns Array[Int](0, 0, 0, 0)
scalalib/overrides-2.11/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.getClassForName("java.lang.Integer")  // returns the Class[_] for java.lang.Integer
scalalib/overrides-2.12/scala/compat/Platform.scala:   *  For example, the following code `scala.compat.Platform.createArray(classOf[Int], 4)` returns an array of the java primitive type `int`.
scalalib/overrides-2.12/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.createArray(classOf[Int], 4).asInstanceOf[Array[Int]] // returns Array[Int](0, 0, 0, 0)
scalalib/overrides-2.12/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.getClassForName("java.lang.Integer")  // returns the Class[_] for java.lang.Integer
scalalib/overrides-2.13/scala/compat/Platform.scala:   *  For example, the following code `scala.compat.Platform.createArray(classOf[Int], 4)` returns an array of the java primitive type `int`.
scalalib/overrides-2.13/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.createArray(classOf[Int], 4).asInstanceOf[Array[Int]] // returns Array[Int](0, 0, 0, 0)
scalalib/overrides-2.13/scala/compat/Platform.scala:   *  val a = scala.compat.Platform.getClassForName("java.lang.Integer")  // returns the Class[_] for java.lang.Integer
```
i.e., only in the overrides of `scala.compat.Platform` itself.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
@@ -16,6 +16,7 @@ import scala.tools.nsc._
  *  @author Sébastien Doeraene
  */
 trait Compat210Component {
+  import Compat210Component.{infiniteLoop, noImplClasses}
 
   val global: Global
 
@@ -25,7 +26,7 @@ trait Compat210Component {
 
   implicit final class SymbolCompat(self: Symbol) {
     def unexpandedName: Name = self.originalName
-    def originalName: Name = sys.error("infinite loop in Compat")
+    def originalName: Name = infiniteLoop()
 
     def isPrivateThis: Boolean = self.hasAllFlags(PRIVATE | LOCAL)
     def isLocalToBlock: Boolean = self.isLocal
@@ -49,18 +50,20 @@ trait Compat210Component {
       self: Compat210Component.this.global.type) {
 
     def enteringPhase[T](ph: Phase)(op: => T): T = self.beforePhase(ph)(op)
-    def beforePhase[T](ph: Phase)(op: => T): T = sys.error("infinite loop in Compat")
+    def beforePhase[T](ph: Phase)(op: => T): T = infiniteLoop()
 
     def exitingPhase[T](ph: Phase)(op: => T): T = self.afterPhase(ph)(op)
-    def afterPhase[T](ph: Phase)(op: => T): T = sys.error("infinite loop in Compat")
+    def afterPhase[T](ph: Phase)(op: => T): T = infiniteLoop()
 
     def delambdafy: DelambdafyCompat.type = DelambdafyCompat
   }
 
   object DelambdafyCompat {
     object FreeVarTraverser {
-      def freeVarsOf(function: Function): mutable.LinkedHashSet[Symbol] =
-        sys.error("FreeVarTraverser should not be called on 2.10")
+      def freeVarsOf(function: Function): mutable.LinkedHashSet[Symbol] = {
+        throw new AssertionError(
+            "FreeVarTraverser should not be called on 2.10")
+      }
     }
   }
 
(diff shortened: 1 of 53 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
```diff
@@ -250,6 +252,12 @@ object Compat210Component {
       }
     }
   }
+
+  private def infiniteLoop(): Nothing =
+    throw new NotImplementedError("Infinite loop in Compat")
+
+  private def noImplClasses(): Nothing =
+    throw new NotImplementedError("No impl classes in this version")
```
gzm0: I think those two should be assertion errors.

javalib/src/main/scala/java/lang/MathJDK8Bridge.scala
```diff
@@ -21,54 +21,52 @@
 package java.lang
 
 private[java] object MathJDK8Bridge {
-  def addExact(a: scala.Int, b: scala.Int): scala.Int = sys.error("stub")
+  def addExact(a: scala.Int, b: scala.Int): scala.Int = ???
```
gzm0: So the idea is that you replace `sys.error("stub")` with `???` if it is a bridge, and with `new jl.Error("stub")` if it is a primitive? I think using `???` is not advisable, since it throws a `NotImplementedError` which is not really appropriate here (the problem is a messed up classpath, not the implementation).

I suggest:
- Replacing these with `jl.Error("stub")`.
- or replacing `sys.error("stub")` with `jl.Error("primitive")` / `jl.Error("bridge")` as appropriate (might be overkill?).

compiler/src/main/scala/org/scalajs/core/compiler/Compat210Component.scala
sjrd: I wasn't sure for those. I'm glad you have a stronger opinion :P

javalib/src/main/scala/java/lang/MathJDK8Bridge.scala
sjrd: That was the idea, yes. I'm fine with `jl.Error("stub")` too.

sjrd:
Updated.

PR #2969, opened 2017-05-27
DO NOT MERGE Complete review of Linker.Config and friends

sjrd:
I started with the desire to simplify those things so that they could legitimately be exposed as sbt settings (instead of `scalaJSOptimizerOptions` which is a bag of "stuff"), and one thing led me to another, until there was no `Linker.Config` at all anymore ^^

Not sure about all of that, especially the last commit. Discussion welcome.

### DIFF
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -195,12 +195,12 @@ object Scalajsld {
       val backendConfig = LinkerBackend.Config()
         .withSourceMap(options.sourceMap)
         .withRelativizeSourceMapBase(options.relativizeSourceMap)
+        .withClosureCompiler(options.fullOpt)
         .withPrettyPrint(options.prettyPrint)
 
       val config = Linker.Config()
         .withOptimizer(!options.noOpt)
         .withParallel(true)
-        .withClosureCompiler(options.fullOpt)
         .withFrontendConfig(frontendConfig)
         .withBackendConfig(backendConfig)
 
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -193,11 +193,11 @@ object Scalajsld {
         .withCheckIR(options.checkIR)
 
       val backendConfig = LinkerBackend.Config()
+        .withSourceMap(options.sourceMap)
         .withRelativizeSourceMapBase(options.relativizeSourceMap)
         .withPrettyPrint(options.prettyPrint)
 
       val config = Linker.Config()
-        .withSourceMap(options.sourceMap)
         .withOptimizer(!options.noOpt)
         .withParallel(true)
         .withClosureCompiler(options.fullOpt)
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -200,12 +200,8 @@ object Scalajsld {
         .withClosureCompiler(options.fullOpt)
         .withPrettyPrint(options.prettyPrint)
 
-      val config = Linker.Config()
-        .withFrontendConfig(frontendConfig)
-        .withBackendConfig(backendConfig)
-
       val linker = Linker(semantics, options.outputMode, options.moduleKind,
-          config)
+          frontendConfig, backendConfig)
 
       val logger = new ScalaConsoleLogger(options.logLevel)
       val outFile = WritableFileVirtualJSFile(options.output)
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -191,6 +191,8 @@ object Scalajsld {
       val frontendConfig = LinkerFrontend.Config()
         .withBypassLinkingErrorsInternal(options.bypassLinkingErrors)
         .withCheckIR(options.checkIR)
+        .withOptimizer(!options.noOpt)
+        .withParallel(true)
 
       val backendConfig = LinkerBackend.Config()
         .withSourceMap(options.sourceMap)
@@ -199,8 +201,6 @@ object Scalajsld {
         .withPrettyPrint(options.prettyPrint)
 
       val config = Linker.Config()
-        .withOptimizer(!options.noOpt)
-        .withParallel(true)
         .withFrontendConfig(frontendConfig)
         .withBackendConfig(backendConfig)
 
--- a/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
+++ b/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
@@ -70,7 +70,7 @@ class MainGenericRunner {
       Seq(ModuleInitializer.mainMethod("PartestLauncher", "main"))
 
     val linkerConfig = Linker.Config()
-      .withSourceMap(false)
+      .withBackendConfig(_.withSourceMap(false))
       .withClosureCompiler(optMode == FullOpt)
 
     val linker = Linker(semantics, OutputMode.ECMAScript51Isolated,
--- a/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
+++ b/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
@@ -71,7 +71,7 @@ class MainGenericRunner {
 
     val linkerConfig = Linker.Config()
       .withBackendConfig(_.withSourceMap(false))
-      .withClosureCompiler(optMode == FullOpt)
+      .withBackendConfig(_.withClosureCompiler(optMode == FullOpt))
 
     val linker = Linker(semantics, OutputMode.ECMAScript51Isolated,
         ModuleKind.NoModule, linkerConfig)
--- a/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
+++ b/partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
@@ -8,7 +8,8 @@ import org.scalajs.core.tools.io._
 import org.scalajs.core.tools.jsdep.ResolvedJSDependency
 import org.scalajs.core.tools.io.IRFileCache.IRContainer
 import org.scalajs.core.tools.linker.{Linker, ModuleInitializer}
-import org.scalajs.core.tools.linker.backend.{OutputMode, ModuleKind}
+import org.scalajs.core.tools.linker.frontend.LinkerFrontend
+import org.scalajs.core.tools.linker.backend._
 
 import org.scalajs.core.ir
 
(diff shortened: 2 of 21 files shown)


gzm0:
Do you also want to consolidate the method hashes in the IR to not have a tree and a pos part anymore?

sjrd:
Yes but that can only be done in 1.x.

gzm0:
I'm not so sure about the last part of this one (removing source maps from the frontend is fine by me).

When I have written this, I did the mistake of trying to consolidate configs between base classes. Now when I look at this, it is much clearer to me that configs should only belong to concrete implementations (and otherwise be completely unrelated).

So the pairs should IMO be:

- `Linker.Config` <-> `Linker.apply`
    Note that the config is independent from `class Linker`
- `LinkerFrontend.Config` <-> `class LinkerFrontend`
    With `sourceMaps` gone this is cleaner. Ideally I think we should:
    - Hide `optimizerFactory` inside the config.
    - Only expose a factory. This will ease transition to plugins will we ever do it.
- `BasicLinkerBackend.Config` <-> `class BasicLinkerBackend` (factory optional here)
- `ClosureLinkerBackend.Config` <-> `class ClosureLinkerBackend` (factory optional here)

The current design of `LinkerBackend` is terrible. It is IMHO a blatant abuse of implementation inheritance. `LinkerBackend` should just be a trait and only be:

```scala
trait LinkerBackend {
  val semantics: Semantics
  val esLevel: ESLevel
  val symbolRequirements: SymbolRequirement
}
```

Further, I think `Linker.Config` should actually hide its internal configs and copy relevant members in `Linker.apply`. The reason is twofold:

- With everything on `Linker.Config`, this is exactly the thing we want to expose in sbt (or the build tool of your choice).
- It will allow us to change the internal APIs independently: This is exactly one of the examples of the different levels of public APIs I have been talking about. We **must** guarantee that `Linker.Config`, `Linker.apply` and the public methods of `GenLinker` / `Linker` remain backwards compatible. However, I would take the freedom for us to (in minor versions?) change the internal structure of the linker (and hence potentially moving / breaking / etc. `BasicBackend.Config` or whatever). Insulating these through `Linker.Config` will allow us to make this change transparent to the users of the full public API.

Does that make sense?

sjrd:
My analysis partially meets your own. I agree that configs should be related to concrete implementations. The existence of `LinkerBackend.config` is nonsensical.

Further than that, I think our design depends on one crucial question:

* Do we want to commit to the front-end/back-end split of the linker as a stable, long-term API?

## If yes

If we do, then `Linker` as a `final class` as it is now is basically frozen for a long time, and we should have a stable API for the traits `LinkerFrontend` and `LinkerBackend`. We can then have `BasicLinkerFrontend` (or maybe `StandardLinkerFrontend`) and `-Backend`. Those would have an `apply` method taking their respective `Config`, and building an actual instance.

In this scenario, I believe that `Linker.apply` taking a `StandardLinkerFrontend.Config` and `StandardLinkerBackend.Config` is a nice convenient factory to do "the standard thing". Furthermore, I do believe `Linker.Config` is not a useful abstraction. And it would be reasonable to expose the standard front-end and back-end configs separately in build tools.

This also implies that `LinkingUnit` must have a pretty stable API. For linker plugins, we will need a stable `LinkingUnit` anyway, though.

## If no

If the front-end/back-end separation is not something we want to commit to in the long term, then we are in a significantly different scenario. In that case, I believe that `Linker` itself should be a `trait` without only one method `def link`. (and probably still `val semantics`; `val esLevel` is borderline already) `GenLinker` appears to me as unnecessary anyway.

In that scenario, we should have a `StandardLinker` with a `Config` of its own, not exposing its internals, as you suggest.

gzm0:
> Do we want to commit to the front-end/back-end split of the linker as a stable, long-term API?

I would say no. Mainly because there is zero benefit for the API user (the one that just wants to link his code) to do so. But also because we know that there is a chance that well add linker plugins in the future, so it's better if we keep the ability to restructure the internals.

I also played with the thought that we should have `Linker` and `StandardLinker` (with all the options), so I think we should do that.

I'm not so sure whether `Linker` should have anything else than the `link` method (in its fully public API). What's the use case for asking a linker for its semantics (or ES level)?

sjrd:
> I would say no. Mainly because there is zero benefit for the API user (the one that just wants to link his code) to do so.

I think that's right.

> But also because we know that there is a chance that well add linker plugins in the future

Given that the optimizer is already in the front-end, IMO all linker plugins would be part of the front-end. I have no idea how I would specify a plugin that works in our back-end.

> What's the use case for asking a linker for its semantics (or ES level)?

AFAICT we only use it as a safety guard in `ClearableLinker`: https://github.com/scala-js/scala-js/blob/de825604b9b2a405913932079e2cdcd4ed95ea19/tools/shared/src/main/scala/org/scalajs/core/tools/linker/ClearableLinker.scala#L82-L90 But then again this might be "for its own sake", in that it doesn't want to expose an interface where its `semantics` and `esLevel` change.

gzm0:
> Given that the optimizer is already in the front-end, IMO all linker plugins would be part of the front-end. I have no idea how I would specify a plugin that works in our back-end.

Yes, I agree. I think it is very likely that we keep the frontend and the backend distinction (with plugins just being additional arguments to the frontend). But it is just something I do not think is worth giving any guarantees about at this point.

> But then again this might be "for its own sake", in that it doesn't want to expose an interface where its semantics and esLevel change.

Yeah, I think it is safe to simplify this and remove the `semantics` and `esLevel` methods.

sjrd:
OK so here is an attempt as describing what we want, in the end.

## Public API, in `org.scalajs.core.tools.linker`

```scala
abstract class Linker {
}

final class ClearableLinker(newLinker: () => Linker) {
  def link(...): Unit
  def clear(): Unit
}

object StandardLinker {

  final class Config {
    val semantics: Semantics
    val moduleKind: ModuleKind
    
    val sourceMap: Boolean
    val parallel: Boolean
    ...

    ...
  }

  object Config {
  }
}

final class Semantics { ... }
object Semantics { ... }

abstract sealed class ModuleKind { ... }
object ModuleKind { ... }
```
Some things to note:
* There is no `class StandardLinker`. That's a private-ish API.
* I have relegated `OutputMode` to an optional config and it made it private. The rationale is that it is tight to our current standard linker backend and emitter. But in practice we know it's not something we want to support for a very long time (we destroyed 3 `OutputMode`s already in our history!). We can configure it with some private-ish APIs (see below)
* The only two mandatory config fields are therefore `Semantics` and `ModuleKind`. These are the ones that semantically define how the code behaves a) internally (`Semantics`) and b) externally (`ModuleKind`).
* `GenLinker` is gone. `ClearableLinker` is unrelated to `Linker`.
* `Linker` is an `abstract class`, not a trait, so that we can hope to add additional methods in a binary compatible way as long as they have a default implementation. This would be necessary for an API that can generate *multiple* JS files in the future, for example. (alternatively, if we drop 2.10 and 2.11 support for the tools, it can stay a `trait` as such things would be default methods)

## Private-ish API in `org.scalajs.core.tools.linker.standard`

This is the "standard" linker. It is decomposed in a front-end and a back-end. The back-end has an `OutputMode`.

```scala
object OutputMode { ... }

object StandardLinker {
}

package object standard {
  implicit class StandardConfigOps(val self: StandardLinker.Config) extends AnyVal {
    val outputMode: OutputMode

  }
}

}

object LinkerFrontend {
  final class Config { ... }
}

}

object LinkerBackend {
  final class Config { ... }
}

```
Notes:
* The `LinkingUnit` abstraction and standard and supported
* There are some additional operations on `StandardLinker.Config`, notably for `OutputMode`

## Private API

## Rationale

The idea is that we have 2 levels of support + 1 level of non-support:

* `linker._`: stable, long-lived support. Basically will always work. This is what you use if you "just want to link your code".
* `linker.standard._`: stable-ish, short-lived support. Can be broken in a minor version (but not patch version). This is what you use if you want to customize your linking, or if you want to write linker plugins.
* the private API: not exposed at all, can't be touched.

## Path to get there

In 0.6.x, mostly leave things as they are in the current state of this PR + add `StandardLinker` from the future API. `Linker.Config` and `Linker.apply` are deprecated in favor of `StandardLinker`. Also add type aliases to `Semantics` and `ModuleKind` so that we can hope to write source-compatible user code between 0.6.x and 1.x, as long as you "just want to link".

In 1.x, just nuke everything and put the new API in place.

WDYT?

gzm0:
A couple of things:

- I think the public `Linker` should not be inherited from. I propose a package private constructor and another Linker type in the `standard` package for inheritance. This way we can even add methods that do **not** have default implementations.
- Why should `ClearableLinker` not extend `Linker`?
- I'm assuming `LinkingException` is deliberately not sealed so plugins can throw subclasses.
- Hiding `OutputMode` is a good idea, but how can I chose ES5 vs ES6? Not at all? Maybe `ESLevel` should be exposed.
- The `sealed` part of `ModuleKind` is not part of the public API (users must expect that we'll add subclasses). How do we express that?
- I'm assuming that `LinkerBackend.Config` is a "typo". IIUC we did want to tie the backend configs to their relevant implementations, right?

Migration plan sounds reasonable.

sjrd:
> I think the public Linker should not be inherited from. I propose a package private constructor and another Linker type in the standard package for inheritance. This way we can even add methods that do not have default implementations.

I didn't think about that. I guess that makes sense, indeed.

> Why should ClearableLinker not extend Linker?

Probably no good reason anymore.

> I'm assuming LinkingException is deliberately not sealed so plugins can throw subclasses.

Yes, I see no reason to finalize/seal it.

> Hiding OutputMode is a good idea, but how can I chose ES5 vs ES6? Not at all? Maybe ESLevel should be exposed.

`ESLevel` is not a powerful enough abstraction. We might decide to have two output modes for the same ES level (e.g., one that leverages all the ES6 features that are useful for performance and code size, but avoids those that are known to perform badly in current engines).

I don't have a foolproof solution. I think people who want the non-default output mode will have to tap a little bit in the `standard` package to select it. I don't think it's a big deal in this case.

> The sealed part of ModuleKind is not part of the public API (users must expect that we'll add subclasses). How do we express that?

LOL. I was wondering. I don't think Scala as a language has a solution for that. We could make it an `abstract class` with a private constructor.

> I'm assuming that LinkerBackend.Config is a "typo". IIUC we did want to tie the backend configs to their relevant implementations, right?

I'm not sure. In general, yes. But I more or less consider the `BasicLinkerBackend` and `ClosureLinkerBackend` as being "the same" concrete implementation. They're just one config flag away from each other.

But we can also have `BasicLinkerBackend.Config` and `ClosureLinkerBackend.Config` if you prefer, and have the code that converts `StandardLinker.Config` to those dispatch on the right implementation as well.

gzm0:
>  I think people who want the non-default output mode will have to tap a little bit in the standard package to select it.

Fair enough.

gzm0:
> But I more or less consider the BasicLinkerBackend and ClosureLinkerBackend as being "the same" concrete implementation.

Fair enough. Now that they are behind `standard._`, I'm much more comfortable with sharing them for now.

> We could make it an abstract class with a private constructor.

We probably need to look at something like this. The main caveat is that we still want to be able to get exhaustiveness checking in `standard._`. So the only thing I can think of is this abomination:

```scala
// tools.linker
abstract class ModuleKind private[linker] () {
  private[linker] def exhaustive(): standard.ModuleKind
}

object ModuleKind {
  val NoModule: ModuleKind = standard.ModuleKind.NoModule
  // etc.
}

// tools.linker.standard
sealed abstract class ModuleKind extends linker.ModuleKind {
  def exhaustive(): standard.ModuleKind = this
}

object ModuleKind {
  object NoModule extends ModuleKind
  // etc.
}
```

So if we need to do a match, one can do:

```scala
mk.exhaustive match {
  case NoModule =>
  // etc. but no default.
}
```

PR #2974, opened 2017-05-29
Remove the JSDependencies plugin from our own build.

sjrd:
This is of course required before we can actually move that plugin in a separate repo, and remove from the core repo.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -639,6 +639,9 @@ object Build {
         Seq(outFile)
       }.taskValue,
 
+      // Give more memory to Node.js, and deactivate source maps
+      jsEnv := new NodeJSEnv(args = Seq("--max_old_space_size=3072")).withSourceMap(false),
+
       jsDependencies += ProvidedJS / "js-test-definitions.js" % "test",
       jsDependencies +=
         "org.webjars" % "jszip" % "2.4.0" % "test" / "jszip.min.js" commonJSName "JSZip",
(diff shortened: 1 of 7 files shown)


project/Build.scala
```diff
@@ -638,6 +638,12 @@ object Build {
         testDefinitionsFile +: (jsExecutionFiles in Test).value
       },
 
+      jsExecutionFiles in Test := {
+        val resourceDir = (resourceDirectory in (testSuite, Test)).value
+        val f = FileVirtualJSFile(resourceDir / "ScalaJSDefinedTestNatives.js")
+        f +: (jsExecutionFiles in Test).value
+      },
```
gzm0: Have you considered factoring this into a method (that takes a key)?

sjrd: I, er, haven't. You mean something like this, I guess:
```scala
def testJSExecutionFilesSetting: Setting[_] = {
  jsExecutionFiles in Test := {
    val resourceDir = (resourceDirectory in (testSuite, Test)).value
    val f = FileVirtualJSFile(resourceDir / "ScalaJSDefinedTestNatives.js")
    f +: (jsExecutionFiles in Test).value
  }
}
```
and then use it both in `testSuite.settings(...)` and `toolsJS.settings(...)`?

gzm0: Yeah, something the like :)

sjrd: OK. Will wait for #2975 to go through before rebasing.

sjrd: Done.

PR #2977, opened 2017-05-30
Add the `js.special` package with `delete`, `debugger` and `globalThis`.

sjrd:
Spawned from https://github.com/scala-js/scala-js/pull/2934#discussion_r118822237, backported to 0.6.x so that it is possible to write codebases cross-compiling between 0.6.x and 1.x.

The `js.special._` name can still be debated ;)


library/src/main/scala/scala/scalajs/js/special/package.scala
```diff
@@ -0,0 +1,83 @@
...
+package scala.scalajs.js
+
+/** Contains special primitives of interoperability with JavaScript which are
+ *  of limited importance or rare usefulness.
+ *
+ *  In theory, all of the members of this package could equally well be part of
+ *  the `scala.scalajs.js` package. They are sligthly "hidden" in this
+ *  `special` package so that they are not used on a daily basis, but only when
+ *  absolutely necessary.
+ */
```
gzm0: I'm wondering if we should add a short comment about typing?

Something like:

> Everything in this package is a "I-know-what-I-am-doing" API. Notably, no attempt is made to guide the user by types that are not hard requirements.

sjrd: Good point, done. and fixed the bincompat check.

PR #2981, opened 2017-05-31
Fix #2786: Prevent Closure from renaming `$classData`.

### LINKED ISSUE #2786: Code which runs with fastOptJS but crashes with fullOptJS
Minimal reproducing example project:
https://github.com/woost/wust2/tree/fullopt-bug

```bash
git clone https://github.com/woost/wust2 -b fullopt-bug --depth=1
cd wust2
./test.sh
```

sjrd:
This property is used to determine whether an arbitrary JS object is a Scala object. If GCC renames it to something too common (e.g., a single letter like `b`), we can erroneously identify a JS object as being a Scala object, and then things go wrong.

This slightly increases the size of `-opt.js` files, but it seems that gziped sizes are virtually not affected.


sjrd:
Tested locally with the reproduction in #2786.

PR #2982, opened 2017-06-01
Remove the scalaJSConsole sbt setting.

sjrd:
sbt does not provide JVM projects with any similar setting, allowing to redirect the output of the processes. Providing one in Scala.js project is therefore useless complexity.

Consequently, we also remove the `JSConsole` argument to `new ScalaJSFramework`, as it is not used anymore. Moreover, the sbt testing interface for the JVM does not offer a similar mechanism allowing to redirect the standard output. Therefore, providing it to the Scala.js testing framework adapter is useless complexity.


sjrd:

PR #2985, opened 2017-06-03
CI: Only run `npm install` if not up-to-date.

sjrd:
Since `npm install` seems to have a high failure rate on the CI servers, we explicitly avoid running it if there is nothing to do. We test that by comparing the sha1 hash of `package.json` to a cache file.


sjrd:
Meh, that's stupid. Jenkins wipes the working directory every time, so this doesn't help :-( I need to find a better way to do this.

PR #2990, opened 2017-06-05
Fix #2382 for good, on all Scala versions.

### LINKED ISSUE #2382: Name clash for $outer pointers of two different nesting levels

sjrd:
We simply remove the guard that prevented the fix from being applied in 2.10 and 2.11. It was necessary in 0.6.x not to break backward binary compatibility, but we can now fix it for good.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala
@@ -80,12 +80,6 @@ trait JSEncoding extends SubComponent { self: GenJSCode =>
     js.Ident(localSymbolName(sym), Some(sym.unexpandedName.decoded))
   }
 
-  /** See comment in `encodeFieldSym()`. */
-  private lazy val shouldMangleOuterPointerName = {
-    val v = scala.util.Properties.versionNumberString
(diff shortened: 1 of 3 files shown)


sjrd:
Hum ... the test fails on 2.10. I'll have to investigate that later.

PR #2991, opened 2017-06-05
Fix #1796: Cleanup the library

### LINKED ISSUE #1796: Privatize and seal everything we can in the library
Was: js.LowPrioAnyImplicits should be sealed

Its non-sealed status means we can't add methods in it without breaking binary compatibility :-(
Of course sealing it is itself a source incompatible change, so this is for the next major release.

sjrd:
Some things that were *not* done, and that we could consider in the future:

* Privatize `StackTrace.JSStackTraceElem`, because it is referenced by `EnvironmentInfo`.
* Move `annotation.JavaDefaultMethod` to `internal._`. Technically it is not internal, as it is meant to be used in user source code. But its legit uses are really rare, and must typically done with extreme care, so we may not want to keep that annotation in plain sight.


library/src/main/scala/scala/scalajs/js/JSNumberOps.scala
```diff
@@ -93,13 +93,13 @@ object JSNumberOps {
...
       "operations. This is almost certainly not what you want. "+
-      "Use `.toDouble` explicitly if you need it.", "0.6.0")
+      "Use `.toDouble` explicitly if you need it.", "forever")
   implicit def enableJSNumberOps(x: Long): JSNumberOps =
     x.toDouble.asInstanceOf[JSNumberOps]
 
   @deprecated("A Long is converted to Double to perform JavaScript "+
       "operations. This is almost certainly not what you want. "+
-      "Use `.toDouble` explicitly if you need it.", "0.6.0")
+      "Use `.toDouble` explicitly if you need it.", "forever")
```
gzm0: Maybe comment in the commit message on why we are keeping these (just like in the other commit)? Or even move these changes to the other commit.

library/src/main/scala/scala/scalajs/js/WrappedDictionary.scala
```diff
@@ -57,7 +57,7 @@ class WrappedDictionary[A](val dict: Dictionary[A])
   }
 
   override def update(key: String, value: A): Unit =
```
gzm0: Shouldn't this be `@inline`?

library/src/main/scala/scala/scalajs/js/PropertyDescriptor.scala
```diff
@@ -10,12 +10,21 @@
 
 package scala.scalajs.js
 
-@native
+import scala.scalajs.js
```
gzm0: Sure about this? I think if we do this, we should do it consistently.

library/src/main/scala/scala/scalajs/js/WrappedDictionary.scala
sjrd: No, the class itself is `@inline`. If the instance of `WrappedDictionary` was not virtualized away, there's no point in inlining `update`. But if it is, then `update` will be inlined automatically.

library/src/main/scala/scala/scalajs/js/PropertyDescriptor.scala
sjrd: We've slowly been adopting this style in new code, yes. At some point we'll need a pass over the entire package to do it consistently, indeed.

sjrd: I can add an additional commit with that change everywhere. It kind of fits the theme of the PR.

library/src/main/scala/scala/scalajs/js/WrappedDictionary.scala
gzm0: Right :)

gzm0: (I think I ask this every time you touch this code, don't I :P)

library/src/main/scala/scala/scalajs/js/PropertyDescriptor.scala
gzm0: I think that would be nice.

library/src/main/scala/scala/scalajs/js/WrappedDictionary.scala
sjrd: If you do, I forget from one time to the next ^^

sjrd:
Updated. The last two commits are new. And I moved the `"0.6.x"` -> `"forever"` change to the commit that deprecates other `Long` to JS number conversions.

gzm0:
Ah, this needs a rebase after #2800. I guess it is fine by you that I merged that one first ;)

sjrd:
> I guess it is fine by you that I merged that one first ;)

Yes, I think that was the better order :)

sjrd:
Rebased.

PR #2992, opened 2017-06-05
DO NOT MERGE Recognize any "standard" JVM-style main object in the sbt plugin.

sjrd:
Instead of recognizing objects extending `js.JSApp` as main objects in the sbt plugin, we now recognize any object with a JVM-style `main` method (with an `Array[String]` parameter). We use a new kind of module initializer, namely `ModuleInitializer.mainMethodWithArgs`, to call them.

This has the nice benefit that we can finally write a cross-platform main object.

For backward source compatibility, we add in `js.JSApp` a forwarder
```scala
def main(args: Array[String]): Unit = main()
```
which allows existing `js.JSApp` to still be recognized and run. `js.JSApp` is nevertheless deprecated, as there is no reason to use it anymore.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/JSApp.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSApp.scala
@@ -2,13 +2,28 @@ package scala.scalajs.js
 
 /** Base class for top-level, entry point main objects.
  *
- *  [[JSApp]] is typically used to mark the entry point of a Scala.js
- *  application. As such, the sbt plugin also recognizes top-level objects
- *  extending [[JSApp]]. It allows to run their [[main]] method with `sbt run`.
+ *  In Scala.js 0.6.x, an top-level object had to extend `js.JSApp` to be
+ *  recognized by the sbt plugin as a "main" object, to be executed with `run`.
+ *  Starting with Scala.js 1.0.0, any object with a standard main method of the
+ *  form
+ *  {{{
+ *  def main(args: Array[String]): Unit = ???
+ *  }}}
+ *  will be recognized by the sbt plugin, just like for a JVM project.
  *
- *  To execute the [[main]] method immediately when your Scala.js file is
- *  loaded, use the `scalaJSUseMainModuleInitializer` setting in the sbt plugin.
+ *  [[JSApp]] is therefore deprecated, and should not be used anymore.
  */
+@deprecated(
+    "Extending js.JSApp is not necessary anymore for an object to be " +
+    "recognized by the Scala.js sbt plugin. " +
+    "Use a normal object with a `def main(args: Array[String]): Unit` " +
+    "instead, which will also be cross-platform. " +
+    "Note that js.JSApp objects and their main() method are not " +
+    "automatically exported to JavaScript either; explicitly export your " +
+    "object if you relied on this behavior.",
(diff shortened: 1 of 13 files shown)


sjrd:

library/src/main/scala/scala/scalajs/js/JSApp.scala
```diff
@@ -2,13 +2,28 @@ package scala.scalajs.js
...
  */
+@deprecated(
+    "Extending js.JSApp is not necessary anymore for an object to be " +
+    "recognized by the Scala.js sbt plugin. " +
+    "Use a normal object with a `def main(args: Array[String]): Unit` " +
+    "instead, which will also be cross-platform. " +
+    "Note that js.JSApp objects and their main() method are not " +
+    "automatically exported to JavaScript either; explicitly export your " +
+    "object if you relied on this behavior.",
+    "1.0.0")
```
gzm0: What's the rollover plan for this?

sjrd: For the "it's not exported any more", there's no plan. People need to stop relying on that behavior for 1.x, and hence also in 0.6.x if they want to cross-compile.

For the case of having the same code work in both 0.6.x and 1.x without deprecation, I do not really have any plan. It's not as bad as the `@ScalaJSDefined` problem, though, which could creep in zillions of places in a codebase. A main method is typically alone in a codebase, so at worst you get 1 deprecation warning for this.

One way to achieve it if truly necessary in a codebase, is to swap out one source file depending on the Scala.js version, from the sbt build. It can be as small as a file that defines a type alias to `js.JSApp` (for 0.6.x) or to an empty trait (for 1.x).

sjrd: Hum, so you were thinking that we would remove `@ScalaJSDefined` altogether before 1.0.0 final? I thought it was pretty much a given that it was going to stay there in limbo in 1.x. `js.JSApp` is in the same situation.

Now, it's an interesting thought to deprecate them in the Milestones and remove them in the RCs. But that means users cannot jump straight from 0.6.latest to 1.0.0 final. They need to take a corner through the latest 1.0.0-Mx.

Unless we're bad-ass and we deprecate `@ScalaJSDefined` in 0.6.18, with a message saying that users should activate `-P:scalajs:sjsDefinedByDefault` and remove the annotations.

If we're willing to do that for `@ScalaJSDefined`, then we should indeed find an equally good solution for `js.JSApp`. I'd be very happy not to it linger in the library forever!

gzm0: I think it is OK to create a "migration release" like we have done before that essentially triggers warnings all over the place but still fully compiles an 0.6.x codebase. So I was assuming we would to that for `@ScalaJSDefined`, yes :)

sjrd: [Follow up discussion on Gitter](https://gitter.im/scala-js/scala-js/core?at=5937101302c480e6721b3f02).

It seems we have a plan over there. That means we shouldn't merge this PR just yet in master. Some of it needs to go in 0.6.x first.

sjrd:
Superseded by #3019.

PR #2993, opened 2017-06-06
Fix #2987: Migrate to sbt-crossproject.

### LINKED ISSUE #2987: Migrate to sbt-crossproject
https://github.com/scala-native/sbt-crossproject

Work-in-Progress at https://github.com/scala-js/scala-js/compare/master...sjrd:sbt-crossproject
Requires sbt-crossproject 0.2.0 to be published.

sjrd:
https://github.com/scala-native/sbt-crossproject

sbt-crossproject was designed to be a backward source compatible replacement for Scala.js' own `crossProject`, with a generalization to arbitrary platforms, notably Scala Native.

This commit drops our own implementation of `CrossProject` and its associated features (e.g., the `%%%` macro), and delegates everything to sbt-crossproject instead.


sjrd:
Good to observe: there is not a single change in `sbt-plugin-test/`, which demonstrates the backward source compatibility.

There is a deprecation warning, though, as `crossProject` should be replaced by `crossProject(JVMPlatform, JSPlatform)`, but sbt-crossproject was designed so that a bare `crossProject` defaults to JVM/JS, with a deprecation warning.

sjrd:
> This is OK for me under the condition that sbtcrossproject enters its RC cycle before SJS and that sbtcrossproject 1.0.0 is released before SJS.

Of course, that's part of the plan! Otherwise we would be in a very precarious situation.

gzm0:
I was expecting a response of that sort.

PR #2996, opened 2017-06-07
Recognize any "standard" JVM-style main object in the sbt plugin.

sjrd:
In addition to recognizing objects extending `js.JSApp`, the sbt plugin now recognizes "standard" main methods as well, i.e., objects with a method of the form
```scala
def main(args: Array[String]): Unit = ...
```
If such an object is used as main class, its main method is called using the new `ModuleInitializer.mainMethodWithArgs`.

This has the nice benefit that we can finally write a cross-platform main object.

We "softly" deprecate `js.JSApp`: its documentation says not to use anymore, but it does not cause a deprecation warning yet.

For backward compatibility, objects extending `js.JSApp` are still recognized, and their `main(): Unit` method will be called.

There are some subtle scenarios regarding backward compatibility:

* If an object *both* extends `js.JSApp` and has a standard main method, the `main()` method of `js.JSApp` is preferred, for backward compatibility.
* If an object has a `main(): Unit` method but does not extend `js.JSApp`, it will not be *discovered*. In that case, we assume the old style and call `main()`. This can happen if the user explicitly sets the `mainClass` setting.
* If an object has *both* a `main(): Unit` method and a `main(args: Array[String]): Unit` method, it *will* now be discovered, and the new style will be assumed. This can potentially break compatibility, as now the standard main will be called. This only happens in case the user has explicitly set `mainClass`. Moreover, it is only problematic if `main()` does not behave the same as `main(Array())`, in which case the codebase is arguably in a bad shape to begin with.
* The very fact of recognizing *more* objects as potential main objects means that the sbt build can break because it cannot automatically choose between several discovered objects. This will however fail to link with a nice error message, so it is not too bad.

Here are some examples.

* Old-style only, `main()` is selected

      object OldStyleOnly extends js.JSApp {
        def main(): Unit = ...
      }

* New-style only, `main(Array[String])` is selected

      object NewStyleOnly {
        def main(args: Array[String]): Unit = ...
      }

* Old and new style combined, `main()` is selected

      object OldAndNewStyle extends js.JSApp {
        def main(): Unit = ...
        def main(args: Array[String]): Unit = ...
      }

* `def main(): Unit` without `js.JSApp` only, `main()` is selected

      object OldStyleNonJSApp {
        def main(): Unit = ...
      }

* Both styles of methods without `js.JSApp`, `main(Array[String])` is selected (potentially silently breaking)

      object OldAndNewStyleNonJSApp {
        def main(): Unit = ...
        def main(args: Array[String]): Unit = ...
      }

* Both styles exist in two different objects, sbt reports an ambiguity and demands that `mainClass` be set explicitly

      object OldStyle extends js.JSApp {
        def main(): Unit = ...
      }
      object NewStyle {
        def main(args: Array[String]): Unit = ...
      }

### DIFF
--- a/TESTING
+++ b/TESTING
@@ -11,16 +11,16 @@ For each major Scala version on a *NIX distro and a Windows distro:
 5. Create a temporary directory and do:
 
         mkdir bin
-        echo 'import scala.scalajs.js.JSApp
-          object Foo extends JSApp {
-
-            def main() = {
+        echo '
+          object Foo {
+            def main(): Unit = {
               println(s"asdf ${1 + 1}")
               new A
             }
 
             class A
-          }' > foo.scala
+          }
+        ' > foo.scala
         scalajsc -d bin foo.scala
 
         scalajsp bin/Foo$.sjsir
@@ -28,10 +28,9 @@ For each major Scala version on a *NIX distro and a Windows distro:
         scalajsp bin/Foo\$A.sjsir
         # Verify output
 
-        scalajsld -o test.js bin
+        scalajsld -o test.js -mm Foo.main bin
         # Verify output
 
-        echo "Foo().main()" >> test.js
         node test.js # Or your favorite thing to run JS
 
         # Expect "asdf 2"
--- a/library/src/main/scala/scala/scalajs/js/JSApp.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSApp.scala
@@ -2,7 +2,12 @@ package scala.scalajs.js
 
 import annotation.{JSExport, JSExportDescendentObjects}
 
-/** Base class for top-level, entry point main objects.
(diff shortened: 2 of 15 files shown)


sjrd:
Supersedes #2992.

gzm0:
Did you manually test the last commit? (I guess that's enough-ish).

sjrd:
Updated:
* 1 binary filter
* Don't touch `HelloWorld.scala`, and instead test the `def main(args: Array[String]): Unit` main method in the `sbt-plugin-test`. The problem is that our CI matrix still tests the `persistLauncher` case on the `helloworld`, and `persistLauncher` does not work with the new-style main (as documented).

Also, yes, I locally tested every one of the cases listed in the commit message ;)

PR #2997, opened 2017-06-07
Fix #2837: Remove the term "Scala.js defined"

### LINKED ISSUE #2837: Terminology: Remove Scala.js defined JS classes
Instead use:

- JS type / class where unambiguous
- native JS type / class
- non-native JS type / class

gzm0:
It is replaced by non-native JS class/type (except for tests).

With this commit, the following regexes (case insensitive) to not return
any result outside of tests and the definition of the @ScalaJSDefined
annotation itself.

```
Scala\.js[- ]defined
SJS ?defined
ScalaJSDefined
```


gzm0:
Let me know if you want me to:
- Fix the test names in this PR (I'd rather not to avoid too many conflicts).
- Not close #2837 with this PR.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1420,9 +1420,9 @@ object PrepJSInterop {
     val JSNativeClass = new OwnerKind(0x04)
     /** A native JS object, which extends js.Any. */
     val JSNativeMod = new OwnerKind(0x08)
-    /** A Scala.js-defined JS class/trait. */
+    /** A non-native JS class/trait. */
     val JSClass = new OwnerKind(0x10)
-    /** A Scala.js-defined JS oobject. */
+    /** A non-native JS oobject. */
```
sjrd: Take the opportunity to fix the `oo` in `oobject`.

compiler/src/test/scala/org/scalajs/core/compiler/test/InternalAnnotationsTest.scala
```diff
@@ -27,8 +27,8 @@ class InternalAnnotationsTest extends DirectTest with TestHelpers {
   }
 
   @Test
-  def sjsDefinedAnonymousClass: Unit = {
-    test("SJSDefinedAnonymousClass")
+  def anonymousJSClass: Unit = {
+    test("AnonymousJSClass", "scala.scalajs.js.annotation.internal.AnonymousJSClass")
   }
```
sjrd: There will be a silent conflict with https://github.com/scala-js/scala-js/pull/2991/commits/b3f5839e7b4f88e0c94457f0d37d7befdb392ab4 here. The fully qualified name won't be necessary.

sjrd:
Can you point me to a surviving "Scala.js-defined" in tests? It seems you did rename things in `compiler` tests and `testSuite`, so I'm not sure what else there is?

sjrd:
Anyway IMO we can close #2837 with this PR, as there won't be any user-facing references to "Scala.js-defined".

gzm0:
> Can you point me to a surviving "Scala.js-defined" in tests

For example here:
https://github.com/scala-js/scala-js/blob/bf0b6f49839be3b9db681a4632813ce12cd7939b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSSymbolTest.scala#L29

There are *plenty* of these left.

gzm0:
Also updated.

sjrd:
> There are plenty of these left.

Ah yes I see. Let's leave them alone for now, indeed.

gzm0:
Updated (there was a line number issue in a test).

compiler/src/test/scala/org/scalajs/core/compiler/test/JSSAMTest.scala
```diff
@@ -49,7 +48,6 @@ class JSSAMTest extends DirectTest with TestHelpers {
       def foo(x: Int): Int
     }
 
-    @ScalaJSDefined
```
sjrd: This change provokes some more line number changes below: at lines [61 and 64](https://github.com/gzm0/scala-js/blob/7edfee83c3b576aa993fe8f21f9a9b6378714c0f/compiler/src/test/scala/org/scalajs/core/compiler/test/JSSAMTest.scala#L61-L66), the line numbers should be 15 and 16, respectively.

Issue #2998, opened 2017-06-08
New Linker API


sjrd:
This is a summary extracted from the discussion in #2969, along with some small refinements I came up with while actually trying to implement that stuff.

## Goal

We want a new, simplified Linker API with one core "feature": a decomposition into 3 levels of API stability:

1. Super-stable API: always backward source and binary compatible across minor versions (except in the event of a catastrophic failure :P). It provides a reference point for people who "just want to link" their codebase, possibly with some configuration, but only configs that are meant to be supported "forever".
2. Stable-ish API: the API used by people who want to write their own linker plugins, or to compose some custom linkers, etc. It also exposes some less-stable configs. That API is supposed to stay backward source and binary compatible across minor releases unless maintaining compatibility would prevent progress. It should not be broken without a reasonably good reason.
3. Unstable, private-ish API: some things that we can freely break in minor releases, but that are technically *public* because they can be useful for some tools, e.g., the IR checker.

## Super-stable API, in `org.scalajs.core.tools.linker`

It is trimmed down to the bare minimum. The complete API surface looks like this:
```scala
abstract class Linker private[linker] () {
  def link(irFiles: Seq[VirtualScalaJSIRFile],
      moduleInitializers: Seq[ModuleInitializer],
      output: WritableVirtualJSFile, logger: Logger): Unit
}

final class ClearableLinker(newLinker: () => Linker) extends Linker {
  def link(...): Unit
  def clear(): Unit
}

class LinkingException(...) extends Exception

object StandardLinker {
  def apply(config: Config): Linker

  final class Config private () {
    val semantics: Semantics
    val moduleKind: ModuleKind
    
    val sourceMap: Boolean
    val parallel: Boolean
    ...
    private[linker] val outputMode: OutputMode

    def withSourceMap(sourceMap: Boolean): Config
    ...
    private[linker] def withOutputMode(outputMode: OutputMode): Config
  }

  object Config {
    def apply(semantics: Semantics, moduleKind: ModuleKind): Config
  }
}

final class Semantics private () { ... }
object Semantics { ... }

abstract class ModuleKind private[linker] () { ... }
object ModuleKind { ... }
```
Some things to note:
* There is no `class StandardLinker`. That's a stable-ish API.
[...]

PR #2999, opened 2017-06-08
Introduce the surface of the new linker API.

sjrd:
This commit introduces the portion of the new linker API described
in #2998 that is targeted for the 0.6.x series. It contains:

* `linker.StandardLinker._`
* `linker.Semantics`
* `linker.ModuleKind`

This is enough to implement the CLI, the sbt plugin, the `QuickLinker` and
partest's `MainGenericRunner` entirely in terms of the new API.

We deprecate `Linker.apply` in favor of `StandardLinker.apply` in order to
nudge other users of the linker towards the new API, assuming they use that
"easy" entry point (instead of manually manipulating front-end and back-end).

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -13,8 +13,8 @@ import sbt._
 
 import org.scalajs.core.tools.sem.Semantics
 import org.scalajs.core.tools.io._
-import org.scalajs.core.tools.linker.{ModuleInitializer, LinkingUnit}
-import org.scalajs.core.tools.linker.backend.{ModuleKind, OutputMode}
+import org.scalajs.core.tools.linker._
+import org.scalajs.core.tools.linker.standard._
 import org.scalajs.core.tools.jsdep.{JSDependencyManifest, ResolvedJSDependency}
 import org.scalajs.core.tools.jsdep.ManifestFilters.ManifestFilter
 import org.scalajs.core.tools.jsdep.DependencyResolver.DependencyFilter
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -232,6 +232,11 @@ object ScalaJSPlugin extends AutoPlugin {
         "`scalaJSUseMainModuleInitializer` is true",
         CTask)
 
+    val scalaJSLinkerConfig = SettingKey[StandardLinker.Config](
+        "scalaJSLinkerConfig",
+        "Configuration of the Scala.js linker",
+        BPlusSetting)
+
     val scalaJSNativeLibraries = TaskKey[Attributed[Seq[VirtualJSFile with RelativeVirtualFile]]](
         "scalaJSNativeLibraries", "All the *.js files on the classpath", CTask)
 
--- a/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
+++ b/cli/src/main/scala/org/scalajs/cli/Scalajsld.scala
@@ -11,15 +11,13 @@ package org.scalajs.cli
 
 import org.scalajs.core.ir.ScalaJSVersions
 
-import org.scalajs.core.tools.sem._
 import org.scalajs.core.tools.io._
 import org.scalajs.core.tools.logging._
 
-import CheckedBehavior.Compliant
+import org.scalajs.core.tools.linker._
+import org.scalajs.core.tools.linker.standard._
 
-import org.scalajs.core.tools.linker.{ModuleInitializer, Linker}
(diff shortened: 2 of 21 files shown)


sjrd:
CI only for now. I need to add a second commit that makes the `StandardLinker.Config` available as an sbt setting.

sjrd:
OK the second commit is in. Available for review :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/standard/OutputModeOps.scala
```diff
@@ -0,0 +1,33 @@
...
+
+    import StandardLinker.Config
+
+    /** Standard output mode. */
+    def outputMode: OutputMode = __self.outputMode
+
+    def withOutputMode(outputMode: OutputMode): Config =
+      __self.withOutputMode(outputMode)
+  }
+}
```
gzm0: I liked it better when this was in the package object...

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -217,6 +217,18 @@ object ScalaJSPlugin extends AutoPlugin {
...
+        "scalaJSLinkerConfig",
+        "Configuration of the Scala.js linker",
+        BPlusSetting)
+
+    val scalaJSLinkerBatchMode = SettingKey[Boolean](
+        "scalaJSLinkerBatchMode",
+        "Whether to run the Scala.js linker in batch mode, where nothing is " +
+        "cached nor incremental. This reduces the memory footprint, which " +
+        "can be particularly useful in a CI environment.",
+        BMinusSetting)
```
gzm0: Hum... I don't think the standard linker disables caching in batch mode. In fact, the reason this setting exists is that the standard linker does not have a batch mode...
Maybe we should simply add batch mode to the config so we can use it later on. WDYT?

sjrd: It is `ClearableLinker` which "disables caching" in batch mode, because it throws away the linker after every `link`. That's a nice thing about `ClearableLinker`, the the implementation of `Linker` it wraps doesn't have to deal with batch mode or not.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/standard/OutputModeOps.scala
sjrd: Me too, actually. But at the same time, in this version one can only
```scala
import org.scalajs.core.tools.linker.standard.OutputModeOps._
```
and, if they don't import anything else in `standard._`, at least they know that their code will keep working regardless of what happens to the rest of `standard`.

But I'm not convinced myself. Maybe it's better in the package object anyway.

sjrd: Updated with the contents of `OutputModeOps` moved back to the package object.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: I'm still unsure about this one. I kind of like that linkers don't have to worry about batch mode. But it's also disappointing that we have everything but one option in `scalaJSLinkerConfig`.

We could put the option inside `StandardLinker.Config`, and we might in the future enhance the linker to actually get rid of its caches more eagerly, e.g., throw out the optimizer as soon as it has done its job, rather than waiting for `link` to be completely finished.

sjrd: One thing in favor of a separate `scalaJSLinkerBatchMode` is that it would typically be set in a CI script, and since we can define it `in Global`, it is easy to just have `set scalaJSLinkerBatchMode in Global := true` from the CI, rather than set it for individual projects.

gzm0: I'm not sure I buy this last argument. *If* it is ok to set `scalaJSLinkerBatchMode in Global` it is also OK to set `scalaJSLinkerConfig in Global` (I'm unsure *whether* it is, but that's orthogonal). Therefore one can just modify the global default config.

IMO being able to write `set scalaJSLinkerBatchMode in Global := true` instead of `set scalaJSLinkerConfig in Global ~= (_.withBatchMode(true))` or even `set every scalaJSLinkerConfig ~= (_.withBatchMode(true))` does not outweigh the benefits of having a single unified config that can span all build systems).

sjrd: OK, you seem to have a stronger opinion than me on this matter, so I'll follow your lead.

sjrd: Done.

PR #3002, opened 2017-06-10
Fix #2989: Get rid of `tools.json` as a public API

### LINKED ISSUE #2989: Get rid of `tools.json` as a public API
`tools.json` is a simple cross-platform abstraction over JSON serialization and deserialization, and also offers a small implicit-based DSL to read/write data structures from/to JSON. It was initially introduced in https://github.com/scala-js/scala-js/pull/951 so that we could cross-compile the Tools.

However, now that `jsdep` has been moved out, the Tools do not use JSON at all anymore! According to `git grep json`, the only parts that still use our JSON abstraction are:
* `jsdependencies-core`, obviously, which was moved to https://github.com/scala-js/jsdependencies and will soon disappear from core
* `testAdapter`

Both are JVM-only projects, and both only use JSON internally (not in their public APIs).

Therefore, I suggest we get rid of `tools.json`.
* (maybe Step 0): move `tools.json` to the `jvm/src` sources and get rid of the JS version. This will maybe allow to remove some abstractions that were necessary for cross-platform-ness already.
* As a first step, we can move it as a private API of `testAdapter`, and duplicate it as a private API in https://github.com/scala-js/jsdependencies.
* Later, we can potentially simplify it down to using `json-simple` directly, and shave off some code.

It remains to be seen whether or not the simplification step will actually simplify things, but that does not preclude the benefit of not having `json` as a public API at all.

sjrd:
This was surprisingly difficult, due to `FrameworkDetector` and `HTMLRunnerTemplate`, which were leaking some useless dependency on `json` right into the sbt plugin.


sjrd:
This PR will have to be rebased on top #3006 for tests to pass.

sjrd:
Rebased.

gzm0:
Feel free to merge this if it passes (unless the rebase required substantial changes).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -641,29 +622,20 @@ object ScalaJSPluginInternal {
...
+          new sbt.testing.TaskDef(td.name, td.fingerprint,
+              td.explicitlySpecified, td.selectors)
         }
 
-        IO.write(output, HTMLRunnerTemplate.render(output.toURI,
-            name.value + " - tests", jsFileURIs, css.toURI,
-            (loadedTestFrameworks in testHtml).value,
-            (definedTests in testHtml).value))
+        HTMLRunnerBuilder.writeToFile(output, title, jsFiles,
+            frameworkImplClassNames, taskDefs.toList)
```
sjrd: The only non-trivial conflict was here, because of the change `testHtmlKey` -> `testHtml` in the base branch. But it was just a matter of taking the new implementation from this commit, and re-replacing `testHtmlKey` by `testHtml`.

PR #3003, opened 2017-06-11
Remove support for `exportsNamespace` and get rid of `envInfo`.

sjrd:


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -25,8 +25,24 @@ import org.scalajs.testsuite.utils.AssertThrows.assertThrows
...
+   *  module-global variable, which we can retrieve as if it were in the global
+   *  scope.
+   */
+  val exportsNamespace: js.Dynamic = {
+    if (Platform.isNoModule)
+      org.scalajs.testsuite.utils.JSUtils.globalObject
+    else if (Platform.isCommonJSModule)
+      js.Dynamic.global.exports
+    else
+      ???
```
gzm0: Maybe add something more descriptive?

library/src/main/scala/scala/scalajs/runtime/EnvironmentInfo.scala
```diff
@@ -1,46 +0,0 @@
...
- *  defined in its scope (and references an object). If so, it uses it as
- *  environment info.
- *  Missing, non-optional fields (according to this facade type) are initialized
- *  to default values, optional fields are kept as in the original object.
- *  Finally, [[js.Object.freeze]] is called on the object to avoid modification.
- *
- *  @groupname envInfo Scala.js environment configuration
- *  @groupprio envInfo 1
- */
-sealed trait EnvironmentInfo extends js.Object {
```
gzm0: I do not understand why you are getting rid of the typing here. It seems like this leads to very un-idiomatic Scala.js.

How about:

```
sealed trait EnviromentInfo // unchanged

object EnvironmentInfo {
  def envInfo: js.UndefOr[EnvironmentInfo] = {
    // ...
  }
}
```

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
sjrd: I thought that `NotImplementedError` was the best possible description in this case. I'll write a proper error message.

library/src/main/scala/scala/scalajs/runtime/EnvironmentInfo.scala
sjrd: Er, yes, good point. <insert shame emoji here>

sjrd:
Updated.

library/src/main/scala/scala/scalajs/runtime/EnvironmentInfo.scala
```diff
@@ -50,3 +41,18 @@ sealed trait EnvironmentInfo extends js.Object {
...
 }
+
+object EnvironmentInfo {
+  /** The value of the `__ScalaJSEnv` variable, if it exists. */
+  val envInfo: js.UndefOr[EnvironmentInfo] = {
+    import js.Dynamic.{global => g}
+
+    // We've got to use selectDynamic explicitly not to crash Scala 2.10
+    if (js.typeOf(g.selectDynamic("__ScalaJSEnv")) == "object" &&
+        js.DynamicImplicits.truthValue(g.selectDynamic("__ScalaJSEnv"))) {
```
gzm0: When is this line needed?

gzm0: `null` I guess? Consider simply comparing to `null` for improved clarity.

sjrd: Indeed. It was a direct translation of the JS code and I haven't given it any thought. Comparing to `null` would be better indeed.

PR #3004, opened 2017-06-11
Forbid top-level exports to invalid JS identifiers.

sjrd:
It won't be possible to export things as invalid JS identifiers when we export to the global scope or to ECMAScript 2015 modules. Therefore, we prevent that in the source code.

In the rare cases that someone does want to do something like that, they can do it manually. For example, in a CommonJS module, one could use
```scala
js.Dynamic.global.exports.`not-an-identifier` = 42;
```
in a module initializer to export under a non-identifier.


sjrd:
According to `git grep JSExportTopLevel`, no we don't. I'll add one.

sjrd:
Updated.

PR #3006, opened 2017-06-12
Enable Scalastyle in shared source directories.

sjrd:
Sadly, the Scalastyle sbt plugin only processes files in the `scalaSource` directory, rather than in all `unmanagedSourceDirectories`. This means that our shared sources were never checked by Scalastyle.

We fix this in this commit. The Scalastyle sbt plugin still being a non-Auto plugin from sbt < 0.13.5, there are a few complications. The problem is that settings of such plugins are applied *after* those defined in `.settings(...)`, hence overriding them (instead of the other way around). We have to resort to the `.settingSets` hack that still exists in 0.13. It won't exist in sbt 1.x, but then neither will non-Auto plugins, so by the time we get there, Scalastyle will have to use an `AutoPlugin`.


tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -237,6 +237,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
 
   private def checkMethodDef(methodDef: MethodDef,
       classDef: LinkedClass): Unit = withPerMethodState {
+    // scalastyle:off return
```
gzm0: Consider setting this for the entire IR checker once (maybe even adding a comment justifying why). As in:

```scala
// scalastyle:off return
// In the IR-Checker we allow returns for early returns for improved readability.
```

sjrd:
Updated.

> Any reason this is targeting master and not 0.6.x?

I was afraid of too many conflicts when merging to master laster, but rebasing on 0.6.x wasn't too painful, so the other direction won't either.

PR #3007, opened 2017-06-12
Fuse `testHtml{Fast,Full}Opt` in a unique `testHtml` task.

sjrd:
The new task honors `scalaJSStage` to decide which stage to use.

This gets rid of 2 hacks in our repo: one in our build, and one in the sbt plugin itself. It also allows `testHtml` to correctly use `jsExecutionFiles` that are dependent on the `scalaJSStage`, such as those produced by the `jsDependencies` plugin. Or more generally, correctly integrate with any custom tasks that produce different results depending on the stage.

The new behavior also makes more sense when we regard the HTML test runners as a peculiar form of JS envs. In that sense, `testHtml` is more similar to `test` than it was before.

### DIFF
--- a/TESTING
+++ b/TESTING
@@ -43,11 +43,11 @@ The following HTML-runners must be manually tested:
     examples/helloworld/helloworld-{2.10|2.11}{|-fastopt}.html
     examples/reversi/reversi-{2.10|2.11}{|-fastopt}.html
 
-The following sbt-plugin generated test runners must be manually tested (in both
-2.10 and 2.11):
+The following sbt-plugin generated test runners must be manually tested (with
+Scala 2.10, 2.11 and 2.12, and in `FastOptStage` and `FullOptStage`):
 
-    testingExample/testHtml{Fast,Full}Opt
-    testSuite/testHtml{Fast,Full}Opt
+    testingExample/testHtml
+    testSuite/testHtml
 
 ## Sourcemaps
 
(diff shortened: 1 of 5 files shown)


sjrd:
Tested locally on `testSuite/testHtml` and `testingExample/testHtml`, both in fastOpt and fullOpt.

julienrf:
Concretely how do you tell `testHtml` to be “in `fastOpt` or `fullOpt`? `testSuite/fastOptJS::testHtml`?

julienrf:
Oh, you have to first do a `scalaJSStage := fastOptStage`?

sjrd:
> Oh, you have to first do a `scalaJSStage := fastOptStage`?

Yes, that's right. Or more "idiomatically", `set scalaJSStage in Global := {Fast,Full}OptStage`.

julienrf:
Then the drawback is that you have to reload your build if you want to change this setting. It might be useful to be able to run the two tasks from the same build, what do you think?

sjrd:
Now it's on par with `test`. You also have to reload your build to run `test` in fullOpt.

Back in 0.5.x we could do these things in the same build, but the scoping of all the settings had become insanely complex, and we had to tap into plenty of sbt internals. It wasn't worth it.

gzm0:
Unfortunately I agree to this PR. As you can see from the changes, we already had to rely on hacks that assumed things about the build configuration we wouldn't (and couldn't) actually enforce. I think the only way to get interactive stages back (as we had in 0.5.x) is for sbt to fully support custom configuration axes, which is highly unlikely. So unfortunately, this is indeed the best we can do :(

PR #3008, opened 2017-06-12
Replace `Def.taskDyn` by `Def.settingDyn` where possible.

sjrd:
Regardless of whether the *result* key is a setting key or a task key, it is possible to use `Def.settingDyn` as long as the dynamic part (the part that decides what `Initialize` to return) only depends on settings. That's because a `TaskKey[T]` is nothing more than a `SettingKey[Task[T]]`, so the type parameter of `settingDyn` can be instantiated to `Task[T]` without any problem.

Assuming https://github.com/sbt/sbt/issues/3258 is eventually implemented, using `Def.settingDyn` will provide a better sbt experience in `inspect tree`.

Not all `taskDyn`s are removed, but the only remaining ones are all related to the launcher generation, and therefore will not survive in Scala.js 1.x.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -295,7 +295,7 @@ object ScalaJSPluginInternal {
   )
 
   private def dispatchTaskKeySettings[T](key: TaskKey[T]) = Seq(
-      key := Def.taskDyn {
+      key := Def.settingDyn {
         val stageKey = stageKeys(scalaJSStage.value)
         Def.task { (key in stageKey).value }
       }.value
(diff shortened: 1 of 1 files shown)


sjrd:
OK so this doesn't work at all, but for a different reason. I had to move `val s = streams.value` inside the anonymous `Def.task { }`, since `streams` is a task key and cannot be left in the `settingDyn`. However that means it now resolves to "the streams of a generic anonymous task". And that in turns that the `cacheDirectory` is the same for `{compile,test}:{fastOptJS,fullOptJS}`, so if one does `fastOptJS` then `fullOptJS`, the latter thinks it has nothing to do ...

It is unclear at this point what can be done about it.

sjrd:
OK fixed it. It turns out there was even no need to turn the `taskDyn`s of `{fast,full}OptJS` into `settingDyn`, because all their dependencies are in the non-dynamic part, which means `inspect tree` can show them *right now*. I just made that more explicit in a first commit.

sjrd:
Just for fun: I actually also managed to make it work with a `settingDyn`, but I had to reimplement all the logic to get the proper `TaskStreams` (what `streams.value` normally gives you) from within the anonymous task. The result is pretty ugly, but it works: https://github.com/scala-js/scala-js/compare/0.6.x...sjrd:no-taskdyn-wip

PR #3009, opened 2017-06-13
Make `js.Dynamic.global` user-defined again.

sjrd:
Instead of it being a primitive, it can simply be a user-defined native JS object annotated with `@JSGlobalScope`.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Dynamic.scala
+++ b/library/src/main/scala/scala/scalajs/js/Dynamic.scala
@@ -76,7 +76,35 @@ sealed trait Dynamic extends js.Any with scala.Dynamic {
 /** Factory for dynamically typed JavaScript values. */
 object Dynamic {
   /** Dynamic view of the global scope. */
-  def global: js.Dynamic = throw new java.lang.Error("stub")
+  @js.native
+  @JSGlobalScope
+  object global extends js.Any with scala.Dynamic { // scalastyle:ignore
+    /** Calls a top-level method (in the global scope). */
+    @JSBracketCall
+    def applyDynamic(name: String)(args: js.Any*): js.Dynamic = js.native
+
+    /** Reads a top-level variable (in the global scope). */
+    @JSBracketAccess
+    def selectDynamic(name: String): js.Dynamic = js.native
+
+    /** Writes to a top-level variable (in the global scope). */
+    @JSBracketAccess
+    def updateDynamic(name: String)(value: js.Any): Unit = js.native
+
+    /* The following method is a protection against someone writing
+     * `js.Dynamic.global(args)`. It that method were not there, that call
+     * would silently desugar into
+     * `js.Dynamic.global.applyDynamic("apply")(args)`, which is very
+     * unexpected and will produce confusing run-time errors. Better to have
+     * a straightforward compile-time error.
+     */
+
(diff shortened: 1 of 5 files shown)


library/src/main/scala/scala/scalajs/js/Dynamic.scala
```diff
@@ -76,7 +76,27 @@ sealed trait Dynamic extends js.Any with scala.Dynamic {
...
+
+    /** Writes to a top-level variable (in the global scope). */
+    @JSBracketAccess
+    def updateDynamic(name: String)(value: js.Any): Unit = js.native
+
+    /** Cannot be called--provides a compile-time error instead of a silent
+     *  run-time error if one tries to do `js.Dynamic.global(something)`.
+     */
+    @deprecated("The global scope cannot be called as function.", "forever")
+    def apply(args: js.Any*): js.Dynamic = js.native
```
gzm0: Why is this method necessary?

sjrd: Because otherwise `js.Dynamic.global(foo)` silently desugars to `js.Dynamic.global.applyDynamic("apply")(foo)`, which compiles and links, but doesn't do what it looks like it's going to do.

Wasn't the comment explicit enough?

gzm0: The "otherwise it desugars to ..." part wasn't clear to me.

sjrd: I'll add a non-Scaladoc comment about that, then.

sjrd:
> Please add a test that selects x on js.Dynamic.global :)

It definitely won't work in 2.10. But I don't think we should care. A *global* variable named `x` is bad enough. Combine that with still using 2.10 to access it, and you're beyond redemption.

gzm0:
> Combine that with still using 2.10 to access it, and you're beyond redemption.

Agreed. However, I do not like it at all to set a precedent that essentially says: "If you use a large enough combination of not recommended but officially supported stuff it might not work anyways."

I would feel more comfortable with officially dropping 2.10 support in 1.0.x (IIRC there have been discussions about that, although I'm not sure the maintenance burden is so high).

sjrd:
It would definitely be an easier decision to make if we dropped 2.10. I think we should probably do it, but that's a discussion for another day.

In this precise case, actually I would argue that we should drop the `x` workaround for `js.Dynamic` altogether, not only for `global`. Yes, that means some 2.10 code won't work anymore. But if we look at this workaround, we were really fighting a usability issue of Scala.js for dead simple code, at a time where 2.10 was all there was. Nowadays, virtually everyone using 2.10 with Scala.js is a library maintainer, and in libraries you're unlikely to access a very specific field named `x` without static typing. And even if you are, well you can always fall back on `selectDynamic("x")`/`applyDynamic("x")(...)`.

The `x` workaround doesn't pull its weight anymore, I think.

library/src/main/scala/scala/scalajs/js/Dynamic.scala
sjrd: I have added the comment, but haven't changed anything else at this point.

gzm0:
Let's leave it like this. Tough luck 2.10 users :)

Issue #2344, opened 2016-04-20
Suggesting `NullOr` similar to `UndefOr` for JS-structures


lastsys:
When handling JSON-objects some fields can contain `null` values. It would be nice to have such values parsed automatically to `Option` where `null` represents `None`.

Example:

``` json
{
  "name": null
}
```

could be defined as

``` scala
trait MyObj extends js.Object {
  val name: js.NullOr[String]
}
```

which means that when accessed we get

``` scala
val name: Option[String] = myObj.name.toOption
```

Not sure what to write if, for some reason, a field can be either `undefined`, `null` or `String`.

sjrd:
That does not really make sense. `String` is *already* nullable, because that's how it works in Scala.

With a nullable type such as string, you'd do `Option(myObj.name)` instead of `myObj.name.toOption` to turn `null` into `None`.

The absence of null-safety in Scala is unfortunate, but it's there anyway. I don't think this would significantly improve the situation. If you really want such a thing, you can define it yourself with something like:
```scala
@js.native
sealed trait OrNull[+A] extends js.Any

object OrNull {
  implicit class OrNullOps[+A](val self: OrNull[A]) extends AnyVal {
    def toOption: Option[A] = Option(self).asInstanceOf[Option[A]]
  }
}
```

lastsys:
Yes, `String` is nullable, and that is the problem. `null` introduces a lot of problems throughout the code and being able to eliminate `null` directly from the source would reduce the risk for errors.

I agree that this is not _necessary_ functionality, but it is nice to have. I would suggest some part of the standard lib with utilities for writing facades to enforce best practices.

sjrd:
What about using `A | Null`, e.g., `String | Null`. You can even get a pretty good API with a user-defined implicit pimp:
```scala
package helloworld

import scala.scalajs.js
import scala.scalajs.js.|

object NullOrImplicits {
  implicit class NullOrOps[A](val self: A | Null) extends AnyVal {
    @inline def isDefined: Boolean = self != null

    @inline private def forceGet: A = self.asInstanceOf[A]

    @inline def get: A =
      if (isDefined) forceGet
      else throw new NoSuchElementException("null")

    @inline def fold[B](ifEmpty: => B)(f: A => B): B =
      if (isDefined) f(forceGet)
      else ifEmpty

    @inline def map[B](f: A => B): B | Null =
      fold[B | Null](null)(f(_))

[...]

PR #3011, opened 2017-06-13
Define `js.UndefOr[A]` as a mere alias to `A | Unit`.

sjrd:
This will give a much better status to `js.UndefOr` in a compiler that supports true union types (understand, Dotty).


sjrd:
Let's see what breaks :stuck_out_tongue_closed_eyes:

sjrd:
Well, this (surprisingly) worked out quite easily, with very few changes. The only user-level change is that 2.10 requires some manual ascriptions with `orElse`.

@gzm0 WDYT? Shall we try and put that into M1 and see what feedback we get?

gzm0:
Ship it!

gzm0:
Also see #2344.

PR #3012, opened 2017-06-14
Fix #2945: Leave it to sbt to buffer logs

### LINKED ISSUE #2945: Scala.js SBT plugin ignores logBuffered setting
History: https://github.com/typelevel/cats-effect/pull/51

The `logBuffered in Test := false` SBT setting is designed to allow test frameworks (and really anything with outputs) to produce feedback incrementally, typically as each individual test case completes, rather than exclusively as the test suite completes.  This is really really useful for extremely long running test suites, both from a usability and from a "it works in Travis" standpoint.  For context, Travis watches stdout/stderr, and if it doesn't see any new output in 3 minutes, it will assume the process is hung and kill it.  You can delay this through various dodgy mechanisms, but it doesn't work all that well.  Incremental output is the more scalable solution for long-running suites comprised of numerous fast test cases.

Unfortunately, the setting has no effect with `.js` projects.  It appears that the setting is being overridden by Scala.js somehow, and it's not entirely clear how.  Alex reported on some more minimized testing [here](https://github.com/typelevel/cats-effect/pull/51#issuecomment-300745288).

gzm0:
When we created the test adapter, we implemented log buffering ourselves
to prevent interleaved output when running in parallel. However, this is
not necessary since sbt implements its own log buffering. To make
matters worse, this interfered with it.

We fully remove log buffering in the test adapter while keeping an
escape hatch (via Java system properites) in case something goes wrong.

### DIFF
--- a/test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSTask.scala
+++ b/test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSTask.scala
@@ -14,6 +14,7 @@ import org.scalajs.core.tools.json._
 import org.scalajs.jsenv._
 
 import scala.collection.mutable
+import scala.util.Try
 
 import sbt.testing._
 
@@ -45,7 +46,17 @@ final class ScalaJSTask private (
     slave.send("execute:" + jsonToString(data))
 
     // Prepare result handler
+    // TODO rip this out once we know it doesn't garble the logs.
+    val shouldBufferLog = {
+      val propName = "org.scalajs.testadapter.bufferlog"
+      Try(System.getProperty(propName, "false").toBoolean).getOrElse(false)
+    }
+
     val logBuffer = mutable.Buffer.empty[LogElement[_]]
+
+    val logger: LogElement[_] => Unit =
+      if (shouldBufferLog) logBuffer += _
+      else _.call(loggers)
 
     val doneHandler: LoopHandler[List[TaskInfo]] = {
       case ("ok", msg) => Some(fromJSON[List[TaskInfo]](readJSON(msg)))
@@ -53,7 +64,7 @@ final class ScalaJSTask private (
 
     val handlerChain = (
         eventHandler(handler) orElse
-        loggerHandler(logBuffer) orElse
+        loggerHandler(logger) orElse
         runner.msgHandler(slave) orElse
         doneHandler)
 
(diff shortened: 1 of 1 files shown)


gzm0:
I have manually tested that:

- Unchanged sbt settings still give nice output.
- Setting `logBuffered in testSuite := true` gives interleaved output (which I take as proof that buffering is disabled).
- Running with `JAVA_OPTS=-Dorg.scalajs.sbttestadapter.bufferlog=true sbt` and `logBuffered in testSuite := true` gives nice output.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSTask.scala
```diff
@@ -45,24 +46,38 @@ final class ScalaJSTask private (
     slave.send("execute:" + jsonToString(data))
 
     // Prepare result handler
-    val logBuffer = mutable.Buffer.empty[LogElement[_]]
+    // TODO rip this out once we know it doesn't garble the logs.
+    val shouldBufferLog = {
+      val propName = "org.scalajs.sbttestadapter.bufferlog"
```
sjrd: This should actually be `org.scalajs.testadapter.bufferlog`, i.e., without the `sbt` in `sbttestadapter`. That's the actual package name of the test adapter. (The directories are wrongly `sbttestadapter`, see #3002 about that.)

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSTask.scala
```diff
@@ -111,10 +125,7 @@ final class ScalaJSTask private (
 }
 
 object ScalaJSTask {
-  private final class LogElement[T](index: Int,
-      log: Logger => (T => Unit), data: T) {
-    def call(arr: Array[Logger]): Unit = log(arr(index))(data)
-  }
+  type LogElement = (Int, Logger => Unit)
```
sjrd: Why not keeping the class with named values? Simply a
```scala
private final case class LogElement(index: Int, log: Logger => Unit)
```
I see several advantages:

* There's a name to that `Int` value
* It would avoid the weird `x._2(loggers(x._1))`
* It would avoid the automatic tuple insertion involved in `logger(index, l => level(l)(msg))` and `logger(index, _.trace(throwable))`

gzm0:
Updated

PR #3014, opened 2017-06-14
Fix #3013: Use `genExpr()` for the qualifier of super calls.

### LINKED ISSUE #3013: Super call to a mixin of an outer class confuses the codegen in 2.12+
Same bug as https://github.com/scala/scala/pull/5944 on the JVM side, for exactly the same reason.

sjrd:
Since Scala 2.12, the `qual` part of a super call is not always `this`. It can be `this.$outer()` or similar things. Therefore, we cannot blindly assume that the receiver is `this`, nor that the super call is a JS super call iff the current class is a JS class.

Instead, we correctly use `genExpr(qual)` and inspect `isRawJSType(qual.tpe)`.


sjrd:
This will need a rebase on top of #3006 for tests to pass.

PR #3019, opened 2017-06-16
Deprecate `js.JSApp` and do not special-case them in the sbt plugin.

sjrd:
For backward source compatibility, we add in `js.JSApp` a forwarder
```scala
def main(args: Array[String]): Unit = main()
```
which allows existing `js.JSApp` to still be recognized and run.

---

This completes the changes that were originally proposed in #2992.

### DIFF
--- a/TESTING
+++ b/TESTING
@@ -13,7 +13,7 @@ For each major Scala version on a *NIX distro and a Windows distro:
         mkdir bin
         echo '
           object Foo {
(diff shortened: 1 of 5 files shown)


sjrd:
Yes, we're very close to having a nice, small, tidy, magic-free sbt plugin :smile:

sjrd:
Updated to fix a Scaladoc link error.

PR #3021, opened 2017-06-16
Fix #2877: Internal JS envs config amenable to bincompat evolution.

### LINKED ISSUE #2877: Investigate an API for JS envs configs amenable to binary-compatible evolutions
Currently the JS envs have all their config options as default arguments of their constructors. As we know, this is impossible to evolve in a binary compatible way (e.g., add new options). We should investigate how to provide an API that is amenable to binary-compatible evolutions. It's tricky because of the OO nature of JS envs, so each subclass can add their own config, etc.

sjrd:
The internal configuration of (abstract) classes in the JS env API is now `protected def`-based, rather than based on constructor parameters.

This can be evolved in binary-compatible ways, as we can add new `protected def`s with default implementations in the superclasses.


sjrd:
> Should we also make the base traits abstract classes?

Do you mean `ComJSEnv` and things like that? I don't think that's necessary, because they are pure interfaces and should remain so. Moreover, I don't think that's even possible. For example we have `ExternalJSEnv` that extends `AsyncJSEnv` only, but its subclasses add `extends ComJSEnv`.

> Also: No-reusing bugs :P

I don't understand what that means :s

sjrd:
I do have some more simplifications in mind, but not for the time-frame of M1. Notably I'd like to remove the synchronous APIs.

gzm0:
>> Also: No-reusing bugs :P
> I don't understand what that means :s

You are "fixing" #2877 multiple times. That's not ideal.

sjrd:
Oh did I do that? Where?

My guess would've been https://github.com/scala-js/scala-js/pull/3001 but that one definitely does not fix #2877.

gzm0:
Ah... Indeed. I must have misread the references in #2877. Having the amended commits still being referenced is so confusing :-/

PR #3023, opened 2017-06-16
Fix #2813: Clean up the public API of the sbt plugin

### LINKED ISSUE #2813: Shave off some (public) sbt keys
There are way too many keys for JS dependency loading and JS Env loading. Candidates for removal:

- `scalaJSConfigurationLibs` (only used for `scalaJSJavaSystemProperties`).
- `loadedJSEnv` (only used due to `LinkingUnitJSEnv`).
- `scalaJSPhantomJSClassloader` (only used for PhanomJS).
- `resolvedJSEnv` (can be replaced by a default value on `jsEnv`).

sjrd:
Ref #2813.


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -495,8 +480,11 @@ object ScalaJSPluginInternal {
 
   val scalaJSTestFrameworkSettings = Seq(
       loadedTestFrameworks := {
-        // use assert to prevent warning about pure expr in stat pos
-        assert(scalaJSEnsureUnforked.value)
+        if (fork.value) {
+          throw new MessageOnlyException(
+              "`test` tasks in a Scala.js project require " +
+              "`fork in Test := false`.")
+        }
```
gzm0: Why do we require it here?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -536,24 +532,13 @@ object ScalaJSPluginInternal {
...
+  val scalaJSTestSettings: Seq[Setting[_]] = (
       scalaJSTestBuildSettings ++
-      scalaJSRunSettings ++
       scalaJSTestFrameworkSettings ++
       scalaJSTestHtmlSettings
   )
 
-  val scalaJSDefaultBuildConfigs = (
-      inConfig(Compile)(scalaJSConfigSettings) ++ // build settings for Compile
-      inConfig(Test)(scalaJSTestBuildSettings)
```
gzm0: I had a hunch that somebody requested this, but I can't find any reference in the history anymore (seems like it got introduced in [5e1f3b3](https://github.com/scala-js/scala-js/commit/5e1f3b3ddbf16ccfde6cecd655f027fa780f4f3d#diff-49772688724bf66a3c342ddc1451f277R525)). So I guess its fine if we nuke it.

sjrd:
`scalaJSIsSnapshotVersion` should indeed go. `scalaJSBinaryVersion` can go too I think (it can be reconstructed with things in `ScalaJSCrossVersion`). I'll add a commit for that.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: Because it's a common dependency of all test task, and because from there on in the graph, things go wrong in sbt tasks. It seems fitting, but maybe there exists some more obscure task in sbt that would do the job better? If yes and unless you have a concrete suggestion now, that can always be improved later.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: Yes someone had requested that back in the dark days of 0.1-SNAPSHOT or something. At the time I could not yet make a strong enough case against reusing the same class files for JVM and JS, and someone was vocal enough to make me add this abomination. It's time to remove i, though.

gzm0: SG

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Oh, so things actually *do* go wrong if we try and run the test stuff in a forked JVM? If so, I'm fine with having it here.

sjrd: Oh yes, it goes horribly wrong, with things like obscure serialization errors.

gzm0: :P Good to know

project/Build.scala
```diff
@@ -359,23 +359,6 @@ object Build {
...
-            Some(Resolver.sftp(
-                s"scala-js-$snapshotsOrReleases",
-                "repo.scala-js.org",
-                s"/home/scalajsrepo/www/repo/$snapshotsOrReleases")(
-                Resolver.ivyStylePatterns) as (user, pass))
-          case _ =>
-            None
-        }
-      }
-  )
```
sjrd: No idea how this survived this long. It hasn't been used in 0.1.

sjrd:
Updated with a commit to remove the version fields.

PR #3031, opened 2017-06-26
Fix #2842: Remove the jsdom environment from the core.

### LINKED ISSUE #2842: Move JSDOMNodeJSEnv out of the core

sjrd:
Node.js with jsdom support moved to the separate repository https://github.com/scala-js/scala-js-env-jsdom-nodejs.

After these changes, `git grep -i jsdom` turns up empty.


sjrd:
https://github.com/scala-js/scala-js-env-jsdom-nodejs/pull/1 must be merged before this gets in.

gzm0:
By the way I'm not 100% convinced that the public interface we expose for `JSEnv` implementations (i.e. `AbstractNodeJSEnv` and `ExternalJSEnv`) is optimal. I'd like to rethink it before 1.0.0 (maybe more composition and less inheritance ;)).

gzm0:
But I'm OK to go forward with this to unblock 1.0.0-M1.

sjrd:
I'd also like to rethink the `JSEnv` public interface. It's definitely not optimal.

PR #3037, opened 2017-06-28
Fix #3036: Workaround GCC's dce being too enthusiastic.

### LINKED ISSUE #2708: fastOptJS code not working in JavaScriptCore(JSC)!

### LINKED ISSUE #3036: Exports are sometimes dce'd by GCC in NoModule
Revival of the problem that we worked around with https://github.com/scala-js/scala-js/commit/6680f72e8167af954d58797d26e943cf3326c4af#diff-3c8817e2f33c9f11365145b6a952af00R346.

The workaround was removed in https://github.com/scala-js/scala-js/commit/d28845a866bca1968658de1556577564cf9b74b3 to fix #2708. It appeared not to be necessary anymore at the time.

It turns out we broke it in master, most likely due to https://github.com/scala-js/scala-js/commit/d07af7c79b0ecca3498c0fe96c8fc223ad4abdb4#diff-ec643f0e632cf718a6fcd31e201493cb.

Therefore, we'll have to reintroduce the workaround, but without recreating #2708. Fortunately, it seems this issue only happens in `NoModule` (most likely because GCC treats `exports` as magical), and #2708 is only relevant in `CommonJSModule`. We can therefore reintroduce the workaround only in `NoModule`.

sjrd:
In `NoModule`, sometimes GCC thinks that `$e` is not used and can therefore be dce'ed away ... along with all our exports. We work around that bug by forcing a non-dce'able read of `$e`.

This workaround existed in the past. It was first introduced in 6680f72e8167af954d58797d26e943cf3326c4af. It was later removed in d28845a866bca1968658de1556577564cf9b74b3 to fix #2708, as it appeared not to be necessary anymore. However, we later broke what it was solving in d07af7c79b0ecca3498c0fe96c8fc223ad4abdb4. This commit reintroduces the previous workaround, but only in `NoModule` so that we do not recreate #2708.

### DIFF
--- a/tools/scalajsenv.js
+++ b/tools/scalajsenv.js
@@ -13,6 +13,8 @@ const $e = exports;
 //!else
 // TODO Do not use global object detection, and rather export with actual `var` declarations
 const $e = (typeof global === "object" && global && global["Object"] === Object) ? global : this;
+// #3036 - convince GCC that $e must not be dce'ed away
+this["__ScalaJSWorkaroundToRetainExportsInGCC"] = $e;
 //!endif
 
 // Linking info - must be in sync with scala.scalajs.runtime.LinkingInfo


sjrd:
I manually tested the HTML runner of Reversi and testingExample in fullOpt, obviously.

Issue #2645, opened 2016-11-06
Make test adapter's JSEnv usage more economical


gzm0:
We have had reports of people having their CI overloaded, due to too many JSEnvs being spawned, even when running with `parallelExecution in Test := false`.

This is because each test run brings up at least three JSEnvs:

- Test framework detection (through [FrameworkDetector][1])
- The test master
- A single slave

We should merge these three spawns into a single one.

Merging the slave into the master is relatively easy, since the master already is able to do what the slave can do. We just need to organize the task dispatcher better to reuse the already launched master. There might be some subtleties while handling the com queue (as the master might be allowed to send other messages back).

Merging the test framework detection into the master will require substantial changes to the test adapter and the interface: Instead of creating a master once the framework we want to run with is known, we need to create a master upfront, have it perform framework detection and then instantiate and multiplex between frameworks as necessary.

Note that this will have the pleasant side-effect, that testing with multiple test frameworks in the same project will have substantially reduced JSEnv consumption as well: Currently, we spawn a JSEnv per thread per test framework per project. Afterwards we'll only spawn a JSEnv per thread per project.

Note that we cannot go below this boundary because of the following reasons:
- Per thread: We do not *want* to go below this to exploit parallelism if requested.
- Per project: We must load the JS code into the JSEnv upfront. So there is no way for us to re-use a running JSEnv with a different project (and hence different JS code).

[1]: https://github.com/scala-js/scala-js/blob/master/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/FrameworkDetector.scala

gzm0:
Assigning this to 1.0.0 so we do take a moment and think whether there are any incompatible changes (on the test adapter interface) we want to make. The thing itself clearly does not need to be fixed in 1.0.0.

gzm0:
I'm working on incremental changes to this that I hope to be able to submit to the 0.6.x branch. Thankfully we can evolve the JVM/JS interface without breaking anything.

sjrd:
@gzm0 Do you still plan to have this is the immediate future, or should we delay until 0.6.19? It's the last issue scheduled for 0.6.18, and we need to publish 0.6.18 before 1.0.0-M1.

Otherwise I'll start a release cycle for 0.6.18.

gzm0:
[...]

Issue #2751, opened 2017-01-25
js.Date functions can return NaN, so they should be typed as Double


sjrd:
**Was: On Firefox, (some) js.Date functions can return -0, which cause CCEs**

Reproducible with `testSuite/testHtmlFastOpt` with Firefox. Also:
```scala
    import js.DynamicImplicits._

    val utc = java.util.Date.UTC(0, 0, 0, 0, 0, 0)
    val date = new js.Date(utc)

    println("getUTCDate = " + date.getUTCDate())
    println("getUTCMonth = " + date.getUTCMonth())
    println("getUTCFullYear = " + date.getUTCFullYear())

    val dyndate = date.asInstanceOf[js.Dynamic]
    val h = dyndate.getUTCHours()
    println(js.typeOf(h))
    println(h)
    println(1.0 / h)

    println("getUTCHours = " + date.getUTCHours())
    println("getUTCMinutes = " + date.getUTCMinutes())
    println("getUTCSeconds = " + date.getUTCSeconds())
```
gives:
```
getUTCDate = 31
getUTCMonth = 11
getUTCFullYear = 1899
number
0
-Infinity
uncaught exception: scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: 0 is not an instance of java.lang.Integer
```

sjrd:
Fixing this requires a source-incompatible change, so it can't go in 0.6.x.

sjrd:
So, actually, it's much worse than that: *all* methods of `js.Date` can return `NaN` on *all* engines, if the `js.Date` was created with an invalid string:
```js
C:\Users\Sepi>node
> var d = new Date("foobar")
undefined
> d.getYear()
NaN
> d.getFullYear()
NaN
> d.getDay()
NaN
> d.getDate()
NaN
> d.getMonth()
NaN
```
etc.

So whether Firefox is allowed to return -0 in some cases is pretty irrelevant at this point. All the methods must clearly marked as returning `Double`s.

Also, it's OK to call one of the `setX` method with `NaN`, in which case the date will be set to an invalid date, and therefore all methods will return `NaN` after that.

PR #3040, opened 2017-07-01
Add an RPC layer to the test interface

gzm0:
This replaces the ad-hoc message passing with a simple RPC layer and is
the first step in implementing #2645. It also mostly removes
dependencies on tools.json by replacing the serialization with a binary
protocol.

The next steps will be to consolidate functionality into a single JS
controller so that we do not need to restart the JS VMs so often.

Note that this change is potentially dangerous as it exercises
functionality in JSEnvs that has never been used before. Notably:

- Calling {`send`|`close`} and `receive` concurrently.
- Relying on blocking calls to `receive` to throw if another thread
  calls `close`.

Since in the 0.6.x branch most of the enviroments are still in the core
repository, the risk is minimal.


gzm0:
Let's see what the CI has to say. I haven't tested this on all JSEnvs.

gzm0:
I have tried to manually test this (using `multiTestJS/test` in the sbt test project) with the selenium envs and it hangs indefinitely. I suspect scala-js/scala-js-env-selenium#55 to be the culprit.

The way I see it is that if `scala-js-env-selenium` does not pass the test kit we publish, it is essentially too broken for us to support it. We have the following options:

- I go and fix `scala-js-env-selenium`, we block this PR on it.
- We submit this, but probably break all users of `scala-js-env-selenium`.
- We retarget this to 1.0.0, giving us more time to fix envs.

Note that for 1.0.0 I will increase the rigidness of the `ComJSEnvRunner` interface API significantly and harden the tests. This is necessary to give reasonable failure behavior to the test adapter. Currently its behavior in case of a channel break is essentially unspecified.

sjrd:
> We submit this, but probably break all users of scala-js-env-selenium.

That option is definitely not acceptable. We have users who exclusively use scala-js-env-selenium for their tests, because it's the only thing that works for them.

> We retarget this to 1.0.0, giving us more time to fix envs.

Does it really give us more time? I mean, we'll have to fix the envs before we land this PR, whether or not it lands in 0.6.x or 1.x.

gzm0:
Yeah, you are right. It doesn't seem like there is a real choice. The only difference with targeting 1.0.0 is that I can write a rock-solid test suite first, publish it and then write against it. But indeed, I probably should fix what is there first :-/

gzm0:
So I'll proceed by working on the selenium-env issues. You can start reviewing this already. I don't expect major changes.

sjrd:
OK I'll review as soon as I'm done with 1.0.0-M1.

test-common/src/test/scala/org/scalajs/testcommon/RPCCoreTest.scala
```diff
@@ -0,0 +1,101 @@
...
+  def unboundEndpoint: Unit = {
+    assertRPCException(y.call(eps.simple)(()))
+  }
+
+  @Test
+  def remoteException: Unit = {
+    val msg0 = "My message for the outer exception"
+    val msg1 = "My message for the inner exception"
+    x.attach(eps.simple)(
+        (_: Unit) => throw new Exception(msg0, new Exception(msg1)))
```
sjrd: Both `simpleEndpoint` and `remoteException` do an `x.attach(eps.simple)`. How does that not fail?

test-common/src/test/scala/org/scalajs/testcommon/SerializerTest.scala
```diff
@@ -0,0 +1,19 @@
+package org.scalajs.testcommon
+
+import org.junit.Test
+import org.junit.Assert._
+
+class SerializerTest {
+  def roundTrip[T : Serializer](x: T): T =
```
sjrd: Here also the space before colon should have been caught by Scalastyle.

test-interface/src/main/scala/org/scalajs/testinterface/internal/TaskInfoBuilder.scala
```diff
@@ -0,0 +1,24 @@
+package org.scalajs.testinterface.internal
+
+import sbt.testing._
+
+import org.scalajs.testcommon.{TaskInfo, Serializer}
+
+private[internal] object TaskInfoBuilder {
+  def detachTask(task: Task, runner: Runner): TaskInfo = {
```
sjrd: Why are these methods called `detachTask` and `attachTask`?

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -0,0 +1,187 @@
...
+    promise.complete(x)
+    promise.future
+  }
+}
+
+private[scalajs] object RPCCore {
+  type OpCode = Byte
+
+  /** Exception thrown if a remote invocation fails. */
+  class RPCException(c: Throwable) extends Exception(null, c)
```
gzm0: Not statically, no. Yes, you can call `getClass.getName` and get `RPCException` back, but no publicly exposed type contains `RPCException`. So I don't see how this is a problem (anymore than returning an instance of a private type for `Object`).

I specifically do not want anyone to be able to match on this. It is only here for debugging purposes (but for this it's important to retain the stack trace of the RPC client). An alternative would be to throw just an `Exception` but that's also not very nice :S

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
```diff
@@ -0,0 +1,290 @@
...
+import java.io._
+
+private[scalajs] trait Serializer[T] {
+  def serialize(x: T, out: Serializer.SerializeState): Unit
+  def deserialize(in: Serializer.DeserializeState): T
+}
+
+private[scalajs] object Serializer extends SerializerPlatformExtensions {
+  /* Serialization states. Serialization is currently stateless, but in the
+   * future we might want to deduplicate things.
```
gzm0: In the sense that the same value will always get serialized into the same bytes, no matter where it is in the bytestream. It might be a good idea to change this (for example for package prefixes). Since a lot of data seems to be redundant.

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
```diff
@@ -0,0 +1,290 @@
...
+  def serialize[T : Serializer](t: T): String =
+    withOutputStream(Serializer.serialize(t, _))
+
+  def deserialize[T : Serializer](s: String): T =
+    withInputStream(s)(Serializer.deserialize[T](_))
+
+  @inline
+  def withInputStream[T](s: String)(body: DataInputStream => T): T = {
+    val in = new DataInputStream(
+        new ByteArrayInputStream(s.getBytes(latin1Charset)))
```
gzm0: Ah, ah, this is how you escape the deprecation of converting Chars to Bytes :) Of course *much* better.

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
```diff
@@ -0,0 +1,290 @@
...
+
+    def deserialize(in: DeserializeState): TaskDef = {
+      new TaskDef(in.get[String](), in.get[Fingerprint](), in.get[Boolean](),
+          in.get[List[Selector]]().toArray)
+    }
+  }
+
+  implicit object StatusSerializer extends Serializer[Status] {
+    def serialize(x: Status, out: SerializeState): Unit =
+      out.s.writeInt(x.ordinal)
```
gzm0: I made a distinction between "data" primitive values and "control" primitive values. Maybe not a good idea?

test-interface/src/main/scala/org/scalajs/testinterface/internal/TaskInfoBuilder.scala
gzm0: Because they attach/associate with the runner. The task might need to report things to a runner, so it needs to retain a reference to it. I recommend you read the Scaladoc of `serializeTask` and `deserializeTask`.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
sjrd: Ah OK you're right.

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
sjrd: OK that's not at all clear to me from the current comment. Could you add that explanation in the comment?

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
sjrd: Ah. I'm not sure the distinction is relevant. At least I don't think it is relevant enough to outweigh the potential confusion ;)

test-interface/src/main/scala/org/scalajs/testinterface/internal/TaskInfoBuilder.scala
sjrd: Ah I see. OK, makes sense.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -0,0 +1,187 @@
...
+
+  /** Subclass should call this whenever a new message arrives */
+  final protected def handleMessage(msg: String): Unit = {
+    Serializer.withInputStream(msg) { in =>
+      val opCode = in.readByte()
+      val callId = in.readLong()
+
+      opCode match {
+        case RPCCore.REPLY_OK =>
+          Option(pending.remove(callId)).foreach { p =>
```
gzm0: It can happen in a race with `close`.

gzm0: It can happen in a race with `close`.

sjrd: Whether the duplicate post was intended or not, this is hilarious 😄

sjrd: That said, I think it deserves a comment ;)

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -0,0 +1,187 @@
...
+  final def detach(ep: RPCEndpoint): Unit = {
+    val old = endpoints.remove(ep.opCode)
+    require(old != null, "Endpoint was not attached.")
+  }
+
+  /** Close the communication channel. */
+  def close(): Unit = {
+    for {
+      (callId, p) <- pending.asScala
+      failing <- Option(pending.remove(callId))
```
gzm0: I'm using `keySet` which is guaranteed to be safe for this kind of operations: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html#keySet()

test-common/src/main/scala/org/scalajs/testcommon/RPCEndpoint.scala
```diff
@@ -0,0 +1,27 @@
...
+  implicit val reqSerializer: Serializer[Req]
+  implicit val respSerializer: Serializer[Resp]
+}
+
+private[scalajs] object RPCEndpoint {
+  /** Helper type for readability */
+  type EP[Rq, Rp] = RPCEndpoint { type Req = Rq; type Resp = Rp }
+
+  def ep[Rq, Rp](opc: RPCCore.OpCode)(implicit rqs: Serializer[Rq],
+      rps: Serializer[Rp]): EP[Rq, Rp] = {
```
gzm0: Much better indeed.

test-common/src/main/scala/org/scalajs/testcommon/Serializer.scala
gzm0: Got it. Will remove.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSTask.scala
```diff
@@ -9,127 +9,80 @@
...
-    // Flush log buffer
-    if (shouldBufferLog) {
-      runner.loggerLock.synchronized {
-        logBuffer.foreach(_.call(loggers))
+    def log[T](level: Logger => (T => Unit))(log: LogElement[T]): Unit = {
+      if (shouldBufferLog) {
+        logBuffer += new ScalaJSTask.LogElement(log.index, level, log.x)
+      } else {
+        level(loggers(log.index))(log.x)
       }
```
gzm0: No it didn't. The diff is just a bit out of sync. It's below.

test-common/src/test/scala/org/scalajs/testcommon/RPCCoreTest.scala
gzm0: JUnit instantiates the test class separately for each test. It's a feature :)

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
sjrd: Except you're not using `keySet`, or I'm completely missing something.

Did you mean to use `callId <- pending.keySet.asScala` as the first generator? In that case I could definitely agree it's fine.

test-common/src/test/scala/org/scalajs/testcommon/SerializerTest.scala
gzm0: `enableScalastyleInSharedSources` FTW!!!!

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
gzm0: Yes, that's what I meant and what I have done - in my local working copy :P

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ComJSEnvRPC.scala
```diff
@@ -0,0 +1,67 @@
...
+          try {
+            handleMessage(runner.receive())
+          } catch {
+            case _: TimeoutException =>
+          }
+        }
+      } catch {
+        case _: ComJSEnv.ComClosedException       =>
+        case _: InterruptedException              =>
+        case _: Exception if synchronized(closed) =>
```
gzm0: Good catch. Needed to make `closed` volatile, because I actually can't synchronize here :) Another thread holding that monitor is waiting for this thread to terminate :)

gzm0:
Updated.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSRunner.scala
```diff
@@ -218,20 +181,19 @@ final class ScalaJSRunner private[testadapter] (
   private def createRemoteRunner(): Unit = {
     assert(master == null)
 
-    master = framework.libEnv.comRunner(masterLauncher)
-    master.start(framework.logger, framework.jsConsole)
+    val runner = framework.libEnv.comRunner(masterLauncher)
+    runner.start(framework.logger, ConsoleJSConsole)
```
gzm0: Good catch. Oversight from 1.0 -> 0.6 rebase.

gzm0:
Updated.

sjrd:
I assume we still need to fix the Selenium envs before merging this, is that right?

gzm0:
> I assume we still need to fix the Selenium envs before merging this, is that right?

Yes. I'm working on them.

gzm0:
There is a big problem with this PR: RPCs don't lend themselves very well to implement message passing. So this actually destroys the ordering guarantee we previously had for logging: Since each logger invocation is "fire-and-forget", there is no guarantee that the logger gets invoked in the right order.

With Selenium, this actually manifests: Logger messages come in the wrong order.
I suspect that messages queue up on the browser and when they are fetched, all invocations get dispatched to the logger at once. So in total we have 3 problems:

1. Logger invocations may be out of order.
1. Logger invocations may happen after `execute` is complete. This is not legal.
1. Slave/Master message channel may reorder messages.

(1) we can fix by not dispatching the actual RPC execution to the`ExecutionContext`. Not super clean, but would work. (2) we can fix by waiting on all (!) log calls. (3) is tricky, since we don't want to invoke `send` inside `receiveMessage` callback. So the only way I can currently see we can fix it is by allowing for "streaming RPCs" that you can open, repeatedly send a message (with ordering guarantee) and close. But that in turn would require an implementation of some (simple form of) reactive streams.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -0,0 +1,192 @@
...
+                s"Unknown opcode: $opCode"))
+            send(makeReply(callID, repl))
+          } { bep =>
+            val ep: bep.endpoint.type = bep.endpoint
+            import ep._
+
+            import scala.concurrent.ExecutionContext.Implicits.global
+
+            futureFromTry(Try(deserialize[ep.Req](in)))
+              .flatMap(bep.exec)
```
gzm0: This reorders messages.

sjrd: Ah yes, I see.

gzm0:
Updated. This fully works on chrome over selenium. Testing FF now.

gzm0:
Note: I'll squash the commits as soon as review is good. Just thought it is easier to review like this.

gzm0:
Works on FF. I could even reproduce #2751 :)

gzm0:
Just for good measure I also tested selenium (on Chrome) before these changes. Works too.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -133,13 +160,13 @@ private[scalajs] abstract class RPCCore {
...
-    result.map(makeMsg(ReplyOK, id, _)) match {
+    result.map(makeRpcMsg(ReplyOK, id, _)) match {
       case Success(m) => m
-      case Failure(t) => makeMsg(ReplyErr, id, t)
+      case Failure(t) => makeRpcMsg(ReplyErr, id, t)
     }
   }
 
-  private def makeMsg[T: Serializer](opCode: OpCode, id: Long,
+  private def makeRpcMsg[T: Serializer](opCode: OpCode, id: Long,
```
sjrd: We usually keep acronyms all-uppercase in camelCase identifiers, so this should be named `makeRPCMsg`.

gzm0:
> However, how does it ensure that all log messages have been received before the execute-is-finished message?

1. `RPCCore.send` guarantees immediate forwarding of the message to the underlying channel.
2. The channel guarantees ordering.
3. `RPCCore.attach(_: MsgEndpoint)` guarantees immediate forwarding of the message to the registered callback.

Therefore, if all calls to `send` are done inside the `execute` method (or its future continuation), the completion message must come afterwards.

gzm0:
Updated.

sjrd:
Ah I see. OK, that makes sense.

gzm0:
I might have spoken too soon. It seems that in the sbt-plugin-test, we actually try to send a message to the slave once it is completed already :-/

sjrd:
Hum, where does that happen?

test-interface/src/main/scala/org/scalajs/testinterface/internal/Slave.scala
```diff
@@ -5,119 +5,70 @@ import js.annotation._
...
-
-      body
-    } finally {
-      canSendRunnerMessage = false
-    }
-  }
+  JSRPC.attach(JSSlaveEndpoints.newRunner)(newRunner _)
+  JSRPC.attachAsync(JSSlaveEndpoints.execute)(execute _)
+  JSRPC.attach(JSSlaveEndpoints.stopSlave)(stopSlave _)
+  JSRPC.attach(JSSlaveEndpoints.msg)(runner.receiveMessage _)
```
gzm0: It "NPE"s here. `runner` is already set to `null`, but it still get's a message.

test-interface/src/main/scala/org/scalajs/testinterface/internal/Slave.scala
```diff
@@ -5,119 +5,70 @@ import js.annotation._
...
-    res
+    promise.future
   }
 
-  private def incomingRunnerMessage(msg: String): Try[Unit] = {
+  private def stopSlave(req: Unit): Unit = {
     ensureRunnerExists()
-    Try { runner.receiveMessage(msg); () }
+    try runner.done()
+    finally runner = null
```
gzm0: `runner` gets set to `null` here...

sjrd: Seems like either of two things hold here:
* either it is not valid to send a message to a slave after `done()` has been called on it, in which case sbt-plugin-test should not try to do that, or
* we should simply not set `runner` to `null` here, and instead just let it live on.

gzm0:
> either it is not valid to send a message to a slave after done() has been called on it, in which case sbt-plugin-test should not try to do that.

I think this is what we should do. But I don't see where sbt-plugin-test does that. It sends a message from the slave to the master but doesn't reply.

gzm0:
I have fixed one timing issue, where the slave did not get put early enough into the slave map, so the result message couldn't be routed. Let's see if this fixes it.

gzm0:
Updated. The problem was the following line:

```scala
JSRPC.attach(JSSlaveEndpoints.msg)(runner.receiveMessage _)
```

It bound to the current **value** of `runner`, not the the variable `runner`. I have indirected via a member method to ensure proper binding of `this`. Note that

```scala
JSRPC.attach(JSSlaveEndpoints.msg) { msg =>
  val r = runner
  r.receiveMessage(msg)
}
```

would not expose this behavior.

gzm0:
Even Scala 2.12.2 JVM exposes this behavior. This is a scalac bug: The [spec][1] clearly states that:
`math.sin _` is equivalent to `x => math.sin(x)`.

[1]: https://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html

sbt-plugin-test/testFramework/src/main/scala/sbttest/framework/DummyTask.scala
```diff
@@ -34,7 +36,8 @@ final class DummyTask(
 
   def execute(eventHandler: EventHandler, loggers: Array[Logger],
       continuation: Array[Task] => Unit): Unit = {
-    continuation(execute(eventHandler, loggers))
+    val tasks = execute(eventHandler, loggers)
+    runner.taskBlock.foreach(_ => continuation(tasks))
```
sjrd: Will other testing frameworks have to do that, now? Or is it specific to the Dummy framework?

test-adapter/src/main/scala/org/scalajs/sbttestadapter/RemoteException.scala
```diff
@@ -13,13 +13,13 @@ import sbt.testing._
 
 import org.scalajs.core.tools.json._
 
-import FingerprintSerializers._
-
+@deprecated("Unused. Is going to be removed.", "0.6.19.")
```
sjrd: This should now be 0.6.20.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/RemoteException.scala
```diff
@@ -13,13 +13,13 @@ import sbt.testing._
...
 final class RemoteException private (msg: String, _toString: String,
     cause: Throwable, val originalClass: String) extends Exception(msg, cause) {
   override def toString(): String = _toString
 }
 
+@deprecated("Unused. Is going to be removed.", "0.6.19.")
```
sjrd: Same here.

gzm0:
Filed as scala/bug#10452

sbt-plugin-test/testFramework/src/main/scala/sbttest/framework/DummyTask.scala
gzm0: This is specific to the dummy framework. It tests that if a framework relies on receiving messages from the master to run tests, it can wait on the messages.

sjrd: OK good :)

gzm0:
Updated.

PR #3043, opened 2017-07-06
A portion of the new linker API of #2998

### LINKED ISSUE #2998: New Linker API

sjrd:
This is part of #2998, though there is still a bunch of things to do. In particular, the `linker._` package still need to be cleaned up, with things renamed and/or moved to `standard._`. Nevertheless, it is a substantial step in the direction of the new API.


sjrd:
Rebased + new commit (at the top) that removes `RuntimeClassNameMapper.custom()`.

sjrd:
@gzm0 Any chance you could have a look at this PR, one of these days?

gzm0:
Yep, will do. I have a couple of comments pending that I'll release.

gzm0:
In case that wasn't clear, that's not everything :)

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/LinkerBackendPlatformExtensions.scala
```diff
@@ -0,0 +1,30 @@
...
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package org.scalajs.core.tools.linker.backend
+
+import org.scalajs.core.tools.linker.backend.closure.ClosureLinkerBackend
+
+object LinkerBackendPlatformExtensions {
```
sjrd: No, there's no need to. The `implicit def` that converts a `LinkerBackend.Config` into a `LinkerBackendPlatformExtensions.ConfigExt` is defined in the companion object of `LinkerBackend.Config`. So even though the target type is not in the import scope, the implicit conversion itself is in the implicit scope, so the implicit pimp works without import. The same mechanism is already applied for `StandardConfig.closureCompiler`.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -965,9 +965,11 @@ object Trees {
...
 
-  /** A hash of a tree (usually a MethodDef). Contains two SHA-1 hashes */
-  final class TreeHash(val treeHash: Array[Byte], val posHash: Array[Byte]) {
-    assert(treeHash.length == 20)
-    assert(posHash.length == 20)
+  /** A hash of a tree (usually a MethodDef).
+   *
+   *  Contains a SHA-1 hash.
+   */
+  final class TreeHash(val hash: Array[Byte]) {
```
sjrd: It can probably be `AnyVal`, indeed. I'll have to check whether it would be boxed more often than not, but I don't think so.

Inlining would make it more annoying to ensure invariants, not to mention being typesafe for methods like `hashesEqual` and `hashAsVersion`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
```diff
@@ -75,83 +70,12 @@ final class Linker(frontend: LinkerFrontend, backend: LinkerBackend)
...
-          optimizer = optimizer,
-          parallel = parallel,
-          closureCompilerIfAvailable = closureCompilerIfAvailable,
-          frontendConfig = frontendConfig,
-          backendConfig = backendConfig)
-    }
+object Linker {
+  def apply(frontendConfig: LinkerFrontend.Config,
+      backendConfig: LinkerBackend.Config): Linker = {
+    apply(LinkerFrontend(frontendConfig), LinkerBackend(backendConfig))
```
sjrd: Probably useless, indeed.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/frontend/LinkerFrontendPlatformExtensions.scala
```diff
@@ -0,0 +1,27 @@
...
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+
+package org.scalajs.core.tools.linker.frontend
+
+import org.scalajs.core.tools.linker.frontend.optimizer._
+
+private[frontend] object LinkerFrontendPlatformExtensions {
```
sjrd: Idk. Some possible alternatives:
* `LinkerFrontendPlatform`
* Just `Platform`, and throw everything platform-specific from the `frontend` package in it.

gzm0:
I don't understand why you want to remove `ESLevel`. Now the backend is leaking the `OutputMode` and the `ModuleKind` into the frontend. I think it is a valuable separation of concerns to hide things the frontend doesn't need to know from it.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Linker.scala
```diff
@@ -75,83 +70,12 @@ final class Linker(frontend: LinkerFrontend, backend: LinkerBackend)
...
-      new Config(
-          optimizer = true,
-          parallel = true,
-          closureCompilerIfAvailable = false,
-          frontendConfig = LinkerFrontend.Config(),
-          backendConfig = LinkerBackend.Config())
-    }
-  }
+  def apply(frontend: LinkerFrontend, backend: LinkerBackend): Linker =
+    new Linker(frontend, backend)
```
sjrd: Yes. The `class Linker` in this file is actually what will eventually become the class `standard.StandardLinker`, and this `apply` method will be moved to the object `standard.StandardLinker`. And the current `linker.GenLinker` will become `linker.Linker`.

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/frontend/LinkerFrontendPlatformExtensions.scala
sjrd: Yes, indeed. `Extensions` are ultimately part of the public API because they are inherited. Whereas this thing is an implementation detail that need never be seen.

sjrd:
> I don't understand why you want to remove ESLevel. Now the backend is leaking the OutputMode and the ModuleKind into the frontend. I think it is a valuable separation of concerns to hide things the frontend doesn't need to know from it.

The problem is that it is not completely clear that the frontend does not need to know about the output mode and module kind. For example, if we add an output mode that technically *assumes* ES2015+, but is optimized for speed rather than "for exploiting ES2015 features as much as possible", the frontend phases, and the optimizer in particular, could have to make different decisions in that mode versus `ECMAScript6`.

Similarly, the frontend may need to know things about the module kind. For example, we might want the Analyzer to report linking errors if an `@JSImport` class/object without `globalFallback` is reachable, but the module kind is `NoModule`.

I can reintroduce `ESLevel` under `standard._` if you want. But in the presence of `OutputMode`, it is redundant. Its only use case would be to write custom phases (in plugins) that want to take decisions only depending on the assumed ES version, without being affected by the addition or removal of `OutputMode`s in the future. But for that use case `ESLevel` should have an `Ordering`, so that I can take decisions based on "Is it ES2015 *or later*"?

gzm0:
Ok, fair enough, I buy this. We can introduce `ESLevel` again if plugins actually start proliferating.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
sjrd: So, actually, `TreeHash`es are always inside an `Option`, so making them `AnyVal`s would not improve anything. On the other hand, it would slightly degrade our ability to evolve them in binary compatible ways. So I'd say we keep them as is.

Does that make sense?

sjrd:
Updated with:
* Removal of `Linker.apply(frontendConfig, backendConfig)`
* `LinkerFrontendPlatform` and `LinkerBackendPlatform`

and also changed the PR description so that it won't close automatically the issue. (there was no *commit* with such a message)

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: SG.

PR #3045, opened 2017-07-09
Clean up the IR

sjrd:
I still have some changes in store for the future, but the last two commits definitely deserve a "review milestone", as in they should be reviewed and merged before further changes.


ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -55,15 +62,22 @@ class PrintersTest {
 
     assertPrintEquals("O", ClassType(ObjectClass))
 
-    assertPrintEquals("O[]", ArrayType(ObjectClass, 1))
-    assertPrintEquals("I[][]", ArrayType("I", 2))
+    assertPrintEquals("O[]", ArrayType(ArrayTypeRef(ObjectClass, 1)))
+    assertPrintEquals("I[][]", ArrayType(ArrayTypeRef("I", 2)))
```
gzm0: Use the helper you created?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2029,7 +2029,8 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
 
         case NewArray(tpe, lengths) =>
           genCallHelper("newArrayObject",
-              genClassDataOf(tpe), js.ArrayConstr(lengths map transformExpr))
+              genClassDataOf(tpe.arrayTypeRef),
+              js.ArrayConstr(lengths map transformExpr))
```
gzm0: `lengths.map(transformExpr)`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -514,15 +514,20 @@ private[optimizer] abstract class OptimizerCore(
...
         transformExpr(BinaryOp(BinaryOp.!==, expr, Null()))
 
-      case IsInstanceOf(expr, tpe) =>
+      case IsInstanceOf(expr, typeRef) =>
         trampoline {
           pretransformExpr(expr) { texpr =>
             val result = {
-              // TODO This cast is suspicious
-              if (isSubtype(texpr.tpe.base, tpe.asInstanceOf[Type])) {
+              // TODO The hard conversion of ClassRef into ClassType is suspicious
```
gzm0: I don't think this is suspicious looking at it now. If you look at how its used, it's simply a workaround to not implement `isSubRefType`. Completely acceptable if we document it (NB that `typeRefTpe` cannot be a JS type by construction, does the IR checker verify this?).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -875,17 +879,22 @@ private[optimizer] abstract class OptimizerCore(
...
         pretransformExpr(expr) { texpr =>
-          tpe match {
-            case ClassType(ObjectClass) =>
+          typeRef match {
+            case ClassRef(ObjectClass) =>
               cont(texpr)
             case _ =>
-              // TODO This cast is suspicious
-              if (isSubtype(texpr.tpe.base, tpe.asInstanceOf[Type])) {
+              // TODO The hard conversion of ClassRef into ClassType is suspicious
```
gzm0: Same comment here about subtypeing. Maybe we should lift this into a helper `isSubRefType` or something and document stuff there.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -174,17 +178,19 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
 
         (tree, export.jsName, pos)
       }).toList
     }
 
-    private def genTopLevelOrStaticMethodExports(classSym: Symbol,
-        destination: ExportDestination): List[(js.Tree, String, Position)] = {
+    private def genTopLevelOrStaticMethodExports[A <: js.IRNode: ClassTag](
+        classSym: Symbol,
+        destination: ExportDestination): List[(A, String, Position)] = {
```
gzm0: Just to follow up the conversation we had over the weekend: I think it is completely OK to "leak" the implementation detail of which tree we use for a given `ExportDestination`. IMO this is still better than having these strange cast incantations requiring `ClassTag` and what not.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -667,7 +671,7 @@ abstract class GenJSCode extends plugins.PluginComponent
             case lit: js.StringLiteral    => js.JSBracketSelect(selfRef, lit)
             case js.ComputedName(tree, _) => js.JSBracketSelect(selfRef, tree)
           }
-          js.Assign(select, jstpe.zeroOf(fdef.tpe))
+          js.Assign(select, jstpe.zeroOf(fdef.ftpe))
```
sjrd: This will naturally disappear once we rebase on top of the changes from the 0.6.x branch, yes. I can't remove it right now before we rebase, because `.tpe` simply does not typecheck.

gzm0: Ah :P

ir/src/main/scala/org/scalajs/core/ir/Types.scala
```diff
@@ -15,16 +15,16 @@ import Trees._
...
 object Types {
 
-  /** Type of an term (expression or statement) in the IR.
+  /** Type of a term (expression or statement) in the IR.
    *
-   *  There is a many-to-one relationship from [[ReferenceType]]s to types,
+   *  There is a many-to-one relationship from [[TypeRef]]s to `Type`s,
    *  because:
    *
    *  - `scala.Byte`, `scala.Short` and `scala.Int` collapse to [[IntType]]
```
sjrd: `AsInstanceOf(_, ClassType("scala.Short"))` is not a valid tree. It would have to be `Unbox(_, 'I') << 16 >> 16`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: But problem is that not all `ClassRef`s have a corresponding `ClassType`. Things that don't:
* `ClassRef`s of primitive types
* `ClassRef`s of JS types

However now I think this I understand why this is valid here: because for all `ClassRef`s *that are admissible in an `Is/AsInstanceOf`*, their erased `Type` is indeed a `ClassType` (except `j.l.Object`, special-cased above).

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -15,35 +15,43 @@ import Position.NoPosition
...
      */
     def encodedName: String
-
-    def pos: Position
   }
 
   case class Ident(name: String, originalName: Option[String])(
-      implicit val pos: Position) extends PropertyName {
+      implicit val pos: Position)
+      extends IRNode with PropertyName {
```
sjrd: You mean because `PropertyName` already extends `IRNode` on its own? It feels more explicit to me to specify `extends IRNode` here nevertheless. Same for `ComputedName`.

ir/src/test/scala/org/scalajs/core/ir/PrintersTest.scala
```diff
@@ -918,22 +918,26 @@ class PrintersTest {
...
         """
-          |class LTest extends O {
+          |module class LTest extends O {
           |  val x$1: int
           |  var y$1: int
+          |  export top module "pkg.Foo"
           |}
         """,
-        ClassDef("LTest", ClassKind.Class, Some(ObjectClass), Nil, None,
+        ClassDef("LTest", ClassKind.ModuleClass, Some(ObjectClass), Nil, None,
```
sjrd: Yes, because an `export top module` does not make sense in a non-module class. The IR would still correctly print, of course, but it would look weird.

sjrd:
Updated.

For the casts in `genTopLevelOrStaticExports`, a type member in `ExportDestination` was not helping, because there was no way to reconcile the if/else logic, depending on `destination` and producing different kinds of IR nodes, with the actual value of the type member in `destination`. The Scala type system is simply not powerful enough for that (it would requires some kind of occurrence typing). Instead I had to use a typeclass-like mechanism, except it's explicit rather than implicit, where the typeclass also has methods to deal with the if/else logic in an object-oriented way.

The result is, IMO, worse than before. I have put the changes in a separate commit so that you can see what that gives. We should either squash that commit or throw it away. I'm in favor of throwing it away, TBH.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
gzm0: Ok.

gzm0:
> I'm in favor of throwing it away, TBH.

Yes, I agree. I thought Scala handles this better:

```scala
trait A { type X }
object A1 extends A { type X = Int }
object A2 extends A { type X = String}

def foo(x: Int, y: A): y.X = y match {
  case A1 => x: y.X
  case A2 => x.toString: y.X
}
```

But it doesn't :)

sjrd:
Rebased, and dropped the commit with the typeclass-ish thing for `genTopLevelOrStaticExports`.

PR #3048, opened 2017-07-10
Fix #3047: Refuse to compile overloads collapsing through @JSName.

### LINKED ISSUE #3047: Two methods with the same JS name must not compile
Duh ... I have no idea how this went unnoticed:
```scala
@ScalaJSDefined
class Foo extends js.Object {
  @JSName("bar")
  def foobar(): Int = 24

  def bar(): Int = 42
}
```
Compiles without error. The IR is:
```scala
js class Lhelloworld_Foo extends sjs_js_Object {
  def "constructor"() {
    super()
  }
  def "bar"(): any = {
    Lhelloworld_Foo::bar__Lhelloworld_Foo__I(this)
  }
  static def foobar__Lhelloworld_Foo__I($this: any): int = {
    24
  }
  static def bar__Lhelloworld_Foo__I($this: any): int = {
    42
  }
}
```
The `foobar` method was simply disregarded when generating the dispatching code for `"bar"`. If we switch the order of the definitions, then `bar` is disregarded, i.e., the last one is always kept.

This is because of the test at https://github.com/scala-js/scala-js/blob/b54a26c1c2126bedb5b6dd8a450a5e6891e06612/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala#L332
It seems this test is utterly useless for correct code, and actively harmful for incorrect code such as the one above. git blame says that this test has been there since the original implementation of `@ScalaJSDefined`. It was probably needed before I figured I needed `!sym.isBridge` anyway.

Dropping the mentioned test fixes the issue, and the existing test suite still passes.

sjrd:
This is potentially a source breaking change, as it may be that a codebase happened to compile to something that worked. However, it is binary compatible.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -323,14 +323,8 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     }
 
     private def genJSClassDispatcher(classSym: Symbol, name: JSName): js.Tree = {
-      var alts: List[Symbol] = Nil
-      for {
-        sym <- classSym.info.members
-        if !sym.isBridge && jsNameOf(sym) == name
-      } {
-        val tpe = sym.tpe
-        if (!alts.exists(alt => tpe.matches(alt.tpe)))
-          alts ::= sym
+      val alts = classSym.info.members.toList.filter { sym =>
+        !sym.isBridge && jsNameOf(sym) == name
       }
 
       assert(!alts.isEmpty,
@@ -559,12 +553,9 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         // 2. The optional argument count restriction has triggered
         // 3. We only have (more than once) repeated parameters left
         // Therefore, we should fail
-        reporter.error(pos,
-            s"""Cannot disambiguate overloads for exported method ${alts.head.name} with types
-               |  ${alts.map(_.typeInfo).mkString("\n  ")}""".stripMargin)
+        reportCannotDisambiguateError(alts)
         js.Undefined()
       } else {
-
         val altsByTypeTest = groupByWithoutHashCode(alts) {
           case ExportedSymbol(alt) =>
             typeTestForTpe(computeExportArgType(alt, paramIndex))
@@ -625,6 +616,29 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
           }
         }
       }
+    }
+
+    private def reportCannotDisambiguateError(alts: List[Exported]): Unit = {
+      val currentClass = currentClassSym.get
+
+      // Find a position that is in the current class for decent error reporting
+      val pos = alts.collectFirst {
+        case ExportedSymbol(sym) if sym.owner == currentClass => sym.pos
+        case alt: ExportedBody                                => alt.pos
+      }.getOrElse {
+        currentClass.pos
+      }
+
+      val kind =
+        if (isScalaJSDefinedJSClass(currentClass)) "method"
+        else "exported method"
+
+      val name = alts.head.name
+      val altsTypesInfo = alts.map(_.typeInfo).mkString("\n  ")
+
+      reporter.error(pos,
+          s"Cannot disambiguate overloads for $kind $name with types\n" +
+          s"  $altsTypesInfo")
     }
 
     private def computeExportArgType(alt: Symbol, paramIndex: Int): Type = {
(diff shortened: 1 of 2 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -323,14 +323,8 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
-      var alts: List[Symbol] = Nil
-      for {
-        sym <- classSym.info.members
-        if !sym.isBridge && jsNameOf(sym) == name
-      } {
-        val tpe = sym.tpe
-        if (!alts.exists(alt => tpe.matches(alt.tpe)))
-          alts ::= sym
+      val alts = classSym.info.members.toList.filter { sym =>
+        !sym.isBridge && jsNameOf(sym) == name
```
gzm0: Doesn't the style guide say use parens over braces for lambdas?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -627,6 +618,29 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+    private def reportCannotDisambiguateError(alts: List[Exported]): Unit = {
+      val currentClass = currentClassSym.get
+
+      // Find a position that is in the current class for decent error reporting
+      val pos = alts.collectFirst {
+        case ExportedSymbol(sym) if sym.owner == currentClass => sym.pos
+        case alt: ExportedBody                                => alt.pos
+      }.getOrElse {
+        currentClass.pos
+      }
```
gzm0: Why is the `collectFirst` here necessary? Don't we have: `require(alts.nonEmpty)`? This gives the impression that there is an `Exported` which is none of the two cases.

Consider replacing this with:

```scala
val pos = alts.head match {
  // same
}
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
sjrd: Only if the lambda is a one-liner, from its `param =>` up to the end of its body.

So this is forbidden:
```scala
foo { a => a + 1}
```
but this is OK:
```scala
foo { a =>
  a + 1
}
```

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
sjrd: See the comment above the `collectFirst`? We're trying to find one of the alternatives that is actually defined in the current class to get its position. If there is no such alternative (because they are all inherited), we need to fall back on the class' position, otherwise we report an error potentially in a different file, which might not even be part of the current compilation run (bad).

sjrd: Said in other words: the guard `if sym.owner == currentClass` could be `false` for all `alts`.

gzm0: Right :) In that case, can you add a test for the `getOrElse` case? Seems like we don't have that right now.

sjrd: So, er ... actually I cannot come up with any source program that would enter the `getOrElse`. The reason is that `genJSClassDispatchers` is only called for `JSName`s that were found in the current class, on concrete methods. So by construction there will be at least one `alt` coming from the current class.

That said, removing the `getOrElse` means I would need a brutal `.get` or a `throw` in the `getOrElse`. It doesn't feel right to do either, when there is a perfectly reasonable thing to do in `getOrElse`. It is also future-proof and "I missed some corner case"-proof.

WDYT? Should I add a comment in there telling that (we think) there is no source code that can trigger this code?

gzm0: Hmmm... What about something like this:

```scala
trait Foo extends js.Object {
  def foo: js.UndefOr[Int] = js.undefind
}

trait Bar extends js.Object {
  @JSName("foo")
  def bar: js.UndefOr[Int] = js.undefined
}

class A extends Foo with Bar
```

Shouldn't this complain (but probably wont at the moment)?

gzm0: Hmmm... Probably not, since it doesn't really generate any real methods. So a trait never actually defines any method.

TBH, I actually would prefer the `.get` in the spirit of putting an assert on a condition we believe to be true. It is essentially the equivalent of `assert alts.exists(_.sym.owner == currentClass)`.

Alternatively we could put the assert and still fallback, so a user that gets hit by this could disable the asserts (although we don't have a precedent for this actually happening).

sjrd: > Hmmm... Probably not, since it doesn't really generate any real methods. So a trait never actually defines any method.

Indeed. The above code is valid.

> Alternatively we could put the assert and still fallback, so a user that gets hit by this could disable the asserts (although we don't have a precedent for this actually happening).

It's not possible to disable a Scala `assert`. Not after translation to bytecode, anyway. This is different from Java asserts, which have run-time support in the VM.

sjrd: I found one:
```scala
@js.native
@JSGlobal
class Foo extends js.Object {
  def foo(x: Int): Int = js.native

  @JSName("foo")
}

@ScalaJSDefined
class Bar extends Foo {
  def foo(): Int = 42
}
```
results in
```
[error] /localhome/doeraene/projects/scalajs/examples/helloworld/HelloWorld.scala:21: Cannot disambiguate overloads for method bar with types
[error]   (x: Int)Int
[error]   (x: Int)Int
[error] class Bar extends Foo {
[error]       ^
[error] one error found
```
thanks to the code in the `getOrElse`.

I'll add that as a test.

sjrd:
Updated.

gzm0:
Nice! :)

PR #3049, opened 2017-07-10
Use a fixed structure rather than functions for runtime class name.

sjrd:
Arbitrary functions cannot be compared for equality, which means that currently, there is no way to correctly determine whether two instances of `Semantics` are equivalent. This is an issue for caches, which cannot invalidate their outputs if the input `Semantics` have changed.

Moreover, the existing `Semantics.runtimeClassName` field was a function from `LinkedClass` to `String`, but `LinkedClass`es are a concept of `linker.standard._` whereas `Semantics` is in `linker._`.

We replace the unconstrained `LinkedClass => String` function by an ADT `RuntimeClassNameMapper`. Ultimately, it will only have 3 possibilities, plus an `andThen` combinator:

* Keep all class names unchanged
* Discard all class names and replace everything by `""`
* Replace a prefix by another prefix (typically, both would end with `.` to match packages).

In 0.6.x, however, we have a "custom" combinator to keep supporting the deprecated arbitrary `LinkedClass => String`.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -627,6 +627,12 @@ object Build {
       base = file("tools/js"),
       settings = myScalaJSSettings ++ commonToolsSettings ++ Seq(
           crossVersion := ScalaJSCrossVersion.binary,
+
+          /* We need RuntimeClassNameMapper.custom() in QuickLinker
+           * TODO Remove this in 1.x.
+           */
+          scalacOptions in Test -= "-Xfatal-warnings",
+
           resourceGenerators in Test += Def.task {
             val base = (resourceManaged in Compile).value
             IO.createDirectory(base)
@@ -1498,12 +1504,26 @@ object Build {
         jsDependencies += ProvidedJS / "ScalaJSDefinedTestNatives.js" % "test",
         skip in packageJSDependencies in Test := false,
 
-        scalaJSSemantics ~= (_.withRuntimeClassName(_.fullName match {
-          case "org.scalajs.testsuite.compiler.ReflectionTest$RenamedTestClass" =>
-            "renamed.test.Class"
-          case fullName =>
-            fullName
-        })),
+        scalaJSSemantics ~= { sems =>
+          import Semantics.RuntimeClassNameMapper
+
+          sems.withRuntimeClassNameMapper(
+              RuntimeClassNameMapper.custom(_.fullName match {
+                case "org.scalajs.testsuite.compiler.ReflectionTest$RenamedTestClass" =>
+                  "renamed.test.Class"
+                case fullName =>
+                  fullName
+              }).andThen(
+                  RuntimeClassNameMapper.regexReplace(
+                      raw"""^org\.scalajs\.testsuite\.compiler\.ReflectionTest\$$Prefix""".r,
+                      "renamed.test.byprefix.")
+              ).andThen(
+                  RuntimeClassNameMapper.regexReplace(
+                      raw"""^org\.scalajs\.testsuite\.compiler\.ReflectionTest\$$OtherPrefix""".r,
+                      "renamed.test.byotherprefix.")
+              )
+          )
+        },
 
         javaOptions in Test += "-Dscalajs.scalaVersion=" + scalaVersion.value,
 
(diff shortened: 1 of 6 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
```diff
@@ -107,21 +119,98 @@ final class Semantics private (
...
+
+      override def hashCode(): Int = 369581025 // generated at random
+    }
+
+    def keepAll(): RuntimeClassNameMapper = KeepAll
+
+    def discardAll(): RuntimeClassNameMapper = DiscardAll
+
+    def replacePrefix(from: String, to: String): RuntimeClassNameMapper =
+      ReplacePrefix(from, to)
```
gzm0: How did you come up with these things? I'm just wondering if, for example, discarding everything with a prefix is a use-case as well.

project/Build.scala
```diff
@@ -627,6 +627,10 @@ object Build {
       base = file("tools/js"),
       settings = myScalaJSSettings ++ commonToolsSettings ++ Seq(
           crossVersion := ScalaJSCrossVersion.binary,
+
+          // We need RuntimeClassNameMapper.custom() in QuickLinker
+          scalacOptions in Test -= "-Xfatal-warnings",
```
gzm0: Can you add a TODO for 1.0 to nuke this?

tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
sjrd: There are probably zillions of use cases. We'll probably only find out about them once people ask us for some more.

`keepAll()` and `discardAll()` are obvious. `replacePrefix()` I always thought was the typical use case, but now that I really think about it, I'm not sure it even *has* a use case :-s Maybe we should just remove that one for now.

It's difficult to test anything without at least one non-trivial mapper, though. Probably your discard everything that has a given prefix is more useful.

Or we could have something regex-replace-based. Basically something that invokes [`java.util.regex.Matcher.replaceAll`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#replaceAll(java.lang.String)) for a given pattern and replacement string.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
```diff
@@ -107,21 +119,124 @@ final class Semantics private (
...
+     *
+     *  Given an input class name `className`, the mapper will return a new
+     *  class name equivalent to
+     *  {{{
+     *  regex.replaceAllIn(className, replacement)
+     *  }}}
+     */
+    def regexReplace(regex: scala.util.matching.Regex,
+        replacement: String): RuntimeClassNameMapper = {
+      regexReplace(regex.pattern, replacement)
```
gzm0: trailing parens to `pattern`?

tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
```diff
@@ -107,21 +119,124 @@ final class Semantics private (
...
 }
 
 object Semantics {
   private val HashSeed =
     scala.util.hashing.MurmurHash3.stringHash(classOf[Semantics].getName)
 
+  sealed abstract class RuntimeClassNameMapper {
+    import RuntimeClassNameMapper._
+
+    def andThen(that: RuntimeClassNameMapper): RuntimeClassNameMapper = {
```
gzm0: I guess this is fine for now, but at some point people might want the ability to "abort" the matching chain (as in: do these replacements, then drop the rest). If this is requested, we can add a `orElse` method I assume (and adjust internals).

tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
sjrd: This one is a `val` in a Scala class.

gzm0: Ah, right :)

PR #3059, opened 2017-07-14
Fix #3058: Use the proper type for the "thisLocalVar" of tailrec.

### LINKED ISSUE #3058: Wrong IR type for `this` local of @tailrec methods in the presence of self type in 2.12
With a combination of:

* Scala 2.12
* An `@tailrec` method
* In a class or trait that has a self-type

the codegen produces code with invalid IR.

Reproduction (`Parent` can be a `class` or a `trait`):
```scala
trait Parent { this: Child =>
  @tailrec final def bar(i: Int, acc: Int): Int = {
    println(s"bar($i, $acc)")
    if (i <= count)
      bar(i + 1, acc + i)
    else
      acc
  }
}

class Child extends Parent {
  def count: Int = 5
}

  def main(): Unit = {
    val obj = new Child
    println(obj.bar(1, 0))
  }
}
```
gives the IR error:
```
[error] HelloWorld.scala(13:7:VarRef): Variable _$this of type ClassType(Lhelloworld_Parent) typed as ClassType(Lhelloworld_Child)
```
The tree that we receive from scalac is properly typed with `_$this` of type `Child` (the self-type), but we erroneously assume `currentClassType` for the type of `_$this` in some places (in the `VarDef` and in `genThis()`), rather than using the type of the symbol.

sjrd:
In Scala 2.12, if a tail-recursive method is in a class or trait with a self-type, the type of the `_$this` local variable is the self type. We previously assumed it was always the current class type, which produced invalid IR in that case.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -136,7 +136,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     // Per method body
     private val currentMethodSym = new ScopedVar[Symbol]
-    private val thisLocalVarIdent = new ScopedVar[Option[js.Ident]]
+    private val thisLocalVarInfo = new ScopedVar[Option[(js.Ident, jstpe.Type)]]
     private val fakeTailJumpParamRepl = new ScopedVar[(Symbol, Symbol)]
     private val enclosingLabelDefParams = new ScopedVar[Map[Symbol, List[Symbol]]]
     private val isModuleInitialized = new ScopedVar[VarBox[Boolean]]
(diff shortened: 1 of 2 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1959,14 +1962,15 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-      thisLocalVarIdent.fold[js.Tree] {
+      thisLocalVarInfo.fold[js.Tree] {
         if (tryingToGenMethodAsJSFunction) {
           throw new CancelGenMethodAsJSFunction(
               "Trying to generate `this` inside the body")
         }
         js.This()(currentClassType)
-      } { thisLocalIdent =>
-        js.VarRef(thisLocalIdent)(currentClassType)
+      } { thisLocalInfo =>
```
gzm0: `case (thisLocalIdent, thisLocalTpe) =>`?

sjrd: I hesitated. Done.

PR #3061, opened 2017-07-14
Fix #3060: Do not check field existence for classes with no instances.

### LINKED ISSUE #3060: The IR checker must not check field existence on classes without instances
The following snippet reports a spurious IR checking error:
```scala
class Outer(f: Int) {
  class Inner {
    def get(): Int = f
  }
}

  def main() {
    println(classOf[Outer])
    test(null)
  }

  def test(outer: Outer): Unit = {
    if (outer == null) {
      println("null")
    } else {
      val inner = new outer.Inner
      println(inner.get())
    }
  }
}
```
The error is:
```
[error] HelloWorld.scala(13:22:Select): Class Lhelloworld_Outer does not have a field helloworld$Outer$$f$f
```
In fact `Outer` *does* have such a field, so how come the IR checker got confused?

The reason is that `Outer` is *needed at all* (because of the `classOf[Outer]`) but *does not have any instances*. It is therefore trimmed down by the `BaseLinker` to an empty `class` definition, which does not have any field. The IR checker then complains about the access to that field from `Inner`, which *does* have instances (at least as far as our reachability analysis is concerned) and hence whose `get()` method is reachable.

The fix is easy: the IR checker must simply ignore checking field existence in classes that have no instances.

sjrd:
If a class has no instances, the `BaseLinker` will trim it down and remove all its fields. However, it is still possible that there is some reachable code that reads or writes those fields (although such code would inevitably crash with an NPE if actually executed). The IR checker must not complain in such cases.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
@@ -760,13 +760,32 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
             if (!kind.isClass) {
               reportError(s"Cannot select $item of non-class $cls")
             } else {
-              maybeClass.foreach {
-                _.lookupField(item).fold[Unit] {
-                  reportError(s"Class $cls does not have a field $item")
-                } { fieldDef =>
-                  if (fieldDef.tpe != tree.tpe)
-                    reportError(s"Select $cls.$item of type "+
-                      s"${fieldDef.tpe} typed as ${tree.tpe}")
+              /* Actually checking the field is done only if
+               *
+               * * the class exists, and
+               * * it has instances (including instances of subclasses).
+               *
+               * The latter test is necessary because the BaseLinker can
+               * completely get rid of all the fields of a class that has no
+               * instance. Obviously in such cases, the only value that
+               * `qualifier` can assume is `null`, and the `Select` will fail
+               * with an NPE. But the IR is still valid per se.
+               *
+               * The former case is necessary too because, if the class is not
+               * even "needed at all", the BaseLinker will simply remove the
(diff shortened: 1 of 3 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -760,13 +760,15 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
             } else {
-              maybeClass.foreach {
-                _.lookupField(item).fold[Unit] {
-                  reportError(s"Class $cls does not have a field $item")
-                } { fieldDef =>
-                  if (fieldDef.tpe != tree.tpe)
-                    reportError(s"Select $cls.$item of type "+
-                      s"${fieldDef.tpe} typed as ${tree.tpe}")
+              maybeClass.foreach { c =>
+                if (c.hasInstances) { // see #3060
```
gzm0: Consider moving the reason why we do this inline here. IMO a better position than somewhere in an issue history.

This also makes me thing: Do we need to support a "full check mode" where essentially the `BaseLinker` doesn't remove stuff (or at least does not tear classes apart)? If I were to work on an IR patcher, I would be terribly annoyed by behavior like this.

sjrd: > Consider moving the reason why we do this inline here. IMO a better position than somewhere in an issue history.

Done.

> Do we need to support a "full check mode"?

It might be nice, but it will be hard to define what "full" really means. In general, our IR checker can only check IR that *links* to begin with. So in any case, it is necessary for DCE in general to kick in before the IR checker. Then the problem is that we won't be able to check *everything* in all cases. So where do we draw the line about what should be checked in "full" check mode?

PR #3063, opened 2017-07-15
Fix #3055: Correct codegen for super in anonymous JS classes.

### LINKED ISSUE #3055: Super call in anon JS class produces invalid IR
Extracted from #3053.

Snippet:
```scala
@ScalaJSDefined
class Foo extends js.Object {
  def bar(msg: String): Unit = println(msg)
}

  def main(): Unit = {
    val foo = new Foo {
      override def bar(msg: String): Unit = super.bar("foo: " + msg)
    }
    foo.bar("foobar")
  }
}
```
IR checker error:
```
[error] HelloWorld.scala(20:54:JSSuperBracketCall): JS class type expected but ClassType(Lhelloworld_HelloWorld$$anon$1) found
```

sjrd:
In a named JS class, super method calls (resp. super property accesses) are translated to `JSSuperBracketCall` (resp. `JSSuperBracketSelect`). These IR nodes receive as first parameter a reference to enclosing class, so that they can tell in which prototype they have to start the lookup.

This translation is not valid in anonymous JS classes, because the enclosing class becomes an `AbstractJSType` rather than a `JSClass`. Instead, we now eagerly desugar the nodes in the compiler back-end. This strategy is consistent with how anonymous JS classes are otherwise translated: they are purely a compiler back-end concept, and require no support from the IR.

To translate super property accesses (reads and writes), we need run-time helpers, which we add in `scala.scalajs.runtime`. Their code is unfortunately duplicated with similar helpers used by the emitter to translate `JSSuperBracketSelect`s.


sjrd:
This is the version where we do everything in the compiler back-end. The good thing is: it's extremely easy, with a very small diff. I think that is a huge point in favor of this solution, rather than the one involving changes in the IR.

PR #3068, opened 2017-07-20
Fix #3065: Cross-compile the sbt plugin for sbt 1.0.0-RC2.

### LINKED ISSUE #3065: Build and publish the sbt plugin for sbt 1.0.0-RC2
http://www.scala-sbt.org/1.0/docs/sbt-1.0-Release-Notes.html#sbt+1.0.0-RC2

It turns out the support for cross-building an sbt plugin for 1.0.0-RC2 from a build written in 0.13.16-M1+ is quite good. So we'll be able to publish the sbt plugin of 0.6.x for sbt 1.x, in parallel to sbt 0.13.x.

sjrd:


dwijnand:
```
tmp.sh: 35: tmp.sh: ir/publishLocal: not found
```

wat

ci/matrix.xml
```diff
@@ -277,6 +277,29 @@
         'set scalaJSUseRhino in Global := true' test
   ]]></task>
 
+  <task id="sbtplugin-test-sbt1"><![CDATA[
+    # Publish Scala.js artifacts locally
+    # Then go into standalone project and test
+    sbt ++2.11.11 compiler/publishLocal library/publishLocal javalibEx/publishLocal \
+                  testInterface/publishLocal stubs/publishLocal \
+                  jUnitPlugin/publishLocal jUnitRuntime/publishLocal &&
+    sbt ++2.12.2 ^^1.0.0-RC2
```
sjrd: @dwijnand I forgot a `\` here ;)

sjrd:
Consistent error on the CI server when launching sbt 1.0.0-RC2 in `sbt-plugin-test`:
```
Downloading sbt launcher for 1.0.0-RC2:
  From  http://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/1.0.0-RC2/sbt-launch.jar
    To  /home/jenkins/.sbt/launchers/1.0.0-RC2/sbt-launch.jar

Download failed. Obtain the jar manually and place it at /home/jenkins/.sbt/launchers/1.0.0-RC2/sbt-launch.jar
```
@dwijnand Does that ring any bell?

smarter:
@sjrd You may need to update the `sbt` launcher on that machine? sbt is now published on maven and not bintray (http://repo.typesafe.com/typesafe/ivy-releases is an alias of bintray): https://repo1.maven.org/maven2/org/scala-sbt/sbt-launch/1.0.0-RC2/

sjrd:
Ah OK, thanks. I'll try that tomorrow.

dwijnand:
Yeah update the sbt-extras you have on that box. I changed it about a year ago to download from http://repo.scala-sbt.org/scalasbt/maven-releases for sbt.version 1+.

But Guillaume's right, it appears to be also on Maven Central maybe I should switch to that.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1732,8 +1732,10 @@ private[optimizer] abstract class OptimizerCore(
   private def transformExprsOrSpreads(trees: List[Tree])(
       implicit scope: Scope): List[Tree] = {
 
-    trees match {
-      case (spread: JSSpread) :: rest =>
+    val builder = List.newBuilder[Tree]
+
+    trees.foreach {
```
gzm0: Is this not just a `flatMap`?

ci/matrix.xml
```diff
@@ -277,6 +278,30 @@
         'set scalaJSUseRhino in Global := true' test
   ]]></task>
 
+  <task id="sbtplugin-test-sbt1"><![CDATA[
```
gzm0: Any reason you are not generalizing the `sbtplugin-test` task? The CI script supports any variable. Just use `$sbt` and set it with a `<v>` tag.

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala
```diff
@@ -0,0 +1,68 @@
...
+      part: String): CrossVersion = {
+    cross match {
+      case CrossVersion.Disabled =>
+        CrossVersion.constant(part)
+      case cross: sbt.librarymanagement.Constant =>
+        cross.withValue(part + "_" + cross.value)
+      case cross: CrossVersion.Binary =>
+        cross.withPrefix(part + "_" + cross.prefix)
+      case cross: CrossVersion.Full =>
+        cross.withPrefix(part + "_" + cross.prefix)
```
gzm0: I do not understand this, but the fact that `Binary` and `Full` does the same thing seems suspicious.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.15
+sbt.version=0.13.16-M1
```
sjrd: @dwijnand What are the stability guarantees of 0.13.16-M1? Why was that version not published as 0.13.16?

When can we expect 0.13.16 final? Apparently we're blocking the entire Scala eco-system from even trying out sbt 1.0.0-RC2, so ...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: It is, but `flatMap` will create a lot of garbage lists in this case, since the `JSSpread` case should be fairly rare.

In general we don't have any `flatMap` on collections in `OptimizerCore.scala` (we have one `flatMap`s on `Option`s).

ci/matrix.xml
sjrd: Mostly the `^^` thingy, which I would like to avoid at all in the 0.13.x variant. Also the `echo` is annoying, because if we do it in the 0.13.x job too, then nothing is testing what's stored *in the file* `sbt-plugin-test/project/build.properties` in the repo.

Given my level in bash-fu, I found it easier to copy-paste. But if you have a concrete suggestion for the two above issues, that'd be nice.

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala
sjrd: They syntactically do the same thing. But in the `case Binary`, `cross` is a `Binary` so `cross.withPrefix` constructs a `Binary`. Whereas in `case Full`, `cross.withPrefix` constructs a `Full`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Ok. Can you add a comment?

project/build.properties
dwijnand: > What are the stability guarantees of 0.13.16-M1?

What is meant here by stability?

> Why was that version not published as 0.13.16?

So we could start using the cross-building functionality, without a full release.

> When can we expect 0.13.16 final?

I'll have a chat with @eed3si9n today. We need to get a version out that includes https://github.com/sbt/sbt/pull/3329 (to avoid everyone having to include the workaround), but Eugene said he had an interest in trying to get https://github.com/sbt/sbt/issues/3205 into 0.13.16 too.

Notes:

+ https://github.com/sbt/sbt/compare/v0.13.15...v0.13.16-M1
+ https://github.com/sbt/sbt/compare/v0.13.16-M1...0.13

ci/matrix.xml
gzm0: Use:

```shell
${sbt_version_override:+^^$sbt_version_override}
```

And:

```shell
if [ -n "$sbt_version_override" ]; then echo ...; fi && 
```

project/build.properties
sjrd: > What is meant here by stability?

Most importantly, is whether 0.13.16 final is going to be source and--above all--binary backward compatible with 0.13.16-M1? Or is only backward compatible with 0.13.15, and could potentially break backward compat with 0.13.16-M1?

If we're going to publish an sbt plugin of Scala.js built on top of 0.13.16-M1, we need to make sure it will stay usable by builds using 0.13.16 final and later.

> So we could start using the cross-building functionality, without a full release.

I guess this joins the question above, but: what's the fundamental difference between the 0.13.16-M1 release and a "full" release? Is it just that you don't bother to write release notes and publish documentation for M releases, or are there underlying technical differences?

ci/matrix.xml
sjrd: > `${sbt_version_override:+^^$sbt_version_override}`

Uh! That's fancy. Thanks.

project/build.properties
dwijnand: It'll be binary compatible, for sure.

Yeah it's just the overhead. Including creating windows .msi's on a Windows box, zip/tgz/rpm/deb, etc.. 0.13.16-M1 is probably only out there as a jars and git tag.

sjrd: Thanks. IMO that's fine, then. @gzm0 WDYT?

eed3si9n: We released 0.13.16-M1 because I wanted `^^` facility to be available to plugin authors without going through the RC cycle. I haven't pushed for final because I wanted to make sure that sbt 0.13.16 final is able to cross build against sbt 1.0.0. Now that sbt 1.0.0-RC2 is out we probably can start the process.

The only reported compatibility issue is the forward compatibility breakage of the ScriptedPlugin in https://github.com/foundweekends/giter8/issues/325. This I think is a corner case because Giter8's plugin depends on Scripted directly - https://github.com/foundweekends/giter8/blob/547db187ceadb4cefc504733ce47225f7b7b47aa/build.sbt#L97

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala
sjrd: For reference, this is the definition of `CrossVersion.Binary`: [`Binary.scala`](https://github.com/sbt/librarymanagement/blob/60dfb6e0b1a1220256f866a55ccda09c16344f4e/core/src/main/contraband-scala/sbt/librarymanagement/Binary.scala). Other `CrossVersion`s are similar.

They are similar to our binary-compatible-forever `Config` objects in shape, except in sbt they auto-generate them from [a big JSON file](https://github.com/sbt/librarymanagement/blob/60dfb6e0b1a1220256f866a55ccda09c16344f4e/core/src/main/contraband/librarymanagement.json#L253), using [contraband](http://www.scala-sbt.org/contraband/).

sjrd:
Here is the full log of the (successful) job that tests sbt 1.0.0-RC2: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/185044/console, for manual inspection. It does indeed use 1.0.0-RC2 at the end ;)

project/build.properties
sjrd: Now updated to 0.13.16-RC1.

sbt-plugin/src/main/scala-sbt-0.13/org/scalajs/sbtplugin/SBTCompat.scala
```diff
@@ -0,0 +1,63 @@
...
+  type CompileAnalysis = sbt.inc.Analysis
+
+  class sbtUnchecked extends scala.annotation.Annotation
+
+  val formatImplicits: sbt.Cache.type = sbt.Cache
+
+  def moduleIDWithConfigurations(moduleID: ModuleID,
+      configurations: Option[String]): ModuleID = {
+    moduleID.copy(configurations = configurations)
+  }
```
sjrd: FTR, sbt 0.13.16-RC1 added a backport of `withConfigurations` to `ModuleID`, so we could get away without this particular shim in `SBTCompat`. However that would definitely force our users to upgrade to sbt 0.13.16-RC1+ as well (and I'm pretty sure that failing to do so results in ugly obscure error messages like `NoSuchMethodError`), so I wanted to avoid that.

gzm0:
FTR: I have manually checked that there are no traces of 0.13.16-RC1 in the build artifacts of the sbt-plugin.

gzm0:
Actually, scratch that:

```
ivys/ivy.xml:		<dependency org="org.scala-sbt" name="sbt" rev="0.13.16-RC1" conf="provided->default(compile)"/>
```

gzm0:
I have un-merged this. Sorry for the fast click. I first want to understand how resolvers are going to handle the dependency on the RC-1 library.

sjrd:
Revived PR after the rollback: #3075.

PR #3071, opened 2017-07-22
Fix #3064: Validate the target type of Is/AsInstanceOf nodes.

### LINKED ISSUE #3064: IR check that the rhs of Is/AsInstanceOf nodes makes sense
They cannot be primitive types nor JS types.

sjrd:
They must not be primitive types nor JS types.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
@@ -893,9 +893,11 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
 
       case IsInstanceOf(expr, cls) =>
         typecheckExpr(expr, env)
+        checkIsAsInstanceTargetType(cls)
 
       case AsInstanceOf(expr, cls) =>
         typecheckExpr(expr, env)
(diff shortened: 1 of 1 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -1059,6 +1061,28 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+        } else {
+          val kind = lookupClass(encodedName).kind
+          if (kind.isJSType) {
+            reportError(
+                s"JS type $encodedName is not a valid target type for " +
+                "Is/AsInstanceOf")
+          }
+        }
+
+      case ArrayType(_, _) =>
```
gzm0: Don't we need to check that the referenced class exists? Is the Info checker doing that?

sjrd: The reachability analysis does that, because the class data of any target type of a `Is/AsInstanceOf` is listed in the "referenced class data" fields of infos (which are checked by the info checker). If it's not found by the reachability analysis, you'll get a linking error before you even get to the IR checker.

PR #3073, opened 2017-07-22
Reduce the maximum memory consumption in the bootstrap test.

sjrd:
We do this by restarting sbt more often, so that irJS, testSuite and toolsJS are full-optimized in 3 separate VMs.


sjrd:
In order to produce the list of testing frameworks and detected tests, `toolsJS/test` calls `testSuite/test:loadedTestFrameworks` and `testSuite/test:definedTests`, which needs `testSuite/test:fullOptJS`.

PR #3074, opened 2017-07-23
Inline init methods in the JS constructors.

sjrd:
This a rebased/updated version of #2142.

At definition site, instead of having a constructor that only creates fields, and an `init___XYZ` method for the actual constructor, we inline the body of the `init___XYZ` method inside the JavaScript constructor, giving:
```js
class $c_Foo extends $c_O {
  constructor(args) {
    super();
    this.x$1 = 0;
    this.y$1 = 0;
    // body of the init method
  }
}
```
At call site, this allows to emit
```js
new $c_Foo(args);
```
instead of
```js
new $c_Foo().init___XYZ(args);
```
This produces shorter code, and is potentially friendlier to JavaScript JITs.

We can do this for a class when all of the following apply:

* The class is not instantiated from `scalajsenv.js`
* It does not have any instantiated subclasses
* It has exactly one (regular) constructor
* It does not have any exported constructor


gzm0:
Any particular reason this is targeted on master?

sjrd:
Oh yes: not having to deal with Rhino and its own peculiar encoding of stuff!
IMO that's the biggest deal about dropping Rhino (besides 2x faster CI): the fact that it's much easier to improve the emitter.

tools/scalajsenv.js
```diff
@@ -922,7 +922,7 @@ $TypeData.prototype.getClassOf = function() {
 getClassOf() {
 //!endif
   if (!this._classOf)
-    this._classOf = new $c_jl_Class().init___jl_ScalaJSClassData(this);
+    this._classOf = new $c_jl_Class(this);
```
gzm0: This contradicts the commit message :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -133,8 +133,43 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
       yield js.ClassDef(Some(classIdent), parentVar, members)(tree.pos)
   }
 
+  /** Extracts the inlined init constructor, if there is one. */
+  def extractInlinedInit(tree: LinkedClass)(
```
gzm0: Consider "inlineable" (here and in many places). "inlined" sounds like it has already happened. If you are concerned about it sounding like a choice, at least improve the doc.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -133,8 +133,43 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
...
+          s"Found ${constructors.size} constructors in class " +
+          s"${tree.encodedName} which has an inlined init.")
+      (Some(constructors.head), otherMethods)
+    } else {
+      (None, memberMethods)
+    }
+  }
+
+  /** Generates an inlined init. */
+  def genInlinedInit(className: String, method: MethodDef)(
```
gzm0: I'm not sure the general purpose method `genInlinedInit` brings any value here... It's a `genMethod` essentially :S

In any case, the name or the doc needs to improve. It is absolutely not clear what parts of the overall inlined constructor are generated. (and "Generates an inlined init" as a doc comment on `genInlineInit` is frankly useless).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -133,8 +133,43 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
...
+    assert(method.resultType == NoType,
+        s"Found a constructor with type ${method.resultType} at $pos")
+
+    desugarToFunction(className, method.args, methodBody, isStat = true)
+  }
+
   /** Generates the JS constructor for a class. */
-  def genConstructor(tree: LinkedClass)(
+  def genConstructor(tree: LinkedClass,
+      inlinedInit: Option[WithGlobals[js.Function]])(
```
gzm0: `initToInline`? It is not inlined (yet).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -345,11 +345,20 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
     if (linkedClass.hasInstances && kind.isAnyNonNativeClass) {
+      val (linkedInlinedInit, linkedMemberMethods) =
+        classEmitter.extractInlinedInit(linkedClass)(classCache)
+
+      // JS constructor
+      val inlinedInit = linkedInlinedInit.map { m =>
+        val inlinedInitCache = classCache.getInlinedInitCache()
+        inlinedInitCache.getOrElseUpdate(m.version,
+            classEmitter.genInlinedInit(className, m.tree)(inlinedInitCache))
+      }
```
gzm0: Not sure this is the best place to put all this. Have you considered putting all this inside `classEmitter` and simply filtering out the inlined constructor in the for loop below (the entire `emitter` knows about inlined constructors anyways, so it doesn't couple any tighter).

In any case, I do not see a use for the `inlinedInitCache` it gets cached right away afterwards...

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -345,11 +345,20 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
       val ctor = classTreeCache.constructor.getOrElseUpdate(
-          classEmitter.genConstructor(linkedClass)(classCache))
+          classEmitter.genConstructor(linkedClass, inlinedInit)(classCache))
```
gzm0: This is broken. If `inlinedInit` gets updated (i.e. the ctor body), but the class does not, this does not regenerate.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -73,6 +78,52 @@ private[emitter] final class KnowledgeGuardian {
...
+     */
+    val blackList = Set(
+        "jl_ClassCastException",
+        "jl_ArrayIndexOutOfBoundsException",
+        "sjsr_UndefinedBehaviorError",
+        "js_CloneNotSupportedException"
+    )
+
+    val scalaClassDefs = linkingUnit.classDefs.filter(_.kind.isClass)
+    val instantiatedClasses = scalaClassDefs.filter(_.hasInstances)
```
gzm0: Consider inlining this below to make it explicit that it is unused later. Being able to give it a name does not add any value IMO.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -73,6 +78,52 @@ private[emitter] final class KnowledgeGuardian {
...
+            _.isInstanceOf[TopLevelConstructorExportDef])
+      }
+    }
+
+    (for {
+      classDef <- scalaClassDefs
+      if enableInlinedInitFor(classDef)
+    } yield {
+      classDef.encodedName
+    }).toSet
```
gzm0: Use `classDef.filter(enableInlinedInitFor).map(_.encodedName).toSet`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -73,6 +78,52 @@ private[emitter] final class KnowledgeGuardian {
...
+      instantiatedClasses.flatMap(_.superClass).map(_.name).toSet
+
+    val cantHaveInlinedInit = classesWithInstantiatedSubclasses ++ blackList
+
+    def enableInlinedInitFor(classDef: LinkedClass): Boolean = {
+      /* We can enable inlined init if all of the following apply:
+       * - The class is not blacklisted
+       * - It does not have any instantiated subclass
+       * - It has exactly one (regular) constructor
+       * - It does not have an exported constructor
```
gzm0: I like this list. I'd consider not constructing the intermediate set `cantHaveInlinedInit` and just add two set lookups.

gzm0:
> Oh yes: not having to deal with Rhino and its own peculiar encoding of stuff!

Ah, ah, indeed.

gzm0:
Also, just for my understanding: The restriction on exported constructors is that if we create an exported constructor on a class with a single constructor, we essentially create more than one constructor (the real one, and the exported one that typechecks). In ES6, we cannot forward constructors, so both of them need to be real constructors, therefore inlining is hardly desirable (since it would copy the init methods). Sounds about right?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
sjrd: Oh gosh. The whole point of the separate `genInlinedInit` and `inlinedInitCache` was to make this work in all cases ... and it's still broken. Well spotted!

sjrd:
Updated.

> The restriction on exported constructors is that if we create an exported constructor on a class with a single constructor, we essentially create more than one constructor (the real one, and the exported one that typechecks).

Yes, that's the idea. I added a small line about that in the list of criteria necessary for `init` to be inlineable.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -502,25 +529,38 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
         invalidate()
 
       _staticCaches.nonEmpty || _methodCaches.nonEmpty || _cacheUsed
     }
   }
 
-  private final class MethodCache extends knowledgeGuardian.KnowledgeAccessor {
-    private[this] var _tree: WithGlobals[js.Tree] = null
+  private final class MethodCache[TreeType]
+      extends knowledgeGuardian.KnowledgeAccessor {
```
sjrd: I can actually revert that generalization with a type parameter. Now all uses of `MethodCache` are back to using `js.Tree`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -870,9 +945,7 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
                 genNonNativeJSClassConstructor(className),
                 Nil)
           } else {
-            js.Apply(
-                js.New(encodeClassVar(className), Nil) DOT js.Ident("init___"),
-                Nil)
+            js.New(encodeClassVar(className), Nil)
```
gzm0: This implicitly uses the fact that every module class only has a single init method, right? I'm fine with imposing this, but is it IR checked? (Before, one could theoretically add another constructor and instantiate the class by hand, not that it is a good idea).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -444,6 +460,11 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
 
   // Helpers
 
+  private def mergeVersions(v1: Option[String],
+      v2: Option[String]): Option[String] = {
+    v1.flatMap(s1 => v2.map(s2 => s1 + s2))
```
gzm0: Nope :( Consider:

```scala
mergeVersions("1", "11") // --> 111
mergeVersions("11", "1") // --> 111
```

Bad. Maybe we should allow `Any` as version and just rely on the equality method? Then you could just use a tuple here? (or do we need versions to be serializable?)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -73,6 +78,57 @@ private[emitter] final class KnowledgeGuardian {
...
+
+    def enableInlineableInitFor(classDef: LinkedClass): Boolean = {
+      /* We can enable inlined init if all of the following apply:
+       * - The class is not blacklisted
+       * - It does not have any instantiated subclass
+       * - It has exactly one (regular) constructor
+       * - It does not have any exported constructor (since they are
+       *   effectively secondary constructors)
+       *
+       * By construction, this is always true for module classes.
```
gzm0: IR check this :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
gzm0: We probably could allow `Any` versions in the caches?

sjrd: I think here we can solve this case with `s1.length + "-" + s1 + s2`.

gzm0: That works too, yes.

sjrd:
Updated.

Note that, although I do check that module classes have exactly 1 constructor (without parameter), I do not check yet that module classes are not *extended*. Doing this proves to be quite difficult, and reveals that inheritance relationships between classes of all kinds are woefully underchecked. This needs a dedicated PR.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -345,11 +345,27 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+
+      // JS constructor
+      val ctor = {
+        /* The constructor depends both on the class version, and the version
+         * of the inlineable init. When there is no inlineable init, we use ""
+         * as its version.
+         */
+        val ctorCache = classCache.getConstructorCache()
+        val optInlineableInitVersion =
+          linkedInlineableInit.fold[Option[String]](Some(""))(_.version)
```
gzm0: Oh man, I also didn't spot this... In theory this can also cause a collision (if the method has version `""`).

sjrd: Yeah but ... that's not really something we do, having `""` as version. We might even have some implicit guarantee that a version is never `Some("")`.

If you prefer, it's still possible to be more robust with something like:
```scala
ctorVersion = linkedInlineableInit.fold[Option[String]] {
  linkedClass.version.map("1-" + _)
} { inlineableInitVersion =>
  mergeVersions(linkedClass.version, inlineableInitVersion).map("2-" + _)
}
```

gzm0: I'd rather have that, yes. Sorry for not spotting it earlier. The problem with implicit guarantees is that they are implicit :)

sjrd: Done.

PR #3075, opened 2017-07-23
Fix #3065: Cross-compile the sbt plugin for sbt 1.0.0-RC2.

### LINKED ISSUE #3065: Build and publish the sbt plugin for sbt 1.0.0-RC2

sjrd:
Revival of #3068 after the rolled back merge.

After discussion, this is waiting for sbt 0.13.16 final to be released, so that we don't have our stable artifacts depend on sbt's pre-release artifacts.

/cc @dwijnand @eed3si9n

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -260,14 +260,18 @@
   ]]></task>
 
   <task id="sbtplugin-test"><![CDATA[
+    setJavaVersion $java
+    SBT_VER_OVERRIDE=$sbt_version_override
     # Publish Scala.js artifacts locally
     # Then go into standalone project and test
     sbt ++2.11.11 compiler/publishLocal library/publishLocal javalibEx/publishLocal \
                   testInterface/publishLocal stubs/publishLocal \
                   jUnitPlugin/publishLocal jUnitRuntime/publishLocal &&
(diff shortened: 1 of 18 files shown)


dwijnand:
sbt 0.13.16 has been released btw.

sjrd:
Thanks.

I amended the first commit for 0.13.16. Also I removed the shim for `@sbtUnchecked` in `SBTCompat`, since it's now part of sbt 0.13.16 itself (concerns about forward bin-compat don't apply here, because the annotation is only used on expressions within methods, and those don't survive in the bytecode).

@gzm0 Should be good to go, assuming CI passes. Once this is merged I'll release 0.6.19.

dwijnand:
Btw I reviewed your PR and it LGTM too. (I was concerned about someone else breaking from forward compatibility, like I did with MiMa).

gzm0:
LGTM.

Issue #3076, opened 2017-07-25
The linker must check the validity of inheritance relationships


sjrd:
Some examples:

* Module classes cannot be extended
* Superclasses must be classes (not interfaces)
* Implemented interfaces must be interfaces
* A JS type cannot extend a Scala type
* A Scala type cannot extend a JS type

I'm not sure yet whether this should go to 0.6.x or 1.x.

sjrd:
Fun fact: we *are* breaking these invariants already:
```
[error] scala.Dynamic (of kind Interface) is not a valid interface implemented by scala.scalajs.js.Dynamic (of kind AbstractJSType)
[error]   called from core module analyzer
[error] scala.Dynamic (of kind Interface) is not a valid interface implemented by scala.scalajs.js.Dynamic$global$ (of kind NativeJSModuleClass)
[error]   called from core module analyzer
```
That happens because in PrepJSInterop [we explicitly introduce a loophole for `scala.Dynamic`](https://github.com/scala-js/scala-js/blob/e4ab4b231b32715e0bc8df0a2e64e5886ee2a70f/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala#L466-L469).

Since I think our IR itself should not be encumbered by this loophole, the compiler back-end should remove `scala.Dynamic` from the superinterfaces of JS types.

This means that adding the checks can only be done in 1.x.

PR #3083, opened 2017-08-02
Fix #3078: New Analyzer, bye bye Infos.

### LINKED ISSUE #3078: New Analyzer, bye bye Infos
Scary title, uh?

Background: `Infos` are summaries of each method in a class, telling which other methods it calls, which classes it instantiates, etc. Those Infos are used by the `Analyzer` (aka reachability analysis or DCE) to perform its job. `Infos` are computed at compile-time, and stored in `.sjsir` files. That way, at link-time, only the Infos need to be read from all the `.sjsir` files, then the Analyzer can run, and only used classes need to be fully deserialized into Trees.

Sounds like a good idea, right? And it is, from a performance and memory consumption point of view. However, it has one major drawback: we cannot improve the Analyzer in binary compatible versions of Scala.js in ways that would require *more* information to be stored in the Infos. Indeed, we could deserialize old `.sjsir` files not containing all the infos we need, and be stuck. This is the reason things like #2227 cannot be implemented in minor versions of Scala.js.

How do we solve this? Easy: we compute the infos at link-time rather than at compile-time, and we don't store them in `.sjsir` files. New problem: we need to deserialize all the trees on the classpath, which costs memory and time. New solution: be *much* smarter.

## Goals of the new analyzer

* Not require Infos to be stored in `.sjsir` files
* Be at least as fast as the current analyzer
* Be at most as much memory-hungry as the current analyzer

### Read files and construct infos lazily

Currently, we read the infos of all the `.sjsir` files on the classpath, and the trees of those that are "needed at all" after `BaseLinker`. Ideally we should only read even the infos of classes that are "needed at all". That is in theory possible, because a class that is not needed at all does not participate in the reachability algorithm anyway.

How do we do that? Well, we start the reachability without having loaded anything. We only load the entry points of the classpath. Entry points are:

* `symbolRequirements`
* *Top-level* exports
* Static initializers of Scala classes

The first thing, we know in advance in which classes they are, by definition, so we know which files we have to read. For top-level exports and static initializers, we don't know, as that information is in the files themselves.

We can fix that with an extremely stripped down notion of "infos". The Infos reduce to 1 bit, telling whether or not the class has any entry point. That way, we only need to read that 1 bit from all `.sjsir` files on the classpath.

After that, we can have the reachability analysis algorithm start from the entry points. And every time it refers to a new class (via `lookupClass`), we load the corresponding `.sjsir` file, deserialize it to a tree, and derive the infos from the tree. At this point we know said class is needed at all, by construction. Of course, we cache the tree and infos for incremental runs.

For batch mode, we spend some additional time on deriving the infos from the trees, but we deserialize less, so it's unclear what the impact is on processing time. In an incremental run, we already win on processing time because:
* The time spent to derive the infos at link-time is only for changed classes, which necessarily come right from the compiler, which would previously have to derive those infos itself
* We save the time spent to serialize and deserialize the infos, as well as the I/O those additional bytes incur

**Achievement:**
* Infos are not in `.sjsir` files anymore
* less memory consumption, since only the infos of needed-at-all classes are deserialized and kept in memory
* faster in incremental runs
* potential cost: processing time in batch mode maybe increases

### Parallelism

When we lazily load a class, the reachability algorithm per se is blocked on several tasks to be performed on a single file:

1. Read the bytes from the disk, and uncompress them if them come from jars (not exactly true in our case for .jars, because all the files in jars are eagerly loaded and uncompressed in memory by `IRFileCache`)
2. Deserialize to build the tree
3. Derive the infos from the tree

These operations don't need to touch the data structure used by the reachability algorithm. Therefore, in theory they can be done in parallel. But that's assuming the algorithm can continue doing some work while waiting for the given infos.

That's possible because a reachability algorithm is essentially a fixed point computation, which can be organized with a work list of edges to be processed. Currently the analyzer maintains this work list implicitly on the stack, through recursion. However, it can be (partially) turned into a reified work list processed by a loop. If we do this, when we want to touch a new class, instead of processing the edge that goes to that class right away, we store it in a to-do list for that class, to be processed later when the class has been loaded.

Transforming the algorithm to support that won't be trivial, but it will allow to deserialize the trees and derive the infos in parallel with the actual reachability analysis.

**Achievement:**
* Faster in both batch and incremental runs, due to deserialization being performed in parallel. This should in particular have a major impact on batch mode, and reclaim the potential loss from the previous section.

sjrd:
This is a big rewrite of all the infrastructure around `Infos`, and how they are used by the `Analyzer` to compute the reachability analysis.

Previously, Infos were a "specified" part of the IR, and were stored in .sjsir files. This was limiting the evolution of the reachability analysis in binary compatible releases, as explained in #3078.

With this commit, Infos become almost an implementation detail of the `Analyzer`. They are also known from the `BaseLinker` and `Refiner`, who need to provide inputs to the `Analyzer`, but are otherwise unknown from the rest of the codebase. Obviously, they are not stored in .sjsir files anymore.

The `Analyzer` will now lazily read the full `ClassDef`s of classes it needs, and derive Infos from them at link time. In the Refiner, Infos are also lazily derived from `LinkedClass`es, whereas they were eagerly recomputed in `GenIncOptimizer` before. However, whereas before all Infos on the classpath were read and deserialized, now the Analyzer is smarter: it only reads trees and computes Infos for classes that are actually touched by the reachability analysis algorithm. In other words, only classes that are "needed at all" are deserialized. In fact, with this commit, the concept of "needed at all" disappears: classes that appear in the resulting `Analysis` are by construction "needed at all".

In order for the Analyzer to know its entry points, .sjsir files store a tiny structure `EntryPointsInfo` containing the encoded name of the class, and a flag indicating whether the class has any entry point. That structure is all that is deserialized for all .sjsir files on the classpath.

For incremental runs, the `BaseLinker` and `Refiner` obviously cache the computed Infos between runs, and avoid recomputing them if their inputs (`ClassDef`s or `LinkedClass`es, resp.) have not changed.

Since Infos become an implementation detail of the `Analyzer` and its "friends" `BaseLinker` and `Refiner`, the `-i`/`--info` option of `scalajsp` is removed (both in the CLI and in the sbt plugin). The `InfoPrinter` class is consequently removed as well.

Moreover, since Infos are computed at link time, there is no point in *checking* them anymore, which means that the `InfoChecker` is removed.


sjrd:
Ready for review :)

I haven't done performance tests yet. I'm currently on vacation with only a Windows laptop, unfit for that kind of tests. Maybe I'll have the courage to do some tests over ssh, but most likely it will have to wait until after I'm back on August 16.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
```diff
@@ -95,3 +107,137 @@ final class Refiner(config: CommonPhaseConfig) {
...
+      }
+    }
+
+    def cleanAfterRun(): Unit = {
+      linkedClassesByName = null
+      cache.retain((_, linkedClassCache) => linkedClassCache.cleanAfterRun())
+    }
+  }
+
+  private class LinkedClassInfoCache {
```
sjrd: The things that happen in this class are really ugly, but I don't seem to find a nicer way. I'm happy to take suggestions ;)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/LinkedMember.scala
```diff
@@ -23,8 +22,7 @@ import ir.Infos
  *  reliably to determine at phase P+1 whether a linked member coming from
  *  phase P must be reprocessed.
  */
-final class LinkedMember[+T <: MemberDef](
-    val info: Infos.MethodInfo,
+final class LinkedMember[+T <: IRNode](
```
sjrd: It might make sense to rename this into `Versioned[+T]` at this point. Or else leave `LinkedMember` as is and add `LinkedTopLevelExport`.

Opinion?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -196,48 +246,40 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-        cls.linkClasses()
-        cls.ancestors
-      }.distinct
+      superClass = data.superClass.map(lookupClassAndCheckCycles(_))
+      interfaces = data.interfaces.map(lookupClassAndCheckCycles(_))
 
-      for (ancestor <- ancestors)
-        ancestor.descendants += this
-    }
+      // TODO #3076: Validate the super class and implemented interfaces
```
sjrd: The current PR does not improve nor worsen the situation wrt. #3076. Addressing that issue should be done in a separate PR. The new architecture of the Analyzer will make it easier, though, as we get here only for classes that really need to be linked (needed at all), which means we can eagerly report errors if things don't match. With the previous architecture, this was virtually impossible to do.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -1118,10 +1122,10 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
 
   private def lookupInfo(className: String)(
       implicit ctx: ErrorContext): Infos.ClassInfo = {
-    unit.infos.getOrElse(className, {
+    inputProvider.loadInfo(className).getOrElse {
       reportError(s"Cannot find info for class $className")
       Infos.ClassInfo(className)
-    })
+    }
```
sjrd: This part is still necessary because the IR checker needs the Infos of some classes that are not "needed at all", and are therefore not part of the `LinkingUnit`. An obvious example is abstract JS types that only appear in the signatures of methods. The IR checker wants to validate that `sjs_js_Function1` has type `any` when it appears in a signature, but the reachability analysis otherwise does not care about `js.Funtion1` (unless its `classOf` is needed for some reason).

Arguably, a future version of the Analyzer would at least *link* such classes, so that a `LinkingUnit` is internally consistent, and the IR checker does not need this "hack". Since the current PR does not make the situation better nor worse in this regard, this should be done later, if at all.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/BaseLinker.scala
```diff
@@ -37,61 +36,33 @@ import Analysis._
...
+      val encodedNameToFile = mutable.Map.empty[String, VirtualScalaJSIRFile]
+      for (irFile <- irInput) {
+        // Remove duplicates. Just like the JVM
+        val encodedName = irFile.entryPointsInfo.encodedName
+        if (!encodedNameToFile.contains(encodedName))
+          encodedNameToFile += encodedName -> irFile
       }
-    }
+      encodedNameToFile
+    })
```
gzm0: Why not put this into the input provider?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
gzm0: I think the main reason this is so ugly is that there is no `MethodInfoCache`.

If you add a `MethodInfoCache`, updating the class def becomes:
- Update all MethodInfoCaches
- Regenerate ClassInfo pulling it in from the method caches (don't even bother caching it, it's just copying stuff).

Have you considered that?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -114,7 +114,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
   private lazy val inlinedRTLongHiField =
     inlinedRTLongRecordType.fields(1).name
 
-  def optimize(thisType: Type, originalDef: MethodDef): LinkedMember[MethodDef] = {
+  def optimize(thisType: Type, originalDef: MethodDef): MethodDef = {
```
gzm0: Nice!

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -158,38 +154,45 @@ trait VirtualRelativeScalaJSIRFile extends VirtualScalaJSIRFile
...
+    extends VirtualBinaryFile with VirtualScalaJSIRFile {
 
-      case e: IOException =>
-        throw new IOException(s"Failed to deserialize info of $path", e)
-    } finally {
-      stream.close()
+  override def entryPointsInfo: ir.EntryPointsInfo = {
+    // Overridden to read only the necessary parts
+    enrichIOExceptions {
+      val stream = inputStream
```
sjrd: It is declared without `()`, so I cannot call it with `()`. It's been there since 0.5.0-M1, basically (or at least, `VirtualTextFile.reader`, which is similar, has been). I guess it can be argued that opening the stream itself does not have *side effects* per se, since it does not modify the world. It is clearly not *pure* though, since something else modifying the world will cause that method to behave differently (e.g., throw a `FileNotFoundException`). It comes down to the distinction whether `()` signals *side effects* or *impurity*. Since getters of a mutable var do not have `()`, it cannot be true that `()` signals impurity, so it must be side effects. And according to that logic, `inputStream` should not have `()` either.

But it's fuzzy and debatable.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -34,73 +38,68 @@ private final class Analyzer(config: CommonPhaseConfig,
...
       optional: Boolean = false): Unit = {
 
-    def withClass(className: String)(body: ClassInfo => Unit) = {
-      val clazz = lookupClass(className)
-      if (!clazz.nonExistent || !optional)
-        body(clazz)
+    def withClass(className: String)(body: ClassInfo => Unit)(
+        implicit from: From): Unit = {
+      if (optional)
+        tryLookupClass(className).foreach(body)
```
sjrd: I don't think so. That additional check is done in `withMethod`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -180,9 +182,57 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+            _errors += MissingClass(info, from)
+            info
+        }
+        _classInfos += encodedName -> info
+        info.linkClasses()
+        info
+    }
+  }
+
+  private class ClassInfo(data: Infos.ClassInfo, val nonExistent: Boolean)
```
sjrd: Factory: 👍 

I think keeping `linkClasses()` separate is desirable, precisely because it highlights the mutability of this algorithm. It is important to think about that mutability because of cycle detection. Cycle detection relies on the mutable aspects and the precise timing of mutations involving `_classInfos +=` and `linkClasses()`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -531,27 +559,19 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-        if (isAnyModuleClass) accessModule()
-        else instantiated()
-      }
-
-      // My methods
-      if (!isJSClass) {
-        for (methodInfo <- methodInfos.values) {
-          if (methodInfo.isExported)
-            callMethod(methodInfo.encodedName)
-        }
```
sjrd: Yes, that's right.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -587,10 +607,21 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+            ancestor.instantiatedSubclasses ::= this
+            ancestor.methodsCalledLog
+          }
+
+          for {
+            log <- allMethodsCalledLogs
+            logEntry <- log
+          } {
+            val methodName = logEntry._1
+            implicit val from = logEntry._2
```
sjrd: Not necessarily, but it's definitely more efficient. Destructuring directly in the for loop creates an awful `withFilter` and other nasty inefficient things. We carefully avoid that kind of caveats in all the algorithms that are not incremental in the linker.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -827,15 +862,13 @@ private final class Analyzer(config: CommonPhaseConfig,
   }
 
   private def createMissingClassInfo(encodedName: String): Infos.ClassInfo = {
```
sjrd: Yes, indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
sjrd: #3088

sjrd:
Still to-do:
* `LinkedMember` -> `Versioned`
* Try `MethodInfoCache` in `Refiner`
* Performance measurements

sjrd:
Updated. I have left the `Versioned` renaming and the per-method Infos computation in `Refiner` as separate commits, for easier review, but I think they should be squashed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
```diff
@@ -149,12 +149,10 @@ private object Refiner {
...
-    private var staticMethodsVersions: List[(String, Option[String])] = Nil
-    private var memberMethodsVersions: List[(String, Option[String])] = Nil
-    private var abstractMethodsVersions: List[(String, Option[String])] = Nil
-    private var exportedMembersVersions: List[(String, Option[String])] = Nil
-    private var topLevelExportsVersions: List[(String, Option[String])] = Nil
+    private val staticMethodsInfoCaches = LinkedMembersInfosCache()
+    private val memberMethodsInfoCaches = LinkedMembersInfosCache()
+    private val abstractMethodsInfoCaches = LinkedMembersInfosCache()
+    private val exportedMembersInfoCaches = LinkedMembersInfosCache()
     private var info: Infos.ClassInfo = _
```
gzm0: Why is this still a var?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
```diff
@@ -166,72 +164,101 @@ private object Refiner {
...
       }
+      result
     }
+  }
 
-    private def linkedTopLevelExportListVersion(
-        topLevelExports: List[Versioned[TopLevelExportDef]]): List[(String, Option[String])] = {
-      topLevelExports.map(e => (e.value.topLevelExportName, e.version))
+  private final class LinkedMembersInfosCache(
+      val caches: mutable.Map[(Boolean, String), LinkedMemberInfoCache])
```
gzm0: Why is this not just a class member?
Also: I didn't know that static and instance methods can share names :) (btw. do we IR check this)?

sjrd: So that `LinkedMembersInfosCache` can be an `AnyVal`, reducing indirections.

And yes, static and instance members can share names. We "do" IR-check this in the sense that we don't do anything about it ;) If they were exclusive, we would have to check exclusivity.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
sjrd: In case `loadInfo()` is called twice during the same run (see also `if (!cacheUsed)` in `update`). Currently, the analyzer and the IR checker should never call it twice for the same class (even together), but I'd rather not endanger performance in subsequent changes.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -180,9 +180,59 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+          case None =>
+            val data = createMissingClassInfo(encodedName)
+            val info = newClassInfo(data, nonExistent = true)
+            _errors += MissingClass(info, from)
+            info
+        }
+    }
+  }
+
+  private def newClassInfo(data: Infos.ClassInfo, nonExistent: Boolean)(
```
sjrd: I had not. I can do that, indeed.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/Versioned.scala
```diff
@@ -0,0 +1,25 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js tools             **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2014, LAMP/EPFL   **
```
sjrd: Right. I think there are sbt plugins that enforce that this kind of stuff is consistent all over a codebase.

sjrd:
Updated.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
gzm0: Ah, right. Didn't see the assignment is guarded.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/Refiner.scala
gzm0: Ah, did miss that this is an `AnyVal`.

sjrd:
Yes, let's wait until I get to do those perf tests. I'll try to do that tomorrow.

sjrd:
Results of the benchmarks: https://gist.github.com/sjrd/5c5c9b9f3f8da398af9a560dd20978bb

Everything is done on the test suite.

**warm, compile**

the time taken by `GenJSCode` to compile all of `test:compile`. The time difference between master and new-analyzer seems to be slightly in favor of new-analyzer, but with such a variance that it's probably non-significant

**cold, batch** (with `checkIR`)

`testSuite/test:fastOptJS` on a cold sbt/JVM, obviously in batch mode. The results here are dominated by the disappearance of the Check Infos subphase, which took a whole second. Timings of individual subphases cannot be compared, because the Check Infos part forced all the files to be loaded in memory, including their Trees and Infos, which means there was much less work for the other subphases to do.

Because the presence of Check Infos completely destroys everything else about the analysis, all the subsequent measurements are done with checkIR disabled.

**cold, batch, without checkIR**

Here we see that Basic Linking is slightly faster (~200ms) in new-analyzer, despite having to recompute the Infos. This could be caused by not having to construct the in-memory representation of the unused Infos, but somehow I don't buy it. The difference is probably not significant.

The Refiner is significantly slower in new-analyzer in this configuration. But that's also because the Refiner stole some work previously done by the the Optimizer: computing the Infos of optimized methods. And indeed, if we count the improved timings of the Optimizer from master to new-analyzer, it seems to compensate for the longer time in Refiner.

**warm, batch, without checkIR**

Here, in addition to the JVM being warm, the global IR cache has also already been filled with a lot of things. So even with a clean instance of `Linker`, some significant I/O work is removed. This explains the huge difference in timings between this set of measurements and the previous one.

Between master and new-analyzer, the timings are inconclusive for Basic Linking. Refiner has similar characteristics as the previous set.

**warm, degenerate incremental, without checkIR**

Here, we only `rm` the `-fastopt.js` file, and re-link. This means that all inputs are unchanged, and the linker is fully up-to-date in its incremental data structures.

Times for Basic Linking are inconclusive again. The Refiner seems to be slower in new-analyzer, which probably means that using the cached Infos in Refiner is somewhat less efficient than directly using those from the Optimizer.

**warm, test-suite test-classes re-compiled, no checkIR**

In this scenario, we force re-compilation of all files in the test-suite, but otherwise keep a warmed up linker. This forces the Basic Linking step to recompute all the Infos of the test-suite, but not of the standard library.

Here it seems that the Basic Linking step is basically not impacted at all between master and new-analyzer, despite new-analyzer having to recompute all those Infos. The Refiner seems slightly faster, but by a small margin.

**warm, change one source, no checkIR**

This is supposed to be a "typical" incremental run, where one source file is changed.

The Basic Linking step seems slightly faster in new-analyzer. The Refiner is also faster in new-analyzer, by a somewhat larger margin.

**Conclusion**

Overall I think these results are good enough. At the very least, they show that we don't have any regression in incremental runs (we even seem to be a bit faster). The batch runs are a bit slower in some configurations, as expected, but by a rather small margin.

@gzm0 Does that satisfy you?

gzm0:
LGTM Ship It :)

Issue #2184, opened 2016-01-19
run/literals.scala has linker *warnings* (pbl in refiner)


sjrd:
I suspect the new scheme for symbol requirements to have a problem somewhere.
```
!! 784 - run/literals.scala                        [output differs]
% diff /localhome/jenkins/a/workspace/scalajs-task-worker/partest/fetchedSources/2.11.7/test/files/run/literals-run.log /localhome/jenkins/a/workspace/scalajs-task-worker/partest/fetchedSources/2.11.7/test/files/run/literals.check
@@ -10,16 +10,3 @@ literals.scala:37: warning: Octal escape literals are deprecated, use \u0000 ins
 literals.scala:37: warning: Octal escape literals are deprecated, use \u0000 instead.
       "\0x61\0x62".getBytes(io.Codec.UTF8.charSet) sameElements Array[Byte](0, 120, 54, 49, 0, 120, 54, 50),
             ^
-Referring to non-existent method java.lang.Integer.compareTo(java.lang.Byte)scala.Int
-  called from java.nio.ByteBuffer.compareTo(java.nio.ByteBuffer)scala.Int
-  called from java.nio.ByteBuffer.equals(java.lang.Object)scala.Boolean
-  called from scala.runtime.BoxesRunTime$.equals(java.lang.Object,java.lang.Object)scala.Boolean
-  called from scala.collection.GenMapLike$class.liftedTree1$1(scala.collection.GenMapLike,scala.collection.GenMap)scala.Boolean
-  called from scala.collection.GenMapLike$class.equals(scala.collection.GenMapLike,java.lang.Object)scala.Boolean
-  called from scala.collection.AbstractMap.equals(java.lang.Object)scala.Boolean
-  called from scala.runtime.BoxesRunTime$.equals(java.lang.Object,java.lang.Object)scala.Boolean
-    (already seen, not repeating call stack)
-involving instantiated classes:
-  java.nio.HeapByteBuffer
-  scala.runtime.BoxesRunTime$
-  scala.collection.mutable.HashMap
```

PR #3086, opened 2017-08-07
Redesign primitive numeric types in the IR.

### LINKED ISSUE #3085: Linking error after the optimizer for someInt.toDouble.compareTo(double)
The following snippet:
```scala
object Main {
  def main(args: Array[String]): Unit =
    foo(5)

  def foo(x: Int): Unit =
    bar(x.toDouble)

  @inline def bar(x: Double): Unit =
    println(x.compareTo(5.5))
}
```
causes the following linking error after the optimizer has run:
```
[error] Referring to non-existent method java.lang.Integer.compareTo(java.lang.Double)scala.Int
[error]   called from helloworld.HelloWorld$.main([java.lang.String)scala.Unit
[error]   called from core module module initializers
[error]   helloworld.HelloWorld$
```
This is because the type of `x` in `bar()` is refined to be `int` because of the inlining. And then we search for `compareTo__D__B` in `j.l.Integer` instead of `j.l.Double`, which causes the breakage.

This issue is similar to #2184 and #2780, and to the reason we define `compareTo(Byte/Short)` in `j.l.Integer` ([here](https://github.com/scala-js/scala-js/blob/v0.6.19/javalanglib/src/main/scala/java/lang/Integer.scala#L34-L47)). However, the present issue shows that it is more severe than I previously thought. It is not limited to `Byte` and `Short` being treated as `Int`s because there are no `byte`s and `short`s in the IR. In general, it is also problematic because `int <: double` (and `float <: double`) in the IR, which allows the optimizer to refine a `double` into an `int`.

This means that this issue would not be fixed simply by adding `byte`s and `short`s in the IR. We need those ugly bridges in `Integer` (and `Float`) anyway.

Worse, there is a more annoying issue. The following snippet:
```scala
object Main {
  def main(args: Array[String]): Unit =
    foo(5)

  def foo(x: Int): Unit =
    bar(x.toDouble)

  @inline def bar(x: Double): Unit =
    foobar(x)

  @inline def foobar(x: Comparable[java.lang.Double]): Unit =
    println(x.compareTo(5.5))
}
```
links, but fails at run-time with a CCE:
```
[error] scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: 5.5 is not an instance of java.lang.Integer
[error]     at $c_sjsr_UndefinedBehaviorError.$c_jl_Throwable.fillInStackTrace__jl_Throwable (StackTrace.scala:52:51)
[error]     at $c_sjsr_UndefinedBehaviorError.fillInStackTrace__jl_Throwable (UndefinedBehaviorError.scala:24:34)
[error]     at $c_sjsr_UndefinedBehaviorError.$c_jl_Throwable.init___T__jl_Throwable (Throwables.scala:12:19)
[error]     at $c_sjsr_UndefinedBehaviorError.init___T__jl_Throwable (Throwables.scala:156:46)
[error]     at $c_sjsr_UndefinedBehaviorError.init___jl_Throwable (UndefinedBehaviorError.scala:20:5)
[error]     at $throwClassCastException (scalajsenv.js:172:1)
[error]     at $asInt (scalajsenv.js:619:1)
[error]     at $c_Lhelloworld_HelloWorld$.main__AT__V (Integer.scala:8:13)
[error]     at Object.<anonymous> (helloworld-fastopt.js:2912:30)
[error]     at Object.<anonymous> (helloworld-fastopt.js:2913:4)
```
Here, the problem is that we end up calling the bridge `j.l.Integer.compareTo(Object)` instead of `j.l.Double.compareTo(Object)`. But the former does not accept a `Double` as its parameter; it only accepts `Int`s.

That problem cannot be fixed by adding the relevant method in `j.l.Integer`, because *it already exists*, but does the wrong thing.

I'm afraid this issue is unfixable while the IR specifies that `int <: double` *and* `int <: Integer`/`double <: Double`. Since `int <: Integer` is critical to JS interop, I think we have to abandon `int <: double` in the IR. Instead, explicit conversions of the form `UnaryOp(UnaryOp.IntToDouble, x)` will have to be inserted, even though `FunctionEmitter` would erase them at the end of the day.

sjrd:
* Add types for `char`, `byte` and `short` in the IR.
* Remove the subtyping relationships between primitive types, for example `int <: double`.

Instead of the subtyping relationships, we introduce explicit widening conversions as `UnaryOp`s. In order to avoid widening `int`s to `double`s for the purpose of comparing them, we separate `Num_<` and friends into `Int_<` and `Double_<`. For floats, we do introduce the conversions in that case.

There is one subtlety for `char` (`CharType`): it is *not* a subtype of `any`. That's because `char`s must always be boxed in `j.l.Character` when they are assigned to an `any`.

This fixes #3085, and provides a cleaner solution to the old issues #2184 and #2780.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -1233,13 +1233,13 @@ abstract class GenJSCode extends plugins.PluginComponent
           js.BinaryOp(js.BinaryOp.===, js.IntLiteral(lo), numRef)
 
         case (lo, hi) if lo == hi - 1 =>
-          val lhs = js.BinaryOp(js.BinaryOp.===, numRef, js.IntLiteral(lo))
-          val rhs = js.BinaryOp(js.BinaryOp.===, numRef, js.IntLiteral(hi))
+          val lhs = js.BinaryOp(js.BinaryOp.Int_==, numRef, js.IntLiteral(lo))
+          val rhs = js.BinaryOp(js.BinaryOp.Int_==, numRef, js.IntLiteral(hi))
           js.If(lhs, js.BooleanLiteral(true), rhs)(jstpe.BooleanType)
 
         case (lo, hi) =>
-          val lhs = js.BinaryOp(js.BinaryOp.Num_<=, js.IntLiteral(lo), numRef)
-          val rhs = js.BinaryOp(js.BinaryOp.Num_<=, numRef, js.IntLiteral(hi))
+          val lhs = js.BinaryOp(js.BinaryOp.Int_<=, js.IntLiteral(lo), numRef)
+          val rhs = js.BinaryOp(js.BinaryOp.Int_<=, numRef, js.IntLiteral(hi))
           js.BinaryOp(js.BinaryOp.Boolean_&, lhs, rhs)
           js.If(lhs, rhs, js.BooleanLiteral(false))(jstpe.BooleanType)
       }
@@ -1799,7 +1799,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
           val sym = tree.symbol
           val rhsTree =
-            if (rhs == EmptyTree) genZeroOf(sym.tpe)
+            if (rhs == EmptyTree) jstpe.zeroOf(toIRType(sym.tpe))
             else genExpr(rhs)
 
           rhsTree match {
@@ -1912,7 +1912,13 @@ abstract class GenJSCode extends plugins.PluginComponent
               js.Skip()
             case BooleanTag =>
               js.BooleanLiteral(value.booleanValue)
-            case ByteTag | ShortTag | CharTag | IntTag =>
+            case ByteTag =>
+              js.ByteLiteral(value.byteValue)
+            case ShortTag =>
+              js.ShortLiteral(value.shortValue)
+            case CharTag =>
+              js.CharLiteral(value.charValue)
+            case IntTag =>
               js.IntLiteral(value.intValue)
             case LongTag =>
               js.LongLiteral(value.longValue)
@@ -2342,10 +2348,17 @@ abstract class GenJSCode extends plugins.PluginComponent
       val r = toTypeKind(to)
 
       if (l.isValueType && r.isValueType) {
-        if (cast)
-          genConversion(l, r, expr)
-        else
+        if (cast) {
+          /* It is unclear whether this case can be reached for all type
+           * conversions, but scalac handles all cases, so we do too.
+           * Three known user code patterns that become code handled by this
+           * case are `byte.##`, `short.##` and `char.##`, which become, e.g.,
+           * `char.toChar().$asInstanceOf[Int]`.
+           */
+          genConversion(l.toIRType, r.toIRType, expr)
+        } else {
           js.BooleanLiteral(l == r)
+        }
       } else if (l.isValueType) {
         val result = if (cast) {
           val ctor = ClassCastExceptionClass.info.member(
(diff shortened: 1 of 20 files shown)


sjrd:
CI only for now.

sjrd:
Ready for review.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2050,12 +2055,16 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
         contTree(finishTransformExpr(targs.head) match {
           case ClassOf(ClassRef(cls)) =>
             cls match {
-              case "B" | "S" | "C" | "I" | "D" => IntLiteral(0)
-              case "L"                         => LongLiteral(0L)
-              case "F"                         => FloatLiteral(0.0f)
-              case "Z"                         => BooleanLiteral(false)
-              case "V"                         => Undefined()
-              case _                           => Null()
+              case "C" => IntLiteral(0) // yes, as an Int
```
gzm0: Say why?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4215,15 +4306,15 @@ private[optimizer] object OptimizerCore {
...
+        RefinedType(tpe, isExact = false, isNullable = true)
       case NullType =>
         RefinedType(tpe, isExact = true, isNullable = true)
-      case _ =>
-        RefinedType(tpe, isExact = false, isNullable = true)
+      case NothingType | UndefType | BooleanType | CharType | ByteType |
+          LongType | StringType | NoType =>
+        RefinedType(tpe, isExact = true, isNullable = false)
+      case ShortType | IntType | FloatType | DoubleType | RecordType(_) =>
+        RefinedType(tpe, isExact = false, isNullable = false)
```
gzm0: I don't fully understand this. If there are not subtyping relationships between primitive types, why are some of them not exact?

sjrd:
> Does this PR cause any change in size at all or can we optimize it all away?

We do have some more intermediate variables, because the optimizer cannot *alias* some things anymore. However, that should be removed by Closure. (we could also implement a simple aliasing mechanism in the emitter if we want). For example, given the IR:
```scala
def f(x: int) {
  val y: double = (double) x
  println(y)
}
```
previously the generated JS would be
```js
function f(x) {
  println(x)
}
```
but now it is
```js
function f(x) {
  var y = x;
  println(y)
}
```
(simplified view).

sjrd:
On the other hand, we can optimize more unbox operations. Previously `Unbox(x, 'B')` and `Unbox(x, 'S')` would never be optimized away, because `x` would be typed as `int` at best. Now they can if `x` is known to be a primitive `byte`/`short`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Because that's what [the user-space implementation](https://github.com/scala-js/scala-js/blob/4093c9bf41607320eae260f9d7dbd3c8cf02fc0d/scalalib/overrides-2.12/scala/collection/mutable/ArrayBuilder.scala#L79) does as well ;)
Ultimately that's because an `Array[Char]` stores primitive JS numbers, so we don't want the JS array being built to contain boxed chars.

I can add a comment.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
sjrd: I also thought so, and the CI saved me from my mistake. In the *type system* they're not subtypes, but *at run-time* there's another story: a small `int` will still answer `true` to `x.isInstanceOf[Byte]`. Since `isExact` is used to optimize away `isInstanceOf`s that answer `false`, we need those to be considered non-exact.

In fact, I think now even `ByteType` needs to be non-exact :-s Right now I think I broke `(1.toByte: Any).isInstanceOf[Int]`. It would answer `true` at run-time but the optimizer will replace it with `false`.

sjrd:
Updated.

sjrd:
Re-updated with more tests for equality folding, and a related fix.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3579,74 +3629,11 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     /** Gen JS code for a coercion */
     private def genCoercion(tree: Apply, receiver: Tree, code: Int): js.Tree = {
```
gzm0: It's still here :)

sjrd: Ah, somehow I interpreted that you wanted the first overload of `genConversion` to be inlined, not this one. I prefer this one left as a separate method, because its call site is a series of `if (isXYZOp(code)) genXYZ(tree)`. I think it makes sense not to break that pattern.

gzm0: Ok, fair enough.

sjrd:
Rebased (entirely automatic).

PR #3087, opened 2017-08-09
Upgrade to sbt-crossproject 0.2.1.

sjrd:
This will be necessary to support sbt 1.x.


sjrd:
I don't think there's any point in reviewing this, so I'll merge.

PR #3089, opened 2017-08-11
Hijack `j.l.Character` and make `char <: Character` in the IR.

sjrd:
This brings `Char` on par with all the other primitive types as far as the IR is concerned. A primitive `char` (`CharType`) is a subtype of `j.l.Character`. This removes the complexity in the compiler and optimizer related to Chars being "special".

Instead, it is now the Emitter's responsibility to deal with Chars. Indeed, the JS representation of primitive `char`s is *not*, in fact, a subtype of the JS representation of `j.l.Character`s. The root cause being that the `toString()` representation is different. Therefore, now the Emitter auto-boxes primitive `char`s when used in a place where a non-`char` is expected.

This makes the Emitter a bit more complex, and definitely acutely aware of `char`'s. However, it allows the IR to be much cleaner, with no special primitive type that does not behave like the others.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -927,12 +927,6 @@ abstract class GenJSCode extends plugins.PluginComponent
            */
           jstpe.ClassType(encodeClassFullName(tpe.valueClazz))
 
-        case _ if f.tpe.typeSymbol == CharClass =>
-          /* Will be initialized to null, which will unbox to '\0' when
-           * read.
-           */
-          jstpe.ClassType(ir.Definitions.BoxedCharacterClass)
-
         case _ =>
           /* Other types are not boxed, so we can initialize them to
            * their true zero.
@@ -3777,6 +3771,7 @@ abstract class GenJSCode extends plugins.PluginComponent
               (StringClass, StringClass),
               (BoxedDoubleClass, NumberReflectiveCallClass),
               (BoxedBooleanClass, BooleanReflectiveCallClass),
+              (BoxedCharacterClass, CharacterReflectiveCallClass),
               (BoxedLongClass, LongReflectiveCallClass)
           )
           implMethodSym = matchingSymIn(reflBoxClass)
@@ -3805,6 +3800,7 @@ abstract class GenJSCode extends plugins.PluginComponent
               } else {
                 val reflBoxClassPatched = {
                   def isIntOrLongKind(kind: TypeKind) = kind match {
+                    case CharKind     => false
                     case _:INT | LONG => true
                     case _            => false
                   }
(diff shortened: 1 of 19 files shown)


sjrd:
This PR builds on top of #3086. Only the second commit belongs to this PR and should be reviewed here.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/JSExportStaticTest.scala
```diff
@@ -260,7 +260,7 @@ class JSExportStaticTest {
 
     assertEquals('\0',
         JSExportStaticTest.StaticExportFields.uninitializedVarChar)
-    assertEquals(null, statics.uninitializedVarChar)
+    assertEquals('\0', statics.uninitializedVarChar)
```
sjrd: This is a progression.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2260,14 +2310,19 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
               "Invalid tree in JSDesugar.transformExpr() of class " +
               tree.getClass)
       }
+
+      if (preserveChar || tree.tpe != CharType)
+        baseResult
+      else
+        genCallHelper("bC", baseResult)
```
gzm0: Hum, doesn't this unbox too often? For example, if you have a `Block(..., x)` of type `CharType`. When you transform `x` you already unbox there (since you transform it with `preserveChar = false`, but then you also unbox the resulting block. I think you need to follow the expected types of the original trees :-/

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5594,8 +5576,11 @@ abstract class GenJSCode extends plugins.PluginComponent
   private lazy val HijackedNumberClasses =
     Seq(BoxedByteClass, BoxedShortClass, BoxedIntClass, BoxedLongClass,
         BoxedFloatClass, BoxedDoubleClass)
-  private lazy val HijackedBoxedClasses =
-    Seq(BoxedUnitClass, BoxedBooleanClass) ++ HijackedNumberClasses
+
+  private lazy val HijackedBoxedClasses = {
+    Seq(BoxedUnitClass, BoxedBooleanClass, BoxedCharacterClass) ++
+    HijackedNumberClasses
```
sjrd: This one contains scalac `Symbol`s.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Uh oh, indeed. The problem is not here, though. It's only that `Block`/`If`/etc. should never just forward `preserveChar`. They must use their `tpe` as expected type for the embedded expressions.

It's weird that nothing in the CI caught this. Probably because it's very hard or impossible to write user code that results in some IR that would expose this issue. I'll try to craft a failing test case.

sjrd:
Updated.

sjrd:
Rebased.

PR #3094, opened 2017-08-15
Remove reflective call classes and handle everything in the compiler.

sjrd:
The compiler already had to do some special-casing for primitives and hijacked classes in the codegen for reflective calls. For primitives, it relied on the presence of dedicated classes in the run-time library.

Now, the compiler directly generates code for what used to be the body of the target method in reflective call class. This allows to entirely remove the reflective call classes from the library.


test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/ReflectiveCallTest.scala
```diff
@@ -158,6 +158,56 @@ class ReflectiveCallTest {
...
+    assertTrue(fCompareToFloat(5.5f, 6.5f) < 0)
+
+    def fCompareToDouble(x: { def compareTo(y: java.lang.Double): Int }, y: Double): Int =
+      x.compareTo(y)
+    assertTrue(fCompareToDouble(5.5, 6.5) < 0)
+  }
+
+  @Test def should_work_with_concat_for_primitives(): Unit = {
+    assumeFalse("Reflective call prim.+(String) broken on the JVM",
+        Platform.executingInJVM)
```
sjrd: https://github.com/scala/bug/issues/10469

gzm0: Could you add a reference to the bug in the code, please?

sjrd: Done.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3766,69 +3806,100 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+                case CharKind     => false
+                case _:INT | LONG => true
+                case _            => false
+              }
+              val ignoreBecauseItMustBeAnInt = {
+                primitiveClass == DoubleClass &&
+                toTypeKind(methodInPrimClass.tpe.resultType) == DoubleKind &&
+                isIntOrLongKind(toTypeKind(sym.tpe.resultType))
+              }
+              if (ignoreBecauseItMustBeAnInt) {
```
gzm0: How does this work to call `toInt` reflectively on a (true) double?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3766,69 +3806,100 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        }
+
+        addCallForPrimitive(BooleanClass)
+        addCallForPrimitive(LongClass)
+        addCallForPrimitive(CharClass)
+
+        /* For primitive numeric types that box as JS numbers, find the first
+         * one that matches. It will be able to handle the subsequent cases.
+         */
+        Seq(DoubleClass, IntClass, FloatClass, ShortClass, ByteClass).find(
```
gzm0: Will this work properly with strict floats?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3766,69 +3806,100 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+            }
+          }
+        }
+
+        addCallForPrimitive(BooleanClass)
+        addCallForPrimitive(LongClass)
+        addCallForPrimitive(CharClass)
+
+        /* For primitive numeric types that box as JS numbers, find the first
+         * one that matches. It will be able to handle the subsequent cases.
```
gzm0: This matching logic is very hard to understand and verify it is correct. Have you considered handling the sets of relevant methods individually? There are not that many:

- Conversions `toX`
- Unary ops
- String concat
- Comparisons
- Binary number ops
- Binary int ops

I'm not sure that it is feasible. Maybe an acceptable compromise is to order the classes according to the runtime subtyping relationship and generate all cases. This would get rid of the strange `ignoreBecauseItMustBeAnInt` and probably be easier to understand :-/

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: In that case, `toTypeKind(methodInPrimClass.tpe.resultType)` is `IntKind`, so the second line of the condition is false.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Yes, even a strict `Float` is first matched as a `Double`, so just like for `Int`s, `Short`s and `Byte`s, most operations are handled by the `Double` case. `Int` has a few more cases to itself, typically bitwise operations. The only thing that's left for `Float`, `Short` and `Byte` is their specific `compareTo` method.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: `ignoreBecauseItMustBeAnInt` was always there. It handles the case
```scala
def foo(x: Any { def /(x: Int): Int }): Int = x / 2
foo(5)
```
Normally, `x` would be matched as a `Double`, because `Double` has a `def (x: Int)`. However, that would perform a double division and return 2.5 instead of the expected 2. Taking the result type into account allows us to patch that corner case. The same trick allows addition to wrap around 32-bits as it should. It's a very ad hoc case that has been there since https://github.com/scala-js/scala-js/commit/23f4ed3224517a30dea7345ea0fa78f82f83a8a5, i.e., the very day we decided not to box `Int`s anymore.

Generating all cases in order of subtyping check does *not* do the correct thing, for example for `compareTo(Object)`. Also we would still have the int vs double operation issue (e.g., 32-bit wrap around or integer division), only reversed. I don't think this would help in the slightest.

I don't quite understand what you mean by handling operations by their categories. From what I understand, it doesn't seem to simplify anything, only making the code longer by having to deal with the categories.

gzm0: Right, I see now, this didn't become any more convoluted than it already was.

PR #3098, opened 2017-08-17
Scala 2.12.3 and 2.13.0-M2

sjrd:


sjrd:
The failure is due to https://github.com/scala/bug/issues/10473

@gzm0 We have two options here:
* Either keep compiling and publishing our 2.12 artifacts with 2.12.2, which means changing back `Build.scalaVersionsUsedForPublishing` and `publish.sh` to use 2.12.2 instead of 2.12.3.
* Or filter out the binary incompatibility reported by MiMa: transitively, `RuntimeLong` still extend `java.io.Serializable` through `java.lang.Number`, so there is no hard incompatibility.

sjrd:
I have added a commit explicitly removing `with java.io.Serializable`, with a binary compat filter, to address this issue. It's unfortunate that we have to do this, but the end result is a little bit cleaner since it brings `RuntimeLong` more in sync with `java.io.Long`.

ci/checksizes.sh
```diff
@@ -11,13 +11,13 @@ case $FULLVER in
...
     ;;
-  2.12.2)
+  2.12.3)
     VER=2.12
     ;;
   2.13.0-M1)
     VER=2.13.0-M1
     ;;
-  2.10.3|2.10.4|2.10.5|2.10.6|2.11.0|2.11.1|2.11.2|2.11.4|2.11.5|2.11.6|2.11.7|2.11.8|2.12.0|2.12.1)
+  2.10.3|2.10.4|2.10.5|2.10.6|2.11.0|2.11.1|2.11.2|2.11.4|2.11.5|2.11.6|2.11.7|2.11.8|2.12.0|2.12.1|2.12.3)
```
gzm0: 2.12.2, not 2.12.3

sjrd: Duh, well spotted! Fixed.

Issue #2171, opened 2016-01-14
Sanitize state persistence in sbt plugin


gzm0:
Historically, we have used settings to persist things (notably the incremental optimizer/linker) between multiple sbt runs (but per scope).

However, this hack did not allow us to clean up once values got replaced (n.b. on reload).

We should fix this and ensure we do it properly.

gzm0:
I should take a look at this again, since with the advent of #2645 (improved test adapter) this will become even more relevant.

@eed3si9n You might be able to help me here: What we need is a clean way to create, use and clean up resources on a per-scope basis. We currently abuse settings and finalizers for this: We store the resource handle in a setting, so it is persistent across interactive invocations and we rely on its finalizer to clean up if a command reloads the setting (see [example][1]). This is of course sub-optimal since resources might be left dangling. Is there a better way to do this?

As a bonus point: We would like to emit statistics once an sbt execution (i.e. a full graph execution) has terminated. Is there a way to do that (same [example][1] shows our current hack to do this).

[1]: https://github.com/scala-js/scala-js/blob/9c35263b29ff84143180a1b8b5c241b60f3763d0/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L344-L351

sjrd:
For the bonus point, I think that's exactly what the `onComplete` setting provides. See in https://github.com/sbt/sbt/blob/1.0.x/main/src/main/scala/sbt/Keys.scala:
```scala
val onComplete = SettingKey[() => Unit]("on-complete", "Hook to run when task evaluation completes.  The type of this setting is subject to change, pending the resolution of SI-2915.", DSetting)
```

gzm0:
Assigning this to v1.0.0-M2 to remember that we should try extra hard to remove all internal keys. Candidates are:

- `scalaJSClearCacheStats`
- `scalaJSIRCacheHolder`

sjrd:
Fixed in 6bbb4ee076698beccf4575518c344e60058df7d9.

PR #3100, opened 2017-08-17
Directly expose everything useful as public in `ScalaJSPlugin`.

sjrd:
By "useful", we mean everything that is still public in `ScalaJSPluginInternal` in Scala.js 1.x. Exposing them directly in `ScalaJSPlugin` will allow us to not expose `ScalaJSPluginInternal` at all in Scala.js 1.x anymore.

There are two exceptions:
* the duo `scalaJSClearCacheStats` and `scalaJSIRCacheHolder`, because we still hope to be able to remove them (see #2171)
* `scalaJSPatchIncOptions`, which we'll silently not expose in 1.x anymore (we can reintroduce it later if asked for)

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -28,6 +28,9 @@ import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
 import org.scalajs.jsenv.phantomjs.PhantomJSEnv
 
 object ScalaJSPlugin extends AutoPlugin {
+  /** The global Scala.js IR cache */
+  val globalIRCache: IRFileCache = new IRFileCache()
+
   override def requires: Plugins = plugins.JvmPlugin
 
   /* The following module-case double definition is a workaround for a bug
@@ -69,7 +72,7 @@ object ScalaJSPlugin extends AutoPlugin {
       /* We take the semantics from the linker, since they depend on the stage.
        * This way we are sure we agree on the semantics with the linker.
        */
-      import ScalaJSPluginInternal.{scalaJSLinker, scalaJSRequestsDOMInternal}
+      import ScalaJSPluginInternal.scalaJSRequestsDOMInternal
       val semantics = scalaJSLinker.value.semantics
       val withDOM = scalaJSRequestsDOMInternal.value
       new RhinoJSEnv(semantics, withDOM, internal = ())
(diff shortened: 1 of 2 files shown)


sjrd:
retest this please

sjrd:
Rebased.

sjrd:
In principle I agree. It's however a trade-off with the following observation. Any named key is anyway part of the global namespace of keys for sbt at run-time (in the sbt console, for example). Therefore, they basically *are* public, from the perspective of compatibility guarantees as well as access rights. With that in mind, I think that all named keys should have a statically typed public definition available, so that they exist in .sbt files as well as in the console. sbt core seems to agree with that view: there's a *lot* of seemingly super-internal stuff in [`Keys.scala`](https://github.com/sbt/sbt/blob/1.x/main/src/main/scala/sbt/Keys.scala) that are public nevertheless (and `Keys._` is imported by default in .sbt files just as much as `autoImport._`).

Of course we could try to further reduce the set of named keys that we need to begin with. That's work I've trying to do a lot in previous PRs in the master branch already. For example 5c6e2925d273f7583581d6d39060375ecba71578. I don't think there is much left that we can remove, though. The only ones I would truly consider internal and that no one should "ever" touch would be `scalaJSClearCacheStats` and `scalaJSIRCacheHolder`. But so far I don't see how we can remove them altogether.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -433,13 +493,38 @@ object ScalaJSPlugin extends AutoPlugin {
...
+  def logIRCacheStats(logger: Logger): Unit =
+    logger.debug("Global IR cache stats: " + globalIRCache.stats.logLine)
+
+  /** Configures the given `IncOptions` for use with Scala.js.
+   *
+   *  This ensures that .sjsir files are pruned as needed in incremental
+   *  compiler runs.
+   */
+  def configureIncOptionsForScalaJS(incOptions: IncOptions): IncOptions =
+    SBTCompat.scalaJSPatchIncOptions(incOptions)
```
gzm0: Why does this need to be public?

gzm0:
> The only ones I would truly consider internal and that no one should "ever" touch would be `scalaJSClearCacheStats` and `scalaJSIRCacheHolder`.

I think there are actually hooks in sbt to do these kind of things. I wanted to use them at some point, but didn't get to it. I'd like to be able to remove these keys in the future, if we find a better solution.

So I agree that having the keys exposed in the sbt shell and the global namespace kind of requires us to also expose them in code. But maybe we should make it clear that it's just the name that is reserved and we reverse the right to remove the keys at any point in time.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: This is in case someone completely overrides their `incOptions` with `:=` instead of transforming it and preserving our external hooks. In that case they'll want to install our external hook on their brand new `IncOptions`.

That said, now that I think about it more, it's probably extremely niche. It's probably better to just not expose it now, and do so if someone ever asks for it.

sjrd:
If we're really strict about our compat guarantees, I don't think we'll be able to remove them. We could deprecate them and not use their values in any of our own tasks.

I think if there's hope to remove them, we should file an issue and make an extra effort to do so before 1.0.

gzm0:
We have #2171 for this. I added it to a v1 milestone.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Yeah, sounds more reasonable. Same for the settings below?

sjrd: You mean `baseConfigSettings` and such? Those are known to be useful to people. In particular `testConfigSettings` is even [documented](https://www.scala-js.org/doc/project/testing.html#integration-testing).

gzm0: Ok, fair enough. It's not like they'll go away anyways. So I'm perfectly fine with exposing them.

sjrd:
OK, so, I'll leave `scalaJSClearCacheStats` and `scalaJSIRCacheHolder` hidden in `ScalaJSPluginInternal`, shall I?

gzm0:
> OK, so, I'll leave scalaJSClearCacheStats and scalaJSIRCacheHolder hidden in ScalaJSPluginInternal, shall I?

Yes, I guess that's the most prudent choice.

sjrd:
Updated.

PR #3104, opened 2017-08-18
Implementation of SplittableRandom

andreaTP:
as promised here you have an initial implementation of SplittableRandom

### DIFF
--- a/javalib/src/main/scala/java/util/SplittableRandom.scala
+++ b/javalib/src/main/scala/java/util/SplittableRandom.scala
@@ -1,0 +1,120 @@
+package java.util
+
+/*
+ * This is a clean room implementation derived from the original paper
+ * and Java implementation mentioned there:
+ *
+ * Fast Splittable Pseudorandom Number Generators
+ * by Guy L. Steele Jr., Doug Lea, Christine H. Flood
+ * http://gee.cs.oswego.edu/dl/papers/oopsla14.pdf
+ *
+ */
+private object SplittableRandom {
+
+  private final val DoubleULP = 1.0 / (1L << 53)
+  private final val GoldenGamma = 0x9e3779b97f4a7c15L
+
+  private var defaultGen: Long = new Random().nextLong()
+
+  private def nextDefaultGen(): Long = {
+    val s = defaultGen
+    defaultGen = s + (2 * GoldenGamma)
+    s
+  }
+
+  // This function implements the original MurmurHash 3 finalizer
+  private final def mix64ForGamma(z: Long): Long = {
+    val z1 = (z ^ (z >>> 33)) * 0xff51afd7ed558ccdL
+    val z2 = (z1 ^ (z1 >>> 33)) * 0xc4ceb9fe1a85ec53L
+    z2 ^ (z2 >>> 33)
+  }
+
+  /*
+   * This function implements David Stafford's variant 4,
+   * while the paper version uses the original MurmurHash3 finalizer
+   * reference:
+   * http://zimbry.blogspot.pt/2011/09/better-bit-mixing-improving-on.html
+   */
+  private final def mix32(z: Long): Int = {
+    val z1 = (z ^ (z >>> 33)) * 0x62a9d9ed799705f5L
+    val z2 = (z1 ^ (z1 >>> 28)) * 0xcb24d0a5c88c35b3L
+    (z2 >>> 32).toInt
+  }
+
+  /*
+   * This function implements Stafford's variant 13,
+   * whereas the paper uses the original MurmurHash3 finalizer
+   */
+  private final def mix64(z: Long): Long = {
+    val z1 = (z ^ (z >>> 30)) * 0xbf58476d1ce4e5b9L
+    val z2 = (z1 ^ (z1 >>> 27)) * 0x94d049bb133111ebL
+    z2 ^ (z2 >>> 31)
+  }
+
+  private final def mixGamma(z: Long): Long = {
+    val z1 = mix64ForGamma(z) | 1L
+    val n = java.lang.Long.bitCount(z1 ^ (z1 >>> 1))
+    /* Reference implementation is wrong since we can read in the paper:
+     *
+     * ... Therefore we require that the number of such
+     * pairs, as computed by Long.bitCount(z ^ (z >>> 1)),
+     * exceed 24; if it does not, then the candidate z is replaced by
+     * the XOR of z and 0xaaaaaaaaaaaaaaaaL ...
+     * ... so the new value necessarily has more than 24 bit pairs whose bits differ
+     */
+    if (n <= 24) z1 ^ 0xaaaaaaaaaaaaaaaaL
+    else z1
+  }
+
+}
+
+final class SplittableRandom private (private var seed: Long, gamma: Long) {
+  import SplittableRandom._
+
+  def this(seed: Long) = {
+    this(seed, SplittableRandom.GoldenGamma)
+  }
+
+  private def this(ll: (Long, Long)) = this(ll._1, ll._2)
+
+  def this() = {
+    this({
+      val s = SplittableRandom.nextDefaultGen()
+
+      (SplittableRandom.mix64(s),
+          SplittableRandom.mixGamma(s + SplittableRandom.GoldenGamma))
+    })
+  }
+
+  def split(): SplittableRandom =
+    new SplittableRandom(mix64(nextSeed()), mixGamma(nextSeed()))
+
+  private def nextSeed(): Long = {
+    seed += gamma
+    seed
+  }
+
+  def nextInt(): Int = mix32(nextSeed())
+
+  //def nextInt(bound: Int): Int
+
+  //def nextInt(origin: Int, bound: Int): Int
+
+  def nextLong(): Long = mix64(nextSeed())
+
+  //def nextLong(bound: Long): Long
+
+  //def nextLong(origin: Long, bound: Long): Long
+
+  def nextDouble(): Double =
+    (nextLong() >>> 11).toDouble * DoubleULP
+
+  //def nextDouble(bound: Double): Double
+
+  //def nextDouble(origin: Double, bound: Double): Double
+
+  // this should be properly tested
+  // looks to work but just by chance maybe
+  def nextBoolean(): Boolean = nextInt < 0
+
+}
--- a/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
+++ b/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
@@ -1,0 +1,157 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013-2017, LAMP/EPFL   **
+**  __\ \/ /__/ __ |/ /__/ __ |/_// /_\ \    http://scala-js.org/       **
+** /____/\___/_/ |_/____/_/ | |__/ /____/                               **
+**                          |/____/                                     **
+\*                                                                      */
+package org.scalajs.testsuite.javalib.util
+
+import org.junit.Assert._
+import org.junit.Test
+
+import java.util.SplittableRandom
+
+import org.scalajs.testsuite.utils.AssertThrows._
+
+class SplittableRandomTest {
+
+  @Test def should_correctly_implement_nextLong(): Unit = {
+    val sr1 = new SplittableRandom(205620432625028L)
+    assertEquals(-546649510716590878L, sr1.nextLong())
+    assertEquals(5574037117696891406L, sr1.nextLong())
+    assertEquals(-2877648745898596966L, sr1.nextLong())
+    assertEquals(5734720902145206190L, sr1.nextLong())
+    assertEquals(1684781725002208217L, sr1.nextLong())
+    assertEquals(687902890032948154L, sr1.nextLong())
+    assertEquals(176280366443457561L, sr1.nextLong())
+    assertEquals(-2944062288620903198L, sr1.nextLong())
+    assertEquals(6872063775710978746L, sr1.nextLong())
+    assertEquals(-7374959378916621341L, sr1.nextLong())
+
+    val sr2 = new SplittableRandom(-7374959378916621341L)
+    assertEquals(3241340805431811560L, sr2.nextLong())
+    assertEquals(-2124831722811234979L, sr2.nextLong())
+    assertEquals(7339249063279462363L, sr2.nextLong())
+    assertEquals(1969867631102365324L, sr2.nextLong())
+    assertEquals(81632902222022867L, sr2.nextLong())
+    assertEquals(3451014011249622471L, sr2.nextLong())
+    assertEquals(-1727223780574897556L, sr2.nextLong())
+    assertEquals(-5128686556801302975L, sr2.nextLong())
+    assertEquals(-6412221907719417908L, sr2.nextLong())
+    assertEquals(-110482401893286265L, sr2.nextLong())
+  }
+
+  @Test def should_correctly_implement_nextInt(): Unit = {
+    val sr1 = new SplittableRandom(-84638)
+    assertEquals(962946964, sr1.nextInt())
+    assertEquals(1723227640, sr1.nextInt())
+    assertEquals(-621790539, sr1.nextInt())
+    assertEquals(-1848500421, sr1.nextInt())
+    assertEquals(-614898617, sr1.nextInt())
+    assertEquals(-628601850, sr1.nextInt())
+    assertEquals(-463597391, sr1.nextInt())
+    assertEquals(1874082924, sr1.nextInt())
+    assertEquals(-1206032701, sr1.nextInt())
+    assertEquals(1549874426, sr1.nextInt())
+
+    val sr2 = new SplittableRandom(1549874426)
+    assertEquals(-495782737, sr2.nextInt())
+    assertEquals(-1487672352, sr2.nextInt())
+    assertEquals(-538628223, sr2.nextInt())
+    assertEquals(1117712970, sr2.nextInt())
+    assertEquals(2081437683, sr2.nextInt())
+    assertEquals(2134440938, sr2.nextInt())
+    assertEquals(-2102672277, sr2.nextInt())
+    assertEquals(832521577, sr2.nextInt())
+    assertEquals(518494223, sr2.nextInt())
+    assertEquals(-42114979, sr2.nextInt())
+  }
+
+  @Test def should_correctly_implement_nextDouble(): Unit = {
+    val sr1 = new SplittableRandom(-45)
+    assertEquals(0.8229662358649753, sr1.nextDouble(), 0.0)
+    assertEquals(0.43324117570991283, sr1.nextDouble(), 0.0)
+    assertEquals(0.2639712712295723, sr1.nextDouble(), 0.0)
+    assertEquals(0.5576376282289696, sr1.nextDouble(), 0.0)
+    assertEquals(0.5505810186639037, sr1.nextDouble(), 0.0)
+    assertEquals(0.3944509738261206, sr1.nextDouble(), 0.0)
+    assertEquals(0.3108138671457821, sr1.nextDouble(), 0.0)
+    assertEquals(0.585951421265481, sr1.nextDouble(), 0.0)
+    assertEquals(0.2009547438834305, sr1.nextDouble(), 0.0)
+    assertEquals(0.8317691736686829, sr1.nextDouble(), 0.0)
+
+    val sr2 = new SplittableRandom(45)
+    assertEquals(0.9684135896502549, sr2.nextDouble(), 0.0)
+    assertEquals(0.9819686323309464, sr2.nextDouble(), 0.0)
+    assertEquals(0.5311927268453047, sr2.nextDouble(), 0.0)
+    assertEquals(0.8521356026917833, sr2.nextDouble(), 0.0)
+    assertEquals(0.01880601374789126, sr2.nextDouble(), 0.0)
+    assertEquals(0.37792881248018584, sr2.nextDouble(), 0.0)
+    assertEquals(0.7179744490511354, sr2.nextDouble(), 0.0)
+    assertEquals(0.3448879713662756, sr2.nextDouble(), 0.0)
+    assertEquals(0.023020123407108684, sr2.nextDouble(), 0.0)
+    assertEquals(0.6454709437764473, sr2.nextDouble(), 0.0)
+  }
+
+  @Test def should_correctly_implement_nextBoolean(): Unit = {
+    val sr1 = new SplittableRandom(4782934)
+    assertFalse(sr1.nextBoolean())
+    assertFalse(sr1.nextBoolean())
+    assertTrue(sr1.nextBoolean())
+    assertTrue(sr1.nextBoolean())
+    assertTrue(sr1.nextBoolean())
+    assertFalse(sr1.nextBoolean())
+    assertFalse(sr1.nextBoolean())
+    assertTrue(sr1.nextBoolean())
+    assertTrue(sr1.nextBoolean())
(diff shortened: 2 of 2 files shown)


sjrd:

sjrd:
Could you rebase and retarget this on the 0.6.x branch, please? (note that it's now possible to change the target branch of a PR; no need to close and open an different PR) Otherwise you won't get this in production until Scala.js 1.0.0 is out ;)

andreaTP:
done! aren't you becoming crazy with these two branch under active development? XD

sjrd:

andreaTP:
Sorry I do not see any evidence that it is failing due to my commit... 🙁

sjrd:
Looks pretty evident to me: https://scala-webapps.epfl.ch/jenkins/job/scalajs-task-worker/189601/consoleText

You need to put the test file in [the `require-jdk8` subdirectory](https://github.com/scala-js/scala-js/tree/master/test-suite/shared/src/test/require-jdk8).

andreaTP:
you are right, sorry, Jenkins is not really friendly from phone :-)
to me ready to re-test

sjrd:

sjrd:
Sorry I haven't reviewed this yet. I'll get to it by the end of the week.

andreaTP:
Thanks a lot for notifying! But feel free to keep this low priority ;-)

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * Fast Splittable Pseudorandom Number Generators
+ * by Guy L. Steele Jr., Doug Lea, Christine H. Flood
+ * http://gee.cs.oswego.edu/dl/papers/oopsla14.pdf
+ *
+ */
+private object SplittableRandom {
+
+  private final val DOUBLE_ULP = 1.0 / (1L << 53)
+  private final val GOLDEN_GAMMA = 0x9e3779b97f4a7c15L // odd
```
sjrd: I suggest we keep Scala conventions: `DoubleULP` and `GoldenGamma`.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  //switched from original comparing result of JVM std lib
+  private final def mix64variant13(_z: Long): Long = {
+  //def mix64(_z: Long): Long = {
```
sjrd: Calling this function `mix64variant13` is nonsense: it's the original MumurHash3 finalizer, not "variant 13". I suggest you call it `mix64ForGamma` instead, since it is the function used in `mixGamma`. The comment should say that it is in fact the original MurmurHash3 finalizer function, whereas the paper version uses variant 13 for this purpose.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+  // changed from original trying and comparing result on the JVM for all the numbers mentioned here:
+  // http://zimbry.blogspot.pt/2011/09/better-bit-mixing-improving-on.html
+  // reference was:
+  //def mix32(_z: Long): Int = {
+  //  var z = _z
+  //  z = (z ^ (z >>> 33)) * 0xff51afd7ed558ccdL
+  //  z = (z ^ (z >>> 33)) * 0xc4ceb9fe1a85ec53L
+  //  (z >>> 32).toInt
+  //}
+  // constant here comes from Mix04
```
sjrd: For multi-line comments, please use `/* */`, as shown here: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#non-scaladoc-comments

I don't think it's useful to have the complete reference in comments here. Instead, say that this function implements David Stafford's variant 4, while the paper version uses the original MurmurHash3 finalizer. The name can stay `mix32` IMO.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+  private final def mix32(_z: Long): Int = {
+    var z = _z
+    z = (z ^ (z >>> 33)) * 0x62a9d9ed799705f5L
+    z = (z ^ (z >>> 28)) * 0xcb24d0a5c88c35b3L
+    (z >>> 32).toInt
+  }
+
+  private final def mix64(_z: Long): Long = {
+  //def mix64variant13(_z: Long): Long = {
```
sjrd: Same as the others: the comment `//def mix64variant13 ...` is not useful or even disturbing. Instead, write as a comment above the method signature that this function Stafford's variant 13, whereas the paper uses the original MurmurHash3 finalizer for this purpose.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+    var z = _z
+    z = (z ^ (z >>> 30)) * 0xbf58476d1ce4e5b9L
+    z = (z ^ (z >>> 27)) * 0x94d049bb133111ebL
+    z ^ (z >>> 31)
+  }
+
+  private final def mix64variantMurmurHash3(_z: Long): Long = {
```
sjrd: This is dead code. It's not called anywhere.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+    z = (z ^ (z >>> 33)) * 0xff51afd7ed558ccdL
+    z = (z ^ (z >>> 33)) * 0xc4ceb9fe1a85ec53L
+    z ^ (z >>> 33)
+  }
+
+  private final def mixGamma(_z: Long) = {
+    var z = _z
+    // original: z = mix64variant13(z) | 1L
+    // changed from original version
+    z = mix64variant13(z) | 1L
```
sjrd: I don't see any difference between "original: ..." and the actual code. I think the comment is obsolete and should be removed.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+    val n = java.lang.Long.bitCount(z ^ (z >>> 1));
```
sjrd: Semicolon

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+    // Reference implementation is wrong since after they mention:
+    //
+    // ... Therefore we require that the number of such
+    // pairs, as computed by Long.bitCount(z ^ (z >>> 1)),
+    // exceed 24; if it does not, then the candidate z is replaced by
+    // the XOR of z and 0xaaaaaaaaaaaaaaaaL ...
+    //
+    // so we invert the condition:
+    //if (n >= 24) z ^= 0xaaaaaaaaaaaaaaaaL
+    if (n < 24) z ^= 0xaaaaaaaaaaaaaaaaL
```
sjrd: After much chewing on this, I agree that the original code seems wrong. However, I think the correct test should be `if (n <= 24)` and not `<`. The text you cite says "*exceeds* 24". If `n == 24`, it does not exceed 24, so the correction should be applied. Later in the text, we can also read

> so the new value necessarily has **more than** 24 bit pairs whose bits differ

(emphasis mine), which is also consistent with "it should be strictly greater than 24".

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+
+    z // This result is always odd.
+  }
+
+}
+
+final class SplittableRandom private (var seed: Long, gamma: Long) {
```
sjrd: `seed` should be private.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+  import SplittableRandom._
+
+  private lazy val defaultGen = new Random()
```
sjrd: This is currently unused in your code, but that's because the 0-argument constructor does not follow the original (see my other comment). Anyway, `defaultGen` is originally a *static* `AtomicLong`. It should therefore be a `private var defaultGen: Long` in the companion object. It should also be initialized with some "initial seed", which could be `new Random().nextLong()`, not just `new Random()`.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+  def this(seed: Long) = {
+    this(seed, SplittableRandom.GOLDEN_GAMMA)
+  }
+
+  private def this(ll: (Long, Long)) = this(ll._1, ll._2)
+
+  def this() = { // Preferred
+    this({
+      val s =
+        new Random().nextLong() + (2 * SplittableRandom.GOLDEN_GAMMA)
```
sjrd: This does not follow the original implementation. It should be:
```scala
val s = defaultGen
defaultGen = s + (2 * SplittableRandom.GOLDEN_GAMMA)
```

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+
+        (SplittableRandom.mix64(s),
+        SplittableRandom.mixGamma(s + SplittableRandom.GOLDEN_GAMMA))
```
sjrd: To avoid the tuple creation, I suggest a different scheme:
```scala
object SplittableRandom {
  ...
  private var defaultGen: Long = new Random().nextLong()

  private def nextDefaultGen(): Long = {
    val s = defaultGen
    defaultGen = s + (2 * GOLDEN_GAMMA)
    s
  }
}

private def this(s: Long, dummy: Boolean) =
  this(mix64(s), mixGamma(s + GOLDEN_GAMMA)

private def this() = this(SplittableRandom.nextDefaultGen())
```

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+    })
+  }
+
+  def split(): SplittableRandom = {
+    new SplittableRandom(
+      mix64(nextSeed()),
+      mixGamma(nextSeed())
+    )
```
sjrd: This should fit on one line.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private def nextSeed() = {
```
sjrd: I'd prefer if you put explicit result type to every method, including private ones. The precise type of things is quite important in this kind of code ;)

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
```diff
@@ -0,0 +1,130 @@
...
+    assertTrue(sr2.nextBoolean())
+    assertTrue(sr2.nextBoolean())
+  }
+
+  @Test def should_correctly_implement_split(): Unit = {
+    val sr1 = new SplittableRandom(205620432625028L).split()
+    assertEquals(-2051870635339219700L, sr1.nextLong())
+    assertEquals(-1989219516, sr1.nextInt())
+    assertEquals(0.5689791577483271, sr1.nextDouble(), 0.0)
+  }
```
sjrd: `split()` seriously needs more testing. You only test 1 `split()` for 1 `seed`, and you do not test that the original instance of `SplittableRandom` (the receiver of `split()`) keeps working as expected after the split.

I'd like to see:
* Several input seeds.
* Testing the original `SplittableRandom` as much as the splitted one.
* One case with 2 levels of splits, where the 4 `SplittableRandom`s that are the result of 3 `split()`s in a "binary tree" keep working as expected.

Also, to test that they "work as expected" after the split, a better way is to test that 2 consecutive `nextLong()`s return the expected results. Indeed, as the paper explains on page 465, second column, the results of 2 consecutive `nextLong()`s provide complete knowledge about the internal state of the `SplittableRandom`. So if 2 consecutive `nextLong()`s succeed after the split, we know that `split()` correctly updated the internal state. Testing `nextInt()` and `nextDouble()` is not useful here.

javalib/src/main/scala/java/util/SplittableRandom.scala
andreaTP: ok, on all the rest, but here I prefer to box the S into a private case class instead if you don't mind

andreaTP:
ok, I think this is ready for the next round (will squash at the end)

andreaTP:
please note that first implementation tried to be closer to original paper implementation, although I think is perfectly ok to align the code style to the rest of the codebase!

javalib/src/main/scala/java/util/SplittableRandom.scala
sjrd: That kind of defeats the purpose of not allocating the tuple, since now you're allocating a `GenLong` instead of a `Tuple2` ;)

andreaTP: I do not really like the second dummy Boolean argument TBH, any other proposal will be much appreciated ...

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  case class GenLong(value: Long)
+
+  private final def mix64ForGamma(_z: Long): Long = {
+    var z = _z
+    z = (z ^ (z >>> 33)) * 0xff51afd7ed558ccdL
+    z = (z ^ (z >>> 33)) * 0xc4ceb9fe1a85ec53L
+    z ^ (z >>> 33)
+  }
```
sjrd: For (way) better performance, use `val`s with a different one at each line, i.e.,
```scala
  private final def mix64ForGamma(z: Long): Long = {
    val z1 = (z ^ (z >>> 33)) * 0xff51afd7ed558ccdL
    val z2 = (z1 ^ (z1 >>> 33)) * 0xc4ceb9fe1a85ec53L
    z2 ^ (z2 >>> 33)
  }
```
The reason is that, in Scala.js, `Long`s are reasonably fast only if the optimizer can crazy-optimize them. And the optimizer is not very good with `var`s; it is much better at dealing with `val`s.

The same comments apply to all 4 `mixXyz` methods (`mix64ForGamma`, `mix32`, `mix64` and `mixGamma`).

For `mix64ForGamma` only: it's also worth adding `@inline` on it, since it has only one call site. This will avoid boxing the parameter and result `Long`s.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,123 @@
...
+   * whereas the paper uses the original MurmurHash3 finalizer
+   */
+  private final def mix64(_z: Long): Long = {
+    var z = _z
+    z = (z ^ (z >>> 30)) * 0xbf58476d1ce4e5b9L
+    z = (z ^ (z >>> 27)) * 0x94d049bb133111ebL
+    z ^ (z >>> 31)
+  }
+
+  private final def mixGamma(_z: Long) = {
```
sjrd: The explicit result type is missing, here.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
```diff
@@ -0,0 +1,157 @@
...
+    assertEquals(274792684182118046L, sr6.nextLong())
+    assertEquals(683259797650761389L, sr6.nextLong())
+
+    val sr7 = sr6.split()
+    assertEquals(1682793527903105269L, sr7.nextLong())
+    assertEquals(2140483520539013019L, sr7.nextLong())
+
+    val sr8 = sr6.split()
+    assertEquals(-7468768144124082123L, sr8.nextLong())
+    assertEquals(6163667569279435512L, sr8.nextLong())
```
sjrd: Awesome :)

javalib/src/main/scala/java/util/SplittableRandom.scala
sjrd: I don't think there is any other way.

Then I'd prefer you go back to your first solution with the tuple. It was better.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
andreaTP: I was surprised too to see all tests passing without any problem here ^^

javalib/src/main/scala/java/util/SplittableRandom.scala
andreaTP: ok

andreaTP:
ready for next review

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  private var defaultGen: Long = new Random().nextLong()
+
+  private def nextDefaultGen(): Long = {
+    val s = defaultGen
+    defaultGen = s + (2 * GoldenGamma)
+    s
+  }
+
+  case class GenLong(value: Long)
```
sjrd: This is now dead code and should be removed.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  private final def mix64ForGamma(z: Long): Long = {
```
sjrd: This method is missing its comment saying what it implements. IIRC that's the original MurmurHash 3 finalizer.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+   * http://zimbry.blogspot.pt/2011/09/better-bit-mixing-improving-on.html
+   */
+  private final def mix32(z: Long): Int = {
+    val z1 = (z ^ (z >>> 33)) * 0x62a9d9ed799705f5L
+    val z2 = (z1 ^ (z1 >>> 28)) * 0xcb24d0a5c88c35b3L
+    (z2 >>> 32).toInt
+  }
+
+  /*
+   * function implements Stafford's variant 13,
```
sjrd: I think this comment is missing a "This" at the beginning of the sentence, if I compare it to the one for `mix32`.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+    val z1 = mix64ForGamma(z) | 1L
+    val n = java.lang.Long.bitCount(z1 ^ (z1 >>> 1))
+    /* Reference implementation is wrong since after they mention:
+     *
+     * ... Therefore we require that the number of such
+     * pairs, as computed by Long.bitCount(z ^ (z >>> 1)),
+     * exceed 24; if it does not, then the candidate z is replaced by
+     * the XOR of z and 0xaaaaaaaaaaaaaaaaL ...
+     * ... so the new value necessarily has more than 24 bit pairs whose bits differ
+     *
```
sjrd: Spurious "blank" line which can be removed.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+     */
+    if (n <= 24) z1 ^ 0xaaaaaaaaaaaaaaaaL
+    else z1 // This result is always odd.
```
sjrd: The position of the comment is confusing; it does not make it clear that `z1 ^ 0xaaaaaaaaaaaaaaaaL` is also always odd. I suggest removing that comment, as it is not useful to the reader of the code, IMO (it is useful to the reader of the paper, but that's another story).

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * Fast Splittable Pseudorandom Number Generators
+ * by Guy L. Steele Jr., Doug Lea, Christine H. Flood
+ * http://gee.cs.oswego.edu/dl/papers/oopsla14.pdf
+ *
+ */
+private object SplittableRandom {
+
+  private final val DoubleULP = 1.0 / (1L << 53)
+  private final val GoldenGamma = 0x9e3779b97f4a7c15L // odd
```
sjrd: Similarly to the other comment about "odd", I would remove this comment. Without reading the paper, it is confusing as it reads more as "weird" than "not even".

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+  private final def mix64(z: Long): Long = {
+    val z1 = (z ^ (z >>> 30)) * 0xbf58476d1ce4e5b9L
+    val z2 = (z1 ^ (z1 >>> 27)) * 0x94d049bb133111ebL
+    z2 ^ (z2 >>> 31)
+  }
+
+  private final def mixGamma(z: Long): Long = {
+    val z1 = mix64ForGamma(z) | 1L
+    val n = java.lang.Long.bitCount(z1 ^ (z1 >>> 1))
+    /* Reference implementation is wrong since after they mention:
```
sjrd: s/after/in the paper/ ?
I think that would be clearer. "after" only makes sense if we're already following the paper while reading the code.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+final class SplittableRandom private (private var seed: Long, gamma: Long) {
+  import SplittableRandom._
+
+  def this(seed: Long) = {
+    this(seed, SplittableRandom.GoldenGamma)
+  }
+
+  private def this(ll: (Long, Long)) = this(ll._1, ll._2)
+
+  def this() = { // Preferred
```
sjrd: This comment is confusing. I suggested removing it.

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+    this({
+      val s = SplittableRandom.nextDefaultGen()
+
+      (SplittableRandom.mix64(s),
+        SplittableRandom.mixGamma(s + SplittableRandom.GoldenGamma))
```
sjrd: This line should have 2 more spaces of indentation (to be indented 4 spaces wrt. `(SplittableRandom.mix64(s),`).

javalib/src/main/scala/java/util/SplittableRandom.scala
```diff
@@ -0,0 +1,122 @@
...
+  //def nextLong(bound: Long): Long
+
+  //def nextLong(origin: Long, bound: Long): Long
+
+  def nextDouble(): Double =
+    (nextLong() >>> 11).toDouble * DoubleULP
+
+  //def nextDouble(bound: Double): Double
+
+  //def nextDouble(origin: Double, bound: Double): Double =
```
sjrd: Nitpick: remove ` =` at the end of the line to be consistent with the other non-implemented methods.

test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/SplittableRandom.scala
```diff
@@ -0,0 +1,157 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___      __ ____  Scala.js Test Suite        **
+**    / __/ __// _ | / /  / _ | __ / // __/  (c) 2013, LAMP/EPFL        **
```
sjrd: 2013-2017

sjrd:
Btw, when you squash, could you also rebase once more on top of the latest 0.6.x, please? We've gone through some tricky merges since this PR was opened (see https://github.com/scala-js/scala-js/pull/3109), and it'd be nice if the history was not made even more obscure with a merge branch that jumps over it all :)

andreaTP:
ok, ready for final check

sjrd:

andreaTP:
Great thanks!

PR #3105, opened 2017-08-19
[no-master] Deprecations to prepare for 1.x.

sjrd:


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -74,12 +74,22 @@ object ScalaJSPluginInternal {
   val scalaJSIRCache = TaskKey[globalIRCache.Cache]("scalaJSIRCache",
       "Scala.js internal: Task to access a cache.", KeyRanks.Invisible)
 
-  /** Internal task to calculate whether a project requests the DOM
-   *  (through jsDependencies or requiresDOM) */
-  val scalaJSRequestsDOM = TaskKey[Boolean]("scalaJSRequestsDOM",
+  /** Non-deprecated alias of `scalaJSRequestsDOM` for internal use. */
+  val scalaJSRequestsDOMInternal = TaskKey[Boolean]("scalaJSRequestsDOM",
```
sjrd: This must be `private[sbtplugin]`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/AbstractJSDeps.scala
```diff
@@ -75,6 +75,11 @@ object ProvidedJSModuleID {
     ProvidedJSModuleID(new JSDependency(name, Nil), configurations)
 }
 
+/** Deprecated.
+ *
+ *  For internal reasons, this class is not marked as `@deprecated`, but should
```
gzm0: Same here: What are "internal reasons". Please document (internally). Here it seems OK that the users don't get a warning, since this is probably not really used externally.

library/src/main/scala/scala/scalajs/js/JSApp.scala
```diff
@@ -2,34 +2,73 @@ package scala.scalajs.js
...
+ *  }
+ *  }}}
+ *
+ *  Alternatively, you can define it as
+ *  {{{
+ *  package bar
+ *
+ *  @JSExportTopLevel("bar.Foo")
+ *  object Foo {
+ *    @JSExport
```
gzm0: Why not recommend to just top level export the method? Seems more JS idiomatic in this case.

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
```diff
@@ -42,6 +42,10 @@ protected[testinterface] object HTMLRunner extends js.JSApp {
     }
   }
 
+  @JSExportTopLevel("org.scalajs.testinterface.HTMLRunner")
+  protected def getInstance(): this.type = this
+
+  @JSExport
```
gzm0: Replace this by a single top level main? We can do this safely, since we always pull `testinterface` to the newest version. So it always matches the sbt plugin version.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/AbstractJSDeps.scala
sjrd: Same reason as in the other PR.

library/src/main/scala/scala/scalajs/js/JSApp.scala
sjrd: Because JS is untyped. No compiler will be there to help the developer while they're migrating their codebase, when they are frustrated that they have to address deprecation warnings. This variant lets their migration in Scala.js be 100% compatible for their JS.

The other variant is there for when they have time to *improve* their codebase after the migration, and whether have time to look up their users from JS and ensure everything's fine.

test-interface/src/main/scala/org/scalajs/testinterface/HTMLRunner.scala
sjrd: I was kind of dogfooding my own migration advice from the Scaladoc. But if you prefer I can change it. Anyway now I've seen that it works ;)

library/src/main/scala/scala/scalajs/js/JSApp.scala
sjrd: Ah I completely misunderstood your point, didn't I? You mean just use `@JSExportTopLevel def`, right? Indeed that would be much better.

gzm0: Yes :)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/AbstractJSDeps.scala
gzm0: Ok.

sjrd:
Updated.

PR #3106, opened 2017-08-20
Introduce better names for `OutputMode`s in the public API.

sjrd:
The "Isolated" in `ECMAScript51Isolated` does not mean anything to external users. It really should be `ECMAScript51` from a user point of view.

ECMAScript 2015 is the real, official name of ES 6. Therefore we introduce `ECMAScript2015` as replacement for `ECMAScript6`.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/OutputMode.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/OutputMode.scala
@@ -36,32 +36,37 @@ object OutputMode {
     val esLevel: ESLevel = ESLevel.ES5
   }
 
-  /** Modern output mode compliant with ECMAScript 5.1 in a function scope.
-   *  This is the default output mode used by fastOpt and fullOpt.
-   *  The output must be enclosed in an anonymous function isolating the code
-   *  in a dedicated scope.
+  /** Output mode compliant with ECMAScript 5.1 (deprecated alias).
+   *
+   *  This value is not annotated with `@deprecated` for technical reasons, but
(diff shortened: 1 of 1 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/OutputMode.scala
```diff
@@ -36,33 +36,38 @@ object OutputMode {
     val esLevel: ESLevel = ESLevel.ES5
   }
 
-  /** Modern output mode compliant with ECMAScript 5.1 in a function scope.
-   *  This is the default output mode used by fastOpt and fullOpt.
-   *  The output must be enclosed in an anonymous function isolating the code
-   *  in a dedicated scope.
+  /** Output mode compliant with ECMAScript 5.1 (deprecated alias).
+   *
+   *  This value is not annotated with `@deprecated` for technical reasons, but
```
gzm0: What are the "technical reasons"? Please document them (internally is fine). Also, how are users supposed to notice that we deprecated this?

sjrd: The reason is that one cannot deprecate a `class` or `object` for the purpose of *constructing* it (which is the intended usage for users) without deprecating pattern matching on it or more generally *deconstructing* it, which we internaly need to keep supporting the feature. At least not in a binary compatible way.

For this class I'm not worried that users actually ever wrote it anywhere, because the only documented/advertised way to create it is via the `RuntimeDOM` DSL. So getting the word out is not an issue.

gzm0: Fair enough

PR #3114, opened 2017-08-29
Fix #2171: Sanitize state persistence in the sbt plugin.

### LINKED ISSUE #2171: Sanitize state persistence in sbt plugin

sjrd:
* All the IR caches are explicitly freed when the build is   unloaded.
* The global IR cache stats are cleared every time a sequence of   tasks end.

This removes the need for `scalaJSClearCacheStats` and `scalaJSIRCacheHolder`, although the latter is still needed for binary compatibility in the 0.6.x series since `scalaJSIRCache` needs to stay a task. In 1.x we will be able to turn it into a setting, at which point `scalaJSIRCacheHolder` will be unnecessary.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -28,9 +28,6 @@ import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
 import org.scalajs.jsenv.phantomjs.PhantomJSEnv
 
 object ScalaJSPlugin extends AutoPlugin {
-  /** The global Scala.js IR cache */
-  val globalIRCache: IRFileCache = new IRFileCache()
-
   override def requires: Plugins = plugins.JvmPlugin
 
   /* The following module-case double definition is a workaround for a bug
@@ -227,7 +224,9 @@ object ScalaJSPlugin extends AutoPlugin {
         "Do not set the value of this setting (only use it as read-only).",
         BSetting)
 
-    val scalaJSIRCache = TaskKey[globalIRCache.Cache]("scalaJSIRCache",
+    // This is lazy to avoid initialization order issues
+    lazy val scalaJSIRCache = TaskKey[ScalaJSPluginInternal.globalIRCache.Cache](
+        "scalaJSIRCache",
         "Scala.js internal: Task to access a cache.", KeyRanks.Invisible)
 
     /** Persisted instance of the Scala.js linker.
@@ -504,14 +503,18 @@ object ScalaJSPlugin extends AutoPlugin {
   )
 
   /** Logs the current statistics about the global IR cache. */
-  def logIRCacheStats(logger: Logger): Unit =
+  def logIRCacheStats(logger: Logger): Unit = {
+    import ScalaJSPluginInternal.globalIRCache
     logger.debug("Global IR cache stats: " + globalIRCache.stats.logLine)
+  }
 
   override def globalSettings: Seq[Setting[_]] = {
     Seq(
         scalaJSStage := Stage.FastOpt,
         scalaJSUseRhinoInternal := false,
-        ScalaJSPluginInternal.scalaJSClearCacheStats := globalIRCache.clearStats()
+
+        ScalaJSPluginInternal.scalaJSClearCacheStats :=
+          ScalaJSPluginInternal.globalIRCache.clearStats()
     )
   }
 
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -10,6 +10,7 @@
 package org.scalajs.sbtplugin
 
 import sbt._
+import sbt.Keys._
 
 import org.scalajs.core.tools.sem.Semantics
 import org.scalajs.core.tools.io._
@@ -513,8 +514,29 @@ object ScalaJSPlugin extends AutoPlugin {
         scalaJSStage := Stage.FastOpt,
         scalaJSUseRhinoInternal := false,
 
-        ScalaJSPluginInternal.scalaJSClearCacheStats :=
-          ScalaJSPluginInternal.globalIRCache.clearStats()
+        ScalaJSPluginInternal.scalaJSClearCacheStatsInternal := {},
+
+        // Clear the IR cache stats every time a sequence of tasks ends
+        onComplete := {
+          val prev = onComplete.value
+
+          { () =>
+            prev()
+            ScalaJSPluginInternal.globalIRCache.clearStats()
+          }
+        },
+
+        /* When unloading the build, free all the IR caches.
+         * Note that this runs on `reload`s, for example, but not when we
+         * *exit* sbt. That is fine, though, since in that case the process
+         * is killed altogether.
+         */
+        onUnload := {
+          onUnload.value.andThen { state =>
+            ScalaJSPluginInternal.freeAllIRCaches()
+            state
+          }
+        }
     )
   }
 
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -1,6 +1,9 @@
 package org.scalajs.sbtplugin
 
+import scala.annotation.tailrec
+
 import java.util.IllegalFormatException
+import java.util.concurrent.atomic.AtomicReference
 
 import sbt._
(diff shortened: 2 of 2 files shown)


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -511,7 +552,29 @@ object ScalaJSPlugin extends AutoPlugin {
...
-        ScalaJSPluginInternal.scalaJSClearCacheStats := globalIRCache.clearStats()
+        ScalaJSPluginInternal.scalaJSClearCacheStatsInternal := {},
+
+        // Clear the IR cache stats every time a sequence of tasks ends
+        onComplete := {
+          val prev = onComplete.value
+
+          { () =>
+            prev()
+            globalIRCache.clearStats()
```
gzm0: We should also report the stats here, rather than each time we are kind of done with the cache. This will also allow us to make the `logIRCacheStats` private.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -511,7 +552,29 @@ object ScalaJSPlugin extends AutoPlugin {
...
         scalaJSStage := Stage.FastOpt,
         scalaJSUseRhinoInternal := false,
-        ScalaJSPluginInternal.scalaJSClearCacheStats := globalIRCache.clearStats()
+        ScalaJSPluginInternal.scalaJSClearCacheStatsInternal := {},
+
+        // Clear the IR cache stats every time a sequence of tasks ends
+        onComplete := {
+          val prev = onComplete.value
+
+          { () =>
```
gzm0: Why not use the `andThen` thingy here?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -28,9 +33,45 @@ import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
...
 
+  private val allocatedIRCaches =
+    new AtomicReference[List[globalIRCache.Cache]](Nil)
+
+  /** Allocates a new IR cache linked to the [[globalIRCache]].
+   *
+   *  The allocated IR cache will automatically be freed when the build is
+   *  unloaded.
+   */
+  def newIRCache: globalIRCache.Cache = {
```
gzm0: When should one use `scalaJSIRCache`, when this method?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -28,9 +33,45 @@ import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
...
+  /** The global Scala.js IR cache.
+   *
+   *  If you want to allocate a new cache for your purposes, prefer calling
+   *  [[newIRCache]] rather than
+   *  [[org.scalajs.core.tools.io.IRFileCache.newCache globalIRCache.newCache]].
+   *  The former will automatically register the new cache to be freed when the
+   *  build is unloaded. If you use the latter, it is your responsibility to
+   *  free the cache.
+   */
   val globalIRCache: IRFileCache = new IRFileCache()
```
gzm0: We should consider hiding this. Do we know of any use case that needs it? Easier to open than close later.

sjrd: I don't know any use case, at the moment. Except that it appears in the path dependent type of `scalaJSIRCache`, which is a `TaskKey[globalIRCache.Cache]`. In 0.6.x that has to stay, but in 1.x we could potentially use `IRFileCache#Cache` instead.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: Unless one knows what they're doing, they probably don't want to use `scalaJSIRCache`. Either they want to read `scalaJSIR.value` which already uses `scalaJSIRCache.value`, or they want their own `Cache`. It is unadvisable to use the same `Cache` for two different purposes, since the calls to `cached(files1)` and `cached(files2)` will interfere with each other.

So unless you want the precise set of files returned by `scalaJSIR.value`, you probably want to use `newIRCache`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: Because

> `andThen` is not a member of `() => Unit`

(`andThen` is only defined on `Function1`, but `onComplete` is a `Function0`)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: I tried, but here we only have access to `sLog` (instead of `streams`) because we're in a setting. And things sent to the `sLog` are (apparently) not displayed by `last`. So any `sLog.value.debug(...)` is lost and can never be observed. (a `.info(...)` is printed on the console, but not stored for further redisplay by `last` either)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: FTR, neither GitHub nor Google can find `globalIRCache` anywhere but in our own codebase.

gzm0: In that case, I propose we deprecate it all in 0.6.x and hide it entirely in 1.0.0 unless someone complains. This also solves the problem that it would be one of the few places where the sbt plugin explicitly exposes something from `linker.standard._` (we haven't yet published a version with this, right?).
We can put a comment on `scalaJSIR` that it really is preferable because it is cached :)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Ugh :( I was really hoping we could get rid of the ugly "invoke this whenever you have done something that has used the cache" :(
Let's leave it then for now.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Maybe we should just hide this for now?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: > (we haven't yet published a version with this, right?).

That's right.

We can make it private in 1.0.0, but we can't deprecate it in 0.6.x for the same reason as above: it is part of the path-dependent type of something that is not deprecated. We can hide it back in `ScalaJSPluginInternal` in 0.6.x, though, since it was still there in the latest released version.

sjrd: > This also solves the problem that it would be one of the few places where the sbt plugin explicitly exposes something from linker.standard._

`IRFileCache` is in `org.scalajs.core.tools.io`, not in `standard._`. It has to be, since it is virtually necessary to create the correct type of input to `Linker.link`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: OK

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
gzm0: Right... Shouldn't review code when I'm too tired :P

So I guess we hide it back into internal now and make it private in 1.0.0. I'd really rather just have `scalaJSIR` for people to use and expose this only if someone needs it.

sjrd: OK

sjrd:
Updated.

PR #3118, opened 2017-08-30
Fix #3076: Validate inheritance relationships in the Analyzer.

### LINKED ISSUE #3076: The linker must check the validity of inheritance relationships

sjrd:


sjrd:
This will need some more manual testing to make sure that errors are indeed reported when necessary, but more importantly that the rest of the Analyzer survives when errors are detected.

So I'm putting this in Do Not Merge mode for now.

sjrd:
We most definitely should. I'll try to set that up for this PR.

sjrd:
Updated with tests :)

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+  private def assertHasErrors(analysis: Analysis,
+      expectedErrors: Error*): Unit = {
+    val actualErrors = analysis.errors
+
+    for (expectedError <- expectedErrors) {
+      assertTrue(s"Missing expected error: $expectedError",
+          actualErrors.contains(expectedError))
+    }
+
+    if (actualErrors.size != expectedErrors.size) {
```
gzm0: Hum, so if we have errors `[1,2]` and expect errors `[3,4]` we are not going to report the unexpected errors, but if we expect errors `[3,4,5]`, we do? Does not make sense to me. I recommend just dropping this if.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  @Test
+  def invalidJavaLangObject(): Unit = {
+    val invalidJLObjectDefs = Seq(
+      ClassDef(Ident(ObjectClass), ClassKind.Class, Some(Ident("Lparent")),
+          Nil, None, Nil, Nil)(emptyOptHints),
+      ClassDef(Ident(ObjectClass), ClassKind.ModuleClass, None, Nil, None,
+          Nil, Nil)(emptyOptHints),
+      ClassDef(Ident(ObjectClass), ClassKind.Class, None,
+          List(Ident("Lparent")), None, Nil, Nil)(emptyOptHints)
```
gzm0: Could you add comments on what is wrong with them and/or add named parameters for the things you set?

The named parameter thing applies in general. It is kind of hard to understand what all the `Nil` and `None` are. Maybe define a helper to create `ClassDef` that has all these optional?

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+
+    if (actualErrors.size != expectedErrors.size) {
+      for (actualError <- actualErrors) {
+        assertTrue(s"Unexpected error: $actualError",
+            expectedErrors.contains(actualError))
+      }
+    }
+  }
+
+  private def assertHasError(msg: String, analysis: Analysis)(
```
gzm0: Consider calling this `assertContainsError` to make it more explicit that it need not be exhaustive.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    Analyzer.computeReachability(CommonPhaseConfig(), symbolRequirements,
+        allowAddingSyntheticMethods = true, inputProvider)
+  }
+
+  private def assertNoError(analysis: Analysis): Unit =
+    assertHasErrors(analysis)
+
+  private def assertHasErrors(analysis: Analysis,
```
gzm0: Consider calling this `assertErrors` or `assertErrorsAre` to make it clearer it is exhaustive.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+            Nil, Nil)(emptyOptHints),
+        ClassDef(Ident("LB"), ClassKind.Class, Some(Ident("LA")), Nil, None,
+            Nil, Nil)(emptyOptHints)
+    )
+
+    val analysis = computeAnalysis(classDefs, reqsFactory.classData("LA"))
+
+    assertHasError("CycleInInheritanceChain(LA, LB)", analysis) {
+      case CycleInInheritanceChain(List(a, b), `fromUnitTest`) =>
+        a.encodedName == "LA" && b.encodedName == "LB"
```
gzm0: Double checking: We are OK with imposing the starting point of the inheritance chain in the test here, since we require A to be present and not B (also below).

Also: Consider creating a matcher for `encodedName`.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -0,0 +1,240 @@
...
+            List(Ident("LI1"), Ident("LI2")), None, List(trivialCtor("LA")),
+            Nil)(emptyOptHints)
+    )
+
+    val analysis = computeAnalysis(classDefs,
+        reqsFactory.instantiateClass("LA", "init___") ++
+        reqsFactory.callMethod("LA", "foo__V"))
+
+    assertHasError("ConflictingDefaultMethods(LI1.foo__V, LI2.foo__V)", analysis) {
+      case ConflictingDefaultMethods(List(a, b), `fromAnalyzer`) =>
```
gzm0: Do we really want to impose the order of a, b in the test here?

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -127,6 +127,128 @@ class AnalyzerTest {
...
+          Seq(ClassKind.Interface, ClassKind.ModuleClass, ClassKind.JSClass,
+              ClassKind.NativeJSClass)
+        case ClassKind.Interface =>
+          Seq(ClassKind.Class, ClassKind.Interface)
+        case ClassKind.JSClass | ClassKind.JSModuleClass |
+            ClassKind.NativeJSClass | ClassKind.NativeJSModuleClass |
+            ClassKind.AbstractJSType =>
+          Seq(ClassKind.Class, ClassKind.Interface, ClassKind.AbstractJSType,
+              ClassKind.JSModuleClass)
+      }
```
gzm0: Consider a function, so you can write the two for loops in one.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -127,6 +127,128 @@ class AnalyzerTest {
...
+        ClassKind.HijackedClass,
+        ClassKind.Interface,
+        ClassKind.JSClass,
+        ClassKind.JSModuleClass,
+        ClassKind.NativeJSClass,
+        ClassKind.NativeJSModuleClass,
+        ClassKind.AbstractJSType
+    )
+
+    for (kindA <- kindsA) {
```
gzm0: Consider `kindSub` and `kindBase` for clarity.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -127,6 +127,128 @@ class AnalyzerTest {
...
```
gzm0: Consider `kindCls` and `kindIntf` for clarity.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -127,6 +127,128 @@ class AnalyzerTest {
...
+            ClassKind.Interface =>
+          Seq(ClassKind.Class, ClassKind.ModuleClass, ClassKind.JSClass,
+              ClassKind.NativeJSClass, ClassKind.AbstractJSType)
+        case ClassKind.JSClass | ClassKind.JSModuleClass |
+            ClassKind.NativeJSClass | ClassKind.NativeJSModuleClass |
+            ClassKind.AbstractJSType =>
+          Seq(ClassKind.Class, ClassKind.ModuleClass, ClassKind.HijackedClass,
+              ClassKind.Interface, ClassKind.JSClass, ClassKind.JSModuleClass,
+              ClassKind.NativeJSClass, ClassKind.NativeJSModuleClass)
+      }
```
gzm0: Function & merge for?

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
```diff
@@ -183,6 +305,18 @@ class AnalyzerTest {
     }
   }
 
+  private def validParentForKind(kind: ClassKind): Option[Ident] = {
+    kind match {
```
gzm0: `import ClassKind._`?

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
sjrd: > Double checking

Yes, it's deterministic: the class that is required "from the outside" always appears first in the cycle.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
sjrd: Not necessarily, but the analyzer is currently completely deterministic, so this won't spuriously fail. Should it change, we can revisit this test case to make it more permissive.

Alternatively, we could define `ConflictingDefaultMethods` with a `Set`.

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
sjrd: `assertTrue` throws an exception, so as soon as there is a missing expected error, control flow does not even reach this `if` condition. We reach here only if all expected errors are found.

So no if we expect `[3,4,5]` you still don't see the unexpected errors. All you see is that `3` is missing.

gzm0: Ah, ah :) Right :P

tools/shared/src/test/scala/org/scalajs/core/tools/linker/AnalyzerTest.scala
gzm0: This is not about making the test flaky, but it's about making the test brittle: You do not want to test more than you expect from the specified implementation, so that when you change implementation details, you don't need to change the tests.

Otherwise you run into the risk of:
- Wrong changes to tests
- Maintenance cost

So I'm fine with any of the following:
- Define `ConflictingDefaultMethods` with a `Set`.
- Comparing the list of things in `ConflictingDefaultMethods` in an unordered fashion.
- Specifying the order (if such a specification is possible). I would prefer one of the other two though.

sjrd:
Updated.

PR #3122, opened 2017-09-02
Fix #2645: Use less JSEnv runners in TestAdapter

### LINKED ISSUE #2645: Make test adapter's JSEnv usage more economical

### LINKED ISSUE #3143: Tests on Travis CI succeed then throw a TimeoutException
This happens 90% of the time, the tests complete successful, log "X tests: X passed, 0 failed" then a TimeoutException is thrown by the SJS test adaptor.

It only happens on Travis CI, never locally (probably, simply because local is faster).
It happens with both 0.6.18 and 0.6.20.

Example:
https://travis-ci.org/japgolly/scalajs-react/jobs/277215211#L2643-L2684

gzm0:
Instead of spawning 1 + (2 + P) * F runners per test run, where

  P: level of parallelism
  F: number of testing frameworks

we will only spawn P runners.

The fundamental change that allows us to do so, is that the JS side of
the test adapter is not tied to a purpose anymore: Before, the JVM side
could either start a master or a slave for a given tesing framework (or
a special purpose short-lived piece of code).

We change the JS side ("bridge") to not have any initial state and be
able to do everything as requested via RPC.

Notably, after this change the bridge
- can detect frameworks,
- is not tied to a specific framework
- can create an arbitrary type (master/slave) of runner.
- can hold runners for multiple frameworks at the same time.

As a consequence, multiple frameworks can share the JS bridges and the
JS VM that has been used for test detection can be reused.

Note that as a side-effect, this also fixes #3143 by abolishing all
timeouts.


sjrd:
Could you write add a commit message explaining what are the major design changes, please?

sjrd:
If we only spawn P runners, does that mean that one runner can contain both a master and a slave at the same time? If that's the case, the other slaves will suffer important latency issues, since they can only receive new work to do in between the master-slave's duties as slave. In other words, if the master-slave executes Test1 while slave 2 executes Test2, and if Test2 finishes earlier than Test1, slave 2 is idle until Test1 finishes and the master-slave can process its master messages.

gzm0:
> If we only spawn P runners, does that mean that one runner can contain both a master and a slave at the same time?

Yes, but we map each VM to a JVM thread and the work balancing is done by sbt (or whichever client) not by the master/slave. The master/slave distinction is for internal bookkeeping of the frameworks.

Indeed from the testing interface perspective, the master/slave distinction is misleading as both the master or slaves can do whatever a test framework client can request.

A typical invocation sequence is as follows:

1. Create all tasks (call `tasks`). This can happen on either a master or a slave.
2. Call `execute` on these tasks over different threads. Typically, slaves send messages to the master (asynchronously) to keep track of completed stuff.

Note that in step 2, the master is not on the critical path for test execution, so a separate master would be mostly idling.

test-common/src/main/scala/org/scalajs/testcommon/RunMuxRPC.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /** Multiplexer map.
+   *
+   *  Access to the outer map needs to synchronized.
+   *  Access to the inner map only needs to be synchronize for writing.
+   */
+  private[this] val mux = mutable.Map.empty[RPCCore.OpCode, ConcurrentHashMap[RunId, _]]
+
+  def call[Req](ep: MuxRPCEndpoint[Req], runId: RunId)(req: Req): Future[ep.Resp] =
+    rpc.call(ep)(new RunMux(runId, req))
```
sjrd: Why is the signature of this method not
```scala
def call(ep: RPCEndpoint, runId: RunId)(req: ep.Req): Future[ep.Resp]
```
i.e., exploiting the path-dependent type `eq.Req`? Same for the other methods below.

test-common/src/main/scala/org/scalajs/testcommon/RunMuxRPC.scala
```diff
@@ -0,0 +1,81 @@
...
+    val dispatch = mux.getOrElseUpdate(opCode, newDispatchMap())
+    val old = dispatch.asInstanceOf[DispatchMap].put(runId, ex)
+    require(old == null, s"Duplicate endpoint for opcode $opCode run $runId")
+  }
+
+  def detach(ep: Endpoint, runId: RunId): Unit = synchronized {
+    val opCode = ep.opCode
+
+    val dispatch = mux.getOrElse(opCode,
+      throw new IllegalArgumentException(s"No endpoint attached for opCode $opCode"))
```
sjrd: 4 spaces

test-common/src/main/scala/org/scalajs/testcommon/RunMuxRPC.scala
```diff
@@ -0,0 +1,81 @@
...
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.util.Try
+
+import java.util.concurrent.ConcurrentHashMap
+
+import FutureUtil.futureFromTry
+
+/** Helper above an [[RPCCore]] that allows to multiplex between runs. */
+private[scalajs] final class RunMuxRPC(rpc: RPCCore) {
```
sjrd: I'm not sure I really understand what this class does :s

test-common/src/main/scala/org/scalajs/testcommon/RunnerMux.scala
```diff
@@ -0,0 +1,20 @@
+package org.scalajs.testcommon
+
+private[scalajs] final class RunMux[+T](val runId: RunMux.RunId, val value: T)
+
+private[scalajs] object RunMux {
+  type RunId = Int
```
sjrd: `RunID`

test-interface/src/main/scala/org/scalajs/testinterface/internal/FrameworkLoader.scala
```diff
@@ -19,4 +19,31 @@ private[internal] object FrameworkLoader {
...
+      name.split('.').foldLeft[js.UndefOr[js.Dynamic]](exportsNamespace) {
+        (prev, part) => prev.map(_.selectDynamic(part))
+      }.isDefined
+    }
+
+    def frameworkExists(name: String): Boolean =
+      frameworkExistsInReflect(name) || frameworkExistsInExportsNamespace(name)
+
+    for (frameworkNames <- names)
+    yield frameworkNames.find(frameworkExists(_))
```
sjrd: 2 spaces

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -0,0 +1,165 @@
...
+    runner.start(config.logger, config.console)
+    new ManagedRunner(threadId, runner)
+  }
+}
+
+object TestAdapter {
+  final class Config private (
+    val timeout: Duration,
+    val logger: Logger,
+    val console: JSConsole
```
sjrd: 4 spaces

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -0,0 +1,165 @@
...
+  private[this] val nextRunId = new AtomicInteger(0)
+  private[this] val runs = TrieMap.empty[RunMux.RunId, Unit]
+
+  /** Creates an `sbt.testing.Framework` for each framework that can be found.
+   *
+   *  The returned Frameworks bind to this TestAdapter and are only valid until
+   *  [[close]] is called.
+   */
+  def loadFrameworks[TestFramework](
+      frameworkNames: List[List[String]]): List[Option[Framework]] = {
```
sjrd: The `TestFramework` type parameter is not used.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -0,0 +1,165 @@
...
+
+      runs.clear()
+      if (seenRuns.nonEmpty) {
+        throw new IllegalStateException(
+            s"close() called with incomplete runs: $seenRuns")
+      }
+    }
+  }
+
+  private[testadapter] def timeout(): Duration = config.timeout
```
sjrd: Looks like this method shouldn't have `()`. It's a pure getter.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/RunnerAdapter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+      slaves.clear()
+      testAdapter.runDone(runId)
+    }
+  }
+
+  private def getRunnerRPC(): RunMuxRPC = {
+    val mRunner = testAdapter.getRunnerForThread()
+
+    if (mRunner != master && !slaves.contains(mRunner.id)) {
```
sjrd: This actually looks like it will never call `JSEndpoints.createSlaveRunner` if we happen to call this method from the thread of the master, even if we intend to use that `RunMuxRPC` for slave tasks.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/ScalaJSFramework.scala
```diff
@@ -21,76 +21,36 @@ import org.scalajs.testcommon._
...
 ) extends Framework {
 
   def this(frameworkName: String, libEnv: ComJSEnv, logger: Logger,
       jsConsole: JSConsole) = {
     this(frameworkName, libEnv, ModuleKind.NoModule, None, logger, jsConsole)
   }
 
-  private[this] val frameworkInfo = fetchFrameworkInfo()
+  private[this] val adapter = new TestAdapter(libEnv, moduleKind, moduleIdentifier,
+    TestAdapter.Config().withLogger(logger).withJSConsole(jsConsole))
```
sjrd: 4 spaces

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TaskAdapter.scala
```diff
@@ -0,0 +1,54 @@
...
+package org.scalajs.testadapter
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+import org.scalajs.testcommon._
+
+import sbt.testing._
+
+private[scalajs] final class TaskAdapter(taskInfo: TaskInfo, runId: RunMux.RunId,
```
sjrd: `private[testadapter]`?

test-adapter/src/main/scala/org/scalajs/sbttestadapter/RunnerAdapter.scala
```diff
@@ -0,0 +1,120 @@
...
+import scala.util._
+
+import org.scalajs.testcommon._
+
+import sbt.testing._
+
+import FutureUtil._
+import TestAdapter.ManagedRunner
+
+private[scalajs] final class RunnerAdapter private (runnerArgs: RunnerArgs,
```
sjrd: `private[testadapter]`?

test-adapter/src/main/scala/org/scalajs/sbttestadapter/RunnerAdapter.scala
```diff
@@ -0,0 +1,120 @@
...
+      Await.result(
+          mRunner.com.call(JSEndpoints.createSlaveRunner)(runnerArgs),
+          testAdapter.timeout)
+    }
+
+    mRunner.mux
+  }
+}
+
+private[scalajs] object RunnerAdapter {
```
sjrd: `private[testadapter]`?

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -0,0 +1,165 @@
...
+      case ModuleKind.CommonJSModule =>
+        val moduleIdent = moduleIdentifier.getOrElse {
+          throw new IllegalArgumentException(
+              "The module identifier must be specified for CommonJS modules")
+        }
+        s"""require("${escapeJS(moduleIdent)}").""" // note the final '.'
+    }
+
+    val code = s"""${prefix}org.scalajs.testinterface.internal.startBridge();"""
+    val launcher = new MemVirtualJSFile(s"testMaster.js").withContent(code)
```
sjrd: `testMaster.js` does not feel appropriate anymore. Maybe `startTestBridge.js`?

(also the `s` interpolator is useless)

japgolly:
Hey @gzm0. I just realised that you cc'd me here earlier. Am I related to this PR? Or is there something I can help with?
(This PR sounds excellent btw)

gzm0:
> Am I related to this PR? Or is there something I can help with?

Just wanted to keep you in the loop since you requested this :)

japgolly:
> Just wanted to keep you in the loop since you requested this :)
>
> EDIT: Or at least so I remember :P

Oh did I? It's possible because I do think the goals of this PR are fantastic :)

Personally I worked around this by upgrading from 16GB RAM to 32GB. Not everyone has that luxury though so :+1: :+1: :+1:

gzm0:
Jenkins, test this please

gzm0:
Rebased on top of #3150.

gzm0:
I am rebasing this to not block on the performance regression (#3123). I still do not know how to solve the performance regression, and this should get in, especially if we do not intend to revert the code that caused #3123.

So this should be ready for review. It'll just need to be rebased on the towards 0.22 commit.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -0,0 +1,191 @@
...
+  /** Called by [[RunnerAdapter]] when the run is completed. */
+  private[testadapter] def runDone(runID: RunMux.RunID): Unit = {
+    val old = runs.remove(runID)
+    require(old.nonEmpty, s"Tried to remove nonexistent run $runID")
+  }
+
+  private[testadapter] def getRunnerForThread(): ManagedRunner = {
+    // Prevent runners from being started after closing started.
+    // Otherwise we might leak runners.
+    require(!closing.get(), "We are closing. Cannot create new runner.")
```
sjrd: What happens if `close()` is called right after this `require` has been tested? It seems that in that case we would be silently leaking stuff.

gzm0: Yes, indeed. This require is just to catch bugs, it doesn't actually enforce the synchronization requirement. Creating a run on a returned `Framework` or calling `close` when there are incomplete runs is undefined behavior.

I can of course make this threadsafe, but I'll probably need a lock, which I'd rather avoid here (since this method will be called for each test case, so I'd like to minimize potential contention).

sjrd: OK, that sounds reasonable.

sjrd:
There are only a couple binary compatibility issues.

gzm0:
There were some trailing whitespaces. Updated.

gzm0:
Aaaaand of course I nuked the binary compat change with the force push from the other machine :-/ git totally needs an evolve extension :)

japgolly:
Fantastic!

I just gave this a try, it gets stuck pretty easily so far.

On one attempt I saw
```
[error] PhantomJS has crashed. Please read the bug reporting guide at
[error] <http://phantomjs.org/bug-reporting.html> and file a bug report.
```
and then after the rest of the tests completed, it just hung.

On another attempt I saw 

```

  phantomjs://code/phantomjs-launcher9060106815877187942.js:9 in onError


  phantomjs://code/phantomjs-launcher9060106815877187942.js:11 in onError
  file:///tmp/shipreq-sjs-22.sbt/webapp-client-project/target/scala-2.12/webapp-client-project-test-fastopt.js:26478 (in function "handleMessage__T__V")

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
  file:///tmp/shipreq-sjs-22.sbt/webapp-client-project/target/scala-2.12/webapp-client-project-test-fastopt.js:181656

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
  file:///tmp/1510141923375-0/comSetup.js:30

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
```
and it hung again.

I ran jstack and uploaded the output here:
https://www.dropbox.com/sh/ouqmuqouqtvoivv/AAAgDaapef7LVPHhJyx6LWgca?dl=0

sjrd:
@japgolly Would you mind filing that as a separate issue please? In a closed PR we will lose track of your comment very easily.

japgolly:
No prob, I'll do it now

PR #3139, opened 2017-09-11
Move the Int to Long conversion to `RuntimeLong.fromInt(x)`.

sjrd:
First, we get rid of `RuntimeLong.Zero`. Instead, the `Emitter` instantiates `new RuntimeLong(0, 0)`
explicitly and stores it in a global variable `$L0`. This allows the `RuntimeLong` module to have a pure constructor, which can therefore be elided when it is not needed after inlining.

Then, we can move the Int to Long conversion to `RuntimeLong.fromInt(x)`. Previously, it was an overloaded constructor `new RuntimeLong(x)`. This made sense in order to avoid loading the module instance of `RuntimeLong`. However, now that `RuntimeLong` has an elidable module accessor, this optimization is not necessary anymore.

With this change, `RuntimeLong` always has exactly one constructor, and it can therefore benefit from the "inlined constructor" optimization.

These changes do not appear to have any measurable impact on the performance on `Long`s, though.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
@@ -250,14 +250,12 @@ final class Emitter private (config: CommonPhaseConfig,
    *  zero of type `Long`.
    */
   private def emitInitializeL0(): js.Tree = {
-    import TreeDSL._
     implicit val pos = Position.NoPosition
 
     // $L0 = new RuntimeLong(0, 0)
     js.Assign(
         jsGen.envField("L0"),
-        js.Apply(
-            js.New(jsGen.encodeClassVar(LongImpl.RuntimeLongClass), Nil) DOT LongImpl.initFromParts,
+        js.New(jsGen.encodeClassVar(LongImpl.RuntimeLongClass),
             List(js.IntLiteral(0), js.IntLiteral(0)))
     )
   }
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
@@ -147,16 +147,18 @@ final class Emitter private (config: CommonPhaseConfig,
 
       /* Emit all the classes, in the appropriate order:
        *
-       * - First, all class definitions, which depend on nothing but their
-       *   superclasses.
-       * - Second, all static field definitions, which depend on nothing,
-       *   except those of type Long which need to instantiate RuntimeLong.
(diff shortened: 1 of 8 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -165,6 +165,18 @@ final class Emitter private (config: CommonPhaseConfig,
...
+      builder.addJSTree({
+        import TreeDSL._
+        implicit val pos = Position.NoPosition
+        js.Assign(
+            jsGen.envField("L0"),
+            js.Apply(
+                js.New(jsGen.encodeClassVar(LongImpl.RuntimeLongClass), Nil) DOT LongImpl.initFromParts,
+                List(js.IntLiteral(0), js.IntLiteral(0)))
+        )
+      })
```
gzm0: Consider moving the tree definition itself to a helper to keep this method cleaner. Also, consider updating the comment that static field definitions also includes `L0`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2267,8 +2267,8 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
           genLongZero()
         case LongLiteral(value) =>
           val (lo, hi) = LongImpl.extractParts(value)
-          genNewLong(LongImpl.initFromParts,
-              js.IntLiteral(lo), js.IntLiteral(hi))
+          js.New(encodeClassVar(LongImpl.RuntimeLongClass),
+              List(js.IntLiteral(lo), js.IntLiteral(hi)))
```
gzm0: Can we find a way that this does not blow up spectacularly if somebody has the great idea to synthesize a class in IR that inherits from `RuntimeLong`? (strictly speaking, this is legal, the IRChecker allows it).

sjrd: Hum ... add the notion of `final` in the IR? The other alternative would be to special-case `RuntimeLong` in the IR checker.

We have the same problem with `java.lang.Class`, btw.

sjrd:
Updated.

I haven't addressed the "blow up spectacularly" part, since it's not clear what's the best way forward.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
gzm0: Probably that'll be necessary :S Not in this PR in any case :)

PR #3141, opened 2017-09-12
DO NOT MERGE Optimize the constructor of `RuntimeLong` with a hack.

sjrd:
We avoid generating the super constructor call and creating the fields with the zero of their types. This leaves only the content of the init method (which we know is inlined for RuntimeLong), which directly assigns `this.lo` and `this.hi`.

We can do this because the superclasses of RuntimeLong do not declare any fields (hence the super constructor call is unnecessary) and because the init method is known to "behave well" wrt. the initialization of its fields (which means there is no need to initialize them first to their zero).

This optimization brings a 17% improvement on the SHA-512 benchmark.

We cannot apply this trick in ECMAScript 2015, because calling the `super()` constructor is mandatory for non-root classes.

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
@@ -551,9 +551,6 @@ object RuntimeLong {
   /** The hi part of a (lo, hi) return value. */
   private[this] var hiReturn: Int = _
 
-  /** The instance of 0L, which is used by the `Emitter` in various places. */
-  val Zero = new RuntimeLong(0, 0)
-
   private def toString(lo: Int, hi: Int): String = {
     if (isInt32(lo, hi)) {
       lo.toString()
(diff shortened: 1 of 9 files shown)


sjrd:
This builds on top of #3139, so Do Not Merge for now. Only the third commit belongs to this PR.

sjrd:
Note that virtually all of the perf improvement comes from removing the super call (not from removing the assignment of fields to their zero). And that is something we could do to *all* classes in the world with a bit more effort. We just have to create all the inherited fields of a class in its constructor, rather than calling the super constructor. This requires some global knowledge, though, so it's not completely trivial to implement. But we really should do that. The perf improvements measured here are likely to apply across the board.

I'm submitting this PR as is because I'm on a paper deadline :P But maybe we should refrain from merging it until a global solution is implemented.

sjrd:
Well, I was motivated tonight, so I implemented the global solution. Hence I believe this PR is superseded by #3142, although I might revive the part that avoids the `= 0`'s for `RuntimeLong` later.

PR #3142, opened 2017-09-12
Avoid calling the super JS constructors in Scala classes.

sjrd:
With the exception of "inlined inits", the job of the JS constructor in a Scala class is to create the fields, initialized to the zero of their types.

Previously, the JS constructor of some class `A` extending `B` were created as follows:

1. call the super constructor to create all inherited fields,
2. create the fields directly declared in `A`.

```js
function $c_A() {
  $c_B.call(this);
  this.aField1 = 0;
  this.aField2 = 0;
  <optional inlined init body>
}
```

This was however non-optimal, as calling the super constructor is pretty expensive. Instead, we now make the JS constructor of `A` directly create all its fields, including inherited ones:
```js
function $c_A() {
  this.bField1 = 0;
  this.bField2 = 0;
  this.aField1 = 0;
  this.aField2 = 0;
  <optional inlined init body>
}
```

This optimization is only performed in ECMAScript 5.1, since with ES 2015 classes, calling the super constructor is mandatory.

This optimization is ridiculously effective, as shown by our benchmark suite:

* Tracer is 25% faster
* DeltaBlue is 37% faster
* SHA-512 is 14% faster


tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -337,11 +340,12 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
   }
 
   /** Generates the creation of fields for a Scala class. */
-  private def genFieldDefsOfScalaClass(tree: LinkedClass)(
+  private def genFieldDefsOfScalaClass(tree: LinkedClass,
+      fields: List[FieldDef])(
```
gzm0: Remove `tree` from args, it is unused (`tpe` is also unused). As a consequence, probably rename the method.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -276,31 +268,42 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
...
       implicit globalKnowledge: GlobalKnowledge): WithGlobals[js.Function] = {
 
     implicit val pos = tree.pos
 
-    val fieldDefs = genFieldDefsOfScalaClass(tree)
+    val superCtorCallAndFieldDefs = outputMode match {
+      case OutputMode.ECMAScript51Isolated =>
+        val allFields =
+          globalKnowledge.getAllScalaClassFieldDefs(tree.encodedName)
+        genFieldDefsOfScalaClass(tree, allFields)
```
gzm0: Just to double check: This is safe because we never inline the init of a class that has instantiated subclasses? Maybe add a comment about this?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
sjrd: Well spotted. I would keep the name, though, as generating fields for a JS class is done elsewhere, and doesn't do the same thing.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
sjrd: Ah, I see what you mean. Yes, indeed, that's right.

sjrd:
Updated, including checksizes.

PR #3146, opened 2017-10-01
Further clean up the sbt plugin

sjrd:


tools/jvm/src/main/scala/org/scalajs/core/tools/io/FileVirtualFiles.scala
```diff
@@ -205,8 +207,13 @@ object FileScalaJSIRContainer {
     }
 
     for (ir <- walkForIR(baseDir)) yield {
-      val relDir = ir.getPath.stripPrefix(baseDir.getPath)
-      FileVirtualScalaJSIRFile.relative(ir, relDir)
+      val relPath0 = ir.getPath
+        .stripPrefix(baseDir.getPath)
+        .replace(java.io.File.separatorChar, '/')
+      val relPath =
+        if (relPath0 != "" && relPath0.charAt(0) == '/') relPath0.substring(1)
```
gzm0: Use `stripPrefix("/")` instead?

sjrd:
Updated, but the last commit does something funny to `checksizes`. I'll have a look at that tomorrow.

sjrd:
Seems to be a gzip version thing. Locally I see no difference in `-fastopt.js` nor `-opt.js` between commits. I also get the exact same sizes as the Jenkins build for the `.js` files, but not for the `.js.gz` files.

I have added a first commit to bump the checksizes.

sjrd:
Good point. Done.

PR #3150, opened 2017-10-08
Fix #3123: Avoid context switches in test receiver

### LINKED ISSUE #3123: Performance regression in running tests, v0.6.20
Was: **Performance regression in the optimiser**

I noticed that [upgrading Scala.js to v0.6.20](https://github.com/sparsetech/pine/commit/831541e0ff886c604db3b444286c1066ecc39185) increased the time of running our test suite by ~60%.

This regression can be reliably reproduced with the following commands:

```bash
git clone git@github.com:sparsetech/pine.git
cd pine

git checkout 284b857d0d0233966d2793d9c30e07a9434d724d  # Last commit before the upgrade
sbt clean && sbt pineJS/test  # Takes 29s

git checkout 831541e0ff886c604db3b444286c1066ecc39185  # First commit after the upgrade
sbt clean && sbt pineJS/test  # Takes 46s
```

It seems that the optimiser stage takes considerably longer. I opened a [similar issue before](https://github.com/scala-js/scala-js/issues/2943) which may be related.

### LINKED ISSUE #3143: Tests on Travis CI succeed then throw a TimeoutException

gzm0:
The new RPC interface for the test
adapter (bf2c835982198310b2793204e6c6c194e624772d)
has introduced a performance regression while running tests
(see #3123). This regression was caused by threads overly context
switching (due to the blocking interface of sbt.testing._).

We fix the performance regression by introducing our own future type
that can *only* be awaited. However, instead of simply waiting, it
"helps" the underlying com to perform I/O. As a result, the thread need
not context switch anymore: It will perform I/O until the requested
message has been received.

This commit introduces a little caveat however: The JVM side cannot
arbitrarily receive callbacks anymore. In fact, it can only receive
callbacks while a call is pending. Since all current usages of callbacks
are during pending calls (and will likely be in the future), this is not
a problem.

As a side-effect, this also fixes #3143, by not having timeouts at all
anymore.

### DIFF
--- a/test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
+++ b/test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
@@ -1,5 +1,7 @@
 package org.scalajs.testcommon
 
+import scala.language.higherKinds
+
 import scala.util.{Try, Failure, Success}
 
 import scala.collection.JavaConverters._
@@ -12,7 +14,7 @@ import java.util.concurrent.atomic.AtomicLong
 
 import Serializer.{serialize, deserialize}
 
-private[scalajs] abstract class RPCCore {
+private[scalajs] abstract class RPCCore[F[_]] {
   import RPCCore._
 
   /** Pending calls.
@@ -88,6 +90,9 @@ private[scalajs] abstract class RPCCore {
   /** Subclass needs to implement message sending. */
   protected def send(msg: String): Unit
 
+  /** Sublcass needs to implement its type of future. */
+  protected def toFuture[T](p: Promise[T]): F[T]
+
   /** Used to send a message to the other end. */
   final def send(ep: MsgEndpoint)(msg: ep.Msg): Unit = {
     import ep._
@@ -95,7 +100,7 @@ private[scalajs] abstract class RPCCore {
   }
 
   /** Used to make an actual call to the other end. */
-  final def call(ep: RPCEndpoint)(req: ep.Req): Future[ep.Resp] = {
+  final def call(ep: RPCEndpoint)(req: ep.Req): F[ep.Resp] = {
     import ep._
 
     // Reserve an id for this call.
@@ -115,7 +120,7 @@ private[scalajs] abstract class RPCCore {
     // Actually send message.
     send(msg)
 
-    promise.future
+    toFuture(promise)
   }
 
   final def attach(ep: MsgEndpoint)(ex: ep.Msg => Unit): Unit = {
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -17,7 +17,22 @@ object BinaryIncompatibilities {
   val SbtPlugin = Seq(
   )
 
-  val TestAdapter = Seq(
+  val TestCommon = Seq(
+      // private[scalajs], not an issue.
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+          "org.scalajs.testcommon.RPCCore.call"),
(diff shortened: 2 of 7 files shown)


sjrd:
I will look at this tomorrow. I don't have access to a computer today.

sjrd:

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -12,7 +14,7 @@ import java.util.concurrent.atomic.AtomicLong
 
 import Serializer.{serialize, deserialize}
 
-private[scalajs] abstract class RPCCore {
+private[scalajs] abstract class RPCCore[F[_]] {
```
sjrd: I have a vague feeling that this would be nicer as an abstract type member. Have you considered it? But I'm not too opinionated about this, so feel free to ignore my comment.

sjrd:
There seem to be consistent timeouts (stalls), in `sbt-plugin-test`.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
gzm0: I have indeed though so too but haven't tried out. I'm a bit concerned that `RunMuxRPC` becomes much worse with an abstract type member when used (because it introduces a path dependent type on the contained `RPCCore`).

gzm0:
> There seem to be consistent timeouts (stalls), in sbt-plugin-test.

Yes :( I haven't investigated it yet, but I'm a bit afraid that our "lying" strategy as too much of a lie... :-/

gzm0:
I have indeed lied too much. We currently rely on JS initiated communication when the master replies to a framework level message from the slave.
This shows in the test, because we wait for the confirmation of the "Hello" message from the master. I do not yet know what to do about this.

sjrd:
Could we have an additional catch-all thread that receives messages when no other thread has declared itself willing to do so?

Upon a new thread asking to take control, we could `interrupt()` the catch-all thread to cancel its blocking wait for new messages.

Not sure this would actually work, but it might be an idea.

gzm0:
Ugh... Yeah that could work. However, we then would require `receive` of each `JSEnv` to be interruptible and continuable. Which currently they are definitely not.

Also I'm not sure that this would cause less context switches: The interrupted thread would need to exit the critical section, which would cause a switch again.

gzm0:
I think the problem at this point is that what we have is a streaming RPC with the JVM being the client and JS being the server. If we were to model it that way, all these problems would go away (we would need to buffer framework level messages for delivery, but that is kind of OK).

However implementing streaming RPCs is nasty without streaming futures which we don't have (and I really don't want to implement). Further, I'd really like the `JSEnv` interface to be spec'd asynchronously, since this is much easier for error handling and teardown state, so I also don't want to consider a synchronous solution on the JVM side.

As a side note: This little incident has convinced me (once again) that lying is not OK. Notably, if I were to add the changes in #3122 on top, I'd expect this to work *most of the time*, because there is no dedicated master `JSEnv` anymore (so some call to `execute` would *probably* unblock the situation). However, in the case where the `JSEnv` that is the master happens to run out of work, the same deadlock could occur again -> happy debugging.

gzm0:
Closing this for now, as it does not work.

PR #3159, opened 2017-10-19
Add `js.special.instanceof(a, b)` for `a instanceof b`.

sjrd:
Previously, we had no way to write code equivalent to `a instanceof b` where `b` is a dynamic value (rather than something directly available as `js.constructorOf[C]` for some `C`). Adding this method fills up a hole in our interoperability features.

This will also be necessary to desugar `.isInstanceOf`s on nested JS classes.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/special/package.scala
+++ b/library/src/main/scala/scala/scalajs/js/special/package.scala
@@ -22,6 +22,25 @@ package scala.scalajs.js
  */
 package object special {
 
+  /** Dynamically tests whether a value is an instance of a JavaScript class.
+   *
+   *  This method is the exact equivalent of `x instanceof clazz` in
+   *  JavaScript.
+   *
+   *  Using this method is only necessary when `clazz` is only known at
+   *  run-time. In most cases, you should use
+   *  {{{
+   *  x.isInstanceOf[C]
+   *  }}}
+   *  instead, where `C` is the statically defined class corresponding to
+   *  `clazz`. In particular, the following identity holds for all `x` and `C`:
+   *  {{{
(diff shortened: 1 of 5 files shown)


library/src/main/scala/scala/scalajs/js/special/package.scala
```diff
@@ -22,6 +22,14 @@ package scala.scalajs.js
  */
 package object special {
 
+  /** Dynamically tests whether a value is an instance of a JavaScript class.
+   *
+   *  This method is the exact equivalent of `x instanceof clazz` in
+   *  JavaScript.
```
gzm0: Could you provide a rationale why this is in `special`? All other methods have it an I think it is a very good thing.

sjrd: Done.

sjrd:
> Just to check my understanding: This is necessary, because we need to desugar `isInstanceOf` on nested classes in `PrepJSInterop` (otherwise we have lost the nesting)?

Basically, that's the idea. It's quite complicated, though, and hard to thoroughly explain before you've seen the whole scheme for nested JS classes. But in a nutshell, we need to desugar `x.isInstanceOf[path.C]` early (before erasure) into `instanceof(x, js.constructorOf[path.C])`, and `js.constructorOf[path.C]` must also be desugared early into something like `path.C$jsclass`, where `C$jsclass` is a field used to store the nested class value that was created for the instance `path`. We need to do this before erasure because that phase will throw away the `path.` part, and replace it with `Enclosing#C`.

If you want to know more at this point, I suggest you read the big "strategy" comments at the beginning of `ExplicitInnerJS` and `ExplicitLocalJS` [in my branch for nested JS classes](https://github.com/scala-js/scala-js/compare/master...sjrd:nested-js-classes).

PR #3160, opened 2017-10-23
Make the lexical Env of FunctionEmitter more robust.

sjrd:
The lexical `Env` of `FunctionEmitter` was previously not always complete. It was not a big deal, because it is only used to avoid spilling expressions into temporaris if they are known to be immutable.

However, handling class captures in nested JS classes will require the `Env` to be complete, so that we can distinguish local variables from class captures. It is also internally nicer to have complete `Env`s.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
@@ -879,10 +879,17 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
             arg match {
               case Block(stats :+ expr) =>
                 val (jsStats, newEnv) = transformBlockStats(stats)
-                innerEnv = newEnv
-                val result = rec(expr)(newEnv) // right-to-left, remember?
+                val result = rec(expr)(newEnv)
                 // Put the stats in a Block because ++=: is not smart
                 js.Block(jsStats) +=: extractedStatements
+                innerEnv = stats.foldLeft(innerEnv) { (prev, stat) =>
+                  stat match {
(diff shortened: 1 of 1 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -957,12 +964,11 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
 
               case _ =>
                 val temp = newSyntheticVar()
-                val newEnv = env.withDef(temp, arg.tpe, false)
-                innerEnv = newEnv
                 val computeTemp = pushLhsInto(
                     Lhs.VarDef(temp, arg.tpe, mutable = false), arg,
                     Set.empty)
                 computeTemp +=: extractedStatements
+                innerEnv = innerEnv.withDef(temp, arg.tpe, false)
```
gzm0: Write `mutable = false`? Also a couple of occurrences below.

sjrd:
> Wow. I didn't know this is actually incomplete :)

Me neither. I discovered that through weird wrong codegen in the nested JS classes branch.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: I cannot. This is an *overload* pimped through an implicit class, and apparently that prevents scalac from understanding the named param as such.

gzm0: Blargh :P Ok

PR #3162, opened 2017-10-26
Small compiler refactorings before nested JS classes

sjrd:
This PR contains a series of small refactorings in the compiler, extracted from the work on nested JS classes. The main purpose of doing those separately is so that the diff for nested JS classes does not contain irrelevant changes.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -731,47 +731,8 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
         // If argument is undefined and there is a default getter, call it
         val verifiedOrDefault = if (param.hasFlag(Flags.DEFAULTPARAM)) {
           js.If(js.BinaryOp(js.BinaryOp.===, jsArg, js.Undefined()), {
-            val trgSym = {
-              if (sym.isClassConstructor) {
-                /* Get the companion module class.
-                 * For inner classes the sym.owner.companionModule can be broken,
-                 * therefore companionModule is fetched at uncurryPhase.
-                 */
-                val companionModule = enteringPhase(currentRun.namerPhase) {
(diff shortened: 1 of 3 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -826,28 +761,119 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+          enteringPhase(currentRun.posterasurePhase)(sym.paramss.flatten.map(ParamSpec))
+        val allParamsNow = sym.paramss.flatten.map(ParamSpec)
+
+        def mergeUncurryPosterasure(paramsUncurry: List[ParamSpec],
+            paramsPosterasure: List[ParamSpec]): List[ParamSpec] = {
+          paramsUncurry.zip(paramsPosterasure).map {
+            case (paramUncurry, paramPosterasure) =>
+              if (paramUncurry.isRepeated) paramUncurry
+              else paramPosterasure
+          }
```
gzm0: Write this as a for loop?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -826,28 +761,119 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+        val tpe = if (repeated) repeatedToSingle(sym.tpe) else sym.tpe
+        new ParamSpec(sym, tpe, repeated, hasDefault)
+      }
+    }
+
     private sealed abstract class Exported {
+      def sym: Symbol
       def pos: Position
-      def params: List[Type]
+      def params: Array[ParamSpec]
```
gzm0: Can this be typed as `IndexedSeq[ParamSpec]`? `Array` is mutable -> evil :)

sjrd:
Updated.

PR #3166, opened 2017-10-31
[no-master] Migrate to sbt-platform-deps to provide `%%%`.

sjrd:
~~At this point this is an experiment~~

We preserve binary compatibility and source compatibility by simply de-implicit-ify the conversion from `String` to the class providing `%%%`. This means that existing binaries will keep linking, while recompiled sources will resolve to the new implicit provided by sbt-platform-deps.

We preserve `%%%!` by introducing *another* implicit conversion to a class that has only `%%%!`, but not `%%%`. That implicit conversion is immediately deprecated, since Scala.js 1.x will not support `%%%!`.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -774,6 +774,8 @@ object Build {
           previousArtifactSetting,
           mimaBinaryIssueFilters ++= BinaryIncompatibilities.SbtPlugin,
 
+          addSbtPlugin("org.portable-scala" % "sbt-platform-deps" % "1.0.0"),
+
           /* TODO Remove this in 1.x, since there are no macros in sbt-plugin
            * anymore.
            */
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.16
+sbt.version=0.13.17
--- a/project/build.sbt
+++ b/project/build.sbt
@@ -4,6 +4,8 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.18")
 
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
 
+addSbtPlugin("org.portable-scala" % "sbt-platform-deps" % "1.0.0")
+
 libraryDependencies += "org.scala-js" % "closure-compiler-java-6" % "v20160517"
 
 libraryDependencies += "io.apigee" % "rhino" % "1.7R5pre4"
--- a/sbt-plugin-test/project/build.properties
+++ b/sbt-plugin-test/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.16
+sbt.version=0.13.17
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -9,6 +9,8 @@
 
 package org.scalajs.sbtplugin
 
+import scala.language.implicitConversions
+
 import sbt._
 import sbt.Keys._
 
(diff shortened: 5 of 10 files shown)


sjrd:
Introducing this change in the 0.6.x branch is necessary so that a later version of sbt-scalajs-crossproject can drop [the hack to disable Scala.js' own `%%%`](https://github.com/portable-scala/sbt-crossproject/blob/b399c75e6594e514039ecb061462f73855afe190/sbt-scalajs-crossproject/src/main/scala/scalajscrossproject/ScalaJSCrossPlugin.scala#L17-L18). That in turn will be necessary to be able to remove the binary compatibility hack in Scala.js 1.x from https://github.com/scala-js/scala-js/pull/3197/files#diff-4099fca73cb9eb3c57d3933a51d4d85b

project/Build.scala
```diff
@@ -775,6 +775,19 @@ object Build {
...
 
+          /* Work around https://github.com/sbt/sbt/issues/3393.
+           * Should be:
+           *   addSbtPlugin("org.portable-scala" % "sbt-platform-deps" % "1.0.0-M2")
+           * We inline the fixed definition of addSbtPlugin to be
+           * released with sbt 0.13.17.
+           */
+          libraryDependencies += {
+            Defaults.sbtPluginExtra(
+                "org.portable-scala" % "sbt-platform-deps" % "1.0.0-M2",
```
gzm0: The 0.6 branch may not depend on a milestone build of sbt-platform-deps. This breaks our own binary compatibility guarantee of the sbt plugin, no? (Next release of 0.6 wouldn't be binary compatible with one that uses sbt-platform-deps 1.0.0).

sjrd: Ah, right.

On the other hand, I don't except sbt-platform-deps 1.0.0 to break binary compatibility wrt. 1.0.0-M2. In fact I don't expect sbt-platform-deps 1.0.0 to be any different than 1.0.0-M2.

Maybe we just publish sbt-platform-deps 1.0.0 final right now?

/cc @densh

sjrd: Or we let this PR dormant for a bit while we publish Scala.js 1.0.0-M2, itself depending on sbt-platform-deps 1.0.0-M2, to give it some exposure in the wild. And if we're happy with it, only then we publish 1.0.0 final and come back to the present PR.

gzm0: That sounds very reasonable.

gzm0: Actually, just to double check: That will not make 0.6 / 1.0 cross compilation harder, will It?

sjrd: No, I already checked that one can use `sbt-scalajs-crossproject` 0.3.0 with any of the following:

* Scala.js 0.6.19 (and hence, I kind of assumed, 0.6.21 or the current 0.6.x branch)
* Scala.js 0.6.22-SNAPSHOT from this branch
* Scala.js 1.0.0-M2-SNAPSHOT from the branch of #3197.

And all of that, with the exact same codebase (except the change of version number for `sbt-scalajs`).

sjrd:
Updated with:
* Rebased
* Use sbt-platform-deps 1.0.0 final
* Add a second commit that deprecates `crossProject` in favor of using sbt-crossproject.

project/Build.scala
```diff
@@ -774,6 +774,19 @@ object Build {
           previousArtifactSetting,
           mimaBinaryIssueFilters ++= BinaryIncompatibilities.SbtPlugin,
 
+          /* Work around https://github.com/sbt/sbt/issues/3393.
+           * Should be:
+           *   addSbtPlugin("org.portable-scala" % "sbt-platform-deps" % "1.0.0")
+           * We inline the fixed definition of addSbtPlugin to be
+           * released with sbt 0.13.17.
+           */
```
gzm0: Shouldn't we just upgrade?

sjrd: We would lose the guarantee that sbt-scalajs can be used with sbt 0.13.16.

But maybe that's fine?

gzm0: Oh, I thought we have upgraded patch sbt versions at will in the past... If we haven't, I'm fine with either and leave it up to you.

sjrd:
Updated.

PR #3167, opened 2017-11-01
Fix #3157: Upgrade to Scala 2.12.4.

### LINKED ISSUE #3157: Upgrade to Scala 2.12.4
I published the necessary artifacts for Scala 2.12.4 for Scala.js 0.6.19, 0.6.20 and 1.0.0-M1. We should now update our build and CI for 2.12.4.

sjrd:


partest-suite/src/test/resources/scala/tools/partest/scalajs/2.12.4/run/anyval-box-types.check
```diff
@@ -0,0 +1,52 @@
...
+true
+1
+true
+-1
+true
+1
+true
+false
+true
+true
```
sjrd: The 2 lines above (11 and 12) are `false` in the original .check file, because they test `1.isInstanceOf[Short]` and `1.isInstanceOf[Double]`, which is false on the JVM but true in Scala.js.

[Source of the test](https://github.com/scala/scala/blob/3b7623facb6ce48c28d6afdf84e74a99d45f0e23/test/files/run/anyval-box-types.scala#L14-L15)

PR #3170, opened 2017-11-05
Fix #2528: Check top-level export conflicts in analyzer

### LINKED ISSUE #2528: Move top-level export name conflict check to analyzer
~~Depends on #2527.~~

gzm0:


tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Infos.scala
```diff
@@ -128,9 +131,14 @@ object Infos {
...
 
+    def setTopLevelExportNames(names: List[String]): this.type = {
+      topLevelExportNames = names
+      this
+    }
+
     def result(): ClassInfo = {
       ClassInfo(encodedName, isExported, kind, superClass,
-          interfaces.toList, methods.toList)
+          interfaces.toList, methods.toList, topLevelExportNames.toList)
```
sjrd: `.toList` is redundant. `topLevelExportNames` is already a `List`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -85,6 +85,8 @@ private final class Analyzer(config: CommonPhaseConfig,
         case CyclicDependencyException(chain, from) =>
           _errors += CycleInInheritanceChain(chain, from)
       }
+
+      checkConflictingExports()
```
sjrd: Consider putting this inside the `try/catch` for cyclic dependency errors. When a cyclic dependency error happens, things become pretty fragile and should terminate ASAP. The particular case of `checkConflicitingExports()` is safe, but I would prefer if we kept the codebase in a state where it's clear that nothing should happen after we've detected a cyclic dependency.

gzm0:
Updated.

PR #3171, opened 2017-11-05
Fix #2506: No duplicate vars in export dispatcher

### LINKED ISSUE #2506: Exports overload resolution creates multiple local vars with the same name
Example in `ExportsTest.disambiguate_overloads_involving_longs`:

``` scala
  def "foo"(arg$1: any): any = {
    if (arg$1.isInstanceOf[jl_Integer]) {
      val prep0: int = arg$1.asInstanceOf[I];
      this.$$js$exported$meth$foo__I__O(prep0)
    } else if (arg$1.isInstanceOf[jl_Long]) {
      val prep0: long = arg$1.asInstanceOf[J];
      this.$$js$exported$meth$foo__J__O(prep0)
    } else {
      throw "No matching overload"
    }
  }
```

`prep0` is defined twice (in different scopes). This happens not to be an issue in practice, but it's still technically invalid.

Discovered while addressing #2498, of course ...

**Note**: When we fix this, we should also completely enforce uniqueness of local variable names per method (instead of non-overlapping scopes), as was the initial plan of #2498.

gzm0:
We rely on freshLocalIdent to avoid naming conflicts. As a consequence, we can change the IR checker to enforce local var uniqueness in method scope (not just block scope).

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
@@ -313,21 +313,7 @@ private final class IRChecker(unit: LinkingUnit,
       reportError("Top level export must be static")
 
     checkExportedPropertyName(pName, classDef, isTopLevel)
-
-    for (ParamDef(name, tpe, _, _) <- params) {
-      if (tpe == NoType)
-        reportError(s"Parameter $name has type NoType")
-      else if (tpe != AnyType)
-        reportError(s"Parameter $name of exported method def has type $tpe, "+
-            "but must be Any")
-    }
-
-    if (params.nonEmpty) {
-      for (ParamDef(name, _, _, rest) <- params.init) {
-        if (rest)
-          reportError(s"Non-last rest parameter $name is illegal")
-      }
-    }
+    checkJSParamDefs(params)
 
     def isJSConstructor = {
       !static && (pName match {
@@ -468,20 +454,7 @@ private final class IRChecker(unit: LinkingUnit,
       return
     }
 
-    for (ParamDef(name, tpe, _, _) <- params) {
-      if (tpe == NoType)
-        reportError(s"Parameter $name has type NoType")
-      else if (tpe != AnyType)
-        reportError(s"Parameter $name of exported constructor def has type "+
-            s"$tpe, but must be Any")
-    }
-
-    if (params.nonEmpty) {
-      for (ParamDef(name, _, _, rest) <- params.init) {
-        if (rest)
-          reportError(s"Non-last rest parameter $name is illegal")
-      }
-    }
+    checkJSParamDefs(params)
 
     val thisType = ClassType(classDef.name.name)
     val bodyEnv = Env.fromSignature(thisType, params, NoType)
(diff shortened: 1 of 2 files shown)


tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -1230,23 +1245,13 @@ private final class IRChecker(unit: LinkingUnit,
...
-          implicit val ctx = ErrorContext(p)
-          val name = ident.name
-          if (prevParams.contains(name))
-            reportDuplicateLocalVarName(name)
-          prevParams + (name -> LocalDef(name, tpe, mutable)(p.pos))
-      }
-      new Env(thisType, paramLocalDefs,
+      val paramLocalDefs =
+        for (p @ ParamDef(ident, tpe, mutable, _) <- params)
+          yield ident.name -> LocalDef(ident.name, tpe, mutable)(p.pos)
```
sjrd: Why didn't you call `checkDeclareLocalVar` in here, instead of adding calls to it in various parts dealing with `ParamDef`s?

gzm0: Because:
https://github.com/sjrd/scala-js/commit/f47c51f76944739b43c1928444e99884b1128738 :)

I can check here as well, do you think it is cleaner overall?

sjrd: Hum ... Yes, I think it's cleaner if it's done once and for all in `fromSignature`, don't you think?

gzm0: Yeah, probably you're right... I'll change it.

gzm0: Updated. I tried doing this and really did not like it, notably because we only create `Env`'s lazily when we traverse into the trees. So in some cases, parameter checking would not happen, even if it could. IMO it is cleaner overall if we keep the `Env` side effect free.

sjrd: Ah I see. Let's keep it, then.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -1056,6 +1025,30 @@ private final class IRChecker(unit: LinkingUnit,
...
+        reportError(s"Parameter $name has type $ptpe but must be any")
+    }
+
+    if (params.nonEmpty) {
+      for (ParamDef(name, _, _, rest) <- params.init) {
+        if (rest)
+          reportError(s"Non-last rest parameter $name is illegal")
+      }
+    }
+
```
sjrd: Missing `}` here causing a parse error.

gzm0:
Updated.

sjrd:
Good :) Could you rebase this PR now that #3186 has be fast-forwarded, so that we are sure that CI still passes when combining the two? Mostly asking because we enabled IR-checking of partests in the merge, and this PR introduces more IR-checking, so there's a non-negligible possibility of interaction.

gzm0:
Grrr... I didn't see that comment. Reverting.

gzm0:
Force pushed to revert, reopened #3187.

PR #3178, opened 2017-11-11
Fix #3072: Enable the IR checker by default.

### LINKED ISSUE #3072: Enable the IR checker by default in 1.x
Was: **Enable the IR checker by default in fullOptJS**

Our main motivation for *not* enabling the IR checker by default is that it is slow.

On a batch linking of the test suite (but with a warm sbt JVM), it takes a full second to Info-check and IR-check the test suite.

On an incremental link, it still takes more than 700 ms. (IR checking is not incremental at all, but on an incremental link some I/O is avoided due to caches of the IR files.) That's about 1/6 of the total linking time. For "normal" codebases, it should be bigger, because Info checking is proportional to the classpath size, not to the reachable part thereof. In the test suite, basically the entire classpath is reachable (or so we hope: code coverage).

In fullOptJS, though, this is a tiny amount of time, because GCC is by a very large margin the bottleneck. So we can afford some more processing for the Info and IR checkers; it won't bother anyone.

This would significantly increase the code surface that our IR checker is exposed to, which would hopefully detect things like #3054 and #3058 much earlier.

/cc @andreaTP who suggested this.

sjrd:
This will help catching potential internal bugs during the Milestones and RC cycles.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/StandardLinker.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/StandardLinker.scala
@@ -91,7 +91,7 @@ object StandardLinker {
       this(
           semantics = Semantics.Defaults,
           moduleKind = ModuleKind.NoModule,
-          checkIR = false,
+          checkIR = true,
           optimizer = true,
           parallel = true,
           sourceMap = true,
@@ -191,7 +191,7 @@ object StandardLinker {
      *
      *  - `semantics`: [[Semantics.Defaults]]
      *  - `moduleKind`: [[ModuleKind.NoModule]]
-     *  - `checkIR`: `false`
+     *  - `checkIR`: `true`
      *  - `optimizer`: `true`
      *  - `parallel`: `true`
      *  - `sourceMap`: `true`


sjrd:
Unfortunately we have 2 partests failing: [run/t6272.scala](https://github.com/scala/scala/blob/2.13.x/test/files/run/t6272.scala) and [run/t3877.scala](https://github.com/scala/scala/blob/2.13.x/test/files/run/t3877.scala). It turns out they were not IR-checked (bad). They fail to IR-check on 2.11 (and I guess would on 2.10 too). On 2.12+, the produced IR is correct. Given that they are caused by `lazy` vals, this is not surprising, since scalac significant changed its treatment of lazy vals in 2.12. The errors are:
```
[error] file:/C:/Users/Sepi/Documents/Projets/test/Test.scala(30:7:VarDef): Duplicate local variable name bitmap$0.
[error] file:/C:/Users/Sepi/Documents/Projets/test/Test.scala(2:10:VarDef): Duplicate local variable name bitmap$0.
```
The issue is also reproducible in 0.6.x.

sjrd:
Rebased.

PR #3179, opened 2017-11-11
Remove all primitives from JS types

sjrd:
The semantics of JS types are governed by the interoperability semantics. According to those semantics, a method call of the form `x.m()` should always correspond to the equivalent JS method call. If `m` is statically declared as a primitive, those semantics are broken.

Therefore, we avoid any primitive in JS types.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Any.scala
+++ b/library/src/main/scala/scala/scalajs/js/Any.scala
@@ -195,6 +195,14 @@ object Any extends js.LowPrioAnyImplicits {
     value.asInstanceOf[js.Any]
 
   implicit class ObjectCompanionOps(val __self: js.Object.type) extends AnyVal {
+    /** Tests whether the specified object `o` has a property `p` on itself or
+     *  in its prototype chain.
+     *
+     *  This method is the equivalent of `p in o` in JavaScript.
+     */
+    def hasProperty(o: js.Object, p: String): Boolean =
+      js.special.in(p, o)
+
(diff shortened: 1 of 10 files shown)


library/src/main/scala/scala/scalajs/js/Any.scala
```diff
@@ -195,6 +195,14 @@ object Any extends js.LowPrioAnyImplicits {
...
 
   implicit class ObjectCompanionOps(val __self: js.Object.type) extends AnyVal {
+    /** Tests whether the specified object `o` has a property `p` on itself or
+     *  in its prototype chain.
+     *
+     *  This method is the equivalent of `p in o` in JavaScript.
+     */
+    def hasProperty(o: js.Object, p: String): Boolean =
+      js.special.in(p, o)
+
```
gzm0: Why no overload for symbols?

sjrd: I was thinking I'd keep the surface API intact in this "refactoring" PR. But we could add one, yes. Or maybe better, type `p` as `String | js.Symbol`?

gzm0: Ah, ok. If you want to add it later, that's fine with me. There are probably other spots in this file where Symbol "support" should be added.

I'm not sure typing as `String | js.Symbol` is a good idea. I would expect better behavior from the Scala typer in general with overloads than implicits.

PR #3182, opened 2017-11-14
Don't fix #3180: Blacklist run/t3877.scala and run/t6272.scala.

sjrd:
Those partests are affected by #3180, which we decided not to fix, given that the root cause in scalac also affects Scala/JVM and produces invalid bytecode in some cases.

PR #3183, opened 2017-11-14
Upgrade to Scala 2.10.7 and 2.11.12

sjrd:
~~Depends on #3182. Will need to be rebased on top of it once it's merged.~~


sjrd:
One of our tests fails on the JVM on 2.10.7 and 2.11.12. It's a regression in scalac/JVM: https://github.com/scala/bug/issues/10609

I'll `assume` away that test on those versions.

sjrd:
Updated.

PR #3184, opened 2017-11-14
Fix #1265: Intrinsify `js.Dictionary.apply` and `js.Dynamic.literal`.

### LINKED ISSUE #1265: Make js.Dictionary.apply() an intrinsic
It should work basically the same as `js.Dynamic.literal.applyDynamic`, and emit a raw `JSObjectConstr` when possible.

Alternative: make it an intrinsic, and implement `js.Dynamic.literal` in user-space by delegating to this one.
Pros:
- New methods, defined in user-space, can benefit from the same optimizations if they are inlined and call that intrinsic

Cons:
- It's going to be tricky for the optimizer to recover the structure
- It's not aligned with the way `js.Array.apply()` is handled (but arguably, if we do this, then this one should also become an intrinsic ...)

sjrd:
To be more exact, they are both written in user-space as forwarders to a new method `js.special.objectLiteral`, which is itself an intrinsic.

This improves the generated code for calls to `js.Dictionary.apply` with known variadic arguments. It will also improve the generated code for user-defined helpers that forward themselves to `js.Dictionary.apply` or `js.Dynamic.literal`.

The diff of `scalajs-testsuite-fastopt.js` for the actual intrinsification commit is available [in this gist](https://gist.github.com/sjrd/5a1a20da63fd7ffba9c7aa727628c93a).


gzm0:
Oh, sorry. I just realized that this merge nuked your merge from 0.6 :( feel free to undo this merge.

PR #3185, opened 2017-11-15
Actually use the IR implementation of hijacked classes in the Emitter.

sjrd:
The main topic of this PR is that the `Emitter` now actually uses the IR of hijacked classes. Instead of duplicating the implementation of all the methods of hijacked classes in `scalajsenv.js`, they are now generated by the Emitter from the IR, in a way very similar to default methods.

Once we do this, we can finally fix the IR of `String` to actually contain all its methods, rather than having most of them in `RuntimeString`, with compiler hacks to reroute them that way.


sjrd:
Ready for review. @gzm0 I think you'll like this one :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2336,44 +2379,28 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
+      case ByteType       => Definitions.BoxedByteClass
+      case ShortType      => Definitions.BoxedShortClass
+      case IntType        => Definitions.BoxedIntegerClass
+      case LongType       => Definitions.BoxedLongClass
+      case FloatType      => Definitions.BoxedFloatClass
+      case DoubleType     => Definitions.BoxedDoubleClass
+      case StringType     => Definitions.StringClass
+    }
+
+    /* Ideally, we should figure out this set using a CHA. However, this is
```
gzm0: CHA?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3811,28 +3805,16 @@ abstract class GenJSCode extends plugins.PluginComponent
...
-                      methodIdent,
-                      castCallTrg :: arguments,
-                      toIRType(retTpe)),
-                  retTpe)
-            }
+            val retTpe = methodInStringClass.tpe.resultType
+            val castCallTrg = fromAny(callTrg, StringClass.toTypeConstructor)
+            boxIfNeeded(
+                genApplyMethod(castCallTrg, methodInStringClass, arguments),
+                retTpe)
```
gzm0: Why is string any different here than any other hijacked class?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5451,13 +5428,18 @@ abstract class GenJSCode extends plugins.PluginComponent
   private def isStringType(tpe: Type): Boolean =
     tpe.typeSymbol == StringClass
 
-  protected lazy val isHijackedBoxedClass: Set[Symbol] = {
-    /* This list is a duplicate of ir.Definitions.HijackedBoxedClasses, but
+  protected lazy val isHijackedClass: Set[Symbol] = {
+    /* This list is a duplicate of ir.Definitions.HijackedClasses, but
```
gzm0: Do we still need `ir.Definitions.HijackedBoxedClasses`? If yes, why?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Class Hierarchy Analysis. I thought that was a sufficiently well-known term. I'll elaborate.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: Right now, there are still two cases in the Emitter that use `ir.Definitions.HijackedBoxedClasses`. However that's limited to how the emitter does its job, and that definition should definitely be moved over there. However, I suggest we don't do that in this PR as it is not directly related. There are other things in `Definitions` that should be removed, btw.

I've filed #3195.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: It isn't (apart from the fact that other hijacked classes also have an associated primitive type). There is indeed some factoring out to do here.

sjrd:
Updated.

PR #3188, opened 2017-11-16
 Remove unneeded artifacts from old test interface

gzm0:


gzm0:
I manually tested the HTML runners.

test-adapter/src/main/scala/org/scalajs/testadapter/HTMLRunnerBuilder.scala
```diff
@@ -89,12 +88,11 @@ object HTMLRunnerBuilder {
       frameworkImplClassNames: List[List[String]],
       taskDefs: List[TaskDef]): String = {
 
-    val taskDefsString = Serializer.serialize(taskDefs)
+    def mkVar[T : Serializer](name: String, value: T) =
```
sjrd: `T: Serializer` (I think it won't pass the Scalastyle checks)

gzm0:
Note to self: Remove dependencies on json-simple.

gzm0:
Updated.

PR #3191, opened 2017-11-18
Reduce the code size for toJSArray calls

Atry:
@kahliburke mentioned this problem at https://gitter.im/ThoughtWorksInc/Binding.scala?at=5a0fa86471ad3f8736fba227 and https://gitter.im/scala-js/scala-js?at=5a0b53da505b630c05e202dc

### DIFF
--- a/library/src/main/scala/scala/scalajs/runtime/package.scala
+++ b/library/src/main/scala/scala/scalajs/runtime/package.scala
@@ -32,9 +32,13 @@ package object runtime {
       case col: js.ArrayOps[A]     => col.result()
       case col: js.WrappedArray[A] => col.array
       case _ =>
-        val result = new js.Array[A]
-        col.foreach(x => result.push(x))
-        result
+        @noinline
+        def unhappyPath() = {
+          val result = new js.Array[A]
+          col.foreach(x => result.push(x))
+          result
+        }
+        unhappyPath()
     }
   }
 


Atry:
@kahliburke said at https://gitter.im/scala-js/scala-js?at=5a0b53da505b630c05e202dc
> So I'm still trying to figure out options for reducing size of my sjsir code and resulting long fastOpt times. It's related to macro code generation, which I understand is not a ScalaJS concern directly. I'm curious about some Array wrapping and type checking code and whether there are any solutions to avoiding it, since the macros generate many many of these sorts of calls and it contributes to the size blowing up.

> 
``` scala
return (function() {
var elements$43 = new $c_sjs_js_WrappedArray().init___sjs_js_Array([$m_Lorg_scalajs_dom_package$().document__Lorg_scalajs_dom_raw_HTMLDocument().createTextNode("\n              ")]);
var this$446 = $m_sjsr_package$();
if ($is_sjs_js_ArrayOps(elements$43)) {
  var x2$70 = $as_sjs_js_ArrayOps(elements$43);
  var jsx$44 = x2$70.scala$scalajs$js$ArrayOps$$array$f
} else if ((elements$43 !== null)) {
  var jsx$44 = elements$43.array$6
} else {
  var result$86 = [];
  var f$102 = new $c_sjsr_AnonFunction1().init___sjs_js_Function1((function($this$162, result$87) {
    return (function(x$2$53) {
      return $uI(result$87.push(x$2$53))
    })
  })(this$446, result$86));
  $f_sc_IndexedSeqOptimized__foreach__F1__V(elements$43, f$102);
  var jsx$44 = result$86
};
return new $c_Lcom_thoughtworks_binding_Binding$Constants().init___sjs_js_Array(jsx$44)
```

Atry:
This PR will minimize the `else` block.

Atry:
Another solution is disabling inlining in `fastOptJS`.

sjrd:
Interestingly, I had just wrote this other commit yesterday evening: https://github.com/sjrd/scala-js/commit/2befe7f8272dc0b496b08c304a207b79c5125e1e, which also solves this particular problem (and more).

However, both solutions fail to correctly optimize the case where 0 arguments are given to a Scala varargs which is then transferred to a JS varargs. See:
```scala
import scala.scalajs.js
import js.annotation._

object HelloWorld {
    println("Hello world!")
    println(myMax(5, 6))
    println(myMax())
  }

  @inline
  def myMax(xs: Double*): Double = js.Math.max(toJSVarArgs(xs): _*)

  @inline
  def toJSVarArgs[A](seq: Seq[A]): js.Array[A] = {
    seq match {
      case seq: js.WrappedArray[A] =>
        seq.array
      case _ =>
        val result = new js.Array[A]
        seq.foreach(x => result.push(x))
        result
    }
  }

  @noinline
  def println(x: Any): Unit = Predef.println(x)
}
```
With the `@inline` and no unhappy path, this generates:
```js
$c_Lhelloworld_HelloWorld$.prototype.main__AT__V = (function(args) {
  this.println__O__V("Hello world!");
  this.println__O__V($uD(Math.max(5.0, 6.0)));
  this.println__O__V($uD(Math.max()))
});
```
which is as good as it gets.

If I remove `@inline`, I get:
```js
$c_Lhelloworld_HelloWorld$.prototype.main__AT__V = (function(args) {
  this.println__O__V("Hello world!");
  this.println__O__V($uD(Math.max(5.0, 6.0)));
  var xs = $m_sci_Nil$();
  var jsx$2 = Math;
  var jsx$3 = jsx$2.max;
  var array = this.toJSVarArgs__sc_Seq__sjs_js_Array(xs);
  var jsx$1 = jsx$3.apply(jsx$2, array);
  this.println__O__V($uD(jsx$1))
});
```
If I use your solution with a no-inline unhappy path, we get:
$c_Lhelloworld_HelloWorld$.prototype.main__AT__V = (function(args) {
  this.println__O__V("Hello world!");
  this.println__O__V($uD(Math.max(5.0, 6.0)));
  var xs = $m_sci_Nil$();
  var jsx$2 = Math;
  var jsx$3 = jsx$2.max;
  var array = this.slowPath$1__p1__sc_Seq__sjs_js_Array(xs);
  var jsx$1 = jsx$3.apply(jsx$2, array);
  this.println__O__V($uD(jsx$1))
});
```
which is only marginally better.

I need to think further about this issue.

**Comment by sjrd** [comment]
Btw, you probably want to target this PR at the 0.6.x branch.

**Comment by Atry** [comment]
@sjrd How about adding a `case Nil =>` for this case?

**Comment by sjrd** [comment]
That doesn't work either because `Nil.equals()` is a complicated beast that is not inlineable (it's not just `this eq that` because it needs to be `==` to other types of sequences).

I also tried the `eq` explicitly with
```scala
  @inline
  def toJSVarArgs[A](seq: Seq[A]): js.Array[A] = {
    seq match {
      case seq: js.WrappedArray[A] =>
        seq.array
      case _ =>
        @noinline def slowPath(): js.Array[A] = {
          val result = new js.Array[A]
          seq.foreach(x => result.push(x))
          result
        }
        if (seq eq Nil)
          js.Array()
        else
          slowPath()
    }
  }
```
but it doesn't work out of the box either:
```js
$c_Lhelloworld_HelloWorld$.prototype.main__AT__V = (function(args) {
  this.println__O__V("Hello world!");
  this.println__O__V($uD(Math.max(5.0, 6.0)));
  var xs = $m_sci_Nil$();
  var jsx$2 = Math;
  var jsx$3 = jsx$2.max;
  var array = ((xs === $m_sci_Nil$()) ? [] : this.slowPath$1__p1__sc_Seq__sjs_js_Array(xs));
  var jsx$1 = jsx$3.apply(jsx$2, array);
  this.println__O__V($uD(jsx$1))
});
```
However for this case I think I can add a rule somewhere in the optimizer to make it do the right thing.

sjrd:
I wonder whether it wouldn't be better to change the compiler to emit a `js.WrappedArray([])` rather than `Nil` right from the start, when given to a Scala varargs. We inherited `Nil` from the behavior of Scala/JVM when giving 0 arguments to a Scala varargs, but I've always questioned whether that was a good idea.

Atry:
`Nil` may be not a good idea even for JVM. It creates a polymorphic type, increasing the cost to inline it on HotSpot JVM.

sjrd:
Yes, I agree. That's precisely the reason I've always thought it was not necessarily a good idea.

sjrd:
Superseded by #3198.

PR #3194, opened 2017-11-18
Fix #3175: Propagate failures in TestAdapter

### LINKED ISSUE #3175: Tests often hang after #3122
I built a local copy of 0.6.22-SNAPSHOT to give #3122 a spin. I tested it on my main project which has 37 modules, 10 of which build with Scala.JS.

Running tests, I often saw failures and then SBT hanging, seemingly unable to recover from the failures. In one case, the failure was simply

```
[error] PhantomJS has crashed. Please read the bug reporting guide at
[error] <http://phantomjs.org/bug-reporting.html> and file a bug report.
```
followed by SBT hanging after the other modules' tests complete.

In another instance the failure looked to be in ScalaJS itself:
```

  phantomjs://code/phantomjs-launcher9060106815877187942.js:9 in onError


  phantomjs://code/phantomjs-launcher9060106815877187942.js:11 in onError
  file:///tmp/shipreq-sjs-22.sbt/webapp-client-project/target/scala-2.12/webapp-client-project-test-fastopt.js:26478 (in function "handleMessage__T__V")

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
  file:///tmp/shipreq-sjs-22.sbt/webapp-client-project/target/scala-2.12/webapp-client-project-test-fastopt.js:181656

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
  file:///tmp/1510141923375-0/comSetup.js:30

  phantomjs://code/phantomjs-launcher9060106815877187942.js:13
```
followed, again, by hanging.

I ran jstack and uploaded the output here:
https://www.dropbox.com/sh/ouqmuqouqtvoivv/AAAgDaapef7LVPHhJyx6LWgca?dl=0

gzm0:


test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -184,8 +202,9 @@ object TestAdapter {
...
 
   private[testadapter] final class ManagedRunner(
-      val id: Long, val runner: ComJSRunner) {
-    val com = new ComJSEnvRPC(runner)
-    val mux = new RunMuxRPC(com)
-  }
+    val id: Long,
+    val runner: ComJSRunner,
+    val com: RPCCore,
+    val mux: RunMuxRPC
```
sjrd: 4 spaces

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -58,49 +61,60 @@ final class TestAdapter(jsEnv: ComJSEnv, config: TestAdapter.Config) {
...
+
+  /** Called when a throwable bubbles up the execution stack.
+   *
+   *  We terminate everyting if this happens to make sure nothing hangs waiting
+   *  on an async operation to complete.
+   */
+  private def reportFailure(cause: Throwable): Unit = {
+    config.logger.error("Failure in async execution. Aborting all test runs.")
+    config.logger.trace(cause)
+    stopEverything()
```
sjrd: IIUC, this causes the `close()` methods of the Com's to be caused, which, due to the added `ComJSRPC.super.close()`, ends up calling `failing.promise.failure(new IOException("Channel got closed"))` in `RPCCore.close()`, which is what trickles into actually failing the sbt test interface waiting for that future. Is that right?

gzm0: Yep. Exactly.

gzm0:
Note to self: There is a race condition in RPCCore for non-expected closing: If a call is issued during an invocation of `close()`, that call will be left dangling.

gzm0:
Updated. Note that I have hardened the error propagation substantially so we can properly track causes. Hopefully this allows users to more easily reason about what happened.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -13,19 +13,33 @@ import java.util.concurrent.atomic.AtomicLong
...
+ *
+ *  Note that it does not have timeout handling for calls. Users are expected to
+ *  manage call teardown by calling [[close]] in case of failure. This typically
+ *  means that subclasses need to put an explicit call to [[close]] once they
+ *  are sure to not call [[handleMessage]] anymore.
+ *
+ *  This class guarantees that dispatch handles synchronously when
+ *  [[handleMessage]] is called, so closing can be performed race-free.
+ */
+private[scalajs] abstract class RPCCore(implicit ex: ExecutionContext) {
```
sjrd: Due to a peculiarity of Scala, classes always need a non-implicit parameter list. This means that this signature in fact desugards to `RPCCore()(implicit ex: ExecutionContext)`. IMO it's better to explicitly write the `()` in that case, to make it obvious.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -102,19 +114,28 @@ private[scalajs] abstract class RPCCore {
...
-    // Actually send message.
-    send(msg)
+    if (closeReason != null) {
+      /* In the meantime, someone closed the channel. Help closing.
+       * We need this check to guard against a race between `call` and `close`.
+       */
+      helpClose()
+    } else {
+      // Actually send message.
+      send(msg)
```
sjrd: What happens if `closeReason` is set to non-null between the `if (closeReason != null)` and the `send(msg)`?

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -150,18 +171,32 @@ private[scalajs] abstract class RPCCore {
     require(old != null, "Endpoint was not attached.")
   }
 
-  /** Close the communication channel. */
-  def close(): Unit = {
+  /** Close the communication channel.
+   *
+   *  This does only affect the current calls (i.e. the client part of the
```
sjrd: "This only affects", I believe.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -150,18 +171,32 @@ private[scalajs] abstract class RPCCore {
...
+    closeReason = reason
+    helpClose()
+  }
+
+  private def helpClose(): Unit = {
     /* Fix for #3128: explicitly upcast to java.util.Map so that the keySet()
      * method is binary compatible on JDK7.
      */
     val pendingCallIDs = (pending: java.util.Map[Long, _]).keySet()
+    val exception = new IOException(s"RPC channel got closed", closeReason)
```
sjrd: Unnecessary `s`

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
gzm0: Then whatever just set `closeReason` will continue to call `helpClose`. There it will see this invocation in `pending` and race with the server to either fail or succeed it. In both cases, the pending call will not be left dangling.

sjrd: OK thanks.

gzm0:
Ah, ah, my hack for when we don't have testadapter on the classpath doesn't work anymore :)

gzm0:
Jenkins, test this please

sjrd:
There's one binary incompatibility on 2.11:
```
[info] Scala.js test interface: found 1 potential binary incompatibilities while checking against org.scala-js:scalajs-test-interface_2.11:0.6.21  (filtered 3)
[error]  * method close()Unit in object org.scalajs.testinterface.internal.JSRPC#Com does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.scalajs.testinterface.internal.JSRPC#Com.close")
```

gzm0:
Grr.... I do not get this one locally :( Probably need to clean.

gzm0:
Cleaning did not help, but I just added it.

PR #3197, opened 2017-11-24
Use sbt-platform-deps to provide `%%%` and drop sbt-crossproject.

sjrd:
This means that sbt-scalajs does not directly provide `crossProject` nor `JSPlatform` anymore. Those are provided by sbt-scalajs-crossproject 0.3.0 or later.


sbt-plugin-test/build.sbt
```diff
@@ -105,10 +116,10 @@ lazy val multiTest = crossProject.
...
     },
 
-    // Test crossPlatform (as a setting, it's evaluated when loading the build)
-    crossPlatform ~= { value =>
-      assert(value == JSPlatform,
-          "crossPlatform should be JSPlatform in multiTestJS")
+    // Test platformDepsCrossVersion (as a setting, it's evaluated when loading the build)
+    platformDepsCrossVersion ~= { value =>
+      assert(value eq ScalaJSCrossVersion.binary,
+          "platformDepsCrossVersion should be ScalaJSCrossVersion.binary in multiTestJS")
```
gzm0: I think that test can be removed, no?

sbt-plugin-test/build.sbt
```diff
@@ -125,36 +136,25 @@ lazy val multiTest = crossProject.
...
-      assert(value == JVMPlatform,
-          "crossPlatform should be JVMPlatform in multiTestJVM")
+    // Test platformDepsCrossVersion (as a setting, it's evaluated when loading the build)
+    platformDepsCrossVersion := {
+      val value = platformDepsCrossVersion.value
+      if (!sbtVersion.value.startsWith("0.")) {
+        // In 0.13, CrossVersions do not have a meaningful ==, but they do in 1.0
+        assert(value == CrossVersion.binary,
+            "platformDepsCrossVersion should be CrossVersion.binary in multiTestJVM")
+      }
```
gzm0: This too, no? Or what is this testing?

sbt-plugin-test/build.sbt
```diff
@@ -69,33 +67,46 @@ lazy val noDOM = project.settings(baseSettings: _*).
...
+  unmanagedSourceDirectories in Test +=
+    baseDirectory.value.getParentFile / "shared/src/test/scala",
+
+  testFrameworks ++= Seq(
+      TestFramework("sbttest.framework.DummyFramework"),
+      TestFramework("inexistent.Foo", "another.strange.Bar")
+  )
+)
 
-lazy val multiTest = crossProject.
```
gzm0: Just to check my understanding: We cannot depend on sbt-crossproject for this, since it will depend on SJS?

sjrd: That's right. Or to be more precise, `sbt-scalajs-crossproject` will (already does) depend on SJS. `sbt-crossproject` per se only depends on `sbt-platform-deps`, but also does not provide `JSPlatform` (which is provided by `sbt-scalajs-crossproject`).

sbt-plugin-test/build.sbt
sjrd: No, it is the only test that makes sure that [this line](https://github.com/scala-js/scala-js/pull/3197/files#diff-dd593403434ec9de3670cb5722f10171R452) has the intended effect (and does not go away).

The tests of sbt-platform-deps, in turn, ensure that if `platformDepsCrossVersion` is correctly set up, then `%%%` works as expected. But it is the responsibility of `ScalaJSPlugin` to set up `platformDepsCrossVersion`, and therefore we need to test that.

sbt-plugin-test/build.sbt
sjrd: This one is mostly by symmetry with the other one. Arguably it's also testing that the setup of `platformDepsCrossVersion` in `ScalaJSPlugin` does not affect other projects (like, if we erroneously put it in `override def globalSettings` instead of `projectSettings`).

sjrd:
So, is there still something you want me to change here?

PR #3198, opened 2017-11-24
Do not inline `.toJSArray` to prevent code size blowing up.

sjrd:
For an explicit `.toJSArray`, it is typically not advantageous to inline the method unless the exact type of collection is known. And in most of those cases, the collection would be stack-allocated, which would force inline anyway.

In this commit, we directly implement `.toJSArray` (instead of depending on a method of `runtime`) and mark it as non-inlined.

The method `runtime.genTraversableOnce2jsArray`, used by the code gen for JS varargs, is unchanged.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/JSConverters.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSConverters.scala
@@ -39,7 +39,28 @@ object JSConverters extends JSConvertersLowPrioImplicits {
 
   implicit class JSRichGenTraversableOnce[T](
       val col: GenTraversableOnce[T]) extends AnyVal {
-    @inline final def toJSArray: Array[T] = genTraversableOnce2jsArray(col)
+    final def toJSArray: Array[T] = {
+      /* This is basically a duplicate of `runtime.genTraversableOnce2jsArray`,
+       * except it is not marked `@inline`. We do not want to inline this
+       * method every time someone does `.toJSArray`, for code size reasons
+       * (unlike `genTraversableOnce2jsArray`, which is used by the codegen for
+       * transferring Scala varargs to JS varargs).
+       *
+       * One would think that we could still delegate to
+       * `genTraversableOnce2jsArray` and mark `toJSArray` with `@noinline`
+       * instead, but that would prevent `toJSArray` to be inlined even when
+       * `col` is stack-allocated (and we do want that to happen as in that
+       * case the entire match disappears and `col` can stay stack-allocated).
+       */
+      col match {
+        case col: js.ArrayOps[T]     => col.repr
+        case col: js.WrappedArray[T] => col.array
+        case _ =>
+          val result = new js.Array[T]
+          col.foreach(x => result.push(x))
+          result
+      }
+    }
   }
 
   implicit class JSRichGenIterable[T](


library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -39,7 +39,16 @@ object JSConverters extends JSConvertersLowPrioImplicits {
 
   implicit class JSRichGenTraversableOnce[T](
       val col: GenTraversableOnce[T]) extends AnyVal {
-    @inline final def toJSArray: Array[T] = genTraversableOnce2jsArray(col)
+    final def toJSArray: Array[T] = {
```
gzm0: I think it might be worth adding the explanation why this duplicates `getTraversableOnce2jsArray` in a comment here.

library/src/main/scala/scala/scalajs/js/JSConverters.scala
```diff
@@ -39,7 +39,16 @@ object JSConverters extends JSConvertersLowPrioImplicits {
...
+      col match {
+        case col: js.ArrayOps[T]     => col.repr
+        case col: js.WrappedArray[T] => col.array
+        case _ =>
+          val result = new js.Array[T]
+          col.foreach(x => result.push(x))
+          result
+      }
+    }
```
gzm0: So just to check my understanding: This behaves better than before in case `col` is stack allocated?

sjrd: No. In case `col` is stack allocated, this performs *exactly* the same as before. The difference is when `col` is *not* stack allocated, in which case the new version produces less code than before by not inlining the method.

gzm0: Ah, right. Because `genTraversableOnce2jsArray` is inlineable as well. OK, got it.

sjrd:
Updated with a comment to explain the duplication.

sjrd:
Got this twice on the CI:
```
[info] Scala.js sbt test adapter: found 2 potential binary incompatibilities while checking against org.scala-js:scalajs-sbt-test-adapter_2.10:0.6.21  (filtered 4)
[error]  * method com()org.scalajs.testadapter.ComJSEnvRPC in class org.scalajs.testadapter.TestAdapter#ManagedRunner has a different result type in current version, where it is org.scalajs.testcommon.RPCCore rather than org.scalajs.testadapter.ComJSEnvRPC
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.scalajs.testadapter.TestAdapter#ManagedRunner.com")
[error]  * method this(Long,org.scalajs.jsenv.ComJSRunner)Unit in class org.scalajs.testadapter.TestAdapter#ManagedRunner does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.scalajs.testadapter.TestAdapter#ManagedRunner.this")
```
That's on Java 6, Scala 2.10.7. I don't know why it passed the CI of #3194 (it actually failed in one job, then not on the rerun of that job :-s), nor why it succeeds for Scala 2.10.7 on JDK 7 or 8. This makes no sense. However, I think the only reasonable thing to do right now is filter those out to fix the CI.

sjrd:
Rebased after #3203.

PR #3199, opened 2017-11-24
Fix #3192: Do not reexecute the constructor in `Object.clone()`.

### LINKED ISSUE #3192: Object.clone() reexecutes the contructor of classes with inlineable inits
This is a regression in 1.x caused by the inlineable inits. Test case:
```scala
  def cloneCtorSideEffectsBug(): Unit = {
    var ctorInvokeCount = 0

    // This class has an inlineable init
    class CloneCtorSideEffectsBug(val x: Int) extends java.lang.Cloneable {
      ctorInvokeCount += 1

      override def clone(): CloneCtorSideEffectsBug =
        super.clone().asInstanceOf[CloneCtorSideEffectsBug]
    }

    val o = new CloneCtorSideEffectsBug(54)
    assert(o.x == 54)
    assert(ctorInvokeCount == 1)

    val o2 = o.clone()
    assert(o2.x == 54)
    assert(ctorInvokeCount == 1, "boom")
  }
```
which fails with:
```
[error] TypeError: Cannot read property 'elem$1' of undefined
[error]     at new $c_Lhelloworld_HelloWorld$CloneCtorSideEffectsBug$1 (HelloWorld.scala:22:7)
[error]     at $c_Lhelloworld_HelloWorld$.cloneObject__O__O (HelloWorld.scala:39:61)
[error]     at $c_Lhelloworld_HelloWorld$.cloneCtorSideEffectsBug__V (HelloWorld.scala:25:20)
[error]     at $c_Lhelloworld_HelloWorld$.main__AT__V (HelloWorld.scala:16:28)
[error]     ...
```
This is because the implementation of `Object.clone()`, which is in `runtime.cloneObject()`, has been broken by inlineable inits. The current implementation is at https://github.com/scala-js/scala-js/blob/f9a87328040efa5348455110f90c7be8a7237cc1/library/src/main/scala/scala/scalajs/runtime/package.scala#L19-L27
In that implementation we invoke `new from.constructor()`, which used to be OK because the constructors were not doing anything side-effectful. However, inlined inits have broken that assumption.

In ES2017 (!) there is a new function [`Object.getOwnPropertyDescriptors`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors) which allows to implement a high-fidelity surface clone of any (non-exotic) JavaScript object. It is even listed in the examples:
```js
Object.create(
  Object.getPrototypeOf(obj), 
  Object.getOwnPropertyDescriptors(obj) 
);
```
There is a polyfill available at https://github.com/tc39/proposal-object-getownpropertydescriptors#proposed-solution, but it uses `Reflect.ownKeys()`, which is an ES2015 thing. We can in turn polyfill `Reflect.ownKeys()` with `Object.getOwnPropertyNames()` (ES5) + `Object.getOwnPropertySymbols()` (ES2015 too, but would return `[]` in ES 5.1 since there are no symbols anyway).

Here is a work-in-progress implementation in Scala.js that assumes ES 2015 (it does not polyfill `Reflect.ownKeys`):
```scala
  def cloneObject(o: Object): Object = {
    js.Object.create(js.Object.getPrototypeOf(o.asInstanceOf[js.Object]),
        getOwnPropertyDescriptors(o))
  }

  def getOwnPropertyDescriptors(o: Object): js.Object = {
    val jso = o.asInstanceOf[js.Object]
    val ownKeys = js.Dynamic.global.Reflect.ownKeys(jso).asInstanceOf[js.Array[js.Any]]
    val descriptors = new js.Object
    for (ownKey <- ownKeys) {
      js.Dynamic.global.Object.defineProperty(descriptors, ownKey, js.Dynamic.literal(
          configurable = true,
          enumerable = true,
          writable = true,
          value = js.Dynamic.global.Object.getOwnPropertyDescriptor(jso, ownKey)
      ))
    }
    descriptors
  }
```

sjrd:
We now use the "official" way to create a shallow copy of a JavaScript object, which uses a combination of `Object.create`, `Object.getPrototypeOf` and `Object.getOwnPropertyDescriptors`.

The latter was introduced in ECMAScript 2017, though, so we need to polyfill it, and along with it two other functions from ECMAScript 2015.

Locally tested with variants that ensured that all code paths were taken and correct. Including `testSuite/testHtml` in Firefox, which already implements `Object.getOwnPropertyDescriptors` itself.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/ObjectClone.scala
+++ b/javalanglib/src/main/scala/java/lang/ObjectClone.scala
@@ -7,6 +7,76 @@ import scala.scalajs.js
  *  Called by the hard-coded IR of `java.lang.Object`.
  */
 private[lang] object ObjectClone {
+  private val getOwnPropertyDescriptors: js.Function1[js.Object, js.Object] = {
+    import js.Dynamic.{global, literal}
+
+    // The val f = ...; f.asInstanceOf's are necessary for 2.10 not to crash
+
+    // Fetch or polyfill Object.getOwnPropertyDescriptors
+    if (js.typeOf(global.Object.getOwnPropertyDescriptors) == "function") {
+      val f = global.Object.getOwnPropertyDescriptors
+      f.asInstanceOf[js.Function1[js.Object, js.Object]]
+    } else {
+      // Fetch or polyfill Reflect.ownKeys
+      type OwnKeysType = js.Function1[js.Object, js.Array[js.Any]]
+      val ownKeysFun: OwnKeysType = {
+        if (js.typeOf(global.Reflect) != "undefined" &&
+            js.typeOf(global.Reflect.ownKeys) == "function") {
+          val f = global.Reflect.ownKeys
+          f.asInstanceOf[OwnKeysType]
+        } else {
+          // Fetch Object.getOwnPropertyNames
+          val getOwnPropertyNames = {
+            val f = global.Object.getOwnPropertyNames
+            f.asInstanceOf[OwnKeysType]
+          }
+
+          // Fetch or polyfill Object.getOwnPropertySymbols
+          val getOwnPropertySymbols: OwnKeysType = {
+            if (js.typeOf(global.Object.getOwnPropertySymbols) == "function") {
+              val f = global.Object.getOwnPropertySymbols
+              f.asInstanceOf[OwnKeysType]
+            } else {
+              /* Polyfill for Object.getOwnPropertySymbols.
+               * We assume that if that function does not exist, then symbols
+               * do not exist at all. Therefore, the result is always an empty
+               * array.
+               */
+              { (o: js.Object) =>
+                js.Array[js.Any]()
+              }
+            }
+          }
+
+          // Polyfill for Reflect.ownKeys
+          { (o: js.Object) =>
+            getOwnPropertyNames(o) ++ getOwnPropertySymbols(o)
+          }
+        }
+      }
+
+      // Polyfill for Object.getOwnPropertyDescriptors
+      { (o: js.Object) =>
+        val ownKeys = ownKeysFun(o)
+        val descriptors = new js.Object
+        for (key <- ownKeys) {
+          /* Almost equivalent to the JavaScript code
+           *   descriptors[key] = Object.getOwnPropertyDescriptor(descriptors, key);
+           * except that `defineProperty` will by-pass any existing setter for
+           * the property `key` on `descriptors` or in its prototype chain.
+           */
+          global.Object.defineProperty(descriptors, key, literal(
+              configurable = true,
+              enumerable = true,
+              writable = true,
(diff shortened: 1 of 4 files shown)


gzm0:
QQ: Could we do the same thing for `Object` than for `String`? Seems like the only other hack we need would be for the super class?

sjrd:
There are several other hacks that we would need for `Object`:

* The super class, as you mentioned
* The implementation of `final` methods of `AnyRef` (`wait`, `getClass`, etc.)
* The fact that the implementation of `getClass()` is hard-coded as the IR `GetClass(This())`, but we don't have any user-space way to emit the IR node `GetClass`.

Scala Native does it, [but it's not pretty](https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/scala/java/lang/Object.scala). I'm not convinced that it is better than hard-coding the IR.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ObjectTest.scala
```diff
@@ -74,4 +74,24 @@ class ObjectTest {
...
+        super.clone().asInstanceOf[CloneCtorSideEffectsBug]
+    }
+
+    val o = new CloneCtorSideEffectsBug(54)
+    assert(o.x == 54)
+    assert(ctorInvokeCount == 1)
+
+    val o2 = o.clone()
+    assert(o2.x == 54)
+    assert(ctorInvokeCount == 1, "boom")
```
gzm0: I think the `"boom"` can go now that it is fixed :)

javalanglib/src/main/scala/java/lang/ObjectClone.scala
```diff
@@ -0,0 +1,85 @@
...
+        val ownKeys = ownKeysFun(o)
+        val descriptors = new js.Object
+        for (key <- ownKeys) {
+          global.Object.defineProperty(descriptors, key, literal(
+              configurable = true,
+              enumerable = true,
+              writable = true,
+              value = global.Object.getOwnPropertyDescriptor(o, key)
+          ))
+        }
```
gzm0: How is this different than just assigning the property descriptor to `descriptors`?

sjrd: It is equivalent, *except* if there is an accessor property named `key` in `o` or its prototype chain. If that's the case, the assignment calls the setter, whereas `defineProperty` by-passes the (inherited) setter and creates a data property on `o`.

Since `descriptors` is a brand `new js.Object`, this can only happen if there is an accessor property on `Object.prototype`, which is unlikely but possible.

gzm0: Ah, fair enough. Maybe add a comment to explain this? It's kind of non-obivous :)

sjrd:
Updated.

PR #3215, opened 2017-12-07
Reintroduce tests for our EMCAScript 2015 polyfills.

sjrd:
When we removed support for Rhino, and moved the support of PhantomJS in a separate repository, we accidently got rid of all the tests in our build checking our polyfills for ES 2015 features. For example, the polyfills for `Math.imul`, or the floating-point bits conversions (including `Double.hashCode`).

This commit restores that by introducing in our build a custom `NodeJSEnvForcePolyfills`, which is like `NodeJSEnv` but forcefully `delete`s all the good stuff from ES 2015.

This ensures that we do not introduce regressions in our polyfills in the future.

### DIFF
--- a/ci/matrix.xml
+++ b/ci/matrix.xml
@@ -83,6 +83,34 @@
         ++$scala $testSuite/test
   ]]></task>
 
+  <task id="test-suite-ecma-script5-force-polyfills"><![CDATA[
+    setJavaVersion $java
+    npm install &&
+    sbtretry 'set jsEnv in $testSuite := new NodeJSEnvForcePolyfills()' \
+        ++$scala $testSuite/test &&
+    sbtretry 'set jsEnv in $testSuite := new NodeJSEnvForcePolyfills()' \
+        'set scalaJSStage in Global := FullOptStage' \
+        ++$scala $testSuite/test \
+        $testSuite/clean &&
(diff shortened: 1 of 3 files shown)


ci/matrix.xml
```diff
@@ -280,6 +308,13 @@
       <v n="testSuite">testSuite</v>
     </run>
 
+    <!-- Test suite on ECMAScript5 with forced polyfills tasks -->
+    <run task="test-suite-ecma-script5-force-polyfills">
+      <v n="scala">2.12.4</v>
+      <v n="java">1.8</v>
+      <v n="testSuite">testSuite</v>
```
gzm0: This parameter is unneeded.

sjrd: Hum ... I disagree. It's actually used all over in the task's script. In theory we could call that script with `<v n="testSuite">scalaTestSuite</v>`.

Are you saying that because I don't have another `<run>` for that task with a different `testSuite`, which means I could "inline" the value in the script (instead of `$testSuite`)? But if that's the case, we should also apply this reasoning to `scala` and `java`. For `java`, we should even replace it in the entire matrix since we only have `1.8` at the moment.

gzm0: Ok, fair enough.

PR #3220, opened 2017-12-12
Fix #2949: Remove the CLI.

### LINKED ISSUE #2949: Move the CLI out of the core
This could be controversial but ... what about moving the CLI out of the core? Rationale:

* It does not need to be in core. We build and test everything else without requiring the CLI. By our general policy of "everything that does not need to be in core shouldn't be in the core", the CLI shouldn't be there.
* It is the only thing that depends on a non-critical Scala dependency: `scopt`. This means we typically cannot even build the CLI for new milestones of Scala at the time we upgrade. And after that, it's kind of too late to publish them.
* It is also the only thing that makes us depend on the sbt plugin `sbt-assembly`.
* Out of the core, we could consider using more fancy things, like [`sbt-native-packager`](https://github.com/sbt/sbt-native-packager), which means we might be able to fix #703, our oldest open issue.

FTR, other Scala--`%%`--dependencies that we have are:

* `scala-partest` which is super core to scala/scala, so guaranteed to be published in a timely manner, and
* `scala-parallel-collections`, which currently is considered critical enough to be published in a timely manner, but we'll probably have to move away from those at some point too.

and other sbt plugins we use are:

* `sbt-bintray`
* `sbt-mima-plugin`
* `scalastyle-sbt-plugin`

sjrd:
The CLI has been ported over to the separate repository https://github.com/scala-js/scala-js-cli.


RELEASING.md
```diff
@@ -16,17 +16,15 @@
 1. Perform [manual testing][3] that needs the tagging (source maps).
 1. Publish:
     - Sonatype, bintray (`./script/publish.sh`)
-    - CLI to website (simple `scp`)
+    - [Publish the CLI][4]
```
gzm0: Won't they follow independent release cycles?

sjrd: Yes and no. We still need to publish a new `scalajs-cli` artifact with version number, say, 1.0.2, that depends on `scalajs-tools` 1.0.2, for easy consumption.

At least that's how I see things in the early days of 1.x. We can relax this coupling later.

PR #3217, opened 2017-12-08
Make java.lang.Class agnostic of Scala

sjrd:

PR #3221, opened 2017-12-12
Drop support for Scala 2.10.x.

sjrd:
In this commit, we completely drop support for Scala 2.10.x in Scala.js code. Note that our JVM artifacts are still cross-compiled with 2.10.x, because we still support sbt 0.13.16+ which runs on 2.10. However, it will not be possible to use `enablePlugins(ScalaJSPlugin)` with `scalaVersion := "2.10.x"` anymore.

This includes:

* Drop CI tasks using 2.10.x with JS artifacts (but not with JVM artifacts).
* Remove all workarounds for Scala 2.10 compiler bugs, such as crashes for some usages of `js.Dynamic`.
* Remove all 2.10-specific stuff from `CompatComponent` in the compiler.
* Remove all 2.10-specific assumptions or branches in the test suite.

This commit does not include:

* Drop `js.Dynamic.x` and the associated compiler tricks, which work around the 2.10 *library* (not compiler), and deserve a separate commit anyway.
* Take advantage of new 2.11 language features, such as making the underlying val of value classes private.

Those items are left for further commits.


sjrd:
Nightly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/5975/
Weekly: https://scala-webapps.epfl.ch/jenkins/job/scalajs-matrix-build/5976/

sjrd:
Nightly and Weekly are good, so it's ready for review.

sjrd:
Rebased to address the trivial conflict in `ci/checksizes.sh`.

sjrd:
Rebased and adapted on top of #3217.

ci/matrix.xml
```diff
@@ -159,6 +159,7 @@
   <task id="bootstrap"><![CDATA[
     setJavaVersion $java
     npm install &&
+    sbt ++$scala tools/test &&
```
gzm0: I do not understand why the `tools` (the JVM part IIUC) move here.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -142,7 +142,6 @@ class TreeSet[E] (_comparator: Comparator[_ >: E])
     val boxedFrom = Box(fromElement)
     val boxedTo = Box(toElement)
     val subSetFun = { () =>
-      // the creation of a new TreeSet is to avoid a mysterious bug with scala 2.10
```
gzm0: Hmmm... What happens if you set `var base = inner.range(boxedFrom, boxedTo)`?

sjrd: `range` does not seem to guarantee that it returns a fresh `TreeSet`, independent of `inner`. So in theory it is possible that we receive something that will be mutated under our feet if we do not force the copy into a new `TreeSet`.

I haven't observed the behavior, but it could happen in future versions of the Scala lib in any case.

ci/matrix.xml
sjrd: Because of https://github.com/scala-js/scala-js/blob/98d3b4e9e1182cef89fda312218cee4b212e3cba/project/Build.scala#L594-L597, which means `tools/test` depends on `library/packageBin` (for the same `scalaVersion`), and therefore we cannot do `tools/test` in 2.10 anymore.

sjrd:
Rebased (without conflict) and fixed the commit message.

ci/matrix.xml
gzm0: Fair enough :)

PR #3224, opened 2017-12-13
Fix #3219: Correctly recognize exports for `$xyz` in the back-end.

### LINKED ISSUE #3219: Methods exported under a scalac name starting with `$` are not exported
```scala
import scala.scalajs.js.annotation._

@JSExportTopLevel("Functions")
object Functions {
  @JSExport("$a")
  def f(x: Int): Int = x

  @JSExport
  def +(x: Int): Int = x

  @JSExport("-")
  def plus(x: Int): Int = x

  // OK
  @JSExport("plus")
  def ++(x: Int): Int = x
}
```
Only the last function is actually exported. All the other ones vanish. The problem is already apparent in the IR:
```scala
> scalajsp Functions$.sjsir
module class LFunctions$ extends O {
  def f__I__I(x: int): int = {
    x
  }
  def $$plus__I__I(x: int): int = {
    x
  }
  def plus__I__I(x: int): int = {
    x
  }
  def $$plus$plus__I__I(x: int): int = {
    x
  }
  def $$js$exported$meth$$a__I__O(x: int): any = {
    this.f__I__I(x)
  }
  def $$js$exported$meth$$plus__I__O(x: int): any = {
    this.$$plus__I__I(x)
  }
  def $$js$exported$meth$$minus__I__O(x: int): any = {
    this.plus__I__I(x)
  }
  def $$js$exported$meth$plus__I__O(x: int): any = {
    this.$$plus$plus__I__I(x)
  }
  def init___() {
    this.O::init___();
    mod:LFunctions$<-this
  }
  def "plus"(arg$1: any): any = {
    val prep0: int = arg$1.asInstanceOf[I];
    this.$$js$exported$meth$plus__I__O(prep0)
  }
  export top module "Functions"
}
```

sjrd:
We were erroneously using `sym.unexpandedName` when testing whether a method is an export bridge. However, if the export name happens to start with a `$`, then `unexpandedName` considers the whole thing as an expanded name, and removes the whole prefix as if it were the result of an expansion.

Simply using the `name` for the test fixes the issue. Note that `jsExportInfo` was already using `name`, and not `unexpandedName`.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSGlobalAddons.scala
@@ -109,8 +109,7 @@ trait JSGlobalAddons extends JSDefinitions
 
     /** checks if the given symbol is a JSExport */
     def isExport(sym: Symbol): Boolean =
-      sym.unexpandedName.startsWith(exportPrefix) &&
-      !sym.hasFlag(Flags.DEFAULTPARAM)
+      sym.name.startsWith(exportPrefix) && !sym.hasFlag(Flags.DEFAULTPARAM)
 
     /** retrieves the originally assigned jsName of this export and whether it
      *  is a property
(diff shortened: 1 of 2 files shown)


sjrd:
Commit message fixed.

PR #3226, opened 2017-12-17
Support Charset.aliases() method

cquiroz:
This is a small addition to `Charset` to support the method `aliases()` basically exposing the `aliases` field in the right format.

This is the method in particular:
https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html#aliases()

### DIFF
--- a/javalib/src/main/scala/java/nio/charset/Charset.scala
+++ b/javalib/src/main/scala/java/nio/charset/Charset.scala
@@ -1,12 +1,19 @@
 package java.nio.charset
 
 import java.nio.{ByteBuffer, CharBuffer}
+import java.util.{Collections, HashSet, Arrays}
 
 import scala.scalajs.js
 
 abstract class Charset protected (canonicalName: String,
-    aliases: Array[String]) extends AnyRef with Comparable[Charset] {
+    _aliases: Array[String]) extends AnyRef with Comparable[Charset] {
+
+  private lazy val aliasesSet =
+    Collections.unmodifiableSet(new HashSet(Arrays.asList(_aliases)))
+
   final def name(): String = canonicalName
+
+  final def aliases(): java.util.Set[String] = aliasesSet
 
   override final def equals(that: Any): Boolean = that match {
     case that: Charset => this.name == that.name
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/CharsetTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/CharsetTest.scala
@@ -7,6 +7,8 @@
 \*                                                                      */
 package org.scalajs.testsuite.niocharset
 
+import scala.collection.JavaConverters._
+
 import java.nio.charset._
 
 import org.junit.Test
@@ -16,6 +18,7 @@ import org.scalajs.testsuite.utils.AssertThrows._
 import org.scalajs.testsuite.utils.Platform.executingInJVM
 
 class CharsetTest {
+  def javaSet[A](elems: A*): java.util.Set[A] = Set(elems: _*).asJava
 
   @Test def defaultCharset(): Unit = {
     assertSame("UTF-8", Charset.defaultCharset().name())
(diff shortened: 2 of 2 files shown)


javalib/src/main/scala/java/nio/charset/Charset.scala
```diff
@@ -8,6 +8,9 @@ abstract class Charset protected (canonicalName: String,
     aliases: Array[String]) extends AnyRef with Comparable[Charset] {
```
sjrd: Could you rename this to `_aliases` or `aliasesArray`, so that there isn't a semi-clash between the private `aliases` and the `def aliases()`?

javalib/src/main/scala/java/nio/charset/Charset.scala
```diff
@@ -8,6 +8,9 @@ abstract class Charset protected (canonicalName: String,
...
   final def name(): String = canonicalName
 
+  final def aliases(): java.util.Set[String] =
+    new java.util.HashSet(java.util.Arrays.asList(aliases))
```
sjrd: The Javadoc says that `aliases()` must return an *immutable* set of the aliases. Therefore, you should wrap the produced `HashSet` with [`java.util.Collections.unmodifiableSet`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableSet-java.util.Set-). This also means that you can always reuse the same unmodifiable set, which you can store in a `private lazy val aliasesSet`. Make sure to make it lazy so that all this stuff can be dce'ed away if `aliases()` is not called.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/CharsetTest.scala
```diff
@@ -67,4 +69,24 @@ class CharsetTest {
 
     assertFalse(Charset.isSupported("this-charset-does-not-exist"))
   }
+
+  @Test def aliases(): Unit = {
+    assertEquals(Charset.forName("UTF-8").aliases(),
+      List("UTF8", "unicode-1-1-utf-8").toSet.asJava)
```
sjrd: Use `Set(...)` instead of `List(...).toSet`. Even better, define a local
```scala
def javaSet[A](elems: A*): java.util.Set[A] = Set(elems: _*).asJava
```
and then use `javaSet("UTF8", "unicode-1-1-utf-8")`.

sjrd: Also, we use 4 spaces indentation when continuing a line. See our coding style at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#indentation

cquiroz:
Thanks for your comments. I rebased to 0.6.x, addressed your concerns and squashed
Please let me know if you need more changes

How are these commits added to master?

sjrd:

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/CharsetTest.scala
```diff
@@ -67,4 +70,24 @@ class CharsetTest {
 
     assertFalse(Charset.isSupported("this-charset-does-not-exist"))
   }
+
+  @Test def aliases(): Unit = {
+    assertEquals(Charset.forName("UTF-8").aliases(),
+      javaSet("UTF8", "unicode-1-1-utf-8"))
```
sjrd: This is in fact also the continuation of the line starting with `assertEquals`, so this should be indented 4 spaces.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/CharsetTest.scala
```diff
@@ -67,4 +70,24 @@ class CharsetTest {
...
+    assertEquals(Charset.forName("UTF-16").aliases(),
+      javaSet("UTF_16", "unicode", "utf16", "UnicodeBig"))
+    assertEquals(Charset.forName("UTF-16BE").aliases(),
+      javaSet("X-UTF-16BE", "UTF_16BE", "ISO-10646-UCS-2",
+          "UnicodeBigUnmarked"))
```
sjrd: Here, you'll have 4 spaces of indentation on top of 4 spaces of indentation, so at the end of the day those lines should look like:
```scala
    assertEquals(Charset.forName("UTF-16BE").aliases(),
        javaSet("X-UTF-16BE", "UTF_16BE", "ISO-10646-UCS-2",
            "UnicodeBigUnmarked"))
```

cquiroz:
Ok, I hope I got it right this time 😄 
Changes were squashed

sjrd:

PR #3229, opened 2017-12-19
Fix #3228: Correctly identify module classes for refl. instantiation.

### LINKED ISSUE #3228: Linking error with an @EnableReflectiveInstantiation object inside a class
The following code causes a linking error:
```scala
class Foo {
  @EnableReflectiveInstantiation
  object Bar
}
```
The error is:
```
[error] Cannot access module for non-module helloworld.Foo$Bar$
[error]   called from helloworld.Foo$Bar$.<clinit>()
```
This happens because, as a nested `object`, `Bar` is not really a module class, but the codegen for `@EnableReflectiveInstantiation` believes `Bar$` is a module class, producing the following erroneous IR:
```scala
class Lhelloworld_Foo$Bar$ extends O {
  def init___Lhelloworld_Foo($$outer: Lhelloworld_Foo) {
    this.O::init___()
  }
  static def clinit___() {
    mod:sjs_reflect_Reflect$.registerLoadableModuleClass__T__jl_Class__sjs_js_Function0__V("helloworld.Foo$Bar$", classOf[Lhelloworld_Foo$Bar$], (lambda<>() = {
      mod:Lhelloworld_Foo$Bar$
    }))
  }
}
```

sjrd:
We were using `sym.isModuleClass` to identify module classes when generating the reflective instantiation code, but that also matches nested objects. The appropriate predicate is `isStaticModule(sym)`.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
@@ -55,6 +55,11 @@ class ReflectTest {
   private final val NameTraitDisable =
     Prefix + "TraitDisable"
 
+  private final val NameInnerObject = {
+    Prefix + "ClassWithInnerObjectWithEnableReflectiveInstantiation$" +
+    "InnerObjectWithEnableReflectiveInstantiation"
+  }
+
   @Test def testClassRuntimeClass(): Unit = {
     for {
       name <- Seq(NameClassEnableDirect, NameClassEnableDirectNoZeroArgCtor,
@@ -187,6 +192,23 @@ class ReflectTest {
     }
   }
 
+  @Test def testInnerObjectWithEnableReflectiveInstantiation_issue_3228(): Unit = {
+    assertFalse(Reflect.lookupLoadableModuleClass(NameInnerObject).isDefined)
+
+    val optClassData = Reflect.lookupInstantiatableClass(NameInnerObject)
+    assertTrue(optClassData.isDefined)
+    val classData = optClassData.get
+
+    val optCtor = classData.getConstructor(
+        classOf[ReflectTest.ClassWithInnerObjectWithEnableReflectiveInstantiation])
+    assertTrue(optCtor.isDefined)
+
+    /* We do not actually instantiate that, because it would violate the
+     * uniqueness of the inner object wrt. its enclosing class, which is not
+     * well-defined at the Scala level.
+     */
+  }
+
 }
 
 object ReflectTest {
@@ -307,4 +329,11 @@ object ReflectTest {
   }
 
   trait TraitDisable extends Accessors
+
+  // Regression cases
+
+  class ClassWithInnerObjectWithEnableReflectiveInstantiation {
+    @EnableReflectiveInstantiation
+    object InnerObjectWithEnableReflectiveInstantiation
+  }
 }
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -941,7 +941,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     private def genRegisterReflectiveInstantiation(sym: Symbol)(
         implicit pos: Position): Option[js.Tree] = {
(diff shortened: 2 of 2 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
```diff
@@ -187,6 +192,23 @@ class ReflectTest {
...
+    val classData = optClassData.get
+
+    val optCtor = classData.getConstructor(
+        classOf[ReflectTest.ClassWithInnerObjectWithEnableReflectiveInstantiation])
+    assertTrue(optCtor.isDefined)
+
+    /* We do not actually instantiate that, because it would violate the
+     * uniqueness of the inner object wrt. its enclosing class, which is not
+     * well-defined at the Scala level.
+     */
```
gzm0: This seems like a major oversight in the design of reflective instantiation. Should we even allow to reflectively instantiate non-static things? I understand that Java reflection will totally allow you to do this, but that doesn't mean we should support it.

sjrd: The alternative would be not to register non-static module classes, and therefore test that both `lookupLoadableModuleClass` and `lookupInstantiatableClass` return `None`. Would you prefer that? Or should we emit a compile error (that might be seriously limiting in some cases).

Note that reflectively instantiating non-static *classes* is perfectly well-defined, so I think we should not disallow that.

gzm0: I would suggest to extend the reflection API to allow reflective access to non-static modules (essentially a forwarder to the module accessor).

Not sure if we should extend `LoadableModuleClass` or add another type.

sjrd: I'm not sure how I feel about that. It might be the "best" solution, but the benefit/cost is close to 0. Nobody would actually use that API, AFAICT, and implementing it is non-trivial (need to recover the enclosing instance, find the right getter for the `object`, etc.).

gzm0: In that case, let's just not register it (and hence return `None`). However, we should probably emit a warning, otherwise this might cause some hard to debug problems (think of pass a `Class` to something, it fails to load this thing, why???). Then again, warning means that we need a suppression mechanism which is painful in scalac.

Alternatively we might return a `LoadableModuleClass` where `loadClass` throws an exception. This would also future proof the API in case we want to add "proper" nested module loading later.

sjrd: I don't think returning something for which `loadModule` throws an exception is more future-proof than returning `None`. The API would have to change anyway, because a nested-module-enabled `loadModule` would have to take the enclosing instance as argument.

I'd vote for returning `None`, and not even warn (I would warn if scalac had a mechanism to suppress warnings). Note that the API has *loadable* in its name (`lookupLoadableModuleClass`). We can argue that a nested module is not *loadable* per se, and therefore it is not found by this method call. This is similar to the fact that abstract classes are not found by `lookupInstantiatableClass`, because they're not instantiatable. We should document the behavior more clearly in the Scaladoc, though, for sure.

gzm0: Hmmm... fair enough. Abstract classes are in exactly the same boat, so let's go with returning `None` for now.

However, I really think we should improve the failure behavior of this API so that we are able to "fail differently" if an item is not found or because we can't return it due to "reasons". Showing these reasons will make it much easier for users to fix the problem at hand.

Think of: "Oh this is failing, it can't find the class... Did I get the name wrong (separators, trailing `$`, etc.)? Do I not inherit from the right thing? Wrong annotation?". All this debugging time can be avoided if we simply say: "A.B.C is abstract, can't instantiate" or "A.B.C is a non-static module, can't be loaded (because: ...)". 

Now, how to expose this in the API such that:
- One can still query if something is loadable without having to catch shady exceptions.
- One can reliably assume that if something gets returned it actually is loadable/instantiatable.

I do not know :)

sjrd:
Superseded by #3239, which implements the alternative semantics discussed here.

PR #3234, opened 2017-12-27
User-space implementation of `js.{Any,ThisFunction}.fromFunctionN`.

sjrd:
Previously, those methods were primitives. They did have a user-space body, but it was not used. This was done because most JS functions were the result of applying those implicit conversions. However, now that we have SAM support in all the versions of Scala for our library, we can get away without them being primitives.

If the user's codebase has SAM support, this commit changes virtually nothing. If it doesn't (it uses Scala 2.11 without `-Xexperimental`), then this commit produces slightly worse code than before (there is a useless double-box of functions), but it is still correct.

If we assume that most end-user codebases should use Scala 2.12 anyway, we are not losing much. We could also improve the optimizer to remove the double-box in the future, if necessary. This commit does not do that, however.


library/src/main/scala/scala/scalajs/js/ThisFunction.scala
```diff
@@ -27,31 +27,28 @@ trait ThisFunction extends js.Function
...
+  implicit def fromFunction22[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R](f: scala.Function22[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R]): js.ThisFunction21[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, R] = (x1: T1, x2: T2, x3: T3, x4: T4, x5: T5, x6: T6, x7: T7, x8: T8, x9: T9, x10: T10, x11: T11, x12: T12, x13: T13, x14: T14, x15: T15, x16: T16, x17: T17, x18: T18, x19: T19, x20: T20, x21: T21, x22: T22) => f(x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14, x15, x16, x17, x18, x19, x20, x21, x22)
```
gzm0: Shouldn't all this be in the next commit?

sjrd: No, it shouldn't. The `identity()` wrappings were there because SAM support was not guaranteed. They had nothing to do with whether or not the methods were primitive or not. As soon as we guarantee SAM support when compiling this file, the `identity()` wrappings become irrelevant.

gzm0: Ok

PR #3237, opened 2017-12-27
Use a "minilib" for the unit tests of the linker.

sjrd:
Instead of using the full `scalajs-library.jar` for the unit tests of the linker, we now use a minimal library. That "minilib" contains things like `Object`, `Integer` and `RuntimeLong`: core classes without which it is not possible to link a hello world.

The minilib does not contain `js.Any` nor any of its subtypes, proving that the linker can live without them. That required to tweak a few core methods not to use niceties from our standard library, such as `JSStringOps` and `JSNumberOps`. Instead, we must be careful to only use `js.Dynamic`, and only inside methods, so that even `js.Dynamic` does not leave any trace in method signatures.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Integer.scala
+++ b/javalanglib/src/main/scala/java/lang/Integer.scala
@@ -136,15 +136,11 @@ object Integer {
     (((t2 + (t2 >> 4)) & 0xF0F0F0F) * 0x1010101) >> 24
   }
 
-  @inline def divideUnsigned(dividend: Int, divisor: Int): Int = {
-    import js.JSNumberOps._
-    asInt(dividend.toUint / divisor.toUint)
-  }
+  @inline def divideUnsigned(dividend: Int, divisor: Int): Int =
+    asInt(asUint(dividend) / asUint(divisor))
 
-  @inline def remainderUnsigned(dividend: Int, divisor: Int): Int = {
-    import js.JSNumberOps._
-    asInt(dividend.toUint % divisor.toUint)
-  }
+  @inline def remainderUnsigned(dividend: Int, divisor: Int): Int =
+    asInt(asUint(dividend) % asUint(divisor))
 
   @inline def highestOneBit(i: Int): Int = {
     /* The natural way of implementing this is:
@@ -234,10 +230,14 @@ object Integer {
   @inline def min(a: Int, b: Int): Int = Math.min(a, b)
 
   @inline private[this] def toStringBase(i: scala.Int, base: scala.Int): String = {
-    import js.JSNumberOps._
-    i.toUint.toString(base)
+    asUint(i).asInstanceOf[js.Dynamic]
+      .applyDynamic("toString")(base.asInstanceOf[js.Dynamic])
+      .asInstanceOf[String]
   }
 
   @inline private def asInt(n: scala.Double): scala.Int =
     (n.asInstanceOf[js.Dynamic] | 0.asInstanceOf[js.Dynamic]).asInstanceOf[Int]
+
+  @inline private def asUint(n: scala.Int): scala.Double =
+    (n.asInstanceOf[js.Dynamic] >>> 0.asInstanceOf[js.Dynamic]).asInstanceOf[scala.Double]
 }
--- a/javalanglib/src/main/scala/java/lang/Long.scala
+++ b/javalanglib/src/main/scala/java/lang/Long.scala
@@ -111,8 +111,7 @@ object Long {
   }
 
   // Intrinsic
-  @inline def toString(i: scala.Long): String =
-    toStringImpl(i, 10)
+  @inline def toString(i: scala.Long): String = "" + i
 
   @inline def toUnsignedString(i: scala.Long): String =
     toUnsignedStringImpl(i, 10)
--- a/javalanglib/src/main/scala/java/lang/_String.scala
+++ b/javalanglib/src/main/scala/java/lang/_String.scala
@@ -33,8 +33,12 @@ final class _String private () // scalastyle:ignore
     this.asInstanceOf[SpecialJSStringOps]
 
   @inline
-  def charAt(index: Int): Char =
-    thisJSString.charCodeAt(index).toChar
+  def charAt(index: Int): Char = {
+    this.asInstanceOf[js.Dynamic]
+      .charCodeAt(index.asInstanceOf[js.Dynamic])
+      .asInstanceOf[Int]
+      .toChar
+  }
 
   def codePointAt(index: Int): Int = {
     val high = charAt(index)
@@ -225,7 +229,7 @@ final class _String private () // scalastyle:ignore
 
   @inline
   def length(): Int =
-    thisJSString.length
+    this.asInstanceOf[js.Dynamic].length.asInstanceOf[Int]
 
   @inline
   def matches(regex: String): scala.Boolean =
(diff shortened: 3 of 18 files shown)


javalanglib/src/main/scala/java/lang/_String.scala
```diff
@@ -332,9 +339,6 @@ object _String { // scalastyle:ignore
    *  and that we need to implement these very Scala methods.
    */
   private trait SpecialJSStringOps extends js.Any {
-    def length: Int
-    def charCodeAt(index: Int): Int
-
     def toLowerCase(): String
     def toUpperCase(): String
     def trim(): String
```
gzm0: I don't understand why the lower three methods (and the trait in general) are OK.

minilib/src/main/scala/java/lang/Class.scala
```diff
@@ -0,0 +1,19 @@
+package java.lang
+
+import scala.scalajs.js
+
+final class Class[A] private (private val data: Object) extends Object {
```
gzm0: I think this class (and all classes in `minilib`) deserve a comment why they are here (and not just pulled from `javalib`).

For this class, I guess it is: "Stripped-down version of jl.Class that only provides what the linker needs itself.".

minilib/src/main/scala/java/lang/System.scala
```diff
@@ -0,0 +1,17 @@
...
+
+object System {
+  // Simpler than the original, technically valid but of lesser quality
+  def identityHashCode(x: Object): scala.Int = {
+    (x: Any) match {
+      case null => 0
+      case _:scala.Boolean | _:scala.Double | _:String | () =>
+        x.hashCode()
+      case _ =>
+        if (x.getClass == null)
```
gzm0: isn't this condition inverted?

minilib/src/main/scala/java/lang/Throwable.scala
```diff
@@ -0,0 +1,19 @@
+package java.lang
+
+class Throwable(s: String, e: Throwable)
```
gzm0: I assume here the reason is because we do not want the stack trace resolution machinery?

javalanglib/src/main/scala/java/lang/Integer.scala
```diff
@@ -234,10 +230,14 @@ object Integer {
...
-    i.toUint.toString(base)
+    toUint(i).asInstanceOf[js.Dynamic]
+      .applyDynamic("toString")(base.asInstanceOf[js.Dynamic])
+      .asInstanceOf[String]
   }
 
   @inline private def asInt(n: scala.Double): scala.Int =
     (n.asInstanceOf[js.Dynamic] | 0.asInstanceOf[js.Dynamic]).asInstanceOf[Int]
+
+  @inline private def toUint(n: scala.Int): scala.Double =
```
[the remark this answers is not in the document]
sjrd: Yes, maybe `asUint` is better here.

javalanglib/src/main/scala/java/lang/Long.scala
```diff
@@ -111,8 +111,7 @@ object Long {
   }
 
   // Intrinsic
-  @inline def toString(i: scala.Long): String =
-    toStringImpl(i, 10)
+  @inline def toString(i: scala.Long): String = "" + i
```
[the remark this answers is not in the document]
sjrd: Because `toStringImpl` (transitively) depends on a bunch of things: `JSNumberOps`, `JSStringOps`, `StringRadixInfo` which is a `js.Array` (and whose rhs depends on `Range` and its `foreach`.

Note that with this change, `Long.toString(i)` has now an implementation similar to `Integer.toString(i)`, which is nice.

javalanglib/src/main/scala/java/lang/_String.scala
sjrd: Because they're not reachable. We only care about reachable stuff, since the Analyzer will eliminate everything else.

minilib/src/main/scala/java/lang/Class.scala
sjrd: Yes, indeed. And in this case, the full one wants a static type for its `data`. If we wanted to keep the full one, we would have needed to drop the static type of `data` in the actual lib, which was a pity.

minilib/src/main/scala/java/lang/System.scala
sjrd: No, it isn't. See https://github.com/sjrd/scala-js/blob/a52281c2acce11e6a40f2052d9d64a76f34bc5df/javalanglib/src/main/scala/java/lang/System.scala#L163

minilib/src/main/scala/java/lang/Throwable.scala
sjrd: Yes, that's right. Especially since it is reachable by the very constructor, because the constructor calls `initStackTrace()`.

javalanglib/src/main/scala/java/lang/_String.scala
sjrd: Now, it's true that, for consistency, we could get rid of `SpecialJSStringOps` entirely. But if we do that, it would also make sense to get rid of all dependencies on `JSStringOps`, which would be quite annoying.

library/src/main/scala/scala/scalajs/runtime/RuntimeLong.scala
```diff
@@ -989,9 +993,17 @@ object RuntimeLong {
     @inline def unsignedSafeDoubleHi(x: Double): Int =
       rawToInt(x / TwoPow32)
 
+    /** Converts an unsigned 32-bit `Int` to a `Double` with the correct value. */
```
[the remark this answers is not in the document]
sjrd: Right, I'll use your description and rename it to `asUint` as well.

sjrd:
Updated.

minilib/src/main/scala/java/lang/FloatingPointBits.scala
```diff
@@ -0,0 +1,19 @@
+package java.lang
+
+import scala.scalajs.js
+
+/** Stripped down version of `java.lang.FloatingPointBits` with the bare
+ *  minimum to support a correct `numberHashCode()`.
+ *
+ *  We cannot the full `java.lang.FloatingPointBits` out of the box, because it
```
gzm0: We cannot *use*

minilib/src/main/scala/java/lang/System.scala
```diff
@@ -0,0 +1,27 @@
+package java.lang
+
+/** Stripped down version of `java.lang.System` with the bare minimum to
+ *  support a correct `numberHashCode()`.
+ *
+ *  We cannot the full `java.lang.System` out of the box, because its
```
gzm0: We cannot *use*

minilib/src/main/scala/java/lang/Throwable.scala
```diff
@@ -0,0 +1,28 @@
+package java.lang
+
+/** Stripped down version of `java.lang.Throwable` with the bare minimum to
+ *  support `toString()` and the subclasses.
+ *
+ *  We cannot the full `java.lang.Throwable` out of the box, because its
```
gzm0: We cannot *use*

project/MiniLib.scala
```diff
@@ -0,0 +1,72 @@
+package build
+
+object MiniLib {
+  val Whitelist = {
+    val inJavaLang = List(
+        "Object",
+        // "Class" is overridden in minilib/
```
gzm0: Add // "System" is overridden in minilib/?

PR #3238, opened 2017-12-29
Remove `abstractMethods` from `LinkedClass`.

sjrd:
That field was "dead code" (it was always an empty list), because the Analyzer never marks as reachable an abstract method. So all code paths that were processing reachable abstract methods were dead code.


sjrd:
This highlights that nothing in our linker currently checks that, when calling `x.meth()`, there is a declared (abstract or not) in the static type of `x` or one of its supertypes. The only thing that we check, obviously, is that all *instantiated* subclasses of `x`'s static type have a concrete `meth`.

That's orthogonal to removing the dead code in this PR, though.

PR #3239, opened 2017-12-29
Fix #3228: Disable reflective instantiation for nested modules.

### LINKED ISSUE #3228: Linking error with an @EnableReflectiveInstantiation object inside a class

sjrd:
We were using `sym.isModuleClass` to identify module classes when generating the reflective instantiation code, but that also matches nested objects. Now, nested objects are not compiled as module classes (only static objects are) because there is one instance per enclosing instance. This caused linking errors as soon as the codebase contained a nested object with `@EnableReflectiveInstantiation`.

Since non-static objects are not really *loadable* as per the API of `Reflect` (we would need a reference to the enclosing instance), we fix this issue by not registering non-static objects at all.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
@@ -55,6 +55,11 @@ class ReflectTest {
   private final val NameTraitDisable =
     Prefix + "TraitDisable"
 
+  private final val NameInnerObject = {
+    Prefix + "ClassWithInnerObjectWithEnableReflectiveInstantiation$" +
+    "InnerObjectWithEnableReflectiveInstantiation"
+  }
+
   @Test def testClassRuntimeClass(): Unit = {
     for {
       name <- Seq(NameClassEnableDirect, NameClassEnableDirectNoZeroArgCtor,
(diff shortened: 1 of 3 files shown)


sjrd:
Supersedes #3229.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala
```diff
@@ -307,4 +317,11 @@ object ReflectTest {
   }
 
   trait TraitDisable extends Accessors
+
+  // Regression cases
+
```
gzm0: Link to issue if you classify this as a regression case?

sjrd: It's in the associated test method name, isn't it?

gzm0: Indeed:-/

Issue #2108, opened 2015-12-11
Defer identifier name mangling until desugaring


sjrd:
Currently, the IR maintains a list of keywords that are reserved in JS, and forbids them in its identifiers. This is too low-level a concern for the IR, and is not robust with future versions of JS adding new reserved keywords. We also forbid leading digits.

We should defer the mangling currently performed by `mangleJSName` in [`JSEncoding`](https://github.com/scala-js/scala-js/blob/master/compiler/src/main/scala/org/scalajs/core/compiler/JSEncoding.scala) to the desugaring.

This a backward binary incompatible change, so it will be for the next major version.

gzm0:
Note: Make sure that things are in the right place (e.g. [`isValidJSIdentifier`, `Definitions` -> `Names`][1]) once this is done.

[1]: https://github.com/scala-js/scala-js/pull/3804#discussion_r333098501

Issue #1961, opened 2015-10-20
Honor call-site @inline and @noinline


sjrd:
The feature was introduced in Scala/JVM in https://github.com/scala/scala/pull/4801

gzm0:
Can we move this post-v1?

sjrd:
We should at least figure out now how we are going to represent that information in the IR. Could they be two additional flags in the `ApplyFlags` of `Apply*` nodes? Currently the `ApplyFlags` are all semantically binding, whereas `@inline` and `@noinline` would be optimization hints.

gzm0:
Ah, OK fair enough.

sjrd:
Thinking more about this, I'm confident we can use the `ApplyFlags` for this purpose, so delaying to post-v1.0.0.

gzm0:
I've started looking into this:

- Changes in the IR/optimizer seem trivial.
- It seems we have forgotten flags on the `New` tree. So we cannot do this for class inlining.

PR #3240, opened 2018-01-02
Fix #3210: First-class private methods in the IR.

### LINKED ISSUE #3210: Model private methods more like static methods in the IR
By definition/construction, private methods can never be overridden. Indeed, their very name encodes the enclosing class/interface. Also in the JVM bytecode definition, private methods in a hierarchy, even when they have the same name and signature, still don't have the same identity, so they do not override each other.

This means that encoding them as member methods in the IR is uselessly inefficient:

* We need to use dynamic lookup to resolve them, whereas static lookup would always be sufficient
* They are encoded as member methods in JS, which is also less efficient than top-level functions with a `$this` parameter
* They require a special `__pXyz` part in their `encodedName`, which is ugly and costly to encode

Modeling private methods more like static methods (namespaced per class, and with static resolution) though with an implicit `this` parameter (like default methods), would solve all these issues.

We could then also encode the actual method implementations of non-native JS classes as private methods rather than static methods, which would make more sense.

sjrd:
Previously, private methods were modeled in the IR as normal methods whose name had a special encoding to ensure that they were non-ambiguous across their entire hierarchy.

In this commit, we mark private methods using an explicit flag, similar to static methods. Those combine as the *namespace* of a member, which can be private or not and static or not. Each of the four namespaces can define methods of the same name (with signature), without them being ambiguous.

Only the public non-static namespace is *inherited* by subclasses. Private and/or static methods can only be accessed using their immediately enclosing class. Therefore, private methods can only be called using `ApplyStatically`, which encodes the enclosing class.

In the emitter, private methods are encoded similarly to default methods: they are top-level functions taking an explicit `this` parameter.

In this commit, constructors must always be public, which corresponds to how things were done previously (private constructors had an encoded name without the `__pXyz` marker). We could possibly revisit this in a future commit, by allowing private constructors in the IR as well.


sjrd:
Ready to be reviewed.

gzm0:
I'm not sure if the `MemberFlags` abstraction is a good one here. My main concerns are:

- High risk of coupling between underlying storage abstraction and interface (how flags are stored should be independent of the interface). In fact, the second commit already shows this: Is a namespace really a flag? I don't think so, it's just *stored* as a flag.
- Unneeded degrees of freedom (one can now construct a mutable `MethodDef`).
- High risk of incompatible changes: The argument that flags can be added easily is IMO an argument *against* the flag construct. Something that interacts with the IR *must* handle all flags. So being able to add a flag without code breaking might be considered a disadvantage.

Also, unrelated higher-level comment: Have you considered splitting namespaces in a static and a private/public flag? It seems simpler to me.

sjrd:
Of course, you're throwing at me my own doubts 😅 

> High risk of coupling between underlying storage abstraction and interface (how flags are stored should be independent of the interface). In fact, the second commit already shows this: Is a namespace really a flag? I don't think so, it's just stored as a flag.

The coupling is an issue, but it also brings some advantages in terms of performance of the code manipulating the IR. Abstracting too much would mean more space on the heap for each of those nodes, which means more GC pressure, etc. As case classes, the IR nodes *are* a use-as-stored thing: they cannot be evolved in source or binary compatible ways anyway.

> Unneeded degrees of freedom (one can now construct a mutable MethodDef).

That is unfortunate, indeed.

> High risk of incompatible changes: The argument that flags can be added easily is IMO an argument against the flag construct. Something that interacts with the IR must handle all flags. So being able to add a flag without code breaking might be considered a disadvantage.

I disagree on this point. Most manipulations of the IR only need to pass-through the information in `flags`, which reduces the need to change unrelated code too often. Of course, if we do add flags, we are defining a new version of the IR and hence IR consumers should determine whether they need to do something about it or not. But most places, after vetting, will not need to be changed.

> Have you considered splitting namespaces in a static and a private/public flag? It seems simpler to me.

Not only have I considered it, that is what I started out with. The result was so ugly that I had to throw it away. There might still be traces of it in the `reflogs` of my computer, but they're not worth recovering ;)

sjrd:
> In fact, the second commit already shows this: Is a namespace really a flag? I don't think so, it's just stored as a flag.

Maybe that problem can be solved just by renaming `MemberFlags` to `MemberModifiers` (and similarly `ApplyFlags` to `ApplyModifiers`).

gzm0:
Note to self: Review done except for the optimizer.

sjrd:
Rebased.

linker/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -846,17 +858,19 @@ private final class IRChecker(unit: LinkingUnit,
           }
         }
 
-      case Apply(receiver, Ident(method, _), args) =>
+      case Apply(flags, receiver, Ident(method, _), args) =>
+        if (flags.isPrivate)
+          reportError(s"Illegal flag for Apply: Private")
```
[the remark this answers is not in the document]
sjrd: Because flags can be generalized, for example for #1961. I should probably have mentioned that upfront, but I thought I'd later reuse those `flags` for the call-site inline/noinline hints.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -259,15 +259,18 @@ object Trees {
...
+  case class Apply(flags: ApplyFlags, receiver: Tree, method: Ident,
+      args: List[Tree])(
       val tpe: Type)(implicit val pos: Position) extends Tree
 
   /** Apply an instance method with static dispatch (e.g., super calls). */
-  case class ApplyStatically(receiver: Tree, cls: ClassType, method: Ident,
-      args: List[Tree])(val tpe: Type)(implicit val pos: Position) extends Tree
+  case class ApplyStatically(flags: ApplyFlags, receiver: Tree, cls: ClassType,
+      method: Ident, args: List[Tree])(
+      val tpe: Type)(implicit val pos: Position) extends Tree
```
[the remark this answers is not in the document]
sjrd: `ApplyStatic` doesn't have a `receiver`, though, unlike `ApplyStatically`.

linker/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -424,19 +429,26 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-      val allInfos = for (methodData <- data.methods)
-        yield (methodData.encodedName, new MethodInfo(this, methodData))
-      val (staticMethodInfos, methodInfos) = allInfos.partition(_._2.isStatic)
-      (mutable.Map(methodInfos: _*), mutable.Map(staticMethodInfos: _*))
+    private lazy val nsMethodInfos = {
+      val nsMethodInfos = Array.fill(MemberNamespace.Count) {
+        mutable.Map.empty[String, MethodInfo]
+      }
+      for (methodData <- data.methods) {
+        // TODO It would be good to report duplicates as errors at this point
```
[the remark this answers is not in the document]
sjrd: That's for later. The current master doesn't have this check, and this PR is exclusively a (giant) refactoring.

linker/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -822,26 +842,28 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+      callMethodStatically(methodName.namespace, methodName.encodedName)
     }
 
-    def callStaticMethod(methodName: String)(implicit from: From): Unit = {
-      lookupStaticMethod(methodName).reachStatic()
+    def callMethodStatically(namespace: MemberNamespace, encodedName: String)(
+        implicit from: From): Unit = {
+      assert(!isReflProxyName(encodedName),
+          s"Trying to call statically refl proxy $this.$encodedName")
+      if (namespace != MemberNamespace.Public || isConstructorName(encodedName))
```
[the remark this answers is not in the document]
sjrd: The thing is that constructors are not inherited, so they are always looked up exclusively in the target class, not its parents. I'm still debating whether constructors should be moved to the *private* non-static namespace for that reason: they obey the same resolution rules. But it's kind of weird too. Perhaps a dedicated constructor namespace as you suggest would be better.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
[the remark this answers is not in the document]
gzm0: Ah, fair enough...

linker/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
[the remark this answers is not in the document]
gzm0: Ok, that's fair. Let's keep it that way then.

gzm0:
I think the constructor namespace is worth investigating (because really, they are their own namespace, they cannot collide with non-constructors).

linker/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -879,7 +901,7 @@ private final class Analyzer(config: CommonPhaseConfig,
     }
 
     def reach(inClass: ClassInfo)(implicit from: From): Unit = {
-      assert(!isStatic,
+      assert(!namespace.isStatic,
```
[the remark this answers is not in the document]
sjrd: Yes, I think you're right.

linker/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Analyzer.scala
```diff
@@ -664,22 +678,26 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-      methodInfos += proxyName -> m
+      publicMethodInfos += proxyName -> m
       m
     }
 
-    def lookupStaticMethod(methodName: String): MethodInfo = {
-      tryLookupStaticMethod(methodName).getOrElse {
-        val syntheticData = createMissingStaticMethodInfo(methodName)
+    def lookupStaticLikeMethod(namespace: MemberNamespace,
+        methodName: String): MethodInfo = {
```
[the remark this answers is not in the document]
sjrd: Static-like versus normal is all about the algorithm used to resolve them. Static-like only looks in the target class, while normal looks in parent classes and interfaces (for default methods). We can remove the distinction, but then the very first thing `lookupMethod` will do is test whether it's `Namespace.Public`, and dispatch to one of the two algorithms based on that. So don't expect any actual simplification, there ;)

sjrd:
> I think the constructor namespace is worth investigating (because really, they are their own namespace, they cannot collide with non-constructors).

I'll have a look at that when I get the chance. I don't have much time to code, these days, as I'm writing my thesis (finally ^^). The deadline is May 7, btw.

linker/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Infos.scala
```diff
@@ -25,6 +25,9 @@ object Infos {
   /** Name used for infos of top-level exports. */
   private val TopLevelExportsName = "__topLevelExports"
 
+  final case class NamespacedEncodedName(
```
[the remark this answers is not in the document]
sjrd: Actually, I tried hard not to use a `MemberID` everywhere, in order to avoid several indirections: the allocations, the dispatch on `hashCode()` and `equals()`, etc. It seemed that exploiting the integer nature of `MemberNamespace` to use an array lookup instead of those additional calls was worth it.

I'll give a try at using a generalized `MemberID` and measure the performance difference.

sjrd:
I have added a commit with a dedicated namespace for constructors (`Constructor`), and also for one for the static initializer (`StaticConstructor`), so that you can see what the changes look like.

The commit will probably not pass CI, because I haven't updated IR printer tests and linker tests, nor checksizes. I'll do that if you give me a 👍 on the approach.

gzm0:
I have given this some more thought, sorry for taking so long.

It seems to me that we already have an implicit other dimension of namespaces in the IR: member types.

Basically they are:
- Fields
- Constructors
- Methods
- JS members (methods, fields and properties)

All these do not collide in names already (by mangling and/or scoping). The exception is `__` being disallowed in JS member names, but we maybe should remove this restriction using a similar mechanism than global ref protection (once #2108 is implemented).

As a consequence, I wonder whether we shouldn't split `MethodDef` into 3 trees:
- `MethodDef` (real IR method def)
- `ConstructorDef` (IR level constructor, will not have a name after #2108).
- `JSMethodDef` (method equivalent of `PropertyDef`, probably should do `s/PropdertyDef/JSPropertyDef`).

The first two would have member flags (we can disallow the private namespaces for constructors), `JSMethodDef` wouldn't.

For non-native JS classes, we might have to introduce a `JSField` tree. This would properly encode private / non-private fields (even if we can't *emit* them properly for now at least in ES5). A non-native JS class may then have the following members:

- `MethodDef` (static only, private only?)
- `ConstructorDef`
- `JSMethodDef`
- `JSPropertyDef`
- `JSFieldDef` (public fields)
- `FieldDef` (private only, private fields)

In summary, we'd have the following members (disregarding export defs):

- `FieldDef`
- `ConstructorDef`
- `MethodDef`
- `JSMethodDef`
- `JSFieldDef`
- `JSPropertyDef`
- `JSFieldDef`

The `JS*` members would not have member flags and share a namespace. The first three trees would have member flags and their individual namespace (plus the 4 given by the member flags themselves). Note that JS constructors would still be present with the special name "constructor" as `JSMethodDef`; given that even in ES6, syntactically constructors are basically just that, I think this is fine.

I'm not sure I have 100% grasped all the implications of this, but I wanted to throw the idea out there anyways.

sjrd:
> sorry for taking so long.

Don't worry. I've been working on my thesis text anyway.

It's funny that you should propose finer decomposition like this, because it's more or less the path I've taken in the formalization I've been writing for my thesis. It seems to work well on paper, at least.

I'll try to send you something that looks decent later today.

sjrd:
@gzm0 I've been looking at this recently, and there's one thing that doesn't quite fit in your latest comment:

> The JS* members would not have member flags and share a namespace.

They still need to be partitioned (and namespaced) by static/non-static. But it's true that they can't be private.

gzm0:
> They still need to be partitioned (and namespaced) by static/non-static. But it's true that they can't be private.

True... So I guess `MemberFlags` is really unavoidable. Fair enough. I still think we should split the trees though.

sjrd:
I've got a wip of the separation in several member types in my branch [separate-member-types](https://github.com/sjrd/scala-js/compare/private-methods...sjrd:separate-member-types). It does not yet compile (there are a few missing things in the emitter), but TBH it already feels like a regression all over.

* The separation between `MethodDef` and `ConstructorDef` just means a *lot* of duplicated work. Everything that must be done for one also needs to be done for the other. It gets insane at the optimizer level, so much that I haven't even tried: in the branch, constructors are not optimized.
* The separation between `FieldDef` and `JSFieldDef` also causes mostly duplicated work everywhere. And it also forces reordering of fields between non-exposed ones and exposed ones (because you need all of one kind then all of the other kind), which is not good.
* The separation between `MethodDef` and `JSMethodDef` is OK, as in it does not result in significantly duplicated work, but it also does not seem to be an improvement anywhere.

From what I saw so far, I believe we're much better off with the status quo.

gzm0:
> From what I saw so far, I believe we're much better off with the status quo.

OK, seems fair. The branch indeed is mostly duplication. Thank you for trying.

What needs to be done next here in this case? Is it ready for final review or do things need squashing / deciding? (There seems to be a commit with a constructor namespace).

sjrd:
I think the next step is a final review with the commits as they are, and decide what to squash. I would be for squashing the last 3 commits.

ir/src/main/scala/org/scalajs/ir/Trees.scala
```diff
@@ -1059,29 +1062,131 @@ object Trees {
...
 
     def withMutable(value: Boolean): MemberFlags =
       if (value) new MemberFlags(bits | MutableBit)
       else new MemberFlags(bits & ~MutableBit)
   }
 
   object MemberFlags {
-    private final val StaticShift = 0
-    private final val StaticBit = 1 << StaticShift
+    private final val NamespaceMask = 7
```
gzm0: Express this with shift as well?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
```diff
@@ -75,8 +76,8 @@ object Analysis {
     def isDataAccessed: Boolean
     def linkedFrom: Seq[From]
     def instantiatedFrom: Seq[From]
-    def methodInfos: scala.collection.Map[String, MethodInfo]
-    def staticMethodInfos: scala.collection.Map[String, MethodInfo]
+    def methodInfos(
+        namespace: MemberNamespace): scala.collection.Map[String, MethodInfo]
```
gzm0: Is the choice of exposing this as a method (not a nested `Map`) deliberate? It feels a bit odd. (I completely see that it is easier to _implement_ it this way, but the mere interface feels odd :S). (this is a bump of an old comment).

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -985,31 +995,30 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-    makeSyntheticMethodInfo(encodedName = encodedName)
+    makeSyntheticMethodInfo(encodedName)
   }
 
-  private def createMissingStaticMethodInfo(
+  private def createMissingStaticLikeMethodInfo(namespace: MemberNamespace,
       encodedName: String): Infos.MethodInfo = {
-    makeSyntheticMethodInfo(encodedName = encodedName, isStatic = true)
+    makeSyntheticMethodInfo(encodedName, namespace)
   }
```
gzm0: I feel like these two helpers are not warranted. In total, they are used 3 times and can be replaced by `makeSyntheticMethodInfo`.

Note that one call to `createMissingMethodInfo` seems wrong, as it refers to a constructor.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -427,19 +432,26 @@ private final class Analyzer(config: CommonPhaseConfig,
     var instantiatedSubclasses: List[ClassInfo] = Nil
     var methodsCalledLog: List[(String, From)] = Nil
 
-    lazy val (methodInfos, staticMethodInfos) = {
-      val allInfos = for (methodData <- data.methods)
-        yield (methodData.encodedName, new MethodInfo(this, methodData))
-      val (staticMethodInfos, methodInfos) = allInfos.partition(_._2.isStatic)
-      (mutable.Map(methodInfos: _*), mutable.Map(staticMethodInfos: _*))
+    private lazy val nsMethodInfos = {
```
gzm0: Seems kind of pointless to make this a lazy val if you evaluate it anyways for `publicMethodInfos` below. (this is a bump).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -166,18 +166,39 @@ final class Emitter private (config: CommonPhaseConfig,
...
+            System.err.println(m.value)
+            m.value.flags.namespace == MemberNamespace.Constructor &&
+            m.value.encodedName == ctorName
+          })
+        if (!ctorIsDefined) {
+          implicit val pos = NoPosition
+          val field = jsGen.envField("ct", className + "__" + ctorName).ident
+          builder.addJSTree(js.VarDef(field, None))
+        }
+      }
```
gzm0: Hum, not sure I understand what is going on here... Why do we need this now?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -151,16 +154,16 @@ private final class IRChecker(unit: LinkingUnit,
...
       val methodDef = member.value
       implicit val ctx = ErrorContext(methodDef)
 
-      assert(methodDef.flags.isStatic, "Found non-static member in static defs")
-
-      if (!methodDef.name.isInstanceOf[Ident])
-        reportError(s"Static method ${methodDef.name} cannot be exported")
-      else
-        checkMethodDef(methodDef, classDef)
+      if (methodDef.flags.namespace.isStatic) {
```
gzm0: Consider putting this as a guard of the for loop.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -259,23 +262,26 @@ private final class IRChecker(unit: LinkingUnit,
...
-    for (method <- classDef.memberMethods) {
+    for (method <- classDef.methods) {
       val tree = method.value
       implicit val ctx = ErrorContext(tree)
 
-      assert(!tree.flags.isStatic,
-          "Member or abstract method may not be static")
-      assert(tree.name.isInstanceOf[Ident],
-          "Normal method must have Ident as name")
+      if (!tree.flags.namespace.isStatic) {
```
gzm0: Consider putting this as a guard on the for-loop.

ir/src/main/scala/org/scalajs/ir/Trees.scala
sjrd: No, for this one it is important that it stays at no-shift. I have added a comment.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
sjrd: It is necessary now because constructors became top-level functions, so they are statically referenced, and therefore can cause GCC compile errors if they are called but not defined. Previously, constructors were instance methods, so they were not statically referenced and would not cause GCC errors.

sjrd:
Rebased.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -115,17 +102,22 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
-          container.fold[List[Versioned[MethodDef]]](Nil)(_.optimizedDefs)
-
         val encodedName = linkedClass.encodedName
-        val memberNamespace =
-          if (linkedClass.kind == ClassKind.Interface) getDefaults(encodedName)
-          else getClass(encodedName)
+        val staticLikeContainers = staticLikes(encodedName)
+
+        val publicContainer = classes.get(encodedName).getOrElse {
+          staticLikeContainers(MemberNamespace.Public.ordinal)
```
gzm0: I think this warrants a comment: IIUC this happens for default methods?

Maybe even better than a comment would be an `assert(linkedClass.kind == Interface)`?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -115,17 +102,22 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
+        }
+
+        val newMethods = for (m <- linkedClass.methods) yield {
+          val namespace = m.value.flags.namespace
+          val container =
+            if (namespace == MemberNamespace.Public) publicContainer
+            else staticLikeContainers(namespace.ordinal)
+          container.methods(m.value.encodedName).optimizedMethodDef
+        }
```
gzm0: I'm not sure I understand why this is now necessary. Couldn't we just take all methods in all containers?

Something like:

`classes.getOrElse(encodedName, empty) ++ statikLikeContainers.flatMap(_.defs)`

?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -136,83 +128,61 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
+    val neededStaticLikes = CollOps.emptyParMap[String, LinkedClass]
     val neededClasses = CollOps.emptyParMap[String, LinkedClass]
-    val neededStatics = CollOps.emptyParMap[String, LinkedClass]
-    val neededDefaults = CollOps.emptyParMap[String, LinkedClass]
     for (linkedClass <- linkedClasses) {
       // Update the list of ancestors for all linked classes
       getInterface(linkedClass.encodedName).ancestors = linkedClass.ancestors
 
+      CollOps.put(neededStaticLikes, linkedClass.encodedName, linkedClass)
+
```
gzm0: Just to test my understanding: This is a simplification that we say every class needs it's own static namespace unconditionally. Since changing the static namespace tags the class for update anyways.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -66,8 +66,9 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
 
   private var objectClass: Class = _
   private val classes = CollOps.emptyMap[String, Class]
-  private val statics = CollOps.emptyParMap[String, StaticsNamespace]
-  private val defaults = CollOps.emptyParMap[String, Defaults]
+
+  private val staticLikes =
+    CollOps.emptyParMap[String, Array[StaticLikeNamespace]]
```
gzm0: Would it be possible to type this as `Array[MethodContainer]` instead and get rid of the strange empty `StaticLikeNamespaces` for classes (and put the `Class` instead)?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -309,9 +282,21 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
+            if namespace == MemberNamespace.Public &&
+                (linkedClass.kind.isClass || linkedClass.kind == ClassKind.HijackedClass) =>
+          /* The public non-static namespace for a class is always empty,
+           * because its would-be content must be handled by the `Class`
+           * instead.
+           */
+          -1 // different from all `ordinal` values of legit namespaces
+        case _ =>
+          namespace.ordinal
+      }
```
gzm0: This is not nice :( See comment on top how we can maybe improve this.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -621,31 +613,32 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
             }
           }
 
-        // Super class constructor.
-        case ApplyStatically(This(), ClassRef(cls), methodName, args) =>
-          Definitions.isConstructorName(methodName.name) &&
+        // Delegation to another constructor (super or in the same class)
+        case ApplyStatically(flags, This(), ClassRef(cls), methodName, args)
+            if flags.isConstructor =>
+          val namespace = MemberNamespace.forNonStaticCall(flags).ordinal
```
gzm0: Isn't this always the non static constructor namespace anyways?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -956,25 +923,19 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
-          methodName: String): Option[MethodID] = {
-        val staticsNS = statics(className)
-        registerCallStatic(staticsNS.myInterface, methodName)
-        staticsNS.methods.get(methodName)
+        def inStaticsLike =
+          staticLikes(className)(namespace.ordinal)
+
+        val container =
+          if (namespace != MemberNamespace.Public) inStaticsLike
+          else classes.get(className).getOrElse(inStaticsLike)
```
gzm0: This would also benefit from typing `staticLikes` with `MethodContainer` only (and of course renaming it).

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1962,8 +1961,9 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
     } (cont) (scope.withEnv(OptEnv.Empty))
   }
 
-  private def callIntrinsic(code: Int, optTReceiver: Option[PreTransform],
-      targs: List[PreTransform], isStat: Boolean, usePreTransform: Boolean)(
+  private def callIntrinsic(code: Int, flags: ApplyFlags,
+      optTReceiver: Option[PreTransform], targs: List[PreTransform],
+      isStat: Boolean, usePreTransform: Boolean)(
```
gzm0: It seems like the call-sites of this should guard against bad apply flags, but currently don't.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2375,7 +2380,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
             Assign(lhs, If(cond, th, value)(lhs.tpe)(stat.pos))(ass.pos) :: rest,
             cancelFun)(buildInner)(cont)
 
-      case ApplyStatically(ths: This, superClass, superCtor, args) :: rest
+      case ApplyStatically(flags, ths: This, superClass, superCtor, args) :: rest
```
gzm0: Assert on the flags?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
sjrd: It's necessary to preserve the order of methods as found in the .sjsir, which itself is necessary to preserve stability of the output.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
sjrd: I added a comment, and an assert. The assert is actually a bit more complex than what you suggest, as this code is called for all non-Scala-classes. It's just that the containers must be empty for the ones that are not interfaces.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
sjrd: Unfortunately, if I do that, it becomes very difficult to reason about what happens when a given `encodedName` transitions from being a class to being something else in an incremental run, or vice versa. The timing of when to delete/change/add things becomes very complicated.

sjrd:
For the squashing strategy, is it fine if I squash everything but the first commit?

gzm0:
Yes, please

sjrd:
Done. You might want to have a last look at the commit message for the squashed commit.

gzm0:
LGTM

PR #3242, opened 2018-01-08
Tweak the mechanism to translate Scala varargs into JS varargs.

sjrd:
Previously, we used an overly general method `genTraversableOnce2jsArray` to convert the `Seq` of Scala varargs into a JS array suitable for the `JSSpread` node. We now specialize that method to only accept `Seq`s and rename it to `toJSVarArgs`.

Only handling `Seq`s means we can drop one case of the match, which was indeed not taken in practice for varargs.

We also remove the `@inline` annotation from the method. It is only useful that it be inlined when the argument is a known `js.WrappedArray` (to be expanded by the optimizer), and in that case the heuristics of the optimizer will force the inlining anyway. This change reduces the cases of spurious inlining of that method.


sjrd:

Issue #3033, opened 2017-06-27
Rethink JSEnv implementation APIs


gzm0:
Notably `AbstractNodeJSEnv` and `ExternalJSEnv` can probably be improved. Notably use less inheritance.

sjrd:
Other things that can be improved, IMO:
* Get rid of the synchronous APIs (`jsRunner`). All our JS envs support asynchronous runs. In general, it's always possible to implement an asynchronous API on top of a synchronous API anyway.
* The arbitrary distinction between arguments given to `{js,async,com}Runner` and those given to their `run`/`start` methods. IMO all the arguments should go to the method in `JSEnv`, and that method should start immediately the runner. A `JSRunner` would always be running or stopped (but not yet-to-be-started).

gzm0:
>  IMO all the arguments should go to the method in JSEnv, and that method should start immediately the runner.

Not 100% convinced about that one. Can be hard if you need two runners that communicate with each other. Mainly because you can't get references to both without starting them (and messages already piling up).

An API like this is OK for apparently immutable things like `Future.apply`. But for things that have state (as in `ComJSEnv.send`) this can lead to very difficult setup sequences. Maybe we can add some final helper methods that do this?

sjrd:
How does the current scheme fare better for that use case? AFAICT, if you call `send()` before `start()`, you just block. You might just as well buffer your messages yourself if you have a use case of communicating between two runners.

gzm0:
Stuff to also add:

- Expected thread safety guarantees.
- Exact exception semantics (what get's thrown when, send after close, receive after close, send while close, receive while close).

gzm0:
I have given this quite some thought and think something like this would be sufficient:

```scala
trait JSEnv {
  val name: String // should really be stable -> val.

  def start(files: List[VirtualJSFile], console: JSConsole, logger: Logger): JSRun
}

trait JSRun {
  // Future completes if run completes.
  def future: Future[Unit]

  // Stops run without interrupting an operation. I.e. it inserts a `stop` operation on the event-loop.
  // Idempotent, async.
  def stop(): Unit

  // Forcefully stops the run. If the run is not stopped, will make run fail with error.
  // Idempotent, async.
  def kill(): Unit

  // Send a message to the JS end, async.
  // "Eventually throws" if the run is completed.
  def send(msg: String): Unit

  // Blocking receive. Throws if the run is completed but buffers incoming messages since completion.
  def receive(): String
[...]

PR #3244, opened 2018-01-09
Separate `tools` into `io`, `logging` and `linker`.

sjrd:
They define the packages `tools.io`, `tools.logging` and `tools.linker`, respectively.

This allows `js-envs` and related artifacts, as well as `test-adapter`, not to depend on the `linker` nor `ir` anymore. Therefore, those artifacts can be used without bringing the linker and IR into the classpath, which can be very important for build tools that want to load the linkers of various Scala.js versions in separate class loaders.

Doing the separation required a few prior refactorings, which come as separate commits.

I also intend to change the *packages* of all those things, to remove the `.tools.` part, and probably even the `.core.` part. This would result in a more significant diff, and considerations about source compatibility of the tools across 0.6.x and 1.x, which deserve a separate PR.


tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -122,19 +122,36 @@ trait WritableVirtualJSFile extends WritableVirtualTextFile with VirtualJSFile {
...
- *
- *  This can be a [[VirtualScalaJSIRFile]] (with [[RelativeVirtualFile]]) or a
- *  [[VirtualFileContainer]].
  *
  *  The main difference compared to using individual files
  *  (that are extracted beforehand) is that the fileset can be versioned at a
  *  higher level: the container needs to change its version when any of the
  *  files change. Therefore, the entire extraction process can be cached.
  */
 trait ScalaJSIRContainer extends VirtualFile {
```
gzm0: Why does this not extend `VirtualFileContainer` and directly define `sjsirFiles`?

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -189,9 +206,7 @@ trait VirtualSerializedScalaJSIRFile
  *  This is a generic virtual container for embedded virtual files, especially
  *  one found on a classpath such as a jar, and containing `.sjsir` files.
```
gzm0: Comment should probably drop references to IR.

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -204,43 +219,6 @@ trait VirtualFileContainer extends ScalaJSIRContainer {
...
-  }
-
-  def jsFiles: List[VirtualJSFile with RelativeVirtualFile] = {
-    listEntries(_.endsWith(".js")) { (relPath, stream) =>
-      val file = new EntryJSFile(path, relPath)
-      file.content = IO.readInputStreamToString(stream)
-      file.version = version
-      file
-    }
-  }
```
gzm0: Just to double check: This was dead-code?

js-envs/src/main/scala/org/scalajs/jsenv/Utils.scala
```diff
@@ -27,12 +27,91 @@ private[jsenv] object Utils {
...
         case _                       => null
       })
     }
   }
+
+  // !!! BEGIN COPY-PASTED CODE with ir/.../Utils.scala
+
+  private final val EscapeJSChars = "\\b\\t\\n\\v\\f\\r\\\"\\\\"
+
+  def escapeJS(str: String): String = {
```
gzm0: This shouldn't be in the public interface of `jsEnvs` :-/ But I don't have a better solution for this ATM. Note that `OptDeadline` will vanish completely with #3033 (for which I hope to send out a PR today).

gzm0: Looking at the next commit, maybe we should just move this to `io` as well? Neither `URIUtils` nor this are `io`, but essentially "everything" needs it.

build.sbt
```diff
@@ -4,8 +4,12 @@ val scalajs = Build.root
 val ir = Build.irProject
 val irJS = Build.irProjectJS
 val compiler = Build.compiler
-val tools = Build.tools
-val toolsJS = Build.toolsJS
+val io = Build.io
+val ioJS = Build.ioJS
+val logging = Build.logging
+val loggingJS = Build.loggingJS
```
gzm0: Maybe we should just have a `base` artifact with packages io, logging and utils (where `URIUtils` and `escapeJS` go)?

js-envs/src/main/scala/org/scalajs/jsenv/Utils.scala
sjrd: We cannot move it to `io` because it is used by `ir.Printers`.

build.sbt
sjrd: In that case, `ir` needs to depend on `base` (for `escapeJS`), and that means it depends on all of `io` and `logging`, but it really shouldn't.

(also `compiler` needs to source in `ir/src/` and all its dependencies, so that would mean sourcing in all of `io` and `logging` for no reason)

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
sjrd: Because `VirtualScalaJSIRFile extends ScalaJSIRContainer`. So if `ScalaJSIRContainer extends VirtualFile`, then `VirtualScalaJSIRFile` transitively implements `VirtualFileContainer`, and everything goes berserk.

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
sjrd: Yes, it was dead code.

js-envs/src/main/scala/org/scalajs/jsenv/Utils.scala
gzm0: I'm OK with the duplication in IR. I'm mostly concerned about it being part of the interface of `jsEnvs`.

sjrd: Ah.

OK maybe we can put it in `io.JSUtils` then, and make the other one `private[ir]`.

tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
gzm0: I see... LOL. Implementation inheritance really sucks :P

js-envs/src/main/scala/org/scalajs/jsenv/Utils.scala
gzm0: SG.

sjrd:
Rebased and updated.

test-adapter/src/main/scala/org/scalajs/testadapter/TestAdapter.scala
```diff
@@ -176,18 +170,24 @@ final class TestAdapter(jsEnv: ComJSEnv, jsFiles: Seq[VirtualJSFile],
 }
 
 object TestAdapter {
+  sealed abstract class ModuleIdentifier
```
gzm0: Maybe add a comment that `sealed` is not part of the public API?

sjrd: Done. I also used your trick for the other PR with a non-sealed class whose constructor is `private`, so that is actually not `sealed`. We only have to match on it within our own code, and I would hope that we will always have a least one test for each module kind ;)

PR #3246, opened 2018-01-10
Introduce a true for-in loop in the IR.

sjrd:
This commit adds first-class support for the general `for-in` loop. It can be produced with `js.special.forin`, which is now used to implement `js.Object.properties`.

This is important because `js.Object.properties` cannot encode all uses of `for-in` loops. While it can encode most of the semantics applying to ordinary JS objects, it is not enough for exotic objects, notably proxies. Only exposing `js.Object.properties` was therefore a hole in our interoperability features.

In addition, if another language wants to compile to the IR, they should have a way to encode an efficient `for-in` loop, which was previously not possible, as the only efficient `for-in` loop was based on intrinsifying a method from the Scala.js standard library.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
@@ -199,6 +199,11 @@ object Trees {
     val tpe = NoType // cannot be in expression position
   }
 
+  case class ForIn(obj: Tree, keyVar: Ident, body: Tree)(
+      implicit val pos: Position) extends Tree {
+    val tpe = NoType
+  }
+
   case class TryCatch(block: Tree, errVar: Ident, handler: Tree)(
       val tpe: Type)(implicit val pos: Position) extends Tree
 
--- a/library/src/main/scala/scala/scalajs/js/Any.scala
+++ b/library/src/main/scala/scala/scalajs/js/Any.scala
@@ -194,15 +194,34 @@ object Any extends js.LowPrioAnyImplicits {
      *  object `o`, including properties in its prototype chain.
      *
      *  This method returns the same set of names that would be enumerated by
-     *  a for-in loop in JavaScript, but not necessarily in the same order.
+     *  a for-in loop in JavaScript, in the same order.
      *
-     *  If the underlying implementation guarantees an order for for-in loops,
-     *  then this is guaranteed to be consistent with [[js.Object.keys]], in
-     *  the sense that the list returned by [[js.Object.keys]] is a sublist of
-     *  the list returned by this method (not just a subset).
+     *  This method assumes that all keys enumerated by a for-in loop are
+     *  strings. If this is not the case, calling this method is an undefined
+     *  behavior of kind `ClassCastException`. Note that for all *ordinary*
+     *  objects, the ECMAScript 2015 guarantees that this is the case. It might
+     *  be false if `o` is a proxy object or another exotic object.
+     *
+     *  For ordinary objects, if the underlying implementation guarantees an
+     *  order for for-in loops, then this is guaranteed to be consistent with
+     *  [[js.Object.keys]], in the sense that the list returned by
+     *  [[js.Object.keys]] is a sublist of the list returned by this method
+     *  (not just a subset).
      */
-    def properties(o: js.Any): js.Array[String] =
-      scala.scalajs.runtime.propertiesOf(o)
+    @noinline
+    def properties(o: js.Any): js.Array[String] = {
+      /* DO NOT touch this code without double-checking the optimized code.
+       *
+       * This implementation is carefully crafted so that the optimizer turns
+       * the code into a pattern known not to fall off the performance cliffs.
+       */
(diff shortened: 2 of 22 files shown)


ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -193,6 +193,11 @@ object Trees {
     val tpe = NoType // cannot be in expression position
   }
 
+  case class JSForIn(obj: Tree, keyVar: Ident, body: Tree)(
```
sjrd: Given that even `Debugger` is called that way, and not `JSDebugger`, this node should be called `ForIn`.

library/src/main/scala/scala/scalajs/js/Any.scala
```diff
@@ -194,15 +194,34 @@ object Any extends js.LowPrioAnyImplicits {
...
      */
-    def properties(o: js.Any): js.Array[String] =
-      scala.scalajs.runtime.propertiesOf(o)
+    @noinline
+    def properties(o: js.Any): js.Array[String] = {
+      /* DO NOT touch this code without double-checking the optimized code.
+       *
+       * This implementation is carefully crafted so that the optimizer turns
+       * the code into a pattern known not to fall off the performance cliffs.
+       */
```
gzm0: Hum... Can we make an optimizer test for this? The problem with this comment is that it implies that one also always needs to check the generated code if one changes the optimizer. That can easily be forgotten I think.

library/src/main/scala/scala/scalajs/js/special/package.scala
```diff
@@ -92,6 +92,29 @@ package object special {
...
+   *  JavaScript. A call of the form
+   *  {{{
+   *  forin(obj)(f)
+   *  }}}
+   *  corresponds to the JavaScript statement
+   *  {{{
+   *  for (const x in obj) {
+   *    f(x)
+   *  }
+   *  }}}
```
gzm0: This is not true if the expression `f` is side-effecting (see corresponding comment in compiler). I'm not sure how strict we are with this in comments in general (given that usual Scala semantics apply), but maybe it might be worth dropping a comment.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -625,6 +625,17 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
             }, newLabel)
           }
 
+        case ForIn(obj, keyVar, body) =>
+          unnest(obj) { (newObj, env0) =>
+            implicit val env = env0
+
+            val lhs = genEmptyImmutableLet(transformLocalVarIdent(keyVar))
+            js.ForIn(lhs, transformExprNoChar(newObj), {
```
gzm0: Having an immutable let as lhs here is IMO a bit of an abuse (just like we used to have for LHSes in the desugarer), but I can't come up with anything better.

library/src/main/scala/scala/scalajs/js/special/package.scala
```diff
@@ -92,6 +92,29 @@ package object special {
...
+   *
+   *  `for..in` loops exhibit performance cliffs in most JavaScript engines,
+   *  which means their performance is very dependent on their surroundings.
+   *  Therefore, it is recommended to avoid this method and use
+   *  [[js.Any.ObjectCompanionOps.properties js.Object.properties]] instead, if
```
gzm0: So just to double check: `js.Object.properties` calls this method internally, but just makes sure it does so properly, right?

library/src/main/scala/scala/scalajs/js/Any.scala
sjrd: If we had some infrastructure for optimizer tests, I would definitely do it. 😞

library/src/main/scala/scala/scalajs/js/special/package.scala
sjrd: Ah, right! It's maybe pedantic, but yes, I'll show the temporaries in the code examples to make it crystal clear.

library/src/main/scala/scala/scalajs/js/special/package.scala
sjrd: Yes, that's right.

library/src/main/scala/scala/scalajs/js/Any.scala
gzm0: Ok, fair enough. The optimizer is definitely your call anyways :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Same here. 😞

sjrd:
Updated.

PR #3247, opened 2018-01-10
Make `js.FunctionN`s map to arrow functions in ECMAScript 2015.

sjrd:
Arrow functions have some distinct semantics besides not capturing `this`: they cannot be *constructed* (i.e., called with `new`) and they do not have a `prototype` property. Therefore, they also cannot be used as "parent classes".

As such, we need an interop feature to create true arrow functions when targeting ES 2015+. The existing `js.FunctionN`s are actually a perfect fit for this, since they already do not have access to their `this` parameter.

This is, in theory, a breaking change, as someone could have relied on calling a `js.FunctionN` with `new` (which succeeded before, and now throws). However, without access to the `this` value, there is no non-contrived use case for that, so it is unlikely to be noticeable.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -704,9 +704,10 @@ abstract class GenJSCode extends plugins.PluginComponent
       def selfRef(implicit pos: ir.Position) =
         js.VarRef(selfName)(jstpe.AnyType)
 
-      def lambda(params: List[js.ParamDef], body: js.Tree)(
+      def memberLambda(params: List[js.ParamDef], body: js.Tree)(
           implicit pos: ir.Position) = {
-        js.Closure(captureParams = Nil, params, body, captureValues = Nil)
+        js.Closure(arrow = false, captureParams = Nil, params, body,
+            captureValues = Nil)
       }
 
       val memberDefinitions = classMembers.toList.map {
@@ -722,7 +723,7 @@ abstract class GenJSCode extends plugins.PluginComponent
         case mdef: js.MethodDef =>
           implicit val pos = mdef.pos
           val name = mdef.name.asInstanceOf[js.StringLiteral]
-          val impl = lambda(mdef.args, mdef.body.getOrElse(
+          val impl = memberLambda(mdef.args, mdef.body.getOrElse(
               throw new AssertionError("Got anon SJS class with abstract method")))
           js.Assign(js.JSBracketSelect(selfRef, name), impl)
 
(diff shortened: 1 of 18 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -706,7 +706,8 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       def lambda(params: List[js.ParamDef], body: js.Tree)(
```
gzm0: I think this name is worth changing. It's not a simple `lambda` anymore. It's actually a `thisFunctionLambda` something. Maybe `memberLambda`?

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -848,10 +848,14 @@ object Trees {
   case class This()(val tpe: Type)(implicit val pos: Position) extends Tree
 
   /** Closure with explicit captures.
-   *  The n captures map to the n first formal arguments.
+   *
+   *  @param arrow
+   *    If `true`, the closure is an Array Function (`=>`), which does not have
```
gzm0: s/Array/Arrow/

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -889,20 +891,23 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
                 newClosure,
                 RefinedType(AnyType, isExact = false, isNullable = false)))
           }
 
-          if (params.exists(_.rest)) {
+          if (!arrow && params.exists(_.rest)) {
             /* TentativeClosureReplacement assumes there are no rest
              * parameters, because that would not be inlineable anyway.
-             * So we never try to inline a Closure with a rest parameter.
+             * Likewise, it assumes that there is not binding for `this`, which
```
gzm0: there is *no* binding for `this`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -889,20 +891,23 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
+             * is only true for arrow functions.
+             * So we never try to inline arrow Closures, nor Closures with a
```
gzm0: So we never try to inline *non-arrow* Closures.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
sjrd: `memberLambda` sounds fine :)

sjrd:
Updated.

PR #3248, opened 2018-01-10
Consolidate the notion of transient node in the IR.

sjrd:
Previously, we had two kinds of nodes in the IR used for internal purposes: `UndefinedParam` and `CallHelper`. Those nodes should not be part of the public IR definition.

We get rid of that public display by introducing a unique, generic `Transient` node, which can be used by internal phases for custom purposes. The scalac plugin uses it for `UndefinedParam`, and the optimizer+emitter use it for `CallHelper`. Transient nodes could be used by any other phase, including user-defined tools manipulating IR trees.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -331,7 +331,7 @@ abstract class GenJSCode extends plugins.PluginComponent
               } catch {
                 case e: ir.InvalidIRException =>
                   e.tree match {
-                    case ir.Trees.UndefinedParam() =>
+                    case ir.Trees.Transient(UndefinedParam) =>
                       reporter.error(sym.pos,
                           "Found a dangling UndefinedParam at " +
                           s"${e.tree.pos}. This is likely due to a bad " +
@@ -1879,7 +1879,7 @@ abstract class GenJSCode extends plugins.PluginComponent
             else genExpr(rhs)
 
           rhsTree match {
-            case js.UndefinedParam() =>
+            case js.Transient(UndefinedParam) =>
               // This is an intermediate assignment for default params on a
               // js.Any. Add the symbol to the corresponding set to inform
               // the Ident resolver how to replace it and don't emit the symbol
@@ -1974,7 +1974,7 @@ abstract class GenJSCode extends plugins.PluginComponent
             } else if (undefinedDefaultParams contains sym) {
               // This is a default parameter whose assignment was moved to
               // a local variable. Put a literal undefined param again
-              js.UndefinedParam()(toIRType(sym.tpe))
+              js.Transient(UndefinedParam)(toIRType(sym.tpe))
             } else {
               js.VarRef(encodeLocalSym(sym))(toIRType(sym.tpe))
             }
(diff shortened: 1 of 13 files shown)


sjrd:
This refactoring highlights that `CallHelper`s are a secret common knowledge of the optimizer and the `Emitter`, which is not very nice. At the same time, they are only used for intrinsics. Maybe we should handle those intrinsics directly within the Emitter, instead of having the `OptimizerCore` rewrite them early into `CallHelper`s. But that is very low priority, as it can be done in point releases.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5573,4 +5576,6 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     case class GlobalScope(pos: Position) extends MaybeGlobalScope
   }
+
+  object UndefinedParam
```
gzm0: I think this deserves a comment :) Something like: "Marker object for undefined parameters in JavaScript semantic calls. To be used inside a js.Transient node".

Also make this a case object?

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
```diff
@@ -817,6 +808,13 @@ object Printers {
           print("createjsclass[")
           print(cls)
           printRow(captureValues, "](", ", ", ")")
+
+        // Transient
+
+        case Transient(value) =>
+          print("<transient ")
+          print(value.getClass.getSimpleName)
```
gzm0: I think simply `toString` is more appropriate here. If somebody uses a general purpose class inside a `Transient` node, they are doing it wrong anyways (meaning they can give it a reasonable `toString`).

We might even define an interface (or even abstract class) that needs to be implemented specifically to avoid this pitfall.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/analyzer/Infos.scala
```diff
@@ -487,6 +488,9 @@ object Infos {
             case CreateJSClass(cls, _) =>
               builder.addInstantiatedClass(cls.className)
 
+            case Transient(CallHelper(_, args)) =>
+              args.foreach(traverse)
+
```
gzm0: Waaa... This is indeed awful :P Consider adding a comment that this is should only happen in the refinement phase.

ir/src/main/scala/org/scalajs/core/ir/Printers.scala
sjrd: I like the idea of an interface. (an abstract class is a bit too constraining for something custom by design)

sjrd:
Updated

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Transients.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  final case class CallHelper(helper: String, args: List[Tree])
+      extends Transient.Value {
+
+    def printIR(out: java.io.Writer, irPrinter: IRTreePrinter): Unit = {
+      out.write("$callHelper(")
+      out.write(helper)
+      for (arg <- args) {
+        out.write(", ")
+        irPrinter.print(arg)
```
gzm0: Hmmm... These two lines make the relationship between `irPrinter` and `out` appear magic :-/ Alternative would be to make `IRTreePrinter.print(_: String)` public, but that's also not nice...

Have you considered creating a dedicated type for this? (that has `print(_: String)` and `print(_: Tree)`?)

sjrd: `IRTreePrinter` could implement all the methods of `Writer` by forwarding them to `out`. That would make the relationship clearer.

Just `String` and `Tree` wouldn't be enough. There's also `Type` and several others.

sjrd:
Updated.

gzm0:
Good to merge without review after conflict resolution.

PR #3249, opened 2018-01-10
Stricter typing for `JSSpread`s in the IR.

sjrd:
Previously, `JSSpread` extended `Tree` That was nonsense because `Tree`s are supposed to be valid anywhere in statement/expression position (provided they typecheck), although `JSSpread`s are only valid in very specific positions, structurally speaking.

We reify those restrictions in the types of the IR nodes, by introducing `TreeOrJSSpread` as a common super type for `Tree` and `JSSpread`.

### DIFF
--- a/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
+++ b/ir/src/main/scala/org/scalajs/core/ir/Trees.scala
@@ -30,8 +30,14 @@ object Trees {
     }
   }
 
+  /** Either a `Tree` or a `JSSpread`.
+   *
+   *  This is the type of actual arguments to JS applications.
+   */
+  sealed trait TreeOrJSSpread extends IRNode
+
   /** Node for a statement or expression in the IR. */
-  abstract sealed class Tree extends IRNode {
+  abstract sealed class Tree extends IRNode with TreeOrJSSpread {
     val tpe: Type
   }
 
@@ -475,7 +481,7 @@ object Trees {
 
   // JavaScript expressions
 
-  case class JSNew(ctor: Tree, args: List[Tree])(
+  case class JSNew(ctor: Tree, args: List[TreeOrJSSpread])(
       implicit val pos: Position) extends Tree {
     val tpe = AnyType
   }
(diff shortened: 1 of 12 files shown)


ir/src/main/scala/org/scalajs/core/ir/Trees.scala
```diff
@@ -30,8 +30,14 @@ object Trees {
     }
   }
 
+  /** Either a `Tree` or a `JSSpread`.
+   *
+   *  This is the type of actual arguments to JS applications.
+   */
+  sealed trait TreeOrJSSpread extends IRNode
```
gzm0: Any reason this is not an abstract class?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -1096,6 +1097,11 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
 
       require(!allowSideEffects || allowUnpure)
 
+      def testJSArg(tree: TreeOrJSSpread): Boolean = tree match {
+        case JSSpread(_) => false
```
gzm0: Unrelated to this commit, but couldn't we actually recurse here?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/IRChecker.scala
```diff
@@ -663,12 +663,11 @@ private final class IRChecker(unit: LinkingUnit,
     typecheck(tree, env)
   }
 
-  private def typecheckExprOrSpread(tree: Tree, env: Env): Type = {
+  private def typecheckExprOrSpread(tree: TreeOrJSSpread, env: Env): Unit = {
```
gzm0: Beautiful how this falls into place :)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala
sjrd: Yes, I think we could, in ES 2015 mode and with a bit of care. But as you say, this commit preserves the old behavior.

Also that's because we say that `JSSpread` can only be given a JS `Array`, but in ES 2015, `...spread` accepts any iterable (a thing with a `[Symbol.iterator]()` method), so in fact, in ES 2015 mode, `JSSpread` actually accepts any iterable, whose expansion can definitely have side effects.

ir/src/main/scala/org/scalajs/core/ir/Trees.scala
sjrd: Only the fact that it doesn't quite make sense. It's basically a union type. If in a later version of the IR, we have to add another such union type (because of a new feature in ES, for example), we would need to make the other one a `trait`, and then we would have a hard time arguing why `TreeOrJSSpread` was an abstract class and not the new one ;)

gzm0: Fair enough :)

Issue #3088, opened 2017-08-10
Link all classes that are mentioned in signatures and types


sjrd:
The Analyzer discards some classes that are not used at all, *except* that they are mentioned in the signatures of methods and possibly as intermediate types of `val`s and other expressions (but they are always upcast before calling a method).

These classes are not necessary for the optimizer or the emitter to do their job, but the fact that they are missing makes a `LinkingUnit` internally inconsistent. This is why the IRChecker needs to still receive the Infos of those.

We should investigate what we can do about this, so that the classes are kept in the `LinkingUnit`, to make them internally consistent.

gzm0:
@sjrd are you actively working on this? I will need this to make the linker asynchronous. Otherwise the IR checker also needs to handle asynchronous file loading. And that is neither nice to implement nor does it conceptually make sense.

gzm0:
I don't have time this week and neither on the weekend. So if you have cycles until then, go for it, otherwise I'll take it. Just keep me updated if you make any progress so we do not duplicate work.

gzm0:
I'll have a look at this now for the next hour.

gzm0:
Seems like `Definitions` already has a required method. PR on the way.

PR #3251, opened 2018-01-13
CI ONLY wip for #3088

sjrd:
Just want to see whether it passes the CI. Will need serious vetting of the performance impact of basic linking though!


gzm0:
Obsolete.

PR #3253, opened 2018-01-14
Fix #3252: Filter `scala.Serializable` out of the interfaces of JS types.

### LINKED ISSUE #3252: The .sjsir for SAM JS functions in 2.11 -Xexperimental extend scala.Serializable
The scalac compiler systematically adds `extends scala.Serializable` to SAM-generated classes. This is problem for the .sjsir files that we generate since #3223 for anon JS functions in 2.11 with `-Xexperimental`, because they are `AbstractJSType`s, and those cannot extend Scala trait.

This issue does not cause any visible problem, because these .sjsir files are only ever read by the IR checker when those SAM types appear in signatures. They are not fully *linked*, and therefore the Analyzer does not complain about the erroneous hierarchy. However, the issue surfaces if we try to implement #3088.

sjrd:
`scala.Serializable` is added automatically by scalac on all SAM-generated classes, which includes JS function types. Since ca86e39f0fc22e6a1bedc422c3f65552d243333e, we generate a actual .sjsir file for JS function types in Scala 2.11 with `-Xexperimental`, which are invalid if they extend `scala.Serializable`.

Therefore, this commit explicitly filters out `scala.Serializable`, the same way we filter out `scala.Dynamic`.

There are no tests because it is not observable without #3251.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -896,14 +896,21 @@ abstract class GenJSCode extends plugins.PluginComponent
           hashedMemberDefs, Nil)(OptimizerHints.empty)
     }
 
+    private lazy val jsTypeInterfacesBlacklist: Set[Symbol] =
+      Set(DynamicClass, SerializableClass) // #3118, #3252
+
     private def genClassInterfaces(sym: Symbol, forJSClass: Boolean)(
         implicit pos: Position): List[js.Ident] = {
+
+      val blacklist =
+        if (forJSClass) jsTypeInterfacesBlacklist
+        else Set.empty[Symbol]
+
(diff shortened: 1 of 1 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -898,12 +898,16 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     private def genClassInterfaces(sym: Symbol, forJSClass: Boolean)(
         implicit pos: Position): List[js.Ident] = {
+
+      val blacklist: Set[Symbol] =
+        if (forJSClass) Set(DynamicClass, SerializableClass) // #3118, #3252
```
gzm0: Ok with me, but you might want to consider making this set more static for speed purposes.

PR #3255, opened 2018-01-14
Fix #3033: Rehaul JSEnv API

### LINKED ISSUE #3033: Rethink JSEnv implementation APIs

gzm0:
This is a full rehaul of the JSEnv API. Highlights include:
- No synchronous interface.
- Proper stopping/closing semantics.
- Proper message passing semantics.
- Removal of JSConsole.
- Reduce implementation inheritance.


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
```diff
@@ -2,246 +2,147 @@ package org.scalajs.jsenv.test
...
-
-    for (i <- 0 until n) {
-      com.send("ping")
-      assertEquals("pong", com.receive())
-    }
-
-    com.close()
-    com.await(DefaultTimeout)
-  }
+  // TODO: What should happen if the JS side terminates?
```
gzm0: @sjrd Any advice on this? Also, we need a way to inject VM-side termination into the test, since this is not specified in the `JSEnv` interface (and a VM does not need to support it).

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -9,63 +9,95 @@
...
+    case Input.FilesToLoad(files) => files
+    case _ =>
+      throw new IllegalArgumentException(
+        s"$name does not support input $input")
+  }
 
-  protected class AsyncNodeRunner(files: Seq[VirtualJSFile])
-      extends AsyncExtRunner(files) with AbstractBasicNodeRunner
+  private def env: Map[String, String] = {
+    System.getenv().asScala.toMap + ("NODE_MODULE_CONTEXTS" -> "0") ++ config.env
```
gzm0: This should probably be part of `ExternalJSRun`. I'm just not sure how to deal with the merging of properties.

project/NodeJSEnvForcePolyfills.scala
```diff
@@ -5,17 +5,32 @@ import org.scalajs.jsenv.nodejs._
...
+      onMessage: String => Unit): JSComRun = {
+    nodeJSEnv.startWithCom(patchInput(input), runConfig, onMessage)
+  }
+
+  private def patchInput(input: Input): Input = input match {
+    case Input.FilesToLoad(fs) => Input.FilesToLoad(forcePolyfills +: fs)
+
+    case _ =>
+      throw new IllegalArgumentException(s"Unsupported input: ${input.getClass}")
+  }
```
gzm0: Should we provide a utility for this in JSEnv? I'm just not sure how to specify it so that we are sure we can always support it for all `Input` types.

nodejs-env/src/test/scala/org/scalajs/jsenv/nodejs/NodeJSTest.scala
```diff
@@ -1,79 +1,14 @@
...
-    com.await(DefaultTimeout)
-  }
-
-}
+class NodeJSRunTest extends RunTests(new NodeJSEnv, withCom = false)
+class NodeJSRunComTest extends RunTests(new NodeJSEnv, withCom = true)
+class NodeJSTimeoutRunTest extends TimeoutRunTests(new NodeJSEnv, withCom = false)
+class NodeJSTimeoutRunComTest extends TimeoutRunTests(new NodeJSEnv, withCom = true)
+class NodeJSComTest extends ComTests(new NodeJSEnv)
+class NodeJSTimeoutComTest extends TimeoutComTests(new NodeJSEnv)
```
gzm0: I hate that we have to do the expansion this way, but I haven't figured out a better way. JUnit4 does not support dynamic test generation :-/

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
gzm0: Note that this is notably causing the CI to fail right now in `library/test`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/Run.scala
```diff
@@ -0,0 +1,47 @@
...
+import scala.concurrent._
+import scala.concurrent.duration.Duration
+import scala.concurrent.ExecutionContext.Implicits.global
+
+private[sbtplugin] object Run {
+  /** Starts and waits for a run on the given [[JSEnv]] interruptibly.
+   *
+   *  Interruption can be triggered by typing anything into stdin.
+   */
+  def runInterruptible(jsEnv: JSEnv, input: Input, config: RunConfig): Unit = {
```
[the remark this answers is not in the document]
gzm0: We talked about this:
https://github.com/scala-js/scala-js/issues/3033#issuecomment-326749049 :P

Note that in the current implementation we indeed do *not* redirect stdin.

The problem with "run to completion" is that it is not supported by webby JS envs (PhantomJS, Selenium).

> Node.js application that would actually read things from its standard input.

I really hope nobody builds this in SJS, just as a side note :P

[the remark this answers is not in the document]
sjrd: Ah, right. But the `NodeJSEnv` (not necessarily other JSEnvs) still does guarantee run-to-completion by default, doesn't it? Otherwise even our `helloworld/run` on the CI is not going to work.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
```diff
@@ -2,246 +2,147 @@ package org.scalajs.jsenv.test
...
+  // Is it OK to fail the com in this case?
+  // Ironically, I did not find a way in Java to detect that a TCP socket got
+  // closed gracefully (when reading or writing from it).
+  // Therefore, I actually struggle to implement this properly (without a
+  // catch-all-any-exception-is-fine approach).
```
sjrd: https://stackoverflow.com/a/10241044/1829647

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
sjrd: You can define in `ComTests` a method like:
```scala
def getTerminateVMJSCode(): Option[String] = None
```
which can be overridden by subclasses of `ComTests` to provide the appropriate termination code. By default, `None` indicates that is not supported. The test can then call that method, and `assume` it `isDefined`.

Now about what should be the nice behavior, well ... I guess we should complete the `run.future` with a successful value. Any issue with doing that?

sjrd:
I just discovered https://github.com/brettwooldridge/NuProcess. I wonder whether we could use that to build an even more asynchronous API, especially around stdout and stderr.

It "only" supports Windows, Linux and Mac OS, on JDK 7+.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,66 @@ package org.scalajs.jsenv
...
+   *  onMessage(m_1), ..., onMessage(m_k)
+   *  }}}
+   *  exactly in this order.
+   *  <li>If `t` and `r` keep running indefinitely and `t` sends n messages,
+   *  `r` receives n messages.
+   *  </ul>
+   *
+   *  @param onMessage Callback invoked each time a message is received from the
+   *      JS VM. The implementation may not call this anymore, once
+   *      [[JSRun#future]] of the returned [[JSComRun]] is completed.
```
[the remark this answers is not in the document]
gzm0: Indeed, I need to add the guarantee that all `onMessage` callbacks are completed.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -9,63 +9,95 @@
...
+import org.scalajs.core.ir.Utils.escapeJS
 import org.scalajs.core.tools.io._
 import org.scalajs.core.tools.logging._
 
-import java.io.{ Console => _, _ }
-
-class NodeJSEnv(config: NodeJSEnv.Config) extends AbstractNodeJSEnv {
+import java.io._
 
+final class NodeJSEnv(config: NodeJSEnv.Config) extends JSEnv {
```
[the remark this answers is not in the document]
gzm0: Why? `JSDomNodeJSEnv` does not share any common code (except for the environment inheritance) with the current `NodeJSEnv` anymore. In any case, users that wish to adapt the environment, may use composition (like `NodeJSEnvForcePolyfills`).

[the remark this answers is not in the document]
sjrd: Ah right. I get it, now. You're right, it's perfectly fine.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/Run.scala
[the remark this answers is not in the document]
gzm0: Yes. A JSEnv *may* provide run to completion by default.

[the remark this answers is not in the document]
sjrd: OK. It makes sense, now.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,276 @@
...
+            case s: AwaitingConnection =>
+              throw new IllegalStateException(s"Unexpected state: $s")
+
+            case Connected(_, _, js2jvm) =>
+              try {
+                val len = js2jvm.readInt()
+                val carr = Array.fill(len)(js2jvm.readChar())
+                handleMessage(String.valueOf(carr))
+              } catch {
+                case _: IOException if state == Closing =>
```
gzm0: Note to self: We should also catch `EOFException` here for graceful closing (when other end terminates).

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,276 @@
...
+
+  // TODO replace this with scheduled tasks on the execution context.
+  private[this] val receiver = new Thread {
+    setName("ComRun receiver")
+
+    override def run(): Unit = {
+      try {
+        awaitConnection()
+
+        while (!promise.isCompleted) {
```
gzm0: This is not the right condition to loop on. In the close procedure, it will loop until the async callback is called.

gzm0:
> I just discovered https://github.com/brettwooldridge/NuProcess. I wonder whether we could use that to build an even more asynchronous API, especially around stdout and stderr.

I have given this some thought and I think we shouldn't rely on the extra dependency. Notably I'm not even sure we could provide a better interface with it (did you have something specific in mind)? Further, memory/CPU wise, I don't expect process invocation to be a bottleneck when testing.

sjrd:
The only issue I have that would be solved with that is the fact that the stdout and stderr streams are, well, streams. Which means we wouldn't be able to implement this on the JS platform, for example on Node.js.

Btw for the same reason I'm thinking that we might need to review the API of virtual files about that. Currently it all works well on Node.js because it has synchronous file APIs, but it wouldn't work with a browser API. It's a bit far fetched, though, because then we'd need to make whole linker API asynchronous! Currently the workaround when we use the linker in a browser context is to load the entire classpath in memory before starting to link.

Anyway, I'm fine with leaving it as is and not introduce that dependency.

Andrei-Pozolotin:
in the context of #3273 please add full logging to runners, for example: (0.6.x version)

sample log output
```

JS-VM launch: fetcher @ Thread[main,5,main]
  exec: ./test-tool/node/node
  args: 
  vars: NODE_PATH=./test-tool/node/node_modules
  libs: ./test-tool/webjars/jquery/jquery.js; ./target/junit-tools.js
  code: testFrameworkInfo.js


JS-VM launch: master @ Thread[main,5,main]
  exec: ./test-tool/node/node
  args: 
  vars: NODE_PATH=./test-tool/node/node_modules
  libs: ./test-tool/webjars/jquery/jquery.js; ./target/junit-tools.js
  code: testMaster.js


JS-VM launch: worker @ Thread[main,5,main]
  exec: ./test-tool/node/node
  args: 
  vars: NODE_PATH=./test-tool/node/node_modules
  libs: ./test-tool/webjars/jquery/jquery.js; ./target/junit-tools.js
  code: testSlave.js

```

sample log code
```scala
    trait CustomRunner extends AbstractNodeRunner {
      // provide clean env var
      override protected def getVMEnv() : Map[ String, String ] = {
        env
      }
      // provide runner report
      override protected def startVM() : Process = {
        val envArgs = getVMArgs()
        val envVars = getVMEnv()
        val argsList = executable +: envArgs
        val builder = new ProcessBuilder( argsList : _* )
        builder.environment().clear()
        for ( ( name, value ) <- envVars ) builder.environment().put( name, value )
        logger.debug( runnerReport( executable, envArgs, envVars, libs, code ) )
        builder.start()
      }
    }

  def hasStackMethod( method : String ) : Boolean = {
    val trace = Thread.currentThread().getStackTrace()
    trace.find( entry => method == entry.getMethodName ).isDefined
  }

  def hasFetcher = hasStackMethod( "fetchFrameworkInfo" )
  def hasMaster = hasStackMethod( "createRemoteRunner" )
  def hasWorker = hasStackMethod( "createSlave" )

  def runnerType = {
    if ( hasFetcher ) "fetcher"
    else if ( hasMaster ) "master"
    else if ( hasWorker ) "worker"
    else "unknown"
  }

  def runnerReport(
    envExec : String, envArgs : Seq[ String ], envVars : Map[ String, String ],
    libs : Seq[ ResolvedJSDependency ], code : VirtualJSFile
  ) : String = {
    s"""
JS-VM launch: ${runnerType} @ ${Thread.currentThread()}
  exec: ${envExec}
  args: ${envArgs.mkString( "; " )}
  vars: ${envVars.map { case ( key, value ) => s"$key=$value" }.mkString( "; " )}
  libs: ${libs.map( _.lib.path ).mkString( "; " )}
  code: ${code.path}
"""
  }

```

BTW 1), please rename `slave` -> `worker` - it is not politically correct in the US :-)

BTW 2), do you really need to start JS-VM 3 times? :-)

gzm0:
> BTW 2), do you really need to start JS-VM 3 times? :-)

We start a VM for each thread we get invoked on in order to emulate the same amount of parallelism for testing. I realize now that this might be annoying if you invoke tests from a thread pool that is not just for testing invocation.

If you need another way to control parallelism, please open an issue and we can discuss. However, it is probably non-trivial to do this without risk of starving the system.

Andrei-Pozolotin:
>please open an issue and we can discuss

in 0.6.22 `TestAdapter` is already much better then `ScalaJSFramework` (except 5 second delay we discussed in gitter), so, no issue for now, thank you.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
gzm0: Hah, turns out inheriting the environment is the default of process builder. So I just need to remove the line that clears it.

nodejs-env/src/test/scala/org/scalajs/jsenv/nodejs/NodeJSTest.scala
gzm0: I hacked JUnit a lot :) Now we have a way.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,66 @@ package org.scalajs.jsenv
...
   /** Human-readable name for this [[JSEnv]] */
-  def name: String
+  val name: String
 
-  /** Prepare a runner with the specified JavaScript files. */
-  def jsRunner(files: Seq[VirtualJSFile]): JSRunner
+  /** Starts a new (asynchronous) JS run.
+   *
+   *  This must throw an [[IllegalArgumentException]] immediately, if the
+   *  type of [[input]] is unknown or a config value cannot be supported.
```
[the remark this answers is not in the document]
gzm0: Ah, I see. I didn't think that we wanted to add configuration values that are specified that they must be used (e.g. stuff that alters behavior). I don't think it is a good idea to say that config values can simply be ignored.

I think we have two options here:
- Commit to not adding config values that alter observable behavior (but are informational, like a thread pool).
- Add a validator construct to which `JSEnv`s can tell the options they know about. If other options are present (in non-default values), the validator can fail even for a `JSEnv` that wasn't compiled with that option.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,66 @@ package org.scalajs.jsenv
...
+   * 
+   *  and `r` observes `onMessage(m_k)` (k <= n) but not `onMessage(m_{k+1})`,
+   *  `r` must observe
+   *  {{{
+   *  onMessage(m_1), ..., onMessage(m_k)
+   *  }}}
+   *  exactly in this order.
+   *  <li>If `t` and `r` keep running indefinitely and `t` sends n messages,
+   *  `r` receives n messages.
+   *  </ul>
```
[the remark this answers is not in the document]
gzm0: >at-most once ordered delivery

This doesn't account for the requirement that if `m_2` is delivered `m_1` must be delivered, does it?

js-envs/src/main/scala/org/scalajs/jsenv/RunConfig.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  def withOnOutputStream(onOutputStream: OnOutputStream): RunConfig =
+    copy(onOutputStream = Some(onOutputStream))
+
+  def withInheritOut(inheritOutput: Boolean): RunConfig = {
+    if (onOutputStream.isEmpty && !inheritOutput) {
+      throw new IllegalArgumentException("You may not set inheritOutput to false " +
+          "if onOutputStream is not set.")
+    }
```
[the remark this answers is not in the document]
gzm0: Ah, maybe. I deferred changing this until we know if we implement the validator construct. If yes, it should definitely go in there.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -9,63 +9,95 @@
...
+      files.foreach {
+        case file: FileVirtualJSFile =>
+          val fname = file.file.getAbsolutePath
+          writer.write(s"""require("${escapeJS(fname)}");\n""")
+        case f =>
+          IO.writeTo(f, writer)
+          writer.write('\n')
+      }
+    } finally {
+      writer.close()
```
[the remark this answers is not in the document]
gzm0: Yes, its might be over-prudent, but it's OK.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/Support.scala
```diff
@@ -0,0 +1,50 @@
...
+          |  function startsWithAnyOf(s, prefixes) {
+          |    for (var i = 0; i < prefixes.length; i++) {
+          |      // ES5 does not have .startsWith() on strings
+          |      if (s.substring(0, prefixes[i].length) === prefixes[i])
+          |        return true;
+          |    }
+          |    return false;
+          |  }
+          |  var nodeWillDeduplicateEvenForOneArgument = startsWithAnyOf(
+          |      process.version, ["v0.", "v1.", "v2.0."]);
```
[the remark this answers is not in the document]
gzm0: Done.

gzm0:
Note to self: Stuff to do after this PR:

- Change `TestAdapter` to take `Input`.
- Change sbt plugin to use `Input`.
- Reduce threads by passing on a special purpose thread pool.
- Investigate better logging.

gzm0:
Jenkins, retest this please.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+object ComRun {
+  /** Starts a [[JSComRun]] using the provided [[JSRun]] launcher.
+   *
+   *  @param config Configuration for the run.
+   *  @param onMessage callback upon message reception.
+   *  @param startRun [[JSRun]] launcher. Gets passed a [[VirtualJSFile]] that
+   *      initializes `scalaJSCom` on `global`. Requires Node.js libraries.
+   */
+  def start(config: RunConfig, onMessage: String => Unit)(
+      startRun: VirtualJSFile => JSRun): JSComRun = {
```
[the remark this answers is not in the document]
gzm0: Maybe we should just change `startRun` it to `(Input, RunConfig) => JSRun`. This will allow to simply pass in an eta-expansion and allows `ComRun` to adapt the `Input` and `RunConfig` if necessary.

gzm0:
Rebased onto master.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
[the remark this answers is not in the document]
sjrd: Ah, indeed, it doesn't.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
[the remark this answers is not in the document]
sjrd: It should probably be an `Input`, yes.

Not sure why you want to give the `RunConfig` back to `startRun`, though. Why would `start` want to change the `RunConfig`?

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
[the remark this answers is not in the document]
sjrd: My point is that this `private def write` does not *own* `out`. Therefore, it is not supposed to close `out`. Maybe the client who calls `write` intends to keep using `out` after `write` has returned.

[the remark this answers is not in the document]
gzm0: Fair enough :) Removed.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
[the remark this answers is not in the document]
gzm0: I realize right now that this is much worse: The specification that `scalaJSCom` needs to be on global is bad: It is difficult to make that happen with proper JS modules. The only way I can think of is:
- Create a module, say `sjsCom.js`
- Add `scalaJSCom = require('sjsCom.js')` to the top of every file

I recall now that this is why this did not take an `Input`: It is the callers responsibility to assemble this file in a context such that the execution works properly (JSDOMNodeJSEnv has an example of this).

So maybe we should pass a String instead that is specified to be an IIFE that returns the `scalaJSCom`?

(This doesn't solve the problem of making the specification work with proper JS modules, but at least we have this problem now in the caller, which at least knows the type of the actual input).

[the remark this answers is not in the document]
sjrd: > So maybe we should pass a String instead that is specified to be an IIFE that returns the scalaJSCom?

It's not very nice, but it would get the job done.

Otherwise, it's not *that* hard to find the global object in JavaScript: https://stackoverflow.com/a/3277192/1829647. And it will probably get easier in the future: https://github.com/tc39/proposal-global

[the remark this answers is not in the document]
sjrd: Also: since `scalaJSCom` is setup by the `JSEnv`, each `JSEnv` can independently know how to find the global object for its environment. For example, `NodeJSEnv` knows it can use `global`. A weird `JSEnv` for an unknown environment for which accessing the global object is done through `harry` can set up `scalaJSCom` accordingly.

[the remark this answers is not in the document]
gzm0: Ok, fair enough. But then I get the feeling that what we currently have is actually the best option: `ComSupport` is Node.js specific anyways, and indeed can safely use `global`. However, returning it as an `Input` is funky, since then the client needs to merge them. Returning a `String` is not super nice.

gzm0:
Updated:

- Removed the double `close`
- Added an `UnsupportedInputException`.

TODO:

- Figure out details about `RunConfig` (what to do with new options).

gzm0:
@sjrd WDYT about adding a verifier for `RunConfig`? I'm a bit split since it is some more code, but probably the most flexible option :S

sjrd:
What exactly do you have in mind? Are you talking about https://github.com/scala-js/scala-js/pull/3255/files#r161516032 or https://github.com/scala-js/scala-js/pull/3255/files#diff-cf497553966fa6b2293facc6a6d947eeR26? Or something else?

gzm0:
The second one. Basically it's about the semantics of `RunOptions`: I do not think we should make it optional for a JSEnv to adhere to them by default. So that leaves us with two ways of maintaining compatibility:

- Restricting ourselves to only add options in the future that can reasonably made optional (n.b. they may not cause a change in behavior, but maybe efficiency).
- Providing a forward compatible way for a `JSEnv` to determine if it handled all the set options.

The second option would be the validator framework where a `JSEnv` can tell the validator what options it supports and the validator checks this with the set options. Since the validator is provided by the base library, it will always know about all options.

Use of the validator would look something like this:

```scala
def start(input: Input, config: RunConfig): Unit = {
  config.validator()
    .supportsA()
    .supportsC()
    .validate()  // fails if B is non-default.
}
```

sjrd:
Right. Yes, I think it should be good to have a validator like that. It doesn't seem to be *that* much more code, is it?

js-envs/src/main/scala/org/scalajs/jsenv/RunConfig.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    /** The caller supports [[onOutputStream]]. */
+    def supportsOnOutputStream(): Validator = copy(onOutputStream = true)
+
+    /** Validates that the bound [[RunConfig]] only sets options that are supported.
+     *
+     *  @throws java.lang.IllegalArgumentException if there are unsupported options.
+     */
+    def validate(): Unit = {
+      val cfg = RunConfig.this
```
sjrd: IMO a more idiomatic API would be that `Validator` is not tight to a `RunConfig` upfront, but is declared in the `object RunConfig` instead and that `validate()` takes the `config: RunConfig` as a formal parameter.

This avoids the `#` types, and allows a `JSEnv` to create its `RunConfig.Validator` once and for all, and reuse it to validate all the `RunConfig`s it receives.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -0,0 +1,165 @@
...
+        run.future.value.get.isFailure)
+    } finally {
+      run.close()
+    }
+  }
+
+  @Test(expected = classOf[RunConfig.FailedValidationForTestException])
+  def ensureValidate: Unit = {
+    val cfg = RunConfig().withFailValidationForTest(true)
+    kit.start("", cfg).close()
```
sjrd: So that's what the dummy `private[jsenv]` exception type and field in `RunConfig` are for? That's pretty twisted. (also the error message that JUnit will be super confusing and not very actionable :-s)

js-envs/src/main/scala/org/scalajs/jsenv/RunConfig.scala
gzm0: Good point.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
gzm0: Yes. I can add a big comment explaining what this does, but I don't know how to do this better so we can still test it. I think it is important that we test this.

sjrd: Neither do I--and I tried finding a better way--, or I would have suggested something :(
And yes, it's probably important, because otherwise custom JSEnv authors will never call that thing.

gzm0: I have
- removed the custom exception (it is not necessary for the test to be effective)
- improved the name of the option
- added some more doc / comments

gzm0:
Updated. Note the additional commit.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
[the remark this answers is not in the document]
gzm0: Un-removed it and made it explicit that this things owns the stream. Otherwise the `BufferedWriter` needs to be flushed explicitly which is not nice.

sjrd:
Summary of CI errors:

* Scaladoc error in jsenvs-test-kit

```
[error] .../js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/JSEnvSuite.scala:13: Tag '@RunWith(classOf[JSEnvSuiteRunner])' is not recognised
[error] /** Conformance test suite for any [[JSEnv]] implementation.
[error] ^
[error] one error found
[error] (jsEnvsTestKit/compile:doc) Scaladoc generation failed
```

* `SourceMapTest` failure in `test-suite-ecma-script6 testSuite=testSuite` (all Scala versions)

```
[info] Test org.scalajs.testsuite.compiler.SourceMapTest.workTest0 started
[error] Test org.scalajs.testsuite.compiler.SourceMapTest.workTest0 failed: java.lang.AssertionError: null, took 0.108999936 sec
[error]     at java.lang.AssertionError.fillInStackTrace(.../test-suite/js/target/scala-2.12/scalajs-test-suite-test-fastopt.js:262754)
[error]     at java.lang.AssertionError.this(.../test-suite/js/target/scala-2.12/scalajs-test-suite-test-fastopt.js:262801)
[error]     at java.lang.AssertionError.this(.../test-suite/js/target/scala-2.12/scalajs-test-suite-test-fastopt.js:309532)
[error]     ...
```
(same for all the other tests in that suite)

gzm0:
Rebased on top of #3306.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,72 @@ package org.scalajs.jsenv
...
+   *
+   *  @throws java.lang.IllegalArgumentException if the value of `input` or
+   *      `config` cannot be supported.
+   */
+  def start(input: Input, config: RunConfig): JSRun
+
+  /** Like [[start]], but initializes a communication channel.
+   *
+   *  Inside the VM this is to provida a global JavaScript object named
+   *  `scalajsCom` that can be used to interact with the message channel. It's
```
sjrd: It's -> Its

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,72 @@ package org.scalajs.jsenv
...
+   *  In this case, [[JSRun#future]] should be failed instead.
+   *
+   *  @throws java.lang.IllegalArgumentException if the value of `input` or
+   *      `config` cannot be supported.
+   */
+  def start(input: Input, config: RunConfig): JSRun
+
+  /** Like [[start]], but initializes a communication channel.
+   *
+   *  Inside the VM this is to provida a global JavaScript object named
```
sjrd: provid**e**

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,72 @@ package org.scalajs.jsenv
...
+   *  {{{
+   *  // initialize com (with callback). May only be called once.
+   *  scalajsCom.init(function(msg) { console.log("Received: " + msg); });
+   *
+   *  // send a message to host system
+   *  scalajsCom.send("my message");
+   *  }}}
+   *
+   *  We describe the expected message delivery guarantees by denoting the
+   *  transmitter as `t` and  the reciever as `r`. Both the JVM and the JS end
```
sjrd: receiver

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,72 @@ package org.scalajs.jsenv
...
+  val name: String
 
-  /** Prepare a runner with the specified JavaScript files. */
-  def jsRunner(files: Seq[VirtualJSFile]): JSRunner
+  /** Starts a new (asynchronous) JS run.
+   *
+   *  This may only throw if value of `input` is unknown or `config` cannot be
+   *  supported. To verify whether a [[RunConfig]] can be supported in a forward
+   *  compatible manner (i.e. when new options are added in later versions)
+   *  implementations of [[JSEnv]]s must use [[RunConfig#Validator]].
```
sjrd: `#` -> `.`

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -11,10 +11,72 @@ package org.scalajs.jsenv
...
+   *  {{{
+   *  onMessage(m_1), ..., onMessage(m_k)
+   *  }}}
+   *  exactly in this order.
+   *  <li>If `t` and `r` keep running indefinitely and `t` sends n messages,
+   *  `r` receives n messages.
+   *  </ul>
+   *
+   *  @param onMessage Callback invoked each time a message is received from the
+   *      JS VM. The implementation may not call this anymore, once
```
sjrd: no comma

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+import org.scalajs.io.{VirtualJSFile, MemVirtualJSFile}
+import org.scalajs.jsenv._
+
+import scala.collection.immutable
+import scala.concurrent._
+import scala.util.{Failure, Success}
+import scala.util.control.NonFatal
+
+// TODO Replace this by a better execution context on the RunConfig.
+import scala.concurrent.ExecutionContext.Implicits.global
```
sjrd: I take it this TODO is for one day, not for this PR?

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+                 } catch {
+                   case _: EOFException =>
+                     // JS end terminated gracefully. Close.
+                     close()
+                 }
+            }
+          }
+        } catch {
+          case _: IOException if state == Closing =>
+            // We got interrupted by a gracful close.
```
sjrd: graceful

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+
+  def close(): Unit = synchronized {
+    val oldState = state
+
+    // Signal receiver thread that it is OK if socket read fails.
+    state = Closing
+
+
```
sjrd: double blank line

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+    oldState match {
+      case c: Connected =>
+        // Interrupts the receiver thread and signals the VM to terminate.
+        closeAll(c)
+
+      case Closing | _: AwaitingConnection =>
```
sjrd: No space after `:` when it's in a `|`-separate list of alternatives.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
```diff
@@ -0,0 +1,302 @@
...
+  private def awaitConnection(): Unit = {
+    var comSocket: Socket = null
+    var jvm2js: DataOutputStream = null
+    var js2jvm: DataInputStream = null
+
+    try {
+      comSocket = serverSocket.accept()
+      serverSocket.close()  // we don't need it anymore.
+      jvm2js = new DataOutputStream(
+        new BufferedOutputStream(comSocket.getOutputStream()))
```
sjrd: 4 spaces

nodejs-env/src/test/scala/org/scalajs/jsenv/nodejs/NodeJSSuite.scala
```diff
@@ -0,0 +1,11 @@
+package org.scalajs.jsenv.nodejs
+
+import org.scalajs.jsenv.test._
+
+import org.junit.runner.RunWith
+
+@RunWith(classOf[JSEnvSuiteRunner])
+class NodeJSSuite extends JSEnvSuite(
+  JSEnvSuiteConfig(new NodeJSEnv)
```
sjrd: 4 spaces

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/Run.scala
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+    }
+
+    val run = jsEnv.start(input, config)
+    try {
+      readThread.start()
+
+      Await.result(
+        Future.firstCompletedOf(List(readPromise.future, run.future)),
```
sjrd: 4 spaces

sjrd:
One question: why is the commit "Always assume test-inteface is present" necessary here? I thought I'd figure it out by the time I was done reviewing the whole PR, but I didn't ^^

gzm0:
> One question: why is the commit "Always assume test-inteface is present" necessary here? I thought I'd figure it out by the time I was done reviewing the whole PR, but I didn't ^^

Because with the new interface, we do not expect a `JSEnv` to provide termination anymore. So if we bring up a `JSEnv` in which the test-interface is not loaded (and we hence silently do not call it), we will wait forever (or we would have to decide on a timeout, which is also not very nice). #3294 will turn this into a link time error, so I think this is acceptable.

sjrd:
Ah I see. OK.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -0,0 +1,176 @@
...
+    /* This test also tests a failure mode where the ExternalJSRun is still
+     * piping output while the client calls close.
+     */
+    val run = kit.start("", RunConfig())
+    run.close()
+
+    if (withCom)
+      Await.result(run.future, Duration.Inf)
+    else
+      Await.ready(run.future, Duration.Inf)
```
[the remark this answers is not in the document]
gzm0: Refactored this into the kit. Same comment than inside the kit.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TestComKit.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import org.scalajs.jsenv._
+
+import org.junit.Assert.fail
+
+import scala.collection.immutable
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+private[test] final class TestComKit(jsEnv: JSEnv) {
```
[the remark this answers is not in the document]
gzm0: The API is not really stable enough. If you think we should provide something, I'd rather think of something cleaner in a subsequent PR.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -0,0 +1,176 @@
...
+    """ hasOutput "<script></script>\n";
+  }
+
+  @Test
+  def jsExitsTest: Unit = {
+    assumeTrue(config.terminateVMJSCode.isDefined)
+
+    val run = kit.start(config.terminateVMJSCode.get, RunConfig())
+    try {
+      Await.result(run.future, Duration.Inf)
```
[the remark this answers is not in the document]
gzm0: Good point. I wrote all code such that it can time out.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutComTests.scala
```diff
@@ -1,162 +1,89 @@
...
 
   @Test
-  def receiveTimeoutTest: Unit = {
-
-    val com = comRunner(s"""
+  def delayedReplyTest: Unit = {
+    val run = kit.start(s"""
       scalajsCom.init(function(msg) {
-        setTimeout(scalajsCom.send, 2000, "Got: " + msg);
+        setTimeout(scalajsCom.send, 20, "Got: " + msg);
```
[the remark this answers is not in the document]
gzm0: Probably. Bumped to 200.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
gzm0: Yep.

gzm0:
Updated: I should have addressed all concerns (except for the public TestKit).

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TestComKit.scala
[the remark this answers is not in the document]
sjrd: OK, let's keep it private then, at least for now.

sjrd:
There's a Scaladoc error:
```
[info] Main Scala API documentation to .../nodejs-env/target/scala-2.11/api...
model contains 11 documentable templates
[error] .../nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala:195: Could not find any member to link for "VirtualJSFile".
[error]   /** Starts a [[JSComRun]] using the provided [[JSRun]] launcher.
[error]   ^
[error] one error found
[error] (nodeJSEnv/compile:doc) Scaladoc generation failed
```

gzm0:
Updated.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TestKit.scala
```diff
@@ -0,0 +1,156 @@
...
+          wait(m)
+        else
+          throw new TimeoutException("Timed out waiting for stdout")
+      }
+
+      stream
+    }
+
+    def onOutputStream(out: Option[InputStream],
+        err: Option[InputStream]) = synchronized {
```
sjrd: Scalastyle says that "public method must have explicit type".

gzm0:
Updated, passes now.

SethTisue:
in the community build I'm now seeing failures like

```
[scala-js] [info] Test org.scalajs.testsuite.compiler.SourceMapTest.workTest0 started
[scala-js] [error] Test org.scalajs.testsuite.compiler.SourceMapTest.workTest0 failed: java.lang.AssertionError: null, took 0.299000064 sec
[scala-js] [error]     at <jscode>.$c_jl_AssertionError.$c_jl_Throwable.fillInStackTrace__jl_Throwable(/Users/tisue/community.211-8/target-0.9.11/project-builds/scala-js-d0f50bf8b3014a5c1286c15f07cc86900faa7b37/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:313660)
[scala-js] [error]     at <jscode>.$c_jl_AssertionError.$c_jl_Throwable.init___T__jl_Throwable(/Users/tisue/community.211-8/target-0.9.11/project-builds/scala-js-d0f50bf8b3014a5c1286c15f07cc86900faa7b37/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:313707)
[scala-js] [error]     at java.lang.AssertionError.this(/Users/tisue/community.211-8/target-0.9.11/project-builds/scala-js-d0f50bf8b3014a5c1286c15f07cc86900faa7b37/test-suite/js/target/scala-2.11/scalajs-test-suite-test-fastopt.js:364376)
[scala-js] [error]     ...
```

?

let me know if you need more details (perhaps in a new ticket?)

sjrd:
@SethTisue A new ticket would be good. We forget closed PRs too easily.

Unless you manage to fix the issue on the community build side with this setting:
```scala
set wantSourceMaps in testSuite := false
```
?

SethTisue:
`set MyScalaJSPlugin.wantSourceMaps in testSuite := false` did the trick 👍

PR #3260, opened 2018-01-17
DO NOT MERGE Fix #3259: Generate overriding bridges in synthesized SAM wrappers.

### LINKED ISSUE #3259: SAM and subtyping causes linking errors
This is the same issue as https://github.com/scala/bug/issues/10512 in upstream Scala, although of course the `AbstractMethodError` manifests as a linking error instead.

The issue was fixed in Scala/JVM 2.12.5 with https://github.com/scala/scala/pull/6087. We need to port some of those changes to fix it on our side too.

sjrd:
Just like Scala/JVM needs to instruct the `LambdaMetaFactory` to generate bridges for methods overridden by the SAM, we need to directly generate appropriate bridges in our synthesized SAM wrappers.

Instead of generating actual *bridges*, in the sense that they call the canonical implementation of the method, we generate bodies for all the bridges just like for the canonical method. This is fine because the bodies are just calling the closure stored in a field of the instance anyway.

Only the last commit is part of this PR. It will have to be rebased on top of #3258.

Tested locally with:
```
> clean
> ++2.12.5-bin-707b0f3
> testSuite/test
> testSuiteJVM/test
```


hrhino:
Thanks! I was going to poke at this over the weekend in case you were busy, but this is better ;)

sjrd:

sjrd:
Rebased.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5154,8 +5152,35 @@ abstract class GenJSCode extends plugins.PluginComponent
...
+        val synthCls = samInfo.synthCls
+        val samBridges = if (synthCls == NoSymbol) {
+          Nil
+        } else {
+          import scala.reflect.internal.Flags.BRIDGE
+          synthCls.info.findMembers(excludedFlags = 0L, requiredFlags = BRIDGE).toList
+        }
+
+        for (sam <- samInfo.sam :: samBridges) {
+          if (seenEncodedNames.add(encodeMethodSym(sam).name))
```
gzm0: Why do we need this?

sjrd: Because the JVM backend does something similar (they call `.disctinct` [here](https://github.com/scala/scala/pull/6087/files#diff-2e0fd5042aa83407a80a7557cc2a666bR1375)). I guess this would happen if a SAM type extends several interfaces, each defining the same method, which would then collapse as the same bridge. The most reliable way for us to detect duplicates at this point is to test uniqueness of the `encodedName`.

sjrd:
Updated.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
gzm0: Hmmm... OK. Not very nice, but fair enough.

PR #3263, opened 2018-01-19
Get rid of the `core` and `tools` package namespaces.

sjrd:
Since the split into `io`, `logging` and `linker`, the `tools` namespaces does not mean anything anymore, so we lift its three components one level up.

This is also a good opportunity to more generally get rid of the `core` namespace. That namespace does not convey any useful meaning, and only make our packages longer for our users and ourselves.

Finally, we also rename `compiler` into `nscplugin`, since nsc is not the only Scala compiler anymore, and there will come a time when compiler plugins for other compilers will exist.

Concretely, this commit introduces the following renamings:

* `org.scalajs.core.ir`            -> `org.scalajs.ir`
* `org.scalajs.core.compiler`      -> `org.scalajs.nscplugin`
* `org.scalajs.core.tools.io`      -> `org.scalajs.io`
* `org.scalajs.core.tools.logging` -> `org.scalajs.logging`
* `org.scalajs.core.tools.linker`  -> `org.scalajs.linker`


sjrd:
For source "compatibility", it is possible for clients of the (now-dead) tools to cross-compile with the following trick:

in some `Compat.scala` file:
```scala
package org.scalajs.core.tools

private[tools] object CompatForMyFooLibrary
```
and then at use site, import as:
```scala
package foo

import org.scalajs._
import org.scalajs.core.tools._
import linker.Linker // for example
```

PR #3274, opened 2018-01-26
Fix #2998: Finish the new Linker API.

### LINKED ISSUE #2998: New Linker API

sjrd:


linker/shared/src/main/scala/org/scalajs/linker/standard/StandardLinkerImpl.scala
```diff
@@ -16,15 +16,16 @@ import java.util.concurrent.atomic.AtomicBoolean
...
 import org.scalajs.linker.frontend.LinkerFrontend
 import org.scalajs.linker.frontend.optimizer.IncOptimizer
 import org.scalajs.linker.backend.{LinkerBackend, BasicLinkerBackend}
 import org.scalajs.linker.irio._
 
-/** The Scala.js linker */
-final class Linker private (frontend: LinkerFrontend, backend: LinkerBackend)
+/** Standard implementation of a Scala.js linker. */
+private[standard] final class StandardLinkerImpl private (
+    frontend: LinkerFrontend, backend: LinkerBackend)
```
gzm0: Can't you make this `private`? The companion should still see it, no?

linker/shared/src/main/scala/org/scalajs/linker/Linker.scala
```diff
@@ -20,7 +20,7 @@ import org.scalajs.linker.irio._
  *  module initializers, link them together, and write the output to a writable
  *  .js file.
  */
-trait GenLinker {
+abstract class Linker {
```
gzm0: The constructor here should be private so that we can add abstract methods in minor versions.

linker/shared/src/main/scala/org/scalajs/linker/standard/LinkedClass.scala
```diff
@@ -66,7 +66,37 @@ final class LinkedClass(
...
+      staticMethods: List[Versioned[MethodDef]],
+      memberMethods: List[Versioned[MethodDef]]
+  ): LinkedClass = {
+    copy(
+        staticMethods = staticMethods,
+        memberMethods = memberMethods
+    )
+  }
+
+  private def copy(
```
gzm0: I'm OK with this change in principle, but I'm not sure we're on the same page here: We should be allowed to change `LinkingUnit` in a binary incompatible fashion in a minor version. So this is not a major concern (if it were, `ir.Trees` would be in a much worse position).

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardLinkerImpl.scala
sjrd: Indeed. Probably came from a transient state.

linker/shared/src/main/scala/org/scalajs/linker/standard/LinkedClass.scala
sjrd: We should be *allowed* to break it in a minor version (not a patch version), but we should *try* not to, as much as possible, in order to break (future) linker plugins.

That's the point of `standard.*` versus the rest: for the rest, we don't care. For `standard`, we try not to break stuff as long as we can avoid it, to preserve the ecosystem of linker plugins.

linker/shared/src/main/scala/org/scalajs/linker/Linker.scala
sjrd: Right, that's indeed what we designed in #2998. But eventually we should be allowed to open this up. We might want to allow users to write their own `Linker` (which would probably wrap an existing standard linker, not reimplement everything). For example that's what I had done in https://github.com/sjrd/scalajs-reflect.

sjrd:
> I'm not sure the commit "Introduce abstract LinkerFrontend and LinkerBackend in standard.*." is necessary. All subpackages of linker except for irio should have the same compatibility guarantees than standard. Therefore, it is completely fine if they refer to each other.

In #2998 we had defined `standard.*` as the "stable-ish" API: something that should stay backward compatible as long as possible, but with allowed breakages:

> That API is supposed to stay backward source and binary compatible across minor releases unless maintaining compatibility would prevent progress. It should not be broken without a reasonably good reason.

Unlike `standard.*`, the other packages besides `linker.*` and `irio.*` are truly unstable, and can break any time without second thought. So `standard.*` does have strong compatibility rules, if not strict *guarantees*, and IMO that's enough not to refer to unstable stuff from the standard stuff.

Btw, the abstract `LinkerFrontend` and `LinkerBackend` were part of the plan laid out in #2998.

> `ClearableLinker` needs to consolidate the batch flag with `StandardLinker.Config` somehow. We can file this as a follow up, or do it in this PR.

We don't want `ClearableLinker` to depend on a whole `StandardLinker.Config`. We should be able to use `ClearableLinker` with non-standard `Linker`s, whose config might be completely different. I'm not sure what we could even change in `ClearableLinker`.

The only thing that comes to mind would be to introduce `StandardLinker.clearable(config: StandardLinker.Config): ClearableLinker`, which would wrap a `StandardLinker(config)` inside a `new ClearableLinker(_, config.batchMode)`.

> sealed on `ModuleIntializer` and `Semantics.RuntimeClassNameMapper` is odd, since all subclasses are private. Note that we explicitly do not give the guarantee that we will never add subclasses.

I'm not sure it's that weird. We have admittedly worse abuses of `sealed` in `js.*` itself: https://github.com/scala-js/scala-js/blob/bad76528021ee7489171e53fc36b4dff5a6b04a4/library/src/main/scala/scala/scalajs/js/UnicodeNormalizationForm.scala#L24. It would be quite annoying while writing the code inside the linker if they weren't sealed, as we wouldn't get the exhaustivity warnings. It's not a big deal for those two, though, as they're not deconstructed very often

> `linker.ModuleKind` is sealed, it shouldn't be.

Meh ... I wouldn't like that one not to be sealed. I can't count all the times I've followed the exhaustivity warnings about `ModuleKind` in the `emitter.*`. But I guess we don't really have a choice.

sjrd:
Updated with:
* `private[linker]` constructor for `Linker`
* `private class StandardLinkerImpl`
* Unseal `ModuleKind` and `OutputMode`

I haven't unsealed `ModuleInitializer` nor `RuntimeClassNameMapper`. IMO it is perfectly fine to have a sealed class in public API, whose cases are not visible in the public API. There is no way for user code ever match on these things without exhaustivity warnings to begin with, so we can always add more cases. In fact, it could even considered the design pattern to follow for write-only data structures.

An alternative would be to use the following pattern:
```scala
abstract class ModuleInitializer private () {
  this: ModuleInitializer.ModuleInitializerImpl =>

  import ModuleInitializer._

  private[linker] def toImpl: ModuleInitializerImpl = this
}

/** Factory for [[ModuleInitializer]]s. */
object ModuleInitializer {
  private[linker] sealed abstract class ModuleInitializerImpl
      extends ModuleInitializer

  private[linker] final case class VoidMainMethod(moduleClassName: String,
      encodedMainMethodName: String)
      extends ModuleInitializerImpl

  private[linker] final case class MainMethodWithArgs(moduleClassName: String,
      encodedMainMethodName: String, args: List[String])
      extends ModuleInitializerImpl
}
```
but then the self type leaks to the public API (it is shown in the Scaladoc), even though it references a class that is not part of the public API.

I also haven't changed `ClearableLinker`, as it's not clear what, if anything, needs to done about it.

Do not merge, as #3277 should go in first.

gzm0:
> In #2998 we had defined standard.* as the "stable-ish" API: something that should stay backward compatible as long as possible, but with allowed breakages:

Ah, I see how you meant this now. You essentially wanted these levels:

1. Stable
2. Stable-ish
3. Unstable
4. Private

I thought we'll only have:

1. Stable
2. Stable-ish
3. Private

While I'm fine with having more levels of compatibility guarantee, I don't think we're drawing the line at the right point: Taking a users perspective, we should answer the question: "What can you do with an API level".

Use cases I can think of from the top of my head:

- Use the linker: super stable API
- Use another output mode: stable-ish API
- Modify the linker (plugin, reporting plugin, etc): stable-ish API? With the current setup, you cannot really do this with the stable-ish API, because you would have to write your own Frontend and Backend, which nobody will.

Bottom-line being: We should either:
- Include a way to create the frontends and backends in the stable-ish API (and add IR etc. to stable-ish), or
- Remove everything except OutputMode from stable-ish.

> The only thing that comes to mind would be to introduce StandardLinker.clearable(config: StandardLinker.Config): ClearableLinker, which would wrap a StandardLinker(config) inside a new ClearableLinker(_, config.batchMode).

Yeah, that sounds reasonable. We can then make `ClearableLinker` abstract and move the implementation to `standard.WrappingClearableLinker` (or something the like)?

About `sealed`:
- I'm OK with keeping sealed for the types that only have private cases.
- For `ModuleKind`, I thought we agreed to make a subclass of `ModuleKind` in `standard._` which is sealed (and convertible to from the base by a private abstract member). Then a user (including our code) can chose against which API they want to build.

We should apply a similar strategy for `Linker`: `linker.Linker` should have a `private[linker]` constructor. However, we should have `linker.standard.Linker` (or similar) that extends `linker.Linker` and *is* open for inheritance. Again making it explicit against which API guarantee level something is written against.

sjrd:
At the highest level, the way I see things is to specify things in terms of guarantees against breakage. Concretely,
* Breaking `linker._` or `irio._` requires a major version bump (basically, it's forbidden)
* Breaking `standard._` or `ir._` requires a minor version bump, and should be avoided if possible
* Breaking something else (the unstable parts) can happen in a patch version, although it should be avoided if it is easy to do so (deprecated forwarders, for example)
* `private[linker]` (or stronger) cannot be accessed anyway, so any breakage is allowed any time

The idea is that `standard._` and `ir._` should (eventually) be enough for someone who wants to write a linker plugin. *Currently*, it requires to write a "complete" front-end or back-end, but that *can* actually be practical if the front-end wraps an existing front-end, and just preprocesses or postprocesses the input/output. It is also enough to work around the disappearance of `linkUnit` in some very special use cases for it, for example, a plugin that reports a graph of dependencies in a `LinkedUnit`.

Something built on top of `standard._` and `ir._` would have a guarantee of working through patch versions, and some hope not to be broken at each minor version.

The unstable APIs are for things that can reasonably be used by "experiments", but said experiments should expect to only work for one specific patch release.

> Remove everything except OutputMode from stable-ish.

Or move `OutputMode` up inside `linker._`. That would mean we can never remove any of its alternatives, though.

gzm0:
> Currently, it requires to write a "complete" front-end or back-end, but that can actually be practical if the front-end wraps an existing front-end, and just preprocesses or postprocesses the input/output.

Yes, indeed. However, in the current setup, IIUC, it is impossible to create a standalone frontend or backend with the stable-ish API (or retrieve it from a StandardLinker). So one cannot write a complete linker without depending on the unstable API or really writing a complete linker. I would argue that the second one is unlikely to happen (at least anytime soon), so we should move "writing a new linker" and "creating a standard frontend/backend" in the same compatibility domain (I'm fine with either stable-ish or unstable, although past experience suggests that this is actually something users want to do, so stable-ish?).

> Or move OutputMode up inside linker._. That would mean we can never remove any of its alternatives, though.

I wouldn't do that. The maintenance burden we risk is just too high.

sjrd:
Rebased after the `OutputMode` -> `ESFeatures` refactoring, which somewhat changes the game.

Now `standard._` only contains stuff that I would later consider relevant when building linker plugins. Even though currently, the `standard` API is not enough to build anything meaningful, I think everything it contains would be part of the things necessary to build something meaningful.

Maybe one reasonable thing we could do to make the current `standard` API practical would be to add `standard.StandardFrontend` and `standard.StandardBackend` objects with a `def apply(config: StandardLinker.Config): ...`. That would then be enough to build a custom linker that has a front-end and/or a back-end that wraps a `StandardFrontend`/`StandardBackend` with some pre- or post-processing, which addresses all the use cases we know to exist so far.

linker/shared/src/main/scala/org/scalajs/linker/StandardLinker.scala
```diff
@@ -46,6 +46,9 @@ object StandardLinker {
         LinkerBackendImpl(backendConfig))
   }
 
+  def clearable(config: Config): ClearableLinker =
+    new ClearableLinker(() => apply(config), config.batchMode)
+
```
gzm0: In that case, shouldn't we make the constructor of `ClearableLinker` private[linker]? IIUC the idea here was that they can share a config.

sjrd: If we do that, there's no way to create a `ClearableLinker` wrapping a non-standard linker, is there?

gzm0: Ah, indeed, we should allow this. Although wondering if it should be part of the super stable API :S

Maybe we should move the implementation of `ClearableLinker` to `standard._` as well and just have an abstract class in `linker`?

sjrd: `ClearableLinker` only relies on a `() => Linker` function to create its linker, so only super-stable stuff. If we move it to `standard._`, and some API receives a `() => Linker`, it could not make a `ClearableLinker` out of it without relying on non-super-stable API.

For example, an alternative build plugin could be configured with a `() => Linker` rather than directly with a `ClearableLinker`. That build plugin might want to wrap that `() => Linker` inside a `ClearableLinker`, but to do that it would have to rely on `standard._` (currently sbt-scalajs only uses `linker._`).

Maybe we could make the constructor `private` and add an `apply` method in `ClearableLinker`'s companion object instead of exposing the constructor. And even make `ClearableLinker` a `trait`, that simply inherits from `Linker` and adds an abstract `def clear(): Unit`. The companion's `apply` would return an instance of a private class implementing that trait.

sjrd:
Rebased and updated with the `ClearableLinker.apply` and `StandardLinker{Frontend,Backend}.apply` I was suggesting above.

PR #3286, opened 2018-02-07
Alternative implementation of `Long`s on top of ES `BigInt`s.

sjrd:
This commit introduces a new way to emit `Long`s in the emitter, using the upcoming ECMAScript `BigInt`s. Although they are arbitrary precision integers, rather than 64-bits, they are designed to allow implementations of 64-bit integers on top of them.

The alternative implementation can be selected with the Java system property `-Dorg.scalajs.linker.backend.emitter.usebigintforlongs=true`, to be passed to sbt (or any other process using the linker). It is not exposed as a more stable API, because `BigInt`s are still very experimental, and not entirely specified yet, so this feature is only meant for expert experimentation.

As of this writing, only a very recent V8 can run the produced .js files, and only with `--harmony-bigint`. Therefore, the present commit does not any automated tests to the CI, which makes this feature even more experimental. The implementation can be locally tested with https://github.com/sjrd/scala-js-env-d8, using a locally built d8 (the V8 debug shell).

The general encoding is straightforward: `Long`s values are encoded as `bigint`s whose value fits in the signed 64-bit integer range. The operations translate as follows:

* `a + b` -> `BigInt.asIntN(64, a + b)` (similar for most operations)
* `a << b` -> `BigInt.asIntN(64, a << BigInt(b & 63))` (similar for `>>`)
* `a >>> b` -> `BigInt.asIntN(64, BigInt.asUintN(64, a) >> BigInt(b & 63))` (because `BigInt` does not have logical shift right)


sjrd:
Usually, I wouldn't put this kind of stuff into master, but in this case, it would be nice to make sure that it doesn't bit-rot. Moreover, the `BigInt` proposal is already at Stage 3, which is close to finalization, and it is important that we be able to provide concrete feedback to the TC39 and to VM implementers about this feature, as it can potentially impact Scala.js in critical ways.

To test this, one needs to
* Build a local copy of d8, as explained at https://github.com/v8/v8/wiki/Building-from-Source
* Clone https://github.com/sjrd/scala-js-env-d8 somewhere
* Run with:

```
$ sbt -Dorg.scalajs.linker.backend.emitter.usebigintforlongs=true -Dbe.doeraene.scalajsd8env.d8executable=/path/to/v8/out.gn/x64.release/d8
...
> set jsEnv in testSuite := new be.doeraene.scalajsd8env.D8JSEnv(be.doeraene.scalajsd8env.D8JSEnv.Config().withArgs(List("--harmony-bigint")))
> testSuiteTest
```
(In my experience, putting `d8` on the `PATH` does not work. It is necessary too use the full path.)

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2558,6 +2668,27 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
       genCallHelper("fround", arg)
     }
 
+    private def wrapBigInt32(tree: js.Tree)(implicit pos: Position): js.Tree =
+      wrapBigIntN(32, tree)
+
+    private def wrapBigInt64(tree: js.Tree)(implicit pos: Position): js.Tree =
+      wrapBigIntN(64, tree)
```
gzm0: Are these two helpers really worth it?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2558,6 +2668,27 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
+    }
+
+    private def wrapBigIntU64(tree: js.Tree)(implicit pos: Position): js.Tree =
+      wrapBigIntUN(64, tree)
+
+    private def wrapBigIntUN(n: Int, tree: js.Tree)(
+        implicit pos: Position): js.Tree = {
+      js.Apply(genIdentBracketSelect(genGlobalVarRef("BigInt"), "asUintN"),
+          List(js.IntLiteral(n), tree))
+    }
```
gzm0: Same here, do we need both?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/InternalOptions.scala
```diff
@@ -10,19 +10,24 @@
...
+      optimizeBracketSelects: Boolean = this.optimizeBracketSelects
+  ): InternalOptions = {
     new InternalOptions(
         optimizeBracketSelects)
   }
 }
 
 private[emitter] object InternalOptions {
+  val UseBigIntForLongs =
+    System.getProperty("org.scalajs.linker.backend.emitter.usebigintforlongs", "false") == "true"
```
gzm0: :((( very sad face.

Could we make a `scalajs` private `OutputMode` instead?

By the way I think this is an excellent test case for the linker API design: If we need to resort to system properties (= global static variables) for this kind of thing, we maybe need to think some more.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4371,6 +4363,12 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
 
 private[optimizer] object OptimizerCore {
 
+  /* TODO If this option ever becomes "public", we need a better way than to
+   * peek into the internals of the emitter.
+   */
+  private val useRuntimeLong =
+    System.getProperty("org.scalajs.linker.backend.emitter.usebigintforlongs", "false") != "true"
```
gzm0: Also very sad face.

project/Build.scala
```diff
@@ -1346,9 +1348,11 @@ object Build {
             }
 
           case _ =>
-            throw new AssertionError(
+            s.log.warn(
                 s"Unknown JSEnv of class ${env.getClass.getName}: " +
-                "don't know what tags to specify for the test suite")
+                "don't know what tags to specify for the test suite, " +
+                "so I will assume that TypedArrays are supported")
+            Seq("unknown-jsenv", "typedarray")
```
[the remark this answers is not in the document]
sjrd: Typed Arrays are specified in ECMAScript 2015. We assume that unknown JS envs are "new", for some definition of "new", and therefore support Typed Arrays.

Source map support is very ad hoc and cannot typically be assumed to work.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
sjrd: Mostly to avoid typos such as 62 or 34. 32 and 64 have a pretty specific meaning. But I can't remove them if you prefer.

sjrd:
> However, I think we should not relax constraints on keeping interfaces clean. So I'm really not a fan of punching holes like the system properties do here :S

Maybe we should turn `OutputMode` into a different kind of data structure. One with options, such as `Semantics`. It could look like the following:
```scala
final class OutputMode private (
    val useECMAScript2015: Boolean,
    val allowUseBigIntsForLongs: Boolean
) {
  def withXYZ // as usual
}
```
There would be standard options and experimental options:
* Standard options start with `use`, and guarantee that the given ES feature will be used. For example, `useECMAScript2015` guarantees that `class`es are used instead of `function`s (which does have some visible semantic implications).
* Experimental options start with `allowUse`: they *allow* the linker to use the given feature, but it is not guaranteed. In particular, if support for the experimental feature is dropped, we can deprecate the field and stop using in the back-end, without violating that contract.

When an experimental option is promoted to a standard option, we add the `useX` variant of `allowUseX` and deprecate the latter.

With this scheme, `OutputMode` can be moved up to `linker._` (as opposed to `standard._`), since we will never need to break its API/contract.

gzm0:
I like your idea about `OutputMode`. It solves two things:

- extensibility
- backwards compatibility

What is a bit less nice that we'll likely end up with mutually exclusive settings at some point (we'll go from `allowECMAScript2016` to `useECMAScript2016` at some point). But I think we can live with that (especially since validation for it is easy to implement).

sjrd:
Rebased on master and updated with the `OutputMode` thing.

linker/shared/src/main/scala/org/scalajs/linker/OutputMode.scala
```diff
@@ -0,0 +1,99 @@
...
+  private val HashSeed =
+    scala.util.hashing.MurmurHash3.stringHash(classOf[OutputMode].getName)
+
+  /** Default configuration of the output mode.
+   *
+   *  - `useECMAScript2015`: false
+   */
+  val Defaults: OutputMode = new OutputMode()
+
+  @deprecated("Use Defaults instead.", "1.0.0")
```
gzm0: This should be deprecated in "1.0.0-M4" instead, no? Ideally we can remove this in 1.0.0

linker/shared/src/main/scala/org/scalajs/linker/OutputMode.scala
```diff
@@ -0,0 +1,99 @@
...
+ *
+ *  - Options whose name is of the form `useX` *force* the linker to use the
+ *    corresponding features, guaranteeing that the specific semantics that
+ *    they provide will be used.
+ *  - Options whose name is of the form `allowX` *allow* the linker to use the
+ *    corresponding features if it supports them. Support for such options can
+ *    be dropped in any subsequent version of the linker, including patch
+ *    versions.
+ */
+final class OutputMode private (
```
gzm0: Should we still call this `OutputMode`? Maybe `OutputConfig` or `OutputOptions` would be more appropriate... Not sure though :S

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
gzm0: It's fine with me to leave them.

linker/shared/src/main/scala/org/scalajs/linker/OutputMode.scala
sjrd: If we want to remove in 1.0.0, there needs to be a replacement available in 0.6.x. We could kind of do it. For `Default`/`Defaults` it is easy. For the `withX` and `x` methods it's also possible, for example:
```scala
sealed abstract class OutputMode {
  def useECMAScript2015: Boolean = this match {
    case ECMAScript51Isolated => false
    case ECMAScript6           => true
  }

  def withUseECMAScript2015(useECMAScript2015: Boolean): OutputMode =
    if (useECMAScript2015) ECMAScript6
    else ECMAScript51Isolated
}
```
Should we do that in the 0.6.x branch?

linker/shared/src/main/scala/org/scalajs/linker/OutputMode.scala
sjrd: Same problem with the 0.6.x branch. If we rename it, we'll need a forward-compatibility alias in the 0.6.x branch.

If we rename it, I would drop the "output" terminology altogether. Maybe something like `ESFeatures`, since it seems it's going in that direction?

gzm0: I like `ESFeatures`, yes.

linker/shared/src/main/scala/org/scalajs/linker/OutputMode.scala
gzm0: If you don't mind. I think it is worth going through that extra little bit to get the legacy out of 1.0.

sjrd: ➡️ #3305

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -252,12 +252,16 @@ final class Emitter private (config: CommonPhaseConfig,
   private def emitInitializeL0(): js.Tree = {
     implicit val pos = Position.NoPosition
 
-    // $L0 = new RuntimeLong(0, 0)
-    js.Assign(
-        jsGen.envField("L0"),
-        js.New(jsGen.encodeClassVar(LongImpl.RuntimeLongClass),
-            List(js.IntLiteral(0), js.IntLiteral(0)))
-    )
+    if (jsGen.useBigIntForLongs) {
```
gzm0: Have you considered putting this condition inside `emitInternal`? Seems cleaner to me.

sjrd: Done. I also added a first commit bumping the timeouts on the CI, because I've noticed several spurious timeouts on the master branch recently.

PR #3291, opened 2018-03-07
Test

sjrd:


gzm0:
Hum: Do you think we can get rid of matrix.xml? The main motivation behind having a separate matrix.xml was to have the CI configuration versioned. I don't think that with `Jenkinsfile` being versioned there is any benefit in having a separate configuration anymore (as opposed to "hard code" what we want to test).

sjrd:
Eventually we'll be able to get rid of `matrix.xml`, but as a path of least resistance, I am first trying to have `Jenkinsfile` load `matrix.xml` and generate the `node {}` invocations from there.

gzm0:
Ok, fair enough. If you think that's easier, that's fine with me.

PR #3292, opened 2018-03-07
Migrate to the Pipeline feature of Jenkins 2.

sjrd:
We replace our makeshift `matrix.xml` + Groovy scripts inside the Jenkins config by a single `Jenkinsfile` script. It defines both the matrix, and how to use to create a pipeline of jobs.

This `Jenkinsfile` is intended to be used with a Multibranch Pipeline (https://jenkins.io/doc/book/pipeline/multibranch/). It contains the logic to run a subset of the configurations for PRs (and push'es to branches), compared to the nightly builds.

Since it is unclear how to distinguish "nightly" from "weekly" builds using this system, we take the opportunity to go back to building both the former 'nightly' and 'weekly' matrices as one 'full' matrix. We also rename the 'pr' matrix into 'quick'. Lately, weekly builds have been faster nightly builds anyway.

Using a groovy script instead of XML to define the matrix allows to do it programmatically, avoiding some repetitions and improving readability.


sjrd:
Can be reviewed at this point, but not merged because the real thing should be based on 0.6.x, not on master.

sjrd:
@gzm0 Should be ready now.

Jenkinsfile
```diff
@@ -283,534 +365,151 @@
...
+  "2.12.1",
+  "2.12.2",
+  "2.12.3"
+]
+
+// The 'quick' matrix
+def quickMatrix = []
+mainScalaVersions.each { scalaVersion ->
+  allJavaVersions.each { javaVersion ->
+    quickMatrix.add([task: "main", scala: scalaVersion, java: javaVersion])
```
gzm0: Have you considered making function calls out of the tasks rather than a string -> string map? We could use [named arguments](http://docs.groovy-lang.org/latest/html/documentation/#_named_arguments) to pass in the params. I believe this will allow us to reduce duplication going forward (e.g. partest optimization levels, ES versions, etc.)

If you rather want to get something out fast, I'm OK with it, but refactoring this across two branches might be painful going forward.

Jenkinsfile
```diff
@@ -1,19 +1,101 @@
...
+// Check whether the job was started by a timer
+// See https://hopstorawpointers.blogspot.ch/2016/10/performing-nightly-build-steps-with.html
+@NonCPS
+def isJobStartedByTimer() {
+  def startedByTimer = false
+  def buildCauses = currentBuild.rawBuild.getCauses()
+  for (buildCause in buildCauses) {
+    if (buildCause != null) {
+      def causeDescription = buildCause.getShortDescription()
+      echo "shortDescription: ${causeDescription}"
```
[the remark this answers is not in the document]
sjrd: Yes, it's useful as a debugging message if anything doesn't go as expected (e.g., an update of Jenkins that would change that string ..., since it's not specified).

Jenkinsfile
sjrd: I have considered it, but elected for this variant which deviates less from the previous matrix specification.

I can change it to functions if you prefer.

gzm0: Meh, let's leave it so we can move forward. It's probably worth to schedule a rework of this once the 0.6.x branch is gone anyways.

PR #3302, opened 2018-03-16
Fix #3295: Do not set __ScalaJSEnv.javaSystemProperties in the sbt plugin.

### LINKED ISSUE #3295: Proposal: Make __ScalaJSEnv.javaSystemProperties the responsibility of the JSEnv
Currently we fully manage Java System Property injection in the [sbt plugin](https://github.com/scala-js/scala-js/blob/097d1a6d785d8c0c6451c50dd686305991bc2aa5/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L280-L299). As a result, we need to understand details about the underlying module structure.

Further, note that many `JSEnv`s further inject information into `__ScalaJSEnv` (source mapper, exit function). I propose that we make `__ScalaJSEnv` generation and injection fully the responsibility of a `JSEnv` by adding `javaSystemProperties: Map[String, String]` (`additionalJavaSystemProperties`?) to the `RunConfig` (after #3255).

As a result
- `__ScalaJSEnv` becomes the sole responsibility of the `JSEnv` and hence can be better isolated.
- The sbt plugin does not need to know about module loading details anymore.
- `JSEnv`s can easily add information about themselves to `javaSystemProperties` if they so desire.

This, together with #3294 would allow us to remove `jsExecutionFiles` and call it `jsInput`. As a result, our core supporting infrastructure would be completely agnostic of the JS module type.

sjrd:
Instead, we set it up in our own build.

That feature was initially included in the sbt plugin because our build needed it, but it has always been a bit hacky. In particular, it requires the sbt plugin to know too much about how JS files are executed, notably whether they are scripts or modules. Moreover, that feature was the only reason left that the sbt plugin needed to know about `__ScalaJSEnv` at all.


gzm0:
Actually: have you considered getting rid of the parsing and only set `javaOptions` in the JVM tests?

sjrd:
I have, but then [this one](https://github.com/sjrd/scala-js/blob/05aac3ffbd0bd54a712412662c5a13d7674ebd1d/project/Build.scala#L685-L688) came up in `commonLinkerSettings`, and I realized it would be more work to introduce a generalized `javaSysProps` in our build, that would work for all JVM projects in addition to JS projects. It felt easier to keep the parsing, which at least does not introduce new keys.

gzm0:
Ok, so let' leave it like this, then.

Issue #2675, opened 2016-12-05
Minimize the number of IIFEs generated by Desugar for Closure IR nodes


Atry:
Was:  Scala.js with Scala 2.12.0 creates too many unnecessary IIFE

I created a web framework [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala), which generates deeply nested blocks for XHTML literal.

These kind of blocks will be converted to IIFE for ECMAScript 5 like this:

``` javascript
                              x(function(a, b, c) {
                                return function(e) {
                                  var f = B();
                                  return S((new G).d(P(C(new D, function(a) {
                                    return function() {
                                      return (new G).d(P(C(new D, function() {
                                        return function() {
                                          return (new F).g(U(Q(), "\n        "));
                                        };
                                      }(a))));
                                    };
                                  }(a, f)))), x(function(a, b, c, e) {
                                    return function(b) {
                                      return E((new G).d(P(C(new D, function(a) {
                                        return function() {
                                          return x(function() {
                                            return function(a) {
                                              return a;
                                            };
                                          }(a));
                                        };
                                      }(a)))), x(function(a, b, c, e) {
                                        return function(a) {
                                          var f = (new F).g((new J).n([b, c, e]));
                                          return (new Eb).Y((new Fb).Y(f, a), x(function() {
                                            return function(a) {
                                              return a;
                                            };
                                          }(f)));
                                        };
                                      }(a, c, e, b)));
                                    };
                                  }(a, b, c, e)));
                                };
                              }(a, b, c))
```
The entire JS file can be found at https://github.com/ThoughtWorksInc/todo/blob/3f173cd/js/js-opt.js

These IIFEs are very slow. In [some benchmarks](https://github.com/Atry/js-framework-benchmark/tree/binding.scala), even the output ECMAScript 2015 code (which creates `const` instead of IIFE) with fast optimizer runs two times faster than ECMAScript 5 code with full optimizer!

I expect that Closure Compiler would inline IIFE, or enabling Closure Compiler for ECMAScript 2015.

sjrd:
I don't really understand what you want Scala.js to do here? Could you make that more precise? We don't develop Closure, so changing what it does is not within our ability.

sjrd:
It's also not clear to me what "the ECMAScript 2015" code is. Generated how, by what, from what?

Atry:
[...]

PR #3308, opened 2018-03-20
Finish the refactoring of `OutputMode` into `ESFeatures`.

sjrd:
Instead of having a sealed abstract class with case objects, that wasn't actually sealed, and was prone to breakages, we have a principled options bag for individual features of ECMAScript that the linker is mandated and/or allowed to use.


Atry:
Hi, @sjrd , did this PR fix #2675 ?

I found a related conditional expression here:

https://github.com/scala-js/scala-js/blob/df91ee0a33a2276c7a0538fd9df6ae313d96f3c5/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala#L1188

Atry:
Will this PR be back-ported to 0.6.x?

sjrd:
This PR did not change any feature. As its name very clearly says, it's only a refactoring.

The 0.6.x "backport" has already landed in 0.6.23. It was https://github.com/scala-js/scala-js/pull/3305.

Atry:
Ah, I see.

The IIFE is introduced when converting closures, which is untouched yet, not `envFieldDef`.

https://github.com/scala-js/scala-js/blob/df91ee0a33a2276c7a0538fd9df6ae313d96f3c5/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala#L2600

PR #3311, opened 2018-03-24
Blacklist LongTest.scala in FullOpt.

sjrd:
`LongTest.scala` uses optimizer-based generative programming to generate a *lot* of tests. Worse, the way it optimizes `Long`s turns that huge amount of test code into an *insane* amount of .js code within methods. Closure then completely loses it with that much code and all the intra-method optimizations it can do about it.

Blacklisting this test when compiling for `FullOpt` dramatically improves the time it takes to fully optimize the test suite (between 4x and 5x on my machine).

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1635,6 +1635,20 @@ object Build {
         Seq(outFile)
       }.taskValue,
 
+      /* Blacklist LongTest.scala in FullOpt, because it generates so much
+       * code, through optimizer-based generative programming, that Closure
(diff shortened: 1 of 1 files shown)


sjrd:
I have reactivated lampscalatst5 for this test. I believe this will solve our timeout issues for good.

sjrd:
Yup, passed on the first try, and no job had to use its `retry` :)

gzm0:
Filed #3312 to add some tests.

PR #3315, opened 2018-03-28
Upgrade to Scala 2.12.5 and 2.13.0-M3.

sjrd:


sjrd:
Ready for review.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/math/BigIntegerConstructorsTest.scala
```diff
@@ -121,7 +121,7 @@ class BigIntegerConstructorsTest {
   }
 
   @Test def testConstructorBytesZero(): Unit = {
-    val aBytes = Array[Byte](0, 0, 0, -0, +0, 0, -0)
+    val aBytes = Array[Byte](0, 0, 0, -0, 0, 0, -0)
```
gzm0: Are we going to add these back when M4 comes?

sjrd: Probably not. They were not consistently used in here. It seems like they were `+0` instead of `0` by "chance".

We could add a dedicated test somewhere else, but that would be redundant with [the partest](https://github.com/scala/scala/pull/6416/files#diff-2ca2635e3d3c286af0041a1b6e77bd0a) anyway.

PR #3317, opened 2018-03-30
Fix #3281: Backport the new `genConversion` from 1.x.

### LINKED ISSUE #3281: [0.6.x] Invalid IR (missing conversion) for adapted Int-to-Long in match with 2.13.0-M3
OK, this is a really, really weird one. Discovered when trying to [upgrade scala-xml to 2.13.0-M3](https://github.com/scala/scala-xml/pull/186). It only happens on 0.6.x, using Scala 2.13.0-M3.

Here is a minimal reproduction:
```scala
object HelloWorld {
    val l: Any = 0 :: Nil
    overloaded(l match {
      case x :: xs => 5
    })
  }

  def overloaded(x: Long): Unit =
    assert(x.equals(5L))

  def overloaded(x: Any): Unit =
    assert(false, "test(Any)")
}
```
The ingredients are:
* An overloaded method, with one overload taking a `Long`
* A call to the overloaded method, whose argument is a `match`
* The rhs of the `case` is an `Int`

Symptom: IR checking error (or weird link-time or run-time misbehavior if the IR checker is not enabled, as is the case in scala-xml):
```
[error] HelloWorld.scala(15:14:If): LongType expected but IntType found for tree of type org.scalajs.core.ir.Trees$IntLiteral
```
Sure enough, the IR is bogus:
```scala
> helloworld/scalajsp helloworld/HelloWorld$.sjsir
module class Lhelloworld_HelloWorld$ extends O {
  def main__AT__V(args: T[]) {
    val l: any = {
      val rassoc$1: int = 0;
      mod:sci_Nil$.$$colon$colon__O__sci_List(rassoc$1)
    };
    this.overloaded__J__V({
      val x1: any = l;
      if (x1.isInstanceOf[sci_$colon$colon]) {
        5 // <== This 5 is missing a conversion to Long
      } else {
        throw new s_MatchError().init___O(x1)
      }
    })
  }
  def overloaded__J__V(x: long) {
    mod:s_Predef$.assert__Z__V(x.equals__O__Z(5L))
  }
  def overloaded__O__V(x: any) {
    mod:s_Predef$.assert__Z__F0__V(false, new sjsr_AnonFunction0().init___sjs_js_Function0((lambda<$this: Lhelloworld_He
lloWorld$ = this>() = {
      $this.$$anonfun$overloaded$1__p1__T()
    })))
  }
  @hints(1) def $$anonfun$overloaded$1__p1__T(): T = {
    "test(Any)"
  }
  def init___() {
    this.O::init___();
    mod:Lhelloworld_HelloWorld$<-this
  }
}
```
`-Xprint:mixin` shows something funny:
```scala
package helloworld {
  object HelloWorld extends Object {
      val l: Object = {
        <synthetic> <artifact> val rassoc$1: Int = 0;
        scala.collection.immutable.Nil.::(scala.Int.box(rassoc$1))
      };
      HelloWorld.this.overloaded({
        case <synthetic> val x1: Object = l;
        case5(){
          if (x1.$isInstanceOf[scala.collection.immutable.::]())
            matchEnd4(5.$asInstanceOf[Long]()) // <== Note the $asInstanceOf[Long] here
          else
            case6()
        };
        case6(){
          matchEnd4(throw new MatchError(x1))
        };
        matchEnd4(x: Long){
          x
        }
      })
    };
    def overloaded(x: Long): Unit = scala.Predef.assert(scala.Long.box(x).equals(scala.Long.box(5L)));
    def overloaded(x: Object): Unit = scala.Predef.assert(false, {
      (() => HelloWorld.this.$anonfun$overloaded$1())
    });
    final <artifact> private[this] def $anonfun$overloaded$1(): String = "test(Any)";
    def <init>(): helloworld.HelloWorld.type = {
      HelloWorld.super.<init>();
      ()
    }
  }
}
```
It converts the `5` into a `Long` using `.$asInstanceOf[scala.Long]`.

By comparison, with 2.13.0-M2, the `5` (or rather the result of the `match`) is converted to `Long` using `.toLong()`:
```scala
package helloworld {
  object HelloWorld extends Object {
      val l: Object = {
        <synthetic> <artifact> val x$1: Int = 0;
        scala.collection.immutable.Nil.::(scala.Int.box(x$1))
      };
      HelloWorld.this.overloaded({
  case <synthetic> val x1: Object = l;
  case5(){
    if (x1.$isInstanceOf[scala.collection.immutable.::]())
      matchEnd4(5)
    else
      case6()
  };
  case6(){
    matchEnd4(throw new MatchError(x1))
  };
  matchEnd4(x: Int){
    x
  }
}.toLong()) // <== .toLong() here
    };
    def overloaded(x: Long): Unit = scala.Predef.assert(scala.Long.box(x).equals(scala.Long.box(5L)));
    def overloaded(x: Object): Unit = scala.Predef.assert(false, {
      (() => HelloWorld.this.$anonfun$overloaded$1())
    });
    final <artifact> private[this] def $anonfun$overloaded$1(): String = "test(Any)";
    def <init>(): helloworld.HelloWorld.type = {
      HelloWorld.super.<init>();
      ()
    }
  }
}
```
and, accordingly, the IR has a `(long)` conversion operator, which the 2.13.0-M3 version is missing.

Now, `$asInstanceOf[prim]` ends up in [genConversion](https://github.com/scala-js/scala-js/blob/43a7c39307ea33870c6750a84c31e1bd9014d177/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L2690-L2713), an obscure method, which for a long time I could not even find a reproducible source code that calls it. As you can see, it is definitely missing a case for `Int` to `Long` conversion, which means it just forwards the value unadapted.

In Scala.js 1.x, `genConversion` was completely rewritten in https://github.com/scala-js/scala-js/commit/9890ee769ee55adc0ef0514760d08e3fe3a3c2a0 to become [this new version](https://github.com/scala-js/scala-js/blob/ec5b328330276b9feb20cccadd75e19d27e887d3/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala#L2816-L2868), which does handle all sorts of primitive conversions. This explains why Scala.js 1.x is not affected. In the same commit, I also [added a comment](https://github.com/scala-js/scala-js/commit/9890ee769ee55adc0ef0514760d08e3fe3a3c2a0#diff-c9a195a739d51194e79f0655bc93bd61R2352) about one case I found where this thing was generated.

Now, it seems that Scala 2.13.0-M3 generates new occurrences of these `$asInstanceOf[prim]`, where previous versions of the compiler used `.toLong`. From the minimization, it can at least happen if a `match` needs to be *adapted* from `Int` to `Long`. The overload is necessary to trigger this: it prevents the typechecker from typechecking the `match` with an *expected* type of `Long`, while at the same time needing to *adapt* the `Int` it found to the matching overload which is `Long`.

I tried with a simple `Int` or with an `if`, and could not reproduce the issue. It seems limited to `match`.

The fix would of course be to extend `genConversion` in 0.6.x to support all possible conversions, like the one in 1.x.

sjrd:
Starting with Scala 2.13.0-M3, if the typer needs to *adapt* an `Int` to a `Long` inside a `match`, typically after overload resolution has taken place, we end up with a tree of the form `(x: Int).$asInstanceOf[Long]` in the back-end. This kind of node is handled by the `genConversion` method, which was previously buggy for the `Int`-to-`Long` conversion (among others).

This commit backports the implementation of `genConversion` from master, which was rewritten in
9890ee7. We needed to adapt the code to rely on `TypeKind`s instead of `jstpe.Type`s, because in 0.6.x the latter do not know the difference between the various sub-`Int` types.

Will need to rebased on top of #3315 when it is merged. Only the last commit belongs to this PR.

The commit is not marked `[no-master]`, because the test should be merged forward. The conflict resolution in `GenJSCode.scala` will be to drop the code coming from 0.6.x and keep the code already in master.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2687,29 +2687,56 @@ abstract class GenJSCode extends plugins.PluginComponent
           arguments)(resultType)
     }
 
-    /** Gen JS code for a conversion between primitive value types */
+    /* This method corresponds to the method of the same name in
+     * BCodeBodyBuilder of the JVM back-end. It ends up calling the method
+     * BCodeIdiomatic.emitT2T, whose logic we replicate here.
+     */
     def genConversion(from: TypeKind, to: TypeKind, value: js.Tree)(
         implicit pos: Position): js.Tree = {
-      def int0 = js.IntLiteral(0)
-      def int1 = js.IntLiteral(1)
-      def long0 = js.LongLiteral(0L)
-      def long1 = js.LongLiteral(1L)
-      def float0 = js.FloatLiteral(0.0f)
-      def float1 = js.FloatLiteral(1.0f)
+      import js.UnaryOp._
 
-      // scalastyle:off disallow.space.before.token
-      (from, to) match {
-        case (INT(_),   BOOL) => js.BinaryOp(js.BinaryOp.Num_!=,  value, int0)
-        case (LONG,     BOOL) => js.BinaryOp(js.BinaryOp.Long_!=, value, long0)
-        case (FLOAT(_), BOOL) => js.BinaryOp(js.BinaryOp.Num_!=,  value, float0)
+      if (from == to || from == NOTHING) {
+        value
+      } else if (from == BOOL || to == BOOL) {
+        throw new AssertionError(s"Invalid genConversion from $from to $to")
+      } else {
+        def intValue = (from: @unchecked) match {
+          case INT(_)   => value
+          case LONG     => js.UnaryOp(LongToInt, value)
+          case FLOAT(_) => js.UnaryOp(DoubleToInt, value)
+        }
 
-        case (BOOL, INT(_))   => js.If(value, int1,   int0  )(jstpe.IntType)
-        case (BOOL, LONG)     => js.If(value, long1,  long0 )(jstpe.LongType)
-        case (BOOL, FLOAT(_)) => js.If(value, float1, float0)(jstpe.FloatType)
+        def doubleValue = from match {
+          case FLOAT(_) | INT(_) => value
+          case LONG              => js.UnaryOp(LongToDouble, value)
+        }
 
-        case _ => value
+        (to: @unchecked) match {
+          case CharKind =>
+            js.BinaryOp(js.BinaryOp.Int_&, intValue, js.IntLiteral(0xffff))
+          case ByteKind =>
+            js.BinaryOp(js.BinaryOp.Int_>>,
+                js.BinaryOp(js.BinaryOp.Int_<<, intValue, js.IntLiteral(24)),
+                js.IntLiteral(24))
+          case ShortKind =>
+            js.BinaryOp(js.BinaryOp.Int_>>,
+                js.BinaryOp(js.BinaryOp.Int_<<, intValue, js.IntLiteral(24)),
+                js.IntLiteral(24))
+          case IntKind =>
+            intValue
+          case LongKind =>
+            from match {
+              case FLOAT(_) =>
+                js.UnaryOp(DoubleToLong, doubleValue)
+              case _ =>
+                js.UnaryOp(IntToLong, intValue)
+            }
+          case FloatKind =>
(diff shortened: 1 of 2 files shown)


sjrd:
Rebased.

PR #3319, opened 2018-03-30
Cleanup various details for io

gzm0:


io/js/src/main/scala/org/scalajs/io/NodeVirtualFiles.scala
```diff
@@ -61,3 +61,62 @@ private[io] object NodeFS extends js.Object {
...
+    try {
+      /* Build a Uint8Array with the content of this jar file.
+       * We know that in practice, NodeVirtualBinaryFile#inputStream returns
+       * an ArrayBufferInputStream, so we just fetch its internal ArrayBuffer
+       * rather than copying.
+       *
+       * Since we have NodeVirtualBinaryFile under our control, in the same
+       * repository, we can make this assumption. Should we change
+       * NodeVirtualBinaryFile, this test will immediately fail, and we can
+       * adapt it.
```
sjrd: This comment is outdated, now that you've put that class in the public API. At least we should say "*the bootstrap test* will immediately fail".

gzm0: Good point. I'm going to hold off on this, since we probably should not put this in the public API until io is async: JSZip 3.0 does not support synchronous operation anymore.

PR #3321, opened 2018-03-30
Cleanups for VirtualFiles

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/JSBuilders.scala
```diff
@@ -9,46 +9,35 @@
...
-trait JSTreeBuilder {
+trait JSBuilder {
   /** Add a JavaScript tree representing a statement.
    *  The tree must be a valid JavaScript tree (typically obtained by
    *  desugaring a full-fledged IR tree).
    */
   def addJSTree(tree: Trees.Tree): Unit
 
+  /** Add a chunk of JavaScript code. */
+  def addTopLevelStatement(originalLocation: URI, code: String): Unit
```
sjrd: The "TopLevel" mention is misleading, because `addJSTree` also emits a *top-level* tree, but the method name does not contain `TopLevel`. They should either both have it or not have it.

linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/JSBuilders.scala
```diff
@@ -9,46 +9,35 @@
...
-
-  def addFile(file: VirtualJSFile): Unit =
-    addPartsOfFile(file)(!_.startsWith("//# sourceMappingURL="))
-
-  def addPartsOfFile(file: VirtualJSFile)(selector: String => Boolean): Unit = {
-    for (line <- file.readLines() if selector(line))
-      addLine(line)
-  }
+  def addTopLevelStatement(originalLocation: URI, code: String): Unit =
+    outputWriter.write(code)
```
sjrd: If `code` does not end with a `\n`, this method should add an addition `\n`. Otherwise, the override in `JSFileBuilderWithSourceMapWriter` will not produce a valid source map, since with its `.split('\n')` and `nextLine()` for each line, it only works correctly if the `code` ends with a `\n`.

Also care must be taken to handle the `code.isEmpty` case consistently as well.

IMO adding the `\n` is better than fixing the code in `JSFileBuilderWithSourceMapWriter`, because a "statement" might not end with a trailing `;` and rely on ASI to be correct.

It wouldn't hurt to add comments in both methods that stress that they must be in sync about this.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureModuleBuilder.scala
```diff
@@ -13,26 +13,40 @@ import scala.collection.mutable
...
+    val path = URIUtil.sourceURIToString(relativizeBaseURI, originalLocation)
+    module.add(new CompilerInput(SourceFile.fromCode(path, code)))
+  }
 
-    new ClosureAstBuilder.ScalaJSSourceAst(root)
+  override def complete(): Unit = {
+    flushTrees()
   }
 
+  def result(): JSModule = module
```
sjrd: Wouldn't it be safer to `flushTrees()` before returning `module`?

linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/JSBuilders.scala
```diff
@@ -49,24 +52,24 @@ class JSFileBuilder(val name: String,
...
+  def addLine(line: String): Unit = {
+    outputWriter.write(line)
+    outputWriter.write('\n')
     sourceMapWriter.nextLine()
   }
 
-  override def addTopLevelStatement(originalLocation: URI, code: String): Unit = {
-    super.addTopLevelStatement(originalLocation, code)
+  def addTopLevelStatement(originalLocation: URI, code: String): Unit = {
+    outputWriter.write(code)
```
sjrd: Ah, so with this commit, there is no need for two methods to stay in sync, which is good. But the logic in this method must still send a `\n` to the `outputWriter` if `code` does not have a trailing `\n`.

gzm0:
Updated. I have manually verified with reversi that source maps work in fastOpt. I was unable to verify this in full opt since there is too much inlining.

gzm0:
Just to be sure: Did you see the last commit I added where I remove `RelativeVirtualFile`?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -12,22 +12,12 @@ import java.io._
...
 
-object FileVirtualScalaJSIRFile extends (File => FileVirtualScalaJSIRFile) {
+object FileVirtualScalaJSIRFile {
   import FileVirtualFile._
 
-  def apply(f: File): FileVirtualScalaJSIRFile =
-    new FileVirtualScalaJSIRFile(f)
-
-  def relative(f: File, relPath: String): FileVirtualScalaJSIRFile
-      with VirtualRelativeScalaJSIRFile = {
```
sjrd: Why not keep this method, renamed into `apply`?

gzm0:
Updated.

PR #3322, opened 2018-03-31
Get rid of the `Continue` IR node, as well as labels on loops.

sjrd:
Where we previously used `continue` statements in the IR, as in:
```javascript
lab: while (cond) {
  ...
  if (c)
    continue lab;
  ...
}
```
We now use a labeled block inside the loop, as well as a return off that labeled block, instead:
```javascript
while (cond) {
  lab[void]: {
    ...
    if (c)
      return(lab) undefined;
    ...
  }
}
```
This allows two important simplifications in the IR:

* We do not have the `Continue` IR node anymore.
* `While` and `DoWhile` loops do not have labels anymore.

With those changes, labels are only used by `Labeled` blocks and their corresponding `Return` statements.

To prevent some regressions in the quality of the generated .js code, we recover JS `continue` statements out of the above structure in the `FunctionEmitter`. We also received some *improvements* for free, because the `tailPosLabels` can kick in for those labels.

Full diff of the `testsuite-test-fastopt.js`: https://gist.github.com/sjrd/3d3f427d4b94072d4811b295c4c0a762

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
@@ -190,6 +190,51 @@ import Transients._
  *  `defaultBreakTargets` is property of the *scope*, and is therefore stored
  *  in the environment.
  *
+ *  Finally, we also recover JavaScript `continue` statements out of labeled
+ *  blocks that are immediately nested in the body of `while` loops. When we
+ *  have:
+ *
+ *  {{{
+ *  while (cond) {
+ *    lab: {
+ *      ...
+ *      if (c)
+ *        return(lab) expr;
+ *      ...
+ *    }
+ *  }
+ *  }}}
+ *
+ *  we make the `while` loop "acquire" the label as its own, and instruct the
+ *  environment that, should an instruction want to return to that label, it
+ *  should use `continue` instead of `break`. This produces:
+ *
+ *  {{{
+ *  lab: while (cond) {
+ *    ...
+ *    if (c)
+ *      { expr; continue lab }
+ *    ...
+ *  }
+ *  }}}
+ *
+ *  Note that the result of `expr` is always discarded, in that case, since the
+ *  labeled block was in statement position (by construction).
+ *
+ *  Similarly to the treatment with `defaultBreakTargets`, we keep track of
+ *  `defaultContinueTargets`, the set of labels for which we can simply write
+ *  `continue` instead of `continue lab`. This allows the above code to finally
+ *  become:
+ *
+ *  {{{
+ *  while (cond) {
+ *    ...
+ *    if (c)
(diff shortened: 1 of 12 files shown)


sjrd:
I was working on the formalization of the IR for my thesis, and was bothered by the `continue`s, which I never *truly* understood myself. And then I realized I could get rid of them, and simplify a few things on the way 😄

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -575,54 +620,69 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
               unnest(cond) { (newCond, env0) =>
                 implicit val env = env0
-                js.If(transformExprNoChar(newCond),
-                    transformStat(body, Set.empty)(
-                        env.withDefaultBreakTargets(bodyBreakTargets)),
-                    js.Break())
+                js.If(transformExprNoChar(newCond), newBody, js.Break())
               }
-            }, newLabel)
+            })
```
gzm0: Don't you need to use `optLabel` here? Otherwise how will the while get a label if it needs one?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -575,54 +620,69 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
           }
 
-        case DoWhile(body, cond, label) =>
+        case DoWhile(body, cond) =>
```
gzm0: Why do we not perform the same label optimization for `DoWhile`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
sjrd: Ouch, yes indeed!

I guess all the cases that are generated by Scala source code end up not needing a label ... which is why it wasn't caught by the CI. But in general, we might need one.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
sjrd: Because it's more complicated in a `DoWhile` due to the fact that the condition happens after the block. That means that in the case `!isExpression(cond)`, we cannot really do anything. In the case `isExpression(cond)`, we could. However, no Scala source code produces `do..while`s with this pattern, so the optimization would be dead code. (we get it for `while` loops--`while(true)` loops in particular--mostly because that's how we compile tail-recursive methods.

tl;dr, not worth it ;) I'll add a comment.

sjrd:
Updated.

PR #3326, opened 2018-04-01
Fix #2751: Fix the definition of `js.Date` for invalid dates.

### LINKED ISSUE #2751: js.Date functions can return NaN, so they should be typed as Double

sjrd:
For invalid dates, created from a string that cannot be parsed, the methods of `js.Date` return `NaN`. Therefore, they must be typed as `Double`s. By extension, this covers up for Firefox returning -0 for some valid date as well, instead of +0 like everyone else.

In the process, we fix the behavior of `j.u.Date.parse` for invalid inputs.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Date.scala
+++ b/library/src/main/scala/scala/scalajs/js/Date.scala
@@ -23,6 +23,12 @@ import scala.scalajs.js.annotation._
  * since 1 January, 1970 UTC.
  *
  * MDN
+ *
+ * @note
+ *   `js.Date` objects can represent an *invalid date*, for example, if they
+ *   are constructed from a `String` that cannot be parsed as a date. Most
+ *   methods of such a `js.Date` will return `NaN` (for those returning a
+ *   `Double`) or other invalid values.
  */
 @js.native
 @JSGlobal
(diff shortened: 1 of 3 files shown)


library/src/main/scala/scala/scalajs/js/Date.scala
```diff
@@ -48,141 +48,144 @@ class Date extends js.Object {
    *
    * MDN
    */
-  def getFullYear(): Int = js.native
+  def getFullYear(): Double = js.native
```
gzm0: I think the `js.Date` class deserves a general comment on why all the members here are typed as `Double`. IMO the MDN comment is *very* misleading: "that represents a single moment in time" is *not* the only thing it does (apparently). It can also represent an invalid time. We should document this.

sjrd: Done.

PR #3327, opened 2018-04-02
Fix the way the bootstrap test prepares `testSuite/test:fastOptJS`.

sjrd:
Since b9ec4ee, it is basically invalid to call `testSuite/test:fastOptJS` if `scalaJSStage in testSuite` is `FullOptStage`, and conversely. It didn't cause any hard failure, but was ill-advised since the blacklisting would not be applied appropriately.

The bootstrap test was erroneously doing just that. This commit fixes that test, and adds a check in the test suite settings to make sure no one tries to do that anymore.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -229,17 +229,26 @@ def Tasks = [
         ++$scala $testSuite/test
   ''',
 
+  /* The reason we do `testSuite/test:fastOptJS` is that the linker tests will
+   * need the "discovered tests" from the test suite, which means we need to
+   * fast-optimize the test suite. We explicitly do it beforehand, so that the
+   * memory necessary to do that is not accumulated with the rest of the tests.
+   * Moreover, for the bootstrap tests to be able to call
+   * `testSuite/test:fastOptJS`, `scalaJSStage in testSuite` must be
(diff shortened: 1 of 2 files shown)


gzm0:
QQ: Couldn't we use the optimizer's constant folding to effectively remove the code from the test? (or is that a problem with the optimizer disabled?)
That would allow us to not rely on the sources hack.

sjrd:
We could ... kind of. We could write all those test methods as:
```scala
@Test testMultiply(): Unit = {
  if (LinkingInfo.developmentMode) {
    ... // tests here
  }
}
```
That would kill all the tests at link time in `fullOptJS`. But *also* in `fastOptJS` with `scalaJSLinkerConfig ~= { _.withSemantics(_.optimized) }`.

sjrd:
Note that the changes in the bootstrap test are a good thing regardless. They now ensure that the discovered tests are found using the `FastOpt` stage of the `testSuite`, which makes sense because the `bootstrapTest` will *anyway* fast-optimize the test suite, not full-optimize it. And, in addition, this speeds up the bootstrap test jobs.

gzm0:
Ok. Hopefully using a configuration for the bootstrap tests will improve this :)

Issue #1701, opened 2015-06-05
Add nice title and logo to doc


gzm0:
The Scala.js doc currently displays the name of the open package as title and has no special logo. This is confusing when it is opened in the browser. Adding a logo and a title prefix saying "Scala.js" will help developers recognize the correct tab faster.

antonsarov:
@gzm0 There is a favicon for API docs pages. Or do you have some other pages in mind? Of course the title prefix is still missing...

![scalajs](https://cloud.githubusercontent.com/assets/5653996/8290109/228f4ab0-1921-11e5-819f-31441376ae2b.png)

vendethiel:
A title prefix has the downside of looking a bit worse than a postfix on google, though. "Scala.js" is small enough that it shouldn't matter, though

gzm0:
@antonsarov Hmmm... doesn't show on Chrome Version 41.0.2272.89 (64-bit) on Ubuntu... I assume your screenshot is from firefox...

antonsarov:
@gzm0 Correct, Firefox on Windows

PR #3328, opened 2018-04-03
Add title prefix on scaladoc

exoego:
Addresses #1701 

This PR adds "Scala.js" and version as title on scaladoc pages, so that developes can recognize pages are for what library and version.

image

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1050,7 +1050,12 @@ object Build {
       ) ++ (
           scalaJSExternalCompileSettings
       ) ++ inConfig(Compile)(Seq(
-          scalacOptions in doc ++= Seq("-implicits", "-groups"),
+          scalacOptions in doc ++= Seq(
+              "-implicits",
+              "-groups",
+              "-doc-title", "Scala.js",
+              "-doc-version", scalaJSVersion
+          ),
 
           // Filter doc sources to remove implementation details from doc.
           sources in doc := {


project/Build.scala
```diff
@@ -1107,7 +1107,7 @@ object Build {
 
       scalaJSExternalCompileSettings,
       inConfig(Compile)(Seq(
-          scalacOptions in doc ++= Seq("-implicits", "-groups"),
+          scalacOptions in doc ++= Seq("-implicits", "-groups", "-doc-title", "Scala.js", "-doc-version", s"${scalaJSVersion}"),
```
sjrd: This line is getting long. I suggest you split it in a meaningful way:
```scala
scalacOptions in doc ++= Seq(
    "-implicits",
    "-groups",
    "-doc-title", "Scala.js",
    "-doc-version", scalaJSVersion
),
```
(also the `s"${foo}"` is just `foo` :wink:)

(note also that we use 4 spaces of indentation when we wrap things like that).

exoego:
Oops. I commited as wrong user.

project/Build.scala
exoego: Done in https://github.com/scala-js/scala-js/pull/3328/commits/6ec2db62b22c552e813c2939d40d7530c2e972b3

exoego:
> Also, consider targeting 0.6.x for this change.

Should I create another PR with same changes against 0.6.x?

sjrd:
> Should I create another PR with same changes against 0.6.x?

No, you can change the target branch of this PR (click on Edit next to the PR title), then `rebase -i 0.6.x` your branch locally and then force-push it. No need to have 2 PRs for the same stuff.

exoego:
Target branch changed to `0.6.x`

PR #3329, opened 2018-04-03
2.13.0-M4 support

julienrf:
The goal of this PR is to be ready to release Scala.js for 2.13.0-M4 as soon as the milestone is out.

The main changes caused by 2.13 are related to the usage of collections, the implementation of custom collections, and the way varargs methods are encoded.

- [x] library
- [x] javalib
- [x] scalalib
- [x] compiler
- [x] testSuite
- [x] scalaTestSuite
- [x] linker
- [x] ir
- [x] partest
- [ ] upgrade to latest Scala (a few things have been fixed upstream, and that will impact Scala.js)
- [ ] squash commits


compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -1395,7 +1395,7 @@ abstract class GenJSCode extends plugins.PluginComponent
 
       val ctorToChildren = secondaryCtors.map { ctor =>
         findCtorForwarderCall(ctor.body.get) -> ctor
-      }.groupBy(_._1).mapValues(_.map(_._2)).withDefaultValue(Nil)
+      }.groupBy(_._1).mapValues(_.map(_._2)).toMap.withDefaultValue(Nil)
```
julienrf: `mapValues` explicitly return a `View` in the new collections (instead of a `Map`), hence the explicit call to `toMap` to evaluate it as a `Map`.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4671,8 +4671,7 @@ abstract class GenJSCode extends plugins.PluginComponent
             tryGenRepeatedParamAsJSArray(arg, handleNil = false).fold {
               genExpr(arg)
             } { genArgs =>
-              genNew(WrappedArrayClass, WrappedArray_ctor,
-                  List(js.JSArrayConstr(genArgs)))
+              genWrappedArrayForVarargs(js.JSArrayConstr(genArgs))
```
julienrf: We now wrap vararg parameters into an `ImmutableArray` instead of a `WrappedArray`, because it has to be a `scala.collection.immutable.Seq` in 2.13.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -4840,7 +4839,7 @@ abstract class GenJSCode extends plugins.PluginComponent
           nme.wrapDoubleArray,
           nme.wrapBooleanArray,
           nme.wrapUnitArray,
-          nme.genericWrapArray).map(getMemberMethod(PredefModule, _)).toSet
+          nme.genericWrapArray).map(getMemberMethod(if (isScala213NewCollections) ScalaRunTimeModule else PredefModule, _)).toSet
```
julienrf: The wrapping methods are now defined in `ScalaRunTime` instead of `Predef`

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -5534,6 +5551,12 @@ abstract class GenJSCode extends plugins.PluginComponent
     settings.Xexperimental.value
   }
 
+  // The new collections have been introduced in 2.13.0-M4
+  private[scalajs] lazy val isScala213NewCollections = {
+    val v = scala.util.Properties.versionNumberString
+    v.startsWith("2.13") && v != "2.13.0-M3"
+  }
```
julienrf: We check that we are on a 2.13 release that contains the new collections. I’m not sure if I should exclude the previous milestones as well since they are not mentioned either in the build.

javalib/src/main/scala/java/math/BigDecimal.scala
```diff
@@ -195,7 +195,7 @@ object BigDecimal {
   }
 
   private[math] def newArrayOfPows(len: Int, pow: Int): Array[Long] =
-    new Array[Long](len - 1).scanLeft[Long, Array[Long]](1)((z, e) => z * pow)
+    new Array[Long](len - 1).scanLeft(1L)((z, e) => z * pow)/* temporary, see scala/collection-strawman#544 */.toArray
```
julienrf: `scanLeft` now takes only one type parameter instead of two.

javalib/src/main/scala/java/util/ArrayList.scala
```diff
@@ -66,7 +66,7 @@ class ArrayList[E] private (private[ArrayList] val inner: js.Array[E])
   override def addAll(index: Int, c: Collection[_ <: E]): Boolean = {
     c match {
       case other: ArrayList[_] =>
-        inner.splice(index, 0, other.inner: _*)
+        inner.splice(index, 0, other.inner.toSeq: _*)
```
julienrf: `other.inner` is only a `scala.collection.Seq`, not a `scala.collection.immutable.Seq`, hence the `toSeq` call.

javalib/src/main/scala/java/util/HashSet.scala
```diff
@@ -24,7 +26,7 @@ class HashSet[E] extends AbstractSet[E] with Set[E]
     inner.contains(Box(o.asInstanceOf[E]))
 
   override def remove(o: Any): Boolean =
-    inner.remove(Box(o.asInstanceOf[E]))
+    inner.remove(Box(o.asInstanceOf[E])).nonEmpty
```
julienrf: The `remove` operation now returns an `Option` on `mutable.Set`. See also https://github.com/scala/collection-strawman/issues/551

javalib/src/main/scala/java/util/NavigableView.scala
```diff
@@ -121,22 +124,23 @@ private[util] class NavigableView[E](original: NavigableSet[E],
 
   def last(): E = {
     val iter = iterator()
-    if (iter.hasNext) iter.asScala.toTraversable.last
+    if (iter.hasNext) scala.collection.Iterable.from(iter.asScala).last
```
julienrf: `toTraversable` has been removed.

javalib/src/main/scala/java/util/NavigableView.scala
```diff
@@ -121,22 +124,23 @@ private[util] class NavigableView[E](original: NavigableSet[E],
...
     val boxedTo = Box(toElement)
 
     val subSetFun = { () =>
       val toTs =
-        if (toInclusive) innerNow.to(boxedTo)
+        if (toInclusive) innerNow.rangeTo(boxedTo)
         else innerNow.until(boxedTo)
       if (fromInclusive) toTs.from(boxedFrom)
-      else toTs.from(boxedFrom) - boxedFrom
+      else mutable.SortedSet.from(immutable.SortedSet.from(toTs.from(boxedFrom)) - boxedFrom)
```
julienrf: The `-` operator has been removed from mutable sets.

javalib/src/main/scala/java/util/NavigableView.scala
```diff
@@ -121,22 +124,23 @@ private[util] class NavigableView[E](original: NavigableSet[E],
...
       toInclusive: Boolean): NavigableSet[E] = {
     val innerNow = inner()
+    implicit val boxOrdering: Ordering[Box[E]] = innerNow.ordering
     val boxedFrom = Box(fromElement)
     val boxedTo = Box(toElement)
 
     val subSetFun = { () =>
       val toTs =
-        if (toInclusive) innerNow.to(boxedTo)
+        if (toInclusive) innerNow.rangeTo(boxedTo)
```
julienrf: `to` has been renamed to `rangeTo`.

library/src/main/scala-2.10_2.13.0-M3/scala/scalajs/ScalaVersionSpecificRuntime.scala
```diff
@@ -0,0 +1,19 @@
+package scala.scalajs
+
+import scala.collection.GenTraversableOnce
+
+trait ScalaVersionSpecificRuntime {
```
julienrf: This trait contains the parts of `scala.scalajs.Runtime` that are specific to Scala for versions 2.10 to 2.13.0-M3.

I’ve used the same technique for some other files. Alternatively, when there were too many differences, I duplicated the complete implementation.

library/src/main/scala-2.13/scala/scalajs/js/ImmutableArray.scala
```diff
@@ -0,0 +1,47 @@
...
+\*                                                                      */
+
+package scala.scalajs.js
+
+import scala.annotation.unchecked.uncheckedVariance
+import scala.collection.{IterableOnce, SeqFactory, StrictOptimizedSeqFactory, immutable, mutable}
+
+
+/** Equivalent of sci.ImmutableArray for js.Array */
+final class ImmutableArray[+A] private (array: Array[A])
```
julienrf: Note that there are no tests for this collection (apart from varargs usage in tests).

library/src/main/scala-2.13/scala/scalajs/js/ScalaVersionSpecificAny.scala
```diff
@@ -0,0 +1,22 @@
+package scala.scalajs.js
+
+import scala.collection.{BuildFrom, mutable}
+
+trait ScalaVersionSpecificAny {
+
+  implicit def buildFromArray[A]: BuildFrom[Array[_], A, Array[A]] = {
```
julienrf: `CanBuildFrom` has been superseded by `BuildFrom`, with a slightly different API.

library/src/main/scala/scala/scalajs/js/ArrayOps.scala
```diff
@@ -98,7 +83,7 @@ object ArrayOps {
     throw new UnsupportedOperationException(msg)
 
   /** Non-inlined implementation of [[ArrayOps.++]]. */
-  private def concat[A](left: Array[_ <: A], right: Array[_ <: A]): Array[A] = {
+  private def scalajsConcat[A](left: Array[_ <: A], right: Array[_ <: A]): Array[A] = {
```
julienrf: `concat` is now defined in the new collections. We have to use a different name.

project/Build.scala
```diff
@@ -1049,6 +1049,16 @@ object Build {
...
           scalaJSExternalCompileSettings
       ) ++ inConfig(Compile)(Seq(
+
+         unmanagedSourceDirectories += {
+            val dir = sourceDirectory.value
+            val scalaV = scalaVersion.value
+            CrossVersion.partialVersion(scalaV) match {
+              case Some((2, 13)) if scalaV != "2.13.0-M3" => dir / "scala-2.13"
+              case _ => dir / "scala-2.10_2.13.0-M3"
+            }
```
julienrf: I had to create a source directory for Scala versions between 2.10 and 2.13.0-M3, and another for Scala 2.13.0-M4.

project/Build.scala
```diff
@@ -1512,7 +1522,11 @@ object Build {
 
         unmanagedSourceDirectories in Test ++= {
           val testDir = (sourceDirectory in Test).value
-
+          val scalaV = scalaVersion.value
+          (
+            if (scalaV.startsWith("2.13") && scalaV != "2.13.0-M3") testDir / "scala-2.13"
+            else testDir / "scala-2.10_2.13.0-M3"
```
julienrf: Same for `testSuite / Test` source directories.

project/Build.scala
```diff
@@ -1757,8 +1771,10 @@ object Build {
             // Checkout proper ref. We do this anyway so we fail if
             // something is wrong
             val git = Git.open(trgDir)
+            git.fetch().call()
             s.log.info(s"Checking out Scala source version $ver")
-            git.checkout().setName(s"v$ver").call()
+            val ref = if (ver == "2.13.0-M4-pre-20d3c21") "2ec31613fd702cbde6da07dd614ff7fac9af3b64" else s"v$ver"
```
julienrf: Temporary workaround. To remove for 2.13.0.

scala-test-suite/src/test/resources/2.13.0-M4-pre-20d3c21/BlacklistedTests.txt
```diff
@@ -0,0 +1,143 @@
+## Do not compile
```
julienrf: I copied the 2.13.0-M3/BlacklistedTests file and added a few entries for new tests that don’t work in Scala.js.

tools/shared/src/main/scala-2.10_2.13.0-M3/org/scalajs/core/tools/Compat.scala
```diff
@@ -0,0 +1,16 @@
+package org.scalajs.core.tools
+
+import scala.collection.mutable
+
+// Makes 2.13 API available to older versions of the Scala library
+object Compat {
```
sjrd: This should be `private[tools]`.

tools/shared/src/main/scala-2.13/org/scalajs/core/tools/Compat.scala
```diff
@@ -0,0 +1,7 @@
+package org.scalajs.core.tools
+
+import scala.collection.mutable
+
+object Compat {
```
sjrd: This should be `private[tools]`.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/ComplianceRequirement.scala
```diff
@@ -33,7 +33,7 @@ object ComplianceRequirement {
    *  requirements.
    *  @throws BadComplianceException if the semantics are not compliant.
    */
-  final def checkCompliance(requirements: Traversable[ComplianceRequirement],
+  final def checkCompliance(requirements: Iterable[ComplianceRequirement],
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/ComplianceRequirement.scala
```diff
@@ -43,8 +43,8 @@ object ComplianceRequirement {
   }
 
   def mergeFromManifests(
-      manifests: Traversable[JSDependencyManifest]
-  ): Traversable[ComplianceRequirement] = {
+      manifests: Iterable[JSDependencyManifest]
+  ): Iterable[ComplianceRequirement] = {
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/DependencyResolver.scala
```diff
@@ -8,15 +8,15 @@ import JSLibResolveException.Problem
 object DependencyResolver {
 
   type DependencyFilter =
-    Traversable[FlatJSDependency] => Traversable[FlatJSDependency]
+    Iterable[FlatJSDependency] => Iterable[FlatJSDependency]
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/DependencyResolver.scala
```diff
@@ -8,15 +8,15 @@ import JSLibResolveException.Problem
...
 
   /** Constructs an ordered list of JS libraries to include. Fails if:
    *  - Resource names do not identify a unique resource on the classpath
    *  - Dependencies have cycles
    *  - Not all dependencies are available
    */
   def resolveDependencies(
-      manifests: Traversable[JSDependencyManifest],
+      manifests: Iterable[JSDependencyManifest],
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

tools/shared/src/main/scala/org/scalajs/core/tools/jsdep/ManifestFilters.scala
```diff
@@ -4,7 +4,7 @@ package org.scalajs.core.tools.jsdep
 object ManifestFilters {
 
   type ManifestFilter =
-    Traversable[JSDependencyManifest] => Traversable[JSDependencyManifest]
+    Iterable[JSDependencyManifest] => Iterable[JSDependencyManifest]
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/InfoChecker.scala
```diff
@@ -77,14 +77,14 @@ private final class InfoChecker(
     val expectedMethodIDs = expectedMethods.keySet
 
     if (actualMethodIDs != expectedMethodIDs) {
-      val missingMethods = expectedMethodIDs -- actualMethodIDs
+      val missingMethods = expectedMethodIDs diff actualMethodIDs
```
sjrd: `expectedMethodIDs.diff(actualMethodIDs)`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/checker/InfoChecker.scala
```diff
@@ -77,14 +77,14 @@ private final class InfoChecker(
...
       if (missingMethods.nonEmpty) {
         errorCount += 1
         logger.error(
             s"Missing methods in $className: $missingMethods")
       }
 
-      val unexpectedMethods = actualMethodIDs -- expectedMethodIDs
+      val unexpectedMethods = actualMethodIDs diff expectedMethodIDs
```
sjrd: `.diff()`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -488,7 +487,7 @@ abstract class GenIncOptimizer private[optimizer] (semantics: Semantics,
       interfaces = newInterfaces
 
       val methodAttributeChanges =
-        (parentMethodAttributeChanges -- methods.keys ++
+        ((parentMethodAttributeChanges diff methods.keySet) ++
```
sjrd: `.diff()`

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -511,7 +510,7 @@ abstract class GenIncOptimizer private[optimizer] (semantics: Semantics,
               newInterfaces.size != existingInterfaces.size) {
             val allMethodNames = allMethods().keys
             for {
-              intf <- oldInterfaces ++ newInterfaces -- existingInterfaces
+              intf <- (oldInterfaces ++ newInterfaces) diff existingInterfaces
```
sjrd: `.diff()`

tools/shared/src/main/scala/org/scalajs/core/tools/sem/Semantics.scala
```diff
@@ -247,7 +247,7 @@ object Semantics {
       productionMode = false,
       runtimeClassNameMapper = RuntimeClassNameMapper.keepAll())
 
-  def compliantTo(semantics: Traversable[String]): Semantics = {
+  def compliantTo(semantics: Iterable[String]): Semantics = {
```
sjrd: That is not an acceptable change, because it breaks a public API.

You need to use a `type` alias in an `Compat` object instead.

javalib/src/main/scala/java/math/BigDecimal.scala
julienrf: The “temporary” was about the `.toArray` call, which has been fixed by https://github.com/scala/scala/pull/6506 in the meantime.

javalib/src/main/scala/java/util/NavigableView.scala
```diff
@@ -121,22 +124,23 @@ private[util] class NavigableView[E](original: NavigableSet[E],
...
     val iter = iterator()
-    if (iter.hasNext) iter.asScala.toTraversable.last
+    if (iter.hasNext) scala.collection.Iterable.from(iter.asScala).last
     else null.asInstanceOf[E]
   }
 
   def subSet(fromElement: E, fromInclusive: Boolean, toElement: E,
       toInclusive: Boolean): NavigableSet[E] = {
     val innerNow = inner()
+    implicit val boxOrdering: Ordering[Box[E]] = innerNow.ordering
```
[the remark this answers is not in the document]
julienrf: Because the `mutable.SortedSet.from(…)` below (line 143) requires the implicit ordering instance. This is going to disappear with your suggestion of using `….clone() -= …`.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -146,10 +148,10 @@ class TreeSet[E] (_comparator: Comparator[_ >: E])
       var base = new mutable.TreeSet[Box[E]]
       base ++= inner.range(boxedFrom, boxedTo)
       if (!fromInclusive)
-        base = base - boxedFrom
+        base -= boxedFrom
```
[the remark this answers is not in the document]
julienrf: But this code is going to do the wrong thing on 2.10?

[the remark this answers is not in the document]
sjrd: No, obviously the tests already pass with your change to the code. I'm just saying the *comment* is now irrelevant and should be removed.

(I think we removed that very comment in 1.x anyway, already.)

library/src/main/scala/scala/scalajs/runtime/package.scala
```diff
@@ -2,9 +2,7 @@ package scala.scalajs
 
 import scala.annotation.tailrec
 
-import scala.collection.GenTraversableOnce
-
-package object runtime {
+package object runtime extends ScalaVersionSpecificRuntime {
```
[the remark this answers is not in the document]
julienrf: But we don’t have top-level definitions yet… It would be a bad idea to duplicate the whole `runtime` package object definition.

library/src/main/scala/scala/scalajs/runtime/package.scala
```diff
@@ -80,18 +78,6 @@ package object runtime {
     // scalastyle:on return
   }
 
-  @inline final def genTraversableOnce2jsArray[A](
-      col: GenTraversableOnce[A]): js.Array[A] = {
```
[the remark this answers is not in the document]
julienrf: It’s not sufficient: `TraversableOnce` (in 2.12) and `IterableOnce` (in 2.13) have a different API. In particular, `TraversableOnce` has no `iterator` method, whereas `IterableOnce` *only* has an `iterator` method.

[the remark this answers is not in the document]
sjrd: Then, in addition, write the implementation in the Compat object, and call the method in Compat from the method in the package object.

julienrf:
I’ve addressed all the review comments, except the one about `java.util.concurrent.ForkJoinWorkerThread`. I think we will have to copy the old implementation here?

I’ve blacklisted some partest tests that were using `scala.tools.partest.Util`, but I didn’t check if any previously blacklisted tests still don’t pass in 2.13.0-M4. I didn’t check either if some tests were neither blacklisted or whitelisted.

sjrd:
> I’ve addressed all the review comments, except the one about `java.util.concurrent.ForkJoinWorkerThread`. I think we will have to copy the old implementation here?

No, we can simply find another class from the JDK library that we do not implement, and which is not final. We can rewrite the test to use that other class instead. The point of this test is to make sure that the linker does not crash if there is a class in the .sjsir that extends another class that is *not* available.

> I didn’t check either if some tests were neither blacklisted or whitelisted.

The CI does that for you. If you want to test it locally, do:
```bash
$ cd ci/
$ ./check-partest-coverage.sh 2.13.0-pre-...
```

julienrf:
The remaining failing `compiler/test`s are fixed by #3346, or are caused by https://github.com/scala/bug/issues/10837.

julienrf:
Superseded by #3344.

PR #3332, opened 2018-04-05
Get rid of `TopLevelConstructorExportDef`s in the IR.

sjrd:
Another thing I found while formalizing the IR ...

Exports of Scala constructors can be directly encoded as top-level static *functions* instead, since what they did was creating a new instance inside of the desugared body anyway. Therefore, this commit gets rid of them at the IR level, by moving their desugaring from the `ClassEmitter` all the way back to the compiler.

The only piece of observable semantics we lose by doing so is that a constructor export def did set the `prototype` of the function to `$c_LTheClass.prototype`, indirectly allowing JS code to use `x instanceof ExportedCtor` to perform a Scala instance test. However, this "feature" was never documented, nor tested.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
@@ -61,7 +61,7 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     }
 
     def genConstructorExports(
-        classSym: Symbol): List[js.TopLevelConstructorExportDef] = {
+        classSym: Symbol): List[js.TopLevelMethodExportDef] = {
 
       val constructors = classSym.tpe.member(nme.CONSTRUCTOR).alternatives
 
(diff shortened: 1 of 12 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
```diff
@@ -777,27 +777,32 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
...
+        ensureBoxed(call,
+            enteringPhase(currentRun.posterasurePhase)(sym.tpe.resultType))
+      }
+
+      if (isNonNativeJSClass(currentClassSym)) {
+        assert(sym.owner == currentClassSym.get, sym.fullName)
+        boxIfNeeded(genApplyJSClassMethod(receiver, sym, args))
+      } else {
+        if (sym.isClassConstructor)
+          genNew(currentClassSym, sym, args)
```
gzm0: Why don't we need a box here? Because classes that are boxed never have exported constructors?

sjrd: No, it's because `genNew` already bypasses the logic that would have removed a box (in other words it already contains the box). In practice this only makes a difference if one exports the constructor of an `AnyVal`. Exporting the constructors of hijacked classes is not valid, because they have no constructors to begin with.

But the real reason to specialize the code path is that we generally avoid relying on the `resultType` of scalac class constructors, because they're the class in scalac but they must be `NoType` in our IR. And we definitely must not box to `undefined` here ;)

sjrd:
> Maybe we should just stop lying and say we allow exporting constructors and say factories instead?

Do you mean, in the docs, say that `@JSExportTopLevel` on a Scala class exports a factory function, rather than a constructor? That would probably make sense, indeed.

PR #3333, opened 2018-04-05
Fix #2383: Rewire Throwable to extend JavaScript's Error.

### LINKED ISSUE #2383: Support for assert(false, "I'd like this message to be printed to console")
Whenever I use an assert I like to use the optional second argument to give a helpful diagnostic message.  Could this be logged to console.error?

It'd also be fine if I could grab the message in the debugger, but though I'm dumped into this code on an assert fail:

  @elidable(ASSERTION) @inline
  final def assert(assertion: Boolean, message: => Any) {
    if (!assertion)
      throw new java.lang.AssertionError("assertion failed: "+ message)
  }

message is undefined, though I've definitely set it in my assert().

sjrd:
In order for our Throwables to be truly recognized by JavaScript as error classes, they must extend `Error`. With ECMAScript 2015 classes, this is possible (and easy), but with ECMAScript 5.1 functions and prototypes, we cannot *truly* make them errors, because there is no way we can convince ECMAScript to give them an `[[ErrorData]]` internal slot. That is because the `Error` constructor will only do that with a non-undefined `new.target`.

To minimize the gap between ES 5.1 and 2015, we still rewire the prototype of `Throwable` to that of `Error`.

In both cases, this rewiring means that `Throwable` does not inherit the methods of `j.l.Object` anymore, so we need to explicitly introduce bridges for those in `Throwable`.


sjrd:
Locally tested on Firefox and Chrome, in ES 5.1 and 2015.

Gives excellent results in V8 (Node.js and Chrome): we get the message *and* the stack trace.

But a bit disappointing in FF: now we see the trace without the message, whereas before we saw the message without the trace. Oh well ...

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -151,7 +151,10 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
     val parentVarWithGlobals = for (parentIdent <- tree.superClass) yield {
       implicit val pos = parentIdent.pos
       if (!tree.kind.isJSClass) {
-        WithGlobals(encodeClassVar(parentIdent.name))
+        if (className == ThrowableClass)
+          WithGlobals(js.VarRef(js.Ident("Error")))
```
gzm0: So just to double check: The reason we do not need to track `Error` as global here is because it is not dangerous?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -225,15 +228,21 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
       WithGlobals(js.Skip())
     } { parentIdent =>
       val (inheritedCtorDefWithGlobals, inheritedCtorRef) = if (!isJSClass) {
-        (WithGlobals(js.Skip()), envField("h", parentIdent.name))
+        if (className == ThrowableClass) {
+          val inheritableCtorDef =
+            makeInheritableCtorDef(js.VarRef(js.Ident("Error")), "hh")
+          (WithGlobals(inheritableCtorDef), envField("hh", className))
```
gzm0: I don't understand why we need this machinery (rather than just assigning `new Error()` to `prototyp`). Then again, `Error` seems pretty magic :S

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1256,6 +1265,8 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
 }
 
 private object ClassEmitter {
+  private final val ThrowableClass = "jl_Throwable"
```
gzm0: I'm a bit sad that we re-introduce another special type which is not a hijacked class. But since `Throwable` can have subclasses (which is not the case for any hijacked class), this is probably necessary.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ThrowableJSTest.scala
```diff
@@ -0,0 +1,35 @@
...
+package org.scalajs.testsuite.javalib.lang
+
+import scala.scalajs.js
+import scala.scalajs.LinkingInfo.assumingES6
+
+import org.junit.Test
+import org.junit.Assert._
+import org.junit.Assume._
+
+class ThrowableJSTest {
```
gzm0: We seem to not have tests that test catching behavior (i.e. interaction between Scala throwables and `js.JavaScriptException`). I think it would be a good idea to add some of them before this change to make sure we do not alter any behavior.

My current assumption is that `js.JavaScriptException` catches anything as `js.Any` so this change would not alter the behavior. Then again this would be dangerous, since it would basically mean there is no easy way to catch JS throwables without catching fatal exceptions (notably `UndefinedBehaviorError`). But than that would mean the `js.JavaScriptException` would only catch non-throwables, which would in turn be broken by this change (IIUC).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
sjrd: Yes: it's not dangerous, and we only use outside of method bodies.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
sjrd: We always do that machinery. It is necessary so that the `prototype` does not have the *fields* of the superclass. It must only chain to the prototype of the parent class, but not introduce any field. There's nothing specific to `Error` here.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
sjrd: It is way less special than hijacked classes, though, because `Throwable` does not affect the typing rules of the IR.

The closest special type in the IR is `java.lang.Class`. With a cursory glance, it doesn't seem like it *should* be special, but it is the result type of the `ClassOf` and `GetClass` nodes, so it does appear in the typing rules of the IR.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ThrowableJSTest.scala
gzm0: Ah, right. We test the other way around (we test for `Throwable` not `not Error`, obvious in hindsight since JS can throw anything). So yes, this shouldn't break anything (but we probably still should have tests).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
gzm0: > It is way less special than hijacked classes, though, because Throwable does not affect the typing rules of the IR.

Good point. Fair enough, so we can really keep this as an emitter implementation detail. Maybe we could guard the feature by checking that `Throwable` extends `Object` directly and not something else? (IIUC it would be totally valid IR to have a `jl_Throwable` which is something totally different than the java one).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
gzm0: Yeah, that's fair. It's just a bit ugly that we need to introduce `hh`. But OK :)

PR #3339, opened 2018-04-08
Fix #3338: Fix the intrinsic code for `ScalaRunTime.array_apply()`.

### LINKED ISSUE #3338: [0.6.x] Wrong optimization of array_apply() on an Array[Array[Char]]
The following code crashes on the first `isInstanceOf`:
```scala
    import js.DynamicImplicits.truthValue

    val a = Array(Array('a'))
    val b = arrayGet(a, 0)
    assert(b.isInstanceOf[Array[Char]])
    val c = b.asInstanceOf[Array[Char]]
    val d = arrayGet(c, 0)
    assert(d.isInstanceOf[Char])
    assert(d == 'a')
  }

  @inline
  def arrayGet[A](a: Array[A], i: Int): Any = a(i)
```
`a(i)` is compiled down to `ScalaRunTime.array_apply(a, i)` because `a` is a polymorphic array. There's a wrong optimization of `array_apply(a, i)` in this case because `a` is an `Array[Array[Char]]`: the optimizer wants to box the `Array[Char]` it gets from `a[i]` into a `java.lang.Character`.

The problematic test is [here](https://github.com/scala-js/scala-js/blob/a4bce397ba035e0d6b2c8454ae29c8739ece5ad5/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala#L1980). It is missing an `&& depth == 1`.

The bug does not exist in 1.x obviously, because of cc56fe758fbd1e6505ba05494142eb89b159780e

sjrd:


gzm0:
Shouldn't the commit be `[no-master]`?

sjrd:
No, the test should be merged forward. Same story as in #3317.

PR #3342, opened 2018-04-11
Fix #3294: Start testing instrumentation code via module initializer

### LINKED ISSUE #3294: Proposal: Start TestAdapter's bridge via module initializers
In Scala.js 1.0.x, with the advent of `scalaJSMainModuleInitializer` (aff7dcf2440da9f9134f6d85060cfba2fe322553) we have removed:
- The need to export main methods
- The capability of dynamically (i.e. after linking type) deciding which main is run.

I propose to do the same for testing:

- Define `TestAdapter.bridgeModuleInitializer` that starts the bridge.
- Add an sbt setting `scalaJSTestModuleInitializer`
  - defaulting to `true` in the `Test` configuration
  - mutually exclusive with `scalaJSMainModuleInitializer`
  - required to run `test*`
  - wire it to add `TestAdapter.bridgeModuleInitializer` to `scalaJSModuleInitializers`
- Remove `TestAdapter.ModuleIdentifier` without replacement.
- Simply take a `JSEnv.Input` in `TestAdapter` (after #3255) and wire it to `JSEnv.startWithCom`.
- Always add `test-interface` to `libraryDependencies in Test` (since it doesn't have an export anymore, that is fine).

This would simplify the `TestAdapter` and make it less dependent on how we build JS modules in the future. It would align the semantics/capabilities of `test` with `run` / `runMain`.

The major pain point here is how to deal with `testHTML`: It currently invokes a different exported method (from within `testadapter`) that builds the HTML UI. Solutions are:
1. Create another initializer for `testHTML`
1. Make `TestAdapter`'s initializer depend on a global, private flag injected in the template to dispatch to HTML instead.

While solution (1) seems cleaner, it is probably prohibitively annoying to users (and ourselves). Solution (2) seems to be a reasonable trade-off, especially since the HTML template is versioned together with the initializer definition. Therefore, we can change the interaction between them in a fully compatible manner. (Further, note that the HTML template already relies on global variables for test definitions).

If we go with solution (2), we should probably name the initializer `testadapter.moduleInitializer` or something.

gzm0:


test-adapter/src/main/scala/org/scalajs/testadapter/package.scala
```diff
@@ -12,8 +12,13 @@ package org.scalajs
...
 
+import org.scalajs.linker.ModuleInitializer
+
 package object testadapter {
   private[testadapter] implicit class AwaitFuture[T](val t: Future[T]) extends AnyVal {
     def await(): T = Await.result(t, Duration.Inf)
   }
+
+  def testAdapterInitializer: ModuleInitializer =
+    ModuleInitializer.mainMethod("org.scalajs.testinterface.Bridge", "start")
```
gzm0: I don't really like this package level stuff. WDYT? Should I make an object holder for it instead?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -104,6 +104,10 @@ object ScalaJSPlugin extends AutoPlugin {
         "If true, adds the `mainClass` as a module initializer of the Scala.js module",
         APlusSetting)
 
+    val scalaJSUseTestModuleInitializer = SettingKey[Boolean]("scalaJSUseTestModuleInitializer",
+        "If true, adds the a module initializer required for testing to the Scala.js module",
```
sjrd: typo: the a module

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -335,33 +333,53 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
+      scalaJSUseMainModuleInitializer := false,
+
+      // Use test module initializer by default.
+      scalaJSUseTestModuleInitializer := true,
+
+      scalaJSModuleInitializers ++= {
+        if (scalaJSUseTestModuleInitializer.value) {
+          if (scalaJSUseMainModuleInitializer.value) {
+            throw new MessageOnlyException("You may only set one of " +
+                "scalaJSUseMainModuleInitializer and scalaJSUseTestModuleInitializer to true")
```
sjrd: Include `in $configName` in the message where `val configName = configuration.value.name`?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -335,33 +333,53 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
-
-        val moduleIdentifier = scalaJSLinkerConfig.value.moduleKind match {
-          case ModuleKind.NoModule =>
-            ModuleIdentifier.NoModule
-          case ModuleKind.CommonJSModule =>
-            ModuleIdentifier.CommonJSModule(scalaJSLinkedFile.value.data.getPath)
+        if (!scalaJSUseTestModuleInitializer.value) {
+          throw new MessageOnlyException(
+              "You may only use test related functionality in a Scala.js " +
+              "project, if scalaJSUseTestModuleInitializer := true")
```
sjrd: Same. I would include `in $configName` in the message.

sjrd: Speaking of which, you could change the `fork in Test` in the other check above to `fork in $configName` as well.

test-adapter/src/main/scala/org/scalajs/testadapter/package.scala
sjrd: Yeah, I was also surprised to see this here. An object holder would probably be better.

And actually, as the CI discovered, you cannot declare this `ModuleInitializer` here, because `testAdapter` does not know about the `linker`. You have to leave the construction of the `ModuleInitializer` to the sbt plugin, and expose the two strings independently.

So maybe the holder could look like:
```scala
object TestAdapterInitializer {
  val ModuleClassName: String = "org.scalajs.testinterface.Bridge"
  val MainMethodName: String = "start"
}
```

test-common/src/main/scala/org/scalajs/testcommon/TestInterfaceMode.scala
```diff
@@ -0,0 +1,26 @@
+package org.scalajs.testcommon
+
+/** Mode in which the test interface executes. */
+private[scalajs] sealed trait TestInterfaceMode
```
sjrd: sealed abstract class, as we usually use?

test-common/src/main/scala/org/scalajs/testcommon/TestInterfaceMode.scala
```diff
@@ -0,0 +1,26 @@
...
+    def serialize(x: TestInterfaceMode, out: Serializer.SerializeState): Unit = x match {
+      case FullBridge =>
+        out.write(0)
+
+      case HTMLRunner(tests) =>
+        out.write(1)
+        out.write(tests)
+    }
+
+    def deserialize(in: Serializer.DeserializeState): TestInterfaceMode = in.read[Int] match {
```
sjrd: Aren't calls to `read` supposed to have `()`? (same in `case 1 =>` below)

gzm0:
Updated.

sjrd:
Bootstrap tests hang, e.g., [here](https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3342/2/pipeline). I think this happens because the module initializer for the bridge is [not added by the `TestSuiteLinker`](https://github.com/scala-js/scala-js/blob/d98816ebb94d5034b4f23860ebe2b793c4b45e50/test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala#L25).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -335,33 +333,62 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
-
-        val moduleIdentifier = scalaJSLinkerConfig.value.moduleKind match {
-          case ModuleKind.NoModule =>
-            ModuleIdentifier.NoModule
-          case ModuleKind.CommonJSModule =>
-            ModuleIdentifier.CommonJSModule(scalaJSLinkedFile.value.data.getPath)
+        if (!scalaJSUseTestModuleInitializer.value) {
+          throw new MessageOnlyException(
+              s"You may only use `test in $configName` tasks in " +
+              "a Scala.js project, if `scalaJSUseTestModuleInitializer in " +
```
sjrd: Ah, I hadn't seen that before, but there shouldn't be a comma here ;)

(I haven't yet gotten used to all those commas when I read in German :sweat_smile:)

gzm0: Done.

test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala
```diff
@@ -22,7 +22,11 @@ object QuickLinker {
 
     val linker = StandardLinker(config)
 
-    val moduleInitializers = build.TestSuiteLinkerOptions.moduleInitializers
+    val moduleInitializers = {
+      build.TestSuiteLinkerOptions.moduleInitializers :+
+      // Copied from org.scalajs.testadapter.TestAdapaterInitializer.
+      ModuleInitializer.mainMethod("org.scalajs.testinterface.Bridge", "start")
```
gzm0: So, this is obviously not very nice :S The other option we have is to start cross compiling parts of the `test-adapter` package (although it will never make sense to use it in JS, since by definition it is a JS/JVM bridge).

WDYT we should do here?

sjrd: The alternative is worse, IMO. I suggest we keep it this way, with the hard-coded copy of `TestAdapterInitializer` in `TestSuiteLinker`.

gzm0:
Updated (`import org.scalajs.linker.ModuleInitializer` was still around...)

sjrd:
```
java.lang.AssertionError: assertion failed: Bad number of scalaJSModuleInitializers in Test
```

gzm0:
LOL, all my own new stuff keeps breaking :P

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -335,33 +333,62 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
+       */
+      scalaJSUseMainModuleInitializer := false,
+
+      // Use test module initializer by default.
+      scalaJSUseTestModuleInitializer := true,
+
+      scalaJSModuleInitializers ++= {
+        if (scalaJSUseTestModuleInitializer.value) {
+          if (scalaJSUseMainModuleInitializer.value) {
+            val configName = configuration.value.name
```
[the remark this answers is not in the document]
gzm0: I just extracted into vals. That OK?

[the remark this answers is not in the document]
sjrd: Indeed, that works too.

PR #3346, opened 2018-04-23
Fix reported ambiguous method names and make error pos deterministic

julienrf:
- Fix the type test used to determine whether we should use the jsName
  or the name of a symbol.

- Sort positions of alternative ambiguous overloads and pick the last
  one to not rely on the arbitrary order of the hashmap implementation
  used by the underlying compiler.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -621,13 +621,18 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     private def reportCannotDisambiguateError(alts: List[Exported]): Unit = {
       val currentClass = currentClassSym.get
 
-      // Find a position that is in the current class for decent error reporting
-      val pos = alts.collectFirst {
+      /* Find a position that is in the current class for decent error reporting.
+       * If there are more than one, always use the "highest" one (i.e., the
+       * one coming last in the source text) so that we reliably display the
+       * same error in all compilers.
+       */
+      val validPositions = alts.collect {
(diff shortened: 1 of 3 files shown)


sjrd:
With all versions:
```
[error] Test org.scalajs.core.compiler.test.JSExportTest.noConflictingExport failed: should have right errors expected:<newSource1.scala:[4]: error: Cannot disa...> but was:<newSource1.scala:[6]: error: Cannot disa...>
[error]     at org.scalajs.core.compiler.test.util.TestHelpers$CompileTests.hasErrors(TestHelpers.scala:32)
[error]     at org.scalajs.core.compiler.test.JSExportTest.noConflictingExport(JSExportTest.scala:174)
[error]     ...
```
I think the expected message of that test must be changed.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -621,13 +621,18 @@ trait GenJSExports extends SubComponent { self: GenJSCode =>
     private def reportCannotDisambiguateError(alts: List[Exported]): Unit = {
       val currentClass = currentClassSym.get
 
-      // Find a position that is in the current class for decent error reporting
-      val pos = alts.collectFirst {
+      /* Find a position that is in the current class for decent error reporting.
+       * If there are more than one, always use the "highest" one (i.e., the
+       * one coming last in the source text, so that we reliably display the
+       * same error in all compilers.
```
sjrd: I just realized the opening `(` is not closed. You might as well fix it at the same time as adapting the test.

sjrd:
Could you squash the two commits, please? (keeping only the first commit's message, obiously) In the process, I suggest you change "position" to "pos" in the commit message, so that the first line is not too long for GitHub's taste :grin:

Issue #3330, opened 2018-04-04
Simplify VirtualFile API


gzm0:
Grab-bag issue for various things. Notably:

- [x] Remove all the different types of `VirtualFiles`, only have a single binary one.
- [x] Make it async (#3635)
- [x] Introduce a custom Output type for the linker.
- [x] Determine future of Node* virtual files (they are currently in the public API, not sure we want this).
- [x] Make mem file API immutable (factories instead of `withMethods`).
- [x] Consider splitting the hierarchy of readable/writable files. We never need both (and probably wont in the future).
- [x] Organize remainder of virtual file types (writable file / binary file for JSEnvs), also see #3656.
- [x] Make as much stuff final as we can.
- [x] Use `java.nio.file.Path` wherever we can. #3658

PR #3347, opened 2018-04-24
VirtualFile Cleanup

gzm0:
This is a first part of #3330 where we simplify the virtual file API substantially.


linker/shared/src/main/scala/org/scalajs/linker/Output.scala
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  @param sourceMapURI URI to reach the source map from the JavaScript file.
+ *      This is typically a relative URI but is not required. A [[Linker]]
+ *      should use this (to emit into [[jsFile]]), even if [[sourceMap]] is not
+ *      set.
+ *
+ *  @param jsFileURI URI to reach the JavaScript file from the source map. This
+ *      is typically a relative URI but is not required. A [[Linker]] may use
+ *      this even if [[sourceMap]] is not set, but it is typically meaningless.
+ */
```
gzm0: I'm not 100% convinced that this is a reasonable API, but it's the best I could come up with.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -178,16 +180,19 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
             IO.createDirectory(output.getParentFile)
 
-            linker.link(ir, moduleInitializers,
-                AtomicWritableFileVirtualJSFile(output),
-                sbtLogger2ToolsLogger(log))
+
+            val out = Output(AtomicWritableFileVirtualBinaryFile(output))
+              .withSourceMap(AtomicWritableFileVirtualBinaryFile(sourceMapFile))
+              .withSourceMapURI(???)
+              .withJSFileURI(???)
```
gzm0: TODO :)

sbt-plugin/src/main/scala-sbt-0.13/org/scalajs/sbtplugin/SBTCompat.scala
```diff
@@ -37,8 +37,9 @@ private[sbtplugin] object SBTCompat {
 
       def delete(classes: Iterable[File]): Unit = {
         inherited.delete(classes flatMap { classFile =>
-          val scalaJSFiles = if (classFile.getPath endsWith ".class") {
-            val f = FileVirtualFile.withExtension(classFile, ".class", ".sjsir")
+          val scalaJSFiles = if (classFile.getPath.endsWith(".class")) {
+            val f = new File(classFile.getParentFile(),
+                classFile.getName.stripSuffix(".class") + ".sjsir")
```
gzm0: TODO: Move extension logic to sbt plugin as private members.

sjrd:
You did not manage to break the `partestc` CI job ... how disappointing :laughing:

gzm0:
Bisect has shown that introducing Output makes the bootstrap test fail. Still investigating why.

gzm0:
This should be ready for review. I'll start a CI branch for the individual commits as I'm pretty sure some of them wont work :)

partest/src/main/scala/scala/tools/nsc/MainGenericRunner.scala
```diff
@@ -83,9 +83,9 @@ class MainGenericRunner {
     val linker = StandardLinker(linkerConfig)
 
     val sjsCode = {
-      val output = WritableMemVirtualJSFile("partest.js")
-      linker.link(ir, moduleInitializers, output, logger)
-      output
+      val out = WritableMemVirtualBinaryFile("partest.js")
+      linker.link(ir, moduleInitializers, Output(out), logger)
+      out
```
sjrd: I think half of this change (*returning* a `WritableMemVirtualBinaryFile` to put in `sjsCode`) should go to the preceding commit (see build failure of that other commit).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -161,6 +161,14 @@ private[sbtplugin] object ScalaJSPluginInternal {
         val linker = (scalaJSLinker in key).value
         val usesLinkerTag = (usesScalaJSLinkerTag in key).value
 
+        if (!FileVirtualFile.hasExtension(output, ".js")) {
+          throw new MessageOnlyException(
+              s"artifactPath in $key must have '.js' extension")
```
sjrd: Is that really necessary? Can't we just blindly add `.map` at the end of the `path` out `output`, without ensuring it ends in `.js`?

io/jvm/src/main/scala/org/scalajs/io/AtomicFileOutputStream.scala
```diff
@@ -43,8 +44,17 @@ private[io] class AtomicFileOutputStream private (
     if (!tmpFile.renameTo(baseFile)) {
       // Renaming failed. Fallback to copy
       try {
-        IO.copyTo(FileVirtualBinaryFile(tmpFile),
-            WritableFileVirtualBinaryFile(baseFile))
+        try {
+          // Try atomic move copy.
+          Files.copy(tmpFile.toPath, baseFile.toPath,
```
sjrd: There doesn't appear to be such a thing as atomic *copy*. See [here](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#copy-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-):

> Copying a file is not an atomic operation.

What you want is probably an atomic [`move`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#move-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-), which, among other things, says:

> `ATOMIC_MOVE`: The move is performed as an atomic file system operation and all other options are ignored. [...]

so `REPLACE_EXISTING` would be redundant, and

> If the move cannot be performed as an atomic file system operation then `AtomicMoveNotSupportedException` is thrown.

Note that `AtomicMoveNotSupportedException` is not a subtype of `UnsupportedOperationException` so the catch below needs to be adapted.

sjrd: Also, this specific change should probably be in its own commit. It may have been prompted by the removal of `IO.copyTo`, but using an actual `ATOMIC_MOVE` will improve the observable quality of this implementation, so it doesn't fit in a commit that says "Remove most the API that is unused".

io/shared/src/main/scala/org/scalajs/io/VirtualFiles.scala
```diff
@@ -34,6 +34,12 @@ trait VirtualBinaryFile extends VirtualFile {
   def inputStream: InputStream
 }
 
-trait WritableVirtualBinaryFile extends VirtualBinaryFile {
+/** A writable virtual binary file.
+ *
+ *  @note This does **not** extend [[VirtualBinaryFile]] nor [[VirtualFile]],
+ *      mainly because the Scala.js stack does never need these two things
```
sjrd: "does never need" sounds weird. "never needs"?

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
```diff
@@ -10,52 +10,41 @@
...
-
 import java.nio.charset.StandardCharsets
 
-/** A base class for simple in-memory mutable virtual files. */
-class MemVirtualFile(val path: String) extends VirtualFile {
-  private[this] var _version: Option[String] = None
+/** A simple in-memory immutable virtual binary file. */
+class MemVirtualBinaryFile(
+    final val path: String,
+    final val content: Array[Byte],
```
sjrd: It's not really immutable if the `content` is a mutable `Array[Byte]`...

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
```diff
@@ -10,52 +10,41 @@
...
+  def apply(path: String, content: Array[Byte],
+      version: Option[String] = None): MemVirtualBinaryFile = {
+    new MemVirtualBinaryFile(path, content, version)
+  }
 
-  final def withVersion(v: Option[String]): this.type = {
-    version = v
-    this
+  def fromStringUTF8(path: String, content: String,
+      version: Option[String] = None): MemVirtualBinaryFile = {
```
sjrd: Prefer overloads over default parameters. Default parameters cause trouble with binary compatibility.

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
```diff
@@ -10,52 +10,41 @@
...
-      WritableMemVirtualBinaryFile.this.content = this.toByteArray
+      _content = this.toByteArray
     }
   }
-}
 
-object WritableMemVirtualBinaryFile {
-  def apply(path: String): WritableMemVirtualBinaryFile =
-    new MemVirtualBinaryFile(path) with WritableMemVirtualBinaryFile
+  def toReadable(path: String, version: Option[String] = None): MemVirtualBinaryFile =
```
sjrd: Same comment re default parameters.

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
sjrd: Also: `fromUTF8String` instead of `fromStringUTF8`?

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -28,10 +28,12 @@ private[scalajs] class NodeVirtualJarScalaJSIRContainer(path: String)
       (name, entry) <- zip.files.toList
       if name.endsWith(".sjsir")
     } yield {
-      val path = s"${this.path}:$name"
-      new MemVirtualSerializedScalaJSIRFile(path, name)
-        .withContent(new Int8Array(entry.asArrayBuffer()).toArray)
-        .withVersion(this.version)
+      new MemVirtualSerializedScalaJSIRFile(
+        path = s"${this.path}:$name",
```
sjrd: 4 spaces.

test-adapter/src/main/scala/org/scalajs/testadapter/HTMLRunnerBuilder.scala
```diff
@@ -9,53 +9,67 @@
...
 import org.scalajs.testcommon._
 
 /** Template for the HTML runner. */
 object HTMLRunnerBuilder {
 
-  private val cssFile: MemVirtualTextFile = {
+  private def tmpFile(path: String): File = {
+    // - createTempFile requires a prefix of at least 3 chars
+    // - use the path as suffix so the extension stays (some browsers need that)
+    val f = File.createTempFile("tmp-", path)
```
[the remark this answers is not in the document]
gzm0: We should also have at least some of the name for debugging purposes. If something chokes and the includes just say `1234567.js`, it's super annoying.

Unfortunately, there does not seem to be any documentation about allowed characters in the prefix/suffix. Maybe we should just take `/[a-z_-.]*$/i` of the path? We get all extensions and reasonable names, but nothing nasty. (But then again, why not allow tons of other special chars that are not special on most OSs).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Yes we can. I just tried to preserve existing behavior. But probably indeed not necessary.

test-adapter/src/main/scala/org/scalajs/testadapter/HTMLRunnerBuilder.scala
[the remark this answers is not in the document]
sjrd: `/[a-z_-.]*$/i` sounds reasonable. I would add `0-9`.

We could go further and take any character that passes `Character.isJavaIdentifierPart`, but I'm not sure there is much value in taking anything that's non-ASCII.

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
gzm0: > fromUTF8String instead of fromStringUTF8

Not really if you think about it: the `String` is not UTF8, it's a `String`. The result is UTF8 :S

sjrd: Ah good point. My mistake.

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
gzm0: That's fair :) I've hidden the `content`. It doesn't need to be accessible anyways. And if we need it in the future, we can use a readonly `ByteBuffer` to wrap it.

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
```diff
@@ -65,6 +67,9 @@ class MemVirtualBinaryFile(p: String) extends MemVirtualFile(p)
     content = v
     this
   }
+
+  final def withUTF8String(v: String): this.type =
```
sjrd: By your own reasoning in `fromStringUTF8`, this one should be called `withStringUTF8`, shouldn't it?

linker/shared/src/main/scala/org/scalajs/linker/irio/VirtualIRFiles.scala
```diff
@@ -9,6 +9,7 @@
 package org.scalajs.linker.irio
 
 import java.io._
+import java.net.URI
```
sjrd: Spurious change.

io/jvm/src/main/scala/org/scalajs/io/AtomicFileOutputStream.scala
```diff
@@ -40,14 +40,13 @@ private[io] class AtomicFileOutputStream private (
...
-      } finally {
-        tmpFile.delete()
-      }
+    try {
+      // Try atomic move.
+      Files.move(tmpFile.toPath, baseFile.toPath, StandardCopyOption.ATOMIC_MOVE)
+    } catch {
+      case _: AtomicMoveNotSupportedException =>
+        // Fallback to normal copy.
+        Files.copy(tmpFile.toPath, baseFile.toPath, StandardCopyOption.REPLACE_EXISTING)
```
sjrd: I think that in the fallback, we should still `tmpFile.delete()` after the `Files.copy`.

sjrd: Also, there's a typo in the commit message: Automic

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
```diff
@@ -10,52 +10,49 @@
...
-  private[this] var _content: Array[Byte] = new Array[Byte](0)
+object MemVirtualBinaryFile {
+  def apply(path: String, content: Array[Byte]): MemVirtualBinaryFile =
+    apply(path, content, None)
 
-  def content: Array[Byte] = _content
-  def content_=(v: Array[Byte]): Unit = _content = v
+  def apply(path: String, content: Array[Byte],
+      version: Option[String]): MemVirtualBinaryFile = {
+    new MemVirtualBinaryFile(path, content, version)
```
sjrd: Hum ... if we're being pedantic, we would still need a defensive copy of `content` when allocating the `MemVirtualBinaryFile`. Otherwise the caller could still modify the `content` afterwards if it keeps a reference to it :s

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
gzm0: Yes

io/shared/src/main/scala/org/scalajs/io/MemFiles.scala
gzm0: I have declared this as UB :) Making a copy is probably not a good idea because this is of the path of reading jars.

linker/shared/src/main/scala/org/scalajs/linker/Output.scala
```diff
@@ -0,0 +1,63 @@
...
+    this(jsFile, None, None, None)
+
+  def withSourceMap(sourceMap: WritableVirtualBinaryFile): Output =
+    copy(sourceMap = Some(sourceMap))
+
+  def withSourceMapURI(sourceMapURI: URI): Output =
+    copy(sourceMapURI = Some(sourceMapURI))
+
+  def withJSFileURI(jsFileURI: URI): Output =
+    copy(jsFileURI = Some(jsFileURI))
```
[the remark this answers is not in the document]
gzm0: That's the idea I started out with. It indeed seems that setting either of the URIs without setting `sourceMap` is pointless. However, any combination of them can be meaningful.

- None of them --> no linking / no reasonable relative names
- No `sourceMapURI` --> source map should not appear in prod JS, but want back reference.
- No `jsFileURI` --> no meaningful back reference (e.g. in memory JS only), but want source map
- Both of them --> default case

We cannot (and should not) solve this with overloads only. I found it the lesser evil to allow the admittedly degenerate case of having the rel URIs but not the source map.

Maybe we should change the documentation and say it is unspecified what happens with the rel URIs if source map is not set?

sjrd:
Could you remind me what's the status of this PR? Are you waiting for another round of review? (the CI failure looks like a trivial compile error)

Should we try to make it fit into M4? If yes, I guess the async version would be nice-to-have as well, but that might still take time?

We shouldn't wait much too long before publishing a 1.0.0-M4, because of the Scala 2.13.0-M4 support. Ideally not more than a few weeks. Also for thesis purposes, before July 1 (6 weeks from now), because I'll have to submit the final version of my thesis ~July 4, and I'd like to be able to refer to a versioned release instead of a git SHA ;-)

gzm0:
- I'm waiting for your opinion on `LinkerOutput` and review of your requested changes.
- I need to fix the compile error.

gzm0:
Async will not happen in a few weeks :)

sjrd:
Looks good, other than:

* The commit message of introducing `LinkerOutput` still says "introduce `linker.Output`"
* The finer details of the `LinkerOutput` spec, as discussed above.

gzm0:
Updated

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -139,35 +142,44 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
...
+  private def closureOptions(sourceMapURI: Option[URI]) = {
     val options = new ClosureOptions
     options.prettyPrint = config.prettyPrint
     CompilationLevel.ADVANCED_OPTIMIZATIONS.setOptionsForCompilationLevel(options)
     options.setLanguageIn(ClosureOptions.LanguageMode.ECMASCRIPT5)
     options.setCheckGlobalThisLevel(CheckLevel.OFF)
 
-    if (config.sourceMap) {
-      options.setSourceMapOutputPath(outputName + ".map")
+    if (config.sourceMap && output.sourceMap.isDefined) {
```
sjrd:
```
[error] ClosureLinkerBackend.scala:179: not found: value output
[error]     if (config.sourceMap && output.sourceMap.isDefined) {
[error]                             ^
```

PR #3349, opened 2018-04-25
Run tests with `withAllowBigIntsForLongs(true)`.

sjrd:
Now that Node.js v10.0.0 has been released, with a version of v8 that supports BigInts under the flag `--harmony-bigint`, we can actually test that feature on the CI.


gzm0:
Jenkins, retest this please

sjrd:
Rebased + added a commit to fix #3350 instead of not testing it, which should be reviewed.

PR #3352, opened 2018-04-25
Do not delete `global.ArrayBuffer` in `NodeJSEnvForcePolyfills`.

sjrd:
Because Node.js v10.0.0's `readFileSync`, used by the core module loading mechanism, relies on it being there and working.


gzm0:
Why does this target master?

sjrd:
Because 0.6.x doesn't have `NodeJSEnvForcePolyfills`. It uses Rhino for the purpose of testing our ES 2015 polyfills. :wink:

PR #3354, opened 2018-04-26
Reduce acceptTimeout

eatkins:
I noticed that my tests were very slow to start running. While running
with sbt | ts '%H:%M:%.S', I found that there was almost always exactly
a 5 second delay between starting the process and the first output being
printed. I grepped for 5000 and found this line. Changing this parameter
eliminated the delay for me. The parameter seems to only be used in the
awaitConnection method, which just loops until serverSocket.accept
succeeds (with failure being a socket timeout). This makes me fairly
confident that this change won't break anything. I also would expect any
additional overhead of the extra thread wake ups to be negligible.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
@@ -51,7 +51,7 @@ abstract class AbstractNodeJSEnv(
   }
 
   /** Retry-timeout to wait for the JS VM to connect */
-  protected val acceptTimeout = 5000
+  protected val acceptTimeout = 500
 
   protected trait AbstractNodeRunner extends AbstractExtRunner with JSInitFiles {
 


eatkins:
I’m not sure that I fully agree. The only place the parameter is used is within a loop that will keep retrying if the timeout is reached. I don’t know if this is an os x specific issue, but on my machine, the timeout is _always_ hit the first time and this change dramatically improves the ux.

gzm0:
> The only place the parameter is used is within a loop that will keep retrying if the timeout is reached.

This is absolutely correct: The "timeout" is basically just a polling interval on whether the `JSEnv` got terminated (and hence it will never connect). However, why the timeout always trigger is a mystery to me: The only time this should matter is when the `JSEnv` terminates before establishing a connection at all.

Now since this won't make things worse for slower things (it will just poll a bit more), we already decommissioned this code for 1.0.x and @eatkins says it does improve things for him, I suggest we merge this.

sjrd:
Ah, right. I see now. Let's merge it, then.

That said, I wonder whether we'll need to do something about this in 1.x. IIUC that timed-out `accept()` uses an infinite time-out in 1.x, [here](https://github.com/scala-js/scala-js/blob/894a69de3b3a5d80c918fb981a50df7931d98910/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala#L163). If the `accept()` needs to time-out once before it actually works on the second try, what will happen with an infinite timeout?

eatkins:
Not sure if it's helpful, but I didn't see this behavior with scala.js 6.19 and bundler 0.9.0. I started seeing it with 6.22 and bundler 0.11.0.

PR #3355, opened 2018-04-30
Remove FrameworkDetector test.

gzm0:
Since 22a8d7dc52669c537154f21519752a8b6680a174 FrameworkDetector does not exist anymore, so the test is moot.

### DIFF
--- a/sbt-plugin-test/build.sbt
+++ b/sbt-plugin-test/build.sbt
@@ -117,8 +117,6 @@ lazy val multiTest = crossProject.
   jsSettings(baseSettings: _*).
   jsSettings(
     name := "Multi test framework test JS",
-    // Make FrameworkDetector resilient to other output - #1572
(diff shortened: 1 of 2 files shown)


sjrd:
Hum ... even if `FrameworkDetector` itself does not exist, isn't it still valuable to assert that our testing infrastructure as a whole survives and behaves sanely in case something prints to the console?

gzm0:
Sure, if you think there is value in it. However, it seems to me like this would be something better tested on a `JSEnv` level: The test adapter only uses the com and expects usage of it exclusively. So it is very unlikely that it interferes with printing, except if the `JSEnv` uses the same channel for com and printing. However, this should better be tested in the `JSEnv` test kit (which it currently isn't that much AFAIK).

So maybe we should:
- Target `master` with this.
- Improve com/stdout interaction in the test kit.

WDYT?

gzm0:
Superseded by #3466.

PR #3356, opened 2018-05-03
Fix #3340: Introduce org.scalajs.testing package

### LINKED ISSUE #3340: Introduce `org.scalajs.testing` package
Currently we have the following packages that are intimately related:
- `org.scalajs.testcommon`
- `org.scalajs.testadapter`
- `org.scalajs.testinterface`

Since the latter two share common functionality in the first one, there is a lot of `private[scalajs]` stuff in there. This is odd, since it gives the impression that the testing integration treats entire Scala.js as a special snowflake (which it doesn't #3337 is an example of us actively working to not make it special).

If we add a dot to all the packages above (e.g. `org.scalajs.test.common`, etc.) we can make them `private[test]`, which makes much more sense IMO.

gzm0:


junit-runtime/src/main/scala/org/scalajs/junit/JUnitTask.scala
```diff
@@ -3,7 +3,7 @@ package org.scalajs.junit
 import com.novocode.junit.{Ansi, RichLogger}
 import Ansi._
 import sbt.testing._
-import org.scalajs.testinterface.TestUtils
+import org.scalajs.testing.interface.TestUtils
```
sjrd: Ouch, that's actually problematic. It means that a testing framework using `TestUtils` cannot cross-compile for 0.6.x and 1.x anymore. Unless there's an `import` magic that would allow it? But then we should validate it.

Otherwise we need to either introduce type aliases with the new API in 0.6.x, or keep deprecated type aliases for the old API in 1.x.

SethTisue:
in the Scala 2.11 community build, at e.g. https://scala-ci.typesafe.com/view/scala-2.11.x/job/scala-2.11.x-jdk8-integrate-community-build/701/ we now see:

```
[utest] [error] /home/jenkins/workspace/scala-2.11.x-jdk8-integrate-community-build/target-0.9.11/project-builds/utest-f86bda089a2c87e91c70debe2f6d27b54ef8ba1d/utest/shared/src/main/scala/utest/runner/BaseRunner.scala:7: object scalajs is not a member of package org
[utest] [error] import org.scalajs.testinterface.TestUtils
[utest] [error]            ^
```

should I use the 0.6.x branch instead? (and stick with master in the 2.12 and 2.13 community builds?) or what?

SethTisue:
> should I use the 0.6.x branch instead? (and stick with master in the 2.12 and 2.13 community builds?)

done: https://github.com/scala/community-builds/pull/781

sjrd:
`TestUtils` has been deprecated in 0.6.25. Give it some time and testing frameworks will stop using it in favor of `Reflect` or https://github.com/portable-scala/portable-scala-reflect. I might submit a PR to uTest if it's the only project that was affected in the community build.

SethTisue:
@sjrd there's also minitest. I opened https://github.com/monix/minitest/issues/27 on it

PR #3357, opened 2018-05-03
Remove unused helper.

gzm0:


sjrd:
LGTM, but actually:
```bash
$ git grep liftToTry
test-common/src/main/scala/org/scalajs/testcommon/FutureUtil.scala:    def liftToTry: Future[Try[T]] =
```
So this method is not even used anywhere. It could be plain removed instead ;)

gzm0:
Updated.

PR #3344, opened 2018-04-17
CI ONLY Upgrade to Scala 2.13.0-M4

julienrf:
Continuation of #3344.

- [x] library
- [x] javalib
- [x] scalalib
- [x] compiler
- [x] testSuite
- [x] scalaTestSuite
- [x] linker
- [x] ir
- [x] partest
- [x] upgrade to latest Scala (a few things have been fixed upstream, and that will impact Scala.js)
- [x] squash commits

PR #3359, opened 2018-05-10
Preparations for the upgrade to Scala 2.13.0-M4.

sjrd:
This PR contains commits that get the codebase closer to a state where it can support 2.13.0-M4, by improving things that should have been done already anyway.

Most of it is extracted from #3344.


javalib/src/main/scala/java/math/BigDecimal.scala
```diff
@@ -194,8 +194,15 @@ object BigDecimal {
...
-  private[math] def newArrayOfPows(len: Int, pow: Int): Array[Long] =
-    new Array[Long](len - 1).scanLeft[Long, Array[Long]](1)((z, e) => z * pow)
+  private[math] def newArrayOfPows(len: Int, pow: Int): Array[Long] = {
+    val result = new Array[Long](len)
+    var nextValue = 1L
+    for (i <- 0 until len) {
+      result(i) = nextValue
+      nextValue *= pow
+    }
+    result
```
gzm0: Why not just `Array.iterate(1L, len)(_ * pow)`? Shouldn't this expand into the same anyways past optimizer?

test-interface/src/main/scala/org/scalajs/testinterface/TestDetector.scala
```diff
@@ -53,7 +53,7 @@ private[scalajs] object TestDetector {
     def tryLoad(name: String): Option[Framework] =
       tryLoadFromReflect(name).orElse(tryLoadFromExportsNamespace(name))
 
-    names.toStream.map(tryLoad).flatten.headOption
+    names.iterator.map(tryLoad).find(_.isDefined).flatten
```
gzm0: have you considered using `collectFirst { Some(f) => f }` for better readability?

sjrd:
Added a commit with the removal of procedure syntax in examples.

sjrd:
@gzm0 Anything else here?

PR #3361, opened 2018-05-12
Compatibility with Scala 2.13.0-M4.

sjrd:
This PR puts the codebase in a state where we can build the compiler plugins, libraries and test suite with Scala 2.12.0-M4 and its new collections. The tools API and related artifacts are not yet updated.

This is the minimal set of changes necessary to be able to build and back-publish for 2.13.0-M4 when it comes out. We will not be able to publish tools artifact for that version, but that is not critical.


sjrd:
This PR is ready for review. But it must not be merged yet because I will have to adapt it a little bit after they merge https://github.com/scala/scala/pull/6620 upstream.

This PR is enough to build and publish for 2.13.0-M4. So once it's merged, we can do an "emergency" release of 0.6.23 so that we're ready to back-publish it for 2.13.0-M4 when that one is released (any day now, according to the scala/scala team).

library/src/main/scala-new-collections/scala/scalajs/js/WrappedDictionary.scala
```diff
@@ -0,0 +1,162 @@
...
+  def map[B](f: ((String, A)) => (String, B)): WrappedDictionary[B] = {
+    val b = new WrappedDictionary.WrappedDictionaryBuilder[B]
+    val it = this.iterator
+    while (it.hasNext) {
+      b += f(it.next())
+    }
+    b.result()
+  }
+
+  // TODO overload collect, flatMap and concat
```
sjrd: @julienrf How import is this TODO? What are we missing if we do not do it now? And how hard would it be to address?

ci/checksizes.sh
```diff
@@ -37,7 +37,7 @@ REVERSI_OPT_GZ_SIZE=$(stat '-c%s' "$REVERSI_OPT.gz")
 
 case $FULLVER in
   2.10.2)
-    REVERSI_PREOPT_EXPECTEDSIZE=532000
+    REVERSI_PREOPT_EXPECTEDSIZE=533000
```
sjrd: The small increase in code size comes from the indirection through `runtime.toScalaVarArgs`. Even though it is inlined, the argument must be extracted in a temp var by the optimizer to preserve evaluation order (or so it thinks) which slightly increase code size.

sjrd:
Updated after https://github.com/scala/scala/pull/6620 was merged upstream.
But actually we're also depending on when/if/how much of https://github.com/scala/scala/pull/6611 is merged. So this is still a don't merge.

library/src/main/scala-new-collections/scala/scalajs/js/WrappedDictionary.scala
julienrf: Oh, sorry I forgot about that TODO. If we don’t overload `collect`, `flatMap` and `concat` they can only return a `mutable.Map` (since we don’t rely anymore on `CanBuildFrom` to determine the result type. Above you can see that I’ve added an overload of `map` that returns a `WrappedDictionary[B]` when the transformation function returns a tuple `(String, B)`.

sjrd: OK, thanks. We should probably address that, then, because it might be a source breaking change for some clients.

sjrd: Done.

sjrd:
OK, now this PR and its friend #3344 are complete wrt. the upcoming 2.13.0-M4. At least, to the best of my knowledge, and assuming they don't break anything anymore.

sjrd:
Actually ... not ready. They broke it again: https://github.com/scala/scala/pull/6572/files

sjrd:
Good to go again.

Note that I made a PR upstream at https://github.com/scala/scala/pull/6644, but since we override `Array.scala` anyway, we can go ahead whether or not it is merged upstream.

sjrd:
Well, Scala 2.13.0-M4 has reached Maven Central, so emergency just turned into deportation. I'm going to go ahead and merge this and publish 0.6.23. At least it works. If tweaks should be done, they can still be done later.

PR #3367, opened 2018-05-22
Fix #3348: Support hexadecimal notation in Double.parseDouble.

### LINKED ISSUE #3348: Doubles in hex notation fail to parse
reproducer:

`scala.util.Try("0x0p1".toDouble)`

expected:

`Success(0.0)`

actual:

`Failure(java.lang.NumberFormatException: For input string: "0x0p1")`

sjrd:
Since `Float.parseFloat` delegates to `Double.parseDouble`, we trivially get that one for free.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Double.scala
+++ b/javalanglib/src/main/scala/java/lang/Double.scala
@@ -63,22 +63,149 @@ object Double {
 
   @inline def valueOf(s: String): Double = valueOf(parseDouble(s))
 
-  private[this] lazy val doubleStrPat = new js.RegExp("^" +
-      "[\\x00-\\x20]*"   + // optional whitespace
-      "[+-]?"            + // optional sign
-      "(NaN|Infinity|"   + // special cases
-       "(\\d+\\.?\\d*|"  + // literal w/  leading digit
-        "\\.\\d+)"       + // literal w/o leading digit
-       "([eE][+-]?\\d+)?"+ // optional exponent
-      ")[fFdD]?"         + // optional float / double specifier (ignored)
-      "[\\x00-\\x20]*"   + // optional whitespace
+  private[this] lazy val doubleStrPat = new js.RegExp(
+      "^"                   +
+      "[\\x00-\\x20]*("     + // optional whitespace
+      "[+-]?"               + // optional sign
+      "(?:NaN|Infinity|"    + // special cases
+       "(?:\\d+\\.?\\d*|"   + // literal w/  leading digit
+        "\\.\\d+)"          + // literal w/o leading digit
+       "(?:[eE][+-]?\\d+)?" + // optional exponent
+      ")[fFdD]?"            + // optional float / double specifier (ignored)
+      ")[\\x00-\\x20]*"     + // optional whitespace
+      "$")
+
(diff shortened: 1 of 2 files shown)


javalanglib/src/main/scala/java/lang/Double.scala
```diff
@@ -63,22 +63,147 @@ object Double {
...
+      val integralPartStr = match2(2).asInstanceOf[String]
+      val fractionalPartStr = match2(3).asInstanceOf[String]
+      val binaryExpStr = match2(4).asInstanceOf[String]
+
+      if (integralPartStr == "" && fractionalPartStr == "")
+        fail()
+
+      def overflow(): scala.Double =
+        if (signStr == "-") scala.Double.NegativeInfinity
+        else scala.Double.PositiveInfinity
```
gzm0: This helper is unused.

javalanglib/src/main/scala/java/lang/Double.scala
```diff
@@ -63,22 +63,147 @@ object Double {
...
+
+      def underflow(): scala.Double =
+        if (signStr == "-") -0.0
+        else 0.0
```
gzm0: Consider inlining this at its sole usage site. `underflow` is arguably a strange name for that usage, since the original string was basically a form of "0" (and *not* something with a too small fractional part).

javalanglib/src/main/scala/java/lang/Double.scala
```diff
@@ -63,22 +63,147 @@ object Double {
...
+
+      /* Remove leading 0's in `mantissaStr`, because our algorithm assumes
+       * that there is none.
+       */
+      var i = 0
+      while (i != mantissaStr0.length && mantissaStr0.charAt(i) == '0')
+        i += 1
+      val mantissaStr = mantissaStr0.substring(i)
+      if (mantissaStr == "")
+        return underflow()
```
gzm0: Why is this short exit necessary? Wouldn't it be clearer to truncate `mantissaStr` to at least 1 digit and then follow normal execution flow?

javalanglib/src/main/scala/java/lang/Double.scala
sjrd: Ah yes. Initially I had zillions of cases of `return overflow()` and `return underflow()`. I didn't realize that my refactorings led to 1 `underflow()` and 0 `overflow()` ^^

javalanglib/src/main/scala/java/lang/Double.scala
sjrd: Ah, this one is fun. The thing is that if `mantissa == 0.0` makes it to the end, where we multiply it with `correctingPow`, but `correctingPow` is `Infinity` because the binary exponent was too large, then we get `0.0 * Infinity == NaN`. So we need to either explicitly test `mantissa == 0.0` or `correctingPow == Infinty`. Might as well test for the one we find out first.

I have added a comment for that, and a test case that fails if we don't apply the early exit.

sjrd:
Updated.

sjrd:
I had to disable that test on PhantomJS, because it has precision problems on that platform. It's not the first time we find oddities in edge cases of numbers of PhantomJS, so I blame it on the interpreter.

PR #3369, opened 2018-05-24
Fix JUnit issues

sjrd:


junit-test/shared/src/test/scala/org/scalajs/junit/utils/JUnitTest.scala
```diff
@@ -132,10 +132,11 @@ abstract class JUnitTest {
       )
     }
 
-    def assertion(testName: String, message: String): OutputBuilder = {
+    def assertion(testName: String, message: String,
+        exClass: String = "java.lang.AssertionError"): OutputBuilder = {
```
gzm0: Have you considered actually passing a `Class[_]` to avoid typos?

sjrd: I followed the same API as `def exception()`. I'm happy to switch them all to take `Class[_]` instead.

sjrd:
Updated with a third commit.

PR #3371, opened 2018-05-28
Fix some tests of the JDK that were failing on the JVM

sjrd:
The first commit was tested locally on a JDK 9.


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/OutputStreamWriterTest.scala
```diff
@@ -27,20 +26,23 @@ class OutputStreamWriterTest {
     bos.write(1)
     osw.write("ABC")
     assertFalse(bos.flushed)
+    assertEquals(1, bos.size)
```
gzm0: I do not understand this: This basically asserts that `osw` is not allowed to propagate the string until we call flush, no? That's definitely not true...

sjrd: Ah, good point. You're right, that's too restrictive.

sjrd:
Updated.

PR #3374, opened 2018-06-09
Only read full IR from Jars when needed.

gzm0:
Instead of fully reading every .sjsir file in a Jar, we only read the EntryPointsInfo and defer reading the trees.

### DIFF
--- a/linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
+++ b/linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
@@ -9,10 +9,11 @@
 package org.scalajs.linker.irio
 
 import java.io._
-import java.util.zip.{ZipInputStream, ZipEntry}
+import java.util.zip.{ZipFile, ZipEntry}
 
-import scala.annotation.tailrec
+import scala.collection.JavaConverters._
 
+import org.scalajs.ir
 import org.scalajs.io._
 
 final class FileVirtualScalaJSIRFile(f: File, val relativePath: String)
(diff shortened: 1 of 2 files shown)


gzm0:
@sjrd we should probably benchmark this. Do you have a setup that allows you to do this quickly? Otherwise I'll come up with something (but it'll take me a bunch of days for sure).

linker/jvm/src/main/scala/org/scalajs/linker/irio/SharedZipFile.scala
```diff
@@ -0,0 +1,34 @@
...
+package org.scalajs.linker.irio
+
+import java.io._
+import java.util.zip.ZipFile
+
+private[irio] final class SharedZipFile(file: File) {
+  private[this] var _references: Int = 0
+  private[this] var _zipFile: ZipFile = _
+
+  final def acquire(): ZipFile = synchronized {
```
sjrd: Have you considered making `acquire()` and `release()` private, and instead expose an API like:
```scala
final def acquireWithin[A](f: ZipFile => A): A = {
  val zipFile = acquire()
  try f(zipFile)
  finally release()
}
```
?

sjrd:
> Do you have a setup that allows you to do this quickly? Otherwise I'll come up with something (but it'll take me a bunch of days for sure).

Hum no I don't have anything like that. What I typically do is repeatedly `;testSuite/clean;testSuite/test:fastOptJS;last` and other variations on that theme.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -56,46 +57,46 @@ object FileScalaJSIRContainer {
...
-        val out =
-          if (0 <= size && size <= Int.MaxValue) new ByteArrayOutputStream(size.toInt)
-          else new ByteArrayOutputStream()
+  private def entryFile(entry: ZipEntry, fileVersion: Option[String]) = {
+    val filePath = s"$path:${entry.getName()}"
 
+    def withInputStream[A](f: InputStream => A) = {
+      val zf = zipFile.acquire()
+      try {
+        val in = zf.getInputStream(entry)
```
sjrd: Is this actually safe to use in concurrent threads on the same `ZipFile`? If not, I believe the thread should keep the lock on the `SharedZipFile` while reading the input stream for an entry.

linker/jvm/src/main/scala/org/scalajs/linker/irio/SharedZipFile.scala
gzm0: Yes. This will not be useful anymore once we move to a fully async world (`release()` would happen in a `andThen` closure). If you want I can change it then.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
gzm0: Hmmm... Good question. The internet seems not to have any info about this (only about `ZipFileInputStream`).

gzm0:
Ugh, so I can see that "Linker: Compute reachability" is about 25% slower 3.6s vs 4.5s on the test suite. However, we do not measure how long the caching step takes --> I need to add some more instrumentation.

gzm0:
Doesn't look too good:

timing

Seems like we are losing roughly 1s on overall linking (all of it in reachability where we actually need the trees for info calculation) but only gain about .2s in IR cache loading.

```
master
+-------+--------------------------------+---+----------+----------+----------+----------+----------+
|       |                                |N  |0%        |25%       |50%       |75%       |100%      |
+-------+--------------------------------+---+----------+----------+----------+----------+----------+
|[s]    |Basic Linking                   | 10|5.89219989|5.93742292|5.95171196|5.98378530|6.08650735|
|       |Emitter (write output)          | 10|6.35512326|7.46531361|7.81072722|8.71909887|9.30226972|
|       |Inc. optimizer                  | 10|6.73888677|6.95960304|7.19470459|7.35578124|7.86381543|
|       |Inc. optimizer: Incremental part| 10|0.41614853|0.41952407|0.42940568|0.43469447|0.45261667|
|       |Inc. optimizer: Optimizer part  | 10|6.29487406|6.50882908|6.73478953|6.90448631|7.40563476|
|       |Linker: Assemble LinkedClasses  | 10|0.23869883|0.25028340|0.25537785|0.25643068|0.27552008|
|       |Linker: Check IR                | 10|2.02885721|2.03740769|2.04405335|2.05510382|2.09003017|
|       |Linker: Compute reachability    | 10|3.59324072|3.64136444|3.64957458|3.65430791|3.73154059|
|       |Refiner                         | 10|0.57903138|0.59282202|0.59879094|0.60721720|1.23069876|
|       |Refiner: Assemble LinkedClasses | 10|0.02794562|0.02846429|0.02860794|0.02908156|0.03396334|
|       |Refiner: Compute reachability   | 10|0.52841881|0.54064695|0.54636063|0.55059377|1.18107149|
|       |Update IR cache                 | 10|0.46768658|0.46937800|0.47271807|0.49325308|0.51264774|
+-------+--------------------------------+---+----------+----------+----------+----------+----------+

zip
+-------+--------------------------------+---+----------+----------+----------+----------+----------+
|       |                                |N  |0%        |25%       |50%       |75%       |100%      |
+-------+--------------------------------+---+----------+----------+----------+----------+----------+
|[s]    |Basic Linking                   | 10|6.72135542|6.87768546|6.90811067|6.94055720|6.95540503|
|       |Emitter (write output)          | 10|6.37507263|7.13715290|7.69954649|8.36797832|8.55794520|
|       |Inc. optimizer                  | 10|6.80220019|6.95301488|7.13941591|7.66033694|8.56361930|
|       |Inc. optimizer: Incremental part| 10|0.41252946|0.41593995|0.42133626|0.43309482|0.44999743|
|       |Inc. optimizer: Optimizer part  | 10|6.35117402|6.50621396|6.69713315|7.19189428|8.10175720|
|       |Linker: Assemble LinkedClasses  | 10|0.23798810|0.23850145|0.23984600|0.24075389|0.24698443|
|       |Linker: Check IR                | 10|2.04390851|2.08076778|2.08835225|2.10881285|2.12480803|
|       |Linker: Compute reachability    | 10|4.38480595|4.54071994|4.56418927|4.59697888|4.61543471|
|       |Refiner                         | 10|0.59330414|0.60167574|0.60752610|0.63536047|1.21064103|
|       |Refiner: Assemble LinkedClasses | 10|0.02728309|0.02776132|0.02809906|0.02849831|0.02977002|
|       |Refiner: Compute reachability   | 10|0.54257750|0.55138697|0.55708522|0.58207679|1.16180429|
|       |Update IR cache                 | 10|0.21110350|0.21998392|0.22067230|0.22284186|0.22759736|
+-------+--------------------------------+---+----------+----------+----------+----------+----------+
```

sjrd:
Hum, yup, doesn't look so good :( Maybe the back-and-forth inside the zip file is killing it (data locality, perhaps even hard drive reads stalled because the heads are not in the right place).

gzm0:
Also, note that this is a version that opens a zip file for each read. So instead of opening it once, we open in `n + 1` times where `n` is the number of trees. I'll test how much performance is impacted by this. But yes, data locality is another theory indeed.

gzm0:
Closing this, since it does not seem to buy us anything.

PR #3377, opened 2018-06-13
A number of changes for the upcoming Scala 2.13.0-M5.

sjrd:
The changes in this PR allow `testSuite/test` to pass with today's tip of the `2.13.x` branch in scala/scala, i.e., commit https://github.com/scala/scala/commit/1ef688a914f2052ef558596e3beef6403054dc9a.

I have also locally tested that `testSuite/test` still passes with `++2.13.0-M4`.

### DIFF
--- a/javalib/src/main/scala/java/math/Conversion.scala
+++ b/javalib/src/main/scala/java/math/Conversion.scala
@@ -100,7 +100,7 @@ private[math] object Conversion {
           @tailrec
           def innerLoop(): Unit = {
             currentChar -= 1
-            result = Character.forDigit(resDigit % radix, radix) + result
+            result = Character.forDigit(resDigit % radix, radix).toString + result
             resDigit /= radix
             if(resDigit != 0 && currentChar != 0)
               innerLoop()
@@ -111,7 +111,7 @@ private[math] object Conversion {
           var i: Int = 0
           while (i < delta && currentChar > 0) {
             currentChar -= 1
-            result = '0' + result
+            result = "0" + result
             i += 1
           }
           i = tempLen - 1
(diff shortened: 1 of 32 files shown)


sjrd:
Updated for https://github.com/scala/scala/commit/80ee48b1880c8f817581a90f483b394273c237bc. Also `compiler/test` works.

SethTisue:
fyi @som-snytt for the imports changes

som-snytt:
@SethTisue thanks, I had verified that the delta made sense to me. Thanks @sjrd

junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
```diff
@@ -388,7 +388,14 @@ class ScalaJSJUnitPlugin(val global: Global) extends NscPlugin {
               varargsModule,
               definitions.wrapVarargsArrayMethodName(definitions.ObjectTpe),
               Nil, List(array))
-          gen.mkMethodCall(definitions.List_apply, List(wrappedArray))
```
gzm0: In this commit's message, maybe change "resist" to "support"? It feels like it's easier to understand.

javalib/src/main/scala/java/math/Conversion.scala
```diff
@@ -241,7 +241,7 @@ private[math] object Conversion {
         val prev = v
         v /= 10
         currentChar -= 1
-        result = (48 + (prev - v * 10)).toChar + result
+        result = (prev - v * 10).toInt.toString + result
```
gzm0: Where did the 48 + go?

SethTisue:
```
scala> 48.toChar.toString == 0.toInt.toString
res4: Boolean = true
```

junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala
sjrd: Done. Updated and rebased.

Issue #3202, opened 2017-11-26
Formatter's output for e, f and g conversions differs from the JVM in some cases


sjrd:
As reported by @ArturSoler on Gitter:
```scala
val x = 66380.788125
"%.11f".format(x)
```
prints `66380.78812500000` on the JVM but `66380.78812500001` on JS.

JS is "more right" than the JVM in the sense that the true value of `x` is indeed closer to `...01` than `...00`. However it seems that the JVM behavior is spec'ed [in the JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#dndec), which says:

> If the precision is less than the number of digits which would appear after the decimal point in the string returned by `Float.toString(float)` or `Double.toString(double)` respectively, then the value will be rounded using the round half up algorithm. Otherwise, zeros may be appended to reach the precision.

I think it would be possible for us to take that sentence into account to reproduce the behavior of the JVM.

sjrd:
This is much trickier than it looks, and it's also not the last consistency detail between the JVM and JS when it comes to formatting floating point numbers. The core functions for float-to-string conversions in the JVM and in JS all have different details in their specifications when getting close to the precision limit, and also when the exponents get really big. As far as I can tell, there is no way to build on top of JavaScript's built-in conversion functions to produce the JVM specifications. Instead, a completely correct implementation would need to perform the whole conversion from scratch.

Now, converting floats to strings is really tricky. *Papers* have been written on the topic, and the accompanying source codes are usually very long. That's usually to get top speed, which arguably we don't need here. But still, doing the conversions without intermediate loss of precision is insanely hard.

Conclusion: I do not plan to work on this issue. I'll label it as "help wanted" so that perhaps someone in the community can tackle it, but be warned: this is most definitely *not* a good-first-issue. It's hard-core stuff.

sjrd:
I generalized the title of this issue to encompass other discrepancies in the `e`, `f` and `g` conversion specifiers, which all have the same underlying issue, which I described in my previous comment.

PR #3380, opened 2018-06-15
Complete rewrite of java.util.Formatter.

### LINKED ISSUE #1619: String.format does not support thousands grouping
The following code produces different results whether run in browser vs jvm.  ScalaJS does not appear to support the thousands separator.

```
object ScalaJSExample extends js.JSApp{
  def main(): Unit = {
    println(f"${1000}%,d")    // prints 1000, not 1,000
    println(f"${1000F}%,.2f") // prints 1000.00, not 1,000.00
  }
}
```

sjrd:
This class was written a very long time ago, when quality standards were not as high as they are now. In this commit, we entirely rewrite it while aiming for:

* Readability
* Performance
* Fidelity in a lot of corner cases wrt. the specification on the JVM, in particular in error cases

The corresponding test suite is significantly enhanced.

In the process, this commit fixes #1619 (thousands grouping), as well as many other issues, among which:

* Formatting of astral characters with `%c`.
* Formatting of `null` with all the format specifiers.
* Handling of precision in non-numeric strings (basically `substring`).
* Capturing `IOException`s thrown by the target `Appendable`.

Despite the large number of fixes, the following items are still not addressed:

* #3202 and, more generally, various discrepancies between the formatting of floating point numbers wrt. the specification, with `%f`, `%e` and `%g`.
* Hexadecimal format for floating point numbers with `%a`.
* Formatting of date/times with the `%t.` family.
* Formatting of `BigInteger`s and `BigDecimal`s.

### DIFF
--- a/javalib/src/main/scala/java/util/Formatter.scala
+++ b/javalib/src/main/scala/java/util/Formatter.scala
@@ -4,285 +4,744 @@ import scala.annotation.switch
 import scala.scalajs.js
 
 import java.io._
-import java.lang._
 
-final class Formatter(private val dest: Appendable) extends Closeable with Flushable {
+final class Formatter(private[this] var dest: Appendable)
+    extends Closeable with Flushable {
+
   import Formatter._
+  import Flags._
 
-  var closed = false
+  /** If `dest == null`, the real content is in `stringOutput`.
+   *
+   *  A real `StringBuilder` may be created lazily if `out()` is called, which
+   *  will then capture the current content of `stringOutput`.
+   *
+   *  This allows to bypass the allocation of the `StringBuilder`, the call
+   *  through `dest.append()` and more importantly the `try..catch`es in the
+   *  common case where the `Formatter` is created without a specific
+   *  destination.
+   */
+  private[this] var stringOutput: String = ""
 
-  def this() = this(new StringBuilder())
+  private[this] var closed: Boolean = false
+  private[this] var lastIOException: IOException = null
+
+  def this() = this(null: Appendable)
+
+  @inline
+  private def trapIOExceptions(body: => Unit): Unit = {
+    try {
+      body
+    } catch {
+      case th: IOException =>
+        lastIOException = th
+    }
+  }
+
+  private def sendToDest(s: String): Unit = {
+    if (dest eq null)
+      stringOutput += s
+    else
+      sendToDestSlowPath(js.Array(s))
+  }
+
+  private def sendToDest(s1: String, s2: String): Unit = {
+    if (dest eq null)
+      stringOutput += s1 + s2
+    else
+      sendToDestSlowPath(js.Array(s1, s2))
+  }
+
+  private def sendToDest(s1: String, s2: String, s3: String): Unit = {
+    if (dest eq null)
+      stringOutput += s1 + s2 + s3
+    else
+      sendToDestSlowPath(js.Array(s1, s2, s3))
+  }
+
+  @noinline
+  private def sendToDestSlowPath(ss: js.Array[String]): Unit = {
+    trapIOExceptions {
+      ss.foreach(dest.append(_))
+    }
+  }
 
   def close(): Unit = {
-    if (!closed) {
+    if (!closed && (dest ne null)) {
       dest match {
-        case cl: Closeable => cl.close()
+        case cl: Closeable =>
+          trapIOExceptions {
+            cl.close()
+          }
         case _ =>
       }
     }
     closed = true
   }
 
-  def flush(): Unit = ifNotClosed {
-    dest match {
-      case fl: Flushable => fl.flush()
-      case _ =>
+  def flush(): Unit = {
+    checkNotClosed()
+    if (dest ne null) {
+      dest match {
+        case fl: Flushable =>
+          trapIOExceptions {
+            fl.flush()
+          }
+        case _ =>
+      }
     }
   }
 
-  // Begin implem of format()
+  def format(format: String, args: Array[AnyRef]): Formatter = {
+    // scalastyle:off return
 
-  def format(format_in: String, args: Array[AnyRef]): Formatter = ifNotClosed {
-    import js.JSNumberOps._
+    checkNotClosed()
 
-    var fmt: String = format_in
-    var lastImplicitIndex: Int = 0
-    var lastIndex: Int = 0 // required for < flag
+    var lastImplicitArgIndex: Int = 0
+    var lastArgIndex: Int = 0 // required for < flag
 
-    while (!fmt.isEmpty) {
-      fmt match {
-        case RegularChunk(matchResult) =>
-          fmt = fmt.substring(matchResult(0).get.length)
-          dest.append(matchResult(0).get)
+    val fmtLength = format.length
+    var fmtIndex: Int = 0
 
-        case DoublePercent(_) =>
-          fmt = fmt.substring(2)
-          dest.append('%')
-
-        case EOLChunk(_) =>
-          fmt = fmt.substring(2)
-          dest.append('\n')
-
-        case FormattedChunk(matchResult) =>
-          fmt = fmt.substring(matchResult(0).get.length)
-
-          val flags = matchResult(2).get
-          def hasFlag(flag: String) = flags.indexOf(flag) >= 0
-
-          val indexStr = matchResult(1).getOrElse("")
-          val index = if (!indexStr.isEmpty) {
-            Integer.parseInt(indexStr)
-          } else if (hasFlag("<")) {
-            lastIndex
-          } else {
-            lastImplicitIndex += 1
-            lastImplicitIndex
-          }
-          lastIndex = index
-          if (index <= 0 || index > args.length)
-            throw new MissingFormatArgumentException(matchResult(5).get)
-          val arg = args(index-1)
-
-          val widthStr = matchResult(3).getOrElse("")
-          val hasWidth = !widthStr.isEmpty
-          val width = {
-            if (hasWidth) Integer.parseInt(widthStr)
-            else if (hasFlag("-")) throw new MissingFormatWidthException(format_in)
-            else 0
-          }
-
-          val precisionStr = matchResult(4).getOrElse("")
-          val hasPrecision = !precisionStr.isEmpty
-          val precision =
-            if (hasPrecision) Integer.parseInt(precisionStr)
-            else 0
-
-          val conversion = matchResult(5).get.charAt(0)
-
-          // Avoid using conversion.isUpper not to depend on the Unicode database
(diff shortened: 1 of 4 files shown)


javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+          else if (flags.upperCase) "0X"
+          else "0x"
+        }
+        arg match {
+          case arg: Int =>
+            padAndSendToDest(flags, width, prefix,
+                applyUpperCase(flags, java.lang.Integer.toHexString(arg)))
+          case arg: Long =>
+            padAndSendToDest(flags, width, prefix,
+                applyUpperCase(flags, java.lang.Long.toHexString(arg)))
```
gzm0: Seems like it might be worth introducing a `prefix` parameter to `formatNonNumericString`?

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
-  def ioException(): IOException = null
-  def locale(): Locale = ifNotClosed { null }
-  def out(): Appendable = ifNotClosed { dest }
+  private def applyUpperCase(flags: Flags, str: String): String =
+    if (flags.upperCase) str.toUpperCase
+    else str
 
-  override def toString(): String = out().toString()
+  /** This method ignores `flags.zeroPad` and `flags.upperCase`. */
+  private def padAndSendToDest(flags: Flags, width: Int, str: String): Unit = {
```
gzm0: This is a bit magic, that one overload doesn't support `zeroPad` and the other one does. Could we at least `assert(!flags.zeroPad)`?

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+    (if (flags.positivePlus) "+" else "") +
+    (if (flags.positiveSpace) " " else "") +
+    (if (flags.zeroPad) "0" else "") +
+    (if (flags.useGroupingSeps) "," else "") +
+    (if (flags.negativeParen) "(" else "") +
+    (if (flags.useLastIndex) "<" else "")
+  }
+
+  private def computerizedScientificNotation(x: Double, precision: Int,
+      forceDecimalSep: Boolean): String = {
```
gzm0: call this `altFormat`? Otherwise the logic that `altFormat` means `forceDecimalSep` for efg is very implicit.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+      computerizedScientificNotation(x, p - 1, forceDecimalSep)
+    }
+  }
+
+  private def decimalNotation(x: Double, precision: Int,
+      forceDecimalSep: Boolean): String = {
+
+    import js.JSNumberOps._
+
+    // First use JavaScript's native toExponential conversion
```
gzm0: toFixed

javalib/src/main/scala/java/util/Formatter.scala
sjrd: I considered that, but it doesn't fit. `prefix` is the thing that goes after leading space padding, but before 0 padding. If I introduce a `prefix` in `formatNonNumericString`, then I need to take 0 padding into account in that method, and that complicates things. It was cleaner in the end to only repeat `applyUpperCase` in the `x`/`X` conversion.

javalib/src/main/scala/java/util/Formatter.scala
sjrd: I found it was more explicit to name this `forceDecimalSep` here. It is semantically related to the "notation" thing. `altFormat` is a very generic thing which applies to many different conversions, and takes on different *meanings* depending on the conversion. Once we're inside `xyzNotation`, there is a more specific meaning, with a more specific name.

I have made the `forceDecimalSep = altFormat` more explicit at call site (in `efgCommon`) to compensate.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -15,47 +15,12 @@ import org.scalajs.testsuite.utils.Platform._
...
-Double  => JDouble,
-Float   => JFloat,
-Integer => JInteger,
-Long    => JLong,
-Byte    => JByte,
-Short   => JShort,
-Boolean => JBoolean,
-String  => JString
-}
-
```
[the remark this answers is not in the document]
sjrd: Done.

javalib/src/main/scala/java/util/Formatter.scala
sjrd: We actually call it when `flags.zeroPad` is true, *intentionally* disregarding it. I have made that clearer by renaming this overload to `padAndSendToDestNoZeroPad`.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+      else "-Infinity"
+    }
+
+    padAndSendToDest(flags, width, applyUpperCase(flags, str))
+  }
+
+  private def formatNumericString(flags: Flags, width: Int, str: String): Unit = {
+
+    val TransformativeFlags =
+      PositivePlus | PositiveSpace | UseGroupingSeps | NegativeParen
```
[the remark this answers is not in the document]
sjrd: Added a comment. It's only relevant to the fast-paths of `formatNumericString`.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+      re.lastIndex = formatSpecifierIndex
+      val execResult = re.exec(format)
 
-          val flags = matchResult(2).get
-          def hasFlag(flag: String) = flags.indexOf(flag) >= 0
+      if (execResult == null || execResult.index != formatSpecifierIndex) {
+        // Could not parse a valid format specifier
+        val conversion =
+          if (formatSpecifierIndex == fmtLength) "%"
+          else format.substring(formatSpecifierIndex, formatSpecifierIndex + 1)
```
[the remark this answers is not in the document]
sjrd: JVM behavior, yes. Added a comment and explicit tests.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
-            var result: String = ""
-            var i = times
-            while (i > 0) {
-              result += s
-              i -= 1
-            }
-            result
-          }
+  private def parseFlags(flags: String, conversion: Char): Flags = {
+    var bits = if (conversion <= 'Z') UpperCase else 0
```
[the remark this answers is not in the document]
sjrd: I thought about that but I prefer it in `parseFlags`. That way, `parseFlags` contains all the logic to construct the `Flags` from various pieces of the format specifier. I don't find it strange to pass in `conversion` given that `conversion` plays a role in computing the `Flags`, at the end of the day.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
-              }
-            }
+      bits |= bit
+      i += 1
+    }
+
+    new Flags(bits)
+  }
+
+  private def parsePositiveIntSilent(capture: js.UndefOr[String],
```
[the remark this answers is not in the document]
sjrd: It has a slight size and performance impact, yes. I have added named arguments at call site.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
+      conversion: Char, invalidFlags: Int): Unit = {
+
+    @noinline def illegalFlags(): Nothing =
+      throw new IllegalFormatFlagsException(flagsToString(flags))
+
+    if ((flags.bits & invalidFlags) != 0)
+      illegalFlags()
+  }
+
+  private def flagsToString(flags: Flags): String = {
```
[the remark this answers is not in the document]
sjrd: Actually no. I have added a comment about that.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -4,285 +4,702 @@ import scala.annotation.switch
...
-          def strRepeat(s: String, times: Int) = {
-            var result: String = ""
-            var i = times
-            while (i > 0) {
-              result += s
-              i -= 1
-            }
-            result
-          }
+  private def parseFlags(flags: String, conversion: Char): Flags = {
```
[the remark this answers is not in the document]
sjrd: Same as `flagsToString`. I have added comments about that.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -67,174 +32,380 @@ class FormatterTest {
...
+
     if (!executingInJVMOnJDK6)
       expectFormatterThrows(classOf[Exception], "%#s", "hello")
   }
 
-  @Test def `should_fail_s_conversions_without_width`(): Unit = {
-    // Issue #2246
-    expectFormatterThrows(classOf[MissingFormatWidthException], "%-s", "abc")
+  @Test def sConversionWithFormattable(): Unit = {
+    assertFC("foobar", "%s", 0, -1, -1)
```
gzm0: define `assertFC` as a local method here? It is only used in this test. Maybe even define `FormattableClass` here, also inline `"foobar`" into `assertFC`?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -67,174 +32,380 @@ class FormatterTest {
...
+
+    /* We use 1.51e100 in this test, since we seem to have a floating point
+     * imprecision at exactly 1.5e100 that yields to a rounding error towards
+     * 1e+100 instead of 2e+100.
+     */
+    assertF("2e+100", "%.0e", 1.51e100)
+
+    assertF(" 1.20e+100", "%10.2e", 1.2e100)
+
+    // Corner case of round-half-up (currently non-compliant)
```
gzm0: Should we track this as a bug?

sjrd: I explicitly generalized #3202 for this set of issues.

PR #3413, opened 2018-07-21
JSEnv fixes

gzm0:


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/JSEnvSuiteConfig.scala
```diff
@@ -26,42 +26,43 @@ import scala.concurrent.duration._
...
-    copy(terminateVMJSCode = Some(code))
-
   def withSupportsCom(supportsCom: Boolean): JSEnvSuiteConfig =
     copy(supportsCom = supportsCom)
 
   def withSupportsTimeout(supportsTimeout: Boolean): JSEnvSuiteConfig =
     copy(supportsTimeout = supportsTimeout)
 
+  def withSupportsExit(supportsExit: Boolean): JSEnvSuiteConfig =
+    copy(supportsExit = supportsExit)
```
sjrd: Any reason not to put that method where `withTerminateVMJSCode` was? That would keep the ordering so that all fields are always in the same order.

gzm0: Nope, fluke. Good catch.

gzm0:
Updated.

PR #3418, opened 2018-07-24
[no-master] Fix #3417: Handle Windows file separator in VirtualFile.nameFromPath.

### LINKED ISSUE #3417: Crash in jsDependencies resolution on Windows due to some path issue
While trying to test my scala-js application, I was facing this issue where tests were running on Mac but not on windows. The issue was happening because of the `jsDependencies` resolution, while materializing  js files on windows system, the path was not correct.

I tried to reproduce the same with a minimal example where my scala-js app depends on:

     `libraryDependencies += "superfoo" %%% "foo" % "0.1-SNAPSHOT"`

This dummy module `foo` has `moment.js` as a `jsDependency`. I executed tests for the application on windows as well as on Mac.

**Windows** - Didn't work!
-------

```
VirtualFileMaterializer :: path => C:\Users\VIK\.ivy2\local\superfoo\foo_sjs0.6_2.12\0.1-SNAPSHOT\jars\foo_sjs0.6_2.12.jar:moment.js
                           name => C:\Users\VIK\.ivy2\local\superfoo\foo_sjs0.6_2.12\0.1-SNAPSHOT\jars\foo_sjs0.6_2.12.jar:moment.js
```
    Results in Exception:
    `C:\Users\VIK\AppData\Local\Temp\1532419615932-0\C:\Users\VIK\.ivy2\local\superfoo\foo_sjs0.6_2.12\0.1-SNAPSHOT\jars\foo_sjs0.6_2.12.jar:moment.js (The filename, directory name, or volume label syntax is incorrect)`
    
    Cuz it searches for 
           `C:\Users\VIK\AppData\Local\Temp\1532419615932-0\foo_sjs0.6_2.12.jar:moment.js`
    but it gets
           `C:\Users\VIK\AppData\Local\Temp\1532419615932-0\C:\Users\VIK\.ivy2\local\superfoo\foo_sjs0.6_2.12\0.1-SNAPSHOT\jars\foo_sjs0.6_2.12.jar:moment.js`

                           
Mac - Works!
---
```
VirtualFileMaterializer :: path => /Users/vika/.ivy2/local/superfoo/foo_sjs0.6_2.12/0.1-SNAPSHOT/jars/foo_sjs0.6_2.12.jar:moment.js
                           name => foo_sjs0.6_2.12.jar:moment.js
                           
```
                           
The problem is this function in 
https://github.com/scala-js/scala-js/blob/v0.6.24/tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala#L47
```

def nameFromPath(path: String): String = {
    val pos = path.lastIndexOf('/')
    if (pos == -1) path
    else path.substring(pos + 1)
  }
```

vi-kas:
Fixes issue [3417](https://github.com/scala-js/scala-js/issues/3417) 

See issue link for description.
Fixes a path issue which was not os specific. The issue was happening because of the `jsDependencies` resolution, while materializing js files on windows system, the path was not correct.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
@@ -45,7 +45,10 @@ trait VirtualFile {
 object VirtualFile {
   /** Splits at the last slash and returns remainder */
   def nameFromPath(path: String): String = {
-    val pos = path.lastIndexOf('/')
+    val pos0 = path.lastIndexOf('/')
+    val pos =
+      if (pos0 >= 0) pos0
+      else path.lastIndexOf('\\')
     if (pos == -1) path
     else path.substring(pos + 1)
   }


tools/shared/src/main/scala/org/scalajs/core/tools/io/VirtualFiles.scala
```diff
@@ -45,7 +45,10 @@ trait VirtualFile {
 object VirtualFile {
   /** Splits at the last slash and returns remainder */
   def nameFromPath(path: String): String = {
-    val pos = path.lastIndexOf('/')
+    val osName = System.getProperty("os.name")
```
sjrd: You cannot use `System.property("os.name")` here, because the Tools API is cross-compiled, and that will not work in the JS version. You can see that "bootstrap" tests are failing on the CI because of this. Besides, even on Windows, the `path` would be using forward slashes in many situations, so your fix could break other use cases.

What we typically do instead is to handle *both* `\` and `/`, on all platforms. One way to do it here would be
```scala
val pos = Math.max(path.lastIndexOf('/'), path.lastIndexOf('\\'))
```
or, to avoid the second call when it's not necessary:
```scala
val pos0 = path.lastIndexOf('/')
val pos =
  if (pos0 >= 0) pos0
  else path.lastIndexOf('\\')
```

sjrd:
Is there an existing test in this repo that I can run on Windows and that would reproduce the bug? I wonder why typical bootstrap tests don't fail, for example (I can run bootstrap tests on Windows all right).

vi-kas:
@sjrd Thanks, I've refactored and tested following your comments.

sjrd:
Thanks. The code and commit look good. Do you have an answer to my question at https://github.com/scala-js/scala-js/pull/3418#issuecomment-407391200 ?

vi-kas:
No not in **this** repo, do you mean any dummy project to reproduce the bug?

sjrd:
I meant in this repo. So that there is a least one test I could execute on Windows that would convince me that I don't break that feature later, for example. Something we would put in our CI script, if we had one for Windows in 0.6.x.

vi-kas:
Nope, not that I know of.

vi-kas:
Yep, Thank you.

PR #3424, opened 2018-08-12
Fix #3420: Make timeout tests less restrictive

### LINKED ISSUE #3420: Timeout tests have too restrictive timing
Originally they were written to test our own implementation of `setTimeout` for Rhino. Now they are mostly a means to run things after returning to the event loop once. We should rewrite the tests with that in mind.

gzm0:
Originally they were written to test our own implementation of
`setTimeout` for Rhino. Now they are just a means to run things
after returning to the event loop once, so we can reduce them
significantly.

### DIFF
--- a/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutRunTests.scala
+++ b/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutRunTests.scala
@@ -39,109 +39,21 @@ private[test] class TimeoutRunTests(config: JSEnvSuiteConfig, withCom: Boolean) 
   }
 
   @Test
-  def clearTimeoutTest: Unit = {
-
-    val deadline = 300.millis.fromNow
-
-    """
-    var c = setTimeout(function() { console.log("1"); }, 200);
-    setTimeout(function() {
-      console.log("2");
-      clearTimeout(c);
-    }, 100);
-    setTimeout(function() { console.log("3"); }, 300);
-    setTimeout(function() { console.log("4"); },   0);
-    """ hasOutput
-    """|4
-       |2
-       |3
-       |""".stripMargin
-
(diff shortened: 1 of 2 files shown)


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutRunTests.scala
```diff
@@ -38,110 +38,21 @@ private[test] class TimeoutRunTests(config: JSEnvSuiteConfig, withCom: Boolean)
...
-    var i2 = setInterval(function() { console.log("each 3100"); }, 3100);
-    var i3 = setInterval(function() { console.log("each 1300"); }, 1300);
-
-    setTimeout(function() {
-      clearInterval(i1);
-      clearInterval(i2);
-      clearInterval(i3);
-    }, 10000);
+    setInterval(function() { console.log("tick"); }, 20);
     """ hasOutput
```
sjrd: I'm totally confused by the new version of this test. How does it stop after 5 occurrences of the interval? The previous version explicitly called `clearInterval` after some time, which explained the behavior. But now, what makes it stop?

gzm0: The kit makes it stop :) We cannot rely anymore on `JSEnvs` terminating. So once we have seen the output we need, we terminate it.

sjrd: Oh! Maybe add a comment to remind the reader about that in this specific test? Because in this particular case it's non-obvious why the test is even correct if the reader doesn't have that property in mind. (In other tests I've read, it's not a problem for understanding the test.)

gzm0:
Updated. I also fixed the ComTimeoutTests that I forgot.

PR #3425, opened 2018-08-13
Fix #3422: Support nulling-out lazy vals in non-native JS classes.

### LINKED ISSUE #3422: NPE during compilation of a nulling-out `lazy val` in a non-native JS class
```
[error] java.lang.NullPointerException
[error] 	at scala.reflect.internal.Definitions$DefinitionsClass.isPrimitiveValueType(Definitions.scala:1378)
[error] 	at org.scalajs.core.compiler.GenJSCode$JSCodePhase.ensureBoxed(GenJSCode.scala:3979)
[error] 	at org.scalajs.core.compiler.GenJSCode$JSCodePhase.genBoxedRhs$1(GenJSCode.scala:2012)
[error] 	at org.scalajs.core.compiler.GenJSCode$JSCodePhase.genStatOrExpr(GenJSCode.scala:2021)
...
```

To reproduce:
1. checkout https://github.com/benhutchison/metadoc
2. run `js/compile`

sjrd:
We work around the `rhs.tpe eq null` caused by scalac by testing explicitly for this case, when the rhs is a literal `null`. It is decidedly ugly, but it gets the job done.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2009,8 +2009,17 @@ abstract class GenJSCode extends plugins.PluginComponent
               val genQual = genExpr(qualifier)
 
               def genBoxedRhs: js.Tree = {
-                ensureBoxed(genRhs,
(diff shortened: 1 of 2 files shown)


sjrd:
Superseded by #3427.

PR #3427, opened 2018-08-13
Fix #3426: Fully support lazy vals in non-native JS classes.

### LINKED ISSUE #3426: lazy vals in non-native JS classes are accepted by the compiler, but do not work
I don't know whether they should
* be fixed, or
* be rejected by the compiler

but in any case they do not work:

* In 2.10 and 2.11, they always evaluate to `null` (or the zero of their type), without ever be initialized
* In 2.12+, they work if accessed from Scala.js code, but do not work if accessed from JavaScript code (again, not initialized).

Discovered in the CI of #3425 for 2.10 and 2.11, and then I realized they were also wrong in 2.12 if accessed from JS.

sjrd:
This turned out to be relatively simple. There is only a little dance to correctly consider lazy *fields* as non-exposed, but the lazy *def* that accesses it as exposed, both in 2.11- and 2.12+.

The commit also includes a better error message when trying to include a `lazy val` in a JS trait. The code previously failed to compile as well but with an obscure error message.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -2009,8 +2009,18 @@ abstract class GenJSCode extends plugins.PluginComponent
               val genQual = genExpr(qualifier)
 
               def genBoxedRhs: js.Tree = {
-                ensureBoxed(genRhs,
-                    enteringPhase(currentRun.posterasurePhase)(rhs.tpe))
+                val tpeEnteringPosterasure =
+                  enteringPhase(currentRun.posterasurePhase)(rhs.tpe)
+                if ((tpeEnteringPosterasure eq null) && genRhs.isInstanceOf[js.Null]) {
+                  // 2.10.x does not yet have `devWarning`, so use `debugwarn` instead.
+                  debugwarn(
+                      "Working around https://github.com/scala-js/scala-js/issues/3422 " +
+                      s"for ${sym.fullName} at ${sym.pos}")
+                  // Fortunately, a literal `null` never needs to be boxed
+                  genRhs
+                } else {
+                  ensureBoxed(genRhs, tpeEnteringPosterasure)
+                }
               }
 
               if (isScalaJSDefinedJSClass(sym.owner)) {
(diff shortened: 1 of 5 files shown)


sjrd:
@gzm0 I realized the first (most important) fix contains a potential issue in terms of backward binary compatibility. The problematic scenario is the following:

* Using 2.12+ (with 2.10 or 2.11 the whole thing would not have worked to begin with)
* A library `X` compiled with 0.6.24 or earlier defines a JS class `A` with a `lazy val x`
* A codebase `Y` compiled with 0.6.25 or later accesses `a.x` of some `a: A` from the library `X`

In that case, the access in `Y` will compile to `a["x"]`, but `A` does not define the proper getter for `"x"` (it only defines a field) and hence `Y` will see an uninitialized value. If `Y` were compiled with 0.6.24 or earlier, it would bug-compatibly call the Scala.js-specific internal getter.

If we strictly follow our binary compat policy, we cannot merge this change in 0.6.x. We can only fix those issues in 1.x.

Opinion?

sjrd:
There might be a way to hack around the problem for bincompat's sake:

* When *accessing* a `lazy val` of a JS class, pretend it wasn't exposed.

This would force call sites written in Scala.js to delegate to the internal getter, which exists regardless of whether the JS class was compiled by 0.6.24- or 0.6.25+.

Problem: this does not honor `override lazy val`s correctly. It will still do the current behavior which is bonkers: https://scalafiddle.io/sf/ujFBC2U/0

sjrd:
> The do not use a lazy val.

If only ... but the very existence of the bug report #3422 demonstrates that people *do* use `lazy val`s in JS classes. This may happen when they need the JS class for other reasons, but that the `lazy val` itself is only used within Scala.js.

> Further, silently making overriding fail (and behave differently from Scala and JS!) seems a bigger problem than getting an UndefinedBehaviorError (or maybe () if that is allowed in the return type).

It's not a UBE. It's a zero of the type of the `lazy val`, so `0` for an `Int`, `null` for reference types, etc. The field is initialized to the zero of its type; it's the rhs of the `lazy val` which is disregarded.

> Could we protect against this in the linker? Notably once we have private methods?

You mean refuse to link accesses to private methods from a different class? We might be able to, but it could be expensive. We could do it in the IR checking, maybe.

> Maybe we should reach out to the community for this to make sure nobody is upset if we break the binary compat promise a bit?

We can do that. It's worth noting that would set a precedent. So far we *never* broke binary compat of SJS artifacts.

gzm0:
Ah, right. Ok in that case we should keep the broken overrides. Maybe we can warn if a lazy val is overridden in a SJS defined class and/or if a lazy val is not final? (Not in this PR but in this release).

PR #3430, opened 2018-08-18
Deprecate org.scalajs.testinterface.TestUtils and ScalaJSClassLoader.

sjrd:
In favor of

* `scala.scalajs.reflect.Reflect` for JS-only code, or
* `portable-scala-reflect` for portable code.


test-interface/src/main/scala/org/scalajs/testinterface/ScalaJSClassLoader.scala
```diff
@@ -13,9 +13,13 @@ import java.util.Enumeration
 final class ScalaJSClassLoader(
```
gzm0: Shouldn't this also be deprecated? (basically everything public in `testinterface` should be deprecated, IIUC)

sjrd: It cannot be deprecated in 0.6.x because it's possible that someone would still be relying on dynamic instantiation through exports. In that case, they would still have to pattern-match on `ScalaJSClassLoader` in 0.6.x. In 1.x however, it will disappear.

It's not nice because something non-deprecated will disappear, but direct uses of `ScalaJSClassLoader` in the wild are most likely super rare, if they exist at all.

Or we could just deprecate it anyway, and let people live with the deprecation ...

gzm0: But that dynamic instantiation is precisely what we deprecate (and provide an alternative for). So they should get a deprecation warning.

sjrd: OK.

I couldn't actually deprecate `ScalaJSClassLoader` itself, because we need to instantiate it in a few places in the test interface (we can't instantiate a private copy, because that would break existing clients that match against `ScalaJSClassLoader`). So instead I deprecated its only meaningful member `val namespace`, and I added a notice of deprecation in the Scaladoc.

PR #3434, opened 2018-08-23
Support Scala 2.13.0-M5.

sjrd:
There is one non-trivial change: `js.ArrayOps` had to be completely reimplemented for 2.13.0-M5, because the collection framework does not accept collections classes whose `repr` is not themselves anymore. This means that `js.ArrayOps` must be separated from the collection *hierarchy* per se, and instead be an independent `AnyVal` that happens to implement the entire same API. This is a lot of code, but is similar to what `scala.collection.ArrayOps` does for `scala.Array` (only that we have the API of `Buffer` to support, in addition of just `Seq`).

Ideally, all those methods should receive dedicated tests now, since their implementation is not inherited from the collections. However, this commit does not go to such lengths.

The PR also includes
* one commit to implement the `addSuppressed` mechanism of `Throwable` from JDK 7, which is now reached by the new `Future` implementation of 2.13.0-M5
* a few preparatory commits for overrides and similar


sjrd:
The first commit should be amended to include unit tests. I'm already opening the PR so that you can start reviewing it, should you have time to do so before I get to writing them.

I have locally tested with `++2.13.0-pre-53c4be1` and `testSuite/test`. That version [is supposed to become M5](https://contributors.scala-lang.org/t/2-13-0-m5-release-train/2125/12?u=sjrd) in the very short future (we're talking days).

sjrd:
Erm ... your comments (and my answers) disappeared. I think you commented on the commit itself, rather than via the PR view of the commit, so GitHub did not transfer the comments to the amended commits.

I tried pushing the recovered commits from the `reflog` to https://github.com/sjrd/scala-js/tree/recover-the-comments, but it doesn't seem like the comments are there :-(

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+    // scalastyle:off return
+    val len = xs.length
+    var i = from
+    while (i < len) {
+      if (elem == xs(i))
+        return i
+      i += 1
+    }
+    -1
+    // scalastyle:on return
```
sjrd: From my mails: @gzm0 says:

> Why not a tailrec method instead? Should compile to roughly the same code.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+
+    def hasNext: Boolean = pos < xs.length
+
+    def next(): js.Array[A] = {
+      if (pos >= xs.length)
+        throw new NoSuchElementException
+      val r = new js.ArrayOps(xs).slice(pos, pos + groupSize)
+      pos += groupSize
+      r
+    }
```
sjrd: From my mails: @gzm0 says:

> Two things in general here:
>
> * I do not understand when something is marked inline and when it is not (e.g. iterator is not, but should?).
> * How do we ensure that new methods on collections also appear here?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
sjrd: Because the copied from [`scala.collection.ArrayOps.scala`](https://github.com/scala/scala/blob/008d525314ca55b41652c79fb23ba1c9cf78a133/src/library/scala/collection/ArrayOps.scala#L499-L506).

In general, virtually all the methods are copied from their implementation in upstream scala/scala.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
sjrd: > I do not understand when something is marked inline and when it is not (e.g. iterator is not, but should?).

A method is `@inline` if at least one of the following applies:

* It was `@inline` in its implementation upstream
* It delegates to a JS operation
* It is of the form `{ someMethodCall(...); xs }`

Arguably, the last two bullets should be integrated as heuristics in the optimizer, but that's for another day.

> How do we ensure that new methods on collections also appear here?

We do as they do for `scala.collection.ArrayOps`: we monitor the `Seq`, `Buffer`, `Growable` and `Shrinkable` APIs, and we replicate the changes. It's sad, but the new collection design does not seem to allow anything else.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+
+  // A helper for tails and inits.
+  private[this] def iterateUntilEmpty(
+      f: js.Array[A] => js.Array[A]): scala.collection.Iterator[js.Array[A]] = {
+    scala.collection.Iterator.iterate(xs)(f).takeWhile(_.nonEmpty) ++
+    scala.collection.Iterator.single(js.Array[A]())
+  }
+
+  /** An array containing the first `n` elements of this array. */
+  def take(n: Int): js.Array[A] =
```
gzm0: Inline

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+    xs.jsSlice(0, max(n, 0))
+
+  /** The rest of the array without its `n` first elements. */
+  def drop(n: Int): js.Array[A] =
```
gzm0: Inline

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+   *    the longest suffix of this array whose first element does not satisfy
+   *    the predicate `p`.
+   */
+  def dropWhile(p: A => Boolean): js.Array[A] = {
+    val i = indexWhere(x => !p(x))
+    val lo = if (i < 0) xs.length else i
+    xs.jsSlice(lo, xs.length)
+  }
+
+  def iterator: scala.collection.Iterator[A] =
```
gzm0: Inline?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+   *  @see [[scala.collection.Iterator]], method `grouped`
+   *
+   *  @param size
+   *    the number of elements per group
+   *  @return
+   *    An iterator producing arrays of size `size`, except the last will be
+   *    less than size `size` if the elements don't divide evenly.
+   */
+  def grouped(size: Int): scala.collection.Iterator[js.Array[A]] =
+    new js.ArrayOps.GroupedIterator[A](xs, size)
```
gzm0: Inline?

sjrd:
> I have left a bunch of comments where I do not understand (even based on your explanation) why there are no inlines.

Yes I missed those. I've added all the relevant `@inline`s.

> I also don't really know how to review the overrides, as a lot of it is just copying.
> Also since all implementations are from scala/scala it seems to make little sense to fully review them as well.

Sure, makes sense.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,2002 @@
...
+   *  @see [[scala.math.Ordering]]
+   *
+   *  @param ord
+   *    the ordering to be used to compare elements.
+   *  @return
+   *    an array consisting of the elements of this array sorted according to
+   *    the ordering `ord`.
+   */
+  def sorted[B >: A](implicit ord: Ordering[B]): js.Array[A] =
+    new js.WrappedArray(xs).sorted(ord.asInstanceOf[Ordering[A]]).array
```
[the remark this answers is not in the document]
sjrd: Actually `WrappedArray` is already `@inline`, so we would just need to delegate to `WrappedArray`.

However, even with our kick-ass optimizer, the implementations inherited from the collections into `WrappedArray` are not as good as those we can directly write on `js.Array`, which means that `js.ArrayOps` would not be on par with `scala.collection.ArrayOps` if we do that. I did it for `sorted` because the original delegates to `java.util.Arrays.sort` anyway, which in Scala.js goes back to the collections (and also, it's really non-trivial code!).

sjrd:
Addressed the other comments. And also I had the courage to write tests until `distinct` (excluded).

sjrd:
> Also: If we introduce ArrayOps like this, we also should introduce DictionaryOps (which we didn't have because the repr restriction always existed for Map IIRC).

We should do it eventually, but it is not necessary right now. In this PR I'm only aiming for parity in the support of 2.13.0-M5 wrt. the support of 2.12.x and 2.13.0-M4.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
+    xs
+  }
 
-  // Scala notation for a fast concat()
+  def sliceInPlace(start: Int, end: Int): js.Array[A] =
+    takeInPlace(end).dropInPlace(start)
 
-  @inline def ++[B >: A](that: js.Array[_ <: B]): js.Array[B] =
-    ArrayOpsCommon.concat(array, that)
+  @inline private def normalized(n: Int): Int =
```
gzm0: call this `clampIndex`?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
+  def trimStart(n: Int): Unit =
+    xs.splice(0, normalized(n))
+
+  /** Removes the last ''n'' elements of this array.
+   *
+   *  @param n
+   *    the number of elements to remove from the end of this array.
+   */
+  def trimEnd(n: Int): Unit =
+    xs.length = xs.length - normalized(n)
```
gzm0: xs.length -= normalized(n)?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
+
+  @inline
+  def patchInPlace(from: Int, patch: scala.collection.Seq[A],
+      replaced: Int): js.Array[A] = {
+    new js.WrappedArray(xs).patchInPlace(from, patch, replaced)
```
gzm0: Couldn't this just be `xs.splice(from, replaced, patch: _*)`?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
+    xs
+  }
+
+  def filterInPlace(p: A => Boolean): js.Array[A] = {
+    var i = 0
+    var j = 0
+    val len = xs.length
+    while (i < len) {
+      if (p(xs(i))) {
+        if (i != j) {
```
gzm0: Is this necessary for efficiency? IMHO it hurts readability.

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
   }
 
 }
 
-private object ArrayOps {
-  private object ArrayFactory extends SeqFactory[js.Array] {
-    @inline
-    def empty[A]: js.Array[A] = js.Array()
+object ArrayOps {
+  @SerialVersionUID(3L)
```
gzm0: Why do we need this?

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
 
-    @inline
-    def from[A](source: IterableOnce[A]): js.Array[A] =
-      (newBuilder[A] ++= source).result()
+    @inline def apply(n: Int): A = xs(n)
 
-    @inline
-    def newBuilder[A]: mutable.Builder[A, js.Array[A]] =
-      new ArrayOps[A]
+    override protected[this] def className = "ArrayView"
```
gzm0: Shouldn't this return a JS specific prefix? Otherwise `toString` will not be distinguishable from a `scala.Array` view (I suppose).

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
```diff
@@ -8,97 +8,1998 @@
...
+
+    def hasNext: Boolean = pos < xs.length
+
+    def next(): A = {
+      try {
+        val r = xs(pos)
+        pos += 1
+        r
+      } catch {
+        case _: ArrayIndexOutOfBoundsException =>
```
gzm0: This does not work IIUC (and below)

library/src/main/scala-new-collections/scala/scalajs/js/ArrayOps.scala
gzm0: Another thing in general: There are a lot of while loops in this file to replace `foreach` on `js.Array[_]`. From what I have understood, our optimizer should be able to inline `foreach` completely. Shouldn't we use the higher level construct then?

sjrd:
Updated. I have addressed your comments except the fact that we should use `foreach` everywhere. I'll do that tomorrow if I get the chance. I have also added many more tests. Only a few methods are still missing their tests.

sjrd:
OK done. I have used `foreach` wherever possible. There are still a number of `while` loops because we often need the indices.

I have also finished the tests.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -123,7 +123,6 @@ abstract class GenJSCode extends plugins.PluginComponent
 
     override def name: String = phaseName
     override def description: String = GenJSCode.this.description
-    override def erasedTypes: Boolean = true
```
sjrd: I had to add this change to support the new candidate for 2.13.0-M5, i.e., 2.13.0-pre-8a52aa1. It turns out that the inherited value of `erasedTypes` is already always `true` (across all versions of Scala we support) so this line has always been redundant.

sjrd:
Now that 2.13.0-M5 is on Maven Central, I added CI runs for that version in `Jenkinsfile`.

library/src/main/scala-new-collections/scala/scalajs/js/Any.scala
```diff
@@ -181,13 +201,15 @@ object Any extends LowPrioAnyImplicits {
...
+    new js.WrappedArray(array)
+  implicit def wrapDictionary[A](dict: js.Dictionary[A]): js.WrappedDictionary[A] =
+    new js.WrappedDictionary(dict)
 }
 
 sealed trait LowestPrioAnyImplicits {
-  implicit def iterableOps[A](iterable: Iterable[A]): IterableOps[A] =
-    new IterableOps(iterable)
+  implicit def arrayAsIterable[A](array: js.Array[_ <: A]): scala.collection.Iterable[A] =
+    new js.WrappedArray(array)
```
sjrd: And I had to add this new implicit to take precedence over `iterableOps`, in case the expected type is already known to be `scala.collection.Iterable[A]` but the type of the array is a `js.Array[B]` with `B <: A`. This may be due to subtle changes in type inference in Scala 2.13.0-M5.

gzm0:
I have given this another quick look and from what I see it looks fine. Unfortunately this review is not as proper as typically, but given the time pressure, I guess we have to accept that.

Do you mind me asking what the current process, especially in terms of timeline is for Scala milestones. It seems we are having a quite serious problem in there somewhere, since we always end up rushing to keep up with their milestones. This is not good, as it degrades review quality (like this PR). We should make sure we address this somehow.

sjrd:
> Do you mind me asking what the current process, especially in terms of timeline is for Scala milestones. It seems we are having a quite serious problem in there somewhere, since we always end up rushing to keep up with their milestones. This is not good, as it degrades review quality (like this PR). We should make sure we address this somehow.

The problem is the extremely fast pace at which *interfaces* have been shifting in 2.13.x. This many changes in this little time have never happened in the past 5 years. We used to be able to back-publish *several* versions of Scala.js for new milestones of Scala. The rework of the collections in 2.13 has completely destroyed our approach to this issue.

I sincerely hope that things settle down now that M5 is the feature freeze. We shouldn't have to deal with such pressure from now on.

PR #3437, opened 2018-08-26
handle incomplete multibyte sequence for UTF-8

### LINKED ISSUE #3433: UTF-8 CharsetDecoder doesn't handle incomplete character ranges correctly
Trying to consume a bytebuffer with an incomplete character at the end of a buffer should result in rejecting the buffer and returning underflow, rather than handling it as an invalid byte sequence.

As seen on https://scalafiddle.io/sf/WsaQcxb/0

I have not minimized whether the issue is in CharsetDecoder's `decode` method, or in the `decodeLoop` method of the UTF-8 decoder.

martijnhoekstra:
make it return underflow instead of coding error, since there may
still be more input

Fixes #3433

### DIFF
--- a/library/src/main/scala/scala/scalajs/niocharset/UTF_8.scala
+++ b/library/src/main/scala/scala/scalajs/niocharset/UTF_8.scala
@@ -116,17 +116,33 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
             if (length == -1) {
               finalize(CoderResult.malformedForLength(1))
             } else {
-              val decoded = {
-                @inline
-                def inArrayOr0(offset: Int): Int =
-                  if (inPos + offset < inEnd) inArray(inPos + offset)
-                  else 0 // 0 is not a valid next byte
-
-                val b2 = inArrayOr0(1)
-                if (length == 2) decode2(leading, b2)
-                else if (length == 3) decode3(leading, b2, inArrayOr0(2))
-                else decode4(leading, b2, inArrayOr0(2), inArrayOr0(3))
-              }
+              val decoded =
+                if (inPos + 1 >= inEnd)
+                  DecodedMultiByte(CoderResult.UNDERFLOW)
+                else {
+                  val b2 = inArray(inPos + 1)
+                  if (isInvalidNextByte(b2))
+                    DecodedMultiByte(CoderResult.malformedForLength(1))
+                  else if (length == 2)
+                    decode2(leading, b2)
+                  else if (inPos + 2 >= inEnd)
+                    DecodedMultiByte(CoderResult.UNDERFLOW)
+                  else {
+                    val b3 = inArray(inPos + 2)
+                    if (isInvalidNextByte(b3))
+                      DecodedMultiByte(CoderResult.malformedForLength(2))
+                    else if (length == 3)
+                      decode3(leading, b2, b3)
+                    else if (inPos + 3 >= inEnd)
+                      DecodedMultiByte(CoderResult.UNDERFLOW)
+                    else {
+                      val b4 = inArray(inPos + 3)
+                      if (isInvalidNextByte(b4))
+                        DecodedMultiByte(CoderResult.malformedForLength(3))
+                      else decode4(leading, b2, b3, b4)
+                    }
+                  }
+                }
 
               if (decoded.failure != null) {
                 finalize(decoded.failure)
@@ -160,9 +176,12 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
       @inline
       @tailrec
       def loop(): CoderResult = {
+        //mark input position so we can reset on multi-byte failure
+        val startPosition = in.position
+
         @inline
-        def finalize(read: Int, result: CoderResult): CoderResult = {
-          in.position(in.position() - read)
+        def fail(result: CoderResult): CoderResult = {
+          in.position(startPosition)
           result
         }
 
@@ -173,7 +192,7 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
           if (leading >= 0) {
             // US-ASCII repertoire
             if (!out.hasRemaining) {
-              finalize(1, CoderResult.OVERFLOW)
+              fail(CoderResult.OVERFLOW)
             } else {
               out.put(leading.toChar)
               loop()
@@ -182,27 +201,35 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
             // Multi-byte
             val length = lengthByLeading(leading & 0x7f)
             if (length == -1) {
-              finalize(1, CoderResult.malformedForLength(1))
+              fail(CoderResult.malformedForLength(1))
             } else {
-              var bytesRead: Int = 1
-
               val decoded = {
-                @inline
-                def getOr0(): Int =
-                  if (in.hasRemaining) { bytesRead += 1; in.get() }
-                  else 0 // 0 is not a valid next byte
-
-                if (length == 2) decode2(leading, getOr0())
-                else if (length == 3) decode3(leading, getOr0(), getOr0())
-                else decode4(leading, getOr0(), getOr0(), getOr0())
+                if (in.hasRemaining) {
+                  val b2 = in.get()
+                  if (isInvalidNextByte(b2)) DecodedMultiByte(CoderResult.malformedForLength(1))
+                  else if (length == 2) decode2(leading, b2)
+                  else if (in.hasRemaining) {
+                    val b3 = in.get()
+                    if (isInvalidNextByte(b3)) DecodedMultiByte(CoderResult.malformedForLength(2))
+                    else if (length == 3) decode3(leading, b2, b3)
+                    else if (in.hasRemaining) {
+                      val b4 = in.get()
+                      if (isInvalidNextByte(b4)) DecodedMultiByte(CoderResult.malformedForLength(3))
+                      else decode4(leading, b2, b3, b4)
+                    }
+                    else DecodedMultiByte(CoderResult.UNDERFLOW)
+                  }
+                  else DecodedMultiByte(CoderResult.UNDERFLOW)
+                }
+                else DecodedMultiByte(CoderResult.UNDERFLOW)
               }
 
               if (decoded.failure != null) {
-                finalize(bytesRead, decoded.failure)
+                fail(decoded.failure)
               } else if (decoded.low == 0) {
                 // not a surrogate pair
                 if (!out.hasRemaining)
-                  finalize(bytesRead, CoderResult.OVERFLOW)
+                  fail(CoderResult.OVERFLOW)
                 else {
                   out.put(decoded.high)
                   loop()
@@ -210,7 +237,7 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
               } else {
                 // a surrogate pair
                 if (out.remaining < 2)
-                  finalize(bytesRead, CoderResult.OVERFLOW)
+                  fail(CoderResult.OVERFLOW)
                 else {
                   out.put(decoded.high)
                   out.put(decoded.low)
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/nio/charset/UTF_8.scala
```diff
@@ -106,17 +106,29 @@ private[charset] object UTF_8 extends Charset("UTF-8", Array(
...
+                  else {
+                    val b3 = inArray(inPos + 2)
+                    if (isInvalidNextByte(b3))
+                      DecodedMultiByte(CoderResult.malformedForLength(2))
+                    else if (length == 3)
+                      decode3(leading, b2, b3)
+                    else if (inPos + 3 >= inEnd)
+                      DecodedMultiByte(CoderResult.UNDERFLOW)
+                    else
+                      decode4(leading, b2, b3, inArray(inPos + 3))
```
martijnhoekstra: it might be nice to handle the last byte consistently with the rest

maybe then it's also a good idea to make decodeN assume correct continuation bytes so it doesn't need to do work validating that anymore

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/UTF8Test.scala
```diff
@@ -16,26 +16,36 @@ import org.junit.Assert._
 import BaseCharsetTest._
 
 class UTF8Test extends BaseCharsetTest(Charset.forName("UTF-8")) {
-  @Test def decode(): Unit = {
-    def OutSeq(elems: OutPart[CharBuffer]*): Seq[OutPart[CharBuffer]] =
-      Seq[OutPart[CharBuffer]](elems: _*)
-
+  @Test def decode1byte(): Unit = {
```
martijnhoekstra: I grouped up the decode tests, in order to more easily identify test-cases. Happy to undo again if that's your preference, though this might be easier for the next person coming through.

sjrd:
Thanks.

You should target the 0.6.x branch for this PR, otherwise it will never be available in 0.6.x releases.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
```diff
@@ -30,6 +30,57 @@ class BaseCharsetTest(val charset: Charset) {
   protected def testDecode(in: ByteBuffer)(
       outParts: OutPart[CharBuffer]*): Unit = {
 
+    def testDecodeVsSteppedDecode(malformedAction: CodingErrorAction,
+        unmappableAction: CodingErrorAction) = {
+
+      val in1 = in.duplicate
+      val in2 = in.duplicate
```
sjrd: You should take `readOnly` into account, the same way it is done [in `testOneConfig`](https://github.com/scala-js/scala-js/blob/ed20d0dac5d92f4a84edf5837054bae70aecb021/test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala#L40-L44) (so you should probably abstract that away in a `def duplicateIn(in, readOnly)` in order not to repeat it 3 times).

This is important to check that both code paths with `hasArray == true` and `hasArray == false` are handled by the charset decoders. In particular, for UTF-8, this means `decodeLoopArray` and `decodeLoopNoArray`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
```diff
@@ -30,6 +30,57 @@ class BaseCharsetTest(val charset: Charset) {
...
+
+      in2.limit(in2.position)
+
+      val d1 = charset.newDecoder()
+      d1.onMalformedInput(malformedAction)
+      d1.onUnmappableCharacter(unmappableAction)
```
sjrd: You'll probably want to abstract away this 3-line incantation in a separate `def newDecoder(malformedAction, unmappableAction)` inside `testDecode` (or even as `private def` in `BaseCharsetTest`). This incantation is now repeated 3 times (twice in your new code, and once in `testOneConfig`).

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
```diff
@@ -30,6 +30,57 @@ class BaseCharsetTest(val charset: Charset) {
...
+
+
+      val r1 = Try(d1.decode(in1))
+      val out2 = CharBuffer.allocate(in.capacity * 2)
```
sjrd: I suggest clearly grouping the things used to compute `r1` on the one hand, and the things used to compute `r2` on the other hand. At the moment they are interleaved for no obvious reason, so it's harder to understand what `r1` and `r2` represent, respectively.

Even better, put everything used to compute `r1` in the `Try { ... }` for `r1`, and respectively for `r2`. AFAICT they're totally independent, so variables should not be needed in both.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
```diff
@@ -30,6 +30,57 @@ class BaseCharsetTest(val charset: Charset) {
...
+        result = d2.decode(in2, out2, true)
+        if(result.isError) result.throwException
+        result = d2.flush(out2)
+        if(result.isError) result.throwException
+        out2.flip
+        out2
+      }
+
+      (r1, r2) match {
+        case (Success(b1), Success(b2)) => assert(b1.toString == b2.toString)
```
sjrd: Consider using `assertEquals` for better error messages.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
```diff
@@ -30,6 +30,57 @@ class BaseCharsetTest(val charset: Charset) {
...
+        case (Failure(err1), Failure(err2)) => {
+          (err1, err2) match {
+            case (m1: MalformedInputException, m2: MalformedInputException) =>
+              assert(m1.getInputLength == m2.getInputLength)
+            case (u1: UnmappableCharacterException, u2: UnmappableCharacterException) =>
+              assert(u1.getInputLength == u2.getInputLength)
+            case _ => assert(false)
```
sjrd: Consider using `assertSame` for better error messages, like it's done [in `testOneConfig`](https://github.com/scala-js/scala-js/blob/ed20d0dac5d92f4a84edf5837054bae70aecb021/test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala#L94-L96).

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
sjrd: And now that I've wrapped my head around what you're testing here, I suggest giving more descriptive names to `r1` and `r2`. Maybe `fullDecodeResult` and `steppedDecodeResult`, respectively?

javalib/src/main/scala/java/nio/charset/UTF_8.scala
sjrd: I believe we can significantly increase the performance of the "good cases" by doing first a quick check for whether there are enough remaining bytes to have a chance of a correct result. Something like
```scala
val decoded = if (inPos + length <= inEnd) {
  // Enough bytes to decode one character
  if (length == 2)
    decode2(leading, inArray(inPos + 1))
  else if (length == 3)
    decode3(leading, inArray(inPos + 1), inArray(inPos + 2))
  else
    decode4(leading, inArray(inPos + 1), inArray(inPos + 2), inArray(inPos + 3))
} else {
  // Not enough bytes, either it's already known to be malformed, or it is an underflow
  var wellFormedUntilPos = inPos + 1
  while (wellFormedUntilPos < inEnd && !isInvalidNextByte(inArray(wellFormedUntilPos)))
    wellFormedUntilPos += 1
  if (wellFormedUntilPos == inEnd)
    DecodedMultiByte(CoderResult.UNDERFLOW)
  else
    DecodedMultiByte(CoderResult.malformedForLength(wellFormedUntilPos - inPos))
}
```
WDYT?

martijnhoekstra:
I'll have to get back on this on Monday, due to a flat laptop battery an no other device at hand.

test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/BaseCharsetTest.scala
martijnhoekstra: I think directDecodeResult may be even better than fullDecodeResult (considering both should be full).

I'll review all names in this method.

javalib/src/main/scala/java/nio/charset/UTF_8.scala
martijnhoekstra: This adds 1 range check to the end of the buffer halfway through a byte sequence (the initial range check + 1 for in the while per continuation byte), but you spare 1 range check for a 3 byte sequence and 2 range checks for a 4 bytes sequence that is not broken up.

My wild guess would be that for something like vietnamese text, which is relatively likely to have a 2 byte sequence at the end of an input buffer, but relatively unlikely to contain a 3 byte sequence, you variation would be slower, and for chinese, japanese or korean text, that are likely to consist mainly of 3 byte sequences, your variation would be faster, maybe significantly so.

Additionally, many of these checks are almost always the same for longer buffers (and who is going to be decoding in an 8 byte buffer?), so branch prediction could be doing a lot of work, to the point where I don't want to predict any performance characteristic.

Also, your code is a bit more compact (though it took me a bit more time to convince myself it was correct).

Do you want benchmarks for this? If so, are there any benchmarks examples I could follow? I wouldn't want to bet on any performance characteristic differences for japanese vs vienamese vs english text, either embedded in html or plain, and with various buffer sizes.

sjrd: Right. So there isn't a clear winner.

> Do you want benchmarks for this?

Nah, just pick whichever you prefer.

martijnhoekstra:
From jenkins, is this spurious?

> [error] Test org.scalajs.jsenv.test.TimeoutComTests.delayedReplyTest[config = Node.js] failed: Execution took too little time
> 
> [error] Failed: Total 41, Failed 1, Errors 0, Passed 40
> 
> [error] Failed tests:
> 
> [error] 	org.scalajs.jsenv.nodejs.NodeJSSuite

It seems unrelated.

sjrd:
Yes that's a spurious error. That's why we retry every job a second time ^^ In the second run there's a non-spurious error (Scalastyle):
```
[error] .../test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/UTF8Test.scala:20:0: Whitespace at end of line
[error] .../test-suite/shared/src/test/scala/org/scalajs/testsuite/niocharset/UTF8Test.scala:94:66: Space before token :
[info] scalastyle Processed 270 file(s)
[info] scalastyle Found 2 errors
[info] scalastyle Found 0 warnings
[info] scalastyle Found 0 infos
```

martijnhoekstra:
because of re-target, appveyor probably need to be manually restarted.

Fixed the same bug in the read-only buffers, and addressed the feedback. Used buffer mark because I don't trust myself to count to three.

sjrd:
We don't have AppVeyor in the 0.6.x branch, so we'll just ignore that status.

library/src/main/scala/scala/scalajs/niocharset/UTF_8.scala
```diff
@@ -160,20 +172,23 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
...
+        def fail(result: CoderResult): CoderResult = {
+          in.reset()
           result
         }
 
         if (!in.hasRemaining) {
           CoderResult.UNDERFLOW
         } else {
+          //mark input position so we can reset on multi-byte underflow
+          in.mark()
```
sjrd: I would prefer not to use `mark()`. It is unclear whether the charset decoder is allowed to mess up with the mark, since it could interfere with the user using `mark()` and `reset()` for their own purposes.

What about just `val savedPosition = in.position()` instead of `in.mark()`, and `in.position(savedPosition)` instead of `in.reset()`?

library/src/main/scala/scala/scalajs/niocharset/UTF_8.scala
```diff
@@ -228,47 +248,34 @@ private[niocharset] object UTF_8 extends Charset("UTF-8", Array(
...
       }
     }
 
     @inline private def decode4(b1: Int, b2: Int, b3: Int, b4: Int): DecodedMultiByte = {
-      if (isInvalidNextByte(b2))
-        DecodedMultiByte(CoderResult.malformedForLength(1))
-      else if (isInvalidNextByte(b3))
-        DecodedMultiByte(CoderResult.malformedForLength(2))
-      else if (isInvalidNextByte(b4))
+      if (isInvalidNextByte(b4))
```
sjrd: This last test being alone here is extremely confusing. I suggest moving it to the two call sites, so that it is uniform wrt. the other bytes.

library/src/main/scala/scala/scalajs/niocharset/UTF_8.scala
martijnhoekstra: Oh, I hadn't considered that. Your suggestion sounds good.

martijnhoekstra:
re: coding style, yes, absolutely. I'm happy to fix anything that there are linting warnings for myself. What do I invoke to check the automated part that I can do myself?

sjrd:
> What do I invoke to check the automated part that I can do myself?

    > library/scalastyle
    > testSuite/test:scalastyle

But tbh that only checks a very small part of our (very stringent) coding style rules, which is why I typically offer to fix up the style myself. I wish we had an automated way to do it, but we're not there yet.

martijnhoekstra:
Addressed the comments. Did one last rebase onto the latest 0.6.x -- hopefully that'll turn Jenkins green.

martijnhoekstra:
I don't understand what Jenkins is tripping over. The logs say

```
Merging remotes/origin/0.6.x commit 2c79fc32587d8c9ceb4145aa26dbc770351e7a39 into PR head commit 05af64fae68dca512febbdb65f5ad8a833118425
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 05af64fae68dca512febbdb65f5ad8a833118425
 > git merge 2c79fc32587d8c9ceb4145aa26dbc770351e7a39 # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 05af64fae68dca512febbdb65f5ad8a833118425
Command "git merge 2c79fc32587d8c9ceb4145aa26dbc770351e7a39" returned status code 128:
stdout: 
stderr: error: merge is not possible because you have unmerged files.
hint: Fix them up in the work tree, and then use 'git add/rm <file>'
hint: as appropriate to mark resolution and make a commit.
fatal: Exiting because of an unresolved conflict.
```

locally, doing the same:

```
$ git checkout -f 05af64fae68dca512febbdb65f5ad8a833118425
HEAD is now at 05af64fae handle incomplete multibyte sequence for UTF-8
$ git merge 2c79fc32587d8c9ceb4145aa26dbc770351e7a39
Already up to date.
```

sjrd:
I'll investigate. It might be the first time we do a branch retargeting since we use Jenkins 2. Our infra might not be up to dealing with it.

sjrd:
I'm at a loss, and I don't much time right now to investigate this. So I opened a different PR at #3442. It seems that Jenkins is better-behaved over there.

PR #3438, opened 2018-08-26
[no-master] Fix #3432: Use the new API of jsdom v10 in JSDOMNodeJSEnv.

### LINKED ISSUE #3432: Support jsdom v12.x in JSDOMNodeJSEnv
jsdom v12.0.0 dropped support for the legacy API that we are using in `JSDOMNodeJSEnv`. See the [release notes](https://github.com/jsdom/jsdom/blob/master/Changelog.md#1200)

> Removed the old jsdom API, as the new API now has all the capabilities you need.

This means we'll need to update `JSDOMNodeJSEnv` to support jsdom v12.x.

sjrd:
jsdom v12.0.0 dropped support for the "old API" that was legacy from v9. We were still using that old API so that we could support jsdom v9 and v10 (and v11), but this is no longer an option for v12.

This commit migrates to using the "new API" introduced in jsdom v10.0.0. This brings support for jsdom v12, but of course the price to pay is that we drop support for jsdom v9.

After this PR is merged, similar changes should be transferred to https://github.com/scala-js/scala-js-env-jsdom-nodejs.

Locally tested with jsdom v12 with:
```
> testingExample:test:run
> testingExample:test
> jsEnvsTestSuite/testOnly org.scalajs.jsenv.test.JSDOMNodeJSEnvTest
> set requiresDOM in testSuite := true
> testSuite/test
```


js-envs/src/main/scala/org/scalajs/jsenv/nodejs/JSDOMNodeJSEnv.scala
```diff
@@ -65,46 +65,59 @@ class JSDOMNodeJSEnv private[jsenv] (
...
       val jsDOMCode = {
         s"""
            |(function () {
-           |  var jsdom;
-           |  try {
-           |    jsdom = require("jsdom/lib/old-api.js"); // jsdom >= 10.x
-           |  } catch (e) {
-           |    jsdom = require("jsdom"); // jsdom <= 9.x
-           |  }
+           |  var jsdom = require("jsdom");
```
gzm0: Can't we detect this somehow? It seems odd to break compatibility TBH.

sjrd: Sure we can. It's even easy: `typeof jsdom.JSDOM === "function"` will tell us whether it's jsdom >= 10. It's just that if we do that, we need to maintain two completely different implementations forever.

Every now and then, we require people to upgrade their version of sbt. I don't think it's unreasonable to demand that they upgrade to a newer version of another supporting infrastructure.

sjrd: Yes, you're right, the major version upgrade makes this case different than sbt. I'll adapt to keep support for v9.x.

sjrd: Done.

Locally tested with the same commands, with jsdom 9.12.0, 10.0.0 and 12.0.0.

gzm0:
The description of the PR is outdated.

PR #3441, opened 2018-08-28
Deprecate js.JSArrayOps.

sjrd:
The methods in that extension trait all have better alternatives in the normal collection methods. It almost never makes sense to use them. For the rare (virtually non-existent) use cases, it is still possible to use `js.Dynamic` or one's own facade types.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/JSArrayOps.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSArrayOps.scala
@@ -27,6 +27,12 @@ import scala.scalajs.js.annotation._
  *  To enable the use of these functions on js.[[Array]]s, import the implicit
  *  conversion [[JSArrayOps.jsArrayOps]].
  */
+@deprecated(
+    "It almost never makes sense to call native JavaScript Array functions " +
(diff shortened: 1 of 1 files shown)


library/src/main/scala/scala/scalajs/js/JSArrayOps.scala
```diff
@@ -27,6 +27,12 @@ import scala.scalajs.js.annotation._
  *  To enable the use of these functions on js.[[Array]]s, import the implicit
  *  conversion [[JSArrayOps.jsArrayOps]].
  */
+@deprecated(
+    "It almost never makes sense to call native JavaScript Array functions " +
```
gzm0: In what cases does it make sense?

sjrd: For example if you have a sparse array, and you want holes to be skipped. But to get sparse JS arrays in Scala.js you have to mean it.

gzm0: Fair enough :)

PR #3448, opened 2018-09-02
Cleanup internal annotations

sjrd:


sjrd:
I couldn't make the removal of `@JSType` work, so I'll just leave the two first commits. Ready for review.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -82,8 +82,8 @@ private final class IRChecker(unit: LinkingUnit,
               classDef.memberMethods.nonEmpty ||
               classDef.exportedMembers.nonEmpty ||
               classDef.topLevelExports.nonEmpty) {
-            reportError(s"Raw JS type ${classDef.name} cannot "+
-                "have instance members")
+            reportError(
+                s"JS type ${classDef.name} cannot have instance members")
```
gzm0: Report the more precise type (abstract / native)?

sjrd: Done.

PR #3451, opened 2018-09-05
Fix #3450: Remove the stubs library.

### LINKED ISSUE #3450: Move the remaining stubs in a separate repo
Now that `TestUtils` is gone, `scalajs-stubs` is not necessary to the core repo anymore. I believe we should move it to a separate repo. It's tiny at this point; it only contains definitions for `@JSExport`, `@JSExportAll`, `@JSExportTopLevel` and `@EnableReflectiveInstantiation`.

I suggest to put it under `"org.portable-scala" %% "scalajs-stubs"`.

It's interesting to note that the Community Build already uses a couple hacks to treat scalajs-stubs as if it were from a different repo than scala-js/scala-js:
https://github.com/scala/community-builds/blob/f7cea0b957407c94291778578727ccd213832088/configs/community.dbuild#L208-L218

Does that make sense?

sjrd:
It has been moved to the separate repository at https://github.com/scala-js/scala-js-stubs.

Only merge after https://github.com/scala-js/scala-js-stubs/pull/1 has been merged.


library/src/main/scala/scala/scalajs/js/annotation/README.md
```diff
@@ -1,3 +0,0 @@
-**Attention**: Some files in here are also published in the Scala.js stubs JVM library (see the stubs project in the Scala.js build).
-
-If you add (or rename) a file, make sure the files in the stubs project are up to date.
```
gzm0: Shouldn't this file stay? I mean keeping the stubs in sync is still important.

sjrd: Done. I've also *added* an equivalent file in `reflect/annotation/README.md`.

PR #3456, opened 2018-09-12
Fix ill-stylising.

lamusique:
Just fix an ill-formed markdown rendered well.


lamusique:
Did sign the CLA even for just a Wiki thingy.

PR #3459, opened 2018-09-24
Make the ir/ subproject compilable with Dotty.

sjrd:
This will be necessary for Dotty to compile to sjsir, as it will need to bring the sources of scalajs-ir in the dotty compiler, which needs to be compilable by itself.

Changes are due to:

* Dotty requires call sites to use `foo()` rather than `foo` if the method `foo` was declared with `()`.
* Improved reachability analysis in the pattern matcher discovers more stuff.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Hashers.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Hashers.scala
@@ -441,10 +441,6 @@ object Hashers {
           throw new InvalidIRException(tree,
               "Cannot hash a transient IR node (its value is of class " +
               s"${value.getClass})")
-
-        case _ =>
-          throw new IllegalArgumentException(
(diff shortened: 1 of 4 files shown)


ir/src/main/scala/org/scalajs/ir/Hashers.scala
```diff
@@ -441,10 +441,6 @@ object Hashers {
           throw new InvalidIRException(tree,
               "Cannot hash a transient IR node (its value is of class " +
               s"${value.getClass})")
-
-        case _ =>
-          throw new IllegalArgumentException(
-              s"Unable to hash tree of class ${tree.getClass}")
```
sjrd: This `match` is in fact provably exhaustive, and Dotty reports a warning for a `case _ =>` when the rest is otherwise exhaustive.

PR #3463, opened 2018-10-08
Better test kit

gzm0:


gzm0:
Updated:
- Fixed to work without SAM treatment.
- Fixed TimeoutRunTest (missing "\n").
- Fixed to work without `sci.Queue#dequeueOption`

sjrd:
Jenkins, retest this please

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutComTests.scala
```diff
@@ -17,9 +17,16 @@ private[test] class TimeoutComTests(config: JSEnvSuiteConfig) {
     assumeTrue("JSEnv needs com support", config.supportsCom)
   }
 
+  /** Slack for timeout tests (see #3457)
+   *
+   *  Empirically we can observe that timing can be off by ~0.1ms. By cutting
+   *  10ms slack, we definitely account for this without compromising the tests.
+   */
+  private val slack = 10.millis
```
gzm0: I wonder if we should make this part of `JSEnvSuiteConfig`?

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/ComRun.scala
```diff
@@ -7,23 +7,42 @@ import org.junit.Assert._
...
     extends Run(run, out, err, timeout) {
   private[this] var noMessages = false
 
+  /** Calls [[JSComRun#send]] on the underlying run. */
   final def send(msg: String): this.type = {
     run.send(msg)
     this
   }
 
+  /** Waits until the given message sent to the JVM.
```
sjrd: <ins>is</ins> sent

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/TimeoutComTests.scala
sjrd: I don't think we should move it to `JSEnvSuiteConfig`, no. At least not yet. If we discover a case later, we can always move it to the Config later in a binary compatible way, so I'd refrain from over-generalizing too early.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/ComRun.scala
```diff
@@ -7,23 +7,42 @@ import org.junit.Assert._
 
 import org.scalajs.jsenv._
 
+/** A [[JSComRun]] instrumented for testing.
+ *
+ *  Create an instance of this through one of the overloads of
```
sjrd: of this <ins>class</ins>?

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/Run.scala
```diff
@@ -9,22 +9,48 @@ import org.junit.Assert._
 
 import org.scalajs.jsenv._
 
+/** A [[JSRun]] instrumented for testing.
+ *
+ *  Create an instance of this through one of the overloads of
```
sjrd: of this <ins>class</ins>?

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/Run.scala
```diff
@@ -9,22 +9,48 @@ import org.junit.Assert._
...
 class Run private[kit] (run: JSRun, out: IOReader, err: IOReader, timeout: FiniteDuration) extends AutoCloseable {
   private[this] val utf8decoder = {
     StandardCharsets.UTF_8.newDecoder()
       .onMalformedInput(CodingErrorAction.REPLACE)
       .onUnmappableCharacter(CodingErrorAction.REPLACE)
   }
 
+  /** Waits until the given string is output to stdout (in UTF8).
+   *
+   *  @throws java.lang.AssertionError if there is other output or the run terminates.
```
sjrd: another / some other

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/Run.scala
```diff
@@ -9,22 +9,48 @@ import org.junit.Assert._
...
+   *  @throws java.util.concurrent.TimeoutException if there is not enough output for too long.
+   */
   final def expectOut(v: String): this.type = expectIO(out, "stdout", v)
+
+  /** Waits until the given string is output to stderr (in UTF8).
+   *
+   *  @throws java.lang.AssertionError if there is other output or the run terminates.
```
sjrd: same here: another / some other

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/TestKit.scala
```diff
@@ -9,22 +9,64 @@ import java.util.concurrent.Executors
 import org.scalajs.io.MemVirtualBinaryFile
 import org.scalajs.jsenv._
 
+/** TestKit is a utility class to simplify testing of [[JSEnv]]s
```
sjrd: Add a period at the end of the line so that Scaladoc knows where the short description ends.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/TestKit.scala
```diff
@@ -9,22 +9,64 @@ import java.util.concurrent.Executors
...
+ *
+ *  val kit = new TestKit(new MyEnv, 1.second)
+ *  kit.withRun("""console.log("Hello World");""") {
+ *    _.expectOut("Hello World\n")
+ *      .closeRun()
+ *  }
+ *  }}}
+ *
+ *  @note Methods in [[TestKit]] allow to take a string instead of an [[Input]].
+ *      The string is converted into an input form supported by the [[JSEnv]] to
```
sjrd: That phrasing is weird. It suggests that `TestKit` will automagically decide how best to convert the string into an `Input`, depending on what the given `JSEnv` supports. What if in the future we have another kind of `Input`, and a `JSEnv` does not support `ScriptsToLoad`?

I suggest directly saying that a string corresponds to an `Input.ScriptsToLoad` with the string as only script.

gzm0: This is precisely what this phrasing is supposed to mean. We will need something like this to have a portable test suite. It is the reason why I backed down from having an implicit conversion `String => Input`. Instead, the current API will allow us to do something like:

```scala
def tryStart(input: Input): Option[Run] = {
  try {
    Some(start(input))
  } catch {
    case _: UnsupportedInputException => None
}

tryStart(Input.ScriptsToLoad(...)) orElse tryStart(Input.OtherType(...))
```

gzm0:
Updated.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/TestKit.scala
sjrd: Ah, I see. OK.

sjrd:
needs squashing, but otherwise good to go.

PR #3464, opened 2018-10-08
Fix #3201: Provide more detail on "Unknown opcode: 5".

### LINKED ISSUE #3201: Intermittent IllegalStateException: Unknown opcode 5 while running tests
See #3175. The hanging part was solved in #3194, but there are still occasional errors with

which seem to come from the JavaScript side, which means it corresponds to [`JSEndpoints.msgSlave`](https://github.com/scala-js/scala-js/blob/e9cc89c4fe483aa45d2d193c3a8d3772d1292a17/test-common/src/main/scala/org/scalajs/testcommon/JSEndpoints.scala#L15-L16), which is supposed to be attached [`here`](https://github.com/scala-js/scala-js/blob/e952f9ce520372600a00f5dc3b64062fd6bded2c/test-interface/src/main/scala/org/scalajs/testinterface/internal/Bridge.scala#L61).

gzm0:


sjrd:
Jenkins, test this please

sjrd:
Er ... Jenkins refuses to see that PR :( It's like it doesn't even receive any kind of notification about it.

gzm0:
Seems it has picked it up now. Note that I have added a big comment that this is not so nice, but as good as it gets :)

PR #3465, opened 2018-10-09
CI ONLY Fix #2433: Automatic CLA validation in CI

### LINKED ISSUE #2433: Automatic CLA validation in CI
Doesn't look too hard: https://github.com/scala-native/scala-native/pull/139

gzm0:


neVERberleRfellerER:
As an alternative to Jenkin's httpRequest plugin quick and dirty help from node is available, e.g.:

```js
#!/usr/bin/env node
'use strict';

const http = require('https');

const requestOptions = {
    headers: {
        'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36'
    }
};

function httpGet(url, options) {
    return new Promise(function (resolve, reject) {
        http.get(url, options, resp => {
            const statusCode = resp.statusCode;
            if (statusCode !== 200) {
                reject(Error(`HTTP request returned non-OK status code: ${statusCode}`));
                return;
            }

            let data = '';

            resp.on('data', chunk => {
                data += chunk;
            });

            resp.on('end', err => {
                if (err) {
                    reject(err);
                    return;
                }

                resolve(data);
            });
        }).on('error', err => {
            reject(err);
            return;
        });
    });
}

async function isCLASignedForPR(prNumber) {
    const pullRequestInfoUrl = `https://api.github.com/repos/scala-js/scala-js/pulls/${prNumber}`;
    const prInfo = JSON.parse(await httpGet(pullRequestInfoUrl, requestOptions));

    const authorLogin = prInfo.user.login;
    const claInfoUrl = `https://www.lightbend.com/contribute/cla/scala/check/${authorLogin}`;
    const claInfo = JSON.parse(await httpGet(claInfoUrl, requestOptions));

    return claInfo.signed;
}

(async function main() {
    const pullRequestId = process.argv[2];
    try {
        const signed = await isCLASignedForPR(pullRequestId);
        process.exit(signed ? 0 : 1);
    } catch (e) {
        console.log('error: ' + e.message);
        process.exit(2);
    }
})();

```
Output:
```
$ ./checkCLAforPR.js 3468 && echo "signed" || echo "not signed"
signed

$ ./checkCLAforPR.js 9999 && echo "signed" || echo "not signed"
error: HTTP request returned non-OK status code: 404
not signed

$ ./checkCLAforPR.js 3465 && echo "signed" || echo "not signed"
signed
```
Could be launched with sh and exit codes should do the rest

PR #3468, opened 2018-10-12
Fix #3280: Count capturing groups independently on find.

### LINKED ISSUE #3280: Matcher.groupCount behaviour inconsistent between JS and JVM
The following works fine under both platforms:

```scala
val matcher = Pattern.compile("\\d+").matcher("2147483647")

matcher.find()

matcher.groupCount
```

The following fails in `js` but not `jvm`:

```scala
val matcher = Pattern.compile("\\d+").matcher("2147483647")

//matcher.find()

matcher.groupCount
```

There's nothing in the `Matcher` scaladoc that seems to imply that calling `find` before `groupCount` is needed. 

It does, however, seem like an unfixable bug. I'm no javascript expert, but it does rather look that regexes can only expose the group count for a given match, not for a matcher.

neVERberleRfellerER:
Since
https://www.ecma-international.org/ecma-262/6.0/#sec-regexpbuiltinexec
points 19 nad 20 are always hit and set result size to NcapturingParens+1 and
https://www.ecma-international.org/ecma-262/6.0/#sec-notation
states that "NcapturingParens is the total number of left capturing parentheses" then it should be safe to create always-matching regexp by prepending "|" and getting group count from resulting successful match.

Fixes https://github.com/scala-js/scala-js/issues/3280

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Matcher.scala
+++ b/javalib/src/main/scala/java/util/regex/Matcher.scala
@@ -24,6 +24,9 @@ final class Matcher private[regex] (
   private var lastMatchIsValid = false
   private var canStillFind = true
 
+  // Group count
+  private var lastGroupCount: Option[Int] = None
+
   // Append state (updated by replacement methods)
   private var appendPos: Int = 0
 
@@ -160,6 +163,7 @@ final class Matcher private[regex] (
     regexp = pattern.newJSRegExp()
     regexp.lastIndex = prevLastIndex
     lastMatch = null
+    lastGroupCount = None
     startOfGroupCache = None
     this
   }
@@ -172,7 +176,21 @@ final class Matcher private[regex] (
     lastMatch
   }
 
-  def groupCount(): Int = ensureLastMatch.length-1
+  def groupCount(): Int = {
+    if (lastMatch != null) {
+      lastMatch.length-1
+    } else {
+      lastGroupCount match {
+        case Some(n) => n
+
+        case None =>
+          val groupCountRegex = new js.RegExp("|" + pattern0.jsPattern)
+          val newGroupCount = groupCountRegex.exec("").length-1
+          lastGroupCount = Some(newGroupCount)
+          newGroupCount
+      }
+    }
+  }
 
   def start(): Int = ensureLastMatch.index
   def end(): Int = start() + group().length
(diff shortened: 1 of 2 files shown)


neVERberleRfellerER:
I am checking out your CLA but I am slightly confused. I guess it wants my real name, but it is not clear whether Ligtbend's "Privacy policy" applies to data gathered in CLA or if CLA is separate and gathered data are never disclosed and used just in case of "problems". Also it's not clear, if Lightbend can access the submitted data since CLA only mentions EPFL but no Lightbend or third-parties (in context of data access) in general, but form is submitted on Lighbend's website. Could you give me answer to at least that part with full name sharing with third-parties?

neVERberleRfellerER:
I did sign the CLA.

javalib/src/main/scala/java/util/regex/Matcher.scala
```diff
@@ -172,7 +175,18 @@ final class Matcher private[regex] (
...
   }
 
-  def groupCount(): Int = ensureLastMatch.length-1
+  def groupCount(): Int = {
+    lastGroupCount match {
+      case Some(n) => n
+
+      case None =>
+        val groupCountPattern = Pattern.compile("|" + pattern0.jsPattern)
+        val groupCountRegex = groupCountPattern.newJSRegExp()
```
sjrd: You can avoid the `Pattern.compile` step in this case. Just do
```scala
val groupCountRegex = new js.RegExp("|" + pattern0.jsPattern)
```
this will be faster. Flags are irrelevant, and `pattern0.jsPattern` has already been processed for in-pattern flags, so `Pattern.compile` is going to do work for nothing.

gzm0:
> I guess it wants my real name.

Hum, just to make sure if you are concerned about your real name: There is a very real seeming name in your git commit metadata which is publicly visible.

neVERberleRfellerER:
@gzm0 I know. I was just  concerned that it wants real name without any privacy agreement (which was oversight on my side). Just because I use my real name on github does not mean that I automatically want to "give" my real name to another company when I don't know what they are going to use it for, especially when CLA actually mentions only EPFL (which I have no problem with). It's purely ideological problem on my side.

javalib/src/main/scala/java/util/regex/Matcher.scala
```diff
@@ -172,7 +176,20 @@ final class Matcher private[regex] (
     lastMatch
   }
 
-  def groupCount(): Int = ensureLastMatch.length-1
+  def groupCount(): Int = {
+    if (lastMatch != null) lastMatch.length-1
+    else {
```
sjrd: Looking good! One tiny remaining detail: could you format the above as
```scala
if (lastMatch != null) {
  lastMatch.length-1
} else {
  ...
```
to follow [our `if/else` coding style](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#ifelse) which demands that both branches of an `if/else` must have `{}`, or none of them has braces.

Good to merge after that!

PR #3471, opened 2018-10-17
simplify foldLeft to forall

teresy:


teresy:
Hi there @sjrd. Not a bot :) Though I did run a tool to find this. Also checked that it's in line with the [style guide](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#collections). Commit message updated, thanks.

sjrd:
Thank you. Looks good.

Right now we're in the middle of a release process, so I can't merge it, but I'll come back to this as soon as we're through with the release process.

PR #3477, opened 2018-10-24
Fix #3476: Specify that com messages must be valid UTF-16 strings.

### LINKED ISSUE #3476: JSEnv com messages should be limited to valid UTF-16 strings
i.e., unpaired surrogates should not be valid.

We need to change:

* the documentation so that this is made clear
* the test `ComTests.largeMessageTest`, which currently breaks that assumption.

See https://github.com/scala-js/scala-js-env-phantomjs/pull/23 for context.

sjrd:
* Document it in the Scaladoc of `startWithCom`
* Assert that all messages are valid in the `TestKit`
* Fix `ComTests.largeMessageTest` not to use an invalid message

### DIFF
--- a/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
+++ b/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
@@ -103,8 +103,13 @@ private[test] class ComTests(config: JSEnvSuiteConfig) {
 
   @Test
   def largeMessageTest: Unit = {
-    // 1MB data
-    replyTest(new String(Array.tabulate(1024 * 1024)(_.toChar)))
+    /* 1MB data.
+     * (i & 0x7f) limits the input to the ASCII repertoire, which will use
+     * exactly 1 byte per Char in UTF-8. This restriction also ensures that we
+     * do not introduce surrogate characters and therefore no invalid UTF-16
+     * strings.
+     */
+    replyTest(new String(Array.tabulate(1024 * 1024)(i => (i & 0x7f).toChar)))
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
```diff
@@ -104,7 +104,7 @@ private[test] class ComTests(config: JSEnvSuiteConfig) {
   @Test
   def largeMessageTest: Unit = {
     // 1MB data
-    replyTest(new String(Array.tabulate(1024 * 1024)(_.toChar)))
+    replyTest(new String(Array.tabulate(1024 * 1024)(i => (i & 0x7f).toChar)))
```
gzm0: `0x7f` as a constant probably deserves a comment.

js-envs/src/main/scala/org/scalajs/jsenv/JSEnv.scala
```diff
@@ -51,6 +51,11 @@ trait JSEnv {
    *  scalajsCom.send("my message");
    *  }}}
    *
+   *  All messages, sent in both directions, must be valid UTF-16 strings,
+   *  i.e., they must not contain any unpaired surrogate character. The
+   *  behavior a communication channel is unspecified if this requirement is
```
gzm0: behavior of a

sjrd:
Updated.

gzm0:
I think this comment got lost:

> Maybe add a reminder about the UTF-16 requiement on JSComRun#send?

sjrd:
Indeed, I had completely overlooked it. Fixed.

PR #3478, opened 2018-10-24
[no-master] Fix #2175: Add support for ECMAScript 2015 modules.

### LINKED ISSUE #2175: Integrating with JS module systems

sjrd:
This commit adds a third `ModuleKind` for ES modules, namely `ModuleKind.ESModule`. When emitting an ES module, `@JSImport`s and `@JSExportTopLevel`s straightforwardly map to ES `import` and `export` clauses, respectively.

At the moment, imports are always implemented using a namespace import, then selecting fields inside the namespace. This is suboptimal because it can prevent advanced DCE across ES modules. Improving on this is left for future work.

The Node.js-based environment is adapted to interpret files whose name ends with `.mjs` as ES modules rather than scripts. This aligns with how Node.js itself identifies ES modules as of version 10.x, although it is still experimental, so that could change in the future.

For the 0.6.x branch, the `TestAdapter` assumes that it can force the `JSEnv` to interpret its launcher as an ES module by giving it the `.mjs` extension as well. This is wrong in general, but there does not seem to be a good way to deal with this issue. In 1.x, this will be a non-issue since the `TestAdapter` does not require any launcher.

Although setting `scalaJSLinkerConfig.moduleKind` to `ModuleKind.ESModule` is enough for the Scala.js linker to emit a valid ES module, two additional settings are required to *run* or *test* using Node.js:
```scala
artifactPath in (proj, Compile, fastOptJS) :=
  (crossTarget in (proj, Compile)).value / "somename.mjs"

jsEnv := {
  new org.scalajs.jsenv.NodeJSEnv(
      org.scalajs.jsenv.NODEJSEnv.Config()
        .withArguments(List("--experimental-modules"))
  )
}
```
The first setting is necessary to give the `.mjs` extension to the file produced by Scala.js, which in turn is necessary for Node.js to accept it as an ES module.

The second setting will be necessary until Node.js declares its support for ES module as non-experimental.

ES modules are incompatible with a few features, which are all gone in Scala.js 1.x:

* `EnvironmentInfo.exportsNamespace`: an ES module cannot read its   own exports namespace, short of importing itself (which requires   it to know its file name). The value of `exportsNamespace` is `undefined` in an ES module.
* Generation of a `main` launcher script by the sbt plugin. Attempting to use one will throw an exception in the build.

Moreover, the version of the Closure Compiler that we use does not support ES modules yet, so we deactivate GCC when emitting an ES module.

At this point, the emission of ES modules can be considered stable, but the support in `NodeJSEnv` is experimental (since the support of ES modules in Node.js is itself experimental).

Running the full test suite with ES modules requires Node.js 10.2.0 or later. It has been tested with v10.12.0.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
@@ -134,12 +134,37 @@ abstract class AbstractNodeJSEnv(
      */
     override protected def writeJSFile(file: VirtualJSFile,
         writer: Writer): Unit = {
+
+      def writeImport(file: File): Unit = {
+        val uri = file.toURI.toASCIIString
+        val importerFile = new MemVirtualJSFile("importer.js")
+        importerFile.content = {
+          s"""
+            |import("${escapeJS(uri)}").catch(e => {
+            |  /* Make sure to fail the process, but give time to Node.js to
+            |   * display a good stack trace before that.
+            |   */
+            |  setTimeout(() => process.exit(1), 100);
+            |  throw e;
+            |});
+          """.stripMargin
+        }
+        val f = libCache.materialize(importerFile)
+        writer.write(s"""require("${escapeJS(f.getAbsolutePath)}");\n""")
+      }
+
       file match {
         case file: FileVirtualJSFile =>
           val fname = file.file.getAbsolutePath
-          writer.write(s"""require("${escapeJS(fname)}");\n""")
+          if (fname.endsWith(".mjs"))
+            writeImport(file.file)
+          else
+            writer.write(s"""require("${escapeJS(fname)}");\n""")
         case _ =>
-          super.writeJSFile(file, writer)
+          if (file.path.endsWith(".mjs"))
+            writeImport(libCache.materialize(file))
+          else
+            super.writeJSFile(file, writer)
       }
     }
 
--- a/test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
+++ b/test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
@@ -133,19 +133,6 @@ final class TestAdapter(jsEnv: ComJSEnv, config: TestAdapter.Config) {
     // Otherwise we might leak runners.
     require(!closed, "We are closed. Cannot create new runner.")
 
-    // !!! DUPLICATE code with ScalaJSPlugin.makeExportsNamespaceExpr
-    val orgExpr = config.moduleKind match {
-      case ModuleKind.NoModule =>
-        "typeof(org) != 'undefined' ? org : {}"
-
-      case ModuleKind.CommonJSModule =>
-        val moduleIdent = config.moduleIdentifier.getOrElse {
-          throw new IllegalArgumentException(
-              "The module identifier must be specified for CommonJS modules")
-        }
-        s"""require("${escapeJS(moduleIdent)}").org || {}"""
-    }
-
     /* #2752: if there is no testing framework at all on the classpath,
      * the testing interface will not be there, and therefore the
      * `startBridge` function will not exist. We must therefore be
@@ -153,19 +140,55 @@ final class TestAdapter(jsEnv: ComJSEnv, config: TestAdapter.Config) {
      * If it is not present, we will simply exit; `loadFrameworks` is prepared
      * to deal with this case.
      */
-    val code = s"""
-      (function() {
-        "use strict";
-        var namespace = $orgExpr;
+    val startBridgeFun = """
+      function startBridge(namespace) {
         namespace = namespace.scalajs || {};
         namespace = namespace.testinterface || {};
         namespace = namespace.internal || {};
         var bridge = namespace.startBridge || function() {};
         bridge();
-      })();
+      }
     """
 
-    val launcher = new MemVirtualJSFile("startTestBridge.js").withContent(code)
+    val code = config.moduleKind match {
+      case ModuleKind.NoModule =>
+        s"""
+          (function() {
+            "use strict";
+            $startBridgeFun
+            startBridge(typeof(org) != 'undefined' ? org : {});
+          })();
+        """
+
+      case ModuleKind.ESModule =>
+        val moduleIdent = config.moduleIdentifier.getOrElse {
+          throw new IllegalArgumentException(
+              "The module identifier must be specified for ES modules")
+        }
+        val uri = new java.io.File(moduleIdent).toURI.toASCIIString
+        s"""
+          import * as mod from "${escapeJS(moduleIdent)}";
+          $startBridgeFun
+          startBridge(mod.org || {});
+        """
+
+      case ModuleKind.CommonJSModule =>
+        val moduleIdent = config.moduleIdentifier.getOrElse {
+          throw new IllegalArgumentException(
+              "The module identifier must be specified for CommonJS modules")
+        }
+        s"""
+          (function() {
+            "use strict";
+            $startBridgeFun
+            startBridge(require("${escapeJS(moduleIdent)}").org || {});
+          })();
+        """
+    }
+
+    val ext = if (config.moduleKind == ModuleKind.ESModule) ".mjs" else ".js"
+
+    val launcher = new MemVirtualJSFile("startTestBridge" + ext).withContent(code)
     val runner = jsEnv.comRunner(launcher)
     val com = new ComJSEnvRPC(runner)
     val mux = new RunMuxRPC(com)
--- a/test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
+++ b/test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
@@ -26,7 +26,7 @@ class ModulesTest {
 
   @Test def testImportModuleItself(): Unit = {
     val qs = QueryString
-    assertTrue(qs.isInstanceOf[js.Object])
+    assertEquals("object", js.typeOf(qs))
 
     val dict = js.Dictionary("foo" -> "bar", "baz" -> "qux")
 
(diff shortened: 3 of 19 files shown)


sjrd:
Looking at the mess I made of `Emitter` and `ClassEmitter`, I'm thinking we should perhaps deprecate `@JSExportTopLevel` with `.`'s inside, i.e., with qualified names. The automatic creation of namespaces is painful, and does not have any direct equivalent in ES. The difference of treatment for `TopLevelFieldExportDef` between unqualified and qualified name is particularly striking me as a bad smell.

gzm0:
Yes, I think deprecating top-level exports with dots is a reasonable idea.

In the worst case, one can always write:

```scala
object Foo {
  @JSExportTopLevel("foo")
  val foo = new js.Object {
    val bar = new js.Object {
      def hello(s: String): String = "Hello " + s
    }
  }
}
```

To prevent too much mess from building up, maybe already disallow this, when exporting to an ESModule?

gzm0:
Is it worth targeting 0.6.x with this? It seems there is a lot of complexity because of old interfaces. (N.B. we do not have `Input` yet).

sjrd:
> To prevent too much mess from building up, maybe already disallow this, when exporting to an ESModule?

Our own test bridge uses a qualified export in 0.6.x :( So I don't think we can disallow it right now. Unless we change the export of the bridge now? Can we do that?

> Is it worth targeting 0.6.x with this? It seems there is a lot of complexity because of old interfaces. (N.B. we do not have Input yet).

I think it is worth it, yes. I was afraid of the complexity initially, but in fact it's not too bad. The main issue is the invocation of the bridge in the test adapter, but it's limited to one method. There is a small price to pay for putting it in 0.6.x, but I believe users will sufficiently benefit from this feature to be worth that price.

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
```diff
@@ -134,12 +134,37 @@ abstract class AbstractNodeJSEnv(
...
+        importerFile.content = {
+          s"""
+            |import("${escapeJS(uri)}").catch(e => {
+            |  /* Make sure to fail the process, but give time to Node.js to
+            |   * display a good stack trace before that.
+            |   */
+            |  setTimeout(() => process.exit(1), 100);
+            |  throw e;
+            |});
+          """.stripMargin
```
gzm0: Hum, how is this supposed to work? There is absolutely no guarantee that once this file is required, anything is loaded...

So if there is code like this:
```javascript
require("foo_importer.js");
foo.blah();
```

The second line might randomly fail because `foo` is not loaded yet.

test-adapter/src/main/scala/org/scalajs/sbttestadapter/TestAdapter.scala
```diff
@@ -133,39 +133,62 @@ final class TestAdapter(jsEnv: ComJSEnv, config: TestAdapter.Config) {
...
+            "use strict";
+            $startBridgeFun
+            startBridge(typeof(org) != 'undefined' ? org : {});
+          })();
+        """
+
+      case ModuleKind.ESModule =>
+        val moduleIdent = config.moduleIdentifier.getOrElse {
+          throw new IllegalArgumentException(
+              "The module identifier must be specified for CommonJS modules")
```
gzm0: ES / EcmaScript?

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -26,7 +26,7 @@ class ModulesTest {
 
   @Test def testImportModuleItself(): Unit = {
     val qs = QueryString
-    assertTrue(qs.isInstanceOf[js.Object])
+    assertEquals("object", js.typeOf(qs))
```
gzm0: Why is this necessary?

tools/jvm/src/main/scala/org/scalajs/core/tools/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -58,8 +58,8 @@ final class ClosureLinkerBackend(
   val symbolRequirements: SymbolRequirement = emitter.symbolRequirements
 
   private val needsIIFEWrapper = moduleKind match {
-    case ModuleKind.NoModule       => true
-    case ModuleKind.CommonJSModule => false
+    case ModuleKind.NoModule                             => true
+    case ModuleKind.ESModule | ModuleKind.CommonJSModule => false
```
gzm0: I thought closure doesn't support es modules?

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Trees.scala
```diff
@@ -241,4 +290,53 @@ object Trees {
...
+   *  The `_1` parts of bindings are therefore the identifier names that are
+   *  imported, as specified in `export` clauses of the module. The `_2` parts
+   *  are the names under which they are imported in the current module.
+   *
+   *  Special cases:
+   *  - When `_1 == _2.name`, there is shorter syntax in ES, i.e.,
+   *    `import { binding } from 'from'`.
+   *  - When `_1 == "default"`, it is equivalent to a default import.
+   */
+  case class Import(bindings: List[(String, Ident)], from: StringLiteral)(
```
gzm0: So `_1` isn't an `Ident` because there are different naming rules I assume? If yes, why don't we need an `Export` (or `ExportIdent`) tree?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1068,31 +1068,38 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
...
-    val defProp =
-      genIdentBracketSelect(js.VarRef(js.Ident("Object")), "defineProperty")
-
-    // optional getter definition
-    val getterDef = {
-      js.StringLiteral("get") -> js.Function(Nil, {
-        js.Return(genSelectStatic(cd.encodedName, field))
-      })
-    }
+    if (moduleKind == ModuleKind.ESModule && fullName.indexOf('.') < 0) {
```
gzm0: `!contains('.')`?

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1138,9 +1145,16 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
    */
   private def genCreateNamespaceInExports(qualName: String)(
       implicit pos: Position): (js.Tree, js.Tree) = {
-    val (createNamespace, namespace, lastPart) =
-      genCreateNamespaceInExportsAndGetNamespace(qualName)
-    (createNamespace, genBracketSelect(namespace, lastPart))
+    if (moduleKind == ModuleKind.ESModule && qualName.indexOf('.') < 0) {
```
gzm0: `!qualName.contains('.')`?

gzm0:
What is our merging strategy for this? It seems it might make sense to mark this `[no-master]` since most of it will not be useful in master (only JS trees from what I see). Especially if we disallow nested top-level exports (which we should IMO, as it seems to be very unidiomatic JS).

sjrd:
> What is our merging strategy for this? It seems it might make sense to mark this [no-master] since most of it will not be useful in master (only JS trees from what I see). Especially if we disallow nested top-level exports (which we should IMO, as it seems to be very unidiomatic JS).

There's a few more things than just the JS trees, e.g., tests, but yeah not much. I've done a test cherry-pick on master, and there are more conflicts than clean merges. It looks like this might be a case where a forward *port* is more appropriate than a forward merge. I'll mark the commit [no-master].

js-envs/src/main/scala/org/scalajs/jsenv/nodejs/AbstractNodeJSEnv.scala
gzm0: Hmmm... I see; ok I guess for the 0.6 series with it's unclear semantics of what files passed to a `JSEnv` mean this is OK.

tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Trees.scala
```diff
@@ -21,6 +21,55 @@ import ir.Position.NoPosition
...
+   *  `Other_ID_Continue`. For example,
+   *  `isValidIdentifierName(0x2118.toChar.toString)` will return `false`
+   *  instead of `true`.
+   *
+   *  In theory, it does not really account for code points with the Unicode
+   *  properties `Pattern_Syntax` and `Pattern_White_Space`, which should be
+   *  rejected. However, with the current version of Unicode (9.0.0), there
+   *  seems to be no such character that would be accepted by this method.
+   */
+  final def isValidIdentifierName(name: String): Boolean = {
```
gzm0: Now this can be private, no?

sjrd: It technically can, but I would leave it public. A user of the API would potentially want to call this method to validate inputs, rather than crashing when trying to construct an `ExportName`.

Maybe expose it as `ExportName.isValidExportName` instead?

gzm0: That's fair. Yes I like that name better even if it matches less ES terminology. But in terms of our interface it is definitely cleaner.

PR #3483, opened 2018-11-05
Simplify JUnit runner / task relationship

gzm0:
Runners kept a task count which was unnecessary. As a result, we do
not need a slave/master distinction anymore. Lastly, we can hide the
runners from the tasks and only forward the RunSettings.


gzm0:
> Is it because any given Task is guaranteed to be executed on exactly one node?

Yes, basically. Every `Task` may be executed once only on a single node only.

PR #3486, opened 2018-11-07
Disallow namespaced top-level exports.

sjrd:
`@JSExportTopLevel` with namespaces (containing a `.`) were deprecated in 9404d59827747bd36bb64e09b494bbaa4b423203 because they do not have any appropriate equivalent in ECMAScript modules. This commit disallows them.

Afterwards, the definition of the IR is implicitly simplified, and we rename the field `fullName` to `exportName` now that it cannot contain namespaces anymore.

We simplify the `ClassEmitter` as we remove all the handling of namespaces in top-level exports.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
@@ -1057,12 +1057,17 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
 
   def genTopLevelExports(tree: LinkedClass)(
       implicit globalKnowledge: GlobalKnowledge): WithGlobals[List[js.Tree]] = {
-    val exportsWithGlobals = tree.topLevelExports.map { topLevelExport =>
-      topLevelExport.value match {
-        case e: TopLevelJSClassExportDef =>
-          WithGlobals(genTopLevelJSClassExportDef(tree, e))
-        case e: TopLevelModuleExportDef =>
-          WithGlobals(genTopLevelModuleExportDef(tree, e))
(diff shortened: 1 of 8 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1075,12 +1075,10 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
 
   def genTopLevelJSClassExportDef(cd: LinkedClass,
       tree: TopLevelJSClassExportDef): js.Tree = {
```
gzm0: I think you can inline this method above.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1090,62 +1088,45 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
    */
   def genTopLevelModuleExportDef(cd: LinkedClass,
       tree: TopLevelModuleExportDef): js.Tree = {
```
gzm0: I think you can inline this method above.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
sjrd: Done.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
sjrd: Done

PR #3488, opened 2018-11-10
Fix #2175: Add support for ECMAScript 2015 modules.

### LINKED ISSUE #2175: Integrating with JS module systems

sjrd:
This is a port of b744d12e0c6d8af74960e0ce5071c8e0011249a5.

This commit adds a third `ModuleKind` for ES modules, namely `ModuleKind.ESModule`. When emitting an ES module, `@JSImport`s and `@JSExportTopLevel`s straightforwardly map to ES `import` and `export` clauses, respectively.

At the moment, imports are always implemented using a namespace import, then selecting fields inside the namespace. This is suboptimal because it can prevent advanced DCE across ES modules. Improving on this is left for future work.

The Node.js-based environment is adapted to interpret files whose name ends with `.mjs` as ES modules rather than scripts. This aligns with how Node.js itself identifies ES modules as of version 10.x, although it is still experimental, so that could change in the future.

This mechanism goes against the API of `Input`, which is supposed to address this concern. This needs to be fixed.

Although setting `scalaJSLinkerConfig.moduleKind` to `ModuleKind.ESModule` is enough for the Scala.js linker to emit a valid ES module, two additional settings are required to *run* or *test* using Node.js:
```scala
    artifactPath in (proj, Compile, fastOptJS) :=
      (crossTarget in (proj, Compile)).value / "somename.mjs"

    jsEnv := {
      new org.scalajs.jsenv.NodeJSEnv(
          org.scalajs.jsenv.NODEJSEnv.Config()
            .withArguments(List("--experimental-modules"))
      )
    }
```
The first setting is necessary to give the `.mjs` extension to the file produced by Scala.js, which in turn is necessary for Node.js to accept it as an ES module.

The second setting will be necessary until Node.js declares its support for ES module as non-experimental.

The version of the Closure Compiler that we use does not support ES modules yet, so we deactivate GCC when emitting an ES module.

At this point, the emission of ES modules can be considered stable, but the support in `NodeJSEnv` is experimental (since the support of ES modules in Node.js is itself experimental).

Running the full test suite with ES modules requires Node.js 10.2.0 or later. It has been tested with v10.12.0.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -112,6 +112,18 @@ def Tasks = [
         'set scalaJSLinkerConfig in helloworld ~= (_.withModuleKind(ModuleKind.CommonJSModule))' \
         helloworld/run \
         helloworld/clean &&
+    sbtretry ++$scala \
+        'set artifactPath in (helloworld, Compile, fastOptJS) := (crossTarget in helloworld).value / "helloworld-fastopt.mjs"' \
+        'set jsEnv in helloworld := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in helloworld ~= (_.withModuleKind(ModuleKind.ESModule))' \
+        helloworld/run &&
+    sbtretry ++$scala \
+        'set artifactPath in (helloworld, Compile, fullOptJS) := (crossTarget in helloworld).value / "helloworld-opt.mjs"' \
+        'set jsEnv in helloworld := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in helloworld ~= (_.withModuleKind(ModuleKind.ESModule))' \
+        'set scalaJSStage in Global := FullOptStage' \
+        helloworld/run \
+        helloworld/clean &&
     sbtretry ++$scala testingExample/testHtml &&
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
         ++$scala testingExample/testHtml \
@@ -169,6 +181,18 @@ def Tasks = [
     sbtretry 'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.CommonJSModule))' \
         ++$scala $testSuite/test &&
     sbtretry 'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.CommonJSModule))' \
+        'set scalaJSStage in Global := FullOptStage' \
+        ++$scala $testSuite/test \
+        $testSuite/clean &&
+    sbtretry 'set artifactPath in ($testSuite, Test, fastOptJS) := (crossTarget in $testSuite).value / "testsuite-fastopt.mjs"' \
+        'set jsEnv in $testSuite := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.ESModule))' \
+        'set MyScalaJSPlugin.wantSourceMaps in $testSuite := false' \
+        ++$scala $testSuite/test &&
+    sbtretry 'set artifactPath in ($testSuite, Test, fullOptJS) := (crossTarget in $testSuite).value / "testsuite-opt.mjs"' \
+        'set jsEnv in $testSuite := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.ESModule))' \
+        'set MyScalaJSPlugin.wantSourceMaps in $testSuite := false' \
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala $testSuite/test
   ''',
@@ -242,6 +266,18 @@ def Tasks = [
         ++$scala $testSuite/test &&
     sbtretry 'set scalaJSLinkerConfig in $testSuite ~= (_.withESFeatures(_.withUseECMAScript2015(true)))' \
         'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.CommonJSModule))' \
+        'set scalaJSStage in Global := FullOptStage' \
+        ++$scala $testSuite/test \
+        $testSuite/clean &&
+    sbtretry 'set scalaJSLinkerConfig in $testSuite ~= (_.withESFeatures(_.withUseECMAScript2015(true)))' \
+        'set artifactPath in ($testSuite, Test, fastOptJS) := (crossTarget in $testSuite).value / "testsuite-fastopt.mjs"' \
+        'set jsEnv in $testSuite := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.ESModule))' \
+        ++$scala $testSuite/test &&
+    sbtretry 'set scalaJSLinkerConfig in $testSuite ~= (_.withESFeatures(_.withUseECMAScript2015(true)))' \
+        'set artifactPath in ($testSuite, Test, fullOptJS) := (crossTarget in $testSuite).value / "testsuite-opt.mjs"' \
+        'set jsEnv in $testSuite := new org.scalajs.jsenv.nodejs.NodeJSEnv(org.scalajs.jsenv.nodejs.NodeJSEnv.Config().withArgs(List("--experimental-modules")).withSourceMap(false))' \
+        'set scalaJSLinkerConfig in $testSuite ~= (_.withModuleKind(ModuleKind.ESModule))' \
         'set scalaJSStage in Global := FullOptStage' \
         ++$scala $testSuite/test
   ''',
--- a/js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
@@ -29,6 +29,16 @@ object Input {
   /** All files are to be loaded as scripts into the global scope in the order given. */
   final case class ScriptsToLoad(scripts: List[VirtualBinaryFile]) extends Input
 
+  /** All files are to be loaded as ES modules, in the given order.
+   *
+   *  Some environments may not be able to execute several ES modules in a
+   *  deterministic order. If that is the case, they must reject an
+   *  `ESModulesToLoad` input if the `modules` argument has more than one
+   *  element.
+   */
+  final case class ESModulesToLoad(modules: List[VirtualBinaryFile])
+      extends Input
+
   /** All files are to be loaded as CommonJS modules, in the given order. */
   final case class CommonJSModulesToLoad(modules: List[VirtualBinaryFile])
       extends Input
--- a/linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
+++ b/linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
@@ -44,12 +44,15 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
   import config.commonConfig.coreSpec._
 
   require(!esFeatures.useECMAScript2015,
-      s"Cannot use features $esFeatures with the Closure Compiler" +
+      s"Cannot use features $esFeatures with the Closure Compiler " +
       "because they contain ECMAScript 2015 features")
 
   require(!esFeatures.allowBigIntsForLongs,
-      s"Cannot use features $esFeatures with the Closure Compiler" +
+      s"Cannot use features $esFeatures with the Closure Compiler " +
       "because they allow to use BigInts")
+
+  require(moduleKind != ModuleKind.ESModule,
+      s"Cannot use module kind $moduleKind with the Closure Compiler")
 
   private[this] val emitter = {
     new Emitter(config.commonConfig)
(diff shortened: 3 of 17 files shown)


sjrd:
@gzm0 I could use some advice on how to leverage `Input` to accommodate modules better in `NodeJSEnv`. At the moment I've just ported the mechanism from 0.6.x, which is that `NodeJSEnv` interprets "scripts" whose name ends with `.mjs` as modules.

My first instincts was to have `Input.ModulesToLoad(List[VirtualBinaryFile])`. But if I do that, I don't know what to do with `jsExecutionFiles` in the sbt plugin. At the very least, it looks like we should have an other task `jsEnvInput` which would be a `Task[Input]`. `jsEnvInput` could default to `Input.ScriptsToLoad(jsExecutionFiles.value)`, but that doesn't answer the question of what to do when we actually want to load scripts *and* modules.

gzm0:
`jsExecutionFiles` probably needs to be entirely replaced by `jsInput`. As you might observe, nothing in our plugin / setup relies on `jsExecutionFiles` being modifiable anymore.

> but that doesn't answer the question of what to do when we actually want to load scripts and modules.

We cannot do this. IMO we shouldn't lie and pretend there's a way to load scripts and modules. `JSEnv`'s simply don't support this (apart from loading them from a main file, which the user then should do).

The way I see it, we should have:

```scala
class Modules(mainModule: VirtualFile, availbableModules: Map[String, VirtualFile])
```

or something along these lines. This represents what the `JSEnv`s are actually able to do.

sjrd:
> We cannot do this. IMO we shouldn't lie and pretend there's a way to load scripts and modules. JSEnv's simply don't support this (apart from loading them from a main file, which the user then should do).

Hum, by that argument, even the current `ScriptsToLoad(xs)` is a lie. Node.js cannot load several scripts without a main file, which is what the existing `NodeJSEnv` generates (on the std in).

And then it depends on the JS env. For example a DOM-based environment such as `JSDOMNodeJSEnv` would be perfectly able to load scripts and modules (with `<script type="module">`).

> `class Modules(mainModule: VirtualFile, availbableModules: Map[String, VirtualFile])`

That's more than what JSEnvs can do. In general we can't remap arbitrary strings to files. It depends on how the JSEnv resolves module names. Some of them allow to hook up our own resolver, I think, but that would go far ahead. If that's what you propose, I would grind it down to
```scala
class Module(module: VirtualFile)
```

gzm0:
> Hum, by that argument, even the current ScriptsToLoad(xs) is a lie

Absolutely. IMO we should adjust NodeJSEnv to only allow `ScriptsToLoad` with zero or one elements.

> If that's what you propose, I would grind it down to `class Module(module: VirtualFile)`

That would also work for me. However, JSEnvs that cannot re-map resolvers could simply reject any input that contains a non empty map.

sjrd:
Rebased, with proper handling of `Input`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1102,6 +1101,12 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
       case ModuleKind.NoModule =>
         genAssignToNoModuleExportVar(exportName, exportedValue)
 
+      case ModuleKind.ESModule =>
+        val field = envField("e_" + exportName)
```
gzm0: `envField("e", exportName)`? There seems to be an overload with a `subfield` parameter.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -164,6 +165,37 @@ object NodeJSEnv {
...
             writeRequire(module)
+
+        case Input.ESModulesToLoad(modules) =>
+          if (modules.nonEmpty) {
+            val uris = modules.map {
+              case module: FileVirtualBinaryFile =>
+                module.file.toURI
+              case module =>
+                tmpFile(module.path, module.inputStream).toURI
+            }
```
gzm0: Isn't there a requirement on the filename? If yes, you should check it.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -164,6 +165,37 @@ object NodeJSEnv {
...
+
+            val importerFileContent = {
+              s"""
+                |$importChain.catch(e => {
+                |  /* Make sure to fail the process, but give time to Node.js to
+                |   * display a good stack trace before that.
+                |   */
+                |  setTimeout(() => process.exit(1), 100);
+                |  throw e;
```
gzm0: You can simply `console.log(e)` and then exit. See [this answer](https://stackoverflow.com/a/33593443/1149944)

js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
```diff
@@ -29,6 +29,16 @@ object Input {
...
 
+  /** All files are to be loaded as ES modules, in the given order.
+   *
+   *  Some environments may not be able to execute several ES modules in a
+   *  deterministic order. If that is the case, they must reject an
+   *  `ESModulesToLoad` input if the `modules` argument has more than one
+   *  element.
+   */
+  final case class ESModulesToLoad(modules: List[VirtualBinaryFile])
+      extends Input
```
gzm0: Is this really a reasonable loading strategy to add? It seems maybe we should just allow loading a single module and force the callers to make things load hierarchically if necessary?

We shouldn't take additions here lightly as we push the burden of supporting them on all JSEnvs. I'm still wondering the same about `CommonJSModulesToLoad` by the way, but here it is more flagrant since generating a simple `import `chain is not even enough.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -59,8 +59,8 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
   val symbolRequirements: SymbolRequirement = emitter.symbolRequirements
 
   private val needsIIFEWrapper = moduleKind match {
-    case ModuleKind.NoModule       => true
-    case ModuleKind.CommonJSModule => false
+    case ModuleKind.NoModule                             => true
+    case ModuleKind.ESModule | ModuleKind.CommonJSModule => false
```
gzm0: Could you please guard the Closure backend from being invoked with `ESModule`? I guess currently this would just generate a match error if the closure tree converter finds an `import` / `export`. Not ideal.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
```diff
@@ -192,6 +224,17 @@ object NodeJSEnv {
...
+    import java.nio.file.{Files, StandardOpenOption}
+
+    val f = createTmpFile(path)
+    val contentList = new java.util.ArrayList[String]()
+    contentList.add(content)
+    Files.write(f.toPath(), contentList, StandardCharsets.UTF_8,
+        StandardOpenOption.TRUNCATE_EXISTING)
+    f
+  }
+
```
gzm0: Ah, now it's needed :)

project/Build.scala
```diff
@@ -1615,6 +1626,63 @@ object Build {
       scalaJSLinkerConfig ~= { _.withSemantics(TestSuiteLinkerOptions.semantics _) },
       scalaJSModuleInitializers in Test ++= TestSuiteLinkerOptions.moduleInitializers,
 
+      /* The script that calls setExportsNamespaceForExportsTest to provide
+       * ExportsTest with a loopback reference to its own exports namespace.
+       * Only when using an ES module.
+       * See the comment in ExportsTest for more details.
+       */
+      setModuleLoopbackScript in Test := Def.settingDyn[Task[Option[FileVirtualBinaryFile]]] {
```
gzm0: I really don't like this, but I don't have a better idea either :(

Jenkinsfile
```diff
@@ -112,6 +112,18 @@ def Tasks = [
         'set scalaJSLinkerConfig in helloworld ~= (_.withModuleKind(ModuleKind.CommonJSModule))' \
         helloworld/run \
         helloworld/clean &&
+    sbtretry ++$scala \
+        'set artifactPath in (helloworld, Compile, fastOptJS) := (crossTarget in helloworld).value / "helloworld-fastopt.mjs"' \
```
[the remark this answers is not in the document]
sjrd: Because that would tie the behavior of the sbt plugin to something that's 1) Node.js-specific and 2) even still experimental in Node.js. As long as it's experimental and/or not widespread, I'd rather users do that themselves.

js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
sjrd: If you put several `<script type="module">` in a .html page, they will be loaded in sequence, as specified by this strategy. Therefore, this strategy is a good proxy for how front-end applications will behave. Also, for the same reason, it's going to be trivial to implement on all JSEnvs that provide a browser-like environment (including Selenium and jsdom; PhantomJS will probably never support modules at all).

If we *don't* support this, then users will be burdened if they need to replicate in their testing environment the loading sequence of several modules that they have in their web page. When using `NodeJSEnv`, they could do what we do currently in `NodeJSEnv`, but if they use a browser-like JSEnv, they'll be hard-pressed to reproduce the setup!

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
sjrd: [Not really](https://nodejs.org/api/esm.html#esm_interop_with_existing_modules), although if it doesn't have the extension `.mjs` it will be parsed as a CommonJS module, then *exposed* as an ES module. So it can do something meaningful, and we should not prevent users from using that behavior of Node.js, IMO, at least as long as it's experimental in Node.js: giving more freedom to user means they are less dependent on us releasing new versions if and when Node.js changes. I suggest we reconsider once Node.js settles on a final design.

sjrd: To give a bit more sense to the behavior, I was also considering renaming the 3 `Input` types with a more "action"-oriented name, such as:

* `RunScripts`
* `ImportModules`
* `RequireModules`

Then it's clearer what those operations map to, and that the burden of making sure the target files make sense given the JSEnv's peculiarities is in the hands of the user.

gzm0: I think we are thinking about this the wrong way around. We should specify what the file contains, not how to load it. Otherwise the `Input`'s semantics is not independent of the `JSEnv` anymore.

So I would propose the opposite: Irrespective of the extension of the virtual file, interpret is as specified in the input, renaming it if necessary.

Further, it seems that we will always have a need to load multiple things, so maybe we should change `Input` to a single file but repeated. (Although this will make it harder to add an "environment" with other scripts that are available for loading).

In any case, I'm OK with `ESModulesToLoad` in this PR, but I would like to see some effort to avoid proliferation of different `Input` types going forward; as it stands now, I think we are adding them too easily given the burden they put on downstream maintainers (including us).

Lastly, I just realized we should make `NodeJSEnv` more aware of the coupling of `--experimental-modules` and `ESModulesToLoad`. Maybe enable it automatically or give it a higher level option, failing if it is not set.

Jenkinsfile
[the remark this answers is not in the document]
gzm0: Depending on the outcome of the other comment (rename if necessary) this might not be necessary at all anymore. I prefer this, as it would hide NodeJS details inside NodeJS.

nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala
sjrd: Renaming is not possible. That would break links from other files to the renamed file, which can go up to loading twice the same module (because it's *not* the same module anymore if it doesn't have the same URL).

> Further, it seems that we will always have a need to load multiple things, so maybe we should change Input to a single file but repeated.

Probably, yes.

> (Although this will make it harder to add an "environment" with other scripts that are available for loading).

I don't think we need that anyway.

> In any case, I'm OK with ESModulesToLoad in this PR, but I would like to see some effort to avoid proliferation of different Input types going forward; as it stands now, I think we are adding them too easily given the burden they put on downstream maintainers (including us).

I've filed #3506 so that we revisit that.

> Lastly, I just realized we should make NodeJSEnv more aware of the coupling of `--experimental-modules` and `ESModulesToLoad`. Maybe enable it automatically or give it a higher level option, failing if it is not set.

I really don't think so. If say Node.js 12 suddenly supports modules by default, and drops the flag (which they *do* tend to do), we'll block users from upgrading to Node.js 12 until we can release a new version of Scala.js that does not add/require `--experimental-modules` (and if we did add it automatically before, that would break earlier versions of Node.js). Again, as long as it's experimental in Node.js, I would leave `NodeJSEnv` as dumb as possible and let users fill in the blanks.

gzm0: > Renaming is not possible. That would break links from other files to the renamed file, which can go up to loading twice the same module (because it's not the same module anymore if it doesn't have the same URL).

Hmmm... that is fair. Maybe it is entirely wrong to allow virtual files in there. Maybe we should take file names instead and allow a different input type that is literal code...

Let's go ahead with this and defer to #3506 then.

project/Build.scala
gzm0: Hah, turns out this works:

`test.mjs`
```javascript
export const blah = 1;

import loopback from './loopback.mjs';

console.log(loopback.blah);
```

`loopback.mjs`
```javascript
import * as test from './test.mjs';

export default test;
```

Did you try this? (not sure it works in the more involved setting we have).

sjrd: That looks more elegant, indeed.

I'm not sure how to set it up, though. The biggest issue is that we have fastOpt and fullOpt. The loopback module must be different for each, since it must import a different `testsuite-*opt.mjs`. And then that means that the `@JSImport("loopback.mjs", ...)` inside the test suite must be different for fastOpt and fullOpt. And that is quite annoying to set up. We also need to avoid the mere presence of that `@JSImport` when using NoModule or CommonJS. We've precedent for something similar, but it's yet another burden.

I'm also not *convinced* that the circular import trick does not suffer from the asynchronous theoretical problem. After all, circular bindings need to be initialized in *some* order. I would have to dig quite deep into the spec to ascertain that.

I'm not sure it's worth the trouble.

gzm0: Aight, fair enough. Let's go ahead with this and revisit if we encounter issues.

PR #3489, opened 2018-11-11
Handle cancellation correctly in 1.x's `ComRun`

jvican:
In 0.6.x, calling `close` on the test adapter always kills the
underlying processes that are instantiated by the js environments.

Unlike 0.6.x, the latest 1.x test module doesn't kill the underlying
processes. This causes a resource leak that is fatal in Windows
environments (where the underlying process keeps a reference to the
output JS file indefinitely and prevents other tasks in the build tool
from linking again).

This pull request does the following:

1. Invokes `run.close()` in `ComRun`'s' close method.
2. Only runs `close` once, as `ComRun.close` can be called several times.

We're testing cancellation in Bloop for both Scala.js 0.6.x and 1.x and,
after these changes, all tests pass.

### DIFF
--- a/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
+++ b/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala
@@ -119,16 +119,21 @@ private final class ComRun(run: JSRun, handleMessage: String => Unit,
 
   def close(): Unit = synchronized {
     val oldState = state
+    if (oldState == Closing) () // Do nothing if old state is already closing
+    else {
+      // Close the underlying run
+      run.close()
 
-    // Signal receiver thread that it is OK if socket read fails.
-    state = Closing
+      // Signal receiver thread that it is OK if socket read fails.
+      state = Closing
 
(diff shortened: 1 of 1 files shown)


sjrd:
`> nodeJSEnv/test` doesn't seem to pass on the CI.

gzm0:
Heed the comment in the [receiver thread][1]. tl;dr: you cannot close the run immediately when `close` is called, otherwise killing the process might surface a failure. OTOH we cannot ignore failures, since otherwise we'll race for whether we catch a failure or not. The JS side [should terminate][2] once the socket is closed.

[1]: https://github.com/scala-js/scala-js/blob/f35af07a58439563b828e108104b2a938d6ec862/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala#L85-L88
[2]: https://github.com/scala-js/scala-js/blob/f35af07a58439563b828e108104b2a938d6ec862/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala#L289

gzm0:
Closing this because there has not been any follow-up.

PR #3490, opened 2018-11-16
In `NoModule`, use true top-level `var`s for top-level exports.

sjrd:
Before, we needed to detect the global object at run-time (which is not fool-proof) to add top-level exports as properties of the global object. Now, top-level exports are truly defined as global `var`s.

When using ECMAScript 2015 features, we emit global `let`s instead. This is trivial to implement, but annoying to test. `ExportsTest` now duplicates every interaction with the "exports namespace", since in ES 2015 `NoModule` the only way to access the exports is as `global.TheExport`, which cannot be abstracted away. This results in quite a bit of code duplication in the tests, but it can't be helped.

### DIFF
--- a/js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
+++ b/js-envs/src/main/scala/org/scalajs/jsenv/Input.scala
@@ -30,7 +30,7 @@ object Input {
   final case class ScriptsToLoad(scripts: List[VirtualBinaryFile]) extends Input
 }
 
-case class UnsupportedInputException(msg: String, cause: Throwable)
+class UnsupportedInputException(msg: String, cause: Throwable)
     extends IllegalArgumentException(msg, cause) {
   def this(msg: String) = this(msg, null)
   def this(input: Input) = this(s"Unsupported input: $input")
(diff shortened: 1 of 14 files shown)


sjrd:
For when you come back from vacation ;)

sjrd:
> and that keeping `__ScalaJSEnv` is the right thing to do

`__ScalaJSEnv` is definitely not right. It's just the least bad we've got so far. It's orthogonal to this PR, though.

gzm0:
> It's orthogonal to this PR, though.

Well, yes and no. It relates to this PR in the sense that the PR introduces / changes a bunch of machinery solely for the purpose of supporting `__ScalaJSEnv` (the whole JSEnv loading stuff).

sjrd:
Hum, no, not at all. The changes in `JSEnv` and `Input` are entirely for the exports as `var`s or `let`s. The changes to `__ScalaJSEnv` in the build are then made to accommodate the changes in `JSEnv` (not the other way around).

sjrd:
Updated. (separate commit for `UnsupportedInputException`, and remove dead code `tmpFile`)

PR #3501, opened 2018-11-30
Drop support for 2.13.0-M3 and 2.13.0-M4.

sjrd:
The commit also drops a few references to 2.12.0-M4 that were lingering in comments and assumptions.


project/Build.scala
```diff
@@ -71,24 +71,21 @@ object Build {
...
+   *  2.12.x).
    */
   def collectionsEraDependentDirectory(scalaV: String, sourceDir: File): File =
-    if (scalaV == "2.13.0-M4") sourceDir / "scala-m4-collections"
-    else if (hasNewCollections(scalaV)) sourceDir / "scala-new-collections"
+    if (hasNewCollections(scalaV)) sourceDir / "scala-new-collections"
     else sourceDir / "scala-old-collections"
 
   val scalaVersionsUsedForPublishing: Set[String] =
-    Set("2.10.7", "2.11.12", "2.12.6", "2.13.0-M3")
```
gzm0: Shouldn't this be changed to M5 and maybe M5 included below?

scripts/publish.sh
```diff
@@ -7,15 +7,17 @@ else
...
+BIN_VERSIONS="2.10.7 2.11.12 2.12.6"
+LIMITED_VERSIONS="2.13.0-M5"
 CLI_VERSIONS="2.10.7 2.11.12 2.12.6"
 SBT_VERSION="2.10.7"
 SBT1_VERSION="2.12.6"
 SBT1_SBTVERSION="1.0.0"
 
 COMPILER="compiler jUnitPlugin"
 LIBS="library javalibEx ir irJS tools toolsJS jsEnvs jsEnvsTestKit testAdapter stubs testInterface jUnitRuntime"
+LIMITED="compiler jUnitPlugin library javalibEx stubs testInterface jUnitRuntime"
```
gzm0: Could you comment on what limited means? (IIUC it is the bare minimum to compile SJS *for* 2.13 but not *in* 2.13).

sjrd: That's right. It's called Limited here because that's how it was called in `Jenkinsfile` (`limitedCIScalaVersions`). Should I call it `NOTOOLS` or something like that?

project/Build.scala
sjrd: Hum, yes, we can probably add M5 here. Not sure what you mean by "below"?

sjrd: Ah I see. No it shouldn't be in `newScalaBinaryVersionsInThisRelease`, since we released 0.6.26 with the 2.13.0-M5 artifacts.

It is actually pretty rare for `newScalaBinaryVersionsInThisRelease` to be non-empty, because we always try to back-publish old versions of Scala.js for newer versions of Scala. `git blame` says that the last time it was non-empty was during 0.6.16 for Scala 2.13.0-M1 (https://github.com/scala-js/scala-js/commit/d39a999607ae8f16d1bca3a2e2c18b793aaae9e7).

scripts/publish.sh
sjrd: Done.

PR #3505, opened 2018-12-07
Fix #3504: Remove __ScalaJSEnv

### LINKED ISSUE #3504: Remove __ScalaJSEnv
There are currently 3 usages:

- `exitFunction` to fill `System.exit`; this was introduced for #958 as a means to terminate PhantomJS in async tests. Testing support has improved substantially since then, so it is not needed anymore. If a platform agnostic termination is required a library should be created for it.
- `sourceMapper` introduced as a stop-gap solution to support source mapping in Rhino. Unused.
- `javaSystemProperties` introduced as a hack to allow to pass in tags for testing. Already made more SJS internal in the past (by not honoring `javaOptions` in sbt anymore). Should be replaced by a fully SJS internal solution.

Any objections?

gzm0:


test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemJSTest.scala
```diff
@@ -85,29 +85,6 @@ class SystemJSTest {
...
-    if (compliantAsInstanceOf) {
-      try {
-        Int.box(5).asInstanceOf[String]
-        fail("Invalid class cast succeeded in presence of " +
-            "\"compliant-asinstanceofs\" flag.")
-      } catch {
-        case _: ClassCastException => // As expected
-      }
-    }
-    assertEquals(compliantAsInstanceOf, Platform.hasCompliantAsInstanceOfs)
```
sjrd: It's a bit sad that those tests are going away. They were useful to assert that the `linkingInfo` were indeed populated correctly, and that they would agree with the flags independently computed by the build.

I'm not sure what else to do, though.

linker/scalajsenv.js
```diff
@@ -48,6 +48,13 @@ const $linkingInfo = {
     "productionMode": false
 //!endif
   },
+//!if moduleKind == NoModule
+  "moduleKind": 0,
+//!else
+//!if moduleKind == CommonJSModule
+  "moduleKind": 1,
+//!endif
+//!endif
```
sjrd: That change, together with the facade in `LinkingInfo.scala`, actually constitutes a user-observable feature. It should be in its own commit, without at the same touching `Platform` and `SystemJSTest`.

project/Build.scala
```diff
@@ -63,8 +63,6 @@ object MyScalaJSPlugin extends AutoPlugin {
   val isGeneratingEclipse =
     Properties.envOrElse("GENERATING_ECLIPSE", "false").toBoolean
 
-  val wantSourceMaps = settingKey[Boolean]("Whether source maps should be used")
```
sjrd: Er, we really cannot do that. [It's used in the community build.](https://github.com/scala/community-builds/blob/c1e7bb720d876954d94207939b3367ae2c6f9f5c/configs/community.dbuild#L427). :-s

linker/scalajsenv.js
sjrd: I'm actually hesitant about exposing this information. I cannot quite figure out what bothers me about it. Maybe the fact that the set of possible values will change with time, unlike the other linking info fields. Or that it might not be meaningful at all for completely different kinds of outputs such as Wasm.

Could we put this information in the `BuildInfo` like fastOpt/fullOpt and `scalaVersion` instead? And then revisit later if we find a user-level use case for knowing `moduleKind` at run-time?

gzm0: Absolutely.

project/BuildInfoForTestGenerator.scala
```diff
@@ -0,0 +1,35 @@
...
+        final val scalaVersion = raw\"\"\"${scalaVersion}\"\"\"
+        final val isNoModule = ${moduleKind == ModuleKind.NoModule}
+        final val isCommonJSModule = ${moduleKind == ModuleKind.CommonJSModule}
+        final val isFastOpt = ${stage == Stage.FastOpt}
+        final val isFullOpt = ${stage == Stage.FullOpt}
+        final val hasSourceMaps = $wantSourceMaps
+      }
+      """
+
+    IO.write(trg, scalaCode)
```
sjrd: All those generators should make an effort not to write to the file if it already exists and its content matches what we're about to write. Otherwise they'll be recompiled and re-optimized on every single compilation.

gzm0:
I am splitting this PR. (#3520, #3521 are the first two).

gzm0:
> It's a bit sad that those tests are going away. They were useful to assert that the linkingInfo were indeed populated correctly, and that they would agree with the flags independently computed by the build.

I have added a test that actually does this. The previous tests didn't.

> All those generators should make an effort not to write to the file if it already exists and its content matches what we're about to write. Otherwise they'll be recompiled and re-optimized on every single compilation.

It seems that this should happen, but it doesn't (re-running `testSuite/test` does *not* re-trigger fast opt). Maybe sbt's detection has improved in this regard? Hence, I haven't changed this yet, but still can if you want me to.

gzm0:
Updated with a fix to take the right config based on stage. Note that this means parts of the tests need to be recompiled if the stage changes, but I think this is acceptable.

project/Build.scala
```diff
@@ -1686,6 +1508,35 @@ object Build {
         }
       },
 
+      sourceGenerators in Compile += Def.task {
+        val stage = scalaJSStage.value
+
+        val linkerConfig = stage match {
+          case FastOptStage => (scalaJSLinkerConfig in fastOptJS).value
+          case FullOptStage => (scalaJSLinkerConfig in fullOptJS).value
```
sjrd: I think you need `(scalaJSLinkerConfig in (Compile, fastOptJS)).value` (and same for `fullOptJS`), because the above would always fallback to `scalaJSLinkerConfig.value`, since we only redefine it `in (SomeConfig, fullOptJS)`.

gzm0:
Updated.

PR #3509, opened 2018-12-13
Remove the hack to fix percentage signs in Node.js.

sjrd:
Back when that hack was introduced in 1fc8d0526ba9b46049c0deefc3fe3fa93d3271b0, Node.js' `console.log` deduplicated consecutive `%` signs in its argument, even if it was a single argument. At the time we thought that was intentional, and we added our hack to have it behave in a more standard way for Scala.js.

The deduplication with only 1 argument turned out to be a bug, recognized as such in https://github.com/nodejs/node/issues/3396, when it was accidentally fixed in Node.js 4. It is now the documented behavior, see https://nodejs.org/api/util.html#util_util_format_format_args which is referenced from https://nodejs.org/api/console.html#console_console_log_data_args.


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -95,50 +95,19 @@ private[test] class RunTests(config: JSEnvSuiteConfig, withCom: Boolean) {
...
-
-    withRun(codes.mkString("")) {
+    withRun(code) {
       _.expectOut(result)
         .closeRun()
     }
   }
 
-  @Test // Node.js console.log hack didn't allow to log non-Strings - #561
-  def nonStringTest: Unit = {
```
gzm0: I think this test is still useful, no?

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -95,50 +95,19 @@ private[test] class RunTests(config: JSEnvSuiteConfig, withCom: Boolean) {
...
+    val result = strings.mkString("", "\n", "\n")
 
-    val result = strlists.map(_.mkString(" ") + "\n").mkString("")
-
-    withRun(codes.mkString("")) {
+    withRun(code) {
       _.expectOut(result)
         .closeRun()
     }
   }
```
gzm0: Could you refactor this in a separate commit so we don't accidentally introduce a regression?

sjrd:
Hum, I should probably have elaborated more on the commit message.

It's not in 0.6.x because it *is* a breaking change. Our hack was too eager, and was doing too much. In trying the solve the case where only one string argument with `%%` is passed to `console.log`, it "solved" a few other things:

* It was deduplicating `%%` signs also when more than one argument was given
* It was forcing the conversion of the first argument to be a `string`, preventing Node.js' default `console.log` behavior of displaying non-strings in a more elaborate way than just their `toString()`.

So removing the hack altogether restores those two behaviors to the Node.js defaults. Since the tests were effectively testing our modified behavior, the tests need to change (and in the same commit).

All we really care about is that if we pass in *1 string argument*, that it is sent to stdout as-is, without processing any `%`. Cases where we pass more than 1 argument, or where we pass a non-string argument, are of no concern to Scala.js per se, and should be left to the particular engine to deal with as they wish.

gzm0:
Makes sense.

PR #3513, opened 2018-12-17
Upgrade to Google Closure Compiler v20181210.

### LINKED ISSUE #2749: When we update GCC: Try replacing mutable.Map by js.Dictionary in scalajs.reflect.Reflect

### LINKED ISSUE #2815: run/Predef.readLine.scala is broken in --fullOpt
The change in partest to use a module initializer rather than export to call the launcher broke `run/Predef.readLine.scala` in `--fullOpt`. Somehow this simplified the program enough that Closure could see a bit more ... and break things because it's buggy.

How to reproduce:

    > partestSuite/testOnly -- --fullOpt run/Predef.readLine.scala

Source of the test: [`run/Predef.readLine.scala`](https://github.com/scala/scala/blob/2.12.x/test/files/run/Predef.readLine.scala)

Error:
```
[error] TypeError: Cannot read property 'eb' of null
[error]     at sf (...\java\io\BufferedReader.scala:71:15)
[error]     at Object.<anonymous> (...\helloworld-opt.js:3409:3)
[error]     at Object.<anonymous> (...\helloworld-opt.js:3470:4)
[error]     at Module._compile (module.js:541:32)
[error]     at Object.Module._extensions..js (module.js:550:10)
[error]     at Module.load (module.js:458:32)
[error]     at tryModuleLoad (module.js:417:12)
[error]     at Function.Module._load (module.js:409:3)
[error]     at Module.require (module.js:468:17)
[error]     at require (internal/module.js:20:19)
```
After some digging, I found that GCC removes the assignment of the `new BufferedReader` (created by [`Console.withIn`](https://github.com/scala-js/scala-js/blob/502bde1b1d6046ea531df95abba64b61265168ab/scalalib/overrides-2.11/scala/Console.scala#L112)) to the `value` field of the `DynamicVariable`. Hence, it stays the `null` [it was initialized with](https://github.com/scala-js/scala-js/blob/502bde1b1d6046ea531df95abba64b61265168ab/scalalib/overrides-2.11/scala/Console.scala#L25).

Later, the calls to `readLine()` eventually *reads* that field, calls a method on it, and kaboom!

This is clearly a bug in GCC. It looks similar to the one that led to #2749.

sjrd:
Besides benign API changes, the biggest change is that this new version of GCC will not infer global variables that are accessed, and instead raise errors if they are not declared. This means that we need to track the set of *all* referenced global variables when targeting GCC, rather than just the set of dangerous ones. We do that with a new option in `InternalOptions`.

Because of too much lockdown in the ErrorManager API of Closure, we cannot send the errors and warnings to our `logger` anymore. Instead, they will be sent to the standard error, if any. Note that in theory, this never happens.

This commit fixes #2815 as a side effect: the new version of GCC does not exhibit the bug that caused that issue.

After that, we also trivially fix #2749.


sjrd:
Ready for review :)

sjrd:
Full build: https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3513/9/pipeline

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -96,17 +98,28 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
...
+  private def makeExternsForGlobalRefs(globalRefs: Set[String]): String =
+    globalRefs.map("var " + _ + ";\n").mkString
+
   /** Constructs an externs file listing all exported properties in a linking
    *  unit.
    *
    *  This is necessary to avoid name clashes with renamed properties (#2491).
    */
-  private def makeExternsForExports(linkingUnit: LinkingUnit): String = {
+  private def makeExternsForExports(topLevelVarDeclarations: Option[String],
```
gzm0: I'm not sure I understand this: Why are *all* top level var declarations externs for exports? Aren't some top level var decls purely internal to the file?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -96,17 +98,28 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
           closureExterns.asJava, List(module).asJava, options)
     }
 
+    if (result.errors.nonEmpty) {
```
gzm0: What does `result.errors` contain? Can't we send that to the logger?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -133,7 +148,12 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
 
   private def closureCompiler(logger: Logger) = {
     val compiler = new ClosureCompiler
-    compiler.setErrorManager(new LoggerErrorManager(logger))
+    /* FIXME We used to leverage `compiler.setErrorManager` to send errors and
+     * warnings to our `logger`. However, that API has been locked down a bit
+     * too much in the current version of GCC, so we cannot do that anymore.
+     * As it stands, errors and warnings will be send to the standard error.
+     */
+    //compiler.setErrorManager(new LoggerErrorManager(logger))
```
gzm0: I doubt this is going to get fixed especially if this shouldn't happen if there are no bugs. So I think just letting the errors go to stderr and removing this comment altogether is fine.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -201,20 +225,28 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
 private object ClosureLinkerBackend {
   /** Minimal set of externs to compile Scala.js-emitted code with Closure. */
   private val ScalaJSExterns = """
-    /** @constructor */
```
gzm0: If these comments are not necessary anymore, we should maybe also remove them from the tree builder?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: Yeah, right. I had it all wrong. It turns out it's quite possible to send things to the logger. It's just slightly suboptimal in the ordering, but it works. See the updated code.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: Because top-level var declarations are added after the fact *outside* of the IIFE that wraps the entire result of GCC. So from the point of view of GCC, all top-level exports are externs.

Recall that `topLevelVarDeclarations` is the set of names that are top-level-exported when in `NoModule`---empty otherwise. It was introduced in #3490.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: Fixed.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
sjrd: They are still necessary for `function` declarations, which is what we use in the actual code. For externs I replaced them by `var`s so that they do not conflict with externs generated for global variables and top-level exports.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: Right; so this actually makes perfect sense.

PR #3514, opened 2018-12-18
Fix #3181: Enable GCC when using ECMAScript 2015 features.

### LINKED ISSUE #3181: Enable GCC for OutputMode.ECMAScript2015
Based on my [question](https://gitter.im/scala-js/scala-js?at=5a07b68f51915d962afbd867) on Gitter and @sjrd's [reply](https://gitter.im/scala-js/scala-js?at=5a07f5d186d308b755da315d):

> The ES2015 output still deactivates Closure, at the moment. For a long time Closure did not support ES2015 so we had deactivated it. **We need to revisit that one of these days.**

So the issue is for "revisiting" this topic 😄  Also if anybody can point me to some relevant documentation to read, please do.

sjrd:
We update `ClosureLinkerBackend` and in particular `ClosureAstTransformer` so that they understand and can translate ECMAScript 2015 features.

Combined with the version of GCC, which understands ES 2015, this allows us to enable GCC when using ES 2015 features.

The Jenkinsfile is updated to test more configurations, which become meaningful due to this change.

### DIFF
--- a/linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureAstTransformer.scala
+++ b/linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureAstTransformer.scala
@@ -46,6 +46,10 @@ private[closure] class ClosureAstTransformer(relativizeBaseURI: Option[URI]) {
         val node = transformName(ident)
         optRhs.foreach(rhs => node.addChildToFront(transformExpr(rhs)))
         new Node(Token.VAR, node)
+      case Let(ident, mutable, optRhs) =>
+        val node = transformName(ident)
+        optRhs.foreach(rhs => node.addChildToFront(transformExpr(rhs)))
+        new Node(if (mutable) Token.LET else Token.CONST, node)
       case Skip() =>
         new Node(Token.EMPTY)
       case Block(stats) =>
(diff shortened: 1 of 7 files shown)


linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureAstTransformer.scala
```diff
@@ -135,12 +139,116 @@ private[closure] class ClosureAstTransformer(relativizeBaseURI: Option[URI]) {
...
 
+      case ClassDef(className, parentClass, members) =>
+        val membersBlock = new Node(Token.CLASS_MEMBERS)
+        for (member <- members)
+          membersBlock.addChildToBack(transformClassMember(member))
+
+        new Node(
+            Token.CLASS,
+            transformOptToEmpty(className)(transformName(_)),
+            transformOptToEmpty(parentClass)(transformExpr(_)),
```
gzm0: Not sure this helper is worth it.

```scala
className.fold(new Node(Token.EMPTY))(transformName(_))
```

seems easier to understand.

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureAstTransformer.scala
```diff
@@ -189,16 +297,10 @@ private[closure] class ClosureAstTransformer(relativizeBaseURI: Option[URI]) {
         val node = new Node(Token.ARRAYLIT)
         items.foreach(i => node.addChildToBack(transformExpr(i)))
         node
-
```
gzm0: Intentional removal?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -138,8 +134,12 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
...
 
     val content = new java.lang.StringBuilder
-    for (topLevelVarDecls <- topLevelVarDeclarations)
-      content.append(topLevelVarDecls + "\n")
+    for (topLevelVarDecls <- topLevelVarDeclarations) {
+      val varDecls =
+        if (!topLevelVarDecls.startsWith("let ")) topLevelVarDecls
+        else "var " + topLevelVarDecls.stripPrefix("let ")
+      content.append(varDecls + "\n")
+    }
```
gzm0: Oh, dear; what is happening here? Do you think we can improve this by passing on a js.Tree instead?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureModuleBuilder.scala
```diff
@@ -32,8 +33,17 @@ private[closure] class ClosureModuleBuilder(
...
+    tree match {
+      case js.Block(stats) =>
+        for (stat <- transformer.transformBlockStats(stats)(NoPosition))
+          treeBuf += stat
+      case js.Skip() =>
+        // ignore
+      case _ =>
+        treeBuf += transformer.transformStat(tree)(NoPosition)
+    }
+  }
```
gzm0: This is odd: Why is this necessary now? In other words: We might have an incomplete specification of `addJSTree` or an abuse of a `js.Block` instead of calling `addJSTree` multiple times.

gzm0: I see; I think this warrants a comment. Also:

```scala
treeBuf ++= transformer.transformBlockStats(stats)(NoPosition)
```

in the `js.Block` case?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
gzm0: Yeah, much better.

sjrd:
Should I squash the two commits or keep them separate? I'm leaning towards squashing, but not sure.

sjrd:
After sleeping on it, I believe they're best kept separate.

PR #3517, opened 2018-12-27
JUnit cleanups

gzm0:


sjrd:
Can't we keep supporting quiet mode for synchronous tests? It seems to me that it should be easy to just say that async tests do not support quiet mode, but sync tests do. Quiet mode is a public-facing features, but async tests will be private; it seems wrong to deny users a public feature because of some private requirements.

gzm0:
> Can't we keep supporting quiet mode for synchronous tests? It seems to me that it should be easy to just say that async tests do not support quiet mode, but sync tests do. Quiet mode is a public-facing features, but async tests will be private; it seems wrong to deny users a public feature because of some private requirements.

We can, if we really want to. However, note that the current implementation is wrong: quiet mode is supposed to only suppress stdout if the test does not fail.

Taking a step back, we should ask ourselves the question what we want to support here: A minimal JUnit implementation providing what we need or as much as we possibly can. I don't think it is a good idea to support things "half-way". So if we accept `-q` as a flag, it should (to the extent possible) behave just like the JVM JUnit runner and not simply "like it did before".

sjrd:
> Taking a step back, we should ask ourselves the question what we want to support here: A minimal JUnit implementation providing what we need or as much as we possibly can. I don't think it is a good idea to support things "half-way". So if we accept -q as a flag, it should (to the extent possible) behave just like the JVM JUnit runner and not simply "like it did before".

Our implementation of JUnit is advertised as a cross-platform solution for testing with JUnit. So limiting it to what *we* need is clearly not acceptable. We should strive for parity with the JVM version within reasonable bounds. There are of course things which cannot really be properly supported, like timeouts.

gzm0:
Updated commit message of quiet mode removal.

PR #3518, opened 2018-12-27
Directly build the core JS lib as a `js.Tree`. 

sjrd:
The core JS lib used to be written in a .js file with a custom preprocessor to deal with variants in semantics, ES features and module kind. In this commit, we directly generate it as a JS tree instead.

Despite the relative loss of readability, this has a few perks:

* We can code-generate a few things that required a lot of duplication before.
* We can leverage a number of helpers in `JSGen`, which centralizes a bit more some of the encodings that we use.
* There is less duplication between similar code paths such as using ES 2015 or not.
* We can convert the AST into GCC's AST, eliminating the need for the awkward combination of text source code and ASTs.
* We can track the global refs used by the core JS lib itself in a more principled way.


sjrd:
Ready for review. Only the last commit belongs to this PR.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+              MethodDef(static = false, Ident("constructor"),
+                  paramList(arg), {
+                Block(
+                    Apply(Super(), Nil),
+                    If(typeof(arg) === str("number"), {
+                      // arg is the length of the array
+                      Block(
+                          privateFieldSet("u", New(ArrayRef, arg :: Nil)),
+                          For(let(i, 0), i < arg, i.++, {
+                            Assign(BracketSelect(This() DOT "u", i), componentZero)
```
gzm0: Use `:=`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+                  MethodDef(static = false, Ident("get"), paramList(i), {
+                    Block(
+                        boundsCheck,
+                        Return(BracketSelect(This() DOT "u", i))
+                    )
+                  }),
+                  MethodDef(static = false, Ident("set"), paramList(i, v), {
+                    Block(
+                        boundsCheck,
+                        Assign(BracketSelect(This() DOT "u", i), v)
```
gzm0: Use `:=`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+                      }, {
+                        If(that === genClassOf(BoxedUnitClass), {
+                          thatFakeInstance := Undefined()
+                        }, {
+                          thatFakeInstance := ObjectConstr(List(classData -> that))
+                        })
+                      })
+                    })
+                  })
+                }),
```
gzm0: Seems like this could be written as a `switch`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+      genLet(ref.ident, mutable = true, rhs)
+
+    private def paramList(refs: VarRef*): List[ParamDef] =
+      refs.toList.map(ref => ParamDef(ref.ident, rest = false))
+
+    private def str(s: String): StringLiteral = StringLiteral(s)
+
+    private def bool(b: Boolean): BooleanLiteral = BooleanLiteral(b)
+
+    private implicit def int(i: Int): IntLiteral = IntLiteral(i)
```
gzm0: Why is this implicit and the others not? What is special about ints?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+              val target = if (static) classRef else prototype
+              genPropSelect(target, name) := Function(arrow = false, args, body)
+          }
+        }
+
+        Block(ctorFun :: inheritProto :: setMembers)
+      }
+    }
+
+    private def varRef(name: String): VarRef = VarRef(Ident(name))
```
gzm0: Have you considered using symbols with implicit conversions for this? It might quite shorten the file (and maybe even aid readability).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+      def genPolyfillFor(builtinName: String): Tree = builtinName match {
+        case "imul" =>
+          val a = varRef("a")
+          val b = varRef("b")
+          val ah = varRef("ah")
+          val al = varRef("al")
+          val bh = varRef("bh")
+          val bl = varRef("bl")
+          Function(arrow = false, paramList(a, b), Block(
+              const(ah, a >>> 16),
```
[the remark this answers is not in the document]
sjrd: Yes that's right. Actually even if the operand is not 32 bits it will be coerced to 32 bits before doing the shift. The `& 0xffff` was completely redundant.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -0,0 +1,1496 @@
...
+      }
+    }
+
+    private def defineIsHijackedBoxedClasses(): Unit = {
+      val v = varRef("v")
+
+      def defineIsIntLike(name: String, specificTest: Tree): Unit = {
+        buf += envFunctionDef(name, paramList(v), {
+          Return((typeof(v) === str("number")) && specificTest &&
+              ((int(1) / v) !== (int(1) / double(-0.0))))
```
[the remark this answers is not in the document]
sjrd: This *is* the emitter. The only thing after that is the printer to text, so nothing is going to postprocess a `Double.NegativeInfinity`.

sjrd:
Updated.

Issue #958, opened 2014-08-19
Add means to exit PhantomJS explicitly in client code


l15k4:
Hi,

PhantomJSEnv execution exits `phantom.exit()` on `page.onLoadFinished` which makes it almost impossible to run/test asynchronous stuff like websockets because there is no way of waiting until some event occurs in singlethreaded environment. It would make more sense if PhantomJSEnv by default exited phantom with the possible override consisting in exiting phantom explicitly by user code: http://phantomjs.org/api/webpage/handler/on-callback.html

It seems that the best solution would be having PhantomJSEnv provide the option of "override" and utest the explicit call via its JSRunner.

Details https://groups.google.com/d/msg/scala-js/AcwyD5r5svI/-gar4ZDEhCMJ

PR #3520, opened 2019-01-02
Remove System.exit / __ScalaJSEnv.exitFunction

gzm0:
This has been added way back in the days (#958) as a poor man's way to
support asynchronous testing.

Since then, testing has substantially matured. Notably a JSEnv is not
required anymore to be able to terminate automatically.

Therefore, there is no need anymore to support exiting in a platform
agnostic manner. Or if there is a need or such a thing, it should be
put in an external library with JS-style dynamic platform detection.

Unfortunately this makes the JSEnv test suite slightly more
complicated again (and basically reverts
6746cf3a20d2497bbc6c520666a47f7ec1cd28dd).

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Runtime.scala
+++ b/javalanglib/src/main/scala/java/lang/Runtime.scala
@@ -15,26 +15,10 @@ package java.lang
 import scala.scalajs.js
 
 class Runtime private {
-  def exit(status: Int): Unit =
-    halt(status)
-
+  //def exit(status: Int): Unit
   //def addShutdownHook(hook: Thread): Unit
   //def removeShutdownHook(hook: Thread): Unit
-
-  def halt(status: Int): Unit = {
-    val envInfo = EnvironmentInfo.envInfo
-
-    envInfo.flatMap(_.exitFunction).fold {
-      // We don't have an exit function. Fail
-      throw new SecurityException("Cannot terminate a JavaScript program. " +
-          "Define a JavaScript function `__ScalaJSEnv.exitFunction` to " +
-          "be called on exit.")
-    } { exitFunction =>
-      exitFunction(status)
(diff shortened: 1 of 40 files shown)


javalanglib/src/main/scala/java/lang/Runtime.scala
```diff
@@ -15,26 +15,10 @@ package java.lang
...
-      throw new SecurityException("Cannot terminate a JavaScript program. " +
-          "Define a JavaScript function `__ScalaJSEnv.exitFunction` to " +
-          "be called on exit.")
-    } { exitFunction =>
-      exitFunction(status)
-      throw new IllegalStateException("__ScalaJSEnv.exitFunction returned")
-    }
-  }
-
+  //def halt(status: Int): Unit =
```
sjrd: Nitpick: for consistency, there shouldn't be an `=` sign here.

js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/ComTests.scala
```diff
@@ -48,10 +48,11 @@ private[test] class ComTests(config: JSEnvSuiteConfig) {
 
   @Test
   def jsExitsOnMessageTest: Unit = {
-    assumeTrue(config.supportsExit)
+    val exitStat = config.exitJSStatement.getOrElse(
+      throw new AssumptionViolatedException("JSEnv needs exitJSStatement"))
```
sjrd: 4 spaces.

sjrd: (or use braces)

gzm0:
Updated. There was indeed a partest linking in (not using) `sys.exit`.

PR #3524, opened 2019-01-07
Get rid of TypeKinds in the compiler.

sjrd:


compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  def toTypeRef(t: Type): Types.TypeRef =
+    convert(t, 0)(ToTypeRefFinisher)
+
+  private sealed abstract class ConversionFinisher {
+    type Result
+
+    def finish(sym: Symbol, arrayDepth: Int): Result =
+      if (arrayDepth == 0) finishNonArray(sym)
+      else finishArray(sym, arrayDepth)
```
gzm0: This seems to be better suited to be at the call site.

compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
```diff
@@ -0,0 +1,173 @@
...
+  import global._
+  import definitions._
+
+  def toIRType(t: Type): Types.Type =
+    convert(t, 0)(ToIRTypeFinisher)
+
+  def toTypeRef(t: Type): Types.TypeRef =
+    convert(t, 0)(ToTypeRefFinisher)
+
+  private sealed abstract class ConversionFinisher {
```
gzm0: Instead of this, consider having `convert` return a `(Symbol, Int)` and do the finishing at the call site. This will make for much simpler code IMO.

compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
```diff
@@ -0,0 +1,173 @@
...
+          BooleanClass -> Definitions.BooleanClass,
+          CharClass    -> Definitions.CharClass,
+          ByteClass    -> Definitions.ByteClass,
+          ShortClass   -> Definitions.ShortClass,
+          IntClass     -> Definitions.IntClass,
+          LongClass    -> Definitions.LongClass,
+          FloatClass   -> Definitions.FloatClass,
+          DoubleClass  -> Definitions.DoubleClass,
+          NothingClass -> encodeClassFullName(RuntimeNothingClass),
+          NullClass    -> encodeClassFullName(RuntimeNullClass)
```
gzm0: Hmm... Do we nowhere have a method that converts IR types to their boxed representation (or is this this place :P).

sjrd: For `Nothing` and `Null`, it is this place (it's Scala-specific; the IR and linker don't know about `scala.runtime.{Null,Nothing}`)

For the others, we're not talking about their boxed representation here. We're talking about the `ClassRef` name of primitive types. For example when `classOf[Int]` becomes `ClassOf(ClassRef("I"))`. And for that map, it's not the only place. It exists in the IR checker to make sure that encoded method names match their signatures. Other than that, I'm not sure. It's not a common operation to do this mapping, because `ClassRef`s are mostly used in `ClassOf`s (which do not interact with the `Type` version of their argument) and `Is/AsInstanceOf`s (for which primitives are not valid; `Unbox` must be used instead).

So basically, I don't see any obvious duplication here.

compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
sjrd: I added a commit with that suggestion. If you prefer that one, I'll squash the commits.

compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
gzm0: OK :) Thanks for the explanation.

compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
gzm0: I very much prefer that one :) Thanks.

PR #3526, opened 2019-01-08
Fix the name of jl_CloneNotSupportedException in inlineable init's blacklist.

sjrd:
The test accompanying this change does not actually fails before the fix, but it does fail in a hello world. The bug cannot be reproduced in our test suite: the existence of JUnit indirectly refers to `java.lang.Enum`, which uses the secondary constructor of `CloneNotSupportedException`, which supersedes the blacklist.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ObjectJSTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ObjectJSTest.scala
@@ -15,6 +15,8 @@ package org.scalajs.testsuite.javalib.lang
 import org.junit.Test
 import org.junit.Assert._
 
+import org.scalajs.testsuite.utils.AssertThrows._
+
 import scala.scalajs.js
 
 // scalastyle:off disallow.space.before.token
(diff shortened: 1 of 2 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ObjectJSTest.scala
```diff
@@ -30,4 +32,13 @@ class ObjectJSTest {
...
     (js.Array(5)  : Any).asInstanceOf[Object]
   }
+
+  @Test def cloneOnNonScalaObject(): Unit = {
+    class CloneOnNonScalaObject extends js.Object {
+      def boom(): Any = this.clone()
+    }
+
+    val obj = new CloneOnNonScalaObject()
+    assertThrows(classOf[CloneNotSupportedException], obj.boom())
```
gzm0: `(new js.Object).clone()` instead? Or did you want to test this for a non-native JS class specifically?

sjrd: It's `protected` ;)

gzm0: Ah...

PR #3529, opened 2019-01-08
RFC: What if Chars were boxed as Strings of length 1?

sjrd:
Obvious destruction of semantics:

* `('A': Any).isInstanceOf[String]` now returns `true`, as does `("A": Any).isInstanceOf[Char]` (similar to `Int`s vs `Double`s)
* `('A': Any).equals("A")` returns `true` (similar to `Int`s vs `Double`s)
* `('A': Any) == ("A": Any)` returns `true` (*not* similar to `Int`s vs `Double`s, in the sense that for those it's also the case on the JVM, but not for `Char`/`String`)

Advantages:

* Optimized string representation of the VM for boxed chars
* Makes sense from an interop point of view: in JS if we manipulate one char, it *is* a string of length 1 (e.g., `"abc".charAt(1)` returns the *string* `"b"`)

What could possibly go wrong? 🙉


gzm0:
> Optimized string representation of the VM for boxed chars

Is this really faster? We'd probably need to benchmark this quite seriously...

> Makes sense from an interop point of view: in JS if we manipulate one char, it is a string of length 1 (e.g., "abc".charAt(1) returns the string "b")

Definitely nicer. However, how often is this needed?

> What could possibly go wrong?

```scala
val a = 'a': Any
val b = 97: Any
a == b // was true, false now
```

Conceptually, I have one concern that `String` is not an `AnyVal` in Scala. Otherwise one could almost argue that these semantics are more consistent with the number semantics (as in: for value types only the values matter).

Not really sure about this one overall....

sjrd:
> Is this really faster? We'd probably need to benchmark this quite seriously...

Definitely consumes less memory. Faster? No idea.

> > Makes sense from an interop point of view: in JS if we manipulate one char, it is a string of length 1 (e.g., "abc".charAt(1) returns the string "b")
>
> Definitely nicer. However, how often is this needed?

Probably very rarely. As an indicator, every now and then I see a question about why `Long`s cannot be used to type JS numbers that are whole but too large for `Int`s. In contrast, I don't remember ever have seen a question why `Char`s cannot be used to type a JS string of length 1.

AFAIK, the only method that really uses that is `String.prototype.charAt`, and that one we already hide away behind `java.lang.String.charAt`, which is actually implemented on top of `String.prototype.charCodeAt` instead.

> ```scala
> val a = 'a': Any
> val b = 97: Any
> a == b // was true, false now
> ```

No, that's still `true`. `BoxesRunTime.equals` makes sure of it. However, that also means that it also answers `true` for `val a = "a": Any`. Maybe eventually I'll manage to convince Martin and the SIP committee that `Char`s shouldn't be equal to numerics in Scala to begin with, but I tried that already.

> Conceptually, I have one concern that `String` is not an `AnyVal` in Scala. Otherwise one could almost argue that these semantics are more consistent with the number semantics (as in: for value types only the values matter).

True. It can also be argued in a different way, though: for types that are semantically exposed as JavaScript primitive types, only the values matter.

> Not really sure about this one overall....

Nah, me neither. I was talking yesterday with some folks about `Char`s and why they're the only primitive that gets boxed in Scala.js, and I figured I'd give that a try and see what breaks, and what's the sentiment about it. I was pretty happy to be able to do this seemingly very big change with only a few dozens LoC :p

Issue #2629, opened 2016-10-12
Rethink encoding of field names in the IR


sjrd:
See https://github.com/scala-js/scala-js/pull/2628#discussion_r83030921

PR #3531, opened 2019-01-09
DO NOT MERGE Fix #2108: Defer identifier name mangling to the emitter.

### LINKED ISSUE #2108: Defer identifier name mangling until desugaring

sjrd:
Previously, we banished JavaScript keywords in the IR, in several places:

* The name of local variables
* The unprefixed name of fields
* The name of methods (not including their encoded signature)

This commit allows all JavaScript keywords (and special reserved or meaningful identifiers such as `arguments`) in all these places. For fields and methods, it was irrelevant anyway, since they are always suffixed with something, and therefore the keyword is never actually a keyword.

For local variables, we defer their mangling until the emitter, in the pessimistic naming run (so we assume in the first run that no local variable uses a reserved word). The optimizer helps by eagerly renaming such variables on its own, to avoid the pessimistic run. It has to do renaming for inlined stuff anyway, so this is not more job for it. The pessimistic run is still necessary in exported methods (which are not optimized) and if the optimizer is disabled.

There is one place where the set of strict keywords must be avoided in the IR, per semantics: in `JSGlobalRef`. Therefore, `JSGlobalRef` does maintain a fixed set of illegal global variable names, which cannot be accessed. They trigger compile errors.

There is still one thing which must be eagerly mangled: identifiers starting with a digit (for fields, methods, and local variables). That's because scalac thinks they are valid, but they cannot be used at the start of an `IdentifierName` in JavaScript. We still do not allow them in the IR, because the IR can reasonably follow one of the two following rules:

* Either it follows JavaScript's rules, hence `IdentifierName`
* Or it follows the JVM rules, which allow virtually all Unicode characters (with a few exceptions such as `. / < >`).

Supporting the latter in the IR would mean much more preemptive mangling in the emitter, which would slow it down for almost no reason, since scalac uses none of those weird characters.

For that reason, the IR sticks to JavaScript `IdentifierName`s, and it is the scalac back-end's responsibility to encode leading digits away. We choose to prefix a leading digit with the character `〇` (Ideographic Number Zero, U+3007, category `LETTER_NUMBER`), which is a valid first character for an `IdentifierName`. This avoids cascading manglings. At the end of the day, that character will be Unicode-escaped as \u3007 by the pretty-printer to keep the character set of the emitted .js as ASCII (this was already implemented). If we had cascading manglings, it would be awkward to specify interoperability between different languages compiling to the IR.


linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/Trees.scala
```diff
@@ -45,10 +45,40 @@ object Trees {
...
+     *  - `undefined`, because it would be really confusing.
+     *
+     *  It can be expanded in future versions of the linker.
+     */
+    val ReservedLocalNames: Set[String] = {
+      /* JSGlobalRef.ReservedWords already contains all keywords, per spec.
+       * Let's reuse them.
+       */
+      ir.Trees.JSGlobalRef.ReservedWords ++ Set(
+          "arguments", "eval", "undefined"
```
gzm0: Arguments is in both, is this intentional?

sjrd:
Uh. In fact Scala will let through anything that is a `Character.isJavaIdentifierPart()` (even for the first character, although that really should be `isJavaIdentifierStart()`. So in fact there *are* other characters that will blow it, for example non-dollar currency symbols such as `€`.

sjrd:
I pushed a commit with a different strategy, accompanied with some more tests this time.

If it looks good, I'll squash the commits.

gzm0:
Hmmm... But now that we basically need the full JVM set from scalac's side, wouldn't it make more sense to allow all these in the IR? It seems to me that we have intentionally modeled the IR very close to the JVM bytecode in order to make it easy for languages targeting the JVM to target JS. So we should do the same for identifiers. Or is my understanding wrong here?

sjrd:
> I'll come up with yet another version supporting all of the JVM, for comparison.

Well, to do that in a decent way, I'll have to apply significant changes to field and method handling. That's good because it'll force me to deal with #2629. However to address methods I'm basically blocked by private methods (#3210/#3240).

sjrd:
Superseded by #3814 and #3823.

PR #3535, opened 2019-01-12
Erase `scala.runtime.BoxedUnit` to `java.lang.Void`.

sjrd:
This removes one more dependency of the IR specification on the Scala namespace, in the handling of primitives, which is pretty core.

`java.lang.Void` makes sense as the boxed class of `Unit` because

* `scala.Unit` is to `void` as `scala.Int` is to `int`
* Other boxed classes have a `TYPE` static field that represent the `classOf` of their corresponding primitive, and `java.lang.Void` precisely has a `TYPE` field that represents `classOf[Unit]`.

`java.lang.Void` is safe enough to reuse as it cannot have any instance in the JVM spec (barring evil uses of reflection). Now we're giving it exactly one instance: `undefined`.

Since `java.lang.Void` does not extend `java.io.Serializable`, this introduces a weird discrepancy wrt. Scala/JVM: a boxed `()` will not answer true to `.isInstanceOf[Serializable]`. However, it does not seem like something would ever do. An alternative would be to alter `java.lang.Void` and make it extend `java.io.Serializable` anyway. This is a binary compatible extension so we could do it later if necessary.

Another semantic difference introduced by this change is that `((): Any).getClass().getName()` will now return `java.lang.Void` instead of `scala.runtime.BoxedUnit`.


sjrd:
Comment from smarter at the Scala meeting today: by doing this we're potentially exposing ourselves to issues if a future version of the JDK decides to allow instances of `java.lang.Void`.

IMO That shouldn't be a problem because it's very unlikely, and if it happens, then I'm fairly sure it would be to have one singleton instance of `java.lang.Void` which would be a boxed version of `void`, precisely what our only instance is supposed to be anyway.

tarsa:
Will this change affect the output of e.g.: `println(Seq((), (), ()))`?

sjrd:
No, it won't. The change in this PR does not change the string representation of `()`. It only changes the string representation of `((): Any).getClass`.

PR #3540, opened 2019-01-17
Added less granular tags to the fastOptJs and fullOptJS tasks

### LINKED ISSUE #3539: Allow users to limit how many Scala.js linker tasks can run at once in sbt
SBT will happily run lots of fastOptJS or fullOptJS tasks at once. These tasks use a lot of memory so it would be useful for users to be able to limit how many of these tasks can run concurrently to work around resource limits in continuous integration.

DavidGregory084:
Fixes #3539.

This would allow users to restrict how many fastOptJS / fullOptJS tasks are running at once in resource-constrained environments.

Some questions:
 
* Is it okay to add new methods to the `Stage` trait? I imagine that is a problem for bincompat. If not are you happy for me to duplicate some pattern matches on the `stage`?
* I would like to add a `Tags` object to the user-facing `ScalaJSPlugin `which contains these tags, is that okay?
* I'd like to use this with 0.6.x, would you accept a PR for that too?

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -158,11 +158,7 @@ private[sbtplugin] object ScalaJSPluginInternal {
       usesScalaJSLinkerTag in key := {
         val projectPart = thisProject.value.id
         val configPart = configuration.value.name
-
-        val stagePart = stage match {
-          case Stage.FastOpt => "fastopt"
-          case Stage.FullOpt => "fullopt"
-        }
+        val stagePart = stage.tagName
 
         Tags.Tag(s"uses-scalajs-linker-$projectPart-$configPart-$stagePart")
       },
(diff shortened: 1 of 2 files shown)


sjrd:
Thank you for your PR.

Please read our [contributing guide](https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md), in particular the part about commit messages.

> I'd like to use this with 0.6.x, would you accept a PR for that too?

We take PRs *either* for 0.6.x or for master, but we do merge 0.6.x into master every now and then. If you want your PR to apply to both, you should *only* send it for 0.6.x.

> Is it okay to add new methods to the Stage trait?

No, please avoid that. This is a completely empty enum. You can duplicate matches all you like.

> I would like to add a Tags object to the user-facing ScalaJSPlugin which contains these tags, is that okay?

It *should* be in the user-facing `ScalaJSPlugin`, absolutely. Otherwise this is not very usable.

---

I would question the usefulness of having separate tags for fastOptJS and fullOptJS. Supposedly, if someone does not want two `fastOptJS` to run concurrently, they also don't want one `fastOptJS` running concurrently with a `fullOptJS`. I suggest having only one tag called `scalajs-link`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -185,20 +181,19 @@ private[sbtplugin] object ScalaJSPluginInternal {
         val output = (artifactPath in key).value
         val linker = (scalaJSLinker in key).value
         val usesLinkerTag = (usesScalaJSLinkerTag in key).value
+        val scalaJsOptTag = Tags.Tag(s"scalajs-${stage.tagName}")
```
sjrd: We write acronyms in all-uppercase when they are not first "word" of a camelCase identifier. So this would be `scalaJSOptTag`, for example.

DavidGregory084:
Thanks for the detailed review @sjrd! I will close this and reopen against 0.6.x.

PR #3541, opened 2019-01-17
Fix #3539: Enable users to limit concurrent linking tasks in sbt

### LINKED ISSUE #3539: Allow users to limit how many Scala.js linker tasks can run at once in sbt

DavidGregory084:
This is a continuation of #3540. It will enable users to limit how many linking tasks are able to run at once in sbt by using the `concurrentRestrictions` key:

```scala
Global / concurrentRestrictions += Tags.limit(ScalaJSTags.Link, 2)
```

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -58,6 +58,21 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalaJSIsSnapshotVersion = ScalaJSVersions.currentIsSnapshot
     val scalaJSBinaryVersion = ScalaJSCrossVersion.currentBinaryVersion
 
+    /** Declares [[sbt.Tags.Tag Tag]]s which may be used to limit the
+     *  concurrency of build tasks.
+     *
+     *  For example, the following snippet can be used to limit the
+     *  number of linking tasks which are able to run at once:
+     *
+     *  {{{
+     *  Global / concurrentRestrictions += Tags.limit(ScalaJSTags.Link, 2)
+     *  }}}
+     */
+    object ScalaJSTags {
+      /** This tag is applied to the [[fastOptJS]] and [[fullOptJS]] tasks. */
+      val Link = Tags.Tag("scalajs-link")
+    }
+
     // Stage values
     @deprecated("Use FastOptStage instead", "0.6.6")
     val PreLinkStage = Stage.FastOpt
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -333,7 +333,7 @@ object ScalaJSPluginInternal {
 
           val sourceMapFile = FileVirtualJSFile(output).sourceMapFile
           Attributed.blank(output).put(scalaJSSourceMap, sourceMapFile)
-        }.tag(usesLinkerTag)
+        }.tag(usesLinkerTag, ScalaJSTags.Link)
       }.value,
 
       key := key.dependsOn(packageJSDependencies, packageScalaJSLauncherInternal).value,


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -333,7 +334,12 @@ object ScalaJSPluginInternal {
 
           val sourceMapFile = FileVirtualJSFile(output).sourceMapFile
           Attributed.blank(output).put(scalaJSSourceMap, sourceMapFile)
-        }.tag(usesLinkerTag)
+        }
+
+        linkerTask
+          .tag(usesLinkerTag)
+          .tag(scalaJSLinkTag)
```
sjrd: I believe `.tag()` takes varargs, both in sbt 0.13.x and 1.x. If that's true, I suggest keeping the old formatting and just calling
```scala
}.tag(usesLinkerTag, scalaJSLinkTag)
```

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -333,7 +334,12 @@ object ScalaJSPluginInternal {
...
+
```
sjrd: Spurious blank line.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -304,8 +304,9 @@ object ScalaJSPluginInternal {
         val output = (artifactPath in key).value
         val linker = (scalaJSLinker in key).value
         val usesLinkerTag = (usesScalaJSLinkerTag in key).value
+        val scalaJSLinkTag = ScalaJSTags.Link
```
sjrd: I'm not sure this `val` is necessary. I suggest directly using `ScalaJSTags.Link` below. It also helps keeping that series of `val`s dedicated to all the `.value`s.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -58,6 +58,8 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalaJSIsSnapshotVersion = ScalaJSVersions.currentIsSnapshot
     val scalaJSBinaryVersion = ScalaJSCrossVersion.currentBinaryVersion
 
+    val ScalaJSTags = org.scalajs.sbtplugin.ScalaJSTags
```
sjrd: Maybe this could be an `object` inline, rather than duplicating its contents in a separate object? @gzm0 Opinion?

DavidGregory084: Yeah, the only reason I used an object is because it was used in the task definition and I though it would be a bit ugly to use something from the ScalaJSPlugin in the ScalaJSPluginInternal 😃

sjrd: > it would be a bit ugly to use something from the ScalaJSPlugin in the ScalaJSPluginInternal

We do that all the time. Those two objects should be viewed as one, with the non-Internal one containing the public API, and Internal one everything else.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -58,6 +58,10 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalaJSIsSnapshotVersion = ScalaJSVersions.currentIsSnapshot
     val scalaJSBinaryVersion = ScalaJSCrossVersion.currentBinaryVersion
 
+    object ScalaJSTags {
+      val Link = Tags.Tag("scalajs-link")
```
gzm0: I'd like to see some documentation on this. Basically a description that it is a tag on every task that links to allow restricting concurrency.

DavidGregory084: I have added some docs now, please let me know if these look okay

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -58,6 +58,25 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalaJSIsSnapshotVersion = ScalaJSVersions.currentIsSnapshot
     val scalaJSBinaryVersion = ScalaJSCrossVersion.currentBinaryVersion
 
+
+    /**
+      * Declares [[sbt.Tags.Tag Tag]]s which may be used to limit the
```
sjrd: Please follow the coding style for Scaladoc that we use. It's documented at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#scaladoc

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -58,6 +58,25 @@ object ScalaJSPlugin extends AutoPlugin {
     val scalaJSIsSnapshotVersion = ScalaJSVersions.currentIsSnapshot
     val scalaJSBinaryVersion = ScalaJSCrossVersion.currentBinaryVersion
 
+
```
sjrd: Double blank line.

PR #3542, opened 2019-01-17
Add Object.assign method

cquiroz:
Add signature for the `js.Object.assign` method.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Object.scala
+++ b/library/src/main/scala/scala/scalajs/js/Object.scala
@@ -57,6 +57,37 @@ object Object extends Object {
   def apply(): Object = native
   def apply(value: Any): Object = native
 
+  /** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+   *  The `Object.assign()` method is used to copy the values of all enumerable
+   *  own properties from one or more source objects to a target object. It will
+   *  return the target object.
+   *
+   *  Properties in the target object will be overwritten by properties in the
+   *  sources if they have the same key. Later sources' properties will
+   *  similarly overwrite earlier ones.
+   *
+   *  The `Object.assign()` method only copies enumerable and own properties from
+   *  a source object to a target object. It uses `Get` on the source and `Set`
+   *  on the target, so it will invoke getters and setters. Therefore it
+   *  assigns properties versus just copying or defining new properties. This may
+   *  make it unsuitable for merging new properties into a prototype if the merge
+   *  sources contain getters. For copying property definitions, including their
+   *  enumerability, into prototypes `Object.getOwnPropertyDescriptor()` and
+   *  `Object.defineProperty()` should be used instead.
+   *
+   *  Both `String` and `Symbol` properties are copied.
+   *
+   *  In case of an error, for example if a property is non-writable, a
+   *  `TypeError` will be raised, and the target object can be changed if any
+   *  properties are added before error is raised.
+   *
+   *  Note that `Object.assign()` does not throw on null or undefined source
+   *  values.
+   *
+   *  MDN
+   */
+  def assign(t: Object, s: Object*): Object = native
+
   /** Tests whether the object has a property on itself or in its prototype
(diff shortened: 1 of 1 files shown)


cquiroz:
It wasn't very clear if we are doing tests on `js.Object` nor where they are. Please let me know if I should add a test

library/src/main/scala/scala/scalajs/js/Object.scala
```diff
@@ -57,6 +57,37 @@ object Object extends Object {
...
+   *
+   * In case of an error, for example if a property is non-writable, a TypeError
+   * will be raised, and the target object can be changed if any properties are
+   * added before error is raised.
+   *
+   * Note that Object.assign() does not throw on null or undefined source values.
+   *
+   * MDN
+   */
+  def assign(t: Object, s: Object: *): Object = native
```
gzm0: This line has a syntax error (should be `s: Object*`).
Also, this needs the es6 badge. See for example the typedarray package for how to add it.

cquiroz:
I updated the code and added the badge at the top of the scaladoc

library/src/main/scala/scala/scalajs/js/Object.scala
```diff
@@ -57,6 +57,38 @@ object Object extends Object {
   def apply(): Object = native
   def apply(value: Any): Object = native
 
+  /** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+   *
+   * The Object.assign() method is used to copy the values of all enumerable own
```
sjrd: Could you make sure to follow our coding styles for Scaladoc, explained at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#scaladoc?

You'll also need to remove the blank line above this line for the initial sentence to correctly appear in the rendered Scaladoc.

library/src/main/scala/scala/scalajs/js/Object.scala
```diff
@@ -57,6 +57,38 @@ object Object extends Object {
...
+   * properties from one or more source objects to a target object. It will
+   * return the target object.
+   *
+   *
```
sjrd: Double blank line.

library/src/main/scala/scala/scalajs/js/Object.scala
```diff
@@ -57,6 +57,38 @@ object Object extends Object {
...
+   * Properties in the target object will be overwritten by properties in the
+   * sources if they have the same key.  Later sources' properties will similarly
+   * overwrite earlier ones.
+   *
+   * The Object.assign() method only copies enumerable and own properties from a
+   * source object to a target object. It uses [[Get]] on the source and [[Set]]
```
sjrd: Those are probably going to cause errors when generating Scaladoc. Surround them with backticks if necessary. Also in general, it's recommended to surround code with backticks so that it renders appropriately.

cquiroz:
I updated the scaladoc as requested. Let me know if it is fine and I'll squash the commits

library/src/main/scala/scala/scalajs/js/Object.scala
```diff
@@ -57,6 +57,37 @@ object Object extends Object {
   def apply(): Object = native
   def apply(value: Any): Object = native
 
+  /** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
+   * The `Object.assign()` method is used to copy the values of all enumerable
```
sjrd: Ah, almost 😄 There should be one more space between the `*` and the text. That way, the first character of the text is always aligned, from the first line to the last one. So it starts as
```scala
  /** <span class="badge badge-ecma6" style="float: right;">ECMAScript 6</span>
   *  The `Object.assign()` method is used to copy the values of all enumerable
   *  own properties from one or more source objects to a target object. It will
   *  return the target object.
```
Other than that, good to squash and then we can merge it.

cquiroz: Gotcha, fixed and squashed

PR #3543, opened 2019-01-18
 Do not require `ju.Properties` if we only use `System.{get,set,clear}Property`.

sjrd:
It is reasonably common to use `System.getProperty` and other related methods, but not to use `System.{get,set}Properties` nor to directly instantiate `ju.Properties`.

This commit uses a few laziness tricks to avoid making `ju.Properties` reachable (along with `ju.Hashtable` and some others) if the codebase only uses `System.{get,set,clear}Property`.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/System.scala
+++ b/javalanglib/src/main/scala/java/lang/System.scala
@@ -243,7 +243,7 @@ object System {
       sysProp.setProperty("java.vm.specification.vendor", "Oracle Corporation")
       sysProp.setProperty("java.vm.specification.name", "Java Virtual Machine Specification")
       sysProp.setProperty("java.vm.name", "Scala.js")
-      linkingInfo.linkerVersion.foreach(v => sysProp.setProperty("java.vm.version", v))
+      sysProp.setProperty("java.vm.version", linkingInfo.linkerVersion)
       sysProp.setProperty("java.specification.version", "1.8")
       sysProp.setProperty("java.specification.vendor", "Oracle Corporation")
       sysProp.setProperty("java.specification.name", "Java Platform API Specification")
--- a/javalanglib/src/main/scala/java/lang/System.scala
+++ b/javalanglib/src/main/scala/java/lang/System.scala
@@ -234,48 +234,73 @@ object System {
   }
 
   private object SystemProperties {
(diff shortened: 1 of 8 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
```diff
@@ -201,4 +201,79 @@ class SystemTest {
...
+
+  // System properties
+
+  private final val LineSeparatorPropName = "line.separator"
+  private final val ExistingPropName = "org.scalajs.testsuite.existingprop"
+  private final val TestPropName = "org.scalajs.testsuite.testprop"
+
+  @After def resetSystemPropertiesAfterEachTest(): Unit = {
+    System.setProperties(null)
+  }
```
gzm0: Consider putting the system property tests into a separate class so that we don't do this for every test.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/SystemTest.scala
```diff
@@ -201,4 +201,79 @@ class SystemTest {
...
+
+  @Test def getSetClearProperty(): Unit = {
```
gzm0: I find the two names here are a bit misleading. From what I understand is the first test tests the `dict` only case, the second test tests when forcing happens. Maybe add this in a comment to clarify?

sjrd:
Updated to address the two small comments.

> Do you think there is any way we can test the fact that `ju.Properties` is not linked in anymore by default?

There is a way, yes: we could provide the linker unit tests with the path to the full `scalajs-library.jar` (currently they only use the minilib), and have `AnalyzerTest`s that use that full library along with a small method that calls `getProperty`, `setProperty` and `clearProperty`, then assert that the `Analysis` has `!isAnySubclassInstantiated` for `ju.Properties`.

Should we do that?

sjrd:
Well I figured it wasn't too much code, so I added a commit with that analyzer test.

gzm0:
Up to you if you wanna squash or not.

sjrd:
I actually prefer them kept separate.

PR #3544, opened 2019-01-18
Make division by 0 compliant (always).

sjrd:
So far, division by 0 has been an unchecked undefined behavior. Contrary to the other undefined behaviors we have, this one has always been problematic, because Scala as a language and a type system do nothing to guard against it.

This commit makes division by 0 compliant, i.e., throwing an `ArithmeticException`, in all cases. Some optimizations are thrown in to avoid the additional checks when the rhs is a non-zero constant.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
@@ -1300,8 +1300,37 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
       finishTransformBindings(bindingsAndStats, finishTransformStat(result))
     case PreTransUnaryOp(_, lhs) =>
       finishTransformStat(lhs)
-    case PreTransBinaryOp(_, lhs, rhs) =>
-      Block(finishTransformStat(lhs), finishTransformStat(rhs))(stat.pos)
+
+    case PreTransBinaryOp(op, lhs, rhs) =>
+      // Here we need to preserve the side-effects of integer division/modulo
+      import BinaryOp._
+
+      val newLhs = finishTransformStat(lhs)
+
+      def finishNoSideEffects: Tree =
+        Block(newLhs, finishTransformStat(rhs))(stat.pos)
+
+      op match {
+        case Int_/ | Int_% =>
+          rhs match {
+            case PreTransLit(IntLiteral(r)) if r != 0 =>
+              finishNoSideEffects
+            case _ =>
+              Block(newLhs, BinaryOp(op, IntLiteral(0)(stat.pos),
+                  finishTransformExpr(rhs))(stat.pos))(stat.pos)
+          }
+        case Long_/ | Long_% =>
+          rhs match {
+            case PreTransLit(LongLiteral(r)) if r != 0L =>
+              finishNoSideEffects
+            case _ =>
+              Block(newLhs, BinaryOp(op, LongLiteral(0L)(stat.pos),
+                  finishTransformExpr(rhs))(stat.pos))(stat.pos)
+          }
+        case _ =>
+          finishNoSideEffects
+      }
+
     case PreTransLocalDef(_) =>
       Skip()(stat.pos)
     case PreTransRecordTree(tree, _, _) =>
@@ -2887,18 +2916,8 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
       case Int_+ => IntLiteral(int(lhs) + int(rhs))
       case Int_- => IntLiteral(int(lhs) - int(rhs))
       case Int_* => IntLiteral(int(lhs) * int(rhs))
-
-      case Int_/ =>
-        int(rhs) match {
-          case 0 => IntLiteral(0) // Undefined Behavior
-          case r => IntLiteral(int(lhs) / r)
-        }
-
-      case Int_% =>
-        int(rhs) match {
-          case 0 => IntLiteral(0) // Undefined Behavior
-          case r => IntLiteral(int(lhs) % r)
-        }
+      case Int_/ => IntLiteral(int(lhs) / int(rhs))
+      case Int_% => IntLiteral(int(lhs) % int(rhs))
 
       case Int_|   => IntLiteral(int(lhs) | int(rhs))
       case Int_&   => IntLiteral(int(lhs) & int(rhs))
@@ -2917,18 +2936,8 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
       case Long_+ => LongLiteral(long(lhs) + long(rhs))
       case Long_- => LongLiteral(long(lhs) - long(rhs))
       case Long_* => LongLiteral(long(lhs) * long(rhs))
-
-      case Long_/ =>
-        long(rhs) match {
-          case 0 => LongLiteral(0L) // Undefined Behavior
-          case r => LongLiteral(long(lhs) / r)
-        }
-
-      case Long_% =>
-        long(rhs) match {
-          case 0 => LongLiteral(0L) // Undefined Behavior
-          case r => LongLiteral(long(lhs) % r)
-        }
+      case Long_/ => LongLiteral(long(lhs) / long(rhs))
+      case Long_% => LongLiteral(long(lhs) % long(rhs))
 
       case Long_|   => LongLiteral(long(lhs) | long(rhs))
       case Long_&   => LongLiteral(long(lhs) & long(rhs))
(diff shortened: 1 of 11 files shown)


gzm0:
It feels like I wouldn't be doing my job as a reviewer if I weren't to ask "What about performance?". Also, any reason we don't make this a checked behavior?

sjrd:
Performance is only affected when performing an integer division by a non-constant. Integer division with non-constant values is slow to begin with, down to the processor level, with latencies dozens of times as long as other arithmetic operations. I doubt that an additional test makes it any slower in practice.

> Also, any reason we don't make this a checked behavior?

As I said in the commit message, it is fundamentally different in the sense that Scala does not protect against divisions by 0. By opposition:

* `ClassCastException`s are avoided by a strong type system, which confines explicit casts to rare performance-sensitive parts of a codebase.
* `IndexOutOfBoundsException`s are avoided by all the combinators available on arrays and strings, which make explicit iteration rarely necessary.

There is nothing like that for divisions by 0.

sjrd:
That said, we certainly could make it a `CheckedBehavior` instead, but I'm afraid that bugs due to divisions by 0 won't be caught at testing time, unlike other kinds of `CheckedBehavior`s, due to the lack of support by the language. So the dev vs prod distinction is not helpful for those kinds of errors.

gzm0:
Do we have any existing benchmarks that heavily use (non-constant) division? I'd be much more comfortable with this if we had a little data.

> As I said in the commit message, it is fundamentally different in the sense that Scala does not protect against divisions by 0. By opposition:

- `ClassCastExceptions`: Absolutely agree.
- `IndexOutOfBoundsExceptions`: Not sure I agree here. IMO iteration combinators answer a different use-case than indexed access. And indexed access is equally unprotected by the type system. (OTOH it is probably true that indexed access happens less often in a dynamic setting which would not be caught by tests).

Another POV w.r.t. making this a checked exception is whether we want to "allow" catching an `ArithmeticException`. Not sure this is something we should be opinionated about though.

sjrd:
> Do we have any existing benchmarks that heavily use (non-constant) division?

No, we don't. I can dig up the benchmarks I had done for `Long`s, and re-adapt them for `Int`s to get micro-benchmarks of `Int` arithmetic operations.

sjrd:
Before:
```
IntNop30: Node.js;15.967002764543778;0.0034772965450421524
IntXor30: Node.js;20.731441968602343;0.00165648137079802
IntMul30: Node.js;39.244489056232446;0.004249763483063499
IntDiv30_30: Node.js;52.17132073080477;0.006019416801741286
IntDiv30_8: Node.js;54.42764422269611;0.0068740287505001385

IntNop: Node.js;15.97970926334304;0.0014907559579369145
IntXor: Node.js;26.327586056699044;0.015230639751758562
IntMul: Node.js;39.79079199238288;0.010138473968228187
IntDiv32_32: Node.js;53.76725127351898;0.006553566427106843
IntDiv32_8: Node.js;51.82589188025982;0.007484810592308043
```
After:
```
IntNop30: Node.js;15.930427583840775;0.0033313302230299825
IntXor30: Node.js;20.790196298804954;0.0019994371657090972
IntMul30: Node.js;39.65034554569834;0.0074793968097032065
IntDiv30_30: Node.js;58.13404731259431;0.008993087991756238
IntDiv30_8: Node.js;55.84709138463164;0.0092286348808631

IntNop: Node.js;16.065554712730044;0.0035086964257769644
IntXor: Node.js;26.643512520997668;0.005231426795054181
IntMul: Node.js;39.37519856897995;0.008958719902576776
IntDiv32_32: Node.js;55.362975633100724;0.008641793529265954
IntDiv32_8: Node.js;55.644471553663955;0.0071193516063680655
```
`IntDiv30_30` is the "worst" regression, going from 52 to 58. Subtracting `Nop` it's going from 36 to 42, so a  16% slowdown.

gzm0:
Hmmm... 16% can be quite much. OTOH I do buy your argument that this is less likely to be caught in tests that out of bounds etc. I see the following options:

a) Introduce this as a checked behavior with usual defaults (UBE in fast opt, UB in full opt).
  PROS: Faster, consistent with other things
  CONS: Testing might not catch problems

b) Introduce this as a checked behavior with compliant as default.
  PROS: Safer
  CONS: less consistent with other checked behaviors, behavior might be a bit more likely relied on (not sure I buy this one).

c) Introduce this as is, committing to doing b) if the slowdown is prohibitive for users.
  PROS: Less options, equal safety
  CONS: Implicit commitment to b), needs user escalation if there are problems.


Really feels like the options are a bit on par here. Opinions?

sjrd:
Another way to put the 16% percent into perspective: for `ClassCastException`, we had benchmarks with 100% slowdown *on the entire macro benchmark* (not on a micro-benchmark testing only the cast itself). That was at least 2 orders of magnitude more catastrophic.

gzm0:
OK. Fair enough. As you are typically our performance guardian, I'll stick with your opinion here. I assume this means we'll go ahead without a checked behavior for this.

I'll review in detail.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2887,18 +2930,8 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
-          case r => IntLiteral(int(lhs) / r)
-        }
-
-      case Int_% =>
-        int(rhs) match {
-          case 0 => IntLiteral(0) // Undefined Behavior
-          case r => IntLiteral(int(lhs) % r)
-        }
+      case Int_/ => IntLiteral(int(lhs) / checkForIntDiv(int(rhs)))
+      case Int_% => IntLiteral(int(lhs) % checkForIntDiv(int(rhs)))
```
gzm0: Why do we need `checkForIntDiv`? Isn't this the very commit that make sure dividing by zero would throw?

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala
```diff
@@ -288,6 +289,46 @@ class IntTest {
...
+      assertThrows(classOf[ArithmeticException], divNoInline(x, 0))
+      assertThrows(classOf[ArithmeticException], divInline(x, 0))
+    }
+
+    test(0)
+    test(1)
+    test(43)
+    test(-3)
+
+    // Eligible for constant-folded by scalac itself
```
gzm0: constant-folding

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/IntTest.scala
```diff
@@ -288,6 +289,46 @@ class IntTest {
...
+      assertThrows(classOf[ArithmeticException], modNoInline(x, 0))
+      assertThrows(classOf[ArithmeticException], modInline(x, 0))
+    }
+
+    test(0)
+    test(1)
+    test(43)
+    test(-3)
+
+    // Eligible for constant-folded by scalac itself
```
gzm0: constant-folding

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes, and that's why a division by 0 cannot be *constant-folded* anymore. It needs to preserved as a non-constant, which is why `constantFoldBinaryOp_except_String_+` cannot be called on a division by 0. Instead, that case is caught by `foldBinaryOp`.

sjrd:
Updated for the two typos in comments.

sjrd:
> Note that this might "free" up some partests and Scalac unit tests. Feel free investigate this in another PR though.

I just checked and it doesn't change the result of partests nor scalaTestSuite.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
gzm0: I understand that. But checkForIntDiv is simply an assertion so that if something tries to constant fold div by 0 it crashes (calling code guards against this). But there is no need for this, since the division itself will crash.

sjrd: True. We could simply let the division crash on its own. I don't have any preference. Do you prefer letting the division crash?

gzm0: Yeah, mostly because it is less code.

sjrd: Makes sense. Updated.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2876,6 +2905,12 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
       case BooleanLiteral(value) => value
     }
 
+    def illegalArgDivByZero(): Nothing = {
+      throw new IllegalArgumentException(
+          "constFoldBinaryOp_except_String_+ must not be called for a " +
+          "division by zero")
+    }
+
```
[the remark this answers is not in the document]
sjrd: Right ... updated.

PR #3550, opened 2019-01-26
CI ONLY Make Linker interface async

gzm0:


gzm0:
Pre-review by @sjrd. I'm not sure we should merge this as such, but rather with the first changes actually using it. WDYT?

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardLinkerImpl.scala
```diff
@@ -35,33 +35,26 @@ private final class StandardLinkerImpl private (
...
-      if (!_valid) {
-        throw new IllegalStateException(
-          "Linker is invalid due to a previous exception in a component")
-      }
+    checkValid()
+      .flatMap(_ => frontend.link(
+          irFiles, moduleInitializers, backend.symbolRequirements, logger))
+      .flatMap(linkingUnit => backend.emit(linkingUnit, output, logger))
+      .andThen { case t if t.isFailure => _valid = false }
+      .andThen { case t => _linking.set(false) }
```
sjrd: Consider merging those two `.andThen`s.

sbt-plugin-test/build.sbt
```diff
@@ -66,6 +66,8 @@ lazy val noDOM = project.settings(baseSettings: _*).
         val moduleInitializers = scalaJSModuleInitializers.value
 
         Def.task {
+          import scala.concurrent.ExecutionContext.Implicits.global
```
sjrd: The call to `linker.link()` a few lines below (in this `Def.task {}`) should be surrounded with `Await.result()`, otherwise this test is moot.

test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala
```diff
@@ -16,7 +19,7 @@ import org.scalajs.linker.irio._
 object QuickLinker {
   /** Link the Scala.js test suite on Node.js */
   @JSExport
-  def linkTestSuiteNode(irFilesAndJars: js.Array[String], outputPath: String): Unit = {
+  def linkTestSuiteNode(irFilesAndJars: js.Array[String], outputPath: String): js.Promise[Unit] = {
```
sjrd: Why is this change not accompanied by a change in `Build.scala` somewhere? It looks suspicious to me. Aren't we discarding that `Promise` without looking at its result (nor even waiting for it)?

gzm0: Done.

sbt-plugin-test/build.sbt
gzm0: Done.

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardLinkerImpl.scala
gzm0: I did, found this cleaner but I don't care too much. LMK what to do.

gzm0:
Superseded by #3558.

PR #3557, opened 2019-02-03
Fix #3088: Link all classes that are mentioned in signatures and types

### LINKED ISSUE #3088: Link all classes that are mentioned in signatures and types

gzm0:


gzm0:
A couple of questions I have w.r.t. this:

- What about field types?
- Do we need to benchmark this (or do we not have a choice anyways)?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -389,8 +407,12 @@ object Infos {
 
       methodDef.name match {
         case ComputedName(tree, _) =>
-          traverse(tree)
```
sjrd: The deletion of this `traverse(tree)` is suspicious. Was it done by mistake?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -423,6 +423,7 @@ private final class Analyzer(config: CommonPhaseConfig,
     var areInstanceTestsUsed: Boolean = false
     var isDataAccessed: Boolean = false
     var isAnyStaticFieldReachable: Boolean = false
+    var isReferenced: Boolean = false
```
sjrd: `isReferenced` is never actually used. I believe that it can be removed, and simply doing the `lookupClass()` is enough.

That also means that the calls to `reference()` are superfluous.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -750,37 +751,18 @@ private final class IRChecker(unit: LinkingUnit,
...
+            val kind = c.kind
             if (!kind.isClass) {
               reportError(s"Cannot select $item of non-class $cls")
             } else {
-              /* Actually checking the field is done only if
-               *
-               * * the class exists, and
-               * * it has instances (including instances of subclasses).
-               *
-               * The latter test is necessary because the BaseLinker can
```
sjrd: This part of the comment should be preserved.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1211,20 +1193,6 @@ private final class IRChecker(unit: LinkingUnit,
...
-  private def lookupInfo(className: String)(
-      implicit ctx: ErrorContext): Infos.ClassInfo = {
-    inputProvider.loadInfo(className).getOrElse {
-      reportError(s"Cannot find info for class $className")
-      Infos.ClassInfo(className)
-    }
-  }
-
-  private def tryLookupClass(className: String)(
-       implicit ctx: ErrorContext): Either[Infos.ClassInfo, CheckedClass] = {
```
sjrd: After this removal, is the polyfill for `Either.foreach` at the bottom of this file still needed?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -281,6 +284,20 @@ object Infos {
...
+      addReferencedClass(baseNameOf(tpe))
+
+    def addReferencedClass(cls: String): this.type = {
+      referencedClasses += cls
+      this
+    }
+
+    def maybeAddReferencedClass(tpe: Type): this.type = tpe match {
+      case ClassType(cls) => addReferencedClass(cls)
+      case ArrayType(ref) => addReferencedClass(ref)
```
sjrd: We usually use `typeRef` instead of `ref` as the name of a val that represents a `TypeRef`.

sjrd:
> What about field types?

One thing we could do is add their references to each constructor in the class. A bit of a hack, but that will ensure that their types are only reached if the class `hasInstances`.

> Do we need to benchmark this (or do we not have a choice anyways)?

We could choose to avoid populating `referencedClasses` (in `Infos`) if IR-checking is disabled. We could also always disable it in the `Refiner`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
gzm0: Indeed. Fixed.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
gzm0: Done.

gzm0:
Higher level comments are still unresolved:

- Fields: Putting them in constructors needs more work due to the caching in the Refiner (might be doable, need to check).
- Benchmarking: I assume you have seen the analysis on #3558, so I'll try and investigate what makes the emitter slow.

gzm0:
I have simply added field types as a separate info. This seemed easier and notably didn't result in a strange caching situation in the refiner. Speed investigation still pending.

gzm0:
With and without this change, it seems helloworld and reversi fast opts do not change (at all).

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -47,9 +48,10 @@ object Infos {
         superClass: Option[String] = None,
         interfaces: List[String] = Nil,
         methods: List[MethodInfo] = Nil,
+        referencedFieldClasses: List[String] = Nil,
```
sjrd: Nitpick: I would order that before `methods`. We typically do fields before methods.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -737,6 +737,11 @@ private final class Analyzer(config: CommonPhaseConfig,
           (isScalaClass || isJSClass || isNativeJSClass)) {
         isInstantiated = true
 
+        for (className <- data.referencedFieldClasses) {
```
sjrd: Hum, I just realized that this is going to miss references to the types of *static* fields if the class is accessed but not instantiated. At least, here, we should only process instance fields. Static fields should be dealt in a different way.

Now, actually static fields might not need any specific treatment, for a side-reason: they are reached iff they are read or written, but when they are read or written, we'll also reach their type because it is the `tpe` of the `SelectStatic`. It's a bit of a roundabout way to look at things, though. :-s If we rely on that we should document it in a comment somewhere.

gzm0:
Updated. Still investigating the speed issue.

gzm0:
More benchmarking:

benchmarks

Observation: A significantly larger amount of time seems to be taken while emitting trees to the JS builder.

sjrd:
> Observation: A significantly larger amount of time seems to be taken while emitting trees to the JS builder.

It could be the mere fact of iterating through the set of all linked classes, then going to the caches of its constituents, even if at the end of the day there's nothing to actually emit ?

gzm0:
> It could be the mere fact of iterating through the set of all linked classes, then going to the caches of its constituents, even if at the end of the day there's nothing to actually emit ?

Yes, that's what I thought as well. Also we are sorting all linked classes. However, the "emit trees" only contains [this code][1]. So all the JS trees have already been built.

[1]: https://github.com/scala-js/scala-js/blob/c90e9af069f836f4643f11b86016d7492a4db188/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala#L214-L262

sjrd:
Hum. What's `generatedClasses.size` before and after the change? Is it significantly bigger?

Is the final `.js` file actually the same before and after the change? Or did we perhaps actually introduce more code in the output?

gzm0:
You will love this. Size of `generatedClasses`

master: 3243
reach-types: 3283

No difference in fastOptJS file.

sjrd:
Indeed, it does look like good results. But that means it's not enough to explain the increased emission time.

gzm0:
I was being sarcastic... It does not explain the increased emission times *at all*... :(

Btw. this makes me think if we need to worry about codebases that have much more JS types. Because these are basically the ones we reach now and we didn't before...

sjrd:
At this point I'd try hacking up Refiner not to reach the `referencedClasses`, and see if that has an impact on the Emitter.

gzm0:
I have tested two more things:

- [`filter-empty`][1] filter unneeded LinkedClasses at the beginning of the Emitter.
- [`partial-reach`][2] do not reach referenced classes in the Refiner

[1]: https://github.com/scala-js/scala-js/compare/master...gzm0:filter-empty?expand=1
[2]: https://github.com/scala-js/scala-js/compare/master...gzm0:partial-reach?expand=1

Benchmarks:

speed

Size of `generatedClasses`

master: 3243
reach-types: 3283
filter-empty: 3021 (seems the refiner has a bug?)
partial-reach: 3243

I'm a bit at a loss of how to make sense of this. Feels a bit like a cop out to blame it on the JIT...

sjrd:
Yeah, the bump in "Emit trees" since master doesn't make sense, given the reported sizes of `generatedClasses`.

I guess we just have to live with it, then. And given the benchmarks, I would keep the `reach-types` variant.

gzm0:
Rebased.

PR #3558, opened 2019-02-03
Make Linking async

gzm0:


gzm0:
speed

Observations:
- Expected: Reaching more makes linking / analysis slower.
- Expected: Async makes linking faster.
- Unexpected: Reaching more makes emitter slower.
- **Unexpected**: Async makes emitter slower.

Last two points probably need investigation.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -640,10 +709,14 @@ private final class Analyzer(config: CommonPhaseConfig,
         right: ir.Types.TypeRef): Boolean = {
       import ir.Types._
 
+      def getClassInfo(name: String): Option[ClassInfo] =
```
sjrd: This method should have `{}`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -640,10 +709,14 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+        // TODO: This is suspicious: It shouldn't be necessary.
```
sjrd: Any idea why/when it's necessary? What should be the expected implementation?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -1046,20 +1122,66 @@ object Analyzer {
...
+        allowAddingSyntheticMethods, inputProvider, ex)
+    analyzer.computeReachability().map(_ => analyzer)
   }
 
   trait InputProvider {
-    def classesWithEntryPoints(): TraversableOnce[String]
+    def classesWithEntryPoints()(implicit ex: ExecutionContext): Future[TraversableOnce[String]]
 
-    def loadInfo(encodedName: String): Option[Infos.ClassInfo]
+    def loadInfo(encodedName: String)(implicit ex: ExecutionContext): Option[Future[Infos.ClassInfo]]
```
sjrd: Shouldn't this be a `Future[Option[...]]` instead?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -275,123 +361,106 @@ private final class Analyzer(config: CommonPhaseConfig,
...
-      superClass = data.superClass.map(lookupClassAndCheckCycles)
-      interfaces = data.interfaces.map(lookupClassAndCheckCycles)
+    _classInfos(encodedName) = this
 
-      // j.l.Object is special and is validated upfront
-      if (encodedName != ObjectClass) {
-        validateSuperClass()
-        validateInterfaces()
-      }
+    def linked()(implicit from: From): Unit = {
```
sjrd: The past participle confused me several times. Call it `link()` instead?

linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
```diff
@@ -48,54 +48,71 @@ final class BaseLinker(config: CommonPhaseConfig) {
...
+      val linkResult = logger.time("Linker: Assemble LinkedClasses") {
+        assemble(moduleInitializers, analysis)
+      }
+
+      if (checkIR) {
+        logger.time("Linker: Check IR") {
+          val errorCount = IRChecker.check(linkResult, logger)
+          if (errorCount != 0) {
+            throw new LinkingException(
+              s"There were $errorCount IR checking errors.")
```
sjrd: 4 spaces.

linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala
```diff
@@ -379,17 +412,21 @@ class AnalyzerTest {
         ))
     )
 
-    val analysis = computeAnalysis(classDefs,
+    val result = for {
+      analysis <- computeAnalysis(classDefs,
         reqsFactory.instantiateClass("LA", "init___") ++
```
sjrd: 4 spaces.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -28,90 +34,114 @@ import Analysis._
 private final class Analyzer(config: CommonPhaseConfig,
     symbolRequirements: SymbolRequirement,
     allowAddingSyntheticMethods: Boolean,
-    inputProvider: Analyzer.InputProvider)
+    inputProvider: Analyzer.InputProvider,
+    executionContext: ExecutionContext)
```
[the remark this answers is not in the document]
gzm0: I'm concerned that not making `executionContext` explicit might cause unwanted parallelism in this file (same below).

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -1046,20 +1122,66 @@ object Analyzer {
   def computeReachability(config: CommonPhaseConfig,
       symbolRequirements: SymbolRequirement,
       allowAddingSyntheticMethods: Boolean,
-      inputProvider: InputProvider): Analysis = {
+      inputProvider: InputProvider)(implicit ex: ExecutionContext): Future[Analysis] = {
```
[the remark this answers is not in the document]
gzm0: We seem to have in our codebase: `executionContext` (`test-adapter`), `ex` (`test-common`). I don't  really have a strong opinion, but I'd like it to be consistent. LMK what you would like me to rename everything to :)

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -1046,20 +1122,66 @@ object Analyzer {
...
+    def classesWithEntryPoints()(implicit ex: ExecutionContext): Future[TraversableOnce[String]]
 
-    def loadInfo(encodedName: String): Option[Infos.ClassInfo]
+    def loadInfo(encodedName: String)(implicit ex: ExecutionContext): Option[Future[Infos.ClassInfo]]
   }
 
-  private final case class CyclicDependencyException(
-      encodedClassNames: List[String], root: Analysis.ClassInfo)
-      extends Exception(s"Cyclic dependency: $encodedClassNames")
+  private class WorkQueue(ex: ExecutionContext) {
```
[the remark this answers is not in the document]
gzm0: Same argument about unwanted parallelism...

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
gzm0: It is mainly the original code that is suspicious: It assumes the class is already loaded but allows and ignores non-existent classes. That doesn't make any sense to me.

In fact, it seems that this also works around the limitation that note all type refs are linked. I'll have a look.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
gzm0: Well, in the current implementation we do know whether a file exists or not without an async operation (and this will stay that way). So we can make it a `Option[Future[...]]`.

gzm0:
Updated

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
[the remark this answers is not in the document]
sjrd: I'd be happy with `executionContext` or `ec`, with a slight preference for `ec`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
gzm0: So, I had a closer look at this and it seems that this (and the previous code) simply *ignores* if the result type of a potential reflective call target is not loaded. I suspect this could lead to non-deterministic behavior; I will try force load class infos properly here.

gzm0:
OK. I'll go forward with this and try to address the TODO separately.

gzm0:
Updated.

PR #3560, opened 2019-02-07
Update overrides of 2.13 to the latest nightly.

sjrd:
Porting changes from https://github.com/scala/scala/pull/7624.

Locally tested with
```
> ++2.13.0-pre-596c123
> testSuite/test
> compiler/test
```


gzm0:
It seems that overrides are really becoming a problem. Have we investigated separating platform specific things in the Scala library itself? If Scala.js keeping up with Scala is so important, it seems it would warrant that little effort in the Scala library.

sjrd:
I believe overrides have always been an issue when Scala got close to a new major version. It's just that we forgot how that went the last time around, when 2.12 came out. It doesn't help that 2.13 was specifically a library release, so it contains more changes in the stdlib than in other Scala versions.

I'd love to see a more systematic way of dealing with this. I'm not sure platform-specific code in scala/scala would help, because the CI of scala/scala wouldn't catch regressions, and we would be in a similar situation, only doing PRs in scala/scala rather than scala-js/scala-js.

What I've always wanted is for the overrides to be patch-based rather than whole-file based. That way, we would have to change them less often (for example this change could have been avoided, since it isn't near any platform-specific code in the file).

PR #3568, opened 2019-02-19
Improve time logging

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -208,63 +208,72 @@ final class Emitter private (config: CommonPhaseConfig,
...
-        emitModuleInitializer(moduleInitializer, builder)
+        // Emit the module initializers
+        for (moduleInitializer <- unit.moduleInitializers)
+          emitModuleInitializer(moduleInitializer, builder)
 
-      emitPostlude
+        emitPostlude
+      }
 
       trackedGlobalRefs ++ coreJSLibTrackedGlobalRefs
```
sjrd: `coreJSLibTrackedGlobalRefs` does not compile here.

PR #3572, opened 2019-02-22
Rename all ExecutionContext parameters to `ec`

gzm0:
Of course I forgot to rename them in the previous PR.


sjrd:
```
[error] linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala:464: not found: value ex
[error]         Future(classesWithEntryPoints0)(ex)
[error]                                         ^
[error] linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala:468: not found: value ex
[error]         own.orElse(stdlib.flatMap(_.loadInfo(encodedName))).map(Future(_)(ex))
[error]                                                                           ^
[error] two errors found
```

gzm0:
Updated.

PR #3576, opened 2019-02-28
Fix #3575: Do not reverse non-equality comparisons between floats.

### LINKED ISSUE #3575: Aggressive Float/Double Optimizations, e.g. `! ( x <= 0 )`  ⟶ `x > 0`
The follwing code prints `true` on the JVM and `false` on Node:
```scala
def f( x: Double ) = ! (x <= 0)

// ...

println( f(Double.NaN) )
```
It appears that ScalaJS aggressively optimizes `! (x <= 0)` to `x > 0`. With `scalaJSOptimizerOptions ~= (_ withDisableOptimizer true)`, the Node result is consistent with the JVM.

sjrd:
The reversed operations do not have the same result when one of the operands is `NaN`.

And this is officially the highest ratio of tests versus implementation that we ever had, since it is actually *negative* :stuck_out_tongue_closed_eyes:

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/DoubleTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/DoubleTest.scala
@@ -48,4 +48,58 @@ class DoubleTest {
     test(-2147483565.123, -2147483565)
     test(-65.67, -65)
   }
+
+  @Test
+  def noReverseComparisons_issue3575(): Unit = {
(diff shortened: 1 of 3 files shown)


gzm0:
> [...] the highest ratio [...], since it is actually _negative_

`max { ratio } < 0`? Yet another comparison problem?

sjrd:
Ah ah, yes ^^ I should have said the lowest ratio implementation/tests instead :p

PR #3578, opened 2019-03-06
Make VirtualIRFiles async

gzm0:


gzm0:
Needs benchmarks.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRRepo.scala
```diff
@@ -22,26 +23,39 @@ import org.scalajs.linker.irio._
...
   val minilib = new TestIRRepo(StdlibHolder.minilib)
   val fulllib = new TestIRRepo(StdlibHolder.fulllib)
+
+  class InfoLoader(encodedNameToFile: Map[String, VirtualScalaJSIRFile]) {
+    private val infosCache = mutable.Map.empty[String, Future[ClassInfo]]
+
+    def loadInfo(encodedName: String)(
+        implicit ec: ExecutionContext): Option[Future[ClassInfo]] = {
+      infosCache.synchronized {
+        infosCache.get(encodedName).orElse {
```
sjrd: This never fills in the cache. Did you mean `getOrElseUpdate`?

test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala
```diff
@@ -45,10 +44,11 @@ object QuickLinker {
       .withSourceMapURI(relURI(smPath))
       .withJSFileURI(relURI(outputPath))
 
-    linker.link(ir, moduleInitializers, out, new ScalaConsoleLogger).toJSPromise
+    extractIR(irFilesAndJars).flatMap(ir =>
```
sjrd: Consider using `{}` for the `flatMap` here.

linker/shared/src/main/scala/org/scalajs/linker/irio/MemIRFiles.scala
```diff
@@ -12,10 +12,27 @@
...
+  def entryPointsInfo: ir.EntryPointsInfo =
+    withInputStream(ir.Serializers.deserializeEntryPointsInfo)
+
+  def tree: ir.Trees.ClassDef =
+    withInputStream(ir.Serializers.deserialize)
+
+  @inline
+  private def withInputStream[A](f: InputStream => A): A = {
+    val stream = new ByteArrayInputStream(content)
+    try f(stream)
```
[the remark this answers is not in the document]
gzm0: Because the path is not a real path. I can add it if you feel strongly.

linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
```diff
@@ -299,15 +297,16 @@ private object BaseLinker {
...
+          cacheUpdate = Future(irFile.tree).map(t => (t, Infos.generateClassInfo(t)))
         }
       }
+
+      cacheUpdate
     }
 
     /** Returns true if the cache has been used and should be kept. */
-    def cleanAfterRun(): Boolean = synchronized {
+    def cleanAfterRun(): Boolean = {
```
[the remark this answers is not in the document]
gzm0: It never needed to: It does not make sense to call this in a race with other methods. So IMO it's best not to pretend one could.

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -26,18 +28,23 @@ class NodeVirtualScalaJSIRFile(val path: String, val relativePath: String) exten
...
   @inline
-  private def withInputStream[A](f: InputStream => A): A = {
-    val buf = new Uint8Array(NodeFS.readFileSync(path)).buffer
-    val stream = new ArrayBufferInputStream(buf)
-    try VirtualScalaJSIRFile.withPathExceptionContext(path)(f(stream))
-    finally stream.close()
+  private def withInputStream[A](f: InputStream => A)(
+      implicit ec: ExecutionContext): Future[A] = {
+    def read() = {
+      val buf = new Uint8Array(NodeFS.readFileSync(path)).buffer
```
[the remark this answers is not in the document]
gzm0: Yes. I meant to delay all this, but it's probably better to put it in this PR. I'll add commits.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -67,17 +68,22 @@ final class FileVirtualScalaJSIRFile(
...
 
   @inline
-  private def withInputStream[A](f: InputStream => A): A = {
-    val stream = new BufferedInputStream(new FileInputStream(file))
-    try VirtualScalaJSIRFile.withPathExceptionContext(path)(f(stream))
-    finally stream.close()
+  private def withInputStream[A](f: InputStream => A)(
+      implicit ec: ExecutionContext): Future[A] = {
+    def read() = {
+      val stream = new BufferedInputStream(new FileInputStream(file))
```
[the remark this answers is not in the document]
gzm0: Same here about making it async.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRRepo.scala
gzm0: Hah, that seems to always have been the case :P Will fix.

linker/shared/src/main/scala/org/scalajs/linker/irio/MemIRFiles.scala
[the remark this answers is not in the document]
sjrd: I think it's useful anyway, if only so that we have a string to grep for, to find the code that's creating the problematic in-memory file.

linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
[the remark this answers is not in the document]
sjrd: Isn't it necessary at least for the safe publication of `cacheUsed` to other threads?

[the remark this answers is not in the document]
gzm0: Yes, indeed. Good point.

linker/shared/src/main/scala/org/scalajs/linker/irio/MemIRFiles.scala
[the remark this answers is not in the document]
gzm0: Ack. Changed.

gzm0:
Updated with async I/O.

Benchmarks speak slightly in favor of keeping non-async I/O for the JVM side (I also expect it to be less memory heavy).

speed

project/Build.scala
```diff
@@ -1416,7 +1413,7 @@ object Build {
           val run = jsEnv.start(input, config)
           Await.result(run.future, Duration.Inf)
           Set(out)
-        } ((irFiles :+ linkerModule).toSet)
+        } ((cp :+ linkerModule).toSet)
```
gzm0: @sjrd I'm not sure whether this is a problem. Basically we pass directories instead of actual files now.

sjrd:
What do the 4 categories in the benchmark refer to?

gzm0:
`full-async`: this PR
`pseudo-async`: this PR w/o the last commit
`direct`: this PR which `allocateDirect`

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -44,10 +46,46 @@ object Serializers {
     new Deserializer(stream).deserializeEntryPointsInfo()
   }
 
+  def deserializeEntryPointsInfo(mkBuf: Int => ByteBuffer,
+      readMore: ByteBuffer => Future[Unit])(
```
gzm0: TBH, I really dislike this interface, but IDK what to do better (without introducing duplication).

gzm0:
> I believe ScalaJSIRContainer.version should also become a Future[Option[String]] in the last commit.

I hoped to address this in the commit message. Maybe the new commit makes it clearer. LMK if it is still unclear.

ir/src/main/scala/org/scalajs/ir/Serializers.scala
sjrd: Couldn't we just take a `ByteBuffer` (like `deserialize`) and let the `BufferUnderflowException` propagate, documenting that if it is thrown, the deserialization should be attempted again from scratch, with more data?

Callers that have an already complete `ByteBuffer` wouldn't have anything to do (e.g., in-memory, which includes the in-memory extraction of jar contents). And callers who try to be smart about reading the raw data in the first place can do so in the most appropriate way.

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -19,77 +19,217 @@ import scala.concurrent._
...
-    val zip = new JSZip(NodeFS.readFileSync(path))
+      ir.Serializers.deserializeEntryPointsInfo(ByteBuffer.allocateDirect, read)
+    }
+
+    VirtualScalaJSIRFile.withPathExceptionContext(path, result)
+  }
 
+  def tree(implicit ec: ExecutionContext): Future[ir.Trees.ClassDef] = {
+    val result = cbFuture[js.Array[Int]](FS.readFile(path, _)).map { arr =>
+      val stream = new ArrayBufferInputStream(new Uint8Array(arr).buffer)
```
sjrd: This could use a `ByteBuffer` instead, constructed with `TypedArrayBuffer.wrap`.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -25,20 +27,30 @@ trait FileScalaJSIRContainer extends ScalaJSIRContainer {
...
-        List(new FileVirtualJarScalaJSIRContainer(entry))
+        fromJar(entry).map(List(_))
       else
         throw new IllegalArgumentException("Illegal classpath entry " + entry)
-    }
+    }.map(_.flatten)
+  }
+
+  def fromJar(file: File)(implicit ec: ExecutionContext): Future[FileScalaJSIRContainer] =
+    Future(new FileVirtualJarScalaJSIRContainer(file))
```
sjrd: 'Future.successful`?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -25,20 +27,30 @@ trait FileScalaJSIRContainer extends ScalaJSIRContainer {
...
+
+  def fromSingleFile(file: File, relativePath: String)(
+      implicit ec: ExecutionContext): Future[FileScalaJSIRContainer] = {
+    Future(new FileVirtualScalaJSIRFile(file, relativePath))
```
sjrd: 'Future.successful`?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -68,26 +80,38 @@ final class FileVirtualScalaJSIRFile(
     else Some(file.lastModified.toString)
   }
 
-  def entryPointsInfo(implicit ec: ExecutionContext): Future[ir.EntryPointsInfo] =
-    withInputStream(ir.Serializers.deserializeEntryPointsInfo)
+  def entryPointsInfo(implicit ec: ExecutionContext): Future[ir.EntryPointsInfo] = {
+    val chan = AsynchronousFileChannel.open(file.toPath)
+    val result = ir.Serializers.deserializeEntryPointsInfo(
+      ByteBuffer.allocate _, AsyncIO.read(chan, _))
```
sjrd: 4 spaces

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -68,26 +80,38 @@ final class FileVirtualScalaJSIRFile(
     else Some(file.lastModified.toString)
   }
 
-  def entryPointsInfo(implicit ec: ExecutionContext): Future[ir.EntryPointsInfo] =
-    withInputStream(ir.Serializers.deserializeEntryPointsInfo)
+  def entryPointsInfo(implicit ec: ExecutionContext): Future[ir.EntryPointsInfo] = {
+    val chan = AsynchronousFileChannel.open(file.toPath)
```
sjrd: Doesn't `chan` need to be explicitly closed at some point?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -68,26 +80,38 @@ final class FileVirtualScalaJSIRFile(
...
+    val result = ir.Serializers.deserializeEntryPointsInfo(
+      ByteBuffer.allocate _, AsyncIO.read(chan, _))
+    VirtualScalaJSIRFile.withPathExceptionContext(path, result)
+  }
 
-  def tree(implicit ec: ExecutionContext): Future[ir.Trees.ClassDef] =
-    withInputStream(ir.Serializers.deserialize)
+  def tree(implicit ec: ExecutionContext): Future[ir.Trees.ClassDef] = {
+    val chan = AsynchronousFileChannel.open(file.toPath)
```
sjrd: Doesn't `chan` need to be explicitly closed at some point?

project/Build.scala
sjrd: The [doc of `FileFunction`](https://www.scala-sbt.org/1.0/api/sbt/util/FileFunction$.html) is not very clear about this, but I believe it needs an actual set of files, not directories.

sjrd:
Are the `deserialize` overloads taking `InputStream`s still used? If not I believe we could reverse the normal mode of serialize/deserialize to always use `ByteBuffer`s, and potentially reclaim a bit more performance for that case.

gzm0:
> Are the deserialize overloads taking InputStreams still used? If not I believe we could reverse the normal mode of serialize/deserialize to always use ByteBuffers, and potentially reclaim a bit more performance for that case.

This is an excellent idea. However, `ByteBuffer` doesn't have `readUTF` support, so we'd have to build it ourselves (or use a different format). Not sure what to do :S

sjrd:
We can build it ourselves. We already have it anyway in our implementation of `DataInputStream`.

gzm0:
Seems we have an endiannes problem: Node.js reads `534afeca` instead of `cafe534a`.

sjrd:
According to [the JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html), `ByteBuffer`s are always `BIG_ENDIAN` by default. Did we mess up something in our implementation of `TypedArrayBuffer`s, perhaps?

gzm0:
No. The usage is messed up: `TypedArrayBuffer` have the endianness of the system but the SJS serialization is big endian. So we have to explicitly set the buffer to big endian (or require a big endian buffer).

sjrd:
Hum, looks like `TypedArrayBuffer.wrap` sets the endianness to the native endianness of typed arrays, which correspond to the native endianness of the CPU, so typically `LITTLE_ENDIAN`:
https://github.com/scala-js/scala-js/blob/ff7c2a5b625aca42b96ac3f07c86f57362cf2505/javalib/src/main/scala/java/nio/TypedArrayByteBuffer.scala#L234-L238

You could fix it by adding `.order(java.nio.ByteOrder.BIG_ENDIAN)` at https://github.com/scala-js/scala-js/pull/3578/commits/1996b7273207404eee2590c0b07cf4edf9d90bd4#diff-d76a1cead44ae0221c6eb29111c9f774R148

gzm0:
Yeah, I put it in `Serializers` instead. LMK if you have preferences.

gzm0:
Also, this is ready for review. Just needs another round of benchmarks :)

gzm0:
Benchmarks: speed

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -19,77 +19,245 @@ import scala.concurrent._
...
 
-import java.io._
+import java.nio._
 
 import org.scalajs.ir
 
-class NodeVirtualScalaJSIRFile(val path: String, val relativePath: String) extends VirtualScalaJSIRFile {
-  val version: Option[String] =
-    NodeFS.statSync(path).mtime.map(_.getTime.toString).toOption
+trait NodeScalaJSIRContainer extends ScalaJSIRContainer {
```
sjrd: What is this trait useful for? It doesn't seem like anything uses it, or could use `ScalaJSIRContainer` instead.

(I'm only talking about the trait, not its companion object.)

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -19,77 +19,245 @@ import scala.concurrent._
...
+  private def stat(path: String)(implicit ec: ExecutionContext): Future[FS.Stats] =
+    cbFuture[FS.Stats](FS.stat(path, _))
+
+  private def version(stats: FS.Stats): Option[String] =
+    stats.mtime.map(_.getTime.toString).toOption
 
-    VirtualScalaJSIRFile.withPathExceptionContext(path, Future(blocking(read())))
+  private def isNotFound(e: js.Error): Boolean = {
+    val code = e.asInstanceOf[js.Dynamic].code.asInstanceOf[js.UndefOr[String]]
+    code.exists(_ == "ENOENT")
```
sjrd: Consider just doing
```scala
(e.asInstanceOf[js.Dynamic].code: Any) == "ENOENT"
```

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
```diff
@@ -19,77 +19,245 @@ import scala.concurrent._
...
+    val path: String,
+    val relativePath: String,
+    val version: Option[String]
+) extends VirtualScalaJSIRFile with NodeScalaJSIRContainer {
+  import NodeInterop._
 
-  val version: Option[String] =
-    NodeFS.statSync(path).mtime.map(_.getTime.toString).toOption
+  def entryPointsInfo(implicit ec: ExecutionContext): Future[ir.EntryPointsInfo] = {
+    def loop(fd: Int, buf: ByteBuffer): Future[ir.EntryPointsInfo] = {
```
sjrd: I think this loop never terminates if the file itself does not contain enough data for the entry points to be read. There should be a stop when `bytesRead` is 0 or some other condition triggered by Node.js' `fs.read` when the EOF has been reached.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
sjrd: Bump:
```suggestion
    Future.successful(new FileVirtualJarScalaJSIRContainer(file))
```
?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
sjrd: Bump:
```suggestion
    Future.successful(new FileVirtualScalaJSIRFile(file, relativePath))
```
?

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
sjrd: Bump?

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
gzm0: For consistency with the FileIRFile API: A `FileScalaJSIRContainer` exposes its underlying file (and we use this in the sbt plugin). So for Node I think we should be consistent.

linker/js/src/main/scala/org/scalajs/linker/irio/NodeVirtualIRFiles.scala
gzm0: Euh... yes, indeed. Of course the Node.js documentation doesn't mention anything about that :(

gzm0:
Updated. Note that in `FileVirtualIRFiles` the calls to `Future` should not be replaced by calls to `Future.successful`, because they actually do work. I have made this explicit by adding `blocking` markers (that should probably be there anyways).

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -40,12 +42,29 @@ object Serializers {
...
+   *      in the buffer. In this case the buffer's position is unspecified and
+   *      needs to be reset by the caller.
+   */
+  def deserialize(buf: ByteBuffer): ClassDef =
+    withBigEndian(buf)(new Deserializer(_).deserialize())
+
+  @inline
+  private def withBigEndian[T](buf: ByteBuffer)(body: ByteBuffer => T): T = {
+    val o = buf.order()
+    try body(buf)
```
sjrd: Didn't you forget to actually call `buf.order(ByteOrder.BIG_ENDIAN)` here?

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -1168,18 +1179,78 @@ object Serializers {
...
+
+        val char = {
+          if ((a & 0x80) == 0x00) { // 0xxxxxxx
+            a.toChar
+          } else if ((a & 0xE0) == 0xC0 && i < length) { // 110xxxxx
+            val b = buf.get()
+            i += 1
+
+            if ((b & 0xC0) != 0x80) // 10xxxxxx
+              badFormat(f"Expected 2 bytes, found: $b%#02x (init: $a%#02x)")
```
sjrd: Oh, I almost missed this: we need to avoid using the `f` interpolator in the `ir/` project, otherwise dotty cannot compile the sources as part of its own compiler (since it's a macro, and when dotty compiles its own compiler, macros don't work). Explicitly use `"...".format(...)` instead.

See also https://github.com/scala-js/scala-js/pull/3597

(there are a few other `f"..."` in this method)

ir/src/main/scala/org/scalajs/ir/Serializers.scala
gzm0: Yes...

gzm0:
Updated

gzm0:
Updated (removed whitespace).

PR #3582, opened 2019-03-11
Do not exclude bridges when looking for JUnit-related methods.

sjrd:
It turns out that this was only necessary for one very peculiar test method, which was defined as

    @Test def toString()(): Unit = ...

The double-`()` allowed it not to clash with the `toString()` inherited from `Object`, but then for some reason erasure was generating a bridge (although it shouldn't).

Instead of having the JUnit plugin working around that single method, we rename it.

Not excluding bridges is more regular with respect to how JUnit works on the JVM.


sjrd:
That would be necessary to play nice with https://github.com/scala/scala/pull/7843 upstream, should it be merged. While there is discussion on that PR, this one should not be merged.

gzm0:
LGTM

sjrd:
Marking as Ready for review, as https://github.com/scala/scala/pull/7843 was merged upstream.

Locally tested with:
```
> ++2.13.0-pre-b851a9f
> testSuite/test
> compiler/test
```
(`testSuite/test` with that Scala version does not pass on the tip of `0.6.x`)

PR #3584, opened 2019-03-12
Fix #3538: Ignore mixin forwarders when looking for export alternatives.

### LINKED ISSUE #3538: @JSExport on method with variable arguments in trait behaves different from class
In Scala.js 0.6.26 the from JavaScript accessible method `possiblyBroken` throws an Exception when being called from JavaScript. A look in the compiled code shows, that it lacks the code for handling variable arguments that `notBroken1` and `notBroken2` have.

```scala
trait Foo {
  @JSExport def possiblyBroken(args: Int*): Unit = ???
}

class Foo2 {
  @JSExport def notBroken1(args: Int*): Unit = ???
}

@JSExportTopLevel("bar")
object Bar extends Foo2 with Foo {
  @JSExport def notBroken2(args: Int*): Unit = ???
}
```

sjrd:
Mixin forwarders are cloned with a type history that does not go far enough back in time to see varargs. By skipping them when looking for the alternatives of an export, we look further in the class hierarchy to the original definition in the trait, which does expose the entire type history that we need.

Inspired by the discussion at https://github.com/scala/scala/pull/7843.
Supersedes https://github.com/scala/scala/pull/7752.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
@@ -356,6 +356,35 @@ class ExportsTest {
     assertEquals(4, bar.method(2))
   }
 
+  @Test def should_inherit_exports_from_traits_with_value_classes(): Unit = {
+    trait Foo {
+      @JSExport
+      def x: SomeValueClass = new SomeValueClass(5)
+
+      @JSExport
+      def method(x: SomeValueClass): Int = x.i
+    }
+
+    class Bar extends Foo
+
+    val bar = (new Bar).asInstanceOf[js.Dynamic]
+    assertEquals(new SomeValueClass(5), bar.x)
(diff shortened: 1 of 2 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -356,6 +356,34 @@ class ExportsTest {
     assertEquals(4, bar.method(2))
   }
 
+  @Test def should_inherit_exports_from_traits_with_value_classes(): Unit = {
```
sjrd: This test already passes before the changes in this commit. It's only related because it also relies on the type history of export bridges.

sjrd: Well, turns out it *doesn't* pass in 2.10. It seems to be an independent issue, though, so I filed it as #3585.

PR #3598, opened 2019-03-28
WrappedArray overrides knownSize (2.13 stricter refcheck compat)

adriaanm:


sjrd:
As soon as #3599 is in, I will do a merge of 0.6.x into master. Hopefully that will happen today.

PR #3599, opened 2019-03-28
Cross-compile with the deprecated `JavaConverters` without warnings.

sjrd:
In Scala 2.13, `scala.collection.JavaConverters` has been deprecated in favor of `scala.jdk.CollectionConverters.Ops`. This commit introduces some magic to keep cross-compiling across 2.10 through 2.13 without any warning in any version.

Locally tested with
```
> ++2.13.0-pre-bd28253
> testSuite/test
```

### DIFF
--- a/javalib/src/main/scala/java/util/Compat.scala
+++ b/javalib/src/main/scala/java/util/Compat.scala
@@ -56,4 +56,32 @@ private[util] object Compat {
 
   }
 
+  // !!! Duplicate code with org.scalajs.testcommon.RPCCore.JDKCollectionConvertersCompat
+  /** Magic to get cross-compiling access to `scala.jdk.CollectionConverters`
+   *  with a fallback on `scala.collection.JavaConverters`, without deprecation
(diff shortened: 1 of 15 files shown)


javalib/src/main/scala/java/util/Compat.scala
```diff
@@ -56,4 +56,32 @@ private[util] object Compat {
 
   }
 
+  // !!! Duplicate code with org.scalajs.testcommon.RPCCore.CollectionConvertersCompat
+  /** Magic to get cross-compiling access to `scala.jdk.CollectionConverters`
+   *  with a fallback on `scala.collection.JavaConverters`, without deprecation
+   *  warning in any Scala version.
+   */
+  object CollectionConvertersCompat {
```
gzm0: Maybe this name should contain JDK or Java somewhere so it is clear that this is about java collections?

sjrd: Done. Updated.

PR #3600, opened 2019-04-01
DO NOT MERGE Fix the overriding-pairs `Cursor` used by PrepJSInterop for 2.13.

sjrd:
In the upstream commit https://github.com/scala/scala/commit/fde215f606dc63813bea3b6bc64176aa551b40ce, a new abstract method `skipOwnerPair` was introduced in `Cursor`. We now implement it in the same way as `RefChecks` does, since previously we used exactly the same `Cursor` as `RefChecks`. There is some compat code to allow the new code to compile with older Scala versions.

Locally tested with
```
> ++2.13.0-pre-b4926ef
> testSuite/test
```
`compiler/test` is still broken, but not more than the last time I checked.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -684,7 +684,7 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
       }
 
       // Check for consistency of JS semantics across overriding
-      for (overridingPair <- new overridingPairs.Cursor(sym).iterator) {
+      for (overridingPair <- new JSRefChecksCursor(sym).iterator) {
         val low = overridingPair.low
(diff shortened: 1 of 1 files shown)


adriaanm:
Sorry! This really needed to be pulled out because there was too much sharing going on between very different parts of the code base.

adriaanm:
The logic in RefchecksCursor is still wrong, so you may want to hold off merging this. I'm still thinking about a fix.

sjrd:
Thanks for the heads up!

adriaanm:
I'm planning to rollback the Cursor API so that this PR would not be needed anymore. By default, it won't skip any pairs and the bridges subclass will override the method for its purposes.

adriaanm:
I've updated scala/scala#7929 to roll back the breaking changes to RefChecksCursor

sjrd:
Superseded by https://github.com/scala/scala/pull/7929 upstream.

PR #3610, opened 2019-04-08
Enable UndefOr operators for `Unit | _` and `_ | Unit`

Atry:


sjrd:
Could you motivate this please?

Atry:
1. `UndefOr` and `Unit | ?` are essentially the same type
2. Some tools produce `Unit | _` types when converting type definition from TypeScript

sjrd:
So, in fact in Scala.js 1.x `js.UndefOr[A]` *is* literally a type alias for `A | Unit`. I guess we could pimp the `UndefOrOps[A]` to `A | Unit` in 0.6.x to bridge the gap a bit.

That said, tools right now really should emit `js.UndefOr[A]` instead of `A | Unit`/`Unit | A`. I feel like this PR is trying to work around the tools inside the core, whereas in fact we should just fix the tools, not change the core.

Atry:
Is there a plan for releasing Scala.js 1.0?

sjrd:
Not really. We "just" need to clear out https://github.com/scala-js/scala-js/milestone/64

Atry:
Will the milestone break backward compatibility?

sjrd:
It's unlikely that we will break anything at the source level for user-visible stuff, at this point. So the release notes of M7 will be very similar to those of M6, and likewise for future milestones until the final release.

Atry:
Since backward-compatibility will be kept from 1.0.0-M6, why not release now and let new features and internal refactoring go to Scala.js 1.1+?

sjrd:
Because the internal refactorings affect the *binary* compatibility of the artifacts.

Atry:
It is to say it's only necessary to clear out IR related issues, isn't it? Is it possible to breaking change internal compiler API after 1.0?

> Because the internal refactorings affect the _binary_ compatibility of the artifacts.

sjrd:
The *compiler* API we break all the time (in fact we don't even consider it an API at all). But the surface of the linker API, as well as the `JSEnv` API, will also need to preserve binary compatibility after 1.x.

gzm0:
What is the state of this? @Atry would you still like us to consider this for merging or are you OK with using 1.0.x / improving tools?

Atry:
I understand that the Scala.js maintainers are not willing to accept changes that do not help in future Scala 1.0.

sjrd:
I've thought more about this, and in particular how this is going to evolve in Scala 3. Since it will have real union types, `String | Null` will be equivalent to `String`, and therefore the proposed API will not be able to kick in. Unless Scala 3 receives non-nullable types as well, that means the change introduced here will be broken by Scala 3. For that reason I'm going to close this. It could be revisited if Scala 3 gets non-nullable types.

PR #3614, opened 2019-04-09
Drop support for Scala 2.13.0-M5, replace it in the CI by 2.13.0-RC1.

sjrd:
~~Needs to be rebased on top of #3613. Only the last commit belongs to this PR.~~


ci/checksizes.sh
```diff
@@ -54,11 +54,11 @@ case $FULLVER in
...
-  2.13.0-M5)
-    REVERSI_PREOPT_EXPECTEDSIZE=613000
-    REVERSI_OPT_EXPECTEDSIZE=143000
-    REVERSI_PREOPT_GZ_EXPECTEDSIZE=81000
-    REVERSI_OPT_GZ_EXPECTEDSIZE=36000
+  2.13.0-RC1)
+    REVERSI_PREOPT_EXPECTEDSIZE=792000
+    REVERSI_OPT_EXPECTEDSIZE=188000
+    REVERSI_PREOPT_GZ_EXPECTEDSIZE=103000
+    REVERSI_OPT_GZ_EXPECTEDSIZE=47000
```
sjrd: I have not yet investigated the huge size differences. The raw diff of `reversi-fastopt.js` is at https://gist.github.com/sjrd/ababc7a7c106989ad7f907914e4f0285

sjrd: So actually most of the regression is due to https://github.com/scala/scala/commit/afe76c12d705711d54f748f5ebf6a18df22082b9 upstream, which calls into `System.getProperty` from the core collections. This in turns brings the infrastructure for `j.u.Properties` and `Map`s, and that's a lot of code. Perhaps we can backport https://github.com/scala-js/scala-js/commit/111bb36f771e31e56f44ecac32ba5405b99fa618 to fix this problem.

gzm0: #3615 doesn't seem to get the numbers fully down again. Do you know why?

sjrd: Unfortunately, a lot of class hierarchies have been altered, which changes the order of classes in the final .js files, so the diff is extremely difficult to make sense of. I'm not able to identify any single big responsible code. I believe it's the myriad of additional overrides that have been added over the months for performance reasons.

gzm0: OK. Fair enough.

PR #3615, opened 2019-04-09
[no-master] Do not require `ju.Properties` if we only use `System.{get,set,clear}Property`.

sjrd:
Based on #3614.

It is reasonably common to use `System.getProperty` and other related methods, but not to use `System.{get,set}Properties` nor to directly instantiate `ju.Properties`.

This commit uses a few laziness tricks to avoid making `ju.Properties` reachable (along with `ju.Hashtable` and some others) if the codebase only uses `System.{get,set,clear}Property`.

Cherry-picked from 111bb36.


gzm0:
I'm confused. Is this draft or ready for review?

sjrd:
Ready for review, not ready to be *merged* (because of the necessary rebase). I'm using the draft status as a "DO NOT MERGE" that is actually checked by the system. Is that confusing?

gzm0:
> I'm using the draft status as a "DO NOT MERGE" that is actually checked by the system. Is that confusing?

Ah, OK. If I know that it's not :). It's just that there is a big "Ready for review" button; so that's a bit confusing. But if I know that this is not what it means its fine :)

sjrd:
Hum ... the bootstrap tests on the CI are not happy with this. I'm not sure what's going on at the moment.

sjrd:
I don't understand what's failing. I'll close this, as it's only an optimization and fixed in 1.x.

PR #3617, opened 2019-04-12
 Fix #3580: Remove relativePath from VirtualIRFiles

### LINKED ISSUE #3580: Remove relativePath from VirtualIRFiles
Currently, `relativePath` is only used to identify files for the `scalajsp` task. As a result, we have [code like this][1] that basically fakes it.

I propose instead of using the `relativePath` to identify files for `scalajsp`, we use the `encodedName` in the `EntryPointInfo` (with conversion to/from decoded names if feasible).

This will simplify the virtual IR API for a little inconvenience of the debug-only scalajsp.

[1]: https://github.com/scala-js/scala-js/blob/b3f250494d1c49743002bbc766ec9a3de74c5131/test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala#L57-L59

gzm0:


gzm0:
I prefer the option including the second commit. WDYT?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -14,10 +14,11 @@ package org.scalajs.sbtplugin
 
 import scala.annotation.tailrec
 
-import scala.concurrent._
+// sbt.Future collides, but only exists in 0.13 so cannot unimport easily.
+import scala.concurrent.{Future => scFuture, _}
```
sjrd: Isn't it sufficient to use a named import for
```scala
import scala.concurrent.Future
```
? Named imports are supposed to take precedent over wildcard imports.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalajspUtils.scala
```diff
@@ -1,66 +0,0 @@
...
-         * to the subdirectory, '/' included. This means that:
-         *
-         * > scalajsp hello<tab>
-         *
-         * will complete to
-         *
-         * > scalajsp helloworld/
-         *
-         * and further tabs are necessary to show to files and directories
-         * under the helloworld/ subdirectory.
```
sjrd: This behavior would still be welcome with decoded names, with separations based on `.` instead of `/`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Ah, yes, good point.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalajspUtils.scala
gzm0: Well yes, but is it worth the additional code and divergence from standard sbt behavior? (e.g. `testOnly` does not do that).

sjrd: Ah, I was under the impression that it did, but you're right, it doesn't do that.

PR #3618, opened 2019-04-12
RFC Use java.nio.file.Path instead of VirtualBinaryFile

gzm0:


gzm0:
The java.nio.files._ package basically provides a virtual file system API. So I tried to use it...

sjrd:
I like the concept. But that means we'll have to implement those APIs on the JS side ...

Can we also replace `WritableVirtualBinaryFile` et al. with `Path`s?

gzm0:
So I have been looking into this a bit more and it seems we cannot fully implement those APIs on the JS side, because not everything is as async as it needs (e.g. `open` / `close` are not async in Java).

So we seem to have the following options:
- Diverge `JSEnv` from `linker` in terms of I/O handling.
- Implement our own I/O abstraction (my original plans of using `AsynchronousByteChannel` have the same issue: `close` is synchronous). Making both `JSEnv` usage and implementation tedious.
- Something in-between?

As a side note: For linker I/O output, I'm pretty convinced by now that we need our own abstraction unfortunately. I'm actively working on that one.

gzm0:
WIP of our own API is in: https://github.com/gzm0/scala-js/tree/async-write

PR #3621, opened 2019-04-13
Fix #3571: Always load reflective call return types

### LINKED ISSUE #3571: Reflective call target determination depends on whether result types are loaded
From: #3558

> So, I had a closer look at this and it seems that this (and the previous code) simply _ignores_ if the result type of a potential reflective call target is not loaded. I suspect this could lead to non-deterministic behavior; I will try force load class infos properly here.

gzm0:
I have manually verified that the test fails without the fixes.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
@@ -640,19 +640,24 @@ private final class Analyzer(config: CommonPhaseConfig,
       m
     }
 
-    def tryLookupReflProxyMethod(proxyName: String): Option[MethodInfo] = {
+    def tryLookupReflProxyMethod(proxyName: String)(
+        onSuccess: MethodInfo => Unit)(implicit from: From): Unit = {
       if (!allowAddingSyntheticMethods) {
-        tryLookupMethod(proxyName)
+        tryLookupMethod(proxyName).foreach(onSuccess)
       } else {
-        publicMethodInfos.get(proxyName).orElse {
-          findReflectiveTarget(proxyName).map { reflectiveTarget =>
-            createReflProxy(proxyName, reflectiveTarget.encodedName)
+        publicMethodInfos.get(proxyName).fold {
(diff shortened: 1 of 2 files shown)


gzm0:
It seems like this surfaces a bad interaction between default methods and reflective proxies:

```
Caused by: java.lang.AssertionError: assertion failed: Trying to dynamically reach the non-class method ju_Comparator.reversed__ju_Comparator
```

The method that causes the assertion is selected by `findProxyMatch` with the new code.

@sjrd any idea about this?

gzm0:
> The method that causes the assertion is selected by findProxyMatch with the new code.

Seems this was a bug in the new code where it attempted to reach the reflective target and not the generated proxy.

gzm0:
@sjrd I need some help here: The CI shows that the new code erroneously tries to load primitives types. So I'm unsure what to do when I encounter a name of a primitive type; It seems like I'd need something like `isSubtype` in `ir.Types` but for `TypeRef`s?

Note that whatever fix we'll find here will probably have to be backported to 0.6.

sjrd:
Hum, it's true that the code does not seem to handle primitive types. That's weird. In any case, if it encounters any primitive type, then `isMoreSpecific` should always answer `false`. At the IR/JVM level, primitive types are always unrelated to each other. (so `short` is not more specific than `int`, for example)

gzm0:
Even w.r.t. `Object`? Wouldn't e.g. `Int` be more specific than `Object` especially in view of generics and bridges?

sjrd:
Yes, even with `Object`:
```scala
scala> class A { def foo(): Any = "foo"; def foo[X](): Int = 2 }
defined class A

scala> classOf[A].getMethod("foo")
res3: java.lang.reflect.Method = public java.lang.Object A.foo()
```
And also with `java.lang.Integer`:
```scala
scala> class A { def foo(): Integer = 1; def foo[X](): Int = 2 }
defined class A

scala> classOf[A].getMethod("foo")
res4: java.lang.reflect.Method = public java.lang.Integer A.foo()
```

gzm0:
This is ready for final review.

PR #3625, opened 2019-04-25
Do not rely on JavaConverters in our implementation of java.util.

sjrd:
Instead of using some compatibility hacks to access the Java collection converters across all versions of Scala, we simply get rid of them. Instead, we implement our own small set of combinators for operations that we actually use. We use the opportunity to make those combinators zero-cost.

This PR is prompted by yet another breaking change to the API of `JavaConverters` coming in 2.13.x. See https://github.com/scala/scala/pull/7987#discussion_r278158104

### DIFF
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -14,7 +14,7 @@ package java.util
 
 import scala.annotation.tailrec
 
-import Compat.JDKCollectionConvertersCompat.Converters._
+import ScalaOps._
 
 abstract class AbstractList[E] protected () extends AbstractCollection[E]
     with List[E] {
@@ -35,7 +35,7 @@ abstract class AbstractList[E] protected () extends AbstractCollection[E]
     throw new UnsupportedOperationException
 
   def indexOf(o: Any): Int =
-    iterator.asScala.indexWhere(_ === o)
+    this.scalaOps.indexWhere(_ === o)
 
   def lastIndexOf(o: Any): Int = {
     @tailrec
@@ -52,7 +52,7 @@ abstract class AbstractList[E] protected () extends AbstractCollection[E]
 
   def addAll(index: Int, c: Collection[_ <: E]): Boolean = {
     checkIndexOnBounds(index)
-    for ((elem, i) <- c.iterator.asScala.zipWithIndex)
+    for ((elem, i) <- c.iterator().scalaOps.zipWithIndex.scalaOps)
       add(index + i, elem)
     !c.isEmpty
   }
@@ -111,14 +111,14 @@ abstract class AbstractList[E] protected () extends AbstractCollection[E]
       o match {
         case o: List[_] =>
           val oIter = o.listIterator
-          this.asScala.forall(oIter.hasNext && _ === oIter.next()) && !oIter.hasNext
+          this.scalaOps.forall(oIter.hasNext && _ === oIter.next()) && !oIter.hasNext
         case _ => false
       }
     }
   }
 
   override def hashCode(): Int = {
-    this.asScala.foldLeft(1) {
+    this.scalaOps.foldLeft(1) {
       (prev, elem) => 31 * prev + (if (elem == null) 0 else elem.hashCode)
     }
   }
--- a/javalib/src/main/scala/java/util/Collections.scala
+++ b/javalib/src/main/scala/java/util/Collections.scala
@@ -19,7 +19,9 @@ import scala.language.implicitConversions
 
 import scala.annotation.tailrec
 
-import Compat.JDKCollectionConvertersCompat.Converters._
+import scala.collection.mutable
+
+import ScalaOps._
 
 object Collections {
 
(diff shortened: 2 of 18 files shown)


javalib/src/main/scala/java/util/AbstractList.scala
```diff
@@ -52,8 +52,8 @@ abstract class AbstractList[E] protected () extends AbstractCollection[E]
 
   def addAll(index: Int, c: Collection[_ <: E]): Boolean = {
     checkIndexOnBounds(index)
-    for ((elem, i) <- c.iterator.asScala.zipWithIndex)
-      add(index + i, elem)
+    for (elemAndIdx <- c.iterator().scalaOps.zipWithIndex.scalaOps)
```
gzm0: Why not just add `withFilter`?

javalib/src/main/scala/java/util/Collections.scala
```diff
@@ -249,14 +249,14 @@ object Collections {
     min(coll, naturalComparator[T])
 
   def min[T](coll: Collection[_ <: T], comp: Comparator[_ >: T]): T =
-    coll.asScala.min(comp)
+    coll.scalaOps.reduceLeft((a, b) => if (comp.compare(a, b) <= 0) a else b)
```
gzm0: `reduceLeft(math.min)`?

javalib/src/main/scala/java/util/Collections.scala
```diff
@@ -249,14 +249,14 @@ object Collections {
...
 
   // Differs from original type definition, original: [T <: jl.Comparable[_ >: T]]
   def max[T <: AnyRef with jl.Comparable[T]](coll: Collection[_ <: T]): T =
     max(coll, naturalComparator[T])
 
   def max[T](coll: Collection[_ <: T], comp: Comparator[_ >: T]): T =
-    coll.asScala.max(comp)
+    coll.scalaOps.reduceLeft((a, b) => if (comp.compare(a, b) >= 0) a else b)
```
gzm0: `reduceLeft(math.max)`?

javalib/src/main/scala/java/util/Collections.scala
```diff
@@ -534,22 +534,41 @@ object Collections {
...
+      !c1.scalaOps.exists(elem => c2.contains(elem))
     else
-      !c2.asScala.exists(elem => c1.contains(elem))
+      !c2.scalaOps.exists(elem => c1.contains(elem))
   }
 
-  def addAll[T](c: Collection[_ >: T], elements: Array[AnyRef]): Boolean =
-    c.addAll(elements.asInstanceOf[Array[T]].toSeq.asJava)
+  def addAll[T](c: Collection[_ >: T], elements: Array[AnyRef]): Boolean = {
+    val elementsColl = new AbstractCollection[T] {
```
gzm0: Huh, couldn't you use standard array ops to make an Scala iterator out of this and then wrap it as a java iterator?

javalib/src/main/scala/java/util/Hashtable.scala
```diff
@@ -87,25 +87,87 @@ class Hashtable[K, V] private (inner: mutable.HashMap[Box[Any], V])
...
+
+  private class EntrySetIterator extends Iterator[Map.Entry[K, V]] {
+    private var underlying: scala.collection.Iterator[(Box[Any], V)] =
+      Hashtable.this.inner.iterator
+
+    private var isSnapshot: Boolean = false
+
+    private var prev: Box[Any] = null
+
+    private def ensureSnapshot(): Unit = {
```
gzm0: I assume this logic is copied from Scala itself?

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -47,19 +49,16 @@ class Properties(protected val defaults: Properties)
...
-    val thisSet = keySet().asScala.map(_.asInstanceOf[String])
-    val defaultsIterator =
-      if (defaults != null) defaults.propertyNames().asScala
-      else scala.collection.Iterator.empty
-    val filteredDefaults = defaultsIterator.collect {
-      case k: String if !thisSet(k) => k
-    }
-    (thisSet.iterator ++ filteredDefaults).asJavaEnumeration
+    val propNames = mutable.Set.empty[Any]
+    keySet().scalaOps.foreach(propNames += _.asInstanceOf[String]) // CCE mandated by spec
```
gzm0: Consider writing CCE out for ease of understanding.

javalib/src/main/scala/java/util/Collections.scala
sjrd: That would fail to use the supplied `comp` comparator. (Also `math.min` only works for private numbers, not for any `T`.)

javalib/src/main/scala/java/util/Collections.scala
sjrd: That would fail to use the supplied `comp` comparator. (Also `math.max` only works for private numbers, not for any `T`.)

javalib/src/main/scala/java/util/Hashtable.scala
sjrd: More or less, but there are adaptations. I have added a comment.

javalib/src/main/scala/java/util/Hashtable.scala
```diff
@@ -87,25 +87,96 @@ class Hashtable[K, V] private (inner: mutable.HashMap[Box[Any], V])
...
     }
-    inner.asJava.entrySet().asScala.map(new UnboxedEntry(_): ju.Map.Entry[K, V]).asJava
+
+    def remove(): Unit = {
+      if (prev == null)
+        throw new IllegalStateException("next must be called at least once before remove")
+      ensureSnapshot()
+      inner -= prev
+      prev = null
+    }
```
gzm0: It seems this codepath is completely untested or am I missing something?

sjrd: I added a test for the corner cases of the entrySet's iterator.

PR #3633, opened 2019-04-29
Fix #2870: Add warnings to duplicate export annotation.

### LINKED ISSUE #2870: Warn on duplicate JSExport?
Should we warn here:

```scala
@JSExportAll
class A {
  @JSExport
  def a = 1
}
```

or here?

```scala
class A {
  @JSExport
  @JSExport("a")
  def a = 1
}
```

or here?

```scala
class A {
  @JSExport
  @JSExport
  def a = 1
```

exoego:


exoego:
image

🤔 This is odd. JSExportTest.noExportImplicitApply fail on Scala 2.10 & 2.11, but pass on 2.12 & 2.13.
https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3633/1/pipeline#step-178-log-162
https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3633/1/pipeline#step-362-log-110

sjrd:
This often happens with compiler tests. Error messages can be quite fickle. Have you tested it locally with 2.10 or 2.11, in the hello world, to see the reported warnings?

exoego:
> 
> 
> This often happens with compiler tests. Error messages can be quite fickle. Have you tested it locally with 2.10 or 2.11, in the hello world, to see the reported warnings?

Yes, I can see the reported warnings like this
```
[warn] /home/vagrant/IdeaProjects/scala-js/examples/helloworld/HelloWorld.scala:73: Found useless @JSExport, since owner already has @JSExportTopLevel. Consider @JSExport with explicit name
[warn]   def a = 1
[warn]       ^
[warn] /home/vagrant/IdeaProjects/scala-js/examples/helloworld/HelloWorld.scala:78: Found duplicate @JSExport
[warn]   def b = 1
[warn]       ^
```

after adding some warning classes to `HelloWorld.scala` and running command on sbt like below

```
set scalaVersion := "2.10.7"
helloworld/fastOptJS
set scalaVersion := "2.11.12"
helloworld/fastOptJS
```

sjrd:
`set scalaVersion := "2.10.7"` doesn't do anything. It only changes the Scala version of the root project, which is unused.

You need to use `++2.10.7` instead.

exoego:
> You need to use `++2.10.7` instead.

The warnings are reported both on `++2.10.7` and `++2.11.12`.

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -267,6 +267,23 @@ trait PrepJSExports[G <: Global with Singleton] { this: PrepJSInterop[G] =>
         Nil
     }
 
+    val hasDuplicateExportAnnotation = directAnnots.count { annot =>
+      annot.symbol == JSExportAnnotation || annot.symbol == JSExportNamedAnnotation
+    } > 1
+    if (hasDuplicateExportAnnotation) {
+      reporter.warning(sym.pos,
+        s"Found duplicate @JSExport")
+    }
```
gzm0: Hum, wouldn't this warn for this:

```scala
@JSExport("a")
@JSExport("b")
def foo(): Int = ...
```

exoego: Do you mean such case should not be earned?

gzm0: Yes, exactly. It is reasonable to export a method under multiple names.

exoego: Fixed in https://github.com/scala-js/scala-js/pull/3633/commits/6a048ef99cb6c7df95cb08e809b86ffce474feb8

gzm0: Now from what I understand this wouldn't warn on:

```scala
@JSExport("foo")
@JSExport
def foo(): Unit = ...
```

Have you considered using the post-processed `ExportInfo`s? They neatly give you export type, destination and name for grouping.

Something like:

```scala
allExportInfos.groupBy(i => (i.jsName, i.Destination, i.isNamed).filter(_.size > 1).foreach {
  reporter.warning(...)
}
```

exoego: > 
> 
> Now from what I understand this wouldn't warn on:
> 
> ```scala
> @JSExport("foo")
> @JSExport
> def foo(): Unit = ...
> ```

I think that case is covered in https://github.com/scala-js/scala-js/pull/3633/files#diff-c700423ea542bd0634232c51084a9acaR47




> Have you considered using the post-processed `ExportInfo`s? They neatly give you export type, destination and name for grouping.

I will try refactoring using that.

gzm0: You could still report at `sym.pos` instead of `i.pos`, no? I haven't seen how it looks, but I expect it to be much less duplicated logic if you use `exportInfo`.

exoego: Hmm, I tried refactoring using `allExportInfos` but I could not since [lack of IDE support](https://github.com/scala-js/scala-js/issues/3638) give a hurdle for me to understand code base and modify...
I have a little time for cosmetic change, but not enough for refactoring since I will be busy for several weeks.
Could you consider review PR as-is and accept refactoring in future PR? 
Or, you can push changes to the branch since the branch allow edits from maintainers.

gzm0: > Could you consider review PR as-is and accept refactoring in future PR?

I'd rather not. We have a long history on trying to accumulate a minimum of technical debt. (Best example probably being #3240).

> Or, you can refactor and push changes to the branch since the branch allow edits from maintainers.

I'll do that if I have a moment. Otherwise you can take it again. "several weeks" is in my experience not a very long time, especially if you're only contributing in your free time (like I do).

exoego: I found time to refactor using `allExportInfos` !!

exoego:
CI is green and ready to review

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -466,6 +475,13 @@ trait PrepJSExports[G <: Global with Singleton] { this: PrepJSInterop[G] =>
           ignoreInvalid = false)
     }
 
+    if (!unitAnnots.exists(_.symbol == JSExportAllAnnotation)) {
```
gzm0: So this means that the following would not warn, right:

```scala
@JSExportAll
class Foo {
  @JSExport("b")
  @JSExport("b")
  def a = ???
}
```

Consider simply dropping this extra case and live with the less good warning. Alternatively bring in the heavy diagnostic:

```
Found duplicate export for name "a". Exports originating from:
    @JSExport("a")
     ^
  @JSExportAll
   ^
```

(you might want to get inspiration from duplicate member definition error messages, etc.)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -368,6 +368,15 @@ trait PrepJSExports[G <: Global with Singleton] { this: PrepJSInterop[G] =>
...
 
+          def isUnnamedExportInsideExportAll = {
+            isMember && !hasExplicitName && annot.symbol == JSExportAnnotation &&
+              unitAnnots.exists(_.symbol == JSExportAllAnnotation)
+          }
+          if (isUnnamedExportInsideExportAll) {
+            reporter.warning(sym.pos, "Found useless @JSExport inside @JSExportTopLevel."
+              + """ Maybe @JSExport("someName") ?""")
+          }
+
```
gzm0: This would also be rendered unnecessary if we remove the special case below (IIUC).

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
exoego: Simplified in https://github.com/scala-js/scala-js/pull/3633/commits/b0f391d38547cea9fc4372e11767a48dbfd15275

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
exoego: Simplified in https://github.com/scala-js/scala-js/pull/3633/commits/b0f391d38547cea9fc4372e11767a48dbfd15275

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
```diff
@@ -17,6 +17,8 @@ import org.junit.Test
 
 import org.junit.Assume._
 
+import scala.language.postfixOps
```
gzm0: Just write the dots out :)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSExports.scala
```diff
@@ -466,6 +466,17 @@ trait PrepJSExports[G <: Global with Singleton] { this: PrepJSInterop[G] =>
           ignoreInvalid = false)
     }
 
+    allExportInfos.filter(_.destination == ExportDestination.Normal)
+      .groupBy(_.jsName).
+      filter { case (jsName, group) =>
```
gzm0: The dot on the previous line is probably not style conform :) I'll leave it to @sjrd though.

sjrd: Indeed, the dot should be at the beginning of this line.

exoego: Fixed in https://github.com/scala-js/scala-js/pull/3633/commits/723f52104e4ab045cfe5719bc136bed7152353dc

compiler/src/test/scala/org/scalajs/core/compiler/test/JSExportTest.scala
exoego: Removed in https://github.com/scala-js/scala-js/pull/3633/commits/090dc49f91c691ab60c8977431fc5f54a63d6cf2

PR #3635, opened 2019-04-30
Make linker output writing async

gzm0:


gzm0:
Expected to work now, still needs benchmarks.

@sjrd I really do not like this API, it seems much too complicated, but it is the best I could come up with. I do not like that it condemns our output abstraction to be different than normal Java I/O. However, I really do not know what to do better without sacrificing either of the following two:

- Full async support.
- Future proofing for streaming support (i.e. *not* writing the full file at once).

Further, note that this API also will not allow us to partially replace files in incremental runs if we'll ever be able to do this.

sjrd:
I'm not sure what you think is too complicated with the current design. AFAICT it boils down to:
```scala
trait WritableVirtualBinaryFile {
  def newChannel()(implicit ec: ExecutionContext): Future[WriteChannel]

  def writeFull(buf: ByteBuffer)(implicit ec: ExecutionContext): Future[Unit] = ???
}

trait WriteChannel {
  def write(buf: ByteBuffer)(implicit ec: ExecutionContext): Future[Unit]
  def close()(implicit ec: ExecutionContext): Future[Unit]
}
```
which looks quite reasonable. Isn't it?

gzm0:
Well, on its own yes. But it won't work with any other library. Before that, we at least had standard Java streams. Now everybody will have to write a wrapper to wrap this into another abstraction. That's not very user friendly.

sjrd:
> Now everybody will have to write a wrapper to wrap this into another abstraction. That's not very user friendly.

Well ... "everybody" is people who 1) interact with the linker API and 2) are not happy with `Path`-based files *nor* memory-based files. That leaves a very small set of users to which we present an unfriendly API. I wouldn't be too concerned about that.

gzm0:
Benchmarks:

speed

gzm0:
Looks like we have a slight, non-significant slowdown in emitter. If you like I can increase `n` (currently `n = 10`).

sjrd:
The benchmarks look good to me. We should be able to reclaim the possible small regression in the future by writing to disk asynchronously.

sjrd:
Unsurprisingly, this PR needs a rebase now that #3646 is merged.

linker/jvm/src/main/scala/org/scalajs/linker/irio/FileVirtualIRFiles.scala
```diff
@@ -21,6 +21,7 @@ import java.nio.channels._
 import java.util.zip.{ZipInputStream, ZipEntry}
 
 import org.scalajs.ir
+import org.scalajs.io._
```
sjrd: Useless import?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -169,35 +177,36 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
...
       sm.setWrapperPrefix(header)
-      val w = writer(smf.outputStream)
-      try sm.appendTo(w, output.jsFileURI.fold("")(_.toASCIIString))
-      finally w.close()
+      writeToFile(smf) { w =>
+        sm.appendTo(w, output.jsFileURI.fold("")(_.toASCIIString))
+      }
     }
+
+    Future.sequence(List(codeWritten) ++ smWritten).map(_ => ())
```
sjrd: The `Future.sequence` seems overkill. Consider
```scala
smWritten.fold(codeWritten)(_.flatMap(_ => codeWritten))
```

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -17,6 +17,7 @@ import scala.concurrent._
 
 import java.io._
 import java.net.URI
+import java.nio.{ByteBuffer, CharBuffer}
```
sjrd: `CharBuffer` seems unused.

io/shared/src/main/scala/org/scalajs/io/package.scala
```diff
@@ -10,14 +10,14 @@
...
+package org.scalajs
 
 import scala.concurrent._
 
 import scala.util.{Try, Success, Failure}
 
-package object irio {
-  private[irio] implicit class FutureOps[T](val __self: Future[T]) extends AnyVal {
+package object io {
+  private[scalajs] implicit class FutureOps[T](val __self: Future[T]) extends AnyVal {
```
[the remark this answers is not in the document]
sjrd: Actually, you can probably avoid the problem altogether if you first get rid of the `io` package, since everything would then be in the `scalajs-linker` artifact anyway.

gzm0:
Updated and rebased on top of #3655.

gzm0:
Rebased on master

linker/js/src/main/scala/org/scalajs/linker/irio/NodeFS.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @JSImport("fs", JSImport.Namespace)
+  @js.native
+  object FS extends js.Object {
+    def open(path: String, flags: String, callback: CB[Int]): Unit = js.native
+    def close(fd: Int, callback: CB[Unit]): Unit = js.native
+
+    def read(fd: Int, buffer: TypedArray[_, _], offset: Int, length: Int, position: Int,
+      callback: CB[Int]): Unit = js.native
```
sjrd: 4 spaces

linker/js/src/main/scala/org/scalajs/linker/irio/NodeFS.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    def write(fd: Int, buffer: TypedArray[_, _], offset: Int, length: Int,
+      position: Unit, callback: CB[Int]): Int = js.native
```
sjrd: 4 spaces

linker/js/src/main/scala/org/scalajs/linker/irio/NodeFS.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    def readdir(path: String, opts: ReadDirOpt.type, cb: CB[js.Array[Dirent]]): Unit = js.native
```
sjrd: That line is a bit long. Consider wrapping before the `cb` argument.

linker/js/src/main/scala/org/scalajs/linker/irio/NodeFS.scala
sjrd: Also, consider typing `position` as a `js.UndefOr[Nothing]`, and then use `js.undefined` at the call site. It might be more readable. Or even give it its proper type, which is `js.UndefOr[Int]`, even though we don't use it with an `Int` argument in this particular code base.

gzm0:
Updated.

PR #3639, opened 2019-05-06
Fix cross-building plugin issue on Intellij IDEA

### LINKED ISSUE #3638: Can't resolve dependencies on Intellij IDEA 2019.1.1
When I import this repository on Intellij IDEA 2019.1.1 using Scala plugin for the IDE, sbt could not open the project due to following error.

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.portable-scala#sbt-platform-deps;1.0.0: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn] 	Note: Some unresolved dependencies have extra attributes.  Check that these dependencies exist with the requested attributes.
[warn] 		org.portable-scala:sbt-platform-deps:1.0.0 (scalaVersion=2.12, sbtVersion=0.13)
[warn]
[warn] 	Note: Unresolved dependencies path:
[warn] 		org.portable-scala:sbt-platform-deps:1.0.0 (scalaVersion=2.12, sbtVersion=0.13) ((sbt.BuildExtra) Defaults.scala#L2092)
[warn] 		  +- org.scala-js:sbt-scalajs:0.6.28-SNAPSHOT (scalaVersion=2.12, sbtVersion=0.13)
...
[error] (sbtPlugin/*:ssExtractDependencies) sbt.ResolveException: unresolved dependency: org.portable-scala#sbt-platform-deps;1.0.0: not found
[error] (sbtPlugin/*:update) sbt.ResolveException: unresolved dependency: org.portable-scala#sbt-platform-deps;1.0.0: not found
```

But when I use sbt directly on terminal, I can open the project and even can run `compile`, `test` or `fastOptJS` tasks.
* openjdk version "11.0.2" 2019-01-15
* sbt 1.2.8

Intellij IDEA gives better developer experience (code completion, instant compiler error, jump to definition and so on) compared to using sbt only, so I think supporting IDEA is benefical for contributors.

Any ideas to solve?

exoego:
Closes #3638.
Followed the [workaround for cross-building plugin](https://github.com/sbt/sbt/issues/3473#issuecomment-410184709)

As far as I tested, Intellij IDEA now be able to import sbt project.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -482,12 +482,12 @@ noToolsScalaVersions.each { scalaVersion ->
   quickMatrix.add([task: "test-suite-ecma-script6", scala: scalaVersion, java: mainJavaVersion, testSuite: "testSuite"])
 }
 allJavaVersions.each { javaVersion ->
-  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", java: javaVersion])
+  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.12.8", sbt_version_override: "", java: javaVersion])
 }
-quickMatrix.add([task: "tools-cli-stubs", scala: "2.12.8", java: mainJavaVersion])
+quickMatrix.add([task: "tools-cli-stubs", scala: "2.10.7", sbt_version_override: "0.13.17", java: mainJavaVersion])
 quickMatrix.add([task: "partestc", scala: "2.11.0", java: mainJavaVersion])
-quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.10.7", sbt_version_override: "", java: mainJavaVersion])
-quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.12.8", sbt_version_override: "1.0.0", java: mainJavaVersion])
+quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.10.7", sbt_version_override: "0.13.17", java: mainJavaVersion])
+quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.12.8", sbt_version_override: "", java: mainJavaVersion])
 
 // The 'full' matrix
 def fullMatrix = quickMatrix.clone()
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -819,6 +819,12 @@ object Build {
           normalizedName := "sbt-scalajs",
           bintrayProjectName := "sbt-scalajs-plugin", // "sbt-scalajs" was taken
           sbtPlugin := true,
+          sbtVersion in pluginCrossBuild := {
+            scalaVersion.value match {
+              case v if v.startsWith("2.10.") => "0.13.17"
+              case _ => "1.0.0"
+            }
+          },
           scalaBinaryVersion :=
             CrossVersion.binaryScalaVersion(scalaVersion.value),
           previousArtifactSetting,
--- a/sbt-plugin-test/project/build.properties
+++ b/sbt-plugin-test/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.17
+sbt.version=1.0.0
--- a/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala
+++ b/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala
@@ -1,3 +1,15 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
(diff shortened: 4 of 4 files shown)


exoego:
🤔Hmm, the 2nd last build was green, but the last (all comits squashed) was red.
image

Jenkinsfile
```diff
@@ -482,11 +482,11 @@ noToolsScalaVersions.each { scalaVersion ->
...
 }
 allJavaVersions.each { javaVersion ->
-  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", java: javaVersion])
+  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", sbt_version_override: "0.13.17", java: javaVersion])
 }
 quickMatrix.add([task: "tools-cli-stubs", scala: "2.12.8", java: mainJavaVersion])
 quickMatrix.add([task: "partestc", scala: "2.11.0", java: mainJavaVersion])
-quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.10.7", sbt_version_override: "", java: mainJavaVersion])
+quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.10.7", sbt_version_override: "0.13.17", java: mainJavaVersion])
 quickMatrix.add([task: "sbtplugin-test", toolsscala: "2.12.8", sbt_version_override: "1.0.0", java: mainJavaVersion])
```
sjrd: Could you try emptying the `sbt_version_override` for the 2.12.8 case, and change https://github.com/scala-js/scala-js/blob/0.6.x/sbt-plugin-test/project/build.properties to use 1.0.0?

Jenkinsfile
```diff
@@ -482,11 +482,11 @@ noToolsScalaVersions.each { scalaVersion ->
   quickMatrix.add([task: "test-suite-ecma-script6", scala: scalaVersion, java: mainJavaVersion, testSuite: "testSuite"])
 }
 allJavaVersions.each { javaVersion ->
-  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", java: javaVersion])
+  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", sbt_version_override: "0.13.17", java: javaVersion])
```
sjrd: To go with the change below, I'd rather not set `sbt_version_override` here, but change 2.10.7 into 2.12.8. And then change 2.12.8 into 2.10.7 on the line below (the one with `tools-cli-stubs`).

project/Build.scala
```diff
@@ -819,6 +819,12 @@ object Build {
           normalizedName := "sbt-scalajs",
           bintrayProjectName := "sbt-scalajs-plugin", // "sbt-scalajs" was taken
           sbtPlugin := true,
+          sbtVersion in pluginCrossBuild := {
+            scalaVersion.value match {
+              case v if v.startsWith("2.10") || v.startsWith("2.11") => "0.13.17"
```
sjrd: The 2.11 case makes no sense, since this project can never be built for 2.11. I suggest simply removing `|| v.startsWith("2.11")`.

Also we usually include an additional dot on such tests in this codebase, so please use `v.startsWith("2.10.")` instead of `v.startsWith("2.10")`.

exoego: Fixed

Jenkinsfile
exoego: Fixed

Jenkinsfile
exoego: Fixed

Jenkinsfile
```diff
@@ -482,12 +482,12 @@ noToolsScalaVersions.each { scalaVersion ->
   quickMatrix.add([task: "test-suite-ecma-script6", scala: scalaVersion, java: mainJavaVersion, testSuite: "testSuite"])
 }
 allJavaVersions.each { javaVersion ->
-  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.10.7", java: javaVersion])
+  quickMatrix.add([task: "tools-cli-stubs-sbtplugin", scala: "2.12.8", sbt_version_override: "", java: javaVersion])
 }
 quickMatrix.add([task: "tools-cli-stubs", scala: "2.12.8", java: mainJavaVersion])
```
sjrd: This line should now use `2.10.7`. Otherwise we don't test our tools/cli/stubs with 2.10.7 anymore (before it was tested by line 485)

exoego:
Looks like Jenkins is not triggered ???

exoego:
Understood.
Rebased into `0.6.x` and squashed commits with the commit message you suggested.

PR #3641, opened 2019-05-09
Do not rely on the implicit conversion option2Iterable.

sjrd:
`option2Iterable` might be deprecated in 2.13.0-RC2. See https://github.com/scala/scala/pull/8038.

The previous code was doing useless work anyway. We now completely avoid `treeCopy` when there are no exports for any given class.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
@@ -446,9 +446,12 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
           }
         }
 
-        val exports = exporters.get(clsSym).toIterable.flatten
-        // Add exports to the template
-        treeCopy.Template(tree, parents, self, body ++ exports)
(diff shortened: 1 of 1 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -446,9 +446,13 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
-        val exports = exporters.get(clsSym).toIterable.flatten
-        // Add exports to the template
-        treeCopy.Template(tree, parents, self, body ++ exports)
+        // Add exports to the template, if there are any
+        exporters.get(clsSym) match {
+          case None =>
+            tree // nothing to change
+          case Some(exports) =>
+            treeCopy.Template(tree, parents, self, body ::: exports.toList)
+        }
```
gzm0: Wouldn't you want to use a fold for this?

sjrd: Indeed. Updated.

PR #3645, opened 2019-05-12
Fix #3644: Forward cause messages for RPCCore exceptions

### LINKED ISSUE #3644: Report causes of failed testinterface RPC call better
Otherwise this can lead to very hard to debug problems.

gzm0:
This will help guide users better to what failed as the
message (hopefully) will make it clear that it is not the RPC
subsystem itself that failed but merely forwarded exception.

### DIFF
--- a/test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
+++ b/test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
@@ -262,10 +262,10 @@ private[scalajs] object RPCCore {
   type OpCode = Byte
 
   /** Exception thrown if a remote invocation fails. */
-  class RPCException(c: Throwable) extends Exception(null, c)
+  class RPCException(c: Throwable) extends Exception(c)
 
   /** Exception thrown if the channel got closed. */
-  class ClosedException(c: Throwable) extends Exception(null, c)
+  class ClosedException(c: Throwable) extends Exception(c)
 
   private val ReplyOK: Byte = 0.toByte
   private val ReplyErr: Byte = 1.toByte


test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -262,10 +262,10 @@ private[scalajs] object RPCCore {
   type OpCode = Byte
 
   /** Exception thrown if a remote invocation fails. */
-  class RPCException(c: Throwable) extends Exception(null, c)
+  class RPCException(c: Throwable) extends Exception(c.getMessage, c)
```
sjrd: Consider simply using `Exception(c)`, using the constructor [`Exception(Throwable cause)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html#Exception-java.lang.Throwable-) which will automatically incorporate the cause's `toString()` as the `RPCException`'s message.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
```diff
@@ -262,10 +262,10 @@ private[scalajs] object RPCCore {
...
 
   /** Exception thrown if the channel got closed. */
-  class ClosedException(c: Throwable) extends Exception(null, c)
+  class ClosedException(c: Throwable) extends Exception(c.getMessage, c)
```
sjrd: Same here.

test-common/src/main/scala/org/scalajs/testcommon/RPCCore.scala
gzm0: True, that's neater.

gzm0:
> Also I wonder why we even have those exception types? Why not just throw the original exceptions?

Mostly for better distinguishing what happens. It might be very hard to debug stuff when we don't know when stuff traverses an RPC boundary. Another reason is to record a stack trace on the client side as well.

PR #3646, opened 2019-05-12
Fix #3589: Use java.nio.file.Path instead of VirtualBinaryFile

### LINKED ISSUE #3589: Use java.nio.file.Path in JSEnv API

gzm0:


gzm0:
TODO after this (provided it gets accepted)

- [ ] add `JSEnv` test cases for default FS provider paths.
- [ ] get rid of io package.

io/jvm/src/main/scala/org/scalajs/io/FileVirtualFiles.scala
```diff
@@ -12,17 +12,12 @@
 
 package org.scalajs.io
 
-import java.io._
+import java.nio.files._
```
sjrd: I think this was supposed to be `java.nio.file._`.

sjrd:
`NodeJSEnv.scala` refers to `org.scalajs.io.JSUtils.escapeJS`, which it does not have access to anymore, given that the dependency on `io` was dropped. Should we expose `escapeJS` as a utility helper in the `js-envs` project instead of `io`? It seems we don't use it in the linker project anymore; only in JSEnvs in the TestAdapter.

project/Build.scala
```diff
@@ -776,7 +779,7 @@ object Build {
         baseDirectory.value.getParentFile / "test-common/src/main/scala",
       unmanagedSourceDirectories in Test +=
         baseDirectory.value.getParentFile / "test-common/src/test/scala"
-  ).dependsOn(jsEnvs)
+  ).dependsOn(io, jsEnvs)
```
sjrd: If we move `escapeJS` to `js-envs`, then this change might not be needed after all.

gzm0:
Updated: I have decided to leave build dependency changes to later when I clean up IO entirely.

gzm0:
A doc link issue. Updated.

sjrd:
Do you still intended to do

> add JSEnv test cases for default FS provider paths.

or is it good to merge for you?

gzm0:
This is good to merge for me. I'll add these two things later.

PR #3649, opened 2019-05-15
Support the new flat patmat ASTs of Scala 2.12.9.

sjrd:
See https://github.com/scala/scala/pull/8061 upstream.

It is now possible for jumps to the next-case label to be buried in arbitrary positions within a case body. We support that by enclosing a case body with a labeled block, in case at least one `return` to the label could not be caught upstream.

Doing so produces significantly worse code for cases that only have one test (and hence one `return`), which are by far the most common scenario. Therefore, we add an optimization to rewrite cases with a
single `return` of the case to the previous style of `If`. This in turn allows the optimization of the `matchEnd` label to kick in, and production of better code.

When there are more than one `return`, the new flat structure is beneficial for our IR as much as for scalac Trees, and therefore we keep it as is.

Only merge this PR if and when https://github.com/scala/scala/pull/8061 is accepted upstream.

Locally tested with:
```
> set resolvers in Global += "pr" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/"
> ++2.12.9-bin-e77555d-SNAPSHOT
> set scalacOptions in testSuite += "-Ypatmat-flat-ast"
> testSuite/test
```
/cc @retronym

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -144,6 +144,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
     private val fakeTailJumpParamRepl = new ScopedVar[(Symbol, Symbol)]
     private val enclosingLabelDefParams = new ScopedVar[Map[Symbol, List[Symbol]]]
     private val isModuleInitialized = new ScopedVar[VarBox[Boolean]]
+    private val countsOfReturnsToMatchCase = new ScopedVar[mutable.Map[Symbol, Int]]
     private val countsOfReturnsToMatchEnd = new ScopedVar[mutable.Map[Symbol, Int]]
     private val undefinedDefaultParams = new ScopedVar[mutable.Set[Symbol]]
 
@@ -178,6 +179,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
           fakeTailJumpParamRepl := null,
           enclosingLabelDefParams := null,
           isModuleInitialized := null,
+          countsOfReturnsToMatchCase := null,
           countsOfReturnsToMatchEnd := null,
           undefinedDefaultParams := null,
           mutableLocalVars := null,
@@ -1447,13 +1449,14 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
       val sym = dd.symbol
 
       withScopedVars(
-          currentMethodSym          := sym,
-          thisLocalVarIdent         := None,
-          fakeTailJumpParamRepl     := (NoSymbol, NoSymbol),
-          enclosingLabelDefParams   := Map.empty,
-          isModuleInitialized       := new VarBox(false),
+          currentMethodSym := sym,
+          thisLocalVarIdent := None,
+          fakeTailJumpParamRepl := (NoSymbol, NoSymbol),
+          enclosingLabelDefParams := Map.empty,
+          isModuleInitialized := new VarBox(false),
+          countsOfReturnsToMatchCase := mutable.Map.empty,
           countsOfReturnsToMatchEnd := mutable.Map.empty,
-          undefinedDefaultParams    := mutable.Set.empty
+          undefinedDefaultParams := mutable.Set.empty
       ) {
         assert(vparamss.isEmpty || vparamss.tail.isEmpty,
             "Malformed parameter list: " + vparamss)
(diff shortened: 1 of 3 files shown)


sjrd:
The diff of `scalajs-testsuite-fastopt.js` is almost a complete improvement. I found two small regressions by searching the regexp `^\+ *matchEnd.*:` in the diff, to find *new* `matchEnd` labels (i.e., `matchEnd` labels that were previously eliminated, but are not anymore). The two regressions seem to come from for consecutive tests whose conditions are not merged anymore, which is what I already suspected to be a problem in the JVM version of the PR, see my comment at https://github.com/scala/scala/pull/8061#pullrequestreview-237673869.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -2540,15 +2540,25 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+         *
+         * Such labels are not enclosing. Instead, they are forward jumps to a
+         * following case, or to the match-end. For those labels, we generate a
+         * js.Return and keep track of how many such returns we generate, so
+         * that the enclosing `genTranslatedMatch` can optimize away the
+         * labeled blocks in some cases, notably when they are not used at all.
+         *
+         * Next-case labels have no argument, while match-end labels have
+         * exactly one argument, which is the result of the pattern match (of
+         * type BoxedUnit if the match is in statement position). In the latter
```
gzm0: Just to check my understanding: We rely on the arguments to determine if this is a 0 or 1 argument label?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3189,13 +3199,16 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
-      countsOfReturnsToMatchEnd(matchEndSym) = 0
+      countsOfReturnsToPatmatLabel(matchEndSym) = 0
 
       val nextCaseSyms = (cases.tail map (_.symbol)) :+ NoSymbol
 
       val translatedCases = for {
         (LabelDef(_, Nil, rhs), nextCaseSym) <- cases zip nextCaseSyms
       } yield {
+        if (nextCaseSym.exists)
+          countsOfReturnsToPatmatLabel(nextCaseSym) = 0
```
gzm0: Is there a reason we are not using a different map for this? I think it would aid understanding quite a bit.

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3243,8 +3264,91 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+     *  The above rewrite is important for `genOptimizedMatchEndLabeled` below
+     *  to be able to do its job, which in turn is important for the IR
+     *  optimizer to perform a better analysis.
+     *
+     *  This whole thing is only necessary in Scala 2.12.9+, with the new flat
+     *  patmat ASTs. In previous versions, `returnCount` is always 0 because
+     *  all jumps to case labels are already caught upstream by `genCaseBody()`
+     *  inside `genTranslatedMatch()`.
+     */
+    private def genOptimizedCaseLabeled(label: js.Ident,
```
gzm0: Can we write an AST test for this?

compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
```diff
@@ -3243,8 +3264,91 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+                  case js.UnaryOp(js.UnaryOp.Boolean_!, notCond) =>
+                    notCond
+                  case _ =>
+                    js.UnaryOp(js.UnaryOp.Boolean_!, cond)
+                }
+                js.Block(stats1 :+ js.If(notCond, js.Block(stats2), js.Skip())(jstpe.NoType))
+
+              case _ =>
+                default
+            }
```
gzm0: I find reasoning about the correctness of this piece of code very difficult. Notably it's very hard to be sure that it is correct if there is a single `js.Return` but not in the shape we can optimize. So maybe add a `case Nil => default`, `case _ => fail(...)` to aid readability?

sjrd:
Updated.

sjrd:
Since the upstream PR has been approved, I'll merge this now so that Scala.js can be a little bit *ahead* on the community build, for once.

PR #3653, opened 2019-05-18
Misc minor refactoring for optimizations/readability 

exoego:


gzm0:
There seems to be an actual problem with some of the changes (`contains` on `Option` seems to not exist).

gzm0:
Other improvements look good to me (mostly due to readability not speed)

compiler/src/main/scala/org/scalajs/core/compiler/PrepJSInterop.scala
```diff
@@ -1332,7 +1332,7 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
 
       val firstArgIsValid = annot.stringArg(0).isDefined
       if (!firstArgIsValid) {
-        reporter.error(annot.args.head.pos,
+        reporter.error(annot.args(0).pos,
```
sjrd: This commit reverted the wrong change. It should have reverted the change in `OptimizerCore.scala`, not in `PrepJSInterop.scala`.

exoego: Oops. 
Fixed in https://github.com/scala-js/scala-js/pull/3653/commits/fa30b61aa1aa50090db5d9cc69319dd31e0237d8

exoego: Oops. 
Fixed in https://github.com/scala-js/scala-js/pull/3653/commits/e108cb42e9e360b6e352d24d167f258cbe1c1e8e

exoego: Oops. 
Fixed in https://github.com/scala-js/scala-js/pull/3653/commits/67dbe0df56927fa0251ea342165d9b82a1980056

exoego: Oops.
Fixed in https://github.com/scala-js/scala-js/pull/3653/commits/e43e0fb30cba80a23e8a996f535fe9376a4a7b5c

sjrd:
The code changes LGTM now. Could you please rebase on top of the latest 0.6.x, and squash everything in a single commit whose message is the title of this PR? I don't think there's any need to keep those small changes in individual commits.

exoego:
Thanks for reviewing.
Squashed into one commit.

PR #3654, opened 2019-05-19
Add a JSEnv test using the default filesystem

gzm0:
This is a follow-up from #3646.

### DIFF
--- a/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
+++ b/js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
@@ -12,6 +12,10 @@
 
 package org.scalajs.jsenv.test
 
+import java.io.File
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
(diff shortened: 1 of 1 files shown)


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/RunTests.scala
```diff
@@ -142,6 +146,21 @@ private[test] class RunTests(config: JSEnvSuiteConfig, withCom: Boolean) {
     }
   }
 
+  @Test
+  def defaultFilesystem: Unit = {
+    // Tests that a JSEnv works with files from the default filesystem.
+
+    val tmpFile = File.createTempFile("sjs-run-test-defaultfile", ".js")
+    tmpFile.deleteOnExit()
```
sjrd: Could we be more active in deleting this file, for example with a try/finally?

gzm0: Of course. Good point.

PR #3655, opened 2019-05-19
Remove IO artifact entirely

gzm0:
This is a follow-up from #3646.


sjrd:
The last commit probably needs to change something in `scripts/publish.sh`.

gzm0:
>The last commit probably needs to change something in scripts/publish.sh. 

It does remove `io` and `ioJS`. Is there something else that needs to be changed?

gzm0:
> I thought browsers would refuse to load assets with file URIs, and that relative URLs were basically necessary for this use case. 

You are referring to the [same-origin policy][1] it requires that things a page loads come from the same origin. So you cannot load a `file:` URL when loaded from and `http:` URL for example. `file` from `file` is allowed.

> Did you test with Firefox and Chrome, at least?

Yes. Works on both.

[1]: https://en.wikipedia.org/wiki/Same-origin_policy

PR #3663, opened 2019-05-22
Fix #3659: Emit ECMAScript 2015 code by default.

### LINKED ISSUE #3659: Make ECMAScript2015 the default in Scala.js 1.x
There are real advantages to using ECMAScript 2015 now:

* `Throwable`s, by virtue of properly extending JavaScript's `Error`, have an `[[ErrorData]]` internal slot, and therefore receive proper debugging info in JS engines, allowing better display of stack traces and error messages in interactive debuggers.
* Static fields and methods in JS classes are properly inherited. See https://github.com/scala-js/scala-js/issues/2771.
* Overall, Scala.js is *specified* in terms of ES2015. The ES 5.1 output that we emit is already, conceptually, a desugaring of the ES2015 *meaning* of Scala.js programs.

For those reasons, I think we should make it the default in Scala.js 1.x.

Reverting any particular build to emitting ES 5.1 would be as easy as using the following sbt setting:
```scala
scalaJSLinkerConfig in ThisBuild ~= { _.withESFeatures(_.withUseECMAScript2015(false)) }
```

Any objection?

sjrd:


sjrd:
> LGTM, but maybe wait a couple of days for people to actually object on the bug if they have concerns?

Sure.

sjrd:
Added an upgrade of source-map-support to 0.5.7, which is necessary for source maps to correctly work in ES modules. Later versions break `StackTraceTest`, which appears to be due to https://github.com/evanw/node-source-map-support/pull/220 upstream (the only meaningful commit between 0.5.7 and 0.5.8): they start using the original-name information from source maps to give the name to the function. Apparently, our original names are not good enough, or downright wrong. I'll leave fixing that for another day/PR, since it has nothing to do with the present PR.

gzm0:
Shall we merge this?

sjrd:
Yes, ship it!

PR #3671, opened 2019-05-27
Fix #3668: Split the test bridge into a separate artifact.

### LINKED ISSUE #3668: Split org.scalajs.testing.* from test-interface in a separate artifact from sbt.testing.*
The testing subsystem contains the following components:

* `org.scalajs.testing.adapter`: a JVM component implementing `sbt.testing` on the JVM side, by connecting through a protocol to the JS side
* `org.scalajs.testing.interface`: a JS component *using* `sbt.testing` on the JS side, connecting through the same protocol to the JVM side
* `org.scalajs.testing.common`: cross-compiling code that is shared by the two above components
* `sbt.testing`: a JS clone of the standard JVM sbt-testing-interface

Currently, `org.scalajs.testing.interface` and `sbt.testing` live in the same artifact, `scalajs-test-interface`. This artifact plays two very different roles at once:
* the `sbt.testing` part is implemented by concrete testing frameworks, which must therefore depend on it
* the `testing.interface` part needs to be on the classpath when linking tests, because the test adapter relies on its implementation of the protocol.

The protocol of `testing.interface` and `testing.adapter` must be in sync, at all times. In practice, a version *x* of the sbt plugin forces the test adapter version to be *== x*, and the interface version to be *>= x*. That means that we have to deal with version mismatches when defining and implementing the protocol: the interface can be newer than the test adapter. This is a significant problem when we try to fix issues with the protocol itself, such as recently found in #3667.

We can however get rid of this problem for good by splitting `testing.interface` and `sbt.testing` in two different artifacts:
* `sbt.testing` is an independent artifact, depending only on scalajs-library. Concrete testing frameworks depend on that artifact to implement `sbt.testing.Framework`.
* `org.scalajs.testing.interface` (and `testcommon`) are another artifact, that depends on the one for `sbt.testing`

The latter artifact (let's call it `scalajs-testing-interface-impl`) is not on the dependency graph of concrete testing frameworks. Therefore the sbt plugin can always force it to be of version *== x*. With this system, we guarantee that the JVM and JS side of the implementation of the protocol are always exactly equal, and that means we can evolve the protocol itself completely freely, with breaking changes in any directions.

sjrd:
The scalajs-test-interface artifact actually contained two independent concerns:

* The Scala.js version of the `sbt.testing.*` APIs, and
* The JS-side implementation of the test adapter/bridge protocol.

Concrete testing frameworks depend on the former, but need not know about the latter. Moreover, the protocol in the bridge needs to be in sync with the one in the adapter.

Keeping those two concerns together meant that transitive dependencies of testing frameworks could force a version of the bridge that was newer than the adapter. This possibility forced us to lock down the protocol in very serious ways.

This commit splits those two concerns in two different artifacts, so that we can evolve the protocol more freely.

For compatibility, `ScalaJSClassLoader` and `TestUtils` are kept in the scalajs-test-interface artifact. They are both already deprecated, so they will be gone in Scala.js 1.x. The classes that were moved to scalajs-test-bridge are package-private, so there is no issue for them.


project/Build.scala
```diff
@@ -1279,7 +1279,23 @@ object Build {
...
+  lazy val testBridge = Project(
+      id = "testBridge",
+      base = file("test-bridge"),
+      settings = (
+          commonSettings ++ publishSettings ++ myScalaJSSettings ++ fatalWarningsSettings
+      ) ++ Seq(
+          name := "Scala.js test bridge",
+          delambdafySetting,
+          /* By design, the test-bridge has a completely private API (it is
+           * only loaded through a privately-known module initializer entry
```
gzm0: This is only true in 1.0. In 0.6.x this is done through exports IIUC.

gzm0:
> Do we also need to rename the package? It seems wrong to have the same package in different (non-cross) artifacts.

I just realize that `testinterface` would only be present in `test-bridge` modulo the two deprecated things which is probably acceptable. And the artifact containing `sbt.testing` should be called `test-interface` to mirror the JVM name.

So maybe we should simply rename the `org.sjs.testing.interface` package to `org.sjs.testing.bridge` in the master branch and leave this as is? (modulo the comment update)?

sjrd:
> So maybe we should simply rename the org.sjs.testing.interface package to org.sjs.testing.bridge in the master branch and leave this as is? (modulo the comment update)?

Yes, I think that's a good path forward. I've updated the comment.

PR #3674, opened 2019-05-28
Rename `org.scalajs.testing.interface` to `...testing.bridge`.

sjrd:
Follow up to 2240515b60390591fba83be88fab67a31b25adf7.


sjrd:
Good point. I have added a commit for that. (and I have tested `testSuite/testHtml` and the produced .html file)

PR #3675, opened 2019-05-28
Fix #3656: Separate irio into super-stable/stable API

### LINKED ISSUE #3656: Determine compatibility guarantees and naming of irio package
Ideally, we would have a way to make changes to some of the lower level APIs (notably `WritableVirtualBinaryFile`).

However, some other parts definitely need to be super-stable, as they are required to link (and mentioned in the other parts of the super stable API).

gzm0:


sjrd:
So IIUC, after this PR:

* Things in `linker.*` are super-stable
* Things in `linker.irio.*` are stable

Are the latter as stable as `linker.standard.*`? If yes, should we move them to `standard` and get rid of `irio`?

If not, perhaps we should revisit the name `irio` anyway, because the package contains IO for things that are not IR-related, e.g., `OutputFile`s.

linker/shared/src/main/scala/org/scalajs/linker/irio/IRFileCacheImpl.scala
```diff
@@ -58,19 +58,22 @@ final class IRFileCacheImpl extends IRFileCache {
...
         implicit ec: ExecutionContext): Unit = clearOnThrow {
       val result = Seq.newBuilder[PersistedFiles]
 
-      for (file <- files) {
+      for (stableFile <- files) {
+        val file = IRContainerImpl.fromIRContainer(stableFile)
+
         @tailrec
         def putContents(): PersistedFiles = {
+
```
sjrd: Spurious blank line.

linker/shared/src/main/scala/org/scalajs/linker/irio/MemOutputFileImpl.scala
```diff
@@ -18,22 +18,26 @@ import java.io._
...
+import org.scalajs.linker.LinkerOutput
+
+private[linker] final class MemOutputFileImpl extends LinkerOutput.MemFile {
   private var _content: Array[Byte] = _
 
   def content: Array[Byte] = _content
 
-  def newChannel()(implicit ec: ExecutionContext): Future[WriteChannel] =
-    Future.successful(new Channel)
+  private[linker] object impl extends OutputFile {
```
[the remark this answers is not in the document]
gzm0: Then we would expose a less stable type in the stable interface. Because `OutputFile` has all the accessor methods. So this design indeed does not allow to define a type that:
- implements `LinkerOutput.File`
- provides additional API.
- does not expose its implementation.

This is fixable, but probably requires even more types :)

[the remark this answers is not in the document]
gzm0: Seems the only way to fix this is a sealed trait. I'll do that.

gzm0:
Updated. I believe I have addressed all comments.

linker/shared/src/main/scala/org/scalajs/linker/standard/IRFileImpl.scala
```diff
@@ -10,62 +10,27 @@
...
-    shortClassName + "(" + path + ")"
-  }
-}
+import org.scalajs.linker.IRFile
 
 /** A virtual Scala.js IR file.
  *  It contains the class info and the IR tree.
  */
-trait VirtualScalaJSIRFile {
+abstract class IRFileImpl (
```
sjrd: No space before `(` here. We only use a space if there is a visibility modifier for the constructor.

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardIRFileCache.scala
```diff
@@ -57,19 +56,22 @@ final class StandardIRFileCache extends IRFileCache {
...
         implicit ec: ExecutionContext): Unit = clearOnThrow {
       val result = Seq.newBuilder[PersistedFiles]
 
-      for (file <- files) {
+      for (stableFile <- files) {
+        val file = IRContainerImpl.fromIRContainer(stableFile)
+
         @tailrec
         def putContents(): PersistedFiles = {
+
```
sjrd: Spurious blank line.

linker/jvm/src/main/scala/org/scalajs/linker/IRContainerPlatformExtensions.scala
```diff
@@ -0,0 +1,105 @@
...
+          throw new IllegalArgumentException("Illegal classpath entry " + entry)
+        }
+      }
+    }
+
+    (containers.result(), paths.result())
+  }
+
+  private final class JarIRContainer(path: Path, lastModified: FileTime)
+    extends IRContainerImpl(path.toString, Some(lastModified.toString)) {
```
sjrd: Oh there should be 4 spaces here.

gzm0:
updated

PR #3679, opened 2019-05-31
Replace Scala 2.13.0-RC2 by 2.13.0-RC3.

sjrd:
The changes in the overrides had to happen in lockstep, because there were incompatibilities in both directions. Yes, that means that when publishing 0.6.28 and 1.0.0-M8 for RC3, I had to have those changes locally, as dirty files.


gzm0:
> Yes, that means that when publishing 0.6.28 and 1.0.0-M8 for RC3, I had to have those changes locally, as dirty files.

I think it is a very bad idea to do this. Now this means that there are irreproducible artifacts published to 
Maven. That is not good.

sjrd:
I was very reluctant to do it that way, but I didn't see any other solution that would have put me under intense workload and undue stress for days. So at some point I had to make a call on favor of my well-being.

gzm0:
It is good to hear that you favor your well being above the software we write. We all probably far too often do not do that.

That being said, wouldn't not doing anything also have been an option? (Please excuse me if the implicit assumption that this would have had a similarly good effect on your well-being is wrong).

sjrd:
> That being said, wouldn't not doing anything also have been an option?

By not doing anything, you mean, not publishing 0.6.28 and 1.0.0-M8 for 2.13.0-RC3? That would have blocked the entire Scala ecosystem. See https://github.com/scala/scala-dev/issues/629: as long as Scala.js is not published for a new version of Scala, nothing in the ecosystem can be built and published. It is unfortunately not an option anymore; Scala.js has become way too critical to the Scala ecosystem.

gzm0:
Thank you for bringing this bug to my attention. As you know, it is not the first time this problem crops up (example). I have a problem with the current situation because it forces the Scala.js maintainers (i.e. you and me) to make a choice between the following:

1. Compromise on our well-being
1. Compromise on engineering principles
1. Block the entire Scala ecosystem

So far (examples below) we ended up choosing 2 most of the time, but every time this happens, I'm not happy and a bit of 1 happens as well, at least for me. I cannot judge to what extent 3 happens.

Examples where we have compromised on engineering principles due to this pressure:

- https://github.com/scala-js/scala-js/pull/3434#issuecomment-417374330
- https://github.com/scala-js/scala-js/pull/3361#issuecomment-389120406
- https://github.com/scala-js/scala-js/pull/3605#issue-267349253 / scala-js/scala-js#3606
- https://github.com/scala-js/scala-js/pull/3679#issuecomment-498070654 (this PR).
- https://github.com/scala-js/scala-js/commit/8590070e782718f500189acd7d781b5a43e157fe#commitcomment-33046924

Example where I have compromised my well being (through review speed):

- scala-js/scala-js#3625
- scala-js/scala-js#3641

I will try and follow up on scala/scala-dev#629 hoping to find a resolution to this, even though it is not exactly the thing I'd like to spend my own free time on.

PR #3681, opened 2019-06-03
Add java.lang.Byte.decode

exoego:
[`Byte.decode`](https://docs.oracle.com/javase/8/docs/api/java/lang/Byte.html#decode-java.lang.String-) is useful since it can accept hexadecimal and octal numbers.

Test cases (not implementation) are [drawn from OpenJDK](https://github.com/openjdk/jdk/blob/6ef1da93ed67bb839785475335d494af19ed44d4/test/jdk/java/lang/Byte/Decode.java#L62).

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Byte.scala
+++ b/javalanglib/src/main/scala/java/lang/Byte.scala
@@ -81,4 +81,15 @@ object Byte {
 
   @inline def compare(x: scala.Byte, y: scala.Byte): scala.Int =
     x - y
+
+  def decode(s: String): Byte = {
+    val lowerCase = s.toLowerCase
+    val radix = if (lowerCase.contains("0x") || lowerCase.contains("#"))
+      16
+    else if (s.startsWith("0") || s.startsWith("+0") || s.startsWith("-0"))
+      8
+    else
+      10
+    parseByte(lowerCase.replaceAll("[x#]", ""), radix)
+  }
 }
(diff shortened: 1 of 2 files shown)


sjrd:
What's the license of tests in OpenJDK? I'm pretty sure they're under a special thing governed by the Java Test Kit license (or something like that), which is a commercial thing. So we can't use them either.

exoego:
[It's GPL](https://github.com/openjdk/jdk/blob/6ef1da93ed67bb839785475335d494af19ed44d4/LICENSE), not commercial things.

sjrd:
> [It's GPL](https://github.com/openjdk/jdk/blob/6ef1da93ed67bb839785475335d494af19ed44d4/LICENSE), not commercial things.

Arg. I'm sorry, but we really cannot accept this PR. GPL is very, very incompatible with our Apache License, and that is true for the tests just as much as for the implementation. There is no difference: it is one codebase.

I'm afraid I'll have to close this PR without looking at it. And by working with the GPL'ed code for the tests, I'm afraid you've also compromised yourself from implementing that method and/or its tests in a different way, because that wouldn't be a clean room anymore. I know it's very annoying and it can seem drastic, but in the light of the Google v Oracle case, we have to be extremely careful about this licensing stuff.

We can reimplement `Byte.decode` ourselves if it is needed. What prompted you to implemented?

exoego:
Understood.

Motivation of adding `decode`: I want to decode hexadecimal string like `"0x32"` written in some text.
To my knowledge, there is no function for that in Scala's standard lib, but in Java lib.

I will revisit and may draw some tests from GWT project since they have JRE emulation and [they uses Apache License](http://www.gwtproject.org/terms.html)

PR #3683, opened 2019-06-03
[javalib] Add Byte|Integer|Long.decode

exoego:
Revisit of #3681.

This is a cleanroom implementation of [`Byte.decode`](https://docs.oracle.com/javase/8/docs/api/java/lang/Byte.html#decode-java.lang.String-), [`Integer.decode`](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#decode-java.lang.String-), and [`Long.decode`](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#decode-java.lang.String-).
Those are convenient for decoding octal string (`0755`) or hexadecimal string (`0xFFFF`).

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Byte.scala
+++ b/javalanglib/src/main/scala/java/lang/Byte.scala
@@ -81,4 +81,12 @@ object Byte {
 
   @inline def compare(x: scala.Byte, y: scala.Byte): scala.Int =
     x - y
+
+  def decode(s: String): Byte = {
+    val r = Integer.decode(s)
+    if (r < MIN_VALUE || r > MAX_VALUE)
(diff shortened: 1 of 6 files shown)


gzm0:
@sjrd is it ok to look at this or must we ban @exoego from building stuff he has looked at in OpenJDK?

sjrd:
TBH I don't know. :-s If the implementation had been looked at, it would be a definite ban for that area, yes. But even with the tests, the problem is that looking at the OpenJDK tests reveals something about the implementation, in particular about corner cases, and it's got to have some impact on how the implementation here was written.

I'd rather err on the side of caution, and close this. Those APIs and their tests will have to be implemented by someone else, I think.

I'm sorry @exoego. It's just too dangerous.

PR #3687, opened 2019-06-05
Throw UnsupportedEncodingException on bad charset.

exoego:
[According to spec (javadoc)](https://docs.oracle.com/javase/8/docs/api/java/io/OutputStreamWriter.html#OutputStreamWriter-java.io.OutputStream-java.lang.String-), `OutputStreamWriter(OutputStream, String)` should throw `java.io.UnsupportedEncodingException` if the named encoding is not supported.
But current implementation throws `java.nio.charset.UnsupportedCharsetException` instead.

### DIFF
--- a/javalib/src/main/scala/java/io/OutputStreamWriter.scala
+++ b/javalib/src/main/scala/java/io/OutputStreamWriter.scala
@@ -44,8 +44,14 @@ class OutputStreamWriter(private[this] var out: OutputStream,
   def this(out: OutputStream) =
     this(out, Charset.defaultCharset)
 
-  def this(out: OutputStream, charsetName: String) =
-    this(out, Charset.forName(charsetName))
+  def this(out: OutputStream, charsetName: String) = {
+    this(out, try {
+      Charset.forName(charsetName)
+    } catch {
+      case _: UnsupportedCharsetException =>
+        throw new UnsupportedEncodingException(charsetName)
+    })
+  }
 
   def getEncoding(): String =
     if (closed) null else enc.charset.name
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/OutputStreamWriterTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/OutputStreamWriterTest.scala
@@ -134,4 +134,11 @@ class OutputStreamWriterTest {
     testW({ osw => osw.write("ab\ud83d"); osw.close() },
         Array('a', 'b', '?'), alreadyFlushed = true)
   }
+
+  @Test def constructor_throw_UnsupportedEncodingException_if_unsupported_charset_name_given(): Unit = {
+    val ex = expectThrows(classOf[UnsupportedEncodingException],
+      new OutputStreamWriter(new ByteArrayOutputStream(), "UNSUPPORTED-CHARSET"))
+    assertTrue("Cause should be null since constructor does not accept cause",
+      ex.getCause == null)
+  }
 }


sjrd:
I will look at this (and other PRs) after Scala Days is over.

exoego:
Enjoy Scala Days 👍

javalib/src/main/scala/java/io/OutputStreamWriter.scala
```diff
@@ -45,7 +45,11 @@ class OutputStreamWriter(private[this] var out: OutputStream,
     this(out, Charset.defaultCharset)
 
   def this(out: OutputStream, charsetName: String) =
-    this(out, Charset.forName(charsetName))
+    this(out, try {
+      Charset.forName(charsetName)
+    } catch {
+      case _: UnsupportedCharsetException => throw new UnsupportedEncodingException(charsetName)
+    })
```
gzm0: Please but [braces around multi line methods][1]

[1]: https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#braces

javalib/src/main/scala/java/io/OutputStreamWriter.scala
```diff
@@ -45,7 +45,11 @@ class OutputStreamWriter(private[this] var out: OutputStream,
     this(out, Charset.defaultCharset)
 
   def this(out: OutputStream, charsetName: String) =
-    this(out, Charset.forName(charsetName))
+    this(out, try {
+      Charset.forName(charsetName)
+    } catch {
+      case _: UnsupportedCharsetException => throw new UnsupportedEncodingException(charsetName)
```
gzm0: Should we link the original exception as cause? (please reproduce JVM behavior if not specified, maybe even test it?)

javalib/src/main/scala/java/io/OutputStreamWriter.scala
```diff
@@ -45,7 +45,11 @@ class OutputStreamWriter(private[this] var out: OutputStream,
     this(out, Charset.defaultCharset)
 
   def this(out: OutputStream, charsetName: String) =
-    this(out, Charset.forName(charsetName))
+    this(out, try {
```
gzm0: consider putting the try as a separate statement above:

```scala
val charset = try {
 ...
} catch {
  ...
}
```

nested like this is a little funky IMHO (@sjrd has the last word of course).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/OutputStreamWriterTest.scala
```diff
@@ -134,4 +134,9 @@ class OutputStreamWriterTest {
     testW({ osw => osw.write("ab\ud83d"); osw.close() },
         Array('a', 'b', '?'), alreadyFlushed = true)
   }
+
+  @Test def constructor_throw_UnsupportedEncodingException_if_unsupported_charset_name_given(): Unit = {
+    expectThrows(classOf[UnsupportedEncodingException],
+      new OutputStreamWriter(new ByteArrayOutputStream(), "UNSUPPOTED-CHARSET"))
```
gzm0: UNSUPPO**R**TED? Or intended?

javalib/src/main/scala/java/io/OutputStreamWriter.scala
sjrd: Unfortunately in this case there's no choice, because Scala won't accept anything before the `this()` constructor call. So the way @exoego did it here is correct. (However, yes, in general, it would be better with a `val`.)

exoego: As explained by @sjrd, this is a trick to convince Scala compiler.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/io/OutputStreamWriterTest.scala
exoego: Fixed typo in https://github.com/scala-js/scala-js/pull/3687/commits/a1d0dce3e24e6580e8dfdd4b6854d446d47ed736

javalib/src/main/scala/java/io/OutputStreamWriter.scala
exoego: We cannot (therefore should not) link the original exception as cause, since [`UnsupportedEncodingException` does not have constructors to accept cause](https://docs.oracle.com/javase/8/docs/api/java/io/UnsupportedEncodingException.html).

Added tests to verify cause is `null` in https://github.com/scala-js/scala-js/pull/3687/commits/a1d0dce3e24e6580e8dfdd4b6854d446d47ed736

javalib/src/main/scala/java/io/OutputStreamWriter.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/3687/commits/cd3cb270eb7aae92d9517de0aec034f7782a6211

exoego:
Squashed commits and included the detail in the commit message.

javalib/src/main/scala/java/io/OutputStreamWriter.scala
```diff
@@ -45,7 +45,10 @@ class OutputStreamWriter(private[this] var out: OutputStream,
     this(out, Charset.defaultCharset)
 
   def this(out: OutputStream, charsetName: String) =
-    this(out, Charset.forName(charsetName))
+    this(out, try Charset.forName(charsetName)
+    catch {
```
sjrd: The formatting here is really weird. I suggest formatting that method as:
```scala
  def this(out: OutputStream, charsetName: String) = {
    this(out, try {
      Charset.forName(charsetName)
    } catch {
      case _: UnsupportedCharsetException =>
        throw new UnsupportedEncodingException(charsetName)
    })
  }
```

exoego: Fixed in https://github.com/scala-js/scala-js/commit/0e5d092bae45debcfaf5e0229e7d3510281cab39

javalib/src/main/scala/java/io/OutputStreamWriter.scala
gzm0: The method def still needs braces (see @sjrd's example)

sjrd:
(the CI failure is spurious; I have scheduled a re-run)

PR #3688, opened 2019-06-06
Fix #3686: Add `js.import(specifier)` to support ES dynamic imports.

### LINKED ISSUE #3686: Add support for ES dynamic `import()`
The ECMAScript proposal for dynamic imports at https://github.com/tc39/proposal-dynamic-import is [about to reach stage 4](https://github.com/tc39/ecma262/pull/1482) (supposedly it has reached stage 4 during the last TC meeting, but the PR hasn't been merged as of this writing). That means it's time for us to support it.

I would simply suggest a new primitive
```scala
def `import`(specifier: String): js.Primitive[scala.Any]
```
in the package object `js`, so that calls would look like
```scala
js.`import`("foobar.js")
```
It is a bit awkward that Scala imposes backticks to call that method, but I believe it's better than giving it a different name. In the future we might lobby for `import` not to be considered a keyword in Scala if it is followed by an open parenthesis.

---

Note that there is also a Stage 3 proposal for so-called `import.meta`, see https://github.com/tc39/proposal-import-meta. We might want to consider instead a
```scala
object `import` {
  def apply(specifier: String): js.Promise[scala.Any]
}
```
where `apply` is the primitive, so that we can later add a `def meta` in that object. If it's a `def import`, we might have trouble adding a "field" `meta` to it. Although I think Scala would not prevent us from having
```scala
def `import`(specifier: String): js.Promise[scala.Any]
// *and*
def `import`: ImportMeta
sealed trait ImportMeta {
  def meta: ...
}
```
but I never liked the overloading based on whether or not there is an *additional* parameter list :s

---

Another question is the result type of `js.import`: should it be:
* `js.Promise[scala.Any]`?
* `js.Promise[js.Any]`?
* `js.Promise[js.Dynamic]`?
* `js.Promise[js.ModuleInstance]` where `ModuleInstance` is *kind of* `js.Dynamic` but with less stuff (only read-only fields and methods)
* Or even parameterized with `[A <: js.Any]` and returning `js.Promise[A]`?

`scala.Any` is the most conservative one. `A` is the most liberal, allowing to avoid `asInstanceOf`s (but hiding the cast inside the method, in a sense, although it always succeed because `A <: js.Any` anyway).

sjrd:
Tests are rudimentary, because full testing requires asynchronous tests, which are not available in 0.6.x. Full tests should be added in the 1.x series.

The changes in the Closure linker backend are useless at the moment, because dynamic imports are only supported by the parser and pretty-printer of GCC but not the rest of the compiler. Support should eventually come in a later version of GCC, though, at which point it will transfer for free to our GCC backend.

I locally tested the GCC support with a modified version of closure-compiler, see branch at https://github.com/sjrd/closure-compiler/compare/master...sjrd:dynamic-import

#3689 contains further tests on top of the master branch.

### DIFF
--- a/assets/additional-doc-styles.css
+++ b/assets/additional-doc-styles.css
@@ -2,6 +2,10 @@
   background-color: #E68A00;
 }
 
+.badge-ecma2019 {
+  background-color: #E68A00;
+}
+
 .badge-non-std {
   background-color: #B94A48;
 }
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala
@@ -4431,6 +4431,10 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
               genAsInstanceOf(js.JSUnaryOp(js.JSUnaryOp.typeof, arg),
                   StringClass.tpe)
 
+            case JS_IMPORT =>
+              // js.import(arg)
+              js.JSImportCall(arg)
+
             case OBJPROPS =>
               // js.Object.properties(arg)
               genApplyMethod(
--- a/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/JSDefinitions.scala
@@ -118,6 +118,10 @@ trait JSDefinitions {
     lazy val RawJSTypeAnnot = getRequiredClass("scala.scalajs.js.annotation.RawJSType")
     lazy val ExposedJSMemberAnnot = getRequiredClass("scala.scalajs.js.annotation.ExposedJSMember")
 
+    lazy val JSImportModule = getRequiredModule("scala.scalajs.js.import")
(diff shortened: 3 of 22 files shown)


test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/DynamicImportTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  @Test def testDynImportParsesAndExecutes(): Unit = {
+    /* Since we do not have support for asynchronous tests, all we can do here
+     * is test that `js.import` parses, links, and executes without error,
+     * returning a `Promise`. We can't actually wait for the promise to
+     * resolve and hence cannot test its result.
+     *
+     * We will be able to revisit this in Scala.js 1.x.
+     */
+
+    assertEquals("[object Promise]", js.`import`[js.Any]("fs").toString())
```
gzm0: Why test the string representation and not that it is an instance of `js.Promise`?

sjrd: Because I had not thought about it :man_facepalming: 
Updated.

PR #3690, opened 2019-06-07
Various fixes in java.net.URI

exoego:
Noticed these when using `URI` and then running tests from [Apache Harmony](https://github.com/apache/harmony) which is licensed under Apache License 2.0.

Changes
1. IPv4 should not accept the digits greater than 255 (e.g. `256.384.512.777`)
2.  Normalize host in URI.hashCode.
    * This is necessary so that `foo.com` and `FoO.CoM` (which compare equal) produce the same hashCode.
3. Properly compare URIs.
    1. Apply case insensitivity when comparing URI parts
    2. Compare paths, query and fragments when comparing URI parts. This is necessary so that `foo.com/foo` and `FoO.CoM/foo` equals, but `foo.com/foo` and `FoO.CoM/FoO` does not.
    3. Return 0 if both scheme is undefined

### DIFF
--- a/javalib/src/main/scala/java/net/URI.scala
+++ b/javalib/src/main/scala/java/net/URI.scala
@@ -54,7 +54,7 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
   private val _authority = fld(AbsAuthority, RelAuthority).filter(_ != "")
   private val _userInfo = fld(AbsUserInfo, RelUserInfo)
   private val _host = fld(AbsHost, RelHost)
-  private val _port = fld(AbsPort, RelPort).fold(-1)(_.toInt)
+  private val _port = fld(AbsPort, RelPort).map(_.toInt)
 
   private val _path = {
     val useNetPath = fld(AbsAuthority, RelAuthority).isDefined
@@ -101,41 +101,56 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
    */
   @inline
   private def internalCompare(that: URI)(cmp: (String, String) => Int): Int = {
-    @inline def cmpOpt(x: js.UndefOr[String], y: js.UndefOr[String]): Int = {
+    @inline
+    def cmpOpt[T](x: js.UndefOr[T], y: js.UndefOr[T])(comparator: (T, T) => Int): Int = {
       if (x == y) 0
       // Undefined components are considered less than defined components
-      else x.fold(-1)(s1 => y.fold(1)(s2 => cmp(s1, s2)))
+      else x.fold(-1)(s1 => y.fold(1)(s2 => comparator(s1, s2)))
+    }
+    def comparePathQueryFragement(): Int = {
+      val cmpPath = cmpOpt(this._path, that._path)(cmp)
+      if (cmpPath != 0) {
+        cmpPath
+      } else {
+        val cmpQuery = cmpOpt(this._query, that._query)(cmp)
+        if (cmpQuery != 0) cmpQuery
+        else cmpOpt(this._fragment, that._fragment)(cmp)
+      }
     }
 
-    if (this._scheme != that._scheme)
-      this._scheme.fold(-1)(s1 => that._scheme.fold(1)(s1.compareToIgnoreCase))
-    else if (this._isOpaque != that._isOpaque)
-      // A hierarchical URI is less than an opaque URI
-      if (this._isOpaque) 1 else -1
-    else if (_isOpaque) {
-      val ssp = cmp(this._schemeSpecificPart, that._schemeSpecificPart)
-      if (ssp != 0) ssp
-      else cmpOpt(this._fragment, that._fragment)
-    } else if (this._authority != that._authority) {
-      if (this._host.isDefined && that._host.isDefined) {
-        val ui = cmpOpt(this._userInfo, that._userInfo)
-        if (ui != 0) ui
-        else {
-          val hst = this._host.get.compareToIgnoreCase(that._host.get)
-          if (hst != 0) hst
-          else if (this._port == that._port) 0
-          else if (this._port == -1) -1
-          else if (that._port == -1)  1
-          else this._port - that._port
+    val cmpScheme = cmpOpt(this._scheme, that._scheme)(_.compareToIgnoreCase(_))
+    if (cmpScheme != 0) {
+      cmpScheme
+    } else {
+      val cmpIsOpaque = this.isOpaque.compareTo(that.isOpaque) // A hierarchical URI is less than an opaque URI
+      if (cmpIsOpaque != 0) {
+        cmpIsOpaque
+      } else {
+        if (this.isOpaque()) {
+          val cmpSchemeSpecificPart = cmp(this._schemeSpecificPart, that._schemeSpecificPart)
+          if (cmpSchemeSpecificPart != 0) cmpSchemeSpecificPart
+          else comparePathQueryFragement()
+        } else if (this._host.isDefined && that._host.isDefined) {
+          val cmpUserInfo = cmpOpt(this._userInfo, that._userInfo)(cmp)
+          if (cmpUserInfo != 0) {
+            cmpUserInfo
+          } else {
+            val cmpHost = cmpOpt(this._host, that._host)(_.compareToIgnoreCase(_))
+            if (cmpHost != 0) {
+              cmpHost
+            } else {
+              val cmpPort = cmpOpt(this._port, that._port)(_ - _)
+              if (cmpPort != 0) cmpPort
+              else comparePathQueryFragement()
+            }
+          }
+        } else {
+          val cmpAuthority = cmpOpt(this._authority, that._authority)(cmp)
+          if (cmpAuthority != 0) cmpAuthority
+          else comparePathQueryFragement()
         }
-      } else
-        cmpOpt(this._authority, that._authority)
-    } else if (this._path != that._path)
-      cmpOpt(this._path, that._path)
-    else if (this._query != that._query)
-      cmpOpt(this._query, that._query)
-    else
-      cmpOpt(this._fragment, that._fragment)
+      }
+    }
   }
 
   def compareTo(that: URI): Int = internalCompare(that)(_.compareTo(_))
@@ -149,7 +164,7 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
   def getFragment(): String = _fragment.map(decodeComponent).orNull
   def getHost(): String = _host.orNull
   def getPath(): String = _path.map(decodeComponent).orNull
-  def getPort(): Int = _port
+  def getPort(): Int = _port.getOrElse(-1)
   def getQuery(): String = _query.map(decodeComponent).orNull
   def getRawAuthority(): String = _authority.orNull
   def getRawFragment(): String = _fragment.orNull
@@ -166,10 +181,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
     import URI.normalizeEscapes
 
     var acc = URI.uriSeed
-    acc = mix(acc, _scheme.##) // scheme may not contain escapes
-    acc = mix(acc, normalizeEscapes(_schemeSpecificPart).##)
-    acc = mixLast(acc, _fragment.map(normalizeEscapes).##)
-
+    acc = mix(acc, _scheme.map(_.toLowerCase).##) // scheme may not contain escapes
+    if (this.isOpaque()) {
+      acc = mix(acc, normalizeEscapes(this._schemeSpecificPart).##)
+    } else if (this._host.isDefined) {
+      acc = mix(acc, normalizeEscapes(this._userInfo).##)
+      acc = mix(acc, this._host.map(_.toLowerCase).##)
+      acc = mix(acc, this._port.##)
+    } else {
+      acc = mix(acc, normalizeEscapes(this._authority).##)
+    }
+    acc = mix(acc, normalizeEscapes(this._path).##)
+    acc = mix(acc, normalizeEscapes(this._query).##)
+    acc = mixLast(acc, normalizeEscapes(this._fragment).##)
     finalizeHash(acc, 3)
   }
 
@@ -334,7 +358,10 @@ object URI {
   }
 
   // IPv4address   = 1*digit "." 1*digit "." 1*digit "." 1*digit
-  private final val ipv4address = "[0-9]{1,3}(?:\\.[0-9]{1,3}){3}"
+  private final val ipv4address = {
+    val digit = "(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)"
+    s"(?:$digit\\.){3}$digit"
+  }
 
   private final val ipv6address = {
     // http://stackoverflow.com/a/17871737/1149944
@@ -800,21 +827,22 @@ object URI {
   }
 
   /** Upper-cases all URI escape sequences in `str`. Used for hashing */
-  private def normalizeEscapes(str: String): String = {
-    var i = 0
-    var res = ""
-    while (i < str.length) {
-      if (str.charAt(i) == '%') {
-        assert(str.length > i + 2, "Invalid escape in URI")
-        res += str.substring(i, i+3).toUpperCase()
-        i += 3
-      } else {
-        res += str.substring(i, i+1)
-        i += 1
+  private def normalizeEscapes(maybeStr: js.UndefOr[String]): js.UndefOr[String] = {
+    maybeStr.map { str =>
+      var i = 0
+      var res = ""
+      while (i < str.length) {
+        if (str.charAt(i) == '%') {
+          assert(str.length > i + 2, "Invalid escape in URI")
+          res += str.substring(i, i+3).toUpperCase()
+          i += 3
+        } else {
+          res += str.substring(i, i+1)
+          i += 1
+        }
       }
+      res
     }
-
-    res
   }
 
   private final val uriSeed = 53722356
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
@@ -18,7 +18,6 @@ import org.junit.Assert._
 import org.junit.Test
 
 import org.scalajs.testsuite.utils.AssertThrows._
-import org.scalajs.testsuite.utils.Platform.executingInJVM
 
 class URITest {
 
@@ -166,6 +165,8 @@ class URITest {
     val y = new URI("http://example.com/asdf%6A")
     val z = new URI("http://example.com/asdfj")
     val rel = new URI("/foo/bar")
+    val rel2 = new URI("/foo/aaa")
+    val rel3 = new URI("/foo/ccc")
 
     assertTrue(x.compareTo(y) > 0)
     assertTrue(x.compareTo(z) < 0)
@@ -177,6 +178,8 @@ class URITest {
     assertTrue(y.compareTo(rel) > 0)
     assertTrue(z.compareTo(rel) > 0)
     assertEquals(0, rel.compareTo(rel))
+    assertTrue(rel.compareTo(rel2) > 0)
+    assertTrue(rel.compareTo(rel3) < 0)
   }
 
   @Test def should_provide_equals(): Unit = {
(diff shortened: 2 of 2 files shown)


javalib/src/main/scala/java/net/URI.scala
```diff
@@ -107,16 +107,22 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+    if (this._scheme != that._scheme) {
+      val s = this._scheme.fold(-1)(s1 => that._scheme.fold(1)(s1.compareToIgnoreCase))
+      if (s == 0 && isOpaque()) cmpSsp()
+      else s
+    } else if (this._isOpaque != that._isOpaque)
+      // A hierarchical URI is less than an opaque URI
+      if (this._isOpaque) 1 else -1
+    else if (_isOpaque)
+      cmpSsp()
+    else if (this._authority != that._authority) {
```
gzm0: While you are here, please put braces around all if / else blocks (unfortunately this seems undocumented but is a style we follow).

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -107,16 +107,22 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+    @inline def cmpSsp(): Int =  {
       val ssp = cmp(this._schemeSpecificPart, that._schemeSpecificPart)
       if (ssp != 0) ssp
       else cmpOpt(this._fragment, that._fragment)
-    } else if (this._authority != that._authority) {
+    }
+
+    if (this._scheme != that._scheme) {
+      val s = this._scheme.fold(-1)(s1 => that._scheme.fold(1)(s1.compareToIgnoreCase))
+      if (s == 0 && isOpaque()) cmpSsp()
```
gzm0: I'm not sure this now works properly if two URIs have the same scheme but one is opaque an the other is not.

Consider replacing all of this with something like this:

```scala
List(
  this._scheme.fold(-1)(s1 => that._scheme.fold(1)(s1.compareToIgnoreCase)),
  this._isOpaque.toInt - that._isOpaque.toInt, // A hierarchical URI is less than an opaque URI
  ...
).find(_ != 0).getOrElse(0)
```

This makes it unmistakably clear that these are merely priorities for comparison. Also, it will fix similar bugs in other parts when the supplied `cmp` is used.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -341,6 +341,10 @@ class URITest {
       assertEquals(new URI("http://example.com/asdf%6a").hashCode,
           new URI("http://example.com/asdf%6A").hashCode)
     }
+    assertEquals(new URI("http://example.com").hashCode(),
+      new URI("http://Example.CoM").hashCode())
+    assertNotEquals(new URI("http://example.com/example-com").hashCode(),
+      new URI("http://Example.CoM/eXAMplE-cOm").hashCode())
```
gzm0: It seems these tests need some maintenance:

- `executingInJVM` should be dropped (we dropped JDK6/7 support).
- Tests for `equals` and `hashCode` really should use the same test cases since the specification is so intimately related.

I'd be great if you could add another commit for this.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -113,6 +113,15 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
 
+    @inline def cmpPathQueryFragment(): Int = {
+      if (this._path != that._path)
+        cmpOpt(this._path, that._path)
+      else if (this._query != that._query)
+        cmpOpt(this._query, that._query)
+      else
+        cmpOpt(this._fragment, that._fragment)
+    }
+
```
gzm0: Same comment than on the previous commit. Probably easier to fix this in a more principled fashion.

javalib/src/main/scala/java/net/URI.scala
exoego: Fixed

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
exoego: Dropped `executingInJVM` in https://github.com/scala-js/scala-js/pull/3690/commits/065ef673dd763d427160d3365b1eaf4f76f55adb

Refactoring tests for `equals` and `hashCode` in https://github.com/scala-js/scala-js/pull/3690/commits/b9717e97663d6d88eb9184128be7d96e41cf4d01

javalib/src/main/scala/java/net/URI.scala
exoego: Refactored in https://github.com/scala-js/scala-js/pull/3690/commits/b2caa31c242ad85efee3fe748e9a7b925f220a0c and fixed an existing issue in subequent https://github.com/scala-js/scala-js/pull/3690/commits/52d33ceece6e3fb4b655a1ccc0dbf7c4df2fe7d7

javalib/src/main/scala/java/net/URI.scala
exoego: Refactored this same as above in https://github.com/scala-js/scala-js/commit/b2caa31c242ad85efee3fe748e9a7b925f220a0c

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -106,36 +106,45 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
-      } else
-        cmpOpt(this._authority, that._authority)
-    } else if (this._path != that._path)
-      cmpOpt(this._path, that._path)
-    else if (this._query != that._query)
-      cmpOpt(this._query, that._query)
-    else
-      cmpOpt(this._fragment, that._fragment)
+      } else cmpOpt(this._authority, that._authority)
+    ).find(_ != 0).getOrElse(0)
```
gzm0: I think the current version is kind of hard to read.

How about something like this:

```scala
List(
    cmpOpt(this._scheme, that._scheme)(_.compareToIgnoreCase),
    this.isOpaque.compareTo(that.isOpaque), // A hierarchical URI is less than an opaque URI
) ++ (
    if (this.isOpaque) {
      List(
        cmp(this._schemeSpecificPart, that._schemeSpecificPart),
      )
    } else if (this._host.isDefined && that._host.isDefined) {
      List(
        cmpOpt(this._userInfo, that._userInfo)(cmp),
        cmpOpt(this._host, that._host)(_.compareToIgnoreCase),
        cmpOpt(this._port, that._port)(_ - _),  // change _port to js.UndefOr[Int] and fold into -1 only when getPort is called.
      )
    } else {
      List(
         cmpOpt(this._authority, that._authority)(cmp)
      )
    }
) ++ List(
    cmpOpt(this._path, that._path)(cmp),
    cmpOpt(this._query, that._query)(cmp),
    cmpOpt(this._fragment, that._fragment)(cmp),
).find(_ != 0).getOrElse(0)
```

It is unfortunately not as nice as I anticipated, mostly because of the if-branches. Not sure how to improve though. Maybe something with `Option#orElse` or helper functions would work better.

exoego: Refactoring so and cleanup commits.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -101,41 +101,37 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+    ) ++ (
+      if (this.isOpaque()) {
+        List(
+          cmp(this._schemeSpecificPart, that._schemeSpecificPart),
+        )
+      } else if (this._host.isDefined && that._host.isDefined) {
+        List(
+          cmpOpt(this._userInfo, that._userInfo)(cmp),
+          cmpOpt(this._host, that._host)(_.compareToIgnoreCase(_)),
+          cmpOpt(this._port, that._port)(_ - _),  // change _port to js.UndefOr[Int] and fold into -1 only when getPort is called.
```
gzm0: You actually need to make this change for this to compile :)

exoego: Oh, I missed.
Fixed in https://github.com/scala-js/scala-js/pull/3690/commits/dfe44551c417702d25aa841379360d442a559ca3

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -54,7 +54,7 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
   private val _authority = fld(AbsAuthority, RelAuthority).filter(_ != "")
   private val _userInfo = fld(AbsUserInfo, RelUserInfo)
   private val _host = fld(AbsHost, RelHost)
-  private val _port = fld(AbsPort, RelPort).fold(-1)(_.toInt)
+  private val _port = fld(AbsPort, RelPort)
```
gzm0: `_port` should still be an `Int` field (add `.map(_.toInt)`)

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -101,41 +101,37 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+      } else {
+        List(
+          cmpOpt(this._authority, that._authority)(cmp)
+        )
+      }
+    ) ++ List(
+      cmpOpt(this._path, that._path)(cmp),
+      cmpOpt(this._query, that._query)(cmp),
+      cmpOpt(this._fragment, that._fragment)(cmp)
+    )).find(_ != 0).getOrElse(0)
```
gzm0: I'm unsure about the exact indentation rules of this. @sjrd will know :)

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -820,6 +820,14 @@ object URI {
     res
   }
 
+  /** Lower-cases host in `str` only if host exists. Used for hashing */
+  private def normalizeHost(str: String, host: js.UndefOr[String]): String = {
```
gzm0: I'm not sure this is correct. What if the authority part is `FoO.CoM@FoO.CoM`?

javalib/src/main/scala/java/net/URI.scala
sjrd: It's mostly correct, except that every time something is inside `( ... )` (so in this case inside `List( ... )`, the indentation should be 4 spaces instead of 2. So it should be as follows:
```scala
    (List(
        cmpOpt(this._scheme, that._scheme)(_.compareToIgnoreCase(_)),
        this.isOpaque.compareTo(that.isOpaque) // A hierarchical URI is less than an opaque URI
    ) ++ (
        if (this.isOpaque()) {
          List(
              cmp(this._schemeSpecificPart, that._schemeSpecificPart)
          )
        } else if (this._host.isDefined && that._host.isDefined) {
          List(
              cmpOpt(this._userInfo, that._userInfo)(cmp),
              cmpOpt(this._host, that._host)(_.compareToIgnoreCase(_)),
              cmpOpt(this._port, that._port)(_.toInt - _.toInt)
          )
        } else {
          List(
              cmpOpt(this._authority, that._authority)(cmp)
          )
        }
    ) ++ List(
        cmpOpt(this._path, that._path)(cmp),
        cmpOpt(this._query, that._query)(cmp),
        cmpOpt(this._fragment, that._fragment)(cmp)
    )).find(_ != 0).getOrElse(0)
```

sjrd: Btw you should use `:::` instead of `++` to concatenate `List`s. It is more efficient.

javalib/src/main/scala/java/net/URI.scala
exoego: Addressed in https://github.com/scala-js/scala-js/pull/3690/commits/c0eaf768aa58dde27f84927961440e898479bf33

javalib/src/main/scala/java/net/URI.scala
exoego: Indentation in https://github.com/scala-js/scala-js/pull/3690/commits/c2aaaa79c8380a5a1c8fdd21a6f6436ac16fb679
`:::` in https://github.com/scala-js/scala-js/pull/3690/commits/302c4b1d04b09b2e73be88f21e5ee1ef5e8d89c6

javalib/src/main/scala/java/net/URI.scala
exoego: That sounds right.
Addressed in https://github.com/scala-js/scala-js/pull/3690/commits/bba5ecc077c2298f087889cd701f6fe40339b537

exoego:
I will squash recent minor commits when PR is good to go.

javalib/src/main/scala/java/net/URI.scala
gzm0: It feel this re-parses part of the URI again. Have you considered using the pre-parsed fields for this? (as in composing the hash code based on the present fields, rather than re-parsing and changing them)?

exoego: Done in https://github.com/scala-js/scala-js/pull/3690/commits/cfdb791567791ced8e75ee70a407a7eb3d0d30ae
I think *before* is simpler than *after* though re-parsing.

gzm0: Personally I do prefer the new solution. @sjrd opinions?

Not sure we need the `List` with the `fold` rather than keeping mutable accumulation, but otherwise I like it.

sjrd: Yes, I'm not quite sure what the old version was, but the new one is good.

I would also revert to using the mutable accumulator. It was more readable *and* more efficient.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -166,10 +162,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
     var acc = URI.uriSeed
-    acc = mix(acc, _scheme.##) // scheme may not contain escapes
-    acc = mix(acc, normalizeEscapes(_schemeSpecificPart).##)
-    acc = mixLast(acc, _fragment.map(normalizeEscapes).##)
-
+    acc = mix(acc, _scheme.map(_.toLowerCase).##)
+    if (this.isOpaque()) {
+      acc = mix(acc, normalizeEscapes(this._schemeSpecificPart).##)
+    } else if (this._host.isDefined) {
+      acc = mix(acc, this._userInfo.##)
```
gzm0: `userInfo` may contain escapes, so this needs to be normalized.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -166,10 +162,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
+      acc = mix(acc, this._host.map(_.toLowerCase).##)
+      acc = mix(acc, this._port.##)
+    } else {
+      acc = mix(acc, this._authority.##)
```
gzm0: `authority` may contain escapes, so this needs to be normalized.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -166,10 +162,19 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
     import URI.normalizeEscapes
 
     var acc = URI.uriSeed
-    acc = mix(acc, _scheme.##) // scheme may not contain escapes
-    acc = mix(acc, normalizeEscapes(_schemeSpecificPart).##)
-    acc = mixLast(acc, _fragment.map(normalizeEscapes).##)
-
+    acc = mix(acc, _scheme.map(_.toLowerCase).##)
```
gzm0: The comment about escapes should be retained.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -101,41 +101,37 @@ final class URI(origStr: String) extends Serializable with Comparable[URI] {
...
-      cmpOpt(this._path, that._path)
-    else if (this._query != that._query)
-      cmpOpt(this._query, that._query)
-    else
-      cmpOpt(this._fragment, that._fragment)
+    ) ::: List(
+        cmpOpt(this._path, that._path)(cmp),
+        cmpOpt(this._query, that._query)(cmp),
+        cmpOpt(this._fragment, that._fragment)(cmp)
+    )).find(_ != 0).getOrElse(0)
```
sjrd: Looking back at this, I don't think the strategy of building a list with everything then filtering it is the best way. I'm all for functional code where it makes sense, but here it leads to completely unnecessary computations.

Have you considered a more imperative variant, like the following?
```scala
val cmpScheme = cmpOpt(this._scheme, that._scheme)(_.compareToIgnoreCase(_))
if (cmpScheme != 0)
  return cmpScheme
val cmpIsOpaque = this.isOpaque.compareTo(that.isOpaque) // A hierarchical URI is less than an opaque URI
if (cmpIsOpaque != 0)
  return cmpIsOpaque
if (this.isOpaque()) {
  ...
} else {
  ...
}
...
0
```
The new usage of `List` is also not in line with recent efforts to stop using Scala collections in the implementation of the javalib, which is what brought me back here, TBH.

exoego:
> 
> 
> What's the status of this PR?

[The last CI error](https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3690/13/pipeline#step-218-log-1) seems spurious, and I could not rerun so leavee it.

I will update per your request to avoid using Scala collections.

exoego:
Updated and CI passes.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -800,23 +827,23 @@ object URI {
...
-    var res = ""
-    while (i < str.length) {
-      if (str.charAt(i) == '%') {
-        assert(str.length > i + 2, "Invalid escape in URI")
-        res += str.substring(i, i+3).toUpperCase()
-        i += 3
-      } else {
-        res += str.substring(i, i+1)
-        i += 1
+  private def normalizeEscapes(maybeStr: js.UndefOr[String]): js.UndefOr[String] =
```
sjrd: The body of this method should have `{}`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -190,8 +193,33 @@ class URITest {
     assertTrue(x == x)
     assertTrue(y == y)
     assertTrue(z == z)
+  }
+
+  @Test def equals_and_hashCode_should_produces_same_result(): Unit = {
+    val equalsPairs: Seq[(URI,URI)] = Seq(
```
sjrd: Missing space after `,`

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -190,8 +193,33 @@ class URITest {
...
+    val nonEqualPairs: Seq[(URI,URI)] = Seq(
+      (new URI("http://example.com/example-com"), new URI("http://Example.CoM/eXAMplE-cOm")),
+      (new URI("http://example.com@example.com"), new URI("http://EXAMPLE.COM@EXAMPLE.Com")),
+      (new URI("foo:helloWorld%6b%6C"), new URI("foo:helloWorld%6C%6b"))
+    )
+    nonEqualPairs.foreach { case (a, b) =>
+      assertNotEquals(a, b)
+      // Note: hashCode is not restricted to produce same result even if a.equals(b) is false.
+      assertNotEquals("a does not equal to b, but produces same hashCode. Pick different test data",
+        a.hashCode(), b.hashCode())
```
sjrd: 4 spaces of indentation

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
```diff
@@ -337,10 +365,12 @@ class URITest {
   }
 
   @Test def should_provide_hashCode(): Unit = {
-    if (!executingInJVM) { // Fails on JDK6 and JDK7
-      assertEquals(new URI("http://example.com/asdf%6a").hashCode,
-          new URI("http://example.com/asdf%6A").hashCode)
-    }
+    assertEquals(new URI("http://example.com/asdf%6a").hashCode,
+      new URI("http://example.com/asdf%6A").hashCode)
```
sjrd: 4 spaces of indentation (same for the two below)

exoego: Fixed

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
exoego: Fixed

javalib/src/main/scala/java/net/URI.scala
exoego: Fixed

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/net/URITest.scala
exoego: Fixed

PR #3694, opened 2019-06-11
Fix #3607: Include trailing dollars in getSimpleName.

### LINKED ISSUE #3607: Calling getClass.getSimpleName on objects returns name without trailing $
There was some old closed issues about similar issue for getName #157 and #185 
But as I test it now with version 0.6.26, I don't get any trailing $ in object class name.
```
object Test
println(Test.getClass.getSimpleName)    /// result: Test
```

tOverney:


gzm0:
I think this should target 0.6?

tOverney:
indeed!

PR #3700, opened 2019-06-16
 Also run test-common tests in JS

gzm0:
This is possible since we now have async support for JUnit.


gzm0:
Added it to "bootstrap" not the nicest but also not the worst...

sjrd:
Did you consider putting it next to `jUnitTestOutputsJS/test`? They seem to be more or less part of the same "category" of projects (supporting test infrastructure on the JS side).

gzm0:
Rebased and updated.

PR #3702, opened 2019-06-18
[no-master] Enable Scala 2.13.0 in the tools and all the other artifacts.

sjrd:
It's [no-master]  because

* The `@deprecatedName` and `scopt` commits are irrelevant in master.
* The other two commits are virtually all-conflicts if I try to cherry-pick them on master. An independent forward-port will be easier to review.


tools/shared/src/main/scala/org/scalajs/core/tools/linker/CollectionsCompat.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import scala.collection.mutable
+
+private[linker] object CollectionsCompat {
+  implicit class MutableMapCompatOps[K, V](val __self: mutable.Map[K, V])
+      extends AnyVal {
+
+    // filterInPlace replaces retain
+    def filterInPlace(p: (K, V) => Boolean): Unit = {
+      // Believe it or not, this is the implementation of `retain` in 2.12.x:
```
gzm0: Hahahahaha, were you anticipating me complaining :P

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -1027,31 +1029,37 @@ object GenIncOptimizer {
...
+    type ParIterable[V] <: AnyRef
+    type Addable[V] <: AnyRef
 
     def emptyAccMap[K, V]: AccMap[K, V]
     def emptyMap[K, V]: Map[K, V]
     def emptyParMap[K, V]: ParMap[K, V]
     def emptyParIterable[V]: ParIterable[V]
 
     // Operations on ParMap
+    def parMapIsEmpty[K, V](map: ParMap[K, V]): Boolean
```
gzm0: Simply name this `isEmpty`? `parMap` seems superfluous (I understand that overloading is a problem with types that end up being erased, but the whole thing is already terribly inconsistent anyways :S).

tools/shared/src/main/scala/org/scalajs/core/tools/linker/CollectionsCompat.scala
sjrd: Actually I was anticipating myself looking back at this code and being appalled at the inefficiency of it.

sjrd:
Updated.

Besides addressing the review, I have also added the whitelist and blacklist for the `scalaTestSuite`, so that the CI can pass. This discovered two bugs filed as #3704 and #3705, which should be fixed separately, as followups.

scala-test-suite/src/test/resources/2.13.0/BlacklistedTests.txt
```diff
@@ -122,12 +172,20 @@ scala/util/matching/RegexTest.scala
 # Require strict-floats
 scala/math/BigDecimalTest.scala
 
-# Difference of getClass() on primitive values
-scala/collection/immutable/RangeTest.scala
-
-# Bugs
-scala/collection/convert/MapWrapperTest.scala
+# Fails for a BigDecimal range with augmented precision (might be an actual bug)
```
gzm0: Do we need to file an issue for this at least?

sjrd: Filed as #3706.

PR #3707, opened 2019-06-21
Recognize Metals as an IDE for the special treatment of the build.

sjrd:
The same accommodations that were done for Eclipse are also valuable for Metals. Since the environment variable `METALS_ENABLED` is automatically added by Metals when it imports a build, this change gives a better out-of-the-box experience with Metals.

We also now mention Metals-based IDEs as the recommended tools to develop Scala.js in the Developing guide.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -47,8 +47,10 @@ object ExposedValues extends AutoPlugin {
 
 object Build {
 
-  val isGeneratingEclipse =
-    Properties.envOrElse("GENERATING_ECLIPSE", "false").toBoolean
+  val isGeneratingForIDE = {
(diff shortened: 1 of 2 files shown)


project/Build.scala
```diff
@@ -47,8 +47,10 @@ object ExposedValues extends AutoPlugin {
 
 object Build {
 
-  val isGeneratingEclipse =
-    Properties.envOrElse("GENERATING_ECLIPSE", "false").toBoolean
+  val isGeneratingEclipse = {
```
gzm0: Hum, maybe this `val` should be renamed?

sjrd: Done.

PR #3710, opened 2019-07-01
Fix #3709: Add TypedArray.of and TypedArray.from

### LINKED ISSUE #3709: Add TypedArray.of and TypedArray.from
Those are factory ethods to create an instance of *TypedArray* from various arguments.

It is currently missing in `TypedArrayStatic`.
https://github.com/scala-js/scala-js/blob/1d96d3cc24250e874ebc9aa6f78eafb060678512/library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala#L68-L70

# *TypedArray*.from
* https://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%.from
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from

# *TypedArray*.of
* https://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%.of
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/of

exoego:
Closes #3709

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
+++ b/library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
@@ -65,6 +65,20 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
  *  Static information that exists for any concrete TypedArray
  */
 @js.native
-trait TypedArrayStatic extends js.Object {
+trait TypedArrayStatic[T, Repr] extends js.Object {
   val BYTES_PER_ELEMENT: Int = js.native
+
+  /** Returns a new array from a set of elements. */
+  def of(items: T*): Repr = js.native
+
+  /** Creates an array from an `iterable` object. */
+  def from(iterable: js.Iterable[T]): Repr = js.native
+
+  /** Creates an array from an `iterable` object. */
+  def from[E](iterable: js.Iterable[E],
+      mapFn: js.Function1[E, T]): Repr = js.native
+
+  /** Creates an array from an `iterable` object. */
+  def from[D, E](iterable: js.Iterable[E], mapFn: js.ThisFunction1[D, E, T],
+      thisArg: D): Repr = js.native
 }
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -41,17 +41,19 @@ object BinaryIncompatibilities {
   )
 
   val Library = Seq(
+      // New methods in a native JS trait, not an issue
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.*")
   )
 
   val TestInterface = Seq(
-    /* Things that moved to scalajs-test-bridge.
-     * They were private[scalajs] or stricter, so not an issue.
-     */
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testcommon.*"),
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.HTMLRunner"),
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.HTMLRunner$*"),
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.TestDetector"),
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.TestDetector$*"),
-    ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.internal.*")
+      /* Things that moved to scalajs-test-bridge.
+       * They were private[scalajs] or stricter, so not an issue.
+       */
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testcommon.*"),
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.HTMLRunner"),
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.HTMLRunner$*"),
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.TestDetector"),
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.TestDetector$*"),
+      ProblemFilters.exclude[MissingClassProblem]("org.scalajs.testinterface.internal.*")
   )
 }
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
@@ -27,6 +27,10 @@ import js.typedarray._
 trait TypedArrayTest[V, T <: TypedArray[V, T]] {
 
   def bytesPerElement: Int
+  def ofFn(items: V*): T
+  def fromFn(iterable: js.Iterable[V]): T
+  def fromFn[E](iterable: js.Iterable[E])(mapper: js.Function1[E, V]): T
+  def fromFn[D, E](iterable: js.Iterable[E], thisObj: D)(mapper: js.ThisFunction1[D, E, V]): T
   def lenCtor(len: Int): T
   def tarrCtor(tarr: TypedArray[_, _]): T
   def itCtor(arr: js.Iterable[_]): T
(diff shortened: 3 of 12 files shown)


library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -65,6 +65,21 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
...
   val BYTES_PER_ELEMENT: Int = js.native
+
+  /** Returns a new array from a set of elements. */
+  def of(items: T*): Repr = js.native
+
+  /** Creates an array from an `arrayLike` or iterable object. */
+  def from(arrayLike: js.Array[T]): Repr = js.native
+
+  /** Creates an array from an `arrayLike` or iterable object. */
+  def from[E](arrayLike: js.Array[E], mapfn: js.Function1[E, T]): Repr = js.native
```
exoego: [Like TypeScript defines so](https://github.com/microsoft/TypeScript/blob/b36c8a06902ca47da1568167991bd0bc28ced80e/src/lib/es5.d.ts#L1396-L1399), I think `arrayLike` should be a strcture type like
```scala
def from(arrayLike: { 
    def length: Int
    @JSBracketAccess def get(n: Int): T
})
```

But I am not sure what is a Scala.js equilvalent for that.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -65,6 +65,21 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
...
+
+  /** Creates an array from an `arrayLike` or iterable object. */
+  def from[E](arrayLike: js.Array[E], mapfn: E =>T): Repr = js.native
```
sjrd: This signature is erroneous. It tries to accept a Scala function, but only JS functions will work. This overload should be removed.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
sjrd: There is no equivalent in Scala.js. Structural JS types are not supported. Just ignore the array-like's that are not true arrays nor true iterables. If someone really wants to call `from` with a non-iterable array-like, they can cast it to a `js.Array`.

test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
```diff
@@ -36,6 +40,32 @@ trait TypedArrayTest[V, T <: TypedArray[V, T]] {
...
+
+  @Test def should_provide_a_factory_method_from(): Unit = {
+    val x = fromFn(js.Array(intToV(0), intToV(1), intToV(2)))
+    assertEquals(3, x.length)
+    assertEquals(2, x(2))
+  }
+
+  @Test def should_provide_a_factory_method_from_with_mapping_function(): Unit = {
+    val src = js.Array("", "a", "bc")
+    val x = fromFn(src)(s => intToV(s.length))
```
sjrd: You'll need an explicit type for `s` here for this to compile under Scala 2.10 and 2.11.

test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
```diff
@@ -36,6 +40,32 @@ trait TypedArrayTest[V, T <: TypedArray[V, T]] {
...
+    assertEquals(3, x.length)
+    assertEquals(2, x(2))
+  }
+
+  @Test def should_provide_a_factory_method_from_with_mapping_function_and_thisArg(): Unit = {
+    val src = js.Array("", "a", "bc")
+    val x = fromFn(src, 10)((thisArg,s) => intToV(s.length * thisArg))
```
sjrd: You'll need an explicit type for `thisArg` and `s` here for this to compile under Scala 2.10 and 2.11.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -65,6 +65,21 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
...
+
+  /** Creates an array from an `arrayLike` or iterable object. */
+  def from[D,E](arrayLike: js.Array[E], mapfn: js.ThisFunction1[D, E, T], thisArg: D): Repr = js.native
```
sjrd: Coding style: missing space after `,` in `[D,E]`.
Also try to limit line length at 80 characters.

This comment also applies to several definitions in the tests.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -65,6 +65,21 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
  *  Static information that exists for any concrete TypedArray
  */
 @js.native
-trait TypedArrayStatic extends js.Object {
+trait TypedArrayStatic[T, Repr] extends js.Object {
```
sjrd: The commit message should clearly explain why this change is fine for the 0.6.x series, because it's non-obvious. It is clearly a source-breaking change, but it is binary compatible. The additions of methods in the open trait also binary compatible, but only because this is a *native JS trait*.

Ideally we should prevent extending this trait from anywhere but this package, but that's not possible unless we put all the typed array definitions in the same file, and make it `sealed`.

exoego:
Updated per review and squashed commits.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
exoego: Updated the commit message in https://github.com/scala-js/scala-js/pull/3710/commits/11138736c15c11070755c56677c1723cc8b2fb8f

sjrd:
The CI says the following, of course:
```
[info] Scala.js library: found 4 potential binary incompatibilities while checking against org.scala-js:scalajs-library_2.11:0.6.28 
[error]  * method of(scala.collection.Seq)java.lang.Object in trait scala.scalajs.js.typedarray.TypedArrayStatic is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.of")
[error]  * method from(scala.scalajs.js.Iterable,scala.scalajs.js.Function1)java.lang.Object in trait scala.scalajs.js.typedarray.TypedArrayStatic is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.from")
[error]  * method from(scala.scalajs.js.Iterable,scala.scalajs.js.ThisFunction1,java.lang.Object)java.lang.Object in trait scala.scalajs.js.typedarray.TypedArrayStatic is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.from")
[error]  * synthetic method from$default$2()scala.runtime.Nothing# in trait scala.scalajs.js.typedarray.TypedArrayStatic is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.from$default$2")
```
That's because MiMa doesn't know that it's OK to add methods to a native JS trait. You'll have to filter how those in `project/BinaryIncompatibilities.scala` with:
```scala

  val Library = Seq(
      // New methods in a native JS trait, not an issue
      ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.*")
  )
```

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -65,6 +65,17 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
...
   val BYTES_PER_ELEMENT: Int = js.native
+
+  /** Returns a new array from a set of elements. */
+  def of(items: T*): Repr = js.native
+
+  /** Creates an array from an `iterable` object. */
+  def from[E](iterable: js.Iterable[E],
+      mapFn: js.Function1[E, T] = ???): Repr = js.native
```
sjrd: This definition is unsound, because I can call it with:
```scala
val a = js.Array[String]("foo", "bar")
val b = js.typedarray.Int32Array.from(a)
```
and the compiler won't complain (it will simply instantiate `E = String`).

You should keep a separate overload with only 1 argument and no type parameter, for the case where no map function is provided.

test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
```diff
@@ -279,6 +319,11 @@ class Uint8ArrayTest extends TypedArrayTest[Short, Uint8Array] {
 object Uint8ClampedArrayTest extends Requires.TypedArray
 
 class Uint8ClampedArrayTest extends TypedArrayTest[Int, Uint8ClampedArray] {
+  def ofFn(items: Int*): Uint8ClampedArray = Uint8ClampedArray.of(items: _*)
+  def fromFn(iterable: js.Iterable[Int]): Uint8ClampedArray = Uint8ClampedArray.from(iterable)
+  def fromFn[E](iterable: js.Iterable[E])(fn: js.Function1[E, Int]): Uint8ClampedArray = Uint8ClampedArray.from(iterable, fn)
```
sjrd: The CI says that this line is too long.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
exoego: Ah, I did not know it is unsound.

Fixed in https://github.com/scala-js/scala-js/commit/a7dc30d9c1e6a518e8c12615b41c51d680aa5f0a
(commits are squashed as requested always)

test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
exoego: Fixed in https://github.com/scala-js/scala-js/commit/a7dc30d9c1e6a518e8c12615b41c51d680aa5f0a
(commits are squashed as requested always)

project/BinaryIncompatibilities.scala
```diff
@@ -41,6 +41,8 @@ object BinaryIncompatibilities {
   )
 
   val Library = Seq(
+    // New methods in a native JS trait, not an issue
+    ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.scalajs.js.typedarray.TypedArrayStatic.*")
```
sjrd: The indentation should be 4 spaces here (see in `val Tools` above).

exoego: Changed indentations of `Library` and `TestInterface` to `4`.

PR #3711, opened 2019-07-03
Further improve the default import of the build in Metals.

sjrd:
When generating for IDEs, now we:

* Disable source- and resource-generators.
* Use stub definitions for sources that are normally generated (in particular, the `ScalaJSEnvHolder`).
* Do not add the `-P:scalajs:*` options (which is important because we already disable the Scala.js compiler plugin).
* Skip the JVM back-end phases like we do for the normal build.


project/Build.scala
```diff
@@ -110,6 +110,25 @@ object Build {
...
+  private def addScalaJSCompilerOption(option: String): Setting[_] =
+    addScalaJSCompilerOption(Def.setting(option))
+
+  private def addScalaJSCompilerOption(option: Def.Initialize[String]): Setting[_] =
+    addScalaJSCompilerOption(None, option)
+
+  private def addScalaJSCompilerOptionInConfig(config: Configuration,
+      option: String): Setting[_] = {
+    addScalaJSCompilerOption(Some(config), Def.setting(option))
+  }
```
gzm0: Can't we use sbt's `inConfig` for this?

project/Build.scala
```diff
@@ -322,8 +322,12 @@ object Build {
       }
   )
 
-  val noClassFilesSettings: Setting[_] =
-    scalacOptions in (Compile, compile) += "-Yskip:cleanup,icode,jvm"
+  val noClassFilesSettings: Setting[_] = {
+    scalacOptions in (Compile, compile) += {
+      if (isGeneratingForIDE) "-Yskip:jvm"
```
gzm0: Why do we need to tell the IDE what phases to skip?

project/Build.scala
sjrd: No, we can't, because that would also apply the config to the `option`. That's only really a problem for the case with the `Initialize[String]`, but then I thought I would keep consistency.

(Also, `inConfig` always returns a `Seq[Setting[_]]`, and at the existing call sites I needed a single `Setting[_]`, so using `inConfig` would have required more refactoring.)

project/Build.scala
sjrd: Metals uses a phase after `jvm`, so we can't just `-Ystop-after`, or that phase will also be skipped, and some important features of Metals are lost (in particular, go-to-definition). We have to skip `jvm` because it crashes in the `javalanglib`.

sjrd:
> Re-compiler plugin: does Metals not support compiler plugins at all? This seems like a major shortcoming especially when developing for non-JVM platforms (n.b. Scala.js users).

We're not deactivating the Scala.js compiler plugin because of Metals. We also did it for Eclipse. We're deactivating it because IDEs get super confused when using a compiler plugin that is developed in the same build. That's because the .jar they need to load does not exist as long as the compiler plugin hasn't been properly built by sbt. And then even after that, when the compiler changes, the IDEs need to re-load the plugin live, which triggers useless re-compilations inside the IDE, slowing the whole thing down (assuming it doesn't simply crash).

In any other build, Metals is happy to apply the Scala.js compiler plugin, because it's known once and for all and it doesn't change.

gzm0:
Thanks for the explanations.

Issue #3705, opened 2019-06-19
ArrayBuilder.generic.addAll is broken on 2.13


sjrd:
https://github.com/scala-js/scala-js/blob/2e2e2bc0c0b4984a5ea646a9ca8b391f10f123fd/scalalib/overrides-2.13/scala/collection/mutable/ArrayBuilder.scala#L115-L120

The implementation passes `length` as second argument to `slice`, but that argument is an `until`. It should give `offset + length` to be compatible.

Failing test case:
```scala
    val b = mutable.ArrayBuilder.make[Int]
    val arr = Array[Int](1, 2, 3, 4, 5)
    b.addAll(arr, 3, 2)
    assertArrayEquals(Array[Int](4, 5), b.result())
```

This indirectly causes the upstream JUnit test `scala/collection/ArrayOpsTest.scala` to fail.

Note: the call to `ensureSize` is completely useless and should be removed while we're there.

sjrd:
Fixed implementation:
```scala
    /** Add a slice of an array */
    override def addAll(xs: Array[_ <: T], offset: Int, length: Int): this.type = {
      val end = offset + length
      var i = offset
      while (i < end) {
        this += xs(i)
        i += 1
      }
      this
    }
```

sjrd:
If someone wants to tackle this:

Fixing this should be done in the 0.6.x branch.

Tests for this issue should go to [`ArrayBuilderTest`](https://github.com/scala-js/scala-js/blob/0.6.x/test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/ArrayBuilderTest.scala). Since `addAll` did not exist in earlier versions of Scala, you'll need to use some implicit magic similar to what's done at https://github.com/scala-js/scala-js/blob/2e2e2bc0c0b4984a5ea646a9ca8b391f10f123fd/test-suite/shared/src/test/require-sam/org/scalajs/testsuite/compiler/SAMTest.scala#L113-L121

sjrd:
Fixed in 3f7b83b19176aac85a936801d7e57b9176ede0d1.

PR #3712, opened 2019-07-03
Fix #3704 and #3705, two regressions when using Scala 2.13.0.

### LINKED ISSUE #3704: Implement java.util.Collection.removeIf
Scaladoc: https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeIf-java.util.function.Predicate-

It's used by `JSetWrapper.filterInPlace` upstream since https://github.com/scala/scala/commit/9fcbb851ad6634655eda0959b7cbc280a1731d42 (merged in 2.13.0-RC1).

Not having it prevents us from enabling the upstream JUnit test `scala/collection/convert/JSetWrapperTest.scala`, and is potentially a source of *new* linking errors when migrating to Scala 2.13.

sjrd:


project/Build.scala
```diff
@@ -1549,6 +1549,25 @@ object Build {
...
+      val needsTargetFlag = scalaVersion.value match {
+        case PartialVersion("2", "10", n) => n.toInt >= 7
+        case PartialVersion("2", "11", n) => n.toInt >= 12
+        case _                            => false
+      }
+      if (needsTargetFlag)
+        Seq("-target:jvm-1.8")
+      else
+        Nil
+    },
```
sjrd: Updated with the above chunk in the build. It's necessary for scalac to accept the calls to `Predicate.isEqual` in the test suite.

I locally tested with the change with 2.10.6, 2.10.7, 2.11.11 and 2.11.12.

PR #3716, opened 2019-07-08
Optimize calls to `Array.apply()` overloads with literal varargs.

sjrd:
In Scala/JVM, this optimization is performed by the `cleanup` phase, but it comes after our back-end, so we have to reimplement it ourselves.

We then also change some uses of `Array.apply` in `Character.scala` so that the new optimization can kick in.

### DIFF
--- a/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
+++ b/compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
@@ -50,6 +50,45 @@ class OptimizationTest extends JSASTTest {
   }
 
   @Test
+  def testArrayApplyOptimization: Unit = {
+    /* Make sure Array(...) is optimized away completely for several kinds
+     * of data types, with both the generic overload and the ones specialized
+     * for primitives.
+     */
+    """
+    class A {
+      val a = Array(5, 7, 9, -3)
+      val b = Array("hello", "world")
+      val c = Array('a', 'b')
+      val d = Array(Nil)
+      val e = Array(5.toByte, 7.toByte, 9.toByte, -3.toByte)
+    }
+    """.
+    hasNot("any LoadModule of the scala.Array companion") {
+      case js.LoadModule(jstpe.ClassType("s_Array$")) =>
+    }
+
+    /* Using [] with primitives produces suboptimal trees, which cannot be
+     * optimized. We should improve this in the future, if possible. This is
+     * particularly annoying for Byte and Short, as it means that we need to
+     * write `.toByte` for every single element if we want the optimization to
+     * kick in.
+     *
+     * Scala/JVM has the same limitation.
+     */
+    """
+    class A {
+      val a = Array[Int](5, 7, 9, -3)
+      val b = Array[Byte](5, 7, 9, -3)
+    }
+    """.
+    hasExactly(2, "calls to Array.apply methods") {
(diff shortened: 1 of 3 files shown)


compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
```diff
@@ -49,6 +49,46 @@ class OptimizationTest extends JSASTTest {
...
+      val a = Array(5, 7, 9, -3)
+      val b = Array("hello", "world")
+      val c = Array('a', 'b')
+      val d = Array(Nil)
+      val e = Array(5.toByte, 7.toByte, 9.toByte, -3.toByte)
+    }
+    """.
+    hasNot("any of the Array.apply methods") {
+      case js.Apply(js.LoadModule(jstpe.ClassType("s_Array$")), js.Ident(methodName, _), _)
+          if methodName.startsWith("apply__") =>
```
gzm0: Could we make this more restrictive? IIUC the goal of this is to not require scalalib for this. So maybe we should simply check `js.LoadModule(jstpe.ClassType("s_Array$"))` is not there? (or maybe even *any* `js.LoadModule` tree)?

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
```diff
@@ -49,6 +49,46 @@ class OptimizationTest extends JSASTTest {
...
+    """
+    class A {
+      val a = Array[Int](5, 7, 9, -3)
+      val b = Array[Byte](5, 7, 9, -3)
+    }
+    """.
+    hasExactly(2, "calls to Array.apply methods") {
+      case js.Apply(js.LoadModule(jstpe.ClassType("s_Array$")), js.Ident(methodName, _), _)
+          if methodName.startsWith("apply__") =>
+    }
```
gzm0: Why do we need to test this?

sjrd: We don't need to, but I feel it's a nice way to document that limitation for ourselves, when inevitably we'll come back to this commit because we're wondering why a call to `Array()` is not optimized the way we'd like.

sjrd:
> Could you drop a short comment why our backend runs before cleanup?

Because we don't want any other stuff that cleanup does to happen. For example it rewrites reflective calls into actual calls to the Java reflection API, which is a big no no for us. Moreover, there are other phases *before* cleanup that we don't want, such as `delambdafy`.

compiler/src/test/scala/org/scalajs/core/compiler/test/OptimizationTest.scala
sjrd: Done, updated. Just any `LoadModule` didn't work, but since I'm on mobile I couldn't figure out why.

gzm0:
> Because we don't want any other stuff that cleanup does to happen. For example it rewrites reflective calls into actual calls to the Java reflection API, which is a big no no for us. Moreover, there are other phases before cleanup that we don't want, such as delambdafy.

Ah, yes sorry. I meant add it to the commit message. Currently it reads a bit strange, like we can't change when our phase runs. (`cleanup`'s description also says "platform specific" or something the like, so it makes sense it doesn't run).

sjrd:
OK, I updated the commit message. Is it better?

gzm0:
LGTM

PR #3717, opened 2019-07-09
Make the IR of the javalanglib independent of the Scala library.

sjrd:
~~Based on #3720.~~

We do this with a combination of

* Changes in the sources in order to not refer to any non-native code from the Scala library.
* A postprocessor for the IR to erase native JS types to `Object`.

---

The next step will be to do this for `javalib` as well, and then we'll be able to publish a *unique* `scalajs-jdk.jar` containing .sjsir files for the JDK, that is completely agnostic from Scala (and can be dependent on by the `scalajs-library_xy.jar`s of all versions *xy*, or even by other languages targeting the IR).


sjrd:
@gzm0 What do you think about the overall idea?

gzm0:
I think this is an excellent idea. What I do not fully understand is why we need the erasure transformer.

- Aren't native JS types independent of the Scalalib anyways?
- Where do we even need native JS types.

Another concern I have is how do we make sure we don't start using non-native code again?

sjrd:
> Aren't native JS types independent of the Scalalib anyways?

No, because their IR definition either resides in the Scala.js lib (e.g., `js.Object`), or they are defined within the javalanglib itself, but then they *inherit* from something in the Scala.js lib (at best, only from `js.Any`).

> Where do we even need native JS types.

Everywhere we use native JS things from the ES standard library to implement things in `java.lang`. Trivial examples include: methods of `String`, `Math.sin`, `Object.prototype.hasOwnProperty`, `parseInt`, etc. Unless we *systematically* use `js.Dynamic` with casts for arguments *and* for the result type, which is extremely annoying, we cannot avoid *some* references to JS types to end up in the IR.

> Another concern I have is how do we make sure we don't start using non-native code again?

The `JavalibIRCleaner` reports an error and fails the build if it encounters any reference to something defined in the `scala.*` namespace.

gzm0:
You seem to be actively working on this and I'm unsure about the prioritization of it. Is this required:

- for sjs 1.0 (IIUC, we could do this later)
- for other projects you are working on (dotty)
- just something you want to work on

I'd really like to get 1.0 out and try to focus all of our available attention on it.

sjrd:
TBH it's mostly something I like to work on. It's definitely not required for other projects like dotty.

And no, I don't think it's critical for 1.0 either. I believe that if a future version of `scalajs-library_2.12.jar` depends on a newly created `scalajs-javalib.jar`, Ivy resolution will get everything in order. I kind of would like to evaluate if it's actually attainable, though. At least I'd like to get to a point where we don't have any alternative implementation in the `minilib`, but rather can directly use the artifacts of the real lib in the `minilib`. That would provide good validation of the concept.

The changes in the javalib I've been working on recently are clearly work in the direction of this PR, but they're also valuable on their own, as all these changes provide better performance and better correctness for the JDK collections. But that's not high priority, of course.

sjrd:
This is now ready for review. The first commit belongs to #3720.

sjrd:
Rebased.

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    def hasErrors: Boolean = _errorCount != 0
+
+    def errorCount: Int = _errorCount
+  }
+
+  private final class IRFile(val path: Path, val tree: ClassDef)
+
+  private def listIRFiles(path: Path): Seq[IRFile] = {
```
gzm0: We might want to consider exposing this in `IRContainer` (maybe in `standard._` only).

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+      if (fileSize > Int.MaxValue.toLong)
+        throw new IOException(s"IR file too large: $path")
+      val buffer = ByteBuffer.allocate(fileSize.toInt)
+      channel.read(buffer)
+      buffer.flip()
+      Serializers.deserialize(buffer)
+    } finally {
+      channel.close()
+    }
+  }
```
gzm0: Can you use the existing methods in `linker._` for this? (`IRFile`, et al.)?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+      }
+    }
+
+    override def transform(tree: Tree, isStat: Boolean): Tree = {
+      implicit val pos = tree.pos
+
+      val preprocessedTree = tree match {
+        /* Rewrite `throw sjs.runtime.unwrapJavaScriptExeption(arg)` to
+         * `throw arg`, which is OK because the Java lang lib never throws
+         * wrapped JavaScript exceptions.
```
gzm0: Does anything validate this assumption?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+          AsInstanceOf(expr, transformTypeRef(typeRef))
+
+        case LoadJSConstructor(cls) =>
+          genLoadFromLoadSpecOf(cls)
+        case LoadJSModule(cls) =>
+          genLoadFromLoadSpecOf(cls)
+
+        case t: ClassOf =>
+          if (transformTypeRef(t.typeRef) != t.typeRef)
+            System.err.println(s"illegal Class(${t.typeRef}) at $pos")
```
gzm0: This should probably be reported differently? Also below.

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+
+    private def genLoadFromLoadSpecOf(cls: ClassRef)(
+        implicit pos: Position): Tree = {
+      jsTypes.get(cls.className) match {
+        case Some(classDef) =>
+          classDef.jsNativeLoadSpec match {
+            case Some(loadSpec) =>
+              genLoadFromLoadSpec(loadSpec)
+            case None =>
+              System.err.println(s"$cls does not have a load spec at $pos")
```
gzm0: So this means it is not a native class, right? Maybe be more explicit about this?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+      val sig = decodeMethodName(encodedName)
+      val newParamTypes = sig._2.map(transformTypeRef)
+      val newResultType = sig._3.map(transformTypeRef)
+      if (newParamTypes == sig._2 && newResultType == sig._3) {
+        ident
+      } else {
+        Ident(encodeMethodName(sig._1, newParamTypes, newResultType),
+            ident.originalName)(ident.pos)
+      }
+    }
```
gzm0: How do you make sure this does not generate two methods with the same signature?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+
+        case t: Apply =>
+          Apply(t.flags, t.receiver, transformMethodIdent(t.method),
+              t.args)(t.tpe)
+        case t:ApplyStatically =>
+          ApplyStatically(t.flags, t.receiver, transformClassRef(t.cls),
+              transformMethodIdent(t.method), t.args)(t.tpe)
+        case t: ApplyStatic =>
+          ApplyStatic(t.flags, transformClassRef(t.cls),
+              transformMethodIdent(t.method), t.args)(t.tpe)
```
gzm0: These two cases seem bad: If we apply a static method, we should fail, and not re-attach to Object. Or am I missing a case where this happens?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+
+        case NewArray(typeRef, lengths) =>
+          NewArray(transformArrayTypeRef(typeRef), lengths)
+        case ArrayValue(typeRef, elems) =>
+          ArrayValue(transformArrayTypeRef(typeRef), elems)
+
+        case IsInstanceOf(expr, typeRef) =>
+          IsInstanceOf(expr, transformTypeRef(typeRef))
```
gzm0: This is also bad: It breaks the semantics of `isInstnaceOf`. We should fail if we get a JS class here.

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+        implicit pos: Position): TypeRef = typeRef match {
+      case typeRef: ClassRef     => transformClassRef(typeRef)
+      case typeRef: ArrayTypeRef => transformArrayTypeRef(typeRef)
+    }
+
+    private def validateType(tpe: Type)(implicit pos: Position): Unit = {
+      tpe match {
+        case ClassType(cls) =>
+          validateClassName(cls)
+        case ArrayType(ArrayTypeRef(cls, _)) if isScalaClassName(cls) =>
```
gzm0: Unnecessary `isScalaClassName` guard?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,404 @@
...
+ *    `java.lang.Object`.
+ *  - Eagerly dereference `LoadJSConstructor` and `LoadJSModule` by "inlining"
+ *    the JS load spec of the mentioned class ref.
+ *
+ *  Afterwards, we check that the IR does not contain any reference to classes
+ *  undef the `scala.*` package, with the unfortunate exception of
+ *  `scala.scalajs.runtime.UndefinedBehaviorError`.
+ */
+object JavalibIRCleaner {
+  def cleanIR(libraryDir: File, javalibDir: File, outputDir: File,
```
gzm0: Isn't `libraryDir` here simply the compile dependency classpath? If yes, I think we should pass it as such (and use `IRContainer.fromPathClasspath` to load?).

Also: The API is inconsistent: It should either use all `File` or all `Path`.

project/JavalibIRCleaner.scala
sjrd: I have added code in `cleanClassDef` to check that.

project/JavalibIRCleaner.scala
sjrd: No it *is* a JS type (otherwise we wouldn't have gotten in the `case Some(classDef)`) but somehow it does not have a load spec. This shouldn't actually happen ever. It can only happen if the *compiler* did something wrong. I have expanded the error message to say so.

project/JavalibIRCleaner.scala
sjrd: Indeed. We should fail. I have fixed it.

project/JavalibIRCleaner.scala
sjrd: See my other comment: I would like to avoid introducing a dependency on `org.scalajs.linker` in the IR cleaner.

project/JavalibIRCleaner.scala
sjrd: I have fixed the API to take the dependency classpath, and return a `Set[File]`.

For `IRContainer.fromPathClasspath`, see my other comment: I would like to avoid introducing a dependency on `org.scalajs.linker` in the IR cleaner.

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,435 @@
...
+
+    val errorManager = new ErrorManager(logger)
+
+    val javalibDirPath = javalibDir.toPath()
+    val outputDirPath = outputDir.toPath()
+
+    val libraryIRFiles = dependencyClasspath.flatMap(f => listIRFiles(f.toPath()))
+    val javalibIRFiles = listIRFiles(javalibDirPath)
+
+    val jsTypes = analyzeJSTypes(libraryIRFiles ++ javalibIRFiles)
```
gzm0: Don't know why I didn't point this out before: I find the name `analyze` strange here. Maybe just `getJSTypes` or `filter`/`find` `JSTypes`? Nothing is really analyzed :)

javalanglib/src/main/scala/java/lang/SemanticsUtils.scala
```diff
@@ -0,0 +1,58 @@
...
+  def arrayIndexOutOfBoundsCheck(shouldThrow: js.Function0[scala.Boolean],
+      exception: js.Function0[ArrayIndexOutOfBoundsException]): Unit = {
+    genericCheck(linkingInfo.semantics.arrayIndexOutOfBounds, shouldThrow,
+        exception)
+  }
+
+  @inline
+  private def genericCheck(complianceLevel: Int,
+      shouldThrow: js.Function0[scala.Boolean],
+      exception: js.Function0[Throwable]): Unit = {
```
gzm0: Hum, stupid idea: Can we just type this as `js.Function0[RuntimeException]` and get rid of the unwrap exception removal?

project/JavalibIRCleaner.scala
```diff
@@ -0,0 +1,435 @@
...
+        val seenMethodNames = mutable.Set.empty[(MemberNamespace, String)]
+        for (m <- newMemberDefs) {
+          if (m.isInstanceOf[MethodDef]) {
+            if (!seenMethodNames.add((m.flags.namespace, m.name.encodedName))) {
+              reportError(
+                  s"duplicate method name ${m.name.encodedName} after erasure")(
+                  m.pos)
+            }
+          }
+        }
```
gzm0: Hum, is it me or is this pre-transform of the methods? Or am I reading this wrong?

javalanglib/src/main/scala/java/lang/SemanticsUtils.scala
sjrd: I thought about that, but it doesn't work, unfortunately, because `JavaScriptException` is also a subclass of `RuntimeException`. :-(

project/JavalibIRCleaner.scala
sjrd: It iterates over `newMemberDefs` doesn't it?

sjrd: Er ... Yes, you're right -_-'

sjrd:
Updated.

javalanglib/src/main/scala/java/lang/SemanticsUtils.scala
gzm0: ... I have an even more stupid idea: Can we simply move the throw at the call site to help the type checker? We can easily guard against forgetting to write `throw` by requiring the return type to be `Nothing`.

sjrd: I tried that, but I could only get it to work by adding a temporary `val`, and in that case the produced code was worse. :(

gzm0: Ah, yes. I didn't notice it is used twice.

PR #3719, opened 2019-07-10
Do not eagerly throw for `Nothing` in `Class.cast()`.

sjrd:
There is no need to, as the compiler will insert an explicit cast at call site to adapt the erased `A` (`Object`) to `Nothing` (assuming the call is not in statement position).

The test was a bit too restrictive, and has been relaxed not to expect a specific error message, and to only expect an exception when used in expression position.

The new behavior is consistent with Scala/JVM, since of course, on the JVM, `Class.cast()` does not know about `scala.runtime.Nothing$`.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypesTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypesTest.scala
@@ -64,13 +64,12 @@ class RuntimeTypesTest {
     assumeTrue("Assumed compliant asInstanceOf", hasCompliantAsInstanceOfs)
     def test(x: Any): Unit = {
       try {
-        classOf[Nothing].cast(x)
+        val result = classOf[Nothing].cast(x)
         fail("casting " + x + " to Nothing did not fail")
+        identity(result) // discard `result` without warning
       } catch {
-        case th: Throwable =>
-          assertTrue(th.isInstanceOf[ClassCastException])
-          assertEquals(x + " is not an instance of scala.runtime.Nothing$",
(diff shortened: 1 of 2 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypesTest.scala
```diff
@@ -64,13 +64,12 @@ class RuntimeTypesTest {
...
+        val result = classOf[Nothing].cast(x)
         fail("casting " + x + " to Nothing did not fail")
+        identity(result) // discard `result` without warning
       } catch {
-        case th: Throwable =>
-          assertTrue(th.isInstanceOf[ClassCastException])
-          assertEquals(x + " is not an instance of scala.runtime.Nothing$",
-              th.getMessage)
+        case th: ClassCastException =>
+          // ok
```
gzm0: If this is only about the type, should we use `ExpectedException` on the test?

sjrd: If you mean `assertThrows`, that doesn't quite work in this case. Because we need to test that the exception is thrown at the very point of the assignment to `result`, and not any later. If we do a simple `assertThrows`, we cannot check that. In particular we cannot check that the error is thrown before the statement with `identity`.

gzm0: No, I meant JUnit's `ExpectedException`. But indeed, that's a fair point. I didn't get the subtlety about the assignment being necessary.

PR #3720, opened 2019-07-10
Do not check formal arguments of method calls to non-instantiated types.

sjrd:
In a method call of the form `x.m__sig(...args)`, if the type of `x` does not have instances, the base linker might not have linked the classes referenced in `sig` at all. Therefore, we must not try to infer the type signature, as it can crash.

### DIFF
--- a/linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
+++ b/linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
@@ -1,0 +1,106 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker
+
+import scala.concurrent._
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.ir.ClassKind
+import org.scalajs.ir.Definitions._
+import org.scalajs.ir.Trees._
+import org.scalajs.ir.Types._
+
+import org.scalajs.logging._
+
+import org.scalajs.junit.async._
+
+import org.scalajs.linker._
+import org.scalajs.linker.standard._
+
+import org.scalajs.linker.testutils._
+import org.scalajs.linker.testutils.TestIRBuilder._
+
+class IRCheckerTest {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  import IRCheckerTest._
+
+  @Test
+  def testMethodCallOnClassWithNoInstances(): AsyncResult = await {
+    def callMethOn(receiver: Tree): Tree =
+      Apply(EAF, receiver, Ident("meth__LFoo__V"), List(Null()))(NoType)
+
+    val classDefs = Seq(
+        // LFoo will be dropped by base linking
+        classDef("LFoo", superClass = Some(ObjectClass)),
+
+        classDef("LBar",
+            superClass = Some(ObjectClass),
+            memberDefs = List(
+                trivialCtor("LBar"),
+
+                /* This method is called, but unreachable because there are no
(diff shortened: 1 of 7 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -332,6 +332,26 @@ class RegressionTest {
...
+    }
+
+    @noinline def nullBar(): Bar = null
+
+    // the IR checker must not try to infer the signature of these calls
+    nullBar().meth(null)
+    (null: Bar).meth(null)
+    (??? : Bar).meth(null) // scalastyle:ignore
+  }
+
```
gzm0: Could you also add a unit test for this?

sjrd: Er, isn't *that* a unit test? Do you mean a test in `linker/shared/test`?

gzm0: Yes, I meant a test in `linker/shared/test`? Unclear whether the test-suite is unit tests, it really depends on your definition of unit (but doesn't really matter, I meant to ask for a linker test :)).

sjrd: OK. I'll do that when I come back from vacation, probably, because it's going to be annoying to write on mobile ^^

sjrd: Done.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
```diff
@@ -1,15 +1,15 @@
...
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
```
gzm0: Hum, what happened here?

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
```diff
@@ -57,4 +57,7 @@ object TestIRBuilder {
     MethodDef(MemberFlags.empty, Ident("main__AT__V"), List(argsParamDef),
         NoType, Some(body))(emptyOptHints, None)
   }
-}
+
+  def paramDef(name: String, ptpe: Type): ParamDef =
```
gzm0: Use this in `mainMethodDef`?

linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
```diff
@@ -0,0 +1,119 @@
...
+                trivialCtor("LTest$"),
+                MethodDef(MemberFlags.empty, Ident("nullBar__LBar"), Nil, ClassType("LBar"),
+                    Some(Null()))(
+                    emptyOptHints, None),
+                MethodDef(MemberFlags.empty, Ident("test__V"), Nil, NoType, Some(Block(
+                    callMethOn(Apply(EAF, This()(ClassType("LTest$")),
+                        Ident("nullBar__LBar"), Nil)(ClassType("LBar"))),
+                    callMethOn(Null()),
+                    callMethOn(Throw(Null()))
+                )))(emptyOptHints, None)
```
gzm0: Consider using `mainMethodDef` for simplicity.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
sjrd: Somehow the header used Windows line endings before (but only the header). I don't how that happened, nor why `headerCheck` does not complain about that. Now they're fixed to being UNIX line endings.

PR #3722, opened 2019-07-13
Fix #3069: Upgrade to sbt-1.0, drop sbt 0.13, Scala 2.10

### LINKED ISSUE #3069: Upgrade our own build to sbt 1.x, when it's good enough
This should only be done on master, not in 0.6.x. At the same time, we should drop the cross-building of the sbt plugin, and only publish it for sbt 1.x.

Definitively migrating to sbt 1.0 implies dropping support for JDK 6 and 7, since sbt 1 requires JDK 8.

According to release notes, this might also require dropping support for Scala 2.11.0 and 2.11.1. sbt 1 does not support them.

See also: http://www.scala-sbt.org/1.0/docs/sbt-1.0-Release-Notes.html#sbt+1.0.0-RC2

gzm0:


gzm0:
There seems to be a caching issue with `jl_Object`. I had the same on my local machine. Erasing the relevant `target/` directories helped. @sjrd do you know how to do that for both CIs?

sjrd:
Normally the CI always starts from a freshly cloned repo, from an empty directory.

Otherwise you'd need to ssh into the machines. Fabien can give you access of you don't have it.

gzm0:
OK. I can reproduce this locally:

```
$ git -ffdx clean
$ sbt
> helloworld/fastOptJS
[error] Fatal error: java.lang.Object is missing
> helloworld/fastOptJS
[success] ...
```

:S

gzm0:
Ok, I can reproduce this locally:

```
$ git clean -ffdx
$ sbt
> helloworld/fastOptJS
[error] Fatal error: java.lang.Object is missing
> helloworld/fastOptJS
[success] ...
```

gzm0:
I can indeed trace this to the library mappings `show library/compile:packageBin::mappings`.

This does not show `jl_Object` after the first run after a clean, and does show it the second time.

gzm0:
I'm unable to find a mistake in the sbt graph (which would cause a race condition in what is put into `target/scala-X/classes`). Needs further investigation.

gzm0:
It seems that `javalanglib/products` does not copy `Object.sjsir` into `classes`.

gzm0:
Disabling external compile seems to fix the issue.

gzm0:
I suspect [this line][1] is to blame since it assumes only the compile task writes to this directory, which is not true `copyResources` also does.

[1]: https://github.com/scala-js/scala-js/blob/2dceb67560ab13a7ca8d924f7acc731fdfd24528/project/ExternalCompile.scala#L107

gzm0:
More importantly: https://github.com/scala-js/scala-js/blob/2dceb67560ab13a7ca8d924f7acc731fdfd24528/project/ExternalCompile.scala#L60-L62

gzm0:
Updated.

gzm0:
Seems like sbt enforces that `scalaCrossVersions` contains the value we'll set with `++`. We seem to have two options:

- List the values:  a bit annoying, since for the compiler plugin we'll need different ones.
- Use `++ <version> !`: a bit ugly.

@sjrd opinion?

sjrd:
Listing the values seems fine. It should be the same full list of minor versions for all projects, actually. So it shouldn't be different for compiler plugins. It's only different for the sbt plugin.

gzm0:
Jenkins, retest this please

sjrd:
That command doesn't work anymore. You have to click the 🔄 icon in the Jenkins interface instead.

project/ExternalCompile.scala
```diff
@@ -103,14 +107,18 @@ object ExternalCompile {
...
+          // Output is all files in outputDirectory
+          (outputDirectory ** AllPassFilter).get.toSet
         }
 
-        cachedCompile((sources ++ allMyDependencies).toSet)
+        val output = cachedCompile((sources ++ allMyDependencies).toSet)
+        val mapper = Path.rebase(outputDirectory, classesDirectory)
+        val mappings = output.flatMap(f => mapper(f).map(f -> _))
+
+        Sync.sync(s.cacheStoreFactory.make("compile-copy"))(mappings)
```
sjrd: Did you consider putting everything up to and including the can to `Sync.sync` inside the `cachedCompile`? It seems that this should be correct as well, and avoid useless work in `sync` when nothing's changed.

DEVELOPING.md
```diff
@@ -184,5 +184,5 @@ following incantations.
 
     > ++SCALA_VERSION
     > ;compiler/publishLocal;library/publishLocal;testInterface/publishLocal;testBridge/publishLocal;jUnitRuntime/publishLocal;jUnitPlugin/publishLocal
-    > ++2.10.7
+    > ++2.12.8
```
sjrd: 2.12.8 is the default, so that shouldn't even be necessary.

gzm0: Hum, but there is a `++SCALA_VERSION` before it. Or did you want me to change the order?

project/ExternalCompile.scala
gzm0: Indeed. Fixed (even simplified some stuff).

DEVELOPING.md
sjrd: Oh right. I missed that. No never mind, then.

SethTisue:
YESSSSSSS

SethTisue:
now drop 2.11 :-)

sjrd:
@SethTisue That's probably not very realistic at the moment. Too many core libraries still support 2.11.

ramnivas:
@sjrd Assuming removing 2.11 support will help move Scala.js faster, it may be a good idea to survey Scala.js users to see actual usage of 2.11. I find it hard to believe that developers that use Scala.js also use 2.11.

sjrd:
> Assuming removing 2.11 support will help move Scala.js faster,

It won't. As I said above:

> The only real benefit is getting rid of our hacky `@JavaDefaultMethod` annotation.

It's a common misconception that keeping support for old versions of Scala is holding us back. It's simply not true.

dwijnand:
> It's a common misconception that keeping support for old versions of Scala is holding us back. It's simply not true.

The cost of maintaining and waiting for the execution of CI for old versions of Scala is non-negligible.

sjrd:
To some extent, the CI time is a concern, yes. But there are 6 or 7 servers at EPFL dedicated to running the CI for this repo only, so it's OK.

PR #3723, opened 2019-07-14
Avoid constructors of primitive boxes

gzm0:
They are deprecated in Java 9.


sjrd:
It looks like this PR should get 0.6.x, shouldn't it?

Otherwise LGTM.

gzm0:
> It looks like this PR should target 0.6.x, shouldn't it?

Indeed. Done.

PR #3724, opened 2019-07-17
Fix #3697: Improve error message when no main module initializer was found

### LINKED ISSUE #3697: Improve error message when no main module initializer was found
Currently it says: "No main module initializer was specified (possibly because no or multiple main classes were found) [...]".

We should report the classes that were found for easier debugging.

exoego:
Closes #3697

Mentioning `mainClass in Compile := ...` as suggested in https://github.com/scala-js/scala-js/issues/3697#issuecomment-501726873 are not included, since right now I think `mainClass := ...` is sufficient for most use-cases.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -909,15 +909,23 @@ object ScalaJSPluginInternal {
         (scalaJSModuleInitializers in (This, Zero, This)).value,
 
       scalaJSModuleInitializers ++= {
+        val mainClasses = scalaJSDiscoveredMainClasses.value
         if (scalaJSUseMainModuleInitializer.value) {
           Seq(scalaJSMainModuleInitializer.value.getOrElse {
-            throw new MessageOnlyException(
-                "No main module initializer was specified (possibly because " +
-                "no or multiple main classes were found), but " +
-                "scalaJSUseMainModuleInitializer was set to true. " +
-                "You can explicitly specify it either with " +
-                "`mainClass := Some(...)` or with " +
-                "`scalaJSMainModuleInitializer := Some(...)`")
+            if (mainClasses.isEmpty) {
+              throw new MessageOnlyException(
+                  "No main module initializer was specified, but " +
+                  "scalaJSUseMainModuleInitializer was set to true. " +
+                  "You can explicitly specify it either with " +
+                  "`mainClass := Some(...)` or with " +
+                  "`scalaJSMainModuleInitializer := Some(...)`")
+            } else {
+              throw new MessageOnlyException(
+                  s"Multiple main classes (${mainClasses.keys}) were found. " +
+                  "You can explicitly specify the one you want with " +
+                  "`mainClass := Some(...)` or with " +
+                  "`scalaJSMainModuleInitializer := Some(...)`")
+            }
           })
         } else {
           Seq.empty


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -908,21 +908,35 @@ object ScalaJSPluginInternal {
       scalaJSModuleInitializers :=
         (scalaJSModuleInitializers in (This, Zero, This)).value,
 
-      scalaJSModuleInitializers ++= {
+      scalaJSModuleInitializers ++= Def.taskDyn {
```
sjrd: Is that `taskDyn` really necessary? Forcing evaluation of `scalaJSDiscoveredMainClasses` even when `scalaJSUseMainModuleInitializer` is false shouldn't be an issue, I think.

If it's not necessary I would recommend not to use `taskDyn`, as each one deteriorates the debuggability of builds through things like `inspect tree`.

It is is absolutely necessary for correctness, at least use `Def.settingDyn` instead. This is possible because the dynamic part relies only on `scalaJSUseMainModuleInitializer` which is a setting.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -908,21 +908,35 @@ object ScalaJSPluginInternal {
...
-                "No main module initializer was specified (possibly because " +
-                "no or multiple main classes were found), but " +
-                "scalaJSUseMainModuleInitializer was set to true. " +
-                "You can explicitly specify it either with " +
-                "`mainClass := Some(...)` or with " +
-                "`scalaJSMainModuleInitializer := Some(...)`")
-          })
+          Def.task {
+            Seq(scalaJSMainModuleInitializer.value.getOrElse {
+              scalaJSDiscoveredMainClasses.value.keys.toSeq match {
```
sjrd: Consider using
```scala
if (scalaJSDiscoveredMainClasses.value.isEmpty)
```
instead of `.keys.toSeq match { ... }`

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -908,21 +908,35 @@ object ScalaJSPluginInternal {
...
+                case Seq() =>
+                  throw new MessageOnlyException(
+                    """No main module initializer was specified, but
+                      | scalaJSUseMainModuleInitializer was set to true.
+                      | You can explicitly specify it either with
+                      | `mainClass := Some(...)` or with
+                      | `scalaJSMainModuleInitializer := Some(...)`"""
+                      .stripMargin)
```
sjrd: `stripMargin` will remove the spaces and the `|` but will keep the newlines. Consider keeping the old style of leaving the error message as one line, letting the user's terminal wrap around where necessary.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -908,21 +908,35 @@ object ScalaJSPluginInternal {
...
+                case multiple =>
+                  throw new MessageOnlyException(
+                    s"""Multiple main classes (${multiple}) were found.
+                       | Main class should be specified exact once either with
```
sjrd: I'm not quite convinced by the English formulation here. It could be my English (I'm not a native speaker). Perhaps consider:
```suggestion
                       | You can explicitly specify the one you want with
```
(it's more in line with the other error message)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/3724/commits/3a78772b5bc3ad9576b894710c5759fe755f9cf2

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/3724/commits/eaa336487bd07a55b9bd8e5ed9311373157581ff

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
exoego: Applied suggestion in https://github.com/scala-js/scala-js/pull/3724/commits/eaa336487bd07a55b9bd8e5ed9311373157581ff

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
exoego: Honestly, I was unfamiliar difference between `taskDyn` and `settingDyn`.
Fixed in https://github.com/scala-js/scala-js/pull/3724/commits/3cec2d48f6bdff829ea1202a3f87f0f719b28304

sjrd: Did you try without using any `Dyn` at all?

exoego: Yes, tried in https://github.com/scala-js/scala-js/commit/d6e9944923dff56411d179d3fd38cb86bd7661a3
But it turned out that `Dyn` is required since the referenced `scalaJSDiscoveredMainClasses` are `TaskKey`.

sjrd: Hum, what do you mean? Was it just that the macro complained that you can't access a task key's value in an if or lambda? Because that can be dealt with easily by storing `scalaJSDiscoveredMainClasses.value` in a `val` first, before doing the test `if (scalaJSUseMainModuleInitializer.value) {`. That's the normal way to resolve this issue.

Using Dyns is an entirely different kind of monster, that must be used only when the tasks accessed in the dynamic parts *must not* be evaluated when the test fails. It should be very rare. Unfortunately in 0.6.x the plugin has many of them because we took bad design decisions, but if you look at the 1.x plugin, I believe there is only one: the thing that decides whether to evaluate `fastOptJS` or `fullOptJS` based on `scalaJSStage`. That one is of course critical, because if not done properly, every invocation of `test` would evaluate `fullOptJS`! Here we don't care whether `scalaJSDiscoveredMainClasses` is evaluated even if it's not used.

exoego: > Was it just that the macro complained that you can't access a task key's value in an if or lambda?

Yes, I think it was.
Removed `Dyn` in https://github.com/scala-js/scala-js/pull/3724/commits/11e8dc52c613ee8cab13bb7f2df8f2053dc4b7d3

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -909,19 +909,24 @@ object ScalaJSPluginInternal {
...
       scalaJSModuleInitializers ++= {
+        val mainClasses = scalaJSDiscoveredMainClasses.value
         if (scalaJSUseMainModuleInitializer.value) {
           Seq(scalaJSMainModuleInitializer.value.getOrElse {
-            throw new MessageOnlyException(
-                "No main module initializer was specified (possibly because " +
-                "no or multiple main classes were found), but " +
+            if (mainClasses.isEmpty)
+              throw new MessageOnlyException(
+                "No main module initializer was specified, but " +
```
sjrd: This line and the following should be indented with 4 spaces (hence 2 more).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -909,19 +909,24 @@ object ScalaJSPluginInternal {
...
                 "scalaJSUseMainModuleInitializer was set to true. " +
                 "You can explicitly specify it either with " +
                 "`mainClass := Some(...)` or with " +
                 "`scalaJSMainModuleInitializer := Some(...)`")
+            else
+              throw new MessageOnlyException(
+                s"Multiple main classes (${mainClasses.keys}) were found. " +
```
sjrd: This line and the following should be indented with 4 spaces (hence 2 more).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -909,19 +909,24 @@ object ScalaJSPluginInternal {
...
+                "You can explicitly specify the one you want with " +
+                "`mainClass := Some(...)` or with " +
+                "`scalaJSMainModuleInitializer := Some(...)`")
           })
-        } else {
-          Seq.empty
-        }
+        } else Seq.empty
```
sjrd: The else should be formatted as before: with `{}` and with `Seq.empty` on a separate line, according to our coding style.

sjrd:
Cool thanks. Could you squash all the commits in one, with the commit message of the first one? Then we can merge.

(The CI failure is spurious.)

exoego:
Squashed commits.

PR #3725, opened 2019-07-17
Fix #2971: Ensure FunctionEmitter synthetic vars don't name clash

### LINKED ISSUE #2971: Ensure FunctionEmitter synthetic vars don't name clash
This is a follow-up from #2800.

exoego:
Closes #2971


sjrd:
I am pretty sure the proposed fix doesn't work. Did you have a reproduction for the first issue that was fixed by your commit? What if I declare a *local* `val jsx$1` and also some code that will generate a temporary `jsx$1` in the same method? I believe `transformLocalName` will interpret them as the same local val, and hence produce the but anyway.

Similarly, if there's a *local* `$thiz`, the second commit won't help, I believe.

exoego:
> I am pretty sure the proposed fix doesn't work. Did you have a reproduction for the first issue that was fixed by your commit? What if I declare a _local_ `val jsx$1` and also some code that will generate a temporary `jsx$1` in the same method? 

Sorry, added a reproduction test (which fails right now) and updated the changes.s

> Similarly, if there's a _local_ `$thiz`, the second commit won't help, I believe.

Removed the second comitt so this PR will focus #2971 only.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -357,12 +357,12 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
...
-    def newSyntheticVar()(implicit pos: Position): Ident = {
-      /* TODO Integrate this with proper name management.
-       * This is filed as #2971.
-       */
+    @tailrec
+    private def newSyntheticVar()(implicit pos: Position): Ident = {
       syntheticVarCounter += 1
-      Ident("jsx$" + syntheticVarCounter, None)
+      val varName = "jsx$" + syntheticVarCounter
+      if (localVarNames.contains(varName)) newSyntheticVar()
```
sjrd: What if the user-defined local var is encountered *after* we already decided to allocate `jsx$1` to a synthetic one?

exoego: Do you mean something like this?

```scala
while (idx >= 0 && name.charAt(idx) != '$') {
  idx -= 1
  var `jsx$1` = "" + name
  buf += `jsx$1`
}
```

If so, this case passes in this PR.

exoego: The below is fine too.
```scala
while (idx >= 0 && name.charAt(idx) != '$') {
  idx -= 1
  var `jsx$1` = "" + name // user-defined `jsx$1` after synthetic one
  var idx2 = name.length - 2
  while (idx2 >= 0 && name.charAt(idx2) != '$') {
    idx2 -= 1
    buf += `jsx$1`
  }
}
```

sjrd:
Hum, I don't understand how that works :s How come the user-defined local var doesn't clash with the synthetic one, if it comes after? The transformation of the local var wouldn't know to avoid the synthetic one, since it wasn't recorded. What do you think is the data flow that enables it to work?

exoego:
Though I did not look deeply, I guess that the data flow enables it: `newSyntheticVar` are peformed after the transformation of the local var.

sjrd:
Hum, I can't convince myself right now, just on mobile. I'll have to properly review this on a computer, when I'm back from vacation.

sjrd:
Here is a test that breaks the current solution:
```scala
    @noinline def f3(strs: Array[String]): String = {
      val x = if (strs.length > 3 && {
        val y = strs.length
        y / 2 > y - 3
      }) {
        "one"
      } else {
        "two"
      }
      val `jsx$1` = "" + x + "-" + "foo"
      `jsx$1` + "-bar"
    }
    assertEquals("two-foo-bar", f3(Array("one", "two", "three", "four", "five")))
```
In this case, `newSyntheticVar()` is called a first time *before* we reach the user-level definition of `jsx$1`.

exoego:
Why Scala.js does not rename user-defined value name in back quotes?

sjrd:
They wouldn't help, because what would it rename them *to*? You'd have to pick *some* scheme, and that scheme could end up clashing with other stuff.

This is the normal way to do things. You rename only when necessary, when you know what to avoid clashes with.

(Anyway, you can also declare `val jsx$1` without quotes in Scala.)

exoego:
I have no idea to proceed this in one-pass.

sjrd:
> Use Assign if use-defined var decl crush with syntactic var 

Now you're piling a hack on top of the problem. This is not going to end well. For example some temp vars are `const`, which cannot be assigned to a second time. And I wouldn't exclude a situation where the lifetime of a temp var intersects with the lifetime of a user variable with the same name, which would be extremely problematic.

I think you need to take a step back and approach the problem differently from the start. It may be possible to solve this in a principled way by reusing the optimistic and pessimistic run infrastructure.

exoego:
Closing since I no longer work on this.

PR #3734, opened 2019-08-01
Implement `java.util.TreeSet` from scratch.

sjrd:
Previously, it was implemented on top of `s.c.m.TreeSet`, with various hacks to implement JDK features:

* `Iterator.remove()`
* Descending iterators and views
* Projections with exclusive start and/or inclusive end

All those hacks resulted in significantly more memory allocation (at least one `Box` for most elementary operations) as well as unreasonable time complexity for views.

The new implementation, done from scratch, has direct internal support for all these operations:

* It never allocates during lookup and removal
* Iteration never allocates besides the instance of `Iterator`
* It allocates at most one object during `add`

The implementation was initially inspired by the implementation of `s.c.m.TreeMap` in Scala 2.13.0 (in particular the core data structure and algorithms in `RedBlackTree.scala`), but was heavily changed to accommodate the requirements of the JDK API.

The red-black tree implementation supports values associated to keys, although `TreeSet` does not use that feature. It will be useful when we eventually implement `TreeMap`. The original red-black tree in Scala 2.13.0 supported it, so it was easier to port it than not.


javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -12,245 +12,455 @@
...
+      NaturalComparator.unselect(comp)
+
+    def first(): E = {
+      val node = RB.minNodeAfter(tree, fromElement, fromBoundKind)
+      if (node eq null)
+        throw new NoSuchElementException()
+      val key = node.key
+      if (!isWithinUpperBound(key))
+        throw new NoSuchElementException()
+      key
```
gzm0: Use:

```scala
val key = ifWithinUpperBound(RB.minKeyAfter(tree, fromElement, fromBoundKind)
if (key eq null)
  throw NSE
key
```

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -12,245 +12,455 @@
...
-        None, true)
-  }
+    def last(): E = {
+      val node = RB.minNodeAfter(tree, toElement, toBoundKind)
+      if (node eq null)
+        throw new NoSuchElementException()
+      val key = node.key
+      if (!isWithinUpperBound(key))
+        throw new NoSuchElementException()
+      key
```
gzm0: Same comment about using `ifWithinXBound`.

javalib/src/main/scala/java/util/TreeSet.scala
```diff
@@ -12,245 +12,455 @@
...
 
-  def descendingIterator(): Iterator[E] = {
-    new Iterator[E] {
-      private val iter = inner.iterator.toList.reverse.iterator
+  override def clone(): TreeSet[E] =
+    new TreeSet(tree.treeCopy())(comp)
+}
+
+private object TreeSet {
+  private final class Projection[E](
```
gzm0: Have you considered sharing the parts of the order independent implementation? (some of the naming might need to change from `upper/lower` to `to/from`).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentSkipListSetTest.scala
```diff
@@ -404,8 +404,11 @@ class ConcurrentSkipListSetTest {
     val csls = new ConcurrentSkipListSet[TestObj]()
 
     assertEquals(0, csls.size())
-    csls.add(new TestObj(111))
-    expectThrows(classOf[ClassCastException], csls.add(new TestObj(222)))
+    expectThrows(classOf[ClassCastException], {
+      // Throw either when the first or second element is added
+      csls.add(new TestObj(111))
+      csls.add(new TestObj(222))
+    })
```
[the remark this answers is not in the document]
sjrd: The purpose of the weird code removed in the commit was specifically to allow the first addition not to throw in a `ConcurrentSkipListSet`, even though `TreeSet` does throw in that situation.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,957 @@
...
+      null.asInstanceOf[B]
+    } else {
+      deleteNode(tree, z)
+      z.value
+    }
+  }
+
+  def deleteNode[A, B](tree: Tree[A, B], z: Node[A, B]): Unit = {
+    var y = z
+    var yIsRed = y.red
```
gzm0: This variable is completely pointless: It is updated through the algorithm but only use at the end once and readily available as `y.red`.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,957 @@
...
+      } else {
+        xParent = y.parent
+        transplant(tree, y, y.right)
+        y.right = z.right
+        y.right.parent = y
+      }
+      transplant(tree, z, y)
+      y.left = z.left
+      y.left.parent = y
+      y.red = z.red
```
gzm0: I have abandoned reviewing this, due to the lack of reasonable names. Consider Please use semantic names like, for example: `replacement` (node we put in place of the node to delete), `orphan` (node that has just lost its parent because it was moved).

gzm0:
It seems like a large part of the `RedBlackTree` implementation was copied from Scala and is not at the level of quality I expect in the Scala.js repo. Does that match your impression / observation?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentSkipListSetTest.scala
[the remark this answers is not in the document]
gzm0: I see. But why should this behavior be different for `TreeSet`? It seems that the `TreeSet` implementation is buggy: JavaDoc says: "Throws ClassCastExcpetion: if the specified object cannot be compared with the elements currently in this set".

So if there is no element, this probably shouldn't throw. So it seems to me like the test is correct, the `TreeSet` implementation is not.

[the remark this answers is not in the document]
sjrd: The `TreeSet` on the JVM throws when inserting the first element. I thought aligning the behaviors for `TreeSet` was more important than aligning that for `ConcurrentSkipListSet`. But we can also decide not to align `TreeSet` with the behavior on the JVM, and instead always throw only on the second insertion. It's less expensive.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,957 @@
...
+ *  `scala.collection.mutable.RedBlackTree` as found in Scala 2.13.0.
+ */
+private[util] object RedBlackTree {
+
+  // ---- bounds helpers ----
+
+  type BoundKind = Int
+
+  final val InclusiveBound = 0 // This value is meaningful!
+  final val ExclusiveBound = 1 // This value is meaningful!
```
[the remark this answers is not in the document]
sjrd: Ah. It's supposed to mean that the value 1 is used in the code. It cannot be replaced by another value lest it break some of the algorithms. I'll expand the comment.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,957 @@
...
+    def leaf[A, B](key: A, value: B, red: Boolean, parent: Node[A, B]): Node[A, B] =
+      new Node(key, value, red, null, null, parent)
+  }
+
+  // ---- comparator helper ----
+
+  @inline
+  private[this] def compare[A](key1: Any, key2: A)(
+      implicit comp: Comparator[_ >: A]): Int = {
+    comp.compare(key1.asInstanceOf[A], key2)
```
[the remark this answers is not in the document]
sjrd: It will. So will trying the same snippet on the JVM. (Well not UB but a ClassCastException.)

sjrd:
Yes it was indeed copied in large swaths from the Scala repo. It's not the same quality, but I saw value in keeping the names and algorithms the same. That way, if a bug is discovered and fixed in one of the two implementations, it would be easier to transfer it to the other one. Or should we abandon the similarity and focus only on our implementation as an independent code?

sjrd:
Yes it was indeed copied in large swaths from the Scala repo. It's not the same quality, but I saw value in keeping the names and algorithms the same. That way, if a bug is discovered and fixed in one of the two implementations, it would be easier to transfer it to the other one. Or should we abandon the similarity and focus only on our implementation as an independent code?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentSkipListSetTest.scala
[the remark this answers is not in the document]
sjrd: This was specifically done as a fix in JDK 7, according to the changelog at https://www.oracle.com/technetwork/java/javase/compatibility-417013.html#behavioral (look for `5045147` on the page). So I think we should keep the current behavior of our `TreeSet` to eagerly throw when we try to insert an invalid element in an empty set.

sjrd:
I have addressed all the comments that do not involve renaming variables nor making algorithms tail-recursive in `RedBlackTree`. I'll defer those changes until after we've decided whether intrinsic quality or similarity with the upstream implementation is preferable.

sjrd:
> Yes it was indeed copied in large swaths from the Scala repo. It's not the same quality, but I saw value in keeping the names and algorithms the same. That way, if a bug is discovered and fixed in one of the two implementations, it would be easier to transfer it to the other one. Or should we abandon the similarity and focus only on our implementation as an independent code?

@gzm0 Should I try to simplify/improve the implementation, at the expense of deviating from the upstream implementation?

gzm0:
> Should I try to simplify/improve the implementation, at the expense of deviating from the upstream implementation?

I have been thinking about this quite a bit, and I don't think there is a lot of value of keeping the implementations similar. I think the best example are overrides: In theory we should keep them up to date with upstream changes where appropriate; I think in practice this has proven to be very difficult.

So I think sharing code only seems to be valuable if we actually use the same code (by simply pulling individual source files). To continue the overrides analogy, this strategy has worked well for the rest of the scalalib.

Long story short, yes, I think we should improve the implementation if sharing code exactly is infeasible.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+  def fromOrderedKeys[A](xs: Iterator[A], size: Int): Tree[A, Any] = {
+    val maxUsedDepth = 32 - Integer.numberOfLeadingZeros(size) // maximum depth of non-leaf nodes
+
+    def f(level: Int, size: Int): Node[A, Any] = size match {
+      case 0 =>
+        null
+      case 1 =>
+        new Node(xs.next(), (), level == maxUsedDepth && level != 1, null, null, null)
+      case n =>
+        val leftSize = (size - 1) >> 1 // (size - 1) / 2
```
gzm0: Why is the bit fiddling necessary?

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+    def nextResult(node: Node[A, B]): B = node.value
+  }
+
+  // building
+
+  /** Build a Tree suitable for a TreeSet from an ordered sequence of keys. */
+  def fromOrderedKeys[A](xs: Iterator[A], size: Int): Tree[A, Any] = {
+    val maxUsedDepth = 32 - Integer.numberOfLeadingZeros(size) // maximum depth of non-leaf nodes
+
+    def f(level: Int, size: Int): Node[A, Any] = size match {
```
gzm0: call this `createSubtree`?

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+      case 0 =>
+        null
+      case 1 =>
+        new Node(xs.next(), (), level == maxUsedDepth && level != 1, null, null, null)
```
gzm0: What is the advantage of painting these red? Also, either name the argument or put it in a helper to make it clear what it is?

gzm0: Also, use `Node.leaf`.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+      case n =>
+        val leftSize = (size - 1) >> 1 // (size - 1) / 2
+        val left = f(level + 1, leftSize)
+        val x = xs.next()
+        val right = f(level + 1, size - 1 - leftSize)
+        val n = new Node(x, (), false, left, right, null)
```
gzm0: use `Node.apply`.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+      start: A, startKind: BoundKind, end: A, endKind: BoundKind)(
+      implicit comp: Comparator[_ >: A])
+      extends DescendingTreeIterator[A, B, B](tree, start, startKind, end, endKind) {
+
+    def nextResult(node: Node[A, B]): B = node.value
+  }
+
+  // building
+
+  /** Build a Tree suitable for a TreeSet from an ordered sequence of keys. */
```
gzm0: This should at least have a comment saying that unit is used as value. Maybe even take it as argument.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+        val right = f(level + 1, size - 1 - leftSize)
+        val n = new Node(entry.getKey(), entry.getValue(), false, left, right, null)
+        if (left ne null)
+          left.parent = n
+        right.parent = n
+        n
+    }
+
+    new Tree(f(1, size), size)
+  }
```
gzm0: You should share these implementations. Use an accessor function + @inline and it will be zero overhead.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1033 @@
...
+
+  private def copyTree[A, B](n: Node[A, B]): Node[A, B] = {
+    if (n eq null) {
+      null
+    } else {
+      val c = new Node(n.key, n.value, n.red, copyTree(n.left), copyTree(n.right), null)
```
gzm0: Node.apply?

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1019 @@
...
+
+  private def isRed(node: Node[_, _]): Boolean = (node ne null) && node.red
+
+  private def isBlack(node: Node[_, _]): Boolean = (node eq null) || !node.red
+
+  // A red node is non-null and not the root, so node.parent exists and is not null
+  private def redIsLeft(node: Node[_, _]): Boolean = node eq node.parent.left
+
+  // A red node is non-null and not the root, so node.parent exists and is not null
+  private def redIsRight(node: Node[_, _]): Boolean = node eq node.parent.right
```
gzm0: These two helpers should go away now and be replaced by `node.isLeftChild` and `node.isRightChild` respectively.

javalib/src/main/scala/java/util/RedBlackTree.scala
```diff
@@ -0,0 +1,1019 @@
...
+         * last level would have 1 more black node than those ending at the
+         * second-to-last level.
+         * Exception: when the only node is the root, because the root must be
+         * black.
+         */
+        val item = xs.next()
+        val red = level == maxUsedDepth && level != 1
+        Node.leaf(keyOf(item), valueOf(item), red, null)
+      case _ =>
+        val leftSize = (size - 1) >> 1 // opt: (size - 1) / 2 with size > 0
```
gzm0: I do not believe this is a reasonable performance optimization. Any halfway decent compiler will do this on its own. (note that the condition `size > 0` is not necessary).

sjrd: For negative numbers on the left, `>>` does not round in the same direction as `/`, so compilers *don't* do that on their own.

It's a rewriting we often do manually in this codebase. There was one recently in `PriorityQueue`.

gzm0: Right... I missed that... (tested on py which of course has different behavior... :()

javalib/src/main/scala/java/util/RedBlackTree.scala
sjrd: Done.

PR #3735, opened 2019-08-01
Implement `j.u.c.ConcurrentHashMap` on top of `j.u.HashMap`.

sjrd:
This allows to get rid of the dependency on `s.c.m.HashMap`, and leverages the many recent improvements to `j.u.HashMap`.

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
@@ -12,225 +12,227 @@
 
 package java.util.concurrent
 
-import java.lang.{reflect => jlr}
 import java.io.Serializable
 import java.util._
 
-import ScalaOps._
+class ConcurrentHashMap[K, V] private (initialCapacity: Int, loadFactor: Float)
+    extends AbstractMap[K, V] with ConcurrentMap[K, V] with Serializable {
 
-class ConcurrentHashMap[K >: Null, V >: Null]
-    extends AbstractMap[K, V] with ConcurrentMap[K, V] with Serializable { self =>
+  import ConcurrentHashMap._
+
+  def this() =
+    this(HashMap.DEFAULT_INITIAL_CAPACITY, HashMap.DEFAULT_LOAD_FACTOR)
 
   def this(initialCapacity: Int) =
-    this()
-
-  def this(initialCapacity: Int, loadFactor: Float) =
-    this()
-
-  def this(initialCapacity: Int, loadFactor: Float, concurrencyLevel: Int) =
-    this()
+    this(initialCapacity, HashMap.DEFAULT_LOAD_FACTOR)
 
   def this(initialMap: java.util.Map[_ <: K, _ <: V]) = {
-    this()
+    this(initialMap.size())
     putAll(initialMap)
   }
 
-  private val inner = new scala.collection.mutable.HashMap[Box[K], V]()
+  def this(initialCapacity: Int, loadFactor: Float, concurrencyLevel: Int) =
+    this(initialCapacity, loadFactor) // ignore concurrencyLevel
 
-  override def isEmpty(): Boolean = inner.isEmpty
+  private[this] val inner: InnerHashMap[K, V] =
+    new InnerHashMap[K, V](initialCapacity, loadFactor)
 
-  override def size(): Int = inner.size
+  override def size(): Int =
+    inner.size()
+
+  override def isEmpty(): Boolean =
+    inner.isEmpty()
 
   override def get(key: Any): V =
-    inner.get(Box(key.asInstanceOf[K])).getOrElse(null)
+    inner.get(key)
 
-  override def containsKey(key: Any): Boolean = {
-    if (key == null)
-      throw new NullPointerException()
-    inner.exists { case (ik, _) => Objects.equals(key, ik()) }
-  }
+  override def containsKey(key: Any): Boolean =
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -12,225 +12,222 @@
...
-        }
-      }
-    }
-  }
+  def contains(value: Any): Boolean =
+    containsValue(value)
 
   def keys(): Enumeration[K] =
-    inner.keys.iterator.map(_.inner).asJavaEnumeration
+    Collections.enumeration(keySet())
```
gzm0: `inner.keySet()` to avoid another allocation?

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -12,225 +12,222 @@
...
     }
+  }
 
-    def toArray(): Array[AnyRef] = {
-      val result = new Array[AnyRef](size())
-      for (keyAndIdx <- iterator().scalaOps.zipWithIndex.scalaOps)
-        result(keyAndIdx._2) = keyAndIdx._1.asInstanceOf[AnyRef]
-      result
-    }
+  class KeySetView[K, V] private[ConcurrentHashMap] (inner: Set[K])
```
gzm0: Maybe add a comment that this is solely for public API typing purposes?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -67,6 +67,34 @@ class ConcurrentHashMapTest extends MapTest {
...
   }
+
+  @Test def testIteratorsAreWeaklyConsistent(): Unit = {
+    val chm1 = factory.empty[String, String]
+    chm1.put("foo", "bar")
+    val iter1 = chm1.keySet().iterator()
+    chm1.remove("foo")
+    assertTrue(iter1.hasNext())
+    assertEquals("foo", iter1.next())
+    assertFalse(iter1.hasNext())
```
gzm0: I don't think this test case is correct:

From [JDK Docs][1]:

> they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction.

In other words: It is correct if the iterator is empty after `chm1.remove`. It is also correct if "foo" is still present.

Same applies to the test cases below.

[1]: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#Weakly

sjrd: Hum, I interpret that differently. To me it says that additions and value updates can be observed, but that removals will never prevent the iterator from seeing a key that existed when the iterator was created.

Otherwise what meaning do you give to the first part of the sentence, namely

> they are guaranteed to traverse elements as they existed upon construction exactly once

?

gzm0: Ah... I felt this is trying to express the "exactly once" part, but indeed, it feels contradictory the more I think about it. Because:

> reflect any modifications subsequent to construction

clearly includes removals in my understanding :S

Note: I have checked the JDK12 doc, the wording is still the same.

sjrd: Well, it looks like your interpretation is correct. I could observe a removal on the JVM:
```scala
scala> import java.util.concurrent.{ConcurrentHashMap => CHM}
import java.util.concurrent.{ConcurrentHashMap=>CHM}

scala> val m = new CHM[String, Int](); m.put("54", 54); m.put("23", 23); m.put("654", 654)
m: java.util.concurrent.ConcurrentHashMap[String,Int] = {23=23, 654=654, 54=54}
res6: Int = 0

scala> val i = m.keySet().iterator()
i: java.util.Iterator[String] = java.util.concurrent.ConcurrentHashMap$KeyIterator@347c5721

scala> i.next()
res7: String = 23

scala> m.remove("54")
res8: Int = 54

scala> i.next()
res9: String = 654

scala> i.next()
java.util.NoSuchElementException
  at java.util.concurrent.ConcurrentHashMap$KeyIterator.next(ConcurrentHashMap.java:3416)
  ... 28 elided
```
So I guess the only guarantees that really exist are:

* If a key existed when the iterator was created, *and* it is not removed, then eventually the iterator will yield it.
* An iterator will never yield the same key twice.
* If an iterator yields a given key, then the associated value will be any value associated with the key at some point since the iterator was created (but not another, arbitrary value).

Does that make sense?

gzm0: Yes, that does make sense. Shall we:
- Document that this is how we interpret the spec.
- Test against this spec.
?

sjrd: Yes, I'll do that.

sjrd:
Updated. I believe I have addressed all the comments.

PR #3736, opened 2019-08-01
Implement `java.util.PriorityQueue` from scratch.

sjrd:
~~Based on #3734. Only the last commit belongs to this PR.~~

Previously, it was implemented on top of `s.c.m.PriorityQueue`. Unlike other collections implemented on top of Scala collections, `PriorityQueue` had no serious hacks. It did have overhead in terms of allocations of `Box`es, though.

The new implementation, done from scratch, does not box elements, and offers direct support for all the methods of the JDK API.

The test suite of `PriorityQueue` is enhanced in the process. Notably, it includes all the standard tests for `Collection`s in `CollectionTest`.

### DIFF
--- a/javalib/src/main/scala/java/util/PriorityQueue.scala
+++ b/javalib/src/main/scala/java/util/PriorityQueue.scala
@@ -12,136 +12,289 @@
 
 package java.util
 
-import java.lang.Comparable
-import scala.math.Ordering
-import scala.collection.mutable
 import scala.annotation.tailrec
-import scala.language.existentials
 
-class PriorityQueue[E] protected (ordering: Ordering[_ >: E], _comparator: Comparator[_ >: E])
-    extends AbstractQueue[E] with Serializable { self =>
+import scala.scalajs.js
+
+class PriorityQueue[E] private (
+    private val comp: Comparator[_ >: E], internal: Boolean)
+    extends AbstractQueue[E] with Serializable {
+
+  def this() =
+    this(NaturalComparator, internal = true)
 
   def this(initialCapacity: Int) = {
-    this(defaultOrdering[E], null.asInstanceOf[Comparator[_ >: E]])
+    this()
     if (initialCapacity < 1)
       throw new IllegalArgumentException()
   }
 
-  def this() =
-    this(11)
+  def this(comparator: Comparator[_ >: E]) = {
+    this(NaturalComparator.select(comparator), internal = true)
+  }
 
   def this(initialCapacity: Int, comparator: Comparator[_ >: E]) = {
-    this(PriorityQueue.safeGetOrdering[E](comparator), null.asInstanceOf[Comparator[E]])
+    this(comparator)
     if (initialCapacity < 1)
       throw new IllegalArgumentException()
   }
 
   def this(c: Collection[_ <: E]) = {
-    this(defaultOrdering[E], null.asInstanceOf[Comparator[E]])
+    this(c match {
+      case c: PriorityQueue[_] =>
+        c.comp.asInstanceOf[Comparator[_ >: E]]
+      case c: SortedSet[_] =>
+        NaturalComparator.select(c.comparator().asInstanceOf[Comparator[_ >: E]])
+      case _ =>
+        NaturalComparator
+    }, internal = true)
     addAll(c)
   }
 
   def this(c: PriorityQueue[_ <: E]) = {
-    this(PriorityQueue.safeGetOrdering[E](c.comparator()),
-        c.comparator().asInstanceOf[Comparator[E]])
+    this(c.comp.asInstanceOf[Comparator[_ >: E]], internal = true)
     addAll(c)
   }
 
   def this(sortedSet: SortedSet[_ <: E]) = {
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/util/PriorityQueue.scala
```diff
@@ -12,136 +12,181 @@
...
+  private[this] def fixUp(m: Int): Unit = {
+    val inner = this.inner // local copy
+    var k: Int = m
+    while (k > 1 && comp.compare(inner(k >> 1), inner(k)) > 0) {
+      val j = k >> 1
+      val h = inner(k)
+      inner(k) = inner(j)
+      inner(j) = h
+      k = j
+    }
```
gzm0: Write this tailrec?

```scala
def fixUp(m: Int): Unit = {
  if (m > 1) {
    val parent = k >> 1
    if (comp.compare(inner(k >> 1), inner(k)) > 0) {
     // swap
     fixUp(parent)
    }
  }
}
```
Much easier to understand IMHO.

javalib/src/main/scala/java/util/PriorityQueue.scala
```diff
@@ -12,136 +12,181 @@
...
   }
 
+  private[this] def fixDown(m: Int, n: Int): Unit = {
```
gzm0: What is `m`, what is `n`?

gzm0: Why is `n` even a parameter? It's simply the ending index no? So we might as well take the length of `inner`?

javalib/src/main/scala/java/util/PriorityQueue.scala
```diff
@@ -12,136 +12,181 @@
...
+      if (comp.compare(inner(k), inner(j)) <= 0) {
+        return
+      } else {
+        val h = inner(k)
+        inner(k) = inner(j)
+        inner(j) = h
+        k = j
+      }
+    }
+    // scalastyle:on return
```
gzm0: Same here: Why not write this tailrec? I think it will be much easier to understand.

sjrd:
Similarly to what I wrote in `TreeSet`, the names and algorithms have been kept the same as in the Scala version to preserve the similarity. I can change them if you think that intrinsic quality is worth more than the similarity.

sjrd:
I have rebased this PR and made it independent of #3734.

I have added comments in the first commit; then I have added a second commit with tail-recursive implementations of `fixUp` and `fixDown`, so that you can compare which one is best.

sjrd:
Well, I discovered several problems in the implementations of `PriorityQueue.remove(Object)` and `PriorityQueue#Iterator.remove()`, which were very subtle. But now I think I've nailed them. I have added test cases that are *very* whitebox, depending on speicific implementation details, because it was the only sure way to test certain corner cases.

sjrd:
Given that the implementation had completely diverged from Scala's `PriorityQueue` by now, I have squashed the changes to tail-rec, and removed the reference to Scala's implementation in the commit message.

gzm0:
Quick update here: The snapshoting behavior of the iterator kind of bothers me, but I can't come up with any better ideas.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PriorityQueueTest.scala
```diff
@@ -236,4 +241,192 @@ class PriorityQueueTest {
...
   }
 
+  @Test def poll(): Unit = {
+    val pq = newPriorityQueueWith0Until100()
+
+    var nextExpected = 0
+    while (!pq.isEmpty()) {
+      assertEquals(nextExpected, pq.poll())
+      nextExpected += 1
+    }
```
gzm0: assert that nextExpected is 100?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PriorityQueueTest.scala
```diff
@@ -236,4 +241,192 @@ class PriorityQueueTest {
...
+        nextExpected += 1
+      }
+    }
+  }
+
+  @Test def removingAnArbitraryElementWithAnIteratorPreservesPrioritiesCornerCase(): Unit = {
+    /* This test depends on internal details of our implementation. It may have
+     * to be changed in the internal data structure and/or algorithms of
+     * PriorityQueue are modified in the future.
+     */
```
gzm0: I'm not sure I understand why these tests must be whitebox tests in order to test what they test. Could you clarify this in the comments? (or if they don't have to, change them to blackbox tests)?

sjrd: They don't need to be whitebox to test what they test. However, they need to be whitebox to *make sure* that they test what we think they're testing. In particular, the scenario where `Iterator.remove()` causes reordering in the heap such that elements yet to be iterated are moved before the cursor, and elements already iterated are moved after. This scenario is not easy to trigger. It requires some deliberate engineering of how the heap will look like and what specific element will be removed.

sjrd:
Updated.

PR #3739, opened 2019-08-06
Fix #3616: Do not force all JSEnv inputs to be of the same type

### LINKED ISSUE #3616: Re-evaluate JSEnv Input modeling
Forcing every input to be of one type is a potential smell ([example](https://github.com/scala-js/jsdependencies/pull/23/files#diff-2946990a42be4eefbeb99cc3ca3f2a94R405)). We should think about this hard once again before we commit to it.

gzm0:


js-envs-test-kit/src/main/scala/org/scalajs/jsenv/test/kit/TestKit.scala
```diff
@@ -56,15 +56,15 @@ final class TestKit(jsEnv: JSEnv, timeout: FiniteDuration) {
     start(codeToInput(code))
 
   /** Starts a [[Run]] for testing. */
-  def start(input: Input): Run =
+  def start(input: List[Input]): Run =
```
sjrd: It's a bit awkward that we had to change every `Input` into a `List[Input]`. Could we instead keep `Input` as the one and only top type for the the whole input of a JSEnv, and have it contain a `List[InputItem]` or something like that?

gzm0: Yes, the change is a bit awkward. But isn't this the better end state?

IMHO we shouldn't decide on an end state based on how nice the *change* will look like. If you'd like the changes to look nicer, I can offer to add an intermediary commit with the `InputItem` type and then replace it in the next commit. This would separate the functionality from the structural change.

I'm reluctant to introduce a mere wrapper type to a List / Seq since it seems unnecessarily complicated and not what we do in `Linker`. (speaking of which: should we use a `Seq` here?).

sjrd: > But isn't this the better end state?

I'm not completely convinced, but you seem more sure than me about it, so let's take your design.

If we're going to use `Seq`, let's use `scala.collection.immutable.Seq` explicitly.

gzm0:
I have updated this to use `Seq`.

> If we're going to use Seq, let's use scala.collection.immutable.Seq explicitly.

I'm not so convinced about this. This would be inconsistent with other `sbt` keys and with how we do thinks in `Linker`. Is there a reason `JSEnv` should be different?

sjrd:
> I'm not so convinced about this. This would be inconsistent with other sbt keys and with how we do thinks in Linker. Is there a reason JSEnv should be different?

It was to have the same API across Scala 2.12 and 2.13, but perhaps consistency with other APIs is more important indeed.

gzm0:
@sjrd Do you have any idea what happens to the CI where it can't find the linked files? (The failures where it can't find `Promise` are due to `ForcePolyfills`, I know how to fix these).

sjrd:
Of the `java.io.FileNotFoundException`s on the CI are just because the logs have been deleted since. It's a Jenkins exception. We have to delete logs after some time because we fill the disks quite quickly otherwise.

So basically, those errors are safe to ignore.

gzm0:
> So basically, those errors are safe to ignore.

Ah, no. There is an error where Node.js fails to load the fast-opt-ed JS file. We'll see it once I fix the Promise usage.

gzm0:
Incorporated gzm0/scala-js#1. Let's see what the CI says.

PR #3740, opened 2019-08-09
Do not special-case `ArrayBufferInputStream` in `DataInputStream`, then remove it.

sjrd:
#### Do not special-case `ArrayBufferInputStream` in `DataInputStream`.

The special-case fundamentally requires a dependency from the javalib to the Scala.js standard library, which we want to avoid.

---

####  Remove `js.typedarray.ArrayBufferInputStream`.

The only reasons this class existed in the core were that:

* it could be special-cased in `DataInputStream`, and
* it was used in the bootstrapped linker.

Now, neither is true anymore, so the existence of `ArrayBufferInputStream` is questionable.

Since it was part of the public API of the Scala.js library, its removal is a breaking change. Users of the class can still recover its functionality by copy-pasting the 0.6.x implementation into their own projects. I suspect this shouldn't be too much of an issue, because typed arrays are better suited to be treated as `java.nio.Buffer`s than `java.io.InputStream`s anyway.


sjrd:
@gzm0 AFAICT, this is the only thing that really must go in 1.0.0 so that we have the possibility to make the javalib free of any dependency on the Scala library. Everything else is internal and can be changed later.

PR #3741, opened 2019-08-13
Enhance the test suite for `java.util.Map`s.

sjrd:
Motivation: the discussion in #3721 made me realize that the tests for `j.u.Map`s were severely lacking in several aspects, most of all the view-ness of `entrySet()`.

* Add more tests for custom objects, for which `eq` is not the same as `equals()`.
* Add more tests for views, in particular for `entrySet()` views which were hardly covered.
* Do not rely on Scala collections and their Java views for the tests.

In the process, we also fix `m.values().contains(null)` for `NullRejectingHashMap`, which previously failed to throw a `NullPointerException`.

/cc @ekrich

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
@@ -28,29 +28,6 @@ trait MapTest {
 
   def factory: MapFactory
 
-  def testMapApi(): Unit = {
-    should_store_strings()
-    should_store_integers()
-    should_store_doubles_also_in_corner_cases()
-    should_store_custom_objects()
-    should_remove_stored_elements()
-    should_remove_stored_elements_on_double_corner_cases()
-    should_put_or_fail_on_null_keys()
-    should_put_or_fail_on_null_values()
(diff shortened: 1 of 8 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -202,29 +323,50 @@ trait MapTest {
...
+
+    val nullMap = TrivialImmutableMap((null: String) -> "y", "X" -> "z")
+    if (factory.allowsNullKeys) {
+      mp.putAll(nullMap)
+      assertEquals("y", mp.get(null))
+      assertEquals("z", mp.get("X"))
+      assertEquals("one", mp.get("ONE"))
+      assertEquals("b", mp.get("A"))
+    } else {
+      expectThrows(classOf[NullPointerException], mp.putAll(nullMap))
```
gzm0: Is it specified how the map looks after throwing this? If yes, test?

sjrd: No, I could not find any specification for that. All we know is that an NPE will be thrown, but AFAICT the map have stored some or none of the other elements before throwing.

Issue #3596, opened 2019-03-28
Incrementally compiling after certain code changes produces LinkingException


ramnivas:
As reported on the gitter channel, I and many others are seeing `LinkingException`s being thrown. Now I have a reasonably easy reproducible case.

Steps:
1. Clone https://github.com/payalabs/scalajs-tutorial/tree/linking-error
2. From sbt, ~fastOptJS
3. Rename `count` in the `Data` class to `countx`. You will see a compilation error (as expected).
4. Rename `countx` back to `count`.

You will see a `LinkingException`.

This exception won't go until you do one of the following:
1. `clean` followed by `fastOptJS`
2. Change (adding a line is fine) **both** `Data.scala` and `TutorialApp.scala`

One caveat: In our app, when we get an error, `clean` has rarely helped. So there may be another condition under which such an exception is produced, but for now, I don't have a reproducible case for that.

cquiroz:
I see this every now and then but haven't been able to reproduce tit, I'm glad you are able to

sjrd:
Thanks for the report. I'll have a look at it ... when I find time.

gzm0:
I will tentatively classify this as bug and 0.6.28 so we do not forget.

ramnivas:
@gzm0 Thanks for scheduling it for 0.6.28. Lately it has been very frequent issue for us (today we had to do full clean 20 times and it is just 3 PM :-( ).

nafg:
While a proper solution to this would be really nice, it would help a LOT if instead of outputting screenfulls of information I don't need, it would just say "X refers to non-existent Y." Log everything else at debug level or something. If I would see that right away it would take five seconds to do the faux-edit-to-trigger-recompilation dance. The hard part almost is just finding which files to re-jigger in the first place.

ramnivas:
An additional issue triggered by this: Every time we clean classes to fix a linking error, Metals gets into a state where everything is in error and doesn't recover after recompilation.

dispalt:
> it would help a LOT if instead of outputting screenfulls of information I don't need

I agree, my current solution is to wade through the text to find the classes, then change the file by adding a comment or something, usually I can get it to work, or I delete my classes directory.

sjrd:
For what it's worth, I can indeed reproduce this with sbt 1.2.8, as in the reproducing project; but it works fine in sbt 0.13.18. sbt 1.0.4, which is in our CI, exhibits a totally different problem which is clearly an sbt bug

[...]

PR #3744, opened 2019-08-18
Suppress stack trace of LinkingException's in sbt plugin since it is useless for users

exoego:
Addresses https://github.com/scala-js/scala-js/issues/3596#issuecomment-502246833

From Scala.js user perspective, long stack trace of `LinkingException` does not help.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -322,9 +322,14 @@ object ScalaJSPluginInternal {
 
             IO.createDirectory(output.getParentFile)
 
-            linker.link(ir, moduleInitializers,
-                AtomicWritableFileVirtualJSFile(output),
-                sbtLogger2ToolsLogger(log))
+            try {
+              linker.link(ir, moduleInitializers,
+                  AtomicWritableFileVirtualJSFile(output),
+                  sbtLogger2ToolsLogger(log))
+            } catch {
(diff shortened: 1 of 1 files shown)


sjrd:
I don't think this is the best way to improve this. Instead I think the sbt plugin should catch `LinkingException`s, and turn them into sbt's `MessageOnlyException`. The stack trace should be preserved for other use cases than the sbt plugin.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -322,9 +322,14 @@ object ScalaJSPluginInternal {
 
             IO.createDirectory(output.getParentFile)
 
-            linker.link(ir, moduleInitializers,
+            try {
+              linker.link(ir, moduleInitializers,
```
exoego: Not sure this is the right place to catch...

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -322,9 +322,14 @@ object ScalaJSPluginInternal {
...
                 AtomicWritableFileVirtualJSFile(output),
                 sbtLogger2ToolsLogger(log))
```
sjrd: These two lines must also be indented two more spaces now.

exoego: Added indent and mention LinkingException in a squashed commit.

PR #3745, opened 2019-08-19
Fix #3743: Minimize imports from ES Module if named imports

### LINKED ISSUE #3743: Implementation details for named import of a member of a module
Currently, for the following code:
```scala
object RawSemanticUi {
  @js.native
  @JSImport("semantic-ui-react", "Button")
  object Button extends js.Object
}
```
we were expecting to see something like:
```js
import { Button } from "semantic-ui-react"
```
but the output is more like (scalajs 1.0.0-M8, fast and full opt.):
```js
import * as $module from "semantic-ui-react";

// used later as:
const x = $module.Button;
```

The documentations [mentions something similar](https://github.com/scala-js/scala-js/blob/master/library/src/main/scala/scala/scalajs/js/annotation/JSImport.scala#L29-L33).

We understand it may be considered as an optimization (and we have seen `Is it a request for more optimizations? Technically not out of scope, but we have already enough ideas and not enough time for those.`) but we wanted to know if it is the case here, or if there is a technical issue with named imports (as they may be used to improve tree shaking with webpack for example).

exoego:
Closes #3743

I assume using `Trees.Import` instead of `Trees.ImportNamespace` will address this since  `Trees.ImportNamespae` is printed as `import * as ...` in
https://github.com/scala-js/scala-js/blob/a77a0dda2a8d57a3dc434c712f8604fffe88cf65/tools/shared/src/main/scala/org/scalajs/core/tools/javascript/Printers.scala#L580-L584


tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -202,7 +202,8 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
           implicit val pos = pos0
           val from = js.StringLiteral(module)
           val moduleBinding = jsGen.envModuleField(module).ident
-          val importStat = js.ImportNamespace(moduleBinding, from)
+          val bindings = List((js.ExportName(moduleBinding.name), moduleBinding))
+          val importStat = js.Import(bindings, from)
```
sjrd: The trees created here represent the following import:
```js
import { $i_someModule as $i_someModule } from "someModule"
```
I don't think this is going to work.

exoego: I thought `moduleBinding.name` is `Button` in `import { Button } from "semantic-ui-react"`
but it is actualy an alias ?

sjrd: `moduleBinding.name` is always an encoded form of the module path (the part after `from`), as you can see in the computation of `jsGen.envModuleField`, and in how `module` is computed in `foreachImportedModule`.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -150,24 +150,24 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
 
-    def foreachImportedModule(f: (String, Position) => Unit): Unit = {
+    def foreachImportedModule(f: (String, Position,  List[String]) => Unit): Unit = {
       val encounteredModuleNames = mutable.Set.empty[String]
       for (classDef <- orderedClasses) {
-        def addModuleRef(module: String): Unit = {
+        def addModuleRef(module: String, path: List[String]): Unit = {
           if (encounteredModuleNames.add(module))
-            f(module, classDef.pos)
+            f(module, classDef.pos, path)
```
sjrd: This will be problematic if there are two imports from the same module with different paths in the program. The first import will override the others, and you will miss some paths in the generated imports.

For example:
```scala
@js.native
@JSImport("url", "URL")
class URL(input: String) extends js.Object

@js.native
@JSImport("url", "URLSearchParams")
class URLSearchParams extends js.Object
```
There is already a test with two different paths: one for `Namespace` and one for `Default`. It should have failed the test suite, but the failure was hidden by the fact that you lump namespace and default together a bit further down (which is not correct in general either).

I suggest you add specific tests with different paths that are not namespace nor default. For example the `URL` and `URLSearchParams` above.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -198,16 +198,32 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
-        foreachImportedModule { (module, pos0) =>
+        foreachImportedModule { (module, pos0, path) =>
           implicit val pos = pos0
           val from = js.StringLiteral(module)
           val moduleBinding = jsGen.envModuleField(module).ident
-          val importStat = js.ImportNamespace(moduleBinding, from)
-          builder.addJSTree(importStat)
+          path match {
+            // namespace import or default import
+            case Nil | List("default") =>
```
sjrd: Namespace and default imports are not the same thing. A default import is actually much more similar to a named import, whose name happens to be `default`. You cannot treat a default import as a namespace import.

tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/Emitter.scala
```diff
@@ -198,16 +198,32 @@ final class Emitter private (semantics: Semantics, outputMode: OutputMode,
...
+              val importStat = js.ImportNamespace(moduleBinding, from)
+              builder.addJSTree(importStat)
+            case _ =>
+              val bindings = path.map(p => (js.ExportName(p), js.Ident(p)))
```
sjrd: I think you may be misunderstanding what a `path` is. In the following import:
```scala
@js.native
@JSImport("foo", "Bar.Foobar")
```
the `path` will be `List("Bar", "Foobar")`.

Only `path.head` is relevant to generate named imports. further selections in the `path` must remain as deeper selections, and should not participate in the `import` statement.

sjrd: On the same line, but unrelated: simply using `js.Ident(p)` is not safe, as `p` could collide with one of the identifiers used internally by the emitter for other purposes. For example `$uD` or `$c_jl_Class`. You'll have to use appropriate avoidance techniques such as those already present in `jsEnv.envModuleField`.

You could also have the same member imported from two different modules, which would also collide using the current code. Therefore the name of the identifiers should contain the encoded module name as prefix or something like that (always taking care that they really cannot collide with anything else).

exoego:
Looks like Rhino JS engined does not understand `module.exports`.
Tests should be relocated (Where is a proper sbt project ?)
```
org.mozilla.javascript.EcmaError: ReferenceError: "module" is not defined. (/localhome/jenkins/sjsjenkins/workspace/scalajs-main_PR-3745@2/test-suite/js/target/scala-2.10/test-classes/JSImportTest.js#11)
```

sjrd:
Tree tests are correctly located. There's a reason they are in a `require-modules` subdirectory.

The problem is adding `JSImportTest.js` to `jsDependencies`. `jsDependencies` only tolerates scripts, not modules.

We don't really have correct infrastructure for custom JS modules in this build. I strongly recommend that you try and find an existing Node.js module exposing a static field/function in a class, to test the paths. One roundabout way would be to reference the `prototype` field of any class. It's pointless from a programming point of view, but it is good enough for the tests.

exoego:
Note to my self
```sbt
set scalaJSModuleKind in testSuite := ModuleKind.ESModule
testSuite/test
```

sjrd:
Closing for inactivity. Feel free to reopen if there is progress.

PR #3747, opened 2019-08-20
Fix the spec of eq/ne for NaN and +0.0 v -0.0.

sjrd:
Previously, `x eq x` would answer `false` if `x` was `NaN`; and `+0.0 eq -0.0` would answer `true`, although `+0.0.equals(-0.0)` answers `false`.

That behavior was intentionally spec'ed, but arguably the spec had been wrong all along. It was specified like that because we couldn't possibly incur the performance hit of a strict identity function for every single `eq` in the world, an operation that is supposed to be virtually free.

Nevertheless, the spec'ed behavior caused a number of issues, for example when using `eq` to cheaply test whether two values were definitely equal (that was incorrect for +0.0 and -0.0), or when using `eq` to find the exact occurrence of a value in a data structure by reference (that was incorrect for NaN).

In this commit, we change the spec to the much more clearly correct alternative where `NaN eq NaN` is true, and `+0.0 eq -0.0` is false. The full check is implemented with the ES 2015 builtin function `Object.is(x, y)`, which we polyfill when emitting ES 5.1.

In order not to have performance issues, the optimizer is enhanced to optimize `eq` down to JavaScript's `===` when it can prove that it is safe to do so. This happens when one of the following applies:

* Either operand can be proven not to be a primitive number, which can often be done by looking at its type. This is a *very* common scenario, notably in all tests of the form `x eq null` or of the form `someNode eq x` (where `someNode` is a value of some type `Node` of a data structure, for example).
* Both operands are known to be whole primitive numbers: `x eq y` can become `x === y` when `x` and `y` are both `int`s, for example.

These optimizations leave very few uses of `eq` as the full check with `Object.is`, all of which are actually desirable fixes. In the test suite, the optimization applies to 99.8% of the use sites of `eq` and `ne`.

This change means that there is no way to write user code corresponding to ECMAScript's `x === y` for arbitrary `x` and `y` anymore. Although it is possible to write it by hand on top of `eq`, it is a bit absurd to undo the effects of `Object.is` to recover the behavior of `===`. We add a new builtin `js.special.strictEquals(x, y)` for that purpose. It is used in the fast path of `BoxesRunTime.equals(x, y)`, which was the only place in the codebase where we really meant `===` when writing `eq`.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
@@ -502,7 +502,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
         RecordValue(tpe, elems map transformExpr)
 
       case IsInstanceOf(expr, ClassRef(ObjectClass)) =>
-        transformExpr(BinaryOp(BinaryOp.!==, expr, Null()))
+        transformExpr(JSBinaryOp(JSBinaryOp.!==, expr, Null()))
 
       case IsInstanceOf(expr, typeRef) =>
         trampoline {
@@ -512,7 +512,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
 
               if (isSubtype(texpr.tpe.base, typeRefTpe)) {
                 if (texpr.tpe.isNullable)
-                  BinaryOp(BinaryOp.!==, finishTransformExpr(texpr), Null())
+                  JSBinaryOp(JSBinaryOp.!==, finishTransformExpr(texpr), Null())
                 else
                   Block(finishTransformStat(texpr), BooleanLiteral(true))
               } else {
@@ -1137,7 +1137,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
             PreTransTree(finishTransformBindings(bindingsAndStats, tree), tpe)
         }
 
-      case _:PreTransUnaryOp | _:PreTransBinaryOp =>
+      case _:PreTransUnaryOp | _:PreTransBinaryOp | _:PreTransJSBinaryOp =>
         PreTransTree(finishTransformExpr(preTrans))
 
       case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
@@ -1181,7 +1181,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
       case PreTransBlock(_, result) =>
         resolveRecordType(result)
 
-      case _:PreTransUnaryOp | _:PreTransBinaryOp =>
+      case _:PreTransUnaryOp | _:PreTransBinaryOp | _:PreTransJSBinaryOp =>
         None
 
       case PreTransLocalDef(localDef @ LocalDef(tpe, _, replacement)) =>
@@ -1238,6 +1238,8 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
         UnaryOp(op, finishTransformExpr(lhs))
       case PreTransBinaryOp(op, lhs, rhs) =>
         BinaryOp(op, finishTransformExpr(lhs), finishTransformExpr(rhs))
+      case PreTransJSBinaryOp(op, lhs, rhs) =>
+        JSBinaryOp(op, finishTransformExpr(lhs), finishTransformExpr(rhs))
       case PreTransLocalDef(localDef) =>
         localDef.newReplacement
 
@@ -1315,6 +1317,11 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
           finishNoSideEffects
       }
 
+    case PreTransJSBinaryOp(op, lhs, rhs) =>
+      if (op == JSBinaryOp.=== || op == JSBinaryOp.!==)
+        Block(finishTransformStat(lhs), finishTransformStat(rhs))(stat.pos)
+      else // other operators can have side effects that we must preserve
+        finishTransformExpr(stat)
     case PreTransLocalDef(_) =>
       Skip()(stat.pos)
     case PreTransRecordTree(tree, _, _) =>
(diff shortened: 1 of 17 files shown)


sjrd:
@gzm0 This was motivated by recent changes in the JDK collections, notably

* the ugly code in `PriorityQueue.removeExact()`
* the fact that `IdentityHashMap` misbehaves for NaN and zeros

This PR is basically ready, but I'll rebase it on top of #3736 and #3721 once they are merged and hit `master`. Feel free to review it now or to wait for that rebase.

Atry:
```
scala> +0.0 eq -0.0
               ^
       error: the result type of an implicit conversion must be more specific than AnyRef
```

Atry:
```
scala> 0 eq (0: Integer)
res5: Boolean = true

scala> 1000 eq (1000: Integer)
res6: Boolean = false
```

Atry:
Does Scala.js have to conforms all the JVM behaviors, including integer cache?

sjrd:
> error: the result type of an implicit conversion must be more specific than AnyRef

You need to put them in their boxed form to observe it (as `java.lang.Double`s, or as values of generic types that are then cast to `AnyRef` to perform a cheap `eq` test.

> Does Scala.js have to conforms all the JVM behaviors, including integer cache?

To the extent that they are specified, and that they can be relied on in reasonable programs, yes, Scala.js is supposed to conform to all the behaviors of the JVM. You can read Chapter 2 of [my thesis](http://lampwww.epfl.ch/~doeraene/thesis/) for the reasons.

For caching, actually Scala.js has a *stronger* spec than the JVM: it guarantees (observable) caching of all primitive numeric types except `Long`s.

Atry:
> Nevertheless, the spec'ed behavior caused a number of issues, for example when using eq to cheaply test whether two values were definitely equal (that was incorrect for +0.0 and -0.0), or when using eq to find the exact occurrence of a value in a data structure by reference (that was incorrect for NaN).

The usage of `eq` is not portable, since JVM has a weaker spec, therefore, even if this draft is implemented as designed, portable library authors should not use it.

Instead, I propose to add a `scala.scalajs.js.strictEq` function, which is clearly reliable to JS only library.

sjrd:
> The usage of eq is not portable, since JVM has a weaker spec, therefore, even if this draft is implemented as designed, portable library authors should not use it.

Some reasonable uses of `eq`, even with `Double` boxed, are portable (with this PR, but not before), for example:

* Concluding that `x.equals(y)` is true from the cheap test `x eq y`.
* If you have a value `x` of a generic type that you found in a container, you can find its exact location in the container later with `eq`.

Those things are actually used in reasonable code. We should support them.

> Instead, the js.special.strictEquals function should be present as a public API, which is clearly reliable to JS only libraries.

It is made public in this PR.

Atry:
`new Integer(1) eq new Integer(1)` is true in Scala.js. Does that make sense?

sjrd:
It may not *make sense* on its own, but it is a consequence of absolutely critical JavaScript interop. Because `new Integer(1)` returns the primitive JavaScript `number` 1.

gzm0:
As a higher level comment that came up during the first part of the review on this: I'm unclear about the exact semantics of `BinaryOp.===` / `BinaryOp.!==`. IIUC, they are literally the definition of `eq`?

Further, I am not sure I understand why the optimizer is allowed to emit `JSBinaryOp` instead of `BinaryOp`: Isn't this a breach of abstraction w.r.t. how the emitter actually implements classes? (IIUC `Long`s might have or still do receive a similar treatment). Especially since some of them seem to be trivially optimizable by the `Emitter` (n.b. `x === null`).

sjrd:
`BinaryOp.===` and `!==` are by definition `eq` and `ne`, yes. `x === y` if and only if they have the same `typeof` and:

* If they are `number`s, they are equal, with the note that `NaN` *is* equal to itself and that `+0` *is not* equal to `-0`.
* If they are `string`s, they contain the sequence of characters.
* If they are `bigint`, they represent the same mathematical value.
* If they are `boolean`s, they are either both true or both false.
* If they are any other value, they have the same reference identity.

The optimizer is allowed to perform the optimization because the semantics above are fixed and well-defined.

For `Long`s, these semantics result in different observable behavior depending on whether they are compiled as `bigint` or not, sure. But that has always been the case, and is compatible with the fact that caching instances of `java.lang.Long`s is allowed but not required, and therefore left as an implementation detail.

gzm0:
I see. So then the reason we need to give `JSBinaryOp` a non `any` type is that the trees in the optimizer remain well-typed, right? In this case, how would you feel about allowing to explicitly give the `JSBinaryOp` tree a type? Basically saying something like: If the creator of the tree (i.e. the optimizer), knows enough about the operands and the operation in the current ES target, it may optionally type trees stricter than `any`.

A formulation like this would:
- Generalize the notion (and for example allow to type `x + y` as `Double` provided `x` and `y` are).
- Not assume anything about future evolution of ES (except maybe for backwards compatibility).

sjrd:
It's a possibility. It would mean that, should we ever get to run the IR checker after the optimizer, it would have to trust the phases before it for that `tpe`. Not really a checker anymore, in that case. :-s

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3072,7 +3099,46 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
     (op: @switch) match {
       case === | !== =>
-        val positive = (op == ===)
+        // Try to optimize as a primitive JS strict equality
+
+        def canBePrimitiveNum(tpe: RefinedType): Boolean = tpe.base match {
+          case AnyType | ByteType | ShortType | IntType | FloatType | DoubleType =>
+            true
+          case ClassType(cls) =>
+            !tpe.isExact && MaybeHijackedPrimNumberClasses.contains(cls)
```
gzm0: To check my understanding: If the type is exact and a class type of a primitive box, it means the only allowed value is `null` which is not a primitive number. Correct?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3072,7 +3099,46 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
+        def isWhole(tpe: RefinedType): Boolean = tpe.base match {
+          case ByteType | ShortType | IntType =>
+            true
+          case ClassType(cls) =>
+            cls == Definitions.BoxedByteClass ||
+            cls == Definitions.BoxedShortClass ||
+            cls == Definitions.BoxedIntegerClass
+          case _ =>
+            false
+        }
```
gzm0: To check my understanding: We do not have longs here, because we do not have longs in `canBePrimitiveNum`, because we never represent them with primitive numbers. Correct?

gzm0:
> It's a possibility. It would mean that, should we ever get to run the IR checker after the optimizer, it would have to trust the phases before it for that tpe. Not really a checker anymore, in that case. :-s

Right. Indeed that's less than ideal. So in this case the current solution is probably better. Could you add a comment to the `Tree` explaining why we make this assumption?

PR #3751, opened 2019-08-23
Improve error messages for Scala.js newcomers

Atry:


sjrd:
The thing is that `%%%` is not the only possible culprit. There are several possible causes:

* Trying to use `Await`, resulting in obscure linking errors talking about `CompletionLatch` or what not.
* Depending, possibly transitively, on Java libraries that are not supported, which can either requiring depending on additional artifacts, or being out of luck.
* Getting binary incompatible artifacts on the classpath.
* Getting bitten by the incremental compiler messing up somehow.

So I'm afraid that giving one of the possibilities as explanation in the error message can confuse *even more* the user of they are in fact facing one of the other causes.

I've been thinking for a while that a better solution would be to add a dedicated page on the website about linking errors, with their various causes and solutions. And then the error message could refer to that Web page for troubleshooting help.

exoego:
If there are many possible causes for error, I think it is common practice to include a link to dedicated page in an error message, like
`Referring to non-existent class ${info.displayName}. Refer https://www.scala-js.org/path/to/dedicated-page`

sjrd:
> Is it possible to produce different messages when missing SJS IR for classes compiled from Java?

Well, no, because if the IR is *not there*, how on earth would you tell whether it was supposed to have been compiled from Java?

Anyway, even when you do have the IR, nothing in it can it can help you telling whether it was originally a Java library or a Scala library. That's by the design: the IR is designed to be language-agnostic.

sjrd:
Closing for inactivity.

A different PR could be reopened with a message pointing to https://www.scala-js.org/doc/project/linking-errors.html when the sbt plugin catches a `LinkingException` ([here in 0.6.x](https://github.com/scala-js/scala-js/blob/368e57cfa848f261140d81b0b3366b16f39159bc/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L330-L331) or [here in 1.x](https://github.com/scala-js/scala-js/blob/69935b2781e5e8e249491d2483e1abcc42d9c877/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L208-L209)).

PR #3752, opened 2019-08-26
Fix #2066: Emit `x.isInstanceOf[C]` as `x instanceof C` when possible.

### LINKED ISSUE #2066: Optimize isInstanceOf on Scala objects

sjrd:
When the rhs of an `isInstanceOf` refers to a *class* `C` (as opposed to a trait), it is semantically correct to compile it as `x instanceof $c_C`. It turns out that this is much faster.

In fastOpt mode, the optimization also affects `asInstanceOf`s.

In fullOpt mode, most instance tests come from pattern matching, so effectively this optimization is very effective on pattern-matching-intensive workloads.

Benchmarks of the bootstrapped Scala.js optimizer while optimizing the test suite show speedups of up to 47% in fastOpt and 40% in fullOpt, with the maximum for the optimizer core.

[Benchmarks data](https://gist.github.com/sjrd/a8abe0ca0d331c0c4fb9c0a3ba88cfa1)
[Charts](https://github.com/scala-js/scala-js/files/3541154/isInstanceOf.benchmarks.pdf)

This should probably be backported to 0.6.x afterwards. Since the commit doesn't merge remotely cleanly, putting it in 0.6.x first and then dealing with the merge will not be easier.


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/MyDefinitions.scala
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.backend.emitter
+
+import org.scalajs.ir.Definitions._
+
+private[emitter] object MyDefinitions {
```
gzm0: Maybe call this `EmitterDefinitions`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -104,14 +106,27 @@ private[emitter] final class JSGen(val semantics: Semantics,
   }
 
   def genIsInstanceOf(expr: Tree, typeRef: TypeRef)(
+      implicit globalKnowledge: GlobalKnowledge, pos: Position): Tree = {
+    val useInstanceOf = typeRef match {
+      case ClassRef(className) =>
+        className != NumberClass && // the only non-Object super*class* of hijacked classes
```
gzm0: Could we put this in `MyDefinitions` as a `Set` as well, just so that all the hijacked hierarchy stuff is in one place?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -155,9 +170,14 @@ private[emitter] final class JSGen(val semantics: Semantics,
             }
           }
         } else {
-          Apply(
-              envField(if (test) "is" else "as", className),
-              List(expr))
+          if (useInstanceOf) {
+            assert(test)
+            expr instanceof encodeClassVar(className)
```
gzm0: I wonder if it would make sense to dispatch this in genIsInstanceOf instead.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -104,14 +106,27 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+      case ArrayTypeRef(_, _)  =>
+        false
+    }
+    genIsAsInstanceOf(expr, typeRef, test = true, useInstanceOf)
+  }
+
+  def genIsInstanceOf(expr: Tree, typeRef: TypeRef, isScalaClass: Boolean)(
       implicit pos: Position): Tree =
-    genIsAsInstanceOf(expr, typeRef, test = true)
+    genIsAsInstanceOf(expr, typeRef, test = true, useInstanceOf = isScalaClass)
```
gzm0: Hum, isn't this wrong in the case where `typeRef == ClassRef(NumberClass)`?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -585,6 +585,14 @@ final class Emitter private (config: CommonPhaseConfig,
...
 
     if (classEmitter.needInstanceTests(linkedClass)) {
+      if (!linkedClass.hasInstances && kind.isClass) {
+        /* The isInstanceOf implementation will generate
+         * `x instanceof $c_TheClass`, but `$c_TheClass` won't be declared at
+         * all. Define it as a fake class to avoid `ReferenceError`s.
+         */
+        addToMainBase(classEmitter.genFakeClass(linkedClass))
+      }
+
```
gzm0: Hmmm... This makes me wonder:

- If we put this into `GlobalKnowledge` could we simply emit `false` instead?
- Do we still need the normal instance tests (at least the `is` part), if this is a class?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -707,10 +728,14 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
               js.UnaryOp(JSUnaryOp.typeof, obj) === js.StringLiteral("string")
 
             case _ =>
-              var test = {
-                genIsScalaJSObject(obj) &&
-                genIsClassNameInAncestors(className,
-                    obj DOT "$classData" DOT "ancestors")
+              var test = if (tree.kind.isClass) {
+                obj instanceof encodeClassVar(className)
+              } else {
```
gzm0: When is this case needed? (kind of similar question as below).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
sjrd: Yes, it is, but it is never called for `NumberClass`. I have renamed the method and removed the argument to make it clearer. I have also added specific tests for `isInstanceOf[Number]`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
sjrd: OK scratch that. I found a relatively elegant way to do it. Re-updated.

gzm0: > We could, but I think that is a job for the optimizer, not for the emitter.

Fair enough.

PR #3753, opened 2019-08-26
[no-master] Fix #2066: Emit `x.isInstanceOf[C]` as `x instanceof C` when possible.

### LINKED ISSUE #2066: Optimize isInstanceOf on Scala objects

sjrd:
Backport of fa808380af3766472057194252d8d4c162b5493f.

When the rhs of an `isInstanceOf` refers to a *class* `C` (as opposed to a trait), it is semantically correct to compile it as `x instanceof $c_C`. It turns out that this is much faster.

In fastOpt mode, the optimization also affects `asInstanceOf`s.

In fullOpt mode, most instance tests come from pattern matching, so effectively this optimization is very effective on pattern-matching-intensive workloads.


sjrd:
Updated to match the final state of #3752.

PR #3756, opened 2019-08-29
update some java.nio test cases that fail on dotty

mcku:
These tests were failing due to https://github.com/lampepfl/dotty/issues/7132

Updated tests pass both on Scala 2 and Dotty..

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala
@@ -129,7 +129,7 @@ abstract class ByteBufferTest extends BaseBufferTest {
 
     buf.order(ByteOrder.BIG_ENDIAN)
     val charBuf1 = buf.asCharBuffer()
-    assertEquals(createsReadOnly, charBuf1.isReadOnly)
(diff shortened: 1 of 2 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala
```diff
@@ -129,7 +129,7 @@ abstract class ByteBufferTest extends BaseBufferTest {
 
     buf.order(ByteOrder.BIG_ENDIAN)
     val charBuf1 = buf.asCharBuffer()
-    assertEquals(createsReadOnly, charBuf1.isReadOnly)
+    assertEquals(createsReadOnly, charBuf1.asInstanceOf[java.nio.Buffer].isReadOnly)
```
smarter: since this is an upcast, you can use `: java.nio.Buffer`, instead of `asInstanceOf`, the latter is unsafe since it will still compile even if the type of `charBuf1` changes in the future.

sjrd:
I don't like this. There is clearly a bug in dotty (https://github.com/lampepfl/dotty/issues/7132) which should be fixed. Working around it in our codebase is not the right way to go. I could conceive something like that if we did something obscure, but this code is very simple, straightforward stuff that could affect any reasonable codebase.

mcku:
Agreed. closing it..

PR #3757, opened 2019-08-30
Fix #3419 and #3596: Backup and restore .sjsir files ourselves.

### LINKED ISSUE #3419: sbt plugin: incremental compilation issue
In a file `A.scala`:
```scala
class A(val x: Int)
```
In a file `Test.scala`
```scala
class Test {
  def test1(x: A) = 1
  def test2 = test1(null)
}
```
```shell
> sbt compile
[info] Compiling 2 Scala sources to target/scala-2.12/classes ...
[info] Done compiling.
[success] Total time: 1 s, completed Jul 31, 2018 1:57:29 PM

> shasum target/scala-2.12/classes/*
0aada2bc8ba6e2f3cebdf42f7e911931bc774419  target/scala-2.12/classes/A.class
94aded7b8d5330832a7a701fd9ad55e6cb323ef7  target/scala-2.12/classes/A.sjsir
faedd18f93e31fd6e069bf57d7e0cb5945659689  target/scala-2.12/classes/Test.class
84e558abcc97540db40a66cacbd72928e817dd08  target/scala-2.12/classes/Test.sjsir
```
Now I modify `A.scala` as follows:
```scala
class A(val x: Int) extends AnyVal
```

```shell
> sbt compile
[info] Compiling 1 Scala source to target/scala-2.12/classes ...
[info] Done compiling.
[info] Compiling 1 Scala source to target/scala-2.12/classes ...
[error] src/main/scala/Test.scala:3:21: type mismatch;
[error]  found   : Null(null)
[error]  required: A
[error]   def test2 = test1(null)
[error]                     ^
[error] one error found
[error] Total time: 1 s, completed Jul 31, 2018 2:00:13 PM

> shasum target/scala-2.12/classes/*
a19fa4fa7e206eabe387f73d9e0221a445da3647  target/scala-2.12/classes/A$.sjsir
0aada2bc8ba6e2f3cebdf42f7e911931bc774419  target/scala-2.12/classes/A.class
cb0fe0d549ef5256100190b9ac9e3031c52f0299  target/scala-2.12/classes/A.sjsir
faedd18f93e31fd6e069bf57d7e0cb5945659689  target/scala-2.12/classes/Test.class
84e558abcc97540db40a66cacbd72928e817dd08  target/scala-2.12/classes/Test.sjsir
```

When we compile the second time, `A.scala` is recompiled successfully which triggers recompilation of `Test.scala`. The compilation of `Test.scala` fails as expected. `sbt` (aka Zinc) reverts the class files produced during compilation of `A.scala`. It should as well revert the generation of the `.sjsir` files. We are in a state where the class file for `A.scala` are not in sync with the `.sjsir` files. The class files come from the first compilation of `A.scala` and the `.sjsir` files come from the second compilation of `A.scala`.

The issue is that the second compilation of `A.scala` overwrites the class files and `.sjsir` files generated during the first compilation. However Zinc saves the class files in a temporary directory and then restore them when the compilation of `Test.scala` fails. `.sjsir` files are not restored. [Here](https://github.com/sbt/zinc/blob/16e53c2f3933cd9c57f7aee0c5f557ecc34a1592/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala#L82-L109) is the logic in Zinc that deals with the class files.

This issues manifested itself recently in Dotty lampepfl/dotty#4857. Our logic that deals with TASTY files is [copy-pasted](https://github.com/scala-js/scala-js/blob/53a0d61421e73beaef89e73e19b6a552bd519415/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala#L25-L58) from the Scala.js sbt plugin.

sjrd:
Rather than relying on zinc to do it based on the arguments to `delete()`, which it hasn't done since sbt 1.x.

The fix requires sbt 1.2.1+ to be meaningful, so we first upgrade our sbt plugin to use sbt 1.2.1.

### DIFF
--- a/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SJSIRFileManager.scala
+++ b/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SJSIRFileManager.scala
@@ -1,0 +1,117 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.sbtplugin
+
+import java.io.File
+import java.nio.file.Files
+
+import scala.collection.mutable
+
+import sbt._
+
+import xsbti.compile.ClassFileManager
+
+/** A class file manager that prunes .sjsir files as needed.
+ *
+ *  This makes sure that, when a .class file must be deleted, the
+ *  corresponding .sjsir file is also deleted, as well as maintaining the
+ *  transactional aspect of incremental compilation in the presence of
+ *  compilation errors in an intermediate run (as product files need to be
+ *  backed up and later restored).
+ *
+ *  This code is adapted from Zinc `TransactionalClassFileManager`.
+ *  We need to duplicate the logic since forwarding to the default class
+ *  file manager doesn't work: we need to backup sjsir files in a different
+ *  temporary directory as class files.
+ *
+ *  We stole this from `TastyFileManager` in dotty, at
+ *  https://github.com/lampepfl/dotty/blob/0.17.0-RC1/sbt-dotty/src/dotty/tools/sbtplugin/TastyFileManager.scala
+ *
+ *  There are plans to upstream a generic version of this code in zinc in the
+ *  future: https://github.com/sbt/zinc/issues/579
+ */
+private[sbtplugin] final class SJSIRFileManager extends ClassFileManager {
+  private[this] var _tempDir: File = null
+
+  private[this] def tempDir: File = {
+    if (_tempDir == null)
+      _tempDir = Files.createTempDirectory("backup").toFile
+    _tempDir
+  }
+
+  private[this] val generatedSJSIRFiles = new mutable.HashSet[File]
+  private[this] val movedSJSIRFiles = new mutable.HashMap[File, File]
+
+  /* The incremental compiler works along the following general protocol:
+   * 1. it identifies the set of files that have been changed/added/removed
+   * 2. it compiles that set of files
+   *    a) it first `delete`s the products of changed and removed files
+   *    b) then it compiles the files
+   *    c) it registers produced files as `generated`
+   * 3. it detects whether a new round of compilation is necessary, and with
+   *    which files; if necessary, it loops back to 2.
+   * 4. when no new round is necessary, it `complete`s with `success = true` if
+   *    the whole cycle is successful, or `success = false` otherwise.
+   *
+   * Consequently, it is quite possible that, within one cycle, a given file is
+   * `generated` during one iteration, and `delete`d in a subsequent one. In
+   * this case, we would need to revert back to the file not being present in
+   * the final transactional output on `success = false`. This is why we have
+   * the `!generatedSJSIRFiles(t)` test in `delete()` below.
+   */
+
+  override def delete(classes: Array[File]): Unit = {
+    val files = sjsirFiles(classes)
+    val toBeBackedUp = files
+      .filter(t => !movedSJSIRFiles.contains(t) && !generatedSJSIRFiles(t))
+    for (c <- toBeBackedUp)
+      movedSJSIRFiles.put(c, backupFile(c))
+    IO.deleteFilesEmptyDirs(files)
+  }
+
+  override def generated(classes: Array[File]): Unit =
+    generatedSJSIRFiles ++= sjsirFiles(classes)
+
+  override def complete(success: Boolean): Unit = {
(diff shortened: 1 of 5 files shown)


sjrd:
Tested locally with the reproduction from #3596.

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SJSIRFileManager.scala
```diff
@@ -0,0 +1,100 @@
...
+    _tempDir
+  }
+
+  private[this] val generatedSJSIRFiles = new mutable.HashSet[File]
+  private[this] val movedSJSIRFiles = new mutable.HashMap[File, File]
+
+  override def delete(classes: Array[File]): Unit = {
+    val files = sjsirFiles(classes)
+    val toBeBackedUp = files
+      .filter(t => t.exists && !movedSJSIRFiles.contains(t) && !generatedSJSIRFiles(t))
```
gzm0: The existence test is redundant: `sjsirFiles` already only returns files that exist.

gzm0: When is the test for generatedSJSIRFiles relevant? From what I understand, if it ever triggers, this would be a bug no? Because a file deleted prior to one compilation step would be deleted prior to another compilation step of the same run...

gzm0: When is the test for `generatedSJSIRFiles` relevant? From what I understand, if it ever triggers, this would be a bug no? Because a file generated in one compilation step would be deleted in another compilation step of the same run...

Note: I'm following the doc here: https://github.com/sbt/zinc/blob/develop/internal/compiler-interface/src/main/java/xsbti/compile/ClassFileManager.java

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SJSIRFileManager.scala
```diff
@@ -0,0 +1,100 @@
...
+    if (!success) {
+      IO.deleteFilesEmptyDirs(generatedSJSIRFiles)
+      for ((orig, tmp) <- movedSJSIRFiles)
+        IO.move(tmp, orig)
+    }
+
+    generatedSJSIRFiles.clear()
+    movedSJSIRFiles.clear()
+    if (_tempDir != null) {
+      IO.delete(tempDir)
```
gzm0: This will create the tempDir again. Maybe name it differently so it is clear it creates it?

sjrd: Hum, we just tested that it is not null, so it will not recreate it. It will only delete the one that was already created.

We could use `_tempDir` to make it clearer though.

sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SJSIRFileManager.scala
sjrd: OK I figured it out. It is actually necessary. The thing is that incremental compilation works in a loop: each partial compilation can find out that more sources should be recompiled, and because of cycles, it happens that a same source file will be recompiled twice or more in the same cycle. If something has been generated by a first iteration, but deleted in the second iteration, we must not back it up. It must stay deleted at the end of the run (even when `!success`).

In zinc, we find

1. [`Incremental.compile`](https://github.com/sbt/zinc/blob/9200d1e7ec363fc069aa0fc903f548b786410706/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L922) calls `cycle` inside `manageClassfiles`
2. [`manageClassfiles`](https://github.com/sbt/zinc/blob/9200d1e7ec363fc069aa0fc903f548b786410706/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L149) performs a `try..catch` to always call `success` at the end (with `false` if it caught an exception)
3. [`cycle`](https://github.com/sbt/zinc/blob/9200d1e7ec363fc069aa0fc903f548b786410706/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala#L762) is a loop that calls `recompileClasses` during one opération.
4. [`recompileClasses`](https://github.com/sbt/zinc/blob/9200d1e7ec363fc069aa0fc903f548b786410706/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala#L170) calls `delete` once then `generated` once

So we can definitely receive a sequence of calls where `delete` then `generated` are called several times before a single `success` is called.

sjrd:
Updated.

sjrd:
> Therefore I don't think we should block this PR on testing, but maybe we can at least have an issue so we don't forget to add them?

#3758

PR #3759, opened 2019-09-02
[no-master] Disable slow scalaTestSuite tests when running in Rhino.

sjrd:
Because under Rhino, they are *too* slow.


sjrd:
`scalaTestSuite/test` under 2.13.0 is our biggest source of spurious failures in our CI at the moment. This should address that problem.

PR #3761, opened 2019-09-05
Rewrite java.util.regex.GroupStartMap.

sjrd:
First: I'm fine if it takes months before this PR is reviewed/merged. ;-)

I have kept all the intermediate commits, which are successive refactorings, to ease review. Eventually everything should be squashed together, I believe.

At the end of this rewrite, several things have improved:

* We don't use Scala collections anymore. We only use `js.Array`s.
* The parser does not constantly build and deconstruct `Seq`s, improving performance.
* We only parse at most once per instance of `java.util.regex.Pattern`. We can reuse the parse tree across multiple usages of the same `Pattern`.
* The AST data structure is a lot cleaner, and its semantics easier to understand.
* The code paths have been simplified a lot.


javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+
+  def apply(string: String, start: Int): js.Array[Int] = {
+    allMatchingRegex.lastIndex = start
+    val allMatchResult = allMatchingRegex.exec(string)
+    if (allMatchResult == null) {
+      throw new Exception(
+          s"[Internal error] Executed '$allMatchingRegex' on " +
+          s"'$string' at position $start, got an error.\n" +
+          s"Original pattern '${pattern}' did match however.")
+    }
```
gzm0: Shouldn't this be an `AssertionError`?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+
+    // Prepare a `groupNodeMap` array with the correct length filled with -1
+    val len = groupNodeMap.length
```
gzm0: Use `allMatchResult.length`? Then you can avoid passing `groupNodeMap`.

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+    groupStartMap
+  }
+}
+
+private[regex] object GroupStartMapper {
+  def apply(pattern: Pattern): GroupStartMapper = {
+    val parser = new Parser(pattern.jsPattern)
+    val node = parser.parseTopLevel()
+    val groupNodeMap = parser.groupNodeMap
+    val flags = pattern.jsFlags
```
gzm0: Inline this below to avoid polluting the scope?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+    node.setNewGroup(1)
+    val allMatchingRegexStr = node.buildRegex(groupNodeMap)
```
gzm0: Inline?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+
+    def setEndReturnStart(matchResult: js.RegExp.ExecResult,
+        groupStartMap: js.Array[Int], end: Int): Int = {
+
+      val start = matchResult(newGroup).fold {
+        -1
+      } { matched =>
+        val start = end - matched.length
+        if (this.isInstanceOf[GroupNode])
+          groupStartMap(this.asInstanceOf[GroupNode].number) = start
```
gzm0: Can't this go into `propagate` and then we don't need to do instance checks?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+    /** Propagates the start or end position of this node to its descendants.
+     *
+     *  According to the big comment above, `RepeatedNode`s propagate the
+     *  `end`, while other nodes propagate the `start`.
+     */
+    def propagate(matchResult: js.RegExp.ExecResult,
+        groupStartMap: js.Array[Int], start: Int, end: Int): Unit
+
+    def setEndReturnStart(matchResult: js.RegExp.ExecResult,
+        groupStartMap: js.Array[Int], end: Int): Int = {
```
gzm0: `final def propagateFromEnd`?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+
+      val start = matchResult(newGroup).fold {
+        -1
+      } { matched =>
+        val start = end - matched.length
+        if (this.isInstanceOf[GroupNode])
+          groupStartMap(this.asInstanceOf[GroupNode].number) = start
+        start
+      }
+      propagate(matchResult, groupStartMap, start, end)
```
gzm0: What happens if we propagate a -1 here?

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
```diff
@@ -0,0 +1,497 @@
...
+      start
+    }
+
+    def setStartReturnEnd(matchResult: js.RegExp.ExecResult,
```
gzm0: `final def propagateFromStart`?

gzm0:
Note that I have simply reviewed the final result. So feel free to squash (or not).

javalib/src/main/scala/java/util/regex/GroupStartMapper.scala
sjrd: TBH I'm not quite sure. It's doing whatever it did before this PR. I believe it doesn't really matter; at least it doesn't seem to matter in any of our tests.

sjrd:
Squashed :)

PR #3764, opened 2019-09-08
Use scripted for concurrent linker test

gzm0:
This is so that the main sbt-plugin-test does not have a direct dependency on the Linker anymore (see #3805 for details).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -364,7 +364,8 @@ def Tasks = [
         linker/compile:doc jsEnvs/compile:doc \
         jsEnvsTestKit/compile:doc nodeJSEnv/compile:doc \
         testAdapter/compile:doc \
-        sbtPlugin/compile:doc
+        sbtPlugin/compile:doc &&
+    sbt sbtPlugin/scripted
   ''',
 
   "partestc": '''
@@ -393,7 +394,7 @@ def Tasks = [
         noDOM/testHtml multiTestJS/testHtml \
         test \
         noDOM/testScalaJSModuleInitializers \
-        noDOM/clean noDOM/concurrentUseOfLinkerTest \
+        noDOM/clean \
         multiTestJS/test:testScalaJSSourceMapAttribute &&
     sbt 'set scalaJSStage in Global := FullOptStage' \
         noDOM/testHtml multiTestJS/testHtml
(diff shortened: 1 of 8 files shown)


sbt-plugin/src/sbt-test/legacy/legacy/build.sbt
```diff
@@ -14,7 +14,7 @@ version := scalaJSVersion
 
 val versionSettings = Seq(
   version := scalaJSVersion,
-  scalaVersion := "2.11.12"
+  scalaVersion := "2.12.8",
```
sjrd: This change actually concerns me. The sbt plugin test was the only thing in our CI that would ensure that Scala.js works at all when using a `scalaVersion` that is different from the sbt plugin's version of Scala. Using 2.11.12 was quite intentional here. Switching to 2.12.8 digs a large hole in our test coverage.

I understand the difficulty of doing `publishLocal` of two sets of artifacts with differing `scalaVersion`s when using `scripted`, but I think we shouldn't lose the above coverage.

gzm0: Ah... Is this even possible with scripted?

sjrd: I don't know. I'll ask in the gitter channel.

gzm0:
It seems scripted isn't forwarding the ivy home:

```
[info] 	published ivy to /localhome/jenkins/scala-js-sbt-homes/sbt-home/local/org.scala-js/sbt-scalajs/scala_2.12/sbt_1.0/1.0.0-SNAPSHOT/ivys/ivy.xml

 ==== local: tried
[info] [warn]   /home/jenkins/.ivy2/local/org.scala-js/sbt-scalajs/scala_2.12/sbt_1.0/1.0.0-SNAPSHOT/ivys/ivy.xml
```

gzm0:
Culprit seems to be this line:

https://github.com/sbt/sbt/blob/4ff4f6e45e7934ab4331cb8fa8b7dbe4079f2d80/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/RemoteSbtCreator.scala#L42

It doesn't forward relevant envs. Looks like a scripted bug to me.

sjrd:
Should we report it? Is it a blocker for our use case?

gzm0:
It seems to be a blocker for our current CI setup, yes.

gzm0:
Actually, I think I was able to work around it. The values are read as system properties and not environment variables so we can set them.

PR #3781, opened 2019-09-25
Fix #3778: Never inline `clone__O` if the receiver can be an array.

### LINKED ISSUE #3778: 2.13 ListBuffer clone breaks Vector
Repeatable with Scala 2.13.1 and Scala.js 0.6.29

Making use of `ListBuffer.clone` changes the generated JS code and breaks Vector:

Here's the simplified test case:
```scala
import org.scalatest.FunSuite

import scala.collection.mutable.ListBuffer

class UseVector extends FunSuite{
  test("listbuffer clone") {
    val buffer = new ListBuffer[Int]
    buffer.clone
  }

  test("update") {
    val v1 = Vector(0,1,2,3)
    val v2 = v1.updated(2,9)
    assert(v2 === Vector(0,1,9,3))
  }
}
```
Comment out `buffer.clone` and the tests pass. Include `buffer.clone` and the second test fails with
```
[info]   scala.scalajs.js.JavaScriptException: TypeError: $thiz.newSpecificBuilder__scm_Builder is not a function
[info]   at scala.collection.mutable.SeqOps.clone(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:52347:17)
[info]   at scala.collection.immutable.VectorPointer.gotoPosWritable0(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:2299:49)
[info]   at scala.collection.immutable.Vector.gotoPosWritable(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:70389:5)
[info]   at scala.collection.immutable.Vector.updateAt(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:70803:5)
[info]   at {anonymous}()(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:63179:19)
[info]   at scala.scalajs.runtime.AnonFunction0.apply(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:34310:23)
[info]   at org.scalatest.OutcomeOf.outcomeOf(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:1405:7)
[info]   at org.scalatest.Transformer.apply(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:40944:10)
[info]   at org.scalatest.FunSuiteLike$$anon$1.apply(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:35429:72)
[info]   at org.scalatest.FunSuiteLike.invokeWithFixture$1(/home/jon/Code/Jon/JSError/target/scala-2.13/classes/js/js-error-test-fastopt.js:61463:15)
```
This seems to be due to different code generated in `VectorPointer` depending on whether clone is used:
```js
$thiz.display0$und$eq__AO__V($asArrayOf_O($thiz.display0__AO().clone__O(), 1));
```
```js
$thiz.display0$und$eq__AO__V($asArrayOf_O($f_scm_SeqOps__clone__O(this$6), 1));
```
The first works, the second fails inside `SeqOps__clone` with "is not a function" as shown above.

This use of `ListBuffer` `clone` worked correctly in Scala 2.12.9 and Scala.js 0.6.28.

sjrd:
The `clone__O()` method of arrays is very peculiar in our pipeline: it is the only method that is defined in array classes, for which no IR exists. There are several pieces of the linker that special-case `clone__O()` for arrays for that reason, but we were missing one detail: the optimizer must never try to inline `clone__O()` if the receiver can be an array.

Before this commit, the optimizer could erroneously do that, because the CHA done by `GenIncOptimizer` does not have any provision for array classes, and would return the only reachable `clone__O()` method if there was one in the classpath (on `Object` or on an arbitrary other class).

In this commit, we specifically forbid inlining of `clone__O()` if the receiver can be an array.

Since the test suite reaches the `clone__O()` method of many classes, it is impossible to reproduce the problematic behavior in the test suite. Instead, we user linker unit tests that make assertions about the IR produced by the optimizer.

---

I tested locally that the reproducing code in #3778 failed before this commit and is fixed by this commit, in the hello world.

Even though the bug also exists in 0.6.x, I submit the fix to master first, because testing the fix requires linker unit tests, an infrastructure that is not available in the 0.6.x branch. Once reviewed and merged in master, I'll backport the fix to 0.6.x (without tests).

### DIFF
--- a/javalanglib/src/main/scala/java/lang/ObjectClone.scala
+++ b/javalanglib/src/main/scala/java/lang/ObjectClone.scala
@@ -14,8 +14,6 @@ package java.lang
 
 import scala.scalajs.js
 
-import Utils._
-
 /** Implementation of `java.lang.Object.clone()`.
  *
  *  Called by the hard-coded IR of `java.lang.Object`.
@@ -69,7 +67,10 @@ private[lang] object ObjectClone {
       { (o: js.Object) =>
         val ownKeys = ownKeysFun(o)
         val descriptors = new js.Object
-        forArrayElems(ownKeys) { key =>
+        val len = ownKeys.length
+        var i = 0
+        while (i != len) {
+          val key = ownKeys(i)
           /* Almost equivalent to the JavaScript code
            *   descriptors[key] = Object.getOwnPropertyDescriptor(descriptors, key);
            * except that `defineProperty` will by-pass any existing setter for
@@ -81,6 +82,7 @@ private[lang] object ObjectClone {
             val writable = true
             val value = global.Object.getOwnPropertyDescriptor(o, key)
           })
+          i += 1
         }
         descriptors
       }
--- a/linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
+++ b/linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
@@ -1,0 +1,196 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker
+
+import scala.concurrent._
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.ir.ClassKind
+import org.scalajs.ir.EntryPointsInfo
+import org.scalajs.ir.Definitions._
+import org.scalajs.ir.Trees._
+import org.scalajs.ir.Types._
+
+import org.scalajs.logging._
+
+import org.scalajs.junit.async._
+
+import org.scalajs.linker._
+import org.scalajs.linker.standard._
+
+import org.scalajs.linker.testutils._
+import org.scalajs.linker.testutils.IRAssertions._
+import org.scalajs.linker.testutils.TestIRBuilder._
+
+class OptimizerTest {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  import OptimizerTest._
+
+  /** Generic code for the three methods below.
+   *
+   *  Check that `clone()` is never inlined when the result can be an array,
+   *  in several scenarios.
+   */
+  private def testCloneOnArrayNotInlinedGeneric(
+      customMemberDefs: List[MemberDef]): Future[Unit] = {
+
+    val thisFoo = This()(ClassType("LFoo"))
+    val intArrayTypeRef = ArrayTypeRef("I", 1)
+    val intArrayType = ArrayType(intArrayTypeRef)
+    val anArrayOfInts = ArrayValue(intArrayTypeRef, List(IntLiteral(1)))
+    val newFoo = New(ClassRef("LFoo"), "init___", Nil)
+
+    def callCloneOn(receiver: Tree): Tree =
+      Apply(EAF, receiver, "clone__O", Nil)(AnyType)
+
+    val fooMemberDefs = List(
+        trivialCtor("LFoo"),
+
+        // @noinline def witness(): AnyRef = throw null
+        MethodDef(EMF, "witness__O", Nil, AnyType, Some {
+          Throw(Null())
+        })(EOH.withNoinline(true), None),
+
+        // @noinline def reachClone(): Object = clone()
+        MethodDef(EMF, "reachClone__O", Nil, AnyType, Some {
+          Apply(EAF, thisFoo, "clone__O", Nil)(AnyType)
+        })(EOH.withNoinline(true), None),
+
+        // @noinline def anArray(): Array[Int] = Array(1)
+        MethodDef(EMF, "anArray__AI", Nil, intArrayType, Some {
+          anArrayOfInts
+        })(EOH.withNoinline(true), None),
+
+        // @noinline def anObject(): AnyRef = Array(1)
+        MethodDef(EMF, "anObject__O", Nil, AnyType, Some {
+          anArrayOfInts
+        })(EOH.withNoinline(true), None)
+    ) ::: customMemberDefs
+
+    val classDefs = Seq(
+        classDef("LFoo",
+            superClass = Some(ObjectClass),
+            interfaces = List("jl_Cloneable"),
+            memberDefs = fooMemberDefs
+        ),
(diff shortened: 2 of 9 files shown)


gzm0:
What does CHA mean?

sjrd:
Class Hierarchy Analysis. It's the analysis that finds all the potential instantiated subclasses of a class/interface, and subsequently is used to decide what are the possible targets of a call.

linker/shared/src/test/scala/org/scalajs/linker/testutils/IRAssertions.scala
```diff
@@ -0,0 +1,158 @@
...
+        if (matches)
+          actualCount += 1
+        matches
+      })
+      tr.traverse(selfNode)
+      assertEquals(s"AST has the wrong number of $trgName", count, actualCount)
+      this
+    }
+
+    def traverse(pf: Pat): this.type = {
```
gzm0: When is this used?

linker/shared/src/test/scala/org/scalajs/linker/testutils/IRAssertions.scala
```diff
@@ -0,0 +1,158 @@
...
+          for (memberDef <- node.exportedMembers)
+            traverseMemberDef(memberDef.value)
+          for (topLevelExportDef <- node.topLevelExports)
+            traverseTopLevelExportDef(topLevelExportDef.value)
+        }
+      }
+    }
+  }
+
+  abstract class TestTraverser[T](f: IRNode => Boolean) extends Traverser {
```
gzm0: Could we abstract this into a traverser and a finder? (traverser takes a `IRNode => Unit`, finder uses a traverser + control throwable). I think this will completely avoid the `finding` state.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
```diff
@@ -24,6 +26,8 @@ object TestIRBuilder {
   implicit val noPosition: ir.Position = ir.Position.NoPosition
 
   val EAF = ApplyFlags.empty
+  val EMF = MemberFlags.empty
+  val EOH = OptimizerHints.empty
 
   val emptyOptHints: OptimizerHints = OptimizerHints.empty
```
gzm0: There are a bunch of places in the test suite where these can be used (n.b. `LinkerTest`). Also note that `EOH` and `emptyOptHints` are exactly the same.

I'd recommend making a separate commit where you introduce these, and replace ad-hoc usage everywhere. And in the next commit you actually fix the issue.

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
```diff
@@ -64,4 +84,7 @@ object TestIRBuilder {
 
   def mainModuleInitializers(moduleClassName: String): List[ModuleInitializer] =
     ModuleInitializer.mainMethodWithArgs(moduleClassName, "main") :: Nil
+
+  implicit def string2ident(name: String): Ident =
+    Ident(name)
```
gzm0: If you are introducing this, please replace occurrences of `Ident("...")` (again, possibly in a separate commit).

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
```diff
@@ -0,0 +1,246 @@
...
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  import OptimizerTest._
+
+  /** Generic code for the three methods below.
+   *
+   *  Check that `clone()` is never inlined when the result can be an array,
+   *  in several scenarios.
+   */
+  private def testCloneOnArrayNotInlinedGeneric(
```
gzm0: I'm unclear about what this method abstracts. If it is about mere duplication, why are the definitions of `reachClone`, `anArray`, `anObject` and `trivialCtor("LFoo")` not in here?

javalanglib/src/main/scala/java/lang/ObjectClone.scala
```diff
@@ -81,6 +82,7 @@ private[lang] object ObjectClone {
             val writable = true
             val value = global.Object.getOwnPropertyDescriptor(o, key)
           })
+          i += 1
```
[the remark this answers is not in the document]
gzm0: Actually, now I do. Because we put it in the minilib.

gzm0:
> Class Hierarchy Analysis. It's the analysis that finds all the potential instantiated subclasses of a class/interface, and subsequently is used to decide what are the possible targets of a call.

Ah :) Do you think we could spell it out?

linker/shared/src/test/scala/org/scalajs/linker/testutils/IRAssertions.scala
sjrd: Nowhere. Removed.

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
sjrd: Because I'm stupid ... I didn't realize that all those methods were completely duplicated. :sweat: 

Fixed.

sjrd:
> Ah :) Do you think we could spell it out?

It's a pretty standard concept and acronym in optimization circles. But sure, I spelled it out.

linker/shared/src/test/scala/org/scalajs/linker/testutils/IRAssertions.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  type Pat = PartialFunction[IRNode, Boolean]
+
+  private def patToTotal(pat: Pat): IRNode => Boolean =
+    node => pat.applyOrElse(node, (_: IRNode) => false)
+
+  abstract class AbstractIRNodeAssertions[T](selfNode: T) {
+    protected def newTraverser(f: IRNode => Unit): TestTraverser[T]
+
+    protected def newFinder(f: IRNode => Boolean): TestFinder[T] =
```
gzm0: Any reason this is not private?

Also, consider calling this `find` and basically put everything in here (including `patToTotal`).

PR #3786, opened 2019-09-28
Add js.BigInt for ES2019 (ES10)

exoego:
JS's BigInt is available in popular JS runtime like Chrome, Firefox, and Node.js.
Its specification reached stage 4, so I think it is good time to add support it in Scala.js.
https://github.com/tc39/ecma262/pull/1515
https://github.com/tc39/proposals/pull/217

TODO:
- [x] BigInt
- [x] BigInt64Array
- [x] BigUint64Array

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/BigInt.scala
+++ b/library/src/main/scala/scala/scalajs/js/BigInt.scala
@@ -1,0 +1,125 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A built-in object that provides a way to represent whole numbers larger than
+ *  2 ^ 53 - 1, which is the largest number JavaScript can reliably represent
+ *  with the Number primitive.
+ *
+ *  BigInt can be used for arbitrarily large integers.
+ *
+ *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+ */
+@js.native
+@JSGlobal
+final class BigInt private[this] () extends js.Object {
+
+  def +(other: BigInt): BigInt = js.native
+  def *(other: BigInt): BigInt = js.native
+  def -(other: BigInt): BigInt = js.native
+  def %(other: BigInt): BigInt = js.native
+
+  def &(other: BigInt): BigInt = js.native
+  def |(other: BigInt): BigInt = js.native
+  def ^(other: BigInt): BigInt = js.native
+  def <<(other: BigInt): BigInt = js.native
+  def >>(other: BigInt): BigInt = js.native
+  // no >>> since BigInt is always signed
+
+  // scalastyle:off disallow.space.before.token
+  def unary_- : BigInt = js.native
+  def unary_~ : BigInt = js.native
+  // unary_+ is not supported by BigInts
+  // scalastyle:on disallow.space.before.token
+
+  def <(x: BigInt): Boolean = js.native
+  def <=(x: BigInt): Boolean = js.native
+  def >(x: BigInt): Boolean = js.native
+  def >=(x: BigInt): Boolean = js.native
+
+  /** Returns a localized string representation of this BigInt.
+   *
+   *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString
+   *  @param locale A string with a BCP 47 language tag, or an array of such
+   *                strings. For the general form and interpretation of the
+   *                locales argument, see the Intl page.
+   *  @param options An object with some or all of the properties.
+   */
+  def toLocaleString(locale: String,
+      options: BigInt.ToLocaleStringOptions = js.native): String = js.native
+
+  /** Returns a string representation of this BigInt.
+   *
+   *  The trailing "n" is not part of the string.
+   *
+   *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toString
+   */
+  override def toString(): String = js.native
+}
+
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A companion object of BigInt class.
+ *
+ *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+ */
+@js.native
+@JSGlobal
+object BigInt extends js.Object {
+  /** Returns a BigInt from a whole Double.
+   *
+   *  This overload is exposed to allow users to create a BigInt from a Double
+   *  which is greater than Int.MaxValue.
+   *
+   *  @throws RangeError if value has a fractional portion
+   *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+   */
+  def apply(value: Double): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt */
+  def apply(value: Int): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt */
+  def apply(value: String): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asIntN */
+  def asIntN(width: Int, bigint: BigInt): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN */
+  def asUintN(width: Int, bigint: BigInt): BigInt = js.native
+
+  // scalastyle:off line.size.limit
+  /** Type of the `options` parameter of [[BigInt.toLocaleString]].
+   *
+   *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString#Parameters
+   */
+  // scalastyle:on line.size.limit
+  trait ToLocaleStringOptions extends Object {
+    var localeMatcher: js.UndefOr[String] = js.undefined
+    var style: js.UndefOr[String] = js.undefined
+    var currency: js.UndefOr[String] = js.undefined
+    var currencyDisplay: js.UndefOr[String] = js.undefined
+    var useGrouping: js.UndefOr[Boolean] = js.undefined
+    var minimumIntegerDigits: js.UndefOr[Int] = js.undefined
+    var minimumFractionDigits: js.UndefOr[Int] = js.undefined
+    var maximumFractionDigits: js.UndefOr[Int] = js.undefined
+    var minimumSignificantDigits: js.UndefOr[Int] = js.undefined
+    var maximumSignificantDigits: js.UndefOr[Int] = js.undefined
+  }
+}
--- a/library/src/main/scala/scala/scalajs/js/typedarray/BigInt64Array.scala
+++ b/library/src/main/scala/scala/scalajs/js/typedarray/BigInt64Array.scala
@@ -1,0 +1,51 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.scalajs.js.typedarray
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A [[TypedArray]] of signed 64-bit integers represented as [[js.BigInt]].
+ */
+@js.native
+@JSGlobal
+class BigInt64Array private[this] () extends TypedArray[js.BigInt, BigInt64Array] {
+
+  /** Constructs a BigInt64Array with the given length. Initialized to all 0 */
+  def this(length: Int) = this()
+
+  /** Creates a new BigInt64Array with the same elements than the given TypedArray
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(typedArray: TypedArray[js.BigInt, _]) = this()
+
+  /** Creates a new BigInt64Array with the elements in the given array.
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(array: js.Iterable[js.BigInt]) = this()
+
+  /** Creates a BigInt64Array view on the given ArrayBuffer */
+  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = js.native) = this()
+
+}
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *  [[BigInt64Array]] companion
+ */
+@js.native
+@JSGlobal
+object BigInt64Array extends TypedArrayStatic[js.BigInt, BigInt64Array]
--- a/library/src/main/scala/scala/scalajs/js/typedarray/BigUint64Array.scala
+++ b/library/src/main/scala/scala/scalajs/js/typedarray/BigUint64Array.scala
@@ -1,0 +1,53 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.scalajs.js.typedarray
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A [[TypedArray]] of unsigned 64-bit integers represented as [[js.BigInt]].
+ */
+@js.native
+@JSGlobal
+class BigUint64Array private[this] ()
+    extends TypedArray[js.BigInt, BigUint64Array] {
+
+  /** Constructs a BigUint64Array with the given length. Initialized to all 0 */
+  def this(length: Int) = this()
+
+  /** Creates a new BigInt64Array with the same elements than the given TypedArray
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(typedArray: TypedArray[js.BigInt, _]) = this()
+
+  /** Creates a new BigInt64Array with the elements in the given array.
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(array: js.Iterable[js.BigInt]) = this()
+
+  /** Creates a BigInt64Array view on the given ArrayBuffer */
+  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = js.native) = this()
+
+}
+
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *  [[BigUint64Array]] companion
+ */
+@js.native
+@JSGlobal
+object BigUint64Array extends TypedArrayStatic[js.BigInt, BigUint64Array]
--- a/library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
+++ b/library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
@@ -44,6 +44,7 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
   @JSBracketAccess
   def set(index: Int, value: T): Unit = js.native
 
+  // FIXME: Not type-safe. BigInt*Array should accept only BigInt, other TypedArray accept only non-BigInt number.
   /** Set the values of typedArray in this TypedArray */
   def set(typedArray: TypedArray[_, _]): Unit = js.native
 
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
@@ -1,0 +1,134 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.library
+
+import org.junit.Assert._
+import org.junit.Assume._
+import org.junit.{BeforeClass, Test}
+import org.scalajs.testsuite.utils.Platform
+
+import scala.scalajs.js
+
+object BigIntTest {
+  @BeforeClass
+  def assumeRuntimeSupportsBigInt(): Unit = {
+    assumeTrue(Platform.executingInNodeJS)
+    val bigIntConstructor =
+      js.Dynamic.global.BigInt.asInstanceOf[js.UndefOr[js.Any]]
+    assumeTrue(bigIntConstructor.isDefined)
+  }
+}
+
+class BigIntTest {
+
+  @Test def apply(): Unit = {
+    val fromString = js.BigInt("9007199254740992")
+    assertEquals(fromString.toString(), "9007199254740992")
+
+    val fromInt = js.BigInt(2147483647)
+    assertEquals(fromInt.toString(), "2147483647")
+
+    val fromDouble = js.BigInt(4294967295d)
+    assertEquals(fromDouble.toString(), "4294967295")
+  }
+
+  @Test def asIntN(): Unit = {
+    val x = js.BigInt.asIntN(8, js.BigInt("256"))
+    assertEquals(x, js.BigInt("0"))
+  }
+
+  @Test def asUintN(): Unit = {
+    val x = js.BigInt(-123)
+    assertEquals(js.BigInt.asUintN(50, x), js.BigInt("1125899906842501"))
+  }
+
+  @Test def toLocaleString(): Unit = {
+    assumeTrue(Platform.executingInNodeJS)
+
+    val bi = js.BigInt("42123456789123456789")
+    assertEquals(bi.toString(), "42123456789123456789")
+
+    val currency = bi
+      .toLocaleString("de-DE", new js.BigInt.ToLocaleStringOptions {
+        style = "currency"
+        currency = "EUR"
+      })
+    // Internationalization support is limited by default in Node.js.
+    // https://nodejs.org/api/intl.html#intl_internationalization_support
+    // toLocaleString may return as same as by toString
+    assertTrue(
+        currency == "€ 42,123,456,789,123,456,789.00" ||
+        currency == "42123456789123456789")
+  }
+
+  @Test def valueOf(): Unit = {
+    val bi = js.BigInt("42123456789123456789")
+    assertEquals(bi, bi.valueOf())
+  }
+
+  @Test def operators(): Unit = {
+    val previousMaxSafe = js.BigInt("9007199254740991")
+    assertEquals(previousMaxSafe, js.BigInt("9007199254740991"))
+
+    val maxPlusOne = previousMaxSafe + js.BigInt(1)
+    assertEquals(maxPlusOne, js.BigInt("9007199254740992"))
+
+    val theFuture = previousMaxSafe + js.BigInt(2)
+    assertEquals(theFuture, js.BigInt("9007199254740993"))
+
+    val multi = previousMaxSafe * js.BigInt(2)
+    assertEquals(multi, js.BigInt("18014398509481982"))
+
+    val subtr = multi - js.BigInt(10)
+    assertEquals(subtr, js.BigInt("18014398509481972"))
+
+    val mod = multi % js.BigInt(10)
+    assertEquals(mod, js.BigInt("2"))
+
+    // TODO: Scala.js does not recongnize ** as operator
+    //    val bigN = js.BigInt(2) ** js.BigInt(54)
+    //    assertEquals(bigN, js.BigInt("18014398509481984"))
+
+    val negative = -js.BigInt("18014398509481984")
+    assertEquals(negative, js.BigInt("-18014398509481984"))
+
+    val bitAnd = js.BigInt(123) & js.BigInt(31)
+    assertEquals(bitAnd, js.BigInt("27"))
+
+    val bitOr = js.BigInt(123) | js.BigInt(31)
+    assertEquals(bitOr, js.BigInt("127"))
+
+    val bitXor = js.BigInt(123) ^ js.BigInt(31)
+    assertEquals(bitXor, js.BigInt("100"))
+
+    val bitLeftShift = js.BigInt(123) << js.BigInt(31)
+    assertEquals(bitLeftShift, js.BigInt("264140488704"))
+
+    val bitRightShift = js.BigInt(12345678) >> js.BigInt(9)
+    assertEquals(bitRightShift, js.BigInt("24112"))
+
+    val bitNot = ~js.BigInt("42")
+    assertEquals(bitNot, js.BigInt("-43"))
+  }
+
+  @Test def compare_with_bigint(): Unit = {
+    val n = js.BigInt("42")
+    assertTrue(n == js.BigInt("42"))
+    assertTrue(n != js.BigInt("43"))
+    assertTrue(n > js.BigInt("41"))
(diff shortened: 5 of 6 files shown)


sjrd:
Could you try amending the commit (without changes) and force-push the branch again? It seems the CI did not pick up your PR.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * A built-in object that provides a way to represent whole numbers larger than
+ * 2 ^ 53 - 1, which is the largest number JavaScript can reliably represent
+ * with the Number primitive.
+ *
+ * BigInt can be used for arbitrarily large integers.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+ */
```
gzm0: This will need a badge to make it clear it is not ES5.1 compliant.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+@js.native
+@JSGlobal
+class BigInt private[this] () extends js.Object {
```
sjrd: Consider making this class `final`. It's already effectively final because the constructor is `private`, but it's better to be explicit.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  def +(other: BigInt): BigInt = js.native
+  def *(other: BigInt): BigInt = js.native
+  def -(other: BigInt): BigInt = js.native
+  def **(other: BigInt): BigInt = js.native
```
sjrd: This operator won't work, because Scala.js does not recognize `**` as a JavaScript operator at the moment.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+  def %(other: BigInt): BigInt = js.native
+
+  def &(other: BigInt): BigInt = js.native
+  def |(other: BigInt): BigInt = js.native
+  def ^(other: BigInt): BigInt = js.native
+  def <<(other: BigInt): BigInt = js.native
+  def >>(other: BigInt): BigInt = js.native
+  def >>>(other: BigInt): BigInt = js.native
+  // no >>>
```
sjrd: You have a "no `>>>`" comment, but you did define `def >>>` ;)

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+  def unary_~ : BigInt = js.native
+  // no unary_+
+
+  def ==(x: BigInt): Boolean = js.native
+  def ==(x: Short): Boolean = js.native
+  def ==(x: Char): Boolean = js.native
+  def ==(x: Int): Boolean = js.native
+  def ==(x: Long): Boolean = js.native
+  def ==(x: Float): Boolean = js.native
+  def ==(x: Double): Boolean = js.native
```
sjrd: Overloading `==` and `!=` is generally considered bad practice. I don't think there is a strong reason to do it here, so I would remove the overloads.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+  def !=(x: Char): Boolean = js.native
+  def !=(x: Int): Boolean = js.native
+  def !=(x: Long): Boolean = js.native
+  def !=(x: Float): Boolean = js.native
+  def !=(x: Double): Boolean = js.native
+
+  def <(x: BigInt): Boolean = js.native
+  def <(x: Byte): Boolean = js.native
+  def <(x: Short): Boolean = js.native
+  def <(x: Char): Boolean = js.native
```
sjrd: Comparisons with `Char`s and `Long`s will not work. They should not be declared.

In general, I don't think we should expose the comparisons with other types than `BigInt` and `Double`. And even `Double` is border-line, because we're not exposing the converse comparisons where `Double` is on the left and `BigInt` is on the right. I believe it's better to leave them out.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+    maximumFractionDigits.foreach(_v => obj.updateDynamic("maximumFractionDigits")(_v))
+    minimumSignificantDigits.foreach(_v => obj.updateDynamic("minimumSignificantDigits")(_v))
+    maximumSignificantDigits.foreach(_v => obj.updateDynamic("maximumSignificantDigits")(_v))
+    obj.asInstanceOf[BigIntOptions]
+  }
+}
+
+@js.native
+@JSGlobal
+object BigInt extends js.Object {
```
sjrd: The companion object should also have a Scaladoc comment with at the very least the ECMAScript 2020 badge.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+   */
+  def apply(value: String): BigInt = js.native
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+   */
+  def apply(value: BigInt): BigInt = js.native
```
sjrd: Even though this works, it is silly to call this overload as it is basically an identity. I would not expose it.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
+package scala.scalajs.js
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/**
+ * A built-in object that provides a way to represent whole numbers larger than
```
sjrd: Please follow our coding style for Scaladoc comments, as explained at https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#scaladoc

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+   */
+  override def valueOf(): BigInt = js.native
+}
+
+/**
+ * An object with some or all of the following properties.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString#Parameters
+ */
+trait BigIntOptions extends Object {
```
sjrd: This should be inside the `object BigInt` and be called `ToLocaleStringOptions`. Those options are very specific to `toLocaleString`.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+  var useGrouping: js.UndefOr[Boolean] = js.undefined
+  var minimumIntegerDigits: js.UndefOr[Int] = js.undefined
+  var minimumFractionDigits: js.UndefOr[Int] = js.undefined
+  var maximumFractionDigits: js.UndefOr[Int] = js.undefined
+  var minimumSignificantDigits: js.UndefOr[Int] = js.undefined
+  var maximumSignificantDigits: js.UndefOr[Int] = js.undefined
+}
+
+object BigIntOptions {
+  def apply(
```
sjrd: I would not expose this, as it cannot be evolved in a binary compatible way. The above `trait` with the `var`s is enough for users to instantiate the options using
```scala
val opts = new BigInt.ToLocalStringOptions {
  style = "whatever"
  ...
}
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
```diff
@@ -0,0 +1,137 @@
...
+
+package org.scalajs.testsuite.library
+
+import org.junit.Assert._
+import org.junit.Test
+import org.scalajs.testsuite.utils.JSAssert.assertJSUndefined
+
+import scala.scalajs.js
+
+class JsBigIntTest {
```
sjrd: This test class will need a companion object with an `@BeforeClass` def to assume that we're running with Node.js or in the browser, as neither Rhino nor PhantomJS support BigInts.

sjrd: (also it should just be called `BigIntTest`, not `JsBigIntTest`, since it contains the tests for the class `BigInt`, not `JsBigInt`.)

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  // Methods we actually implement
+
+  @Test def apply(): Unit = {
+    val bi = js.BigInt("9007199254740992")
+    assertEquals(bi.toString(), "90071992547409912")
+  }
+
+  @Test def asIntN(): Unit = {
+    val x = js.BigInt(2) ** js.BigInt(64)
```
sjrd: I am really surprised that this could work at all on your side. `**` is not supported as a JS operator at the moment. How did this happen?

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,188 @@
...
+   * @return A string representing the specified BigInt object.
+   *         The trailing "n" is not part of the string.
+   */
+  override def toString(): String = js.native
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/valueOf
+   * @return The wrapped primitive value of a BigInt object.
+   */
+  override def valueOf(): BigInt = js.native
```
sjrd: `valueOf()` is defined in JavaScript for obscure reasons related to how primitive operators work. It is not a useful method to call in user space, as it is basically an identity. So I would not expose it.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
```diff
@@ -0,0 +1,137 @@
...
+  @Test def toLocaleString(): Unit = {
+    val bi = js.BigInt("42123456789123456789")
+    assertEquals(bi.toString(), "90071992547409912")
+
+    val currency = js.BigInt("42123456789123456789")
+      .toLocaleString("de-DE", js.BigIntOptions(
+        style = "currency",
+        currency = "EUR"
+      ))
+    // Options may not be supported
```
sjrd: Could you point me to the part of the spec that says that this may not be supported?

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
```diff
@@ -0,0 +1,137 @@
...
+        || currency == "42.123.456.789.123.456.789,00 €")
+  }
+
+  @Test def valueOf(): Unit = {
+    val bi = js.BigInt("42123456789123456789")
+    assertEquals(bi, bi.valueOf())
+  }
+
+  @Test def operators(): Unit = {
+    val previousMaxSafe = js.BigInt("9007199254740991");
```
sjrd: You have a bunch of stray `;` here ;)

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
exoego: I mean this is not spec issue, but missing implementation.
As of 2019-09-30, only Chrome (V8) support these options and format like "123,456,789 €", but Firefox outputs like "123456789".

sjrd: Then the correct way to write the test is to assert the proper result, but `assume` that we are running in Node.js.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
exoego: `**` is [exponentiation operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Exponentiation).
```
2 ** 3 === 8
2n ** 3n === 8n
```

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: I agree to expose comparison only for `BigInt` and `Double`.
But I do not understand why `Double op BigInt` should not be exposed.

sjrd: Well we could expose `Double op BigInt`, yes, but currently your PR doesn't do that. It only exposes `BigInt op Double`. I'm against exposing one direction but not the other. It should be "commutative".

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
sjrd: I know that `**` is an ECMAScript operator. But it's not supported by Scala.js. Maybe it should, especially if there is no other way to perform `BigInt` exponentiation. But right now it doesn't, so how did this test pass for you locally at all?

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
exoego: Removed in https://github.com/scala-js/scala-js/pull/3786/commits/2cd388f224fff3c4aac0e44f1bd8f2373b314c69

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Added `@note`

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Where is a place Scala.js handle JS operators ?

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Added in https://github.com/scala-js/scala-js/pull/3786/commits/4e0489fc0c5605c937f5b3aec620cdde861e203b

library/src/main/scala/scala/scalajs/js/BigInt.scala
sjrd: That's quite complicated because it requires to change the IR. It spans the entire pipeline from the IR definition to the compiler backend to the optimizer to the emitter. It probably deserves a different PR anyway, so let's try to do everything else in this PR and leave that for later.

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Added css-styed badge.

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: I will leave it for later PR.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/JsBigIntTest.scala
exoego: It seems that current Node.js contains limited locales so `toLocaleString` does not undersntand `currency` or other options.
I think having leaving `o == "123,456,789 €" || o == "123456789"` is fine for current Node.js and  future Node.js that contains all locales.
What do you think?

sjrd: Node.js v12 has enough support for that test to pass, and should enter Active LTS status on October 22nd. I'll check if we can install that on our CI servers, and consider that version required to test Scala.js itself, which would allow us not to have this suboptimal test.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+  def |(other: BigInt): BigInt = js.native
+  def ^(other: BigInt): BigInt = js.native
+  def <<(other: BigInt): BigInt = js.native
+  def >>(other: BigInt): BigInt = js.native
+  // no >>> since BigInt is always signed
+
+  // scalastyle:off disallow.space.before.token
+  def unary_- : BigInt = js.native
+  def unary_~ : BigInt = js.native
+  // no unary_+ not to break asm.js
```
sjrd: The rationale "not to break asm.js" is meaningful in the spec of `BigInt`s, but here it is misleading as it suggests that *Scala.js* doesn't want to break asm.js and therefore does not expose `unary_+`.
```suggestion
  // unary_+ is not supported by BigInts
```

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+  // scalastyle:on disallow.space.before.token
+
+  def <(x: BigInt): Boolean = js.native
+  def <=(x: BigInt): Boolean = js.native
+  def >(x: BigInt): Boolean = js.native
+  def >=(x: BigInt): Boolean = js.native
+
+  /**
```
sjrd: There should be *something* here, even if it is as trivial as

    Returns a localized string representation of this BigInt.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString
+   * @param locale A string with a BCP 47 language tag, or an array of such
+   *               strings. For the general form and interpretation of the
+   *               locales argument, see the Intl page.
+   * @param options An object with some or all of the properties.
+   * @return
```
sjrd: These should have 2 more spaces of indentation before the `@` (after the `*` which are correctly indented).

Also the `@return` should be removed since it is empty.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+   */
+  def toLocaleString(locale: String, options:
+      BigInt.ToLocaleStringOptions = js.native): String = js.native
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toString
+   * @return A string representing the specified BigInt object.
+   *         The trailing "n" is not part of the string.
```
sjrd: The contents of this `@return` would be better used as the regular description of the method.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+ *
+ *  A companion object of BigInt class.
+ *
+ *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+ */
+@js.native
+@JSGlobal
+object BigInt extends js.Object {
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt */
+  def apply(value: Int): BigInt = js.native
```
sjrd: I think we should also expose one more overload that takes a `Double`. It won't be completely "safe", since fractional numbers will cause a `RangeError`, but if we *don't* expose it, then there is no way to create a `BigInt` from a whole `Double` that is greater than `Int.MaxValue`.

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,116 @@
...
+  def apply(value: String): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asIntN */
+  def asIntN(width: Int, bigint: BigInt): BigInt = js.native
+
+  /** @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN */
+  def asUintN(width: Int, bigint: BigInt): BigInt = js.native
+
+  // scalastyle:off line.size.limit
+  /** An object with some or all of the following properties.
```
sjrd:
```suggestion
  /** Type of the `options` parameter of [[BigInt.toLocaleString]].
```
That would be a little bit less useless.

library/src/main/scala/scala/scalajs/js/typedarray/BigInt64Array.scala
```diff
@@ -0,0 +1,51 @@
...
+  def this(typedArray: TypedArray[js.BigInt, _]) = this()
+
+  /** Creates a new BigInt64Array with the elements in the given array.
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(array: js.Iterable[js.BigInt]) = this()
+
+  /** Creates a BigInt64Array view on the given ArrayBuffer */
+  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = ???) = this()
```
sjrd:
```suggestion
  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = js.native) = this()
```

library/src/main/scala/scala/scalajs/js/typedarray/BigUint64Array.scala
```diff
@@ -0,0 +1,53 @@
...
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A [[TypedArray]] of unsigned [[js.BigInt]]
+ */
+@js.native
+@JSGlobal
+class BigUint64Array private[this] ()
+    extends TypedArray[js.BigInt, BigUint64Array] {
+
+  /** Constructs a BigInt64Array with the given length. Initialized to all 0 */
```
sjrd:
```suggestion
  /** Constructs a BigUint64Array with the given length. Initialized to all 0 */
```
Same in the other constructors below.

library/src/main/scala/scala/scalajs/js/typedarray/BigUint64Array.scala
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package scala.scalajs.js.typedarray
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A [[TypedArray]] of unsigned [[js.BigInt]]
```
sjrd:
```suggestion
 *  A [[TypedArray]] of unsigned 64-bit integers represented as [[js.BigInt]].
```

library/src/main/scala/scala/scalajs/js/typedarray/BigInt64Array.scala
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package scala.scalajs.js.typedarray
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** <span class="badge badge-ecma2019" style="float: right;">ECMAScript 2019</span>
+ *
+ *  A [[TypedArray]] of signed [[js.BigInt]]
```
sjrd:
```suggestion
 *  A [[TypedArray]] of signed 64-bit integers represented as [[js.BigInt]].
```

library/src/main/scala/scala/scalajs/js/typedarray/BigUint64Array.scala
```diff
@@ -0,0 +1,53 @@
...
+  def this(typedArray: TypedArray[js.BigInt, _]) = this()
+
+  /** Creates a new BigInt64Array with the elements in the given array.
+   *
+   *  Each elements must be BigInt (no conversion).
+   */
+  def this(array: js.Iterable[js.BigInt]) = this()
+
+  /** Creates a BigInt64Array view on the given ArrayBuffer */
+  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = ???) = this()
```
sjrd:
```suggestion
  def this(buffer: ArrayBuffer, byteOffset: Int = 0, length: Int = js.native) = this()
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.library
+
+import org.junit.Assert._
+import org.junit.Assume._
+import org.junit.{ BeforeClass, Test }
```
sjrd:
```suggestion
import org.junit.{BeforeClass, Test}
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,130 @@
...
+    assumeTrue(Platform.executingInNodeJS)
+    val bigIntConstructor =
+      js.Dynamic.global.BigInt.asInstanceOf[js.UndefOr[js.Any]]
+    assumeTrue(bigIntConstructor.isDefined)
+  }
+}
+
+class BigIntTest {
+
+  // Methods we actually implement
```
sjrd: This comment does not make sense; we don't implement any of these methods, they are natively provided. Please remove it.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,130 @@
...
+  @Test def apply(): Unit = {
+    val bi = js.BigInt("9007199254740992")
+    assertEquals(bi.toString(), "9007199254740992")
+  }
+
+  // TODO: Scala.js does not recongnize ** as operator
+  //  @Test def asIntN(): Unit = {
+  //    val x = js.BigInt(2) ** js.BigInt(64)
+  //    assertEquals(js.BigInt.asIntN(256, x), js.BigInt("18446744073709551616"))
+  //  }
```
sjrd: Looks like this is redundant wrt. to code in `operators()`.

However the name of the commented-out method is `asIntN()` but it tries to test `**`. Since I don't see another test for `asIntN()`, I think this was meant *not* to be commented, and to test `asIntN`.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+class BigIntTest {
+
+  // Methods we actually implement
+
+  @Test def apply(): Unit = {
+    val bi = js.BigInt("9007199254740992")
+    assertEquals(bi.toString(), "9007199254740992")
```
sjrd: Please also test the other overloads of `apply()`.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,130 @@
...
+    val currency = bi
+      .toLocaleString("de-DE", new js.BigInt.ToLocaleStringOptions {
+        style = "currency"
+        currency = "EUR"
+      })
+    // Internationalization support is limited by default in Node.js.
+    // https://nodejs.org/api/intl.html#intl_internationalization_support
+    // toLocaleString may return as same as by toString
+    assertTrue(currency == "€ 42,123,456,789,123,456,789.00" ||
+      currency == "42123456789123456789")
```
sjrd: This should be formatted as follows:
```scala
    assertTrue(
        currency == "€ 42,123,456,789,123,456,789.00" ||
        currency == "42123456789123456789")
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
sjrd: Also, you should swap the arguments to most of your calls to `assertEquals`. The meaning of the arguments is
```scala
assertEquals(expected, actual)
```
This is not just by convention; the error messages reported by JUnit only make sense if the first argument is the expected value, and the second argument is the actual value.

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Added `Double`-overload

library/src/main/scala/scala/scalajs/js/BigInt.scala
```diff
@@ -0,0 +1,125 @@
...
+  /** Returns a localized string representation of this BigInt.
+   *
+   *  @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString
+   *  @param locale A string with a BCP 47 language tag, or an array of such
+   *                strings. For the general form and interpretation of the
+   *                locales argument, see the Intl page.
+   *  @param options An object with some or all of the properties.
+   */
+  def toLocaleString(locale: String, options:
+      BigInt.ToLocaleStringOptions = js.native): String = js.native
```
sjrd: Oops, I think I did not notice this before: the `options:` should be on the next line, together with its type:
```scala
  def toLocaleString(locale: String,
      options: BigInt.ToLocaleStringOptions = js.native): String = js.native
```

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
```diff
@@ -44,6 +44,7 @@ trait TypedArray[T, Repr] extends ArrayBufferView with js.Iterable[T] {
   @JSBracketAccess
   def set(index: Int, value: T): Unit = js.native
 
+  // FIXME: Not type-safe. BigInt*Array should accept only BigInt, other TypedArray accept only non-BigInt number.
```
sjrd: Actually, I think this should be fixed *now*, because it was correct before, but the introduction of the `BigInt` typed arrays makes this incorrect. You can make it correct (although more restrictive than strictly necessary) by using `T` as the first type parameter:
```scala
 def set(typedArray: TypedArray[T, _]): Unit = js.native
```
This won't be backward source compatible, but it is backward binary compatible, so it's fine. It deserves to be mentioned in the commit message though, with something along the lines of

    We restrict the argument type of `TypedArray.set` to have the same
    element type as the target typed array, because `BigInt` typed arrays
    and `number` typed arrays are incompatible. This is a backward source
    incompatible change, but it is binary compatible.

Similarly, I think you should restrict the argument of the constructors of the other typed arrays in a similar way, because it is not valid to give them typed arrays of `BigInt`s.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/BigIntTest.scala
```diff
@@ -0,0 +1,134 @@
...
+      .toLocaleString("de-DE", new js.BigInt.ToLocaleStringOptions {
+        style = "currency"
+        currency = "EUR"
+      })
+    // Internationalization support is limited by default in Node.js.
+    // https://nodejs.org/api/intl.html#intl_internationalization_support
+    // toLocaleString may return as same as by toString
+    assertTrue(
+      currency == "€ 42,123,456,789,123,456,789.00" ||
+      currency == "42123456789123456789")
```
sjrd: Sorry, I probably wasn't very clear the last time. Those two lines should be indented 4 spaces (2 more spaces than now).

exoego: Fixed in https://github.com/scala-js/scala-js/pull/3786/commits/91fd0a6509b6ad566a27a270815bf7e360f979a6

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/3786/commits/9dc3414b54c5e9b4c946902cfc57ad42318e5251 and https://github.com/scala-js/scala-js/pull/3786/commits/7a47c1fbcf1840df72ef595923345421fd983424

library/src/main/scala/scala/scalajs/js/BigInt.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/3786/commits/163d5d94dc7e17ab1aedd7f9db24a4a99e21d8e3

javalib/src/main/scala/java/nio/GenTypedArrayBuffer.scala
```diff
@@ -82,7 +82,7 @@ private[nio] final class GenTypedArrayBuffer[B <: Buffer](
     val typedArray = _typedArray
     val pos = position
     val lim = limit
-    typedArray.set(typedArray.subarray(pos, lim))
+    typedArray.set(typedArray.subarray(pos, lim).asInstanceOf[scala.scalajs.js.Iterable[Nothing]])
```
exoego: I am doing this to leapflog type checking error with less hassle.
Not sure this is fine.

sjrd: I've been playing with this issue, trying to fix up the definition of `TypedArray` to make it work, but it's getting out of hands.

Given that this is not the only problem (there are further compile errors after this), I suggest you revert the change I suggested earlier and go back to the state with the `// FIXME` comment and the wildcards. We might be able to fix it in 1.x with more drastic changes.

exoego: @sjrd 
Reverted the change and left `FIXME` comment along with wildcard.

library/src/main/scala/scala/scalajs/js/typedarray/TypedArray.scala
exoego: This `FIXME` and wildcard are left as intended for 0.6.x to avoid binary incompatibility.
It will be fixed in Scala.js 1.x with some breaking change.

Issue #3714, opened 2019-07-07
Re-structure the build so we do not need source inclusion anymore


gzm0:
Using the techiniques in https://github.com/gzm0/sbtception we can use things without defining their build twice.

This is an alternative to #3661.

gzm0:
I wanted to have a `build` directory and hide most of the things in there. Turns out sbt is quite strict with project directories being sub directories of the root. In practice, this means we'll need to actually move code (unless we want to go even more non-standard).

I see two main options.

### Minimal

Basically a build `core` containing everything up-to and including `sbt-plugin` and a build `full` containing the rest (tests, testing infra, cross builds of stuff in core).

### Semantic

Use this opportunity to group stuff. Rough idea:

- `js-env`: all jsenv stuff including test kit (could also be `core`).
- `core`: compiler plugin, linker, sbt-plugin, libraries, test bridge JVM
- `testing`: jUnit plugin, test bridge JS.
- `tests`: partest, test suite (ex), test suite linker
- `examples`: examples (already here).
- `core-js`: JS build of core stuff.

Both of these options will need an additional `build-common` build containing stuff needed across all builds.

I have a slight preference for the second one, since I believe it will make navigating the SJS source tree easier (not sure compared to now, but definitely compared to `core` and `full` only).

@sjrd WDYT?

sjrd:
> I wanted to have a `build` directory and hide most of the things in there. Turns out sbt is quite strict with project directories being sub directories of the root. In practice, this means we'll need to actually move code (unless we want to go even more non-standard).

Could we not simply have the `project`s themselves be subdirectories of the root, but require their `unmanagedSourceDirectories` to point outside? It seems like this would have minimal changes. We would need something like that for `core-js` anyway, wouldn't we?

sjrd:
Regardless, I would suggest that we keep the number of *builds* to the minimum. So I would advocate for the **Minimal** solution above.

Additionally, I would put `libraries` in the "full" build rather than the "core". Except of course the project that contains only `RuntimeLong` (once it arrives).

gzm0:
> Could we not simply have the projects themselves be subdirectories of the root, but require their unmanagedSourceDirectories to point outside? It seems like this would have minimal changes.

That would mean a non-standard layout and more changes, but yes, that would probably be doable.

> Regardless, I would suggest that we keep the number of builds to the minimum.

Is there any particular reason for that?

[...]

Issue #3661, opened 2019-05-22
Use reflection to load the Scala.js linker in the sbt plugin


gzm0:
Originated here: #3537

https://github.com/scala-js/scala-js/compare/master...sjrd:classload-linker.

PR #3790, opened 2019-09-29
Split linker interface into separate artifact

gzm0:
This is the first step for #3661 / #3714.


sjrd:
The new package structure looks good to me. :)

The `interface.unstable` package makes me wonder whether we should move all `linker.x` subpackages (except `interface` and `standard`) in a subpackage `linker.unstable.x`?

gzm0:
> The interface.unstable package makes me wonder whether we should move all linker.x subpackages (except interface and standard) in a subpackage linker.unstable.x?

Yes, potentially.

gzm0:
This is ready for final review. I think I'll put the actual reflective loading into a separate PR.

We might want to wait with merging until the reflective approach is validated.

linker/shared/src/main/scala/org/scalajs/linker/interface/unstable/OutputFileImpl.scala
```diff
@@ -31,7 +30,17 @@ abstract class OutputFileImpl extends LinkerOutput.File {
         else Future.successful(())
       }
 
-      writeLoop().finallyWith(chan.close())
+      writeLoop()
+      // finallyWith(chan.close())
```
sjrd: Consider extracting that in a
```scala
private def finallyWith[A](f: Future[A], finallyFuture: => Future[Unit]): Future[A]
```
instead of relying on a comment.

linker/js/src/main/scala/org/scalajs/linker/NodeIRContainer.scala
```diff
@@ -23,15 +23,15 @@ import scala.scalajs.js.typedarray.TypedArrayBufferOps._
 
 import java.nio._
 
-import org.scalajs.linker.standard.{IRContainerImpl, IRFileImpl}
+import org.scalajs.linker.interface.{IRContainer, IRFile}
+import org.scalajs.linker.interface.unstable.IRContainerImpl
 import org.scalajs.linker.standard.MemIRFileImpl
 
-abstract class IRContainerPlatformExtensions private[linker] () {
+object NodeIRContainer {
```
sjrd: Shouldn't this refactoring be in its own commit, like the other refactorings in this PR?

Jenkinsfile
```diff
@@ -308,20 +308,22 @@ def Tasks = [
         'set scalaJSStage in testSuite := FastOptStage' \
         ++$scala testSuite/bootstrap:test &&
     sbt ++$scala irJS/mimaReportBinaryIssues \
-        loggingJS/mimaReportBinaryIssues linkerJS/mimaReportBinaryIssues
+        loggingJS/mimaReportBinaryIssues \
+	linkerInterfaceJS/mimaReportBinaryIssues linkerJS/mimaReportBinaryIssues
```
sjrd: There's a tab here, and in quite a few other places in this file.

build.sbt
```diff
@@ -7,7 +7,9 @@ val compiler = Build.compiler
 val logging = Build.logging
 val loggingJS = Build.loggingJS
 val linker = Build.linker
+val linkerInterface = Build.linkerInterface
```
sjrd: For consistency, these declarations should follow the order of declarations in `project/Build.scala`.

gzm0:
FYI, updates are still WIP.

gzm0:
Updated. I think I have addressed everything.

PR #3794, opened 2019-10-01
Fix #3792: Work around a 2.13.1 regression in overload resolution.

### LINKED ISSUE #3792: Compilation failure (IdentityHashMap) on recent Scala 2.13.2 nightly
at e.g. https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/2658/artifact/logs/scala-js-build.log we see:

```
[scala-js] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.16/project-builds/scala-js-7654802acae1122fadcbca880b40e51c7700e137/javalib/src/main/scala/java/util/IdentityHashMap.scala:43: error: ambiguous reference to overloaded definition,
[scala-js] [error] both constructor IdentityHashMap in class IdentityHashMap of type (initialMap: java.util.Map[_ <: K, _ <: V])java.util.IdentityHashMap[K,V]
[scala-js] [error] and  constructor IdentityHashMap in class IdentityHashMap of type (inner: java.util.HashMap[java.util.IdentityHashMap.IdentityBox[K],V])java.util.IdentityHashMap[K,V]
[scala-js] [error] match argument types (java.util.HashMap[java.util.IdentityHashMap.IdentityBox[K],V]) and expected result type AnyRef
[scala-js] [error]     new IdentityHashMap(
[scala-js] [error]     ^
```

Scala version is 2.13.2-bin-59dec61

sjrd:
This commit works around https://github.com/scala/bug/issues/11755. Instead of relying on an arguably sensitive overload, we use an extra argument for explicit disambiguation.

Locally tested with
```scala
> ++2.13.1
> testSuite/test
> compiler/test
```
Irrespective of whether https://github.com/scala/bug/issues/11755 gets fixed upstream and when, we'll need this change before we can enable 2.13.1 on our CI.

### DIFF
--- a/javalib/src/main/scala/java/util/IdentityHashMap.scala
+++ b/javalib/src/main/scala/java/util/IdentityHashMap.scala
@@ -18,7 +18,11 @@ import scala.annotation.tailrec
 
 import ScalaOps._
 
-class IdentityHashMap[K, V](inner: HashMap[IdentityHashMap.IdentityBox[K], V])
+/* The additional `internal` parameter works around
+ * https://github.com/scala/bug/issues/11755
+ */
+class IdentityHashMap[K, V] private (
+    inner: HashMap[IdentityHashMap.IdentityBox[K], V], internal: Boolean)
     extends AbstractMap[K, V] with Map[K, V] with Serializable with Cloneable {
   self =>
 
@@ -26,7 +30,7 @@ class IdentityHashMap[K, V](inner: HashMap[IdentityHashMap.IdentityBox[K], V])
 
   def this(expectedMaxSize: Int) = {
     this(new HashMap[IdentityHashMap.IdentityBox[K], V](
-        expectedMaxSize, HashMap.DEFAULT_LOAD_FACTOR))
+        expectedMaxSize, HashMap.DEFAULT_LOAD_FACTOR), internal = true)
   }
 
   def this() =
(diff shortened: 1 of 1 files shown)


javalib/src/main/scala/java/util/IdentityHashMap.scala
```diff
@@ -18,15 +18,16 @@ import scala.annotation.tailrec
 
 import ScalaOps._
 
-class IdentityHashMap[K, V](inner: HashMap[IdentityHashMap.IdentityBox[K], V])
+class IdentityHashMap[K, V](
+    inner: HashMap[IdentityHashMap.IdentityBox[K], V], internal: Boolean)
```
gzm0: This should be documented as a workaround.

Also, why is this constructor not `private` / `private[java]` / ... if it is internal (I know it gets hidden away because this is only the binary interface, but still).

sjrd: Added a comment, and made the constructor `private`.

sjrd:
> This seems to be a pretty severe regression. Have we considered not supporting 2.13.1?

It's too late for that. 0.6.29 was already successfully published for 2.13.1 (it didn't have `IdentityHashMap` yet).

gzm0:
> It's too late for that. 0.6.29 was already successfully published for 2.13.1 (it didn't have IdentityHashMap yet).

Ah... OK.

PR #3796, opened 2019-10-01
Fix #2433: Configure automatic CLA check as a GitHub Action.

### LINKED ISSUE #2433: Automatic CLA validation in CI

sjrd:
Inspired by https://github.com/scalameta/scalameta/pull/1906


sjrd:
This probably needs to be merged to master before it takes effect. So we can probably not test this before it reaches master ...

PR #3802, opened 2019-10-04
Use a dedicated `PrimRef` for `TypeRef`s of primitive types.

sjrd:
Before, we put primitive types together with class types as `ClassRef`s, using magical "class names" that actually represented primitives. Those special `ClassRef`s for primitives had to be special-cased virtually everywhere, by testing whether the class name was one of the magical names.

Now, we have a dedicated `PrimRef` instead. This simplifies a lot of code paths by using pattern-matching instead of tests for the class name. More interestingly, it allows to completely get rid of the magical class names.

The char codes associated with primitive type refs are now only used by the method signature encodings, and by some emitter-specific code (and although they use the same characters, that is simply by convention; it is not required for correctness).

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
@@ -40,19 +40,19 @@ trait TypeConversions[G <: Global with Singleton] extends SubComponent {
     )
   }
 
-  private lazy val primitiveClassRefNameMap: Map[Symbol, String] = {
+  private lazy val primitiveRefMap: Map[Symbol, Types.NonArrayTypeRef] = {
     Map(
-        UnitClass    -> Definitions.VoidClass,
-        BooleanClass -> Definitions.BooleanClass,
-        CharClass    -> Definitions.CharClass,
-        ByteClass    -> Definitions.ByteClass,
-        ShortClass   -> Definitions.ShortClass,
-        IntClass     -> Definitions.IntClass,
-        LongClass    -> Definitions.LongClass,
-        FloatClass   -> Definitions.FloatClass,
-        DoubleClass  -> Definitions.DoubleClass,
-        NothingClass -> encodeClassFullName(RuntimeNothingClass),
-        NullClass    -> encodeClassFullName(RuntimeNullClass)
+        UnitClass    -> Types.VoidRef,
+        BooleanClass -> Types.BooleanRef,
+        CharClass    -> Types.CharRef,
+        ByteClass    -> Types.ByteRef,
+        ShortClass   -> Types.ShortRef,
+        IntClass     -> Types.IntRef,
+        LongClass    -> Types.LongRef,
+        FloatClass   -> Types.FloatRef,
+        DoubleClass  -> Types.DoubleRef,
+        NothingClass -> Types.ClassRef(encodeClassFullName(RuntimeNothingClass)),
+        NullClass    -> Types.ClassRef(encodeClassFullName(RuntimeNullClass))
     )
   }
 
@@ -67,16 +67,16 @@ trait TypeConversions[G <: Global with Singleton] extends SubComponent {
   def toTypeRef(t: Type): Types.TypeRef = {
     val (base, arrayDepth) = convert(t)
     if (arrayDepth == 0)
-      Types.ClassRef(makeClassRefName(base))
+      makeNonArrayTypeRef(base)
     else
       makeArrayTypeRef(base, arrayDepth)
   }
 
-  private def makeClassRefName(sym: Symbol): String =
-    primitiveClassRefNameMap.getOrElse(sym, encodeClassFullName(sym))
+  private def makeNonArrayTypeRef(sym: Symbol): Types.NonArrayTypeRef =
+    primitiveRefMap.getOrElse(sym, Types.ClassRef(encodeClassFullName(sym)))
 
   private def makeArrayTypeRef(base: Symbol, depth: Int): Types.ArrayTypeRef =
-    Types.ArrayTypeRef(makeClassRefName(base), depth)
+    Types.ArrayTypeRef(makeNonArrayTypeRef(base), depth)
 
   // The following code was modeled after backend.icode.TypeKinds.toTypeKind
 
--- a/ir/src/main/scala/org/scalajs/ir/Definitions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Definitions.scala
@@ -12,39 +12,13 @@
 
 package org.scalajs.ir
 
+import scala.annotation.switch
+
 import Types._
 
 object Definitions {
   /** `java.lang.Object`, the root of the class hierarchy. */
   val ObjectClass = "O"
-
-  // Primitive "classes"
-  val VoidClass = "V"
-  val BooleanClass = "Z"
-  val CharClass = "C"
-  val ByteClass = "B"
-  val ShortClass = "S"
-  val IntClass = "I"
-  val LongClass = "J"
-  val FloatClass = "F"
-  val DoubleClass = "D"
-  val NullClass = "N"
-  val NothingClass = "E" // think "the Empty type", or "throws an Exception"
-
-  /** The set of all primitive classes. */
-  val PrimitiveClasses: Set[String] = Set(
-      VoidClass,
-      BooleanClass,
-      CharClass,
-      ByteClass,
-      ShortClass,
-      IntClass,
-      LongClass,
-      FloatClass,
-      DoubleClass,
-      NullClass,
-      NothingClass
-  )
 
   // Hijacked classes
   val BoxedUnitClass = "jl_Void"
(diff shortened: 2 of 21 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/TypeConversions.scala
```diff
@@ -40,19 +40,19 @@ trait TypeConversions[G <: Global with Singleton] extends SubComponent {
...
+        BooleanClass -> Types.BooleanRef,
+        CharClass    -> Types.CharRef,
+        ByteClass    -> Types.ByteRef,
+        ShortClass   -> Types.ShortRef,
+        IntClass     -> Types.IntRef,
+        LongClass    -> Types.LongRef,
+        FloatClass   -> Types.FloatRef,
+        DoubleClass  -> Types.DoubleRef,
+        NothingClass -> Types.ClassRef(encodeClassFullName(RuntimeNothingClass)),
+        NullClass    -> Types.ClassRef(encodeClassFullName(RuntimeNullClass))
```
gzm0: I know this is already here, but I don't understand why this is not simply the primitive type ref here...

ir/src/main/scala/org/scalajs/ir/Definitions.scala
```diff
@@ -189,21 +163,53 @@ object Definitions {
...
+      case ArrayTypeRef(base, dimensions) =>
+        val encodedBase = encodeTypeRef(base)
+        val builder = new java.lang.StringBuilder(dimensions + encodedBase.length())
+        var i = 0
+        while (i != dimensions) {
+          builder.append('A')
+          i += 1
+        }
+        builder.append(encodedBase)
+        builder.toString()
```
gzm0: Why not simply `"A" * dimensions + encodeTypeRef(base)`?

ir/src/main/scala/org/scalajs/ir/Definitions.scala
```diff
@@ -189,21 +163,53 @@ object Definitions {
...
     }
   }
 
   /** Decodes a [[Types.TypeRef]], such as in an encoded method signature.
    */
   def decodeTypeRef(encodedName: String): TypeRef = {
-    val arrayDepth = encodedName.indexWhere(_ != 'A')
+    var arrayDepth = 0
+    while (encodedName.charAt(arrayDepth) == 'A')
+      arrayDepth += 1
```
gzm0: Why is this changed?

ir/src/main/scala/org/scalajs/ir/Types.scala
```diff
@@ -163,17 +163,41 @@ object Types {
...
+  final val BooleanRef = PrimRef(BooleanType)("Z", Tags.TagBooleanRef)
+  final val CharRef = PrimRef(CharType)("C", Tags.TagCharRef)
+  final val ByteRef = PrimRef(ByteType)("B", Tags.TagByteRef)
+  final val ShortRef = PrimRef(ShortType)("S", Tags.TagShortRef)
+  final val IntRef = PrimRef(IntType)("I", Tags.TagIntRef)
+  final val LongRef = PrimRef(LongType)("J", Tags.TagLongRef)
+  final val FloatRef = PrimRef(FloatType)("F", Tags.TagFloatRef)
+  final val DoubleRef = PrimRef(DoubleType)("D", Tags.TagDoubleRef)
+  final val NullRef = PrimRef(NullType)("N", Tags.TagNullRef)
+  final val NothingRef = PrimRef(NothingType)("E", Tags.TagNothingRef)
```
gzm0: I think it would make sense to provide an accessor per `encodedName` here, just like in an enum. The split of this logic seems a bit strange.

The same applies to tags.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -310,31 +312,47 @@ object Infos {
...
-      addAccessedClassData(baseNameOf(tpe))
+    def maybeAddAccessedClassData(typeRef: TypeRef): this.type = {
+      typeRef match {
+        case ClassRef(cls) =>
+          addAccessedClassData(cls)
+        case ArrayTypeRef(ClassRef(cls), _) =>
+          addAccessedClassData(cls)
+        case _ =>
+      }
+      this
```
gzm0: Create a helper for this pattern match? It appears many times.

Something like:

```scala
private def maybeAdd(typeRef: TypeRef, add: String => this.type): this.type = {
 // ...
}
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -350,15 +352,37 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+    case CharRef    => "C"
+    case ByteRef    => "B"
+    case ShortRef   => "S"
+    case IntRef     => "I"
+    case LongRef    => "J"
+    case FloatRef   => "F"
+    case DoubleRef  => "D"
+    case NullRef    => "N"
+    case NothingRef => "E"
+  }
```
gzm0: I find it a bit strange that we introduce types but the emitter internally still uses char codes in many locations (n.b. the corejslib). Consider using this function everywhere to translate.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1199,29 +1199,16 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
-        case 'B' => ByteType
-        case 'S' => ShortType
-        case 'I' => IntType
-        case 'J' => LongType
-        case 'F' => FloatType
-        case 'D' => DoubleType
-        case 'N' => NullType
-        case 'E' => NothingType
-        case 'O' => AnyType
-        case 'T' => ClassType(BoxedStringClass) // NOT StringType
```
gzm0: Huh... What happened here? (Was this forgotten in the previous PR)?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1236,9 +1223,15 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
       implicit ctx: ErrorContext): Type = {
-    val ArrayTypeRef(baseClassName, dimensions) = arrayTypeRef
-    if (dimensions == 1) classNameToType(baseClassName)
-    else ArrayType(ArrayTypeRef(baseClassName, dimensions - 1))
+    val ArrayTypeRef(base, dimensions) = arrayTypeRef
+    if (dimensions == 1) {
+      base match {
+        case PrimRef(tpe)  => tpe
+        case ClassRef(cls) => classNameToType(cls)
+      }
```
gzm0: Simply call `classRefToType`?

linker/shared/src/test/scala/org/scalajs/linker/testutils/TestIRBuilder.scala
```diff
@@ -71,7 +71,7 @@ object TestIRBuilder {
   }
 
   def mainMethodDef(body: Tree): MethodDef = {
-    val stringArrayType = ArrayType(ArrayTypeRef("T", 1))
+    val stringArrayType = ArrayType(ArrayTypeRef(ClassRef(BoxedStringClass), 1))
```
gzm0: I also do not understand this :)

sjrd: Hum?

1. `"T"` -> `ClassRef("T")` to adapt to the new arguments of `ArrayTypeRef`
2. Replace the literal `"T"` by the named constant `BoxedStringClass` which is equal to `"T"`

Perhaps you thought `"T"` was for *primitive* strings rather than "boxed" strings? It's a historical anomaly that `"T"` stands for `java.lang.String` rather than the primitive `StringType`. It was like that for shortening encoded names as much as possible. This thing will disappear once I get to delaying mangling to the Emitter, which should hopefully come after this PR (this PR came to fix difficulties I encountered when trying to implement the delayed mangling, actually).

ir/src/main/scala/org/scalajs/ir/Types.scala
```diff
@@ -163,17 +163,38 @@ object Types {
...
+  final val CharRef = PrimRef(CharType)
+  final val ByteRef = PrimRef(ByteType)
+  final val ShortRef = PrimRef(ShortType)
+  final val IntRef = PrimRef(IntType)
+  final val LongRef = PrimRef(LongType)
+  final val FloatRef = PrimRef(FloatType)
+  final val DoubleRef = PrimRef(DoubleType)
+  final val NullRef = PrimRef(NullType)
+  final val NothingRef = PrimRef(NothingType)
+
```
gzm0: Actually: Any reason these are not case objects? We could avoid the unchecked pattern matches then, no? Also it would be more consistent with the primitive types.

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -673,6 +673,20 @@ object Serializers {
     }
 
     def writeTypeRef(typeRef: TypeRef): Unit = typeRef match {
+      case typeRef: PrimRef =>
+        (typeRef: @unchecked) match {
```
gzm0: For example here, case objects would help, no?

ir/src/main/scala/org/scalajs/ir/Types.scala
sjrd: Yeah, I thought a lot about that. I really wanted to make them case objects, but if I do that I cannot make `PrimRef` a case class, and it really should be a case class for pattern matching of the form `case PrimRef(tpe)` to be efficient. So I didn't really have a choice, unfortunately.

We could do an abominable thing and make `PrimRef` and `abstract case class`, while making the individuals (non-case) `object`s ... But if I do that I won't be able to argue with Martin that abstract case class are an abomination anymore.

gzm0: Right. And if you simply define an extractor, you do not get exhaustiveness checking.

But, now I can of course come back to a slight variation of `PrimType`: A sub hierarchy (`RefablePrimType`) would allow exhaustiveness checking on the inner type. The `final val`s could be kept as mere aliases or not.

In the same swoop, I'd avoid overriding `equals` and `hashCode`: It feels like premature optimization traded off for some safety.

WDYT?

sjrd: Yes, that's probably reasonable. I'll try that later today when I get the chance.

sjrd: Done.

PR #3804, opened 2019-10-08
Introduce dedicated types for the kinds of `Ident`s and their `Name`s.

sjrd:
~~Based on #3802 -- only the second commit belongs to this PR.~~

Instead of putting everything in generic `Ident`s with `String`s, we now have dedicated types a) of `Name`s and b) of `Ident`s for the following categories:

* Local variables, including parameters and captures
* Labels
* Fields
* Methods
* Classes

The `Name`s such as `LocalName` and `MethodName` are currently implemented as bare `String`s hidden behind a type alias, and protected by validation upon construction. They are all validated to be valid JS identifiers, which was what they already had to be before this commit.

In the future, this new abstraction will make it easier to introduce different rules for validation, and in particular to delay mangling to the emitter, and to have better structure for method names (whose signatures are meaningful).

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Definitions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Definitions.scala
@@ -14,26 +14,68 @@ package org.scalajs.ir
 
 import scala.annotation.switch
 
+import Trees.isValidJSIdentifier
 import Types._
 
 object Definitions {
+  type LocalName <: String
+
+  def LocalName(name: String): LocalName = {
+    require(isValidJSIdentifier(name), "Invalid local name: " + name)
+    name.asInstanceOf[LocalName]
+  }
+
+  type LabelName <: String
+
+  def LabelName(name: String): LabelName = {
+    require(isValidJSIdentifier(name), "Invalid label name: " + name)
+    name.asInstanceOf[LabelName]
+  }
+
+  type FieldName <: String
+
+  def FieldName(name: String): FieldName = {
+    require(isValidJSIdentifier(name), "Invalid field name: " + name)
+    name.asInstanceOf[FieldName]
+  }
+
+  type MethodName <: String
+
+  def MethodName(name: String): MethodName = {
+    require(isValidJSIdentifier(name), "Invalid method name: " + name)
+    name.asInstanceOf[MethodName]
+  }
+
+  type ClassName <: String
+
+  def ClassName(name: String): ClassName = {
+    /* Because of pos/t9392, which generates a class whose full name is
+     * `<local Client>$C`, we cannot actually validate class names. It has
+     * worked so far because that test does not go through the linker. It will
+     * properly work again once we delay mangling to the emitter, which is why
+     * we avoid the validation rather than disabling the test.
+     */
+    // require(isValidJSIdentifier(name), "Invalid class name: " + name)
+    name.asInstanceOf[ClassName]
+  }
+
   /** `java.lang.Object`, the root of the class hierarchy. */
-  val ObjectClass = "O"
+  val ObjectClass: ClassName = ClassName("O")
 
   // Hijacked classes
-  val BoxedUnitClass = "jl_Void"
-  val BoxedBooleanClass = "jl_Boolean"
-  val BoxedCharacterClass = "jl_Character"
-  val BoxedByteClass = "jl_Byte"
-  val BoxedShortClass = "jl_Short"
-  val BoxedIntegerClass = "jl_Integer"
-  val BoxedLongClass = "jl_Long"
-  val BoxedFloatClass = "jl_Float"
-  val BoxedDoubleClass = "jl_Double"
-  val BoxedStringClass = "T"
+  val BoxedUnitClass: ClassName = ClassName("jl_Void")
+  val BoxedBooleanClass: ClassName = ClassName("jl_Boolean")
+  val BoxedCharacterClass: ClassName = ClassName("jl_Character")
+  val BoxedByteClass: ClassName = ClassName("jl_Byte")
+  val BoxedShortClass: ClassName = ClassName("jl_Short")
+  val BoxedIntegerClass: ClassName = ClassName("jl_Integer")
+  val BoxedLongClass: ClassName = ClassName("jl_Long")
+  val BoxedFloatClass: ClassName = ClassName("jl_Float")
+  val BoxedDoubleClass: ClassName = ClassName("jl_Double")
+  val BoxedStringClass: ClassName = ClassName("T")
 
   /** The set of all hijacked classes. */
-  val HijackedClasses: Set[String] = Set(
+  val HijackedClasses: Set[ClassName] = Set(
       BoxedUnitClass,
       BoxedBooleanClass,
       BoxedCharacterClass,
@@ -47,20 +89,26 @@ object Definitions {
   )
 
   /** The class of things returned by `ClassOf` and `GetClass`. */
-  val ClassClass = "jl_Class"
+  val ClassClass: ClassName = ClassName("jl_Class")
+
+  /** Name of a constructor without argument.
+   *
+   *  This is notably the signature of constructors of module classes.
+   */
+  final val NoArgConstructorName: MethodName = MethodName("init___")
 
   /** Name of the static initializer method. */
-  final val StaticInitializerName = "clinit___"
+  final val StaticInitializerName: MethodName = MethodName("clinit___")
 
   /** Encodes a class name. */
-  def encodeClassName(fullName: String): String = {
+  def encodeClassName(fullName: String): ClassName = {
     val base = fullName.replace("_", "$und").replace(".", "_")
     compressedClasses.getOrElse(base, {
       compressedPrefixes.collectFirst {
         case (prefix, compressed) if base.startsWith(prefix) =>
-          compressed + base.substring(prefix.length)
-      } getOrElse {
-        "L" + base
+          ClassName(compressed + base.substring(prefix.length))
+      }.getOrElse {
+        ClassName("L" + base)
       }
     })
   }
@@ -68,7 +116,7 @@ object Definitions {
   // !!! Duplicate logic: this code must be in sync with runtime.StackTrace
 
   /** Decodes a class name encoded with [[encodeClassName]]. */
-  def decodeClassName(encodedName: String): String = {
+  def decodeClassName(encodedName: ClassName): String = {
     val base = decompressedClasses.getOrElse(encodedName, {
       decompressedPrefixes.collectFirst {
         case (prefix, decompressed) if encodedName.startsWith(prefix) =>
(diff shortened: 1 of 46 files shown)


sjrd:
There's a failure in 2.11.0 that I could reproduce up to 2.11.5 or something like that. They don't like the `type LocalName <: String` abstract definitions in `object Definitions`.

Maybe we should get rid of old 2.11.x versions? That would also allow to get rid of some of the workarounds we do with type-tests in pattern matching instead of `@` bindings. Otherwise I can temporarily disable the CI on old 2.11.x versions until the next refactoring comes in, which will actually introduce distinct classes for the different kinds of names.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -819,29 +819,36 @@ private object Emitter {
...
+    val StringArgConstructorName = newMethodName("init___T")
+
     multiple(
-        instantiateClass("O", "init___"),
-        classData("O"),
+        instantiateClass(ObjectClass, NoArgConstructorName),
+        classData(ObjectClass),
 
-        instantiateClass("jl_CloneNotSupportedException", "init___"),
+        instantiateClass(newClassName("jl_CloneNotSupportedException"),
```
gzm0: Is there a reason these class names are not in constants? It seems that everywhere else they go in constants (even if only used once).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -98,12 +99,12 @@ private[emitter] final class KnowledgeGuardian(config: CommonPhaseConfig) {
...
-        "jl_ClassCastException",
-        "jl_ArrayIndexOutOfBoundsException",
-        "sjsr_UndefinedBehaviorError",
-        "jl_CloneNotSupportedException"
+    val blackList = Set[ClassName](
+        newClassName("jl_ArithmeticException"),
+        newClassName("jl_ClassCastException"),
+        newClassName("jl_ArrayIndexOutOfBoundsException"),
+        newClassName("sjsr_UndefinedBehaviorError"),
+        newClassName("jl_CloneNotSupportedException")
```
gzm0: Same comment about constants.

linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/Trees.scala
```diff
@@ -45,7 +43,8 @@ object Trees {
 
   case class Ident(name: String, originalName: Option[String])(
       implicit val pos: Position) extends PropertyName {
-    requireValidIdent(name)
+    if (!ir.Trees.isValidJSIdentifier(name))
+      throw new IllegalArgumentException(s"'$name' is not a valid JS identifier")
```
gzm0: `require(ir.Trees.isValidJSIdentifier(name))`?

linker/shared/src/main/scala/org/scalajs/linker/frontend/MethodSynthesizer.scala
```diff
@@ -90,13 +91,14 @@ private[frontend] final class MethodSynthesizer(
...
-      val targetIdent = targetMDef.name.asInstanceOf[Ident].copy() // for the new pos
+      val targetIdent = targetMDef.name.copy() // for the new pos
       val bridgeIdent = targetIdent
       val params = targetMDef.args.map(_.copy()) // for the new pos
       val currentClassType = ClassType(classInfo.encodedName)
 
       val body = ApplyStatically(
-          ApplyFlags.empty, This()(currentClassType), ClassRef(targetInterface),
+          ApplyFlags.empty, This()(currentClassType),
+          ClassRef(targetInterface),
```
gzm0: Spurious change?

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
```diff
@@ -100,13 +100,15 @@ class OptimizerTest {
 
     for (linkingUnit <- linkToLinkingUnit(classDefs, MainTestModuleInitializers)) yield {
       val linkedClass = linkingUnit.classDefs.find(_.encodedName == MainTestClassDefEncodedName).get
+      val WitnessMethodName = newMethodName("witness__O")
+      val CloneMethodName = newMethodName("clone__O")
       linkedClass.hasNot("any call to Foo.witness()") {
-        case Apply(_, receiver, Ident("witness__O", _), _) =>
+        case Apply(_, receiver, MethodIdent(WitnessMethodName, _), _) =>
```
gzm0: I take it the implicit conversions from `TestIRBuilder` do not work here?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -51,7 +51,7 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
     import factory._
 
     callMethods(LongImpl.RuntimeLongClass, LongImpl.AllIntrinsicMethods.toList) ++
-    instantiateClass("jl_NullPointerException", "init___")
+    instantiateClass(newClassName("jl_NullPointerException"), NoArgConstructorName)
```
gzm0: Use `JavaLangNullPointerException`.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -1993,7 +1991,7 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
             contTree(ArraySelect(array, index)(elemType))
 
           case _ =>
-            defaultApply("array$undapply__O__I__O", AnyType)
+            defaultApply(newMethodName("array$undapply__O__I__O"), AnyType)
```
gzm0: Same comment about making / not making these constants.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -5337,7 +5341,7 @@ private[optimizer] object OptimizerCore {
         }
         usedNamesToNextCounter =
           usedNamesToNextCounter.updated(base, i + 1).updated(result, 1)
-        result
+        result.asInstanceOf[N]
```
gzm0: Hum... Pass in a conversion function to the constructor?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4393,20 +4394,29 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
 
 private[optimizer] object OptimizerCore {
 
+  /** Hacky `LocalName` representing `this`.
+   *
+   *  `this` is normally not a valid `LocalName`, because that is a JS keyword,
+   *  so we cast here.
+   */
+  private val LocalThisName = "this".asInstanceOf[LocalName]
```
gzm0: So basically this happens because the binding infrastructure re-uses LocalName, but allows "this" as an identifier to bind to this? I need to think a bit if I can think of something better here :S

ir/src/main/scala/org/scalajs/ir/Definitions.scala
```diff
@@ -14,26 +14,73 @@ package org.scalajs.ir
 
 import scala.annotation.switch
 
+import Trees.isValidJSIdentifier
```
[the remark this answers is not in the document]
sjrd: Yes I was already considering renaming `Definitions` to `Names`, but perhaps we can do that later. Or if you prefer I can do it in this commit.

I wouldn't move `isValidJSIdentifier` now because it will soon become irrelevant for `Names`, and will only apply to `JSGlobalRef`s, as Names will allow any JVM name, which is virtually any Unicode string.

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -1232,6 +1293,56 @@ object Serializers {
...
+    private def readClassName(): ClassName = {
+      val i = readInt()
+      val s = strings(i)
+      if (!isStringAlreadyValidated(i, ValidatedAsClassNameBit)) {
+        newClassName(s)
+        didValidateString(i, ValidatedAsClassNameBit)
+      }
+      s.asInstanceOf[ClassName]
+    }
+
```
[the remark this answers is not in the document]
sjrd: The idea is that names are at least internally consistent, always. The IR checker validates class structures and type relations, which need some global information, but the structure of names can be validated one by one.

This will become critical for method names in the next (and last) refactoring in the series, because they will have actual structure, which can only *exist* if the encoded name is validated.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -194,15 +195,17 @@ private final class Analyzer(config: CommonPhaseConfig,
...
+
     val classClassInfo = classInfos.get(Definitions.ClassClass)
 
     /* If Class.getSuperclass() is reachable, we can reach the data of all
      * superclasses of classes whose data we can already reach.
      */
     for {
       getSuperclassMethodInfo <-
-        classClassInfo.flatMap(_.publicMethodInfos.get("getSuperclass__jl_Class"))
+        classClassInfo.flatMap(_.publicMethodInfos.get(getSuperclassMethodName))
```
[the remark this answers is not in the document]
sjrd: The constants for non-constructor method names all start with a lowercase, now (matching the actual method name they stand for).

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
sjrd: Yes. I haven't found an easy way to address that here, but I agree we should be able to do better. I would leave that for another commit, however, as it has little to do with the current refactoring.

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
sjrd: No they don't, because we are in a pattern.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -24,16 +24,22 @@ import org.scalajs.linker.backend.emitter.Transients._
 
 object Infos {
 
+  private val ArithmeticExceptionName =
+    newClassName("jl_ArithmeticException")
+
+  private val initStringName =
+    newMethodName("init___T")
```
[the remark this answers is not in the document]
sjrd: All constructor names are now consistently called `TypeArgConstructorName`, where `Type` is the type of the argument.

ir/src/main/scala/org/scalajs/ir/Definitions.scala
```diff
@@ -14,26 +14,73 @@ package org.scalajs.ir
...
 import Types._
 
 object Definitions {
+  type LocalName <: String
+
+  def newLocalName(name: String): LocalName = {
```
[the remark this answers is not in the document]
sjrd: I changed to an apply style.

sjrd:
And rebased.

ir/src/main/scala/org/scalajs/ir/Definitions.scala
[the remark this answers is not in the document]
gzm0: Ok. Let's leave things as they are and re-asses once the name mangling changes are complete.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Yep. SG. I take it we'll have to fix this anyways because going forward, "this" will be a valid identifier?

sjrd:
Nits addressed :)

PR #3805, opened 2019-10-08
Fix #3661: Load Linker Reflectively

### LINKED ISSUE #3661: Use reflection to load the Scala.js linker in the sbt plugin

gzm0:


gzm0:
Note that fullOptJS does currently not work because there is a dependency problem on Guava between closure compiler and jimfs. As a result of the class loading, the older version gets loaded and there are binary incompatibilites.

gzm0:
Hah, now of course we have the problem that we load the wrong Scala version :)

Alternatives:
- Fiddle with the build invocation until it works.
- Make the interface plain Java (can we even do this in Scala)? Main problem would be futures AFAICS.

@sjrd WDYT?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -183,16 +202,21 @@ object ScalaJSPlugin extends AutoPlugin {
 
         scalaJSLinkerConfig := StandardConfig(),
 
+        scalaJSLinkerImpl := LinkerImpl.default(getClass().getClassLoader()),
```
sjrd: What's the point of loading reflectively if you are using the current class loader?

gzm0:
> downloads it separately from the classpath of the sbt plugin

That is of course an even nicer solution. Actually was considering it, but didn't because it needs a custom configuration :) I'll do this.

> To resolve the Guava thing and the Scala version thing, I think we need something similar to what we had done for the JettyClassLoader of PhantomJS in 0.6.x

This will solve the Guava thing, but not the Scala version thing: The linker interface depends on the scala library. So we have to load at least these classes and their transitive dependencies from the parent classloader. Otherwise we'll have a class mismatch.

sjrd:
> That is of course an even nicer solution. Actually was considering it, but didn't because it needs a custom configuration :) I'll do this.

You don't need a custom configuration anymore with sbt 1.x. You can directly invoke the `librarymanagement` API. See here an example how to do that:
https://github.com/sjrd/sbt-dynscalajs/blob/06293c262b24463361d43d37130448b331c0987d/sbt-dynscalajs/src/main/scala/be/doeraene/sbtdynscalajs/DynScalaJSPlugin.scala#L256-L276

> This will solve the Guava thing, but not the Scala version thing: The linker interface depends on the scala library. So we have to load at least these classes and their transitive dependencies from the parent classloader. Otherwise we'll have a class mismatch.

Hum yes but as long as the linker interface you load uses a 2.12.x Scala version, it will correctly link with the Scala 2.12.y used by sbt.

gzm0:
@SethTisue before I investigate the failures here further: Could you have a look at the changes to how we chose the Scala version in this build and check if it is community build compatible? If no, how do I have to adapt it?

SethTisue:
offhand it looks okay to me, I'm not 100% sure, but we can always deal it with later

gzm0:
So the CI failures look like sbt is loading the wrong SJS compiler when compiling the `Test` configuration only. I can reproduce the errors locally: For a JS project, `<project>/compile` works fine, whereas `<project>/test:compile` does not.

project/Build.scala
```diff
@@ -495,7 +506,7 @@ object Build {
         if (v < 8)
           throw new MessageOnlyException("This build requires JDK 8 or later. Aborting.")
         v
-      }
+      },
```
gzm0: Spurious change.

project/Build.scala
```diff
@@ -201,8 +208,8 @@ object Build {
   }
 
   val commonSettings = Seq(
-      scalaVersion := "2.12.8",
       organization := "org.scala-js",
+      scalaVersion := "2.12.8",
```
gzm0: Spurious change

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -57,13 +57,17 @@ object ScalaJSPlugin extends AutoPlugin {
 
     // All our public-facing keys
```
gzm0: Should we make the internal keys private?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -82,9 +86,19 @@ object ScalaJSPlugin extends AutoPlugin {
...
+     *  automatically set up the correct caching behavior.
      */
-    val scalaJSLinker = SettingKey[ClearableLinker]("scalaJSLinker",
-        "Persisted instance of the Scala.js linker", KeyRanks.Invisible)
+    val scalaJSLinker = TaskKey[ClearableLinker]("scalaJSLinker",
+        "Access task for a Scala.js linker. Use this if you want to use the linker.",
+        KeyRanks.Invisible)
+
+    val scalaJSLinkerImpl = TaskKey[LinkerImpl]("scalaJSLinkerImpl",
+        "Factory for a Scala.js linker", KeyRanks.Invisible)
```
gzm0: TODO: Describe this.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -82,9 +86,19 @@ object ScalaJSPlugin extends AutoPlugin {
...
+
+    val scalaJSLinkerBox = SettingKey[CacheBox[ClearableLinker]]("scalaJSLinkerBox",
+        "CacheBox for a Scala.js linker", KeyRanks.Invisible)
```
gzm0: Mark this as internal.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -172,9 +186,14 @@ object ScalaJSPlugin extends AutoPlugin {
...
-    import ScalaJSPluginInternal.globalIRCache
-    logger.debug("Global IR cache stats: " + globalIRCache.stats.logLine)
+  val irCacheStatsLogger: Def.Initialize[Logger => Unit] = Def.setting {
+    val globalIRCacheBox = scalaJSGlobalIRCacheBox.value
+
+    logger: Logger => {
+      globalIRCacheBox.foreach { globalIRCache =>
+        logger.debug("Global IR cache stats: " + globalIRCache.stats.logLine)
+      }
+    }
```
gzm0: I really don't like this. Can we find a better way to do this? Ideally after every task that users the linker :)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -183,16 +202,34 @@ object ScalaJSPlugin extends AutoPlugin {
...
         // Clear the IR cache stats every time a sequence of tasks ends
         onComplete := {
           val prev = onComplete.value
+          val globalIRCacheBox = scalaJSGlobalIRCacheBox.value
 
           { () =>
             prev()
             ScalaJSPluginInternal.closeAllTestAdapters()
-            ScalaJSPluginInternal.globalIRCache.clearStats()
+            globalIRCacheBox.foreach(_.clearStats())
```
gzm0: Would it be acceptable to log IR cache stats just here? That would get rid of the strange logger and also make sure nobody ever forgets it. At the cost of only logging once if multiple linking (or other IR consuming) operations are executed in a single swoop.

sjrd: I think that would be acceptable, yes.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -171,28 +188,44 @@ object ScalaJSPlugin extends AutoPlugin {
...
 
+        scalaJSLinkerImpl := {
+          val s = streams.value
+          val log = s.log
+          val retrieveDir = s.cacheDirectory / "scalajs-linker" / scalaJSVersion
+          val lm = {
+            import sbt.librarymanagement.ivy._
+            val ivyConfig = InlineIvyConfiguration().withLog(log)
+            IvyDependencyResolution(ivyConfig)
+          }
```
gzm0: It seems this does not try an load things from local Ivy:

```
[warn] 	module not found: org.scala-js#scalajs-linker_2.12;1.0.0-SNAPSHOT
[warn] ==== public: tried
[warn]   https://repo1.maven.org/maven2/org/scala-js/scalajs-linker_2.12/1.0.0-SNAPSHOT/scalajs-linker_2.12-1.0.0-SNAPSHOT.pom
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.scala-js#scalajs-linker_2.12;1.0.0-SNAPSHOT: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

@sjrd any idea how to fix that?

sjrd: Er, no, I have no idea :s Perhaps on Gitter: https://gitter.im/sbt/sbt ?

sjrd: Looking at the docs, perhaps all you need is explicitly adding [`Resolver.defaultLocal`](https://www.scala-sbt.org/1.2.8/api/sbt/librarymanagement/Resolver$.html#defaultLocal:sbt.librarymanagement.FileRepository) to the `InlineIvyConfiguration` using its [`withOtherResolvers`](https://www.scala-sbt.org/1.2.8/api/sbt/librarymanagement/ivy/InlineIvyConfiguration.html#withOtherResolvers(otherResolvers:Vector[sbt.librarymanagement.Resolver]):sbt.librarymanagement.ivy.InlineIvyConfiguration) method?

gzm0:
@sjrd: This is ready for review.

sjrd:
@SethTisue could we have a 2.13 community build run based on this branch? Or how can we try it ourselves or help you trying it?

I am a bit worried about the community build since, AFAIK, it basically performs a `++v!` operation, which would force even the `linker` project into 2.13, which would then fail `testSuite/test`.

project/Build.scala
```diff
@@ -732,6 +747,7 @@ object Build {
 
   lazy val jsEnvs: Project = (project in file("js-envs")).settings(
       commonSettings,
+      freezeVersionForToolsSettings,
```
sjrd: Why is this necessary for `jsEnvs` and all the projects below it?

sbt-plugin-test/project/build.sbt
```diff
@@ -1,2 +1,5 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" %
-  org.scalajs.ir.ScalaJSVersions.current)
+val scalaJSVersion = org.scalajs.ir.ScalaJSVersions.current
+
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
+
+libraryDependencies += "org.scala-js" %% "scalajs-linker" % scalaJSVersion
```
sjrd: Why is this necessary? Can we get rid of it? It wouldn't be nice to lose our only test that executes the linker in real real conditions, i.e., in the same conditions as our users.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+      val m = Class.forName("org.scalajs.linker." + clazz, true, loader).getMethod(method, params: _*)
+      require(Modifier.isStatic(m.getModifiers()))
+      require(result.isAssignableFrom(m.getReturnType()))
+      m
+    }
+
+    private def invoke[T](method: Method, args: AnyRef*): T =
+      method.invoke(null, args: _*).asInstanceOf[T]
+
+    /* We load everything eagery to fail immediately, not only when the methods
```
sjrd: Typo: eagerly

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private val clearableLinkerMethod =
+      loadMethod("StandardImpl", "clearableLinker", classOf[ClearableLinker], classOf[StandardConfig])
+
+    private val irFileCacheMethod =
+      loadMethod("StandardImpl", "irFileCache", classOf[IRFileCache])
+
+    private val irContainersMethod = {
+      loadMethod("PathIRContainer", "fromClasspath", classOf[Future[_]],
+        classOf[Seq[Path]], classOf[ExecutionContext])
```
sjrd: 4 spaces

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -126,27 +113,32 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
 
+      scalaJSLinkerBox in key := new CacheBox,
+
       scalaJSLinker in key := {
         val config = (scalaJSLinkerConfig in key).value
+        val box = (scalaJSLinkerBox in key).value
+        val linkerImpl = (scalaJSLinkerImpl in key).value
+        val projectID = thisProject.value.id
+        val configName = configuration.value.name
+        val log = sLog.value
```
sjrd: You should use `streams.value.log` now that `scalaJSLinker` is a task.

SethTisue:
> could we have a 2.13 community build run based on this branch? Or how can we try it ourselves or help you trying it?

I did a run locally just now, and here are the steps I followed:

* clone scala/community-build repo, 2.13.x branch
* edit `proj/scala-js.conf` and replace `uri: "https://github.com/scala-js/scala-js.git#0d6be65846cc67d2807c3cf622b22aa6f19a3254"` with `uri: "https://github.com/gzm0/scala-js.git#load-reflectively"`
* `./narrow scala-js` (this script is brand new, I just added it a couple days ago) to avoid extracting all 200 projects, and since there aren't any downstream projects we need to test. this cuts projs.conf down to just scalacheck, scala-parallel-collections, scala-js
* `./run.sh`

as predicted, this results in

```
[scala-js] Processing subproject: testSuite
...
[scala-js] [error] java.lang.AssertionError: assertion failed: The linker is configured with a Scala version (2.13.2-bin-99afc54) that is not compatible with this build (2.12.x). 
[scala-js] [error] 	at scala.Predef$.assert(Predef.scala:223)
[scala-js] [error] 	at build.MyScalaJSPlugin$.$anonfun$projectSettings$7(Build.scala:109)
```

> Is there a way we can tell the community build not to override the Scala version of some projects?

let's pursue that over at https://github.com/scala/community-build/pull/1005

project/Build.scala
gzm0: Because they depend on `logger` which has this setting. So if we do not give them the setting, versions are inconsistent and `update` can (and does) fail in some CI tasks.

sbt-plugin-test/project/build.sbt
gzm0: It is for MemOutputFile in the concurrent linker test, so we can't get rid of it. How about I'll onboard us onto scripted and move that particular part to a scripted test (keeping the rest in `sbt-plugin-test`)?

sjrd: Yeah, looks like we really need scripted ...

Although if it's just this particular problem, maybe we can use a file on disk instead of an in memory file? Since we already have a means to create a disk output file with `scalaJSLinkerImpl`?

gzm0: Ah yes. That's fair. OTOH, this neatly tested the class loader magic to make sure it is actually possible to mix two "loadings" of the linker. I think in general this would benefit from some more splitting, so how about we try the scripted approach and otherwise use a real file here?

sjrd: Sounds good

gzm0:
Updated.

sjrd:
As I wrote in #3817, I think the changes there won't qualitatively change the challenge with the community build. So if we can validate the community build integration for this PR, the solution will transfer to #3817.

gzm0:
> As I wrote in #3817, I think the changes there won't qualitatively change the challenge with the community build. So if we can validate the community build integration for this PR, the solution will transfer to #3817.

Right. In fact I realize that *our* solution to this does not transfer to #3817: We will need *two* levels of pinning there: The linker needs to be conditionally pinned; the linker private library needs to be *always* pinned.

Another option is to create a project per major scala version for *all* projects. It seems that would solve the version problem for us for all cases (including using scripted with different scala versions). Minor versions could be selected with `++v`.

WDYT?

SethTisue:
(I am committed to helping find a way forward here, but I'm on semi-vacation for a while as I move from CA to NV, so my time is a bit scarce right now)

sjrd:
Using one project per major Scala version for all projects seems reasonable, indeed.

SethTisue:
this is green in the 2.12 community build since the Scala major version happens to align between the build and the meta-build

but in the 2.11 and 2.13 builds, it fails, since the changes here pull in `linker.v2_12` but then dbuild overrides `linker.v2_12`'s Scala version to 2.11

I have some ideas about how to work around it. I'll mess with it

SethTisue:
it's best if I don't have to fork your build, since then my fork would always be falling out of date.

could you offer me a setting I could flip that would cause `scalaJSLinkerImpl` to be set differently in the community build context — namely, by not doing this problematic `fullClasspath in (Build.linker.v2_12, Runtime` thing, but instead retrieving the 2.12 artifact from Maven Central and using that. does that seem plausible?

sjrd:
The artifacts you'll find on Maven Central will not correspond to the version being tested, so that's not a viable path forward I'm afraid.

I see two possible paths:

* The easiest I think would be to tell dbuild not to mess with some projects in the build, in this case any of the `2_12` projects. That includes not setting the Scala version, and not overriding dependencies.
* Otherwise, maybe we can tell dbuild to first build normally and publish locally the relevant 2.12 projects, then go ahead with its normal patching everything behavior, while using `scalaJSLinkerImpl` to load from the locally published version.

SethTisue:
I've opened a separate ticket, https://github.com/scala/community-build/issues/1009, about whether it makes sense to continue using dbuild for Scala.js at all. let's also discuss there

SethTisue:
> I see two possible paths:

I was thinking along these lines as well, but it doesn't seem easy to me, I tried some quick and dirty efforts one but try didn't work and in the other try I got bogged down and it seemed like a time sink without success being assured. so I'd like to discuss on #1009 a bit first, then depending on how that goes, perhaps return to this ticket to go into further details about the stick-with-dbuild option.

PR #3806, opened 2019-10-09
Drop support for old 2.11.x versions.

sjrd:
We only keep 2.11.12, which is the latest 2.11.x version at the time of this commit.


gzm0:
I think this is OK. However, going forward we might want to formalize a bit how long we intend to keep supporting versions...

project/Build.scala
```diff
@@ -409,15 +408,8 @@ object Build {
...
-      // The pattern matcher used to exceed its analysis budget before 2.11.5
-      scalacOptions ++= {
-        scalaVersion.value.split('.') match {
-          case Array("2", "11", x)
-              if x.takeWhile(_.isDigit).toInt <= 4 => Nil
-          case _                                   => Seq("-Xfatal-warnings")
-        }
-      },
+  val fatalWarningsSettings = Def.settings(
+      scalacOptions += "-Xfatal-warnings"
```
gzm0: Should we inline this? Seems a bit pointless like this

sjrd: Well, I think it's still relevant to abstract it away, if anything so that we don't have to copy-paste a string literal, which is not type-checked.

sjrd:
> However, going forward we might want to formalize a bit how long we intend to keep supporting versions...

I think that, for a full release (not milestones), we should indefinitely keep support, or at least until all major Scala libraries have themselves dropped support for any given version of Scala. But as we're still in the milestones for 1.x, I believe it's a nice opportunity to get rid of that legacy. In general nobody should have any issue upgrading to 2.11.12 from 2.11.x, so they can do that before upgrading to a new milesone of Scala.js 1.x.

project/Build.scala
gzm0: OK. Fair enough.

gzm0:
> I think that, for a full release (not milestones), we should indefinitely keep support, or at least until all major Scala libraries have themselves dropped support for any given version of Scala. But as we're still in the milestones for 1.x, I believe it's a nice opportunity to get rid of that legacy. In general nobody should have any issue upgrading to 2.11.12 from 2.11.x, so they can do that before upgrading to a new milesone of Scala.js 1.x.

Agreed.

PR #3809, opened 2019-10-10
Disambiguate overload of FileSystems.newFileSystem for JDK 13+.

SethTisue:
this turned up in the context of scala/community-build#873

### DIFF
--- a/linker/jvm/src/main/scala/org/scalajs/linker/PathIRContainer.scala
+++ b/linker/jvm/src/main/scala/org/scalajs/linker/PathIRContainer.scala
@@ -58,7 +58,9 @@ object PathIRContainer {
 
       blocking {
         // Open zip/jar file as filesystem.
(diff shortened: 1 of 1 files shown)


linker/jvm/src/main/scala/org/scalajs/linker/PathIRContainer.scala
```diff
@@ -58,7 +58,7 @@ object PathIRContainer {
 
       blocking {
         // Open zip/jar file as filesystem.
-        val fs = FileSystems.newFileSystem(path, null)
+        val fs = FileSystems.newFileSystem(path, null: ClassLoader)
```
sjrd: Could you add a comment why the disambiguation is needed? Since we don't test on JDK 13, we might otherwise forget to preserve this ascription if we refactor that code.

SethTisue:
Suggested revisions, made.

PR #3810, opened 2019-10-11
Avoid dotty keyword enum in local variables names

neshkeev:
This patch renames variables that have name `enum` to `enumeration` since
`enum` is a keyword for `Dotty`.

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/CollectionsTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/CollectionsTest.scala
@@ -304,18 +304,18 @@ class CollectionsTest extends CollectionsTestBase {
 
   @Test def enumeration(): Unit = {
     val coll = TrivialImmutableCollection(range: _*)
-    val enum = ju.Collections.enumeration(coll)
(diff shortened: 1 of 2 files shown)


scalalib/overrides/scala/Enumeration.scala
```diff
@@ -214,9 +214,9 @@ abstract class Enumeration (initial: Int) extends Serializable {
       else s"<Unknown name for enum field #$i of class ${getClass}>"
 
     protected def readResolve(): AnyRef = {
-      val enum = thisenum.readResolve().asInstanceOf[Enumeration]
-      if (enum.vmap == null) this
-      else enum.vmap(i)
+      val enumeration = thisenum.readResolve().asInstanceOf[Enumeration]
+      if (enumeration.vmap == null) this
+      else enumeration.vmap(i)
```
sjrd: Don't change files under `scalalib/overrides...`. Those files will never be compiled by Scala 3, and should stay in sync with the corresponding files upstream.

neshkeev: The file has been reverted back to `0.6.x`

neshkeev:
@gzm0 , I changed the commit message

PR #3811, opened 2019-10-11
Drop parenthesis from nullary method overrides

neshkeev:
This patch drops parenthesis from the `def factory` methods, since the
corresponding methods from the parent traits do not have parenthesis.

Dotty does not implicitly inserts parenthesis to nullary methods that so
if a nullary method overrides another nullary method then they both need
to either have parenthesis or not. Otherwise Dotty fails to compile such
methods.


sjrd:
I'm not sure why Jenkins does not pick up this PR. Maybe it's only going to pick it up later.

Let's wait a bit.

neshkeev:
@gzm0 , I changed the commit message

PR #3817, opened 2019-10-22
Include `RuntimeLong.sjsir` as a resource of the linker.

gzm0:


project/Build.scala
```diff
@@ -712,27 +739,121 @@ object Build {
...
+        //  val setVersionSettings = relevantProjects.map { p =>
+        //    scalaVersion in p := "2.12.8"
+        //  }
+        //  val newState =
+        //    appendWithSession(Project.extract(s), setVersionSettings, s)
+        //  val extracted = Project.extract(newState)
+        //  val (_, privateLibProducts) =
+        //    extracted.runTask(products in (linkerPrivateLibrary, Compile), newState)
+        //  privateLibProducts
+        //}
```
gzm0: @dwijnand you recommended against this. How do we do this properly? (We always want to depend on the same scala version build).

dwijnand: `resourceGenerators` takes a list of `Task`s, so you should (in theory - aka, in sbt, with blood, spit and lots of looking at the internals of sbt's macros) be able to wire `products in (linkerPrivateLibrary, Compile)` up so it runs the tasks within the task graph.

For starters you'll have to statically define those (`products in (compiler, Compile)`, `products in (library, Compile)`, etc) and then pull out the underlying task value (`.taskValue`?) and then modify those results to finish defining your resource generator.  Look at the old "tuple syntax" code for some ideas.

gzm0: Thanks for the quick answer. That would work, but I'm not 100% sure what the difference with simply writing `val privateLibProducts = (products in (linkerPrivateLibrary, Compile)).value` would be.

I think the main challenge is that we need to inject different states into parts of the task graph.

For example, if we want to link the test suite in 2.13.x, we need to:

- Compile the SJS compiler for 2.13.x
- Compile the test suite for 2.13.x (using the SJS compiler)
- Compile the SJS compiler for 2.12.8
- Compile the linker library for 2.12.8 (using the SJS compiler)
- Compile the linker for 2.12.8 (using a plain compiler)
- Link the test suite artifacts with the linker.

So the main challenge I see here is that we need to compile the compiler twice. So we need to run these tasks twice under different states. Is that doable?

dwijnand: > I'm not 100% sure what the difference with simply writing...

You're right - that's probably the same.

> So the main challenge I see here is that we need to compile the compiler twice.

You need different projects then. `products in (compiler213, Compile)`, etc.

gzm0: @sjrd what do you say about this? We could do this, but it would probably make the community build integration even more complicated :-/

sjrd: It looks like it's the only safe way, yes.

It's probably not more complicated for the community build than what we are already dealing with in #3805. If we create a `compilerForLinkerPrivateLib` and a `libraryForLinkerPrivateLib` in addition to the `linkerPrivateLib`, those three projects would have to be pinned to 2.12.8, just like the `ir` and `linker` projects.

linker/shared/src/main/scala/org/scalajs/linker/standard/LinkerBackend.scala
```diff
@@ -33,6 +33,15 @@ abstract class LinkerBackend {
...
   val symbolRequirements: SymbolRequirement
 
+  /** Additional IR files to inject for linking, mandated by this back-end.
+   *
+   *  Example: the standard emitter back-end injects `RuntimeLong.sjsir` and
+   *  its companion object, unless it uses `BigInt`s to implement `Long`s.
+   *
+   *  The default implementation in `LinkerBackend` returns an empty sequence.
+   */
+  def injectedIRFiles: Seq[IRFile] = Nil
```
gzm0: We should consider making this abstract: It took me quite a while to debug that a forwarder like this was forgotten in a test. I'd expect most real life usages to run into a similar issue.

linker/jvm/src/main/scala/org/scalajs/linker/backend/emitter/PrivateLibHolder.scala
```diff
@@ -0,0 +1,54 @@
...
+          path = relativeDir + name,
+          version = Some(""), // this indicates that the file never changes
+          content = readResource(name)
+      )
+    }
+  }
+
+  private def readResource(name: String): Array[Byte] = {
+    val inputStream = getClass().getResourceAsStream(name)
+    assert(inputStream != null, f"couldn't load $name from resources")
```
sjrd:
```suggestion
    assert(inputStream != null, s"couldn't load $name from resources")
```

linker/shared/src/main/scala/org/scalajs/linker/standard/LinkerBackend.scala
sjrd: Yes, making it abstract sounds good.

linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
```diff
@@ -171,13 +174,12 @@ object IRCheckerTest {
...
       .withOptimizer(false)
     val linkerFrontend = StandardLinkerFrontend(config)
-    val symbolRequirements = StandardLinkerBackend(config).symbolRequirements
 
     val classDefsFiles = classDefs.map(MemClassDefIRFile(_))
 
     val result = TestIRRepo.minilib.stdlibIRFiles.flatMap { stdLibFiles =>
       linkerFrontend.link(stdLibFiles ++ classDefsFiles, moduleInitializers,
-        symbolRequirements, new ScalaConsoleLogger(Level.Error))
+        noSymbolRequirements, new ScalaConsoleLogger(Level.Error))
```
sjrd: That does not seem right. What about the other things in `Emitter.symbolsRequirements`, such as `new java.lang.Class(Object)`?

project/Build.scala
```diff
@@ -722,6 +741,19 @@ object Build {
...
       ),
+
+      resourceGenerators in Compile += Def.task {
+        val s = streams.value
+        val baseResourceDir = (resourceManaged in Compile).value
+        val resourceDir = baseResourceDir / "org/scalajs/linker/backend/emitter"
+
+        val privateLibProducts = (products in (linkerPrivateLibrary, Compile)).value
+        val mappings = (privateLibProducts ** "*.sjsir").pair(Path.flat(resourceDir))
+        Sync.sync(s.cacheStoreFactory.make("linker-library"))(mappings)
```
sjrd: What does this do? Could you add a comment?

linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
gzm0: What about it? `Emitter.symbolRequirements` should only be symbols needed by the emitter (which we do not check here). The frontend should work with a different backend that has completely different symbol requirements.

gzm0:
Updated

gzm0:
Rebased. There is still the question about IRCheckerTest IIUC.

linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
sjrd: Looking at this more closely, I realize there are two kinds of requirements that the `Emitter` advertises:

* Requirements linked to the compilation strategy that the emitter uses, i.e., the `RuntimeLong` requirements
* Requirements linked to the semantics of certain IR constructs (everything else in `Emitter.symbolRequirements`)

The former can definitely be omitted in the `IRCheckerTest` and any other front-end-only tests.

The latter should *always* be required, and hence I believe they should be moved directly inside `Analyzer`. No matter what the compilation strategy of a back-end is, it *must* for example instantiate a `java.lang.Class` at some point (for `GetClass` and `ClassOf`).

So I guess the proper solution would be to add a first commit that moves all these other requirements into the `Analyzer`. Then the current contents of the PR are fine after that.

project/Build.scala
```diff
@@ -680,6 +688,17 @@ object Build {
       library, irProjectJS, loggingJS,
   )
 
+  lazy val linkerPrivateLibrary: Project = (project in file("linker-private-library")).enablePlugins(
```
sjrd: You should add `linkerPrivateLibrary` to `val allProjects` in `root`.

linker/shared/src/test/scala/org/scalajs/linker/IRCheckerTest.scala
gzm0: Hmmm... I started looking at this and while I agree with this distinction, I believe the current code already meets it: None of the things the Emitter reaches are required in all cases and are merely implementation details of the `CoreJSLib`:

> instantiateClass(ObjectClass, NoArgConstructorName)
> classData(ObjectClass),

Needed for arrays which we do not DCE (but could).

> instantiateClass(CloneNotSupportedExceptionClass, NoArgConstructorName),

Needed for Object.clone() which we do not DCE (but could).

> All exceptions:

Implementation detail of how the Emitter actually throws them. If more stuff were implemented in IR, they would be reached "normally".

> instantiateClass(ClassClass, ObjectArgConstructorName)

Private contract between our implementation of java.lang.Class and the CoreJS lib of how class data is passed. The `ClassClass` itself could arguably be reached by the Analzyer itself *iff* any class data is accessed. But I'm not sure it is worth doing that.

I'm not quite sure how Hijacked classes change this picture, but I'm sure you are :)

sjrd: Hum ... I think you're write for `ObjectClass` and for `CloneNotSupportedException`. However,

> > All exceptions:
> 
> Implementation detail of how the Emitter actually throws them. If more stuff were implemented in IR, they would be reached "normally".

No, those exceptions are thrown as part of the semantics of certain IR Trees. For example `ArraySelect` can throw an `ArrayIndexOutOfBoundsException`. There's no way we could implement that "normally" in IR code, because `ArraySelect` cannot be broken into more primitive IR trees without breaking the IR semantics of arrays. In fact the `Analyzer` already does something smart for `ArithmeticException`, by recognizing `Int_/` and friends and reaching `ArithmeticException` from there. It could do so for `ArraySelect`s, but instead it chooses to always reach `ArrayIndexOutOfBoundsException` in order not to have to track them. (We did the tracking for `ArithmeticException` because they are not gated by undefined behavior, and would therefore always be required, but that does not change the fundamental analysis.)

> > instantiateClass(ClassClass, ObjectArgConstructorName)
> 
> Private contract between our implementation of java.lang.Class and the CoreJS lib of how class data is passed. The `ClassClass` itself could arguably be reached by the Analzyer itself _iff_ any class data is accessed. But I'm not sure it is worth doing that.

This is part of the public contract of the IR. It's even specified at https://lampwww.epfl.ch/~doeraene/thesis/sjsir-semantics/#sec-sjsir-classof

> I'm not quite sure how Hijacked classes change this picture, but I'm sure you are :)

Hijacked classes are always reached because JS code can instantiate them through just creating literal values ^^ As such they are already correctly reached by the `Analyzer`.

gzm0:
Updated. I believe I have addressed all concerns (I moved the `symbolRequirement` thing into a separate issue so we can move forward with this).

gzm0:
Hah, now this is getting interesting. It seems CI fails for Scala versions != 2.12 when deserializing some IR. I'll try to reproduce locally.

gzm0:
So, the problem was that the minor scala version for the linker private library was still set to 2.12.8, but the compiler was defaulted to 2.12.10. Therefore, the build picked up a locally published compiler that used a different IR.

This is another occurrence of a local publish that leaks between CI runs.

PR #3823, opened 2019-10-30
Fix two bugs related to global refs.

sjrd:
Based on #3814. Only the last two commits belong to this PR.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Trees.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Trees.scala
@@ -86,9 +86,15 @@ object Trees {
       ClassIdent(name, None)
   }
 
-  final def isValidJSIdentifier(name: String): Boolean = {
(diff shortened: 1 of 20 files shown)


ir/src/main/scala/org/scalajs/ir/Trees.scala
```diff
@@ -86,9 +86,13 @@ object Trees {
...
 
-  final def isValidJSIdentifier(name: String): Boolean = {
+  /** Tests whether the given name is a valid JavaScript identifier name.
+   *
+   *  This test does *not* exclude keywords.
+   */
+  def isJSIdentifierName(name: String): Boolean = {
     // scalastyle:off return
-    // This method is called on every Name construction; it should be fast.
+    // This method is called on every JSGlobalRef construction; it should be fast.
```
gzm0: It is also called on every `TopLevelExportDef` construction. Maybe find a more general wording?

sjrd:
Rebased and updated.

PR #3825, opened 2019-10-31
Fix #2680: Use Symbols for the private fields of JS classes.

### LINKED ISSUE #2680: Private fields in JS classes can clash with public or JS-defined members
See comment at https://github.com/scala-js/scala-js/pull/2663/files#r91359592

sjrd:
Based on #3814. Only the last commit belongs to this PR.

---

Any encoding of field names as identifiers that we could come up would inevitably be prone to clashes, as we can never know which fields would be declared in superclasses, or worse, in subclasses.

Therefore, we now create one `Symbol` for each private field declaration, and we use that. When Symbols are not supported by the engine, we dynamically generate random names that have the same entropy as UUIDs. This does not guarantee that there will never be any clash, but it makes the likelihood close to zero.

There is no additional test, because there is no way to write a test that would have failed before this commit, and would still be relevant after this commit, given that the compilation has completely changed. Even removing the constraint that a test would need to fail before the commit, I cannot think of any test that would be relevant for this commit.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
@@ -288,19 +288,16 @@ private[emitter] final class JSGen(val semantics: Semantics,
     envField("t", className, item.name, item.originalName)
   }
 
-  /* The similarity with `genSelect` is accidental. It will probably evolve in
-   * the future, to emit selections of symbols that are private to the
-   * Scala.js-generated code.
-   */
-  def genJSPrivateSelect(receiver: Tree, cls: ClassRef, field: irt.FieldIdent)(
+  def genJSPrivateSelect(receiver: Tree, className: ClassName,
+      field: irt.FieldIdent)(
       implicit pos: Position): Tree = {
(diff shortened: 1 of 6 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -288,19 +288,17 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
   }
 
-  // The similarity with `genFieldIdent is accidental. See above.
-  def genJSPrivateFieldIdent(cls: ClassName, field: irt.FieldIdent)(
-      implicit pos: Position): Ident = {
-    Ident(genName(cls) + "__f_" + genName(field.name), field.originalName)
+  def genJSPrivateFieldIdent(className: ClassName, field: irt.FieldIdent)(
+      implicit pos: Position): Tree = {
+    envField("r", genName(className) + "__" + genName(field.name),
+        field.originalName)
```
gzm0: Use a helper that takes a `ClassName` and a `FieldName`?

sjrd:
Rebased and addressed the nit.

PR #3826, opened 2019-10-31
Fix #2627: IR-check duplicate definitions of Scala members.

### LINKED ISSUE #2627: IR Checker does not check for member name collisions

sjrd:
Based on #3814. Only the last 2 commits belong to this PR.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
@@ -85,7 +85,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
               if (classDef.kind == ClassKind.AbstractJSType) "Abstract"
               else "Native"
             reportError(
-                s"$kind JS type ${classDef.name} cannot have instance members")
+                i"$kind JS type ${classDef.name} cannot have instance members")
           }
         case _ =>
           checkScalaClassDef(classDef)
@@ -100,7 +100,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
     for (classCaptures <- classDef.jsClassCaptures) {
       if (classDef.kind != ClassKind.JSClass) {
         reportError(
-            s"Class ${classDef.name} which is not a non-native JS class " +
+            i"Class ${classDef.name} which is not a non-native JS class " +
             "cannot have class captures")
       }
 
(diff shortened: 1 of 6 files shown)


linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1347,6 +1347,36 @@ object IRChecker {
...
+
+      val pi = self.parts.iterator
+      val ai = args.iterator
+      val builder = new java.lang.StringBuilder(processEscapes(pi.next()))
+      while (ai.hasNext) {
+        builder.append(format(ai.next()))
+        builder.append(processEscapes(pi.next()))
+      }
+      builder.toString()
+    }
```
gzm0: `self.s(args.map(format _) : _*)` instead of all this?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -164,6 +165,37 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+      field @ FieldDef(flags, FieldIdent(name, _), _) <- classDef.fields
+    } {
+      implicit val ctx = ErrorContext(field)
+
+      if (!declaredFields.add((flags.namespace.isStatic, name))) {
+        reportError(
+            i"Duplicate definition of field '$name' in class " +
+            i"'${classDef.encodedName}'")
+      }
+    }
```
gzm0: Have you considered simply using a group by?

```scala
for {
  (id, dupes) <- classDef.fields.groupBy(f => (f.flags.namespace, f.name.name))
  if dupes.size > 1
} {
  reportError(...)
}
```

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -164,6 +165,37 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
     }
   }
 
+  private def checkDuplicateMembers(classDef: LinkedClass): Unit = {
+    // Fields
+    val declaredFields = mutable.Set.empty[(Boolean, FieldName)]
+    for {
+      field @ FieldDef(flags, FieldIdent(name, _), _) <- classDef.fields
+    } {
+      implicit val ctx = ErrorContext(field)
```
gzm0: I'm not sure this is the right `ErrorContext`. The field definition on its own is fine, it's the class def that is broken, no?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -164,6 +165,37 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+
+      if (!declaredFields.add((flags.namespace.isStatic, name))) {
+        reportError(
+            i"Duplicate definition of field '$name' in class " +
```
gzm0: Also include the namespace here?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
sjrd: Not sure it's better, but I've applied your suggestion.

sjrd: Indeed, that's better. Updated.

PR #3831, opened 2019-11-04
Check that methods called exist on the static type, even if abstract.

sjrd:
Previously, the following program would link:

```scala
class A
class B {
  def foo(): Int = 1
}
val a: A = new B()
a.foo()
```
even though `foo` is not declared in `A`, because `A` itself is never instantiated.

This is not a problem for the current JavaScript emitter, nor for the optimizer, but it can be problematic in the future for:

* IR-checking that method calls are correct (#1419)
* a potential Wasm back-end, which might want to rely on the abstract methods being declared to build virtual method tables.

This commit performs the bare minimum that needs to be done before 1.0.0 in order to future-proof those two things: in the base linker (only), check that methods called exist on the static type of the argument, potentially as abstract methods.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
@@ -37,6 +37,7 @@ import Infos.{NamespacedMethodName, ReachabilityInfo}
 private final class Analyzer(config: CommonPhaseConfig,
     symbolRequirements: SymbolRequirement,
     allowAddingSyntheticMethods: Boolean,
+    checkAbstractReachability: Boolean,
     inputProvider: Analyzer.InputProvider,
     ec: ExecutionContext)
     extends Analysis {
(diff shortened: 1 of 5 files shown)


linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -942,6 +969,16 @@ private final class Analyzer(config: CommonPhaseConfig,
...
         subclass.callMethodResolved(methodName)
+
+      if (checkAbstractReachability) {
+        /* Also lookup the method as abstract from this class, to make sure it
+         * is *declared* on this type. We do this after the concrete lookup to
+         * avoid work, since a concretely reachable method is already marked as
+         * abstractly reachable.
+         */
+        if (!methodName.isReflectiveProxy)
+          lookupAbstractMethod(methodName).reachAbstract()
```
gzm0: Hum, don't we have to reach on *all* abstract definitions? Otherwise the value of `isAbstractReachable` will be non-deterministic (or at least dependent on the order of interfaces).

linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala
```diff
@@ -483,7 +483,7 @@ object AnalyzerTest {
       loader <- Future.traverse(stdlib.toList)(_.loader).map(_.headOption)
       analysis <- Analyzer.computeReachability(CommonPhaseConfig(),
           symbolRequirements, allowAddingSyntheticMethods = true,
-          inputProvider(loader))
+          checkAbstractReachability = true, inputProvider(loader))
```
gzm0: Could you add a tests for:
- Failure.
- `isAbstractReachable`?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
gzm0: Note: I'm OK with simply not exposing `isAbstractReachable` in which case this would be fine IIUC.

sjrd: No, we don't have to reach all abstract definitions. We just need to make sure that one is reachable. The choice is actually deterministic: it is a function of the superclasses and the interfaces that they implement. Note that the interface list is ordered, so this is deterministic.

Hopefully the test I added will make it obvious.

PR #3833, opened 2019-11-05
Cleanups after the big Name refactoring

sjrd:


ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -22,37 +22,96 @@ import java.nio.CharBuffer
...
+   *
+   *  The contents of an `EncodedName` is guaranteed to be a well-formed UTF-8
+   *  string.
+   *
+   *  @note
+   *    `equals()` and `hashCode()`, along with `==` and `##`, are just as
+   *    broken for `EncodedName` as for `Array`s. Use the methods in the
+   *    companion object instead.
+   */
+  final class EncodedName private (private[ir] val bytes: Array[Byte])
```
gzm0: I like how this gives a neat separation between base encoding and higher level name guarantees. In this line: Have you considered putting this into a separate file (so the utf8 encoding methods would get closer together without increasing noise in this file).

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -106,14 +106,16 @@ object Serializers {
     final val FormatNoPositionValue = -1
   }
 
-  private final class ByteString(val bytes: Array[Byte]) {
+  private final class EncodedNameKey(val encoded: EncodedName) {
```
gzm0: If `EncodedName` gets the equals and hash code directly, this class can simply disappear, right?

ir/src/main/scala/org/scalajs/ir/Names.scala
gzm0: Also, consider simply calling this `UTF8String`.

ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -22,37 +22,96 @@ import java.nio.CharBuffer
...
+  /** An immutable name encoded in UTF-8.
+   *
+   *  The contents of an `EncodedName` is guaranteed to be a well-formed UTF-8
+   *  string.
+   *
+   *  @note
+   *    `equals()` and `hashCode()`, along with `==` and `##`, are just as
+   *    broken for `EncodedName` as for `Array`s. Use the methods in the
+   *    companion object instead.
+   */
```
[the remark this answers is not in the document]
sjrd: Unfortunately, overriding `equals` and `hashCode` is prohibited in an `AnyVal`. The compiler rejects it. I never understood why that restriction exists, but there's nothing I can do about it now.

ir/src/main/scala/org/scalajs/ir/Serializers.scala
sjrd: Yes, but as I said above, `EncodedName` cannot get `equals` and `hashCode` :(

ir/src/main/scala/org/scalajs/ir/Names.scala
sjrd: Done.

ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -22,37 +22,96 @@ import java.nio.CharBuffer
...
+      new EncodedName(validateUTF8(bytes))
+
+    def equals(encoded1: EncodedName, encoded2: EncodedName): Boolean =
+      java.util.Arrays.equals(encoded1.bytes, encoded2.bytes)
+
+    def hashCode(encoded: EncodedName): Int =
+      scala.util.hashing.MurmurHash3.bytesHash(encoded.bytes)
+
+    private[Names] def concat(encoded1: EncodedName,
+        encoded2: EncodedName): EncodedName = {
```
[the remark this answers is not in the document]
sjrd: I called it `++` instead. `+` only means concatenation for `String`s, and it has the problem of allowing `String + Anything` without any check. `++` consistently means concatenation for collections in Scala.

ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -234,10 +234,10 @@ object Names {
     protected def stringPrefix: String = "SimpleMethodName"
 
     /** Returns `true` iff this is the name of an instance constructor. */
-    def isConstructor: Boolean = this == ConstructorSimpleName
+    def isConstructor: Boolean = encoded(0) == '<' && encoded(1) == 'i'
 
     /** Returns `true` iff this is the name of a static initializer. */
-    def isStaticInitializer: Boolean = this == StaticInitializerSimpleName
+    def isStaticInitializer: Boolean = encoded(0) == '<' && encoded(1) == 'c'
```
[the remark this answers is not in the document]
sjrd: Also this avoids defining some methods of `SimpleMethodName` in terms of constant instances of `SimpleMethodName` itself.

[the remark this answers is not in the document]
sjrd: I've reverted that change in that commit, and instead I've added another commit later that uses an even better mechanism, using reference equality.

sjrd:
> Consider dropping the "the" in the message of the new commit so it's not too long.

Good point :)

PR #3837, opened 2019-11-08
Build partest coverage check into partest runner

gzm0:


sjrd:
Hum, IIUC, we are losing one functionality here: the fact that `check-partest-coverage.sh` was not only testing that every actual test is known, but also that we don't "know" a test that does *not*, in fact, exist.

sjrd:
Also: what's the motivation for this change?

gzm0:
> Hum, IIUC, we are losing one functionality here: the fact that check-partest-coverage.sh was not only testing that every actual test is known, but also that we don't "know" a test that does not, in fact, exist.

Ah, indeed. 

> Also: what's the motivation for this change?

Less code, less logic duplication (where is what).

But fair enough, maybe its not worth it.

PR #3838, opened 2019-11-08
Fix #3824: Move UndefinedBehaviorError into the linker library

### LINKED ISSUE #3824: Move UndefinedBehaviorError into linker library
Just adding this for tracking to make clear I'm actively working on this.

gzm0:
If we need UndefinedBehaviorError from a library in the future, we can make it available through a dedicated IR node.


javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -131,48 +123,129 @@ object System {
...
       case src: Array[AnyRef] =>
         dest match {
-          case dest: Array[AnyRef] => impl(src.length, dest.length, (i, j) => dest(i) = src(j))
-          case _                   => mismatch()
+          case dest: Array[AnyRef] =>
+            // bounds checks
+            src(srcPos)
+            src(srcPos + length)
+            dest(destPos)
+            dest(destPos + length)
```
gzm0: @sjrd if you know a better way to do this that will not reference the Scala runtime library, I'd be very happy :)

gzm0:
I forgot to hide `isJSType` (it's indeed unused now).

javalanglib/src/main/scala/java/lang/System.scala
sjrd: I actually think that's a pretty clever way to do it :p

However there are a few issues I think:

* you forgot to check for `length < 0`, which you can do with:
  ```scala
  if (length < 0)
    src(length)
  ```
* you will incorrectly throw if `srcPos + length == src.length` or `destPos + length == dest.length`; you should use `src(srcPos + length - 1)` and `dest(destPos + length - 1)`
* when `length == 0`, you should also accept `srcPos == src.length` and `destPos == dest.length`

Overall, it seems easier to always do the test that was done before in `impl`, but if it's `true`, then do `src(-1)` to force throwing the appropriate exception. You can do that in `impl` without relying on the stdlib with:
```scala
if (srcPos < 0 || destPos < 0 || length < 0 ||
    srcPos > srcLen - length || destPos > destLen - length) {
  f(-1, -1) // throw an ArrayIndexOutOfBoundsException or UBE
}
```
This is not really efficient because it cannot be dead-code-eliminated when the exception is unchecked, but it doesn't matter, because this whole `arraycopy` implementation is by-passed when the optimizer is enabled (it's an intrinsic). We only need a *correct* implementation, not an efficient one.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -617,11 +616,11 @@ object Arrays {
   }
 
   @inline private def checkIndicesForCopyOfRange(
-      len: Int, start: Int, end: Int): Unit = {
+      a: Array[_], start: Int, end: Int): Unit = {
```
sjrd: You should `@specialize` this function if you do that, otherwise it will go through a run-time type test to do `a(start)`.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -815,13 +814,13 @@ object Arrays {
   }
 
   @inline
-  private def checkRangeIndices(length: Int, start: Int, end: Int): Unit = {
+  private def checkRangeIndices(a: Array[_], start: Int, end: Int): Unit = {
```
sjrd: Same here: this method should be `@specialized`.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -617,11 +616,11 @@ object Arrays {
...
 
   @inline private def checkIndicesForCopyOfRange(
-      len: Int, start: Int, end: Int): Unit = {
+      a: Array[_], start: Int, end: Int): Unit = {
     if (start > end)
       throw new IllegalArgumentException("" + start + " > " + end)
-    SemanticsUtils.arrayIndexOutOfBoundsCheck(start < 0 || start > len,
-        new ArrayIndexOutOfBoundsException)
+
+    a(start) // bounds check
```
sjrd: This does not do the same thing as before if `start == a.length`; it now throws. I doubt this is correct.

javalib/src/main/scala/java/util/Arrays.scala
```diff
@@ -815,13 +814,13 @@ object Arrays {
...
     if (start > end)
       throw new IllegalArgumentException("fromIndex(" + start + ") > toIndex(" + end + ")")
-    SemanticsUtils.arrayIndexOutOfBoundsCheck(start < 0,
-        new ArrayIndexOutOfBoundsException("Array index out of range: " + start))
-    SemanticsUtils.arrayIndexOutOfBoundsCheck(end > length,
-        new ArrayIndexOutOfBoundsException("Array index out of range: " + end))
+
+    // bounds checks
+    a(start)
+    a(end)
```
sjrd: This does not do the same thing as before if `end == a.length` (including when `start == end == a.length`); it now throws. I doubt this is correct.

javalanglib/src/main/scala/java/lang/Class.scala
```diff
@@ -84,11 +85,7 @@ final class Class[A] private (data0: Object) extends Object {
 
   @inline // optimize for the Unchecked case, where this becomes identity()
   def cast(obj: Object): A = {
-    SemanticsUtils.asInstanceOfCheck({ () =>
-      (obj != null && !isJSType && !isInstance(obj))
-    }, { () =>
-      new ClassCastException("" + obj + " is not an instance of " + getName)
-    })
+    data.checkCast(obj)
```
sjrd: This is correct but it's less efficient than it used to be for the Unchecked case. Before, the optimizer was able to completely get rid of all the code, and the entire call to `cast(x)` collapsed to `x`. Now there is a call to `checkCast` (which itself doesn't do anything, but still).

That's not a deal-breaker, but we should at least remove the comment next to `@inline`, because it's clearly wrong now.

We could recover the previous optimization by making `Class.cast` an intrinsic; I'm not sure whether it's worth it, though.

sjrd:
After all this, do we still need `LinkingInfo.semantics` at all?

library/src/main/scala/scala/scalajs/runtime/LinkingInfo.scala
```diff
@@ -16,41 +16,15 @@ import scala.scalajs.js
...
 /** Information about link-time configuration of Scala.js. */
 sealed trait LinkingInfo extends js.Object {
-  /** Semantics configuration. */
-  val semantics: LinkingInfo.Semantics
-
   /** Whether we are assuming ECMAScript 6 support or not. */
   val assumingES6: Boolean
 
+  /** Whether we are linking in production mode. */
+  val productionMode; Boolean
```
sjrd:
```suggestion
 val productionMode: Boolean
```

sjrd:
The last commit ("Remove (mostly) unused LinkingInfo.Semantics") should also adapt `OptimizerCore` at
https://github.com/scala-js/scala-js/blob/e9368011ce648279759d40f8a78038ec6b4bb53a/linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala#L3920-L3963

Otherwise LGTM.

gzm0:
So I have been fiddling with the `ArrayOutOfBoundsException` and unfortunately my approach doesn't seem to work if we want the right error messages. From here, we seem to have multiple options:

1. Accept the different error messages.
1. Support the current setup in the IR.
1. Simply throw the errors unconditionally (probably plus benchmarks).

Personally I like option 3 most if the performance change is acceptable. It also is easiest to implement :)

sjrd:
I don't think the error messages are specified, so 1. should be perfectly reasonable.

gzm0:
> I don't think the error messages are specified, so 1. should be perfectly reasonable.

OK. I'll try. I'm a bit concerned that creating meaningless error messages can be confusing to users (e.g. the current code reports -1 as index in `System.arraycopy` even if the offending index is different).

gzm0:
Updated. Note that this needs very likely #3849 for CI to pass.

sjrd:
> Note that this needs very likely #3849 for CI to pass.

I probably won't have time to do a merge PR this weekend, but I could review one if you do it, in case you'd like to make progress this weekend. Otherwise I'll do it on Monday.

gzm0:
Updated (rebased onto the merge commit and added another commit).

javalanglib/src/main/scala/java/lang/Class.scala
```diff
@@ -32,9 +33,22 @@ private trait ScalaJSClassData[A] extends js.Object {
...
+   *  Directly accessing the `data` field from `@inline` methods will cause
+   *  scalac to make the field public and mangle its name. Since the Emitter
+   *  relies on the field being called exactly `data` in some of its
+   *  optimizations, we must avoid that.
+   *
+   *  This non-`@noinline` method can be used to access the field without
+   *  triggering scalac's mangling. Since it is a trivial accessor, the
+   *  Scala.js optimizer will inline it anyway.
+   */
+  private def getData(): ScalaJSClassData[A] = data
```
sjrd: I had to introduce this method. Since its comment why it was necessary.

linker/jvm/src/main/scala/org/scalajs/linker/backend/emitter/PrivateLibHolder.scala
```diff
@@ -19,7 +19,11 @@ import org.scalajs.linker.standard.MemIRFileImpl
 
 object PrivateLibHolder {
   private val relativeDir = "scala/scalajs/runtime/"
-  private val sjsirNames = Seq("RuntimeLong.sjsir", "RuntimeLong$.sjsir")
+  private val sjsirNames = Seq(
+      "RuntimeLong.sjsir",
+      "RuntimeLong$.sjsir",
+      "UndefinedBehaviorError.sjsir"
```
sjrd: I removed the trailing comma here, so that this compiles in 2.11.x and 2.12.1.

javalanglib/src/main/scala/java/lang/Class.scala
gzm0: Hah, I would not have found this so quickly :P I also don't really understand where this is the case (that the `Emitter` needs to know this name). Could you point me to the code?

gzm0: Oh, I found it. The `zero` and `classDataOf` helpers fall back to this. This seems to be an inherent property of the optimization (that we want to retrieve the data again). So it seems we are stuck with it.

(An alternative would be to expose a public method, but because we are too late in the flow, we wouldn't get inlining, etc.)

sjrd: Yes, that's exactly the code and the problem.

gzm0:
LGTM

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -103,13 +103,15 @@ object System {
...
-      })
+      /* Perform dummy swaps to trigger an ArrayIndexOutOfBoundsException or
+       * UBE if the positions / lengths are bad.
+       */
+      if (srcPos < 0 || destPos < 0)
+        f(destPos, srcPos)
+      if (length < 0)
+        f(length, length)
+      if (srcPos > srcLen - length || destPos > destLen - length)
+        f(destPos + length, srcPos + length)
```
sjrd: The `destPos` and `srcPos` were in the wrong order (`f` takes a position in `dest` first, and a position in `src` second). I swapped them to the correct order to fix the latest CI failure (which happened only in Compliant mode without optimizer).

gzm0: Ouch :P

PR #3842, opened 2019-11-11
Support `Match` nodes with non-int-literals in the back-end.

sjrd:
See https://github.com/scala/scala/pull/8451 upstream (not yet merged, but fully approved)

---

Since Scala 2.13.2, the pattern matcher will keep `Match` nodes that match on `String`s and `null`s as is, to be desugared later in `cleanup`. This was implemented upstream in https://github.com/scala/scala/pull/8451

We implement a more general translation that will accept any kind of scalac `Literal`. If the scrutinee is an integer and all cases are int literals, we emit a `js.Match` as before. Otherwise, we emit an `if..else` chain with `===` comparisons.

---

Locally tested with:
```
> set resolvers in Global += "scala-pr-validation" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/"
> ++2.13.2-bin-9f25ca0-SNAPSHOT
> testSuite/test
> compiler/test
[...] copy the big test case from the PR in HelloWorld.scala and:
> helloworld/run
```

---

As noted at https://github.com/scala/scala/pull/8451#issuecomment-552495496, 0.6.x already works out of the box with the change upstream.

It actually raises the question whether we should revert the design of our `js.Match` to accept any kind of `js.Literal`s and really iron out the linker (notably for IR literals that are not JS literals, such as `ClassOf` and `LongLiteral`) instead of changing the compiler back-end. This PR implements the less disruptive change.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -3230,15 +3230,13 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
       implicit val pos = tree.pos
       val Match(selector, cases) = tree
 
-      /* We adapt the selector to IntType so that we can use it in a js.Match,
-       * just like GenBCode does for the JVM. This seems to be redundant,
-       * though, as anything that comes out of the pattern matching has already
-       * been adapted to an Int (along with the cases). However, since GenBCode
-       * adapts, we do the same, to be on the safe side (for example, a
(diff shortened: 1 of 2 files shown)


gzm0:
> It actually raises the question whether we should revert the design of our js.Match to accept any kind of js.Literals and really iron out the linker (notably for IR literals that are not JS literals, such as ClassOf and LongLiteral) instead of changing the compiler back-end.

We should keep this as a a consideration, especially as it would simplify the emitted code IIUC.

PR #3845, opened 2019-11-14
Fix #3827: Review original names

### LINKED ISSUE #3827: Review original names
See https://github.com/scala-js/scala-js/pull/3814#discussion_r340437033

sjrd:
* Encode original names as optional `UTF8String`s.
* Store original names only on definitions, not on every usage.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
@@ -18,6 +18,7 @@ import scala.collection.mutable
 
 import org.scalajs.ir._
 import org.scalajs.ir.Names._
+import org.scalajs.ir.OriginalName.NoOriginalName
 import org.scalajs.ir.Position._
 import org.scalajs.ir.Transformers._
 import org.scalajs.ir.Trees._
@@ -396,15 +397,17 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
 
     // Record names
 
-    def makeRecordFieldIdent(tree: RecordSelect)(
+    def makeRecordFieldIdentForVarRef(tree: RecordSelect)(
         implicit pos: Position): js.Ident = {
 
       val recIdent = (tree.record: @unchecked) match {
         case VarRef(ident)                 => transformLocalVarIdent(ident)
         case Transient(JSVarRef(ident, _)) => ident
-        case record: RecordSelect          => makeRecordFieldIdent(record)
+        case record: RecordSelect          => makeRecordFieldIdentForVarRef(record)
       }
-      makeRecordFieldIdent(recIdent, tree.field.name, tree.field.originalName)
+
+      // Since this is only used for VarRefs, we never need an original name
+      makeRecordFieldIdent(recIdent, tree.field.name, NoOriginalName)
     }
 
     def makeRecordFieldIdent(recIdent: js.Ident,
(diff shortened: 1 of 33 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -404,7 +405,7 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
         case Transient(JSVarRef(ident, _)) => ident
         case record: RecordSelect          => makeRecordFieldIdent(record)
       }
-      makeRecordFieldIdent(recIdent, tree.field.name, tree.field.originalName)
+      makeRecordFieldIdent(recIdent, tree.field.name, NoOriginalName)
```
gzm0: Ugh, this is not nice... So depending on how we construct the full record field identifier we are able to retain the original name or not :(

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -220,6 +220,38 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+        return false
+      var i = 0
+      while (i != name.length) {
+        if (name(i).toInt != jsName.charAt(i).toInt)
+          return false
+        i += 1
+      }
+      true
+      // scalastyle:on return
+    }
```
gzm0: Hum, this looks wrong. It compares an UTF16 string to an UTF8 string byte by byte...

IIUC, this is merely an optimization to *not* generate an `OriginalName` when it is definitely not necessary. If this is the case, please comment.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -4761,8 +4761,8 @@ private[optimizer] object OptimizerCore {
    *  Even though it is not encoded in the type system, `localDef.replacement`
    *  must be a [[ReplaceWithVarRef]] or a [[ReplaceWithRecordVarRef]].
    */
-  private final case class PreTransBinding(localDef: LocalDef,
-      value: PreTransform) {
+  private final case class PreTransBinding(originalName: OriginalName,
+      localDef: LocalDef, value: PreTransform) {
```
gzm0: Is the order of members intended here?

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -5499,6 +5499,12 @@ private[optimizer] object OptimizerCore {
         private[FreshNameAllocator] val usedNamesToNextCounter: Map[N, Int])
   }
 
+  def originalNameForFresh(base: Name, originalName: OriginalName,
+      freshName: Name): OriginalName = {
+    if (originalName.isDefined || (freshName eq base)) originalName
+    else OriginalName(base)
+  }
+
```
gzm0: Providing originalName handling as a separate method seems to make it easy to get it wrong: Since original names are only at definitions sites now, every time one requests a new fresh name, one needs to deal with the original name (and only then non-trivial handling is required). Have you considered moving this into the fresh methods directly?

gzm0:
It might also be worth mentioning somewhere that this forces original names to be well-formed unicode strings (no unmatched surrogates and what not).

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
sjrd: That's a very good point! I changed it in `OptimizerCore`, but I'm not sure where else you think I should do that. In the compiler backend, the `fresh` methods are used only for synthetics and from `encodeLocalSym`, and for the latter we have the `originalNameOfLocal` which forces to do the right thing.

sjrd:
> It might also be worth mentioning somewhere that this forces original names to be well-formed unicode strings (no unmatched surrogates and what not).

I wrote that on the Scaladoc comment of `OriginalName` and in the commit message.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
gzm0: Right, in the compiler backend that does not apply.

PR #3851, opened 2019-11-15
Fix #3840: Rename `envField` to `codegenVar`.

### LINKED ISSUE #3840: Rename `envField` in the emitter
See https://github.com/scala-js/scala-js/pull/3820#discussion_r341662906

sjrd:
`envField` was a historical name that can traced back to when all compiler-generated variables were stored as fields of a global object `ScalaJS`. The name was not too bad until recently, when we started using the mechanism even for some local variables.

This commit renames all "env field" occurrences to "codegen var", since those are essentially codegen-chosen variable names.


gzm0:
Actually the commit subject should be present tense :)

sjrd:
Oops, yes, fixed.

PR #3852, opened 2019-11-15
Finalize all case classes and privatize AnyVal underlying vals.

sjrd:
The case classes for IR Trees and JS Trees are `sealed` instead, because making them `final` triggers bugs with Scala 2.11.x and 2.12.{1-4}, in combination with their `implicit val pos`.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Trees.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Trees.scala
@@ -1120,10 +1120,8 @@ object Trees {
 
   // Miscellaneous
 
-  final class OptimizerHints private (val __private_bits: Int) extends AnyVal {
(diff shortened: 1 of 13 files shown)


gzm0:
> The case classes for IR Trees and JS Trees are `sealed` instead,
because making them `final` triggers bugs with Scala 2.11.x and
2.12.{1-4}, in combination with their `implicit val pos`.


I think this would be a useful comment in the relevant files.

sjrd:
> I think this would be a useful comment in the relevant files.

Indeed, done.

> I have not checked this PR for exhaustiveness, since I assume this was checked using automated tools.

Yes, it was something based on
```
$ git grep 'case class' | grep -v sealed | grep -v final
```
for the first one, and
```
$ git grep 'val __'
```
for the second.

PR #3854, opened 2019-11-16
FF ONLY Merge 0.6.x into master, November 16

gzm0:
```
$ git checkout -b merge-0.6.x-into-master-november-16
```

```
$ export mb=$(git merge-base scala-js/0.6.x scala-js/master)
```

```
$ git log --graph --oneline --decorate $mb..scala-js/0.6.x | cat
*   89928cd32 (scala-js/0.6.x, 0.6.x) Merge pull request #3855 from gzm0/fix-locale-string-test
|\  
| * 08de5c049 Fix BigInt.toLocaleString test
|/  
*   9fba684d2 Merge pull request #3849 from gzm0/array-select-side-effect
|\  
| * f18a9b4c8 Fix #3848: Classify ArraySelect as side effecting
* |   462889c64 Merge pull request #3850 from sjrd/blacklist-compiler-plugin-partest-2.12.9
|\ \  
| |/  
|/|   
| * c6fb0f5e8 Blacklist partest tests involving compiler plugins in 2.12.9.
|/  
*   2af728be3 (sjrd/0.6.x) Merge pull request #3847 from sjrd/fix-checksizes
|\  
| * dfbf71b75 Ignore 2.12.9 in checksizes, not 2.12.10.
|/  
* b13321948 Merge pull request #3786 from exoego/0.6-bigint
* dd2712bd0 Add js.BigInt for ES2019 (ES10)
```

```
$ git merge --no-commit scala-js/0.6.x
CONFLICT (modify/delete): tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala deleted in HEAD and modified in scala-js/0.6.x. Version scala-js/0.6.x of tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/emitter/FunctionEmitter.scala left in tree.
Auto-merging test-suite/js/src/test/scala/org/scalajs/testsuite/typedarray/TypedArrayTest.scala
Auto-merging partest-suite/src/test/resources/scala/tools/partest/scalajs/2.12.9/WhitelistedTests.txt
Auto-merging partest-suite/src/test/resources/scala/tools/partest/scalajs/2.12.9/BlacklistedTests.txt
Auto-merging ci/checksizes.sh
CONFLICT (content): Merge conflict in ci/checksizes.sh
Automatic merge failed; fix conflicts and then commit the result.
```

```
$ git commit
[merge-0.6.x-into-master-november-16 d2b6009d7] Merge '0.6.x' into 'master'
```

PR #3858, opened 2019-11-16
Always attach path context when reading IR

gzm0:
Because we are using buffers under the hood now, BufferUnderflowExceptions can happen as well.

At this point, it seems best to simply attach the path information to all exceptions.


gzm0:
- The single CI failure is due to the accidental push to master.
- This should wait on #3854, then it can be merged.

PR #3863, opened 2019-11-19
FF ONLY Towards 0.6.31.

sjrd:


sjrd:
:man_facepalming: That's why we review even those trivial commits, I guess ...

Fixed.

PR #3866, opened 2019-11-19
Constrain more the arguments of typed arrays' constructors and `set`.

sjrd:
We now only accept arguments whose element type is exactly the same as the typed array being created or modified. The API was previously too relaxed especially when mixing `number`-based typed arrays and `BigInt`-based typed arrays.

The new restrictions are technically stronger than necessary, but they make sense. Copies between typed arrays of different element types should be rare enough that using `js.Dynamic` should be fine.

We also constrain the argument to `set` that was a `js.Iterable` to be a `js.Array`. Contrary to the constructors, which accept any iterable, the `set` methods only accept `Array`s in addition to typed arrays.


gzm0:
> [...] exactly as the same as the typed [...]

Typo

sjrd:
Fixed.

PR #3867, opened 2019-11-19
Fix #3857: Use a test-local `.ivy2` directory in the CI.

### LINKED ISSUE #3857: Locally published binaries leak between CI runs
See https://github.com/scala-js/scala-js/pull/3817#issuecomment-554651054 and https://github.com/scala-js/scala-js/pull/3829#discussion_r344491031

sjrd:
This will isolate the locally-published artifacts of one CI run from other CI runs.

In order to preserve caching of artifacts downloaded from the Internet, we specifically create a symlink for the `cache` subdirectory.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -34,12 +34,12 @@ Otherwise, everything happens within sbt.
 
 Run the normal test suite using the entire Scala.js toolchain using
 
-    > testSuite/test
+    > testSuite2_12/test
 
 In order to test the tests themselves, run the cross-compiling tests on the JVM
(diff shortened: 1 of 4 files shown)


sjrd:
@fsalvi Could you have a look at this PR, to make sure we're not doing something horribly wrong with the CI machines?

sjrd:
Well, unsurprisingly, that unveiled another bug, this time in the scripted tests. I updated the PR with two more commits.

(Also I got an LGTM from Fabien about the CI changes.)

PR #3868, opened 2019-11-19
Swap the tag and position of trees in the sjsir file format.

sjrd:
With the new order, we don't have to store a position in the `None` case of `Option[Tree]`. Moreover, it opens more avenues if in the future we introduce other tags for which we might want different branches.

---

Also: Remove old (dead-code) debugging code in the IR serializers.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Serializers.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Serializers.scala
@@ -68,11 +68,6 @@ object Serializers {
     finally buf.order(o)
   }
 
-  // true for easier debugging (not for "production", it adds 8 bytes per node)
-  private final val UseDebugMagic = false
-  private final val DebugMagic = 0x3fa8ef84
-  private final val PosDebugMagic = 0x65f0ec32
-
   private object PositionFormat {
     /* Positions are serialized incrementally as diffs wrt the last position.
      *
@@ -553,8 +548,6 @@ object Serializers {
               "Cannot serialize a transient IR node (its value is of class " +
               s"${value.getClass})")
       }
-      if (UseDebugMagic)
-        writeInt(DebugMagic)
     }
 
     def writeTrees(trees: List[Tree]): Unit = {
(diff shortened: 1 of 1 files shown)


ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -1003,21 +1007,24 @@ object Serializers {
     }
 
     def readTree(): Tree = {
+      val tag = readByte()
       val pos = readPosition()
```
gzm0: Have you considered moving `readPosition()` into `readTreeFromTag`? That would simplify the call-sites.

sjrd: Good point. Done.

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -1003,31 +1007,33 @@ object Serializers {
     }
 
     def readTree(): Tree = {
-      val pos = readPosition()
-      readTreeFromTag(readByte())(pos)
+      val tag = readByte()
+      readTreeFromTag(tag)
     }
```
gzm0: `def readTree(): Tree = readTreeFromTag(readByte())`?

ir/src/main/scala/org/scalajs/ir/Serializers.scala
```diff
@@ -1003,31 +1007,33 @@ object Serializers {
...
     def readOptTree(): Option[Tree] = {
-      // TODO switch tag and position when we can break binary compat.
-      val pos = readPosition()
       val tag = readByte()
-      if (tag == TagEmptyTree) None
-      else Some(readTreeFromTag(tag)(pos))
+      if (tag == TagEmptyTree)
+        None
+      else
+        Some(readTreeFromTag(tag))
```
gzm0: Keep this in two lines?

sjrd:
I applied the two suggestions.

PR #3874, opened 2019-11-20
Determine supported IR versions automatically

gzm0:


ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
-   *  - `current`
-   */
-  val binaryEmitted: String = current
-
-  /** Versions whose binary files we can support (used by deserializer) */
-  val binarySupported: Set[String] = {
-    Set(binaryEmitted)
+  current = "1.0.0-SNAPSHOT",
+  /** Version of binary IR emitted by this version of Scala.js. */
+  binaryEmitted = "1.0-SNAPSHOT"
```
sjrd: These four lines should have 4 spaces of indentation.

sjrd: And the Scaladoc comments should be on the `val`s in `VersionChecks`. Actual arguments don't care about Scaladoc comments.

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+    final val binaryEmitted: String
+) {
+  import VersionChecks._
+
+  checkConsistent(current, binaryEmitted)
+
+  private val (binaryMajor, binaryMinor, binaryPreRelease) = parseBinary(binaryEmitted)
+
+  private val knownSupportedBinary = {
+    val m = new ConcurrentHashMap[String, Unit]()
```
sjrd: This looks like a prime use case for [`ConcurrentHashMap.newKeySet()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#newKeySet--).
```suggestion
    val m = ConcurrentHashMap.newKeySet[String]()
```
then using `add` instead of `put` and `contains` instead of `containsKey`.

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+private object VersionChecks {
+  private val binaryRE = """^([0-9]+)\.([0-9]+)(-.*)?$""".r
+
+  private def parseBinary(v: String): (Int, Int, Option[String]) = {
+    val m = mustMatch(binaryRE, v)
+    (m.group(1).toInt, m.group(2).toInt, Option(m.group(3)))
+  }
 
+  private def parseFull(v: String): (Int, Int, Int, Option[String]) = {
+    val fullRE = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
```
sjrd: Consider declaring that as a field of `VersionChecks`, similarly to `binaryRE`.

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+}
 
-  // Just to be extra safe
-  assert(binarySupported contains binaryEmitted)
+private object VersionChecks {
+  private val binaryRE = """^([0-9]+)\.([0-9]+)(-.*)?$""".r
+
+  private def parseBinary(v: String): (Int, Int, Option[String]) = {
+    val m = mustMatch(binaryRE, v)
+    (m.group(1).toInt, m.group(2).toInt, Option(m.group(3)))
```
sjrd:
```suggestion
    (m.group(1).toInt, m.group(2).toInt, Option(m.group(3)).map(_.substring(1)))
```
? So that the `-` does not belong to the result string.

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+  }
 
+  private def parseFull(v: String): (Int, Int, Int, Option[String]) = {
+    val fullRE = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
+    val m = mustMatch(fullRE, v)
+    (m.group(1).toInt, m.group(2).toInt, m.group(3).toInt, Option(m.group(4)))
```
sjrd:
```suggestion
    (m.group(1).toInt, m.group(2).toInt, m.group(3).toInt, Option(m.group(4)).map(_.substring(1)))
```
(see above)

ir/src/test/scala/org/scalajs/ir/VersionChecksTest.scala
```diff
@@ -0,0 +1,87 @@
...
+    ok("2.2.0-M1", "2.1")
+    ok("2.1.1-M1", "2.1")
+
+    // binary is newer
+    bad("2.1.0", "2.2")
+    bad("2.2.0-M1", "2.2")
+
+    // binary is pre-release in non-matching version.
+    bad("2.3.0", "2.2-M1") 
+    bad("2.2.1", "2.2-M1")
```
sjrd: Could you also add here
```scala
bad("2.3.0-M1", "2.2-M1") 
bad("2.2.0", "2.2-M1")
```

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+
+  private def checkConsistent(current: String, binary: String) = {
+    val (binaryMajor, binaryMinor, binaryPreRelease) = parseBinary(binary)
+    val (currentMajor, currentMinor, currentPatch, currentPreRelease) = parseFull(current)
+
+    require(currentMajor == binaryMajor, "major(current) != major(binaryEmitted)")
+
+    require(currentMinor >= binaryMinor, "minor(current) < minor(binaryEmitted)")
+
+    require(currentPreRelease.isEmpty ||
```
sjrd: Start the actual argument on the next line, since it spans multiple lines:
```scala
    require(
        currentPreRelease.isEmpty ||
        ...
```

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -12,32 +12,94 @@
...
+        currentMinor > binaryMinor ||
+        currentPatch > 0 ||
+        binaryPreRelease == currentPreRelease,
+        "current is older than binaryEmitted through pre-release")
+
+    require(binaryPreRelease.isEmpty ||
```
sjrd: Same. In addition, add parens or braces around the `&&` subexpression? I find it hard to read when we rely on the precedence of `&&` over `||`, especially in long expressions like this.

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
sjrd: Hum never mind that comment. Our implementation of `ConcurrentHashMap` does not support `newKeySet()`, so this suggestion won't bootstrap.

sjrd:
```
[error] .../sbt-plugin-test/project/project/../../../ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala:58:19: not found: type IRVersionNotSupportedException
[error]         throw new IRVersionNotSupportedException(version, binaryEmitted,
[error]                   ^
[error] one error found
```
It looks like we need to add `IRVersionNotSupportedException.scala` to https://github.com/scala-js/scala-js/blob/master/sbt-plugin-test/project/project/build.sbt

PR #3877, opened 2019-11-21
Fix publishing setup of irJS, loggingJS and linkerInterfaceJS.

### LINKED ISSUE #3876: irJS and loggingJS lack their crossVersion := ScalaJSCrossVersion.binary
This means that both `ir` and `irJS` try to publish under the same artifact name (and same with `logging`/`loggingJS`), so publishing to Sonatype (rightfully) fails.

I'm rolling back 1.0.0-RC1 because of this.

sjrd:
They were actually necessary because `MyScalaJSPlugin` sets `crossVersion` to `CrossVersion.binary`, not to `*ScalaJS*CrossVersion.binary`.

This reverts commit f826ad651e926d4e2a5a275255f66e750eaf1c6a.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -93,11 +93,6 @@ object MyScalaJSPlugin extends AutoPlugin {
         libDeps.filterNot(dep => blacklist.contains(dep.name))
       },
 
-      /* Most of our Scala.js libraries are not cross-compiled against the
-       * the Scala.js binary version number.
-       */
-      crossVersion := CrossVersion.binary,
-
       scalaJSLinkerConfig ~= (_.withCheckIR(true)),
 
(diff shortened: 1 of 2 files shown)


sjrd:
Ouch, `linkerInterface` and `linkerInterfaceJS` are not even listed in `publish.sh` :s

gzm0:
Actually: I'd like to advocate to switch the setup here. The projects that set `crossVersion := CrossVersion.binary` are special and not the other way around. Even if that causes a bit more duplication it's IMO clearer.

sjrd:
> Actually: I'd like to advocate to switch the setup here.

Ah, yes, that's maybe better, indeed.

sjrd:
Updated.

PR #3881, opened 2019-11-22
Fix #3880: Align ScalaJSCrossVersion with ScalaJSVersions

### LINKED ISSUE #3880: ScalaJSCrossVersion is inconsistent with ScalaJSVersions
ScalaJSVersions does not have a patch part in the binary version, however ScalaJSCrossVersion wants one.

Further, it seems that with the current setup, we'd emit `sjs1.0` as a suffix. IIUC this is wrong and we should emit simply `sjs1` as the minor part should be backwards compatible.

gzm0:
Since we dropped the patch version in the `binaryEmitted` version, the regex parsing in `ScalaJSCrossVersion` was wrong.

We move the entire parsing and translation logic into `VersionChecks` where we can properly test it and it is more easily available to non-sbt consumers.

Note that this introduces a behavioral change: Minor pre-release versions are cross versioned with their full version, not their major version only.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -34,6 +34,35 @@ class VersionChecks private[ir] (
 
   private val (binaryMajor, binaryMinor, binaryPreRelease) = parseBinary(binaryEmitted)
 
+  /** The cross binary version
+   *
+   *  This is the version advertised in artifacts released by Scala.js users.
+   *
+   *  For non-pre release versions, this is only the major version, since binary
+   *  minor versions are backwards compatible.
+   *
+   *  For pre-release versions, the story is a bit more complicated:
+   *
+   *  - Any pre-release version with a minor version == 0 is ''before'' the
+   *    final major version is released and typically fully breaking. Therefore,
+   *    the full [[binaryEmitted]] version is used.
+   *  - A SNAPSHOT pre-release version with minor version > 0 is a development
+   *    version with compatible binary version. Nothing should be published
+   *    using this version, but being able to ''read'' artifacts in the major
+   *    line is critical for fast testing. Therefore, only the major version is
+   *    used.
+   *  - A non-SNAPSHOT pre-release version with minor version > 0 is a milestone
+   *    or release candidate version that is published. As such, artifacts may
+   *    be published with it but will not be able to be read by the main line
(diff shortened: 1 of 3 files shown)


ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -34,6 +34,11 @@ class VersionChecks private[ir] (
 
   private val (binaryMajor, binaryMinor, binaryPreRelease) = parseBinary(binaryEmitted)
 
+  /** The binary version as advertised in released artifacts. */
+  final val binaryVersion =
```
sjrd: This does not strike me as a very good name. Essentially this is the version of the ecosystem of libraries. Perhaps name it `ecosystemVersion`? Or `artifactsVersion`?

gzm0: I guess `binaryVersion` is sbt jargon (`CrossVersion.binary` vs `CrossVersion.full`). How would you feel about `binaryCross`? (the `Version` suffix is pointless anyways; everything is a version in this class...).

sjrd: Yeah `binaryCross` sounds reasonable.

gzm0: What shall we do about `ScalaJSCrossVersion.currentBinaryVersion`? Remove?

gzm0:
Oh, also I just noticed that this PR changes the behavior for minor pre-release versions. I guess one could argue both ways:

- Pre-release *should* be binary compatible.
- Pre-release is not stable enough and errors could be there.

WDYT?

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
sjrd: Yes. Google and GitHub searches say that no one uses that.

sjrd:
> Oh, also I just noticed that this PR changes the behavior for minor pre-release versions. I guess one could argue both ways:
>
> * Pre-release should be binary compatible.
> * Pre-release is not stable enough and errors could be there.
>
> WDYT?

Hum, I don't think this change is desirable. When working with 1.2.0-SNAPSHOT, I expect to be able to resolve libraries published with 1.1.x. I shouldn't *publish* myself from that version, but that's not specific to Scala.js.

gzm0:
Updated and fully documented what I think the reasons for all these things are :P

PR #3889, opened 2019-11-28
Move sbt-plugin-test to scripted

gzm0:


gzm0:
I'm not 100% sure what to do about all the `project/` directories. Do you think a symlink would be acceptable?

sjrd:
Symlinks are a big no no on Windows, so copy paste is unfortunately preferable.

sbt-plugin/src/sbt-test/testing/multi-framework/build.sbt
```diff
@@ -0,0 +1,62 @@
+lazy val versionSettings = Def.settings(
+    version := scalaJSVersion,
+    scalaVersion := "2.12.10"
```
sjrd: Consider putting those `inThisBuild(...)` instead.

sbt-plugin/src/sbt-test/testing/multi-framework/project/build.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % sys.props("plugin.version"))
```
sjrd: Here you've inlined `sys.props(...)` but in other tests you used a separate `val scalaJSVersion`. Consider making this uniform, one way or the other.

sbt-plugin/src/sbt-test/testing/multi-framework/test
```diff
@@ -0,0 +1,5 @@
+> root/test
+> multiTestJS/testHtml
+> set scalaJSStage in Global := FullOptStage
```
sjrd: Does this actually *do* anything on its own? Are multiple `>` commands executed within the same session? My impression is that each `>` command reloads sbt, which means that any `set` command on its own is immediately discarded.

sbt-plugin/src/sbt-test/linker/non-existent-classpath/project/build.sbt
```diff
@@ -0,0 +1,5 @@
+val scalaJSVersion = sys.props("plugin.version")
+
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
+
+libraryDependencies += "org.scala-js" %% "scalajs-linker" % scalaJSVersion
```
sjrd: This seems unused.

sbt-plugin/src/sbt-test/cross-version/2.11/test
```diff
@@ -0,0 +1,8 @@
+> run
+> test
+> testHtml
+> clean
+> set scalaJSStage in Global := FullOptStage
```
sjrd: Same concern about the lonely `set` command.

sbt-plugin-test/build.sbt
```diff
@@ -33,20 +28,6 @@ lazy val noDOM = project.settings(baseSettings: _*).
   settings(
     name := "Scala.js sbt test w/o DOM",
     scalaJSUseMainModuleInitializer := true,
-    scalaJSModuleInitializers +=
-      ModuleInitializer.mainMethod("sbttest.noDOM.TestApp", "foo"),
```
sjrd: This change should be accompanied by removing `def foo()` in `TestApp.scala`.

sbt-plugin/src/sbt-test/cross-version/2.13/src/main/scala/sbttest/TestApp.scala
```diff
@@ -0,0 +1,12 @@
...
+
+object TestApp {
+
+  def main(args: Array[String]): Unit = {
+    println(Lib.foo("Hello World"))
+    println(Lib.sq(10))
+  }
+
+  // Used as a manual module initializer.
+  def foo(): Unit = ()
```
sjrd: This is unused. (it should also not appear in the 2.11 `TestApp.scala`, but that should be removed in a previous commit already (see my other comment on the topic).

sbt-plugin/src/sbt-test/cross-version/2.13/test
```diff
@@ -0,0 +1,8 @@
+> run
+> test
+> testHtml
+> clean
+> set scalaJSStage in Global := FullOptStage
```
sjrd: Again the lonely `set`.

sbt-plugin/src/sbt-test/cross-version/2.13/src/main/scala/sbttest/Lib.scala
```diff
@@ -0,0 +1,11 @@
+package sbttest.noDOM
```
sjrd: Perhaps use the opportunity to rename this package. The test doesn't have anything to do with DOM or noDOM anymore.

(same in the other sources, as well as in the 2.11 sources)

sjrd:
Happy to get rid of sbt-plugin-test :tada:

PR #3891, opened 2019-11-29
Generate static forwarders, and use them to call Java-defined stuff.

sjrd:
The kind of things you're *not* supposed to do while in RC ... but #3890 prompted me to investigate, and these changes break the contract of Java APIs at the IR level ...

The PR starts with 2 cleanup commits: one not to special-case impl classes so much (relying more on `isStaticMember` than on `isImplClass`) and a tiny one on the generation of the .sjsir files.

Then it proceeds with two changes impacting the IR contract (not really the IR spec itself, but the relationship between the IR and the world):

* Use static methods as the canonical way to call Java static methods (am I allowed to say "duh" here?)
* Use static methods in module initializers

Both changes break the binary "ecosystem", even if they don't break the IR per se. So either we do them in 1.0.0-RC2 or we stay forever with the status quo.

---

After these changes, the only thing that's not completely like the JVM in terms of JDK libs interop is access to public static *fields*. We access them through a static getter of the same name and type, instead of directly reading the field. This is necessary to ensure proper initialization of the field.

I thought about making these accessors more special, either with special names or with a separate namespace. This would avoid the theoretical possibility between such a field and a documented public static method of the same name, without parameter and with the type of the field as result type. But in the end I think this is *very* unlikely to cause an issue, and fixing it would add quite a lot of complexity.

We still don't have any contract for *setting* public static fields, but who declares *mutable* public static fields?...


sjrd:
partest 2.11 is broken in the last commit. I'll investigate once/if I have an overall go-ahead on the PR.

sjrd:
> Impl classes reach jscode as classes with static methods only?

That's right. Or rather as *interfaces* with static methods only.

> Calls to Java compiled static methods reach jscode as static calls on the class?

Not exactly. They are calls to methods on a fake companion module class, but marked with `isStaticMember = true`. That's why the call itself doesn't need any special case, but `encodeClassName` doesn't append `$` to module classes that are `isJavaDefined`, so that they "erase" to the non-module class. This is exactly how the JVM backend deals with this as well.

> After this PR, the only thing that uses static forwarders is module initializers and writing a Java static method in Scala?

Yes, that's right.
They *could* also be used by another language that calls Scala.js code and compiles to the IR, in exactly the same way that Java code calling Scala sees methods of module classes as static methods in their companion class.

gzm0:
So here are my thoughts about this:

The first two commits are a no-brainer (but IIUC they are also fully compatible).

Regarding the Java calling convention I'm less convinced that it is desirable. The way I look at it right now is the following: Scala introduces a mismatch between Java / Scala "static" calling convention. Both Scala JVM and Scala.js need to bridge this gap in some fashion; this is where things necessarily diverge, because Scala.js is unable to represent the full Java language spec in its IR (no public fields, no static initializers, IIRC).

Independently of that, we observe that calling "Java code" will always be a special case in both the Scala JVM and Scala.js compilers. So from that POV, I'm not sure it's a problem if Scala.js and Scala JVM are different, especially if even after this PR they are different for fields.

Regarding the optimization for static methods making it easier to elide module classes: It seems that this is an optimization strategy we should aim for for *all* of Scala.js code. (a naive approach could be: generate static forwarders in linker, rewrite call sites of form `ApplyStatic(LoadModule(), ...)` in the optimizer).

I'm aware this is probably an incomplete picture, just dumping my current thoughts so we can iterate.

sjrd:
> Regarding the optimization for static methods making it easier to elide module classes: It seems that this is an optimization strategy we should aim for for _all_ of Scala.js code. (a naive approach could be: generate static forwarders in linker, rewrite call sites of form `ApplyStatic(LoadModule(), ...)` in the optimizer).

Yes, that was not really compelling. I removed it from the commit message.

> The first two commits are a no-brainer (but IIUC they are also fully compatible).

They are indeed fully compatible, so if we don't do the last two, I would defer the first two post-v1.0.0.

> Regarding the Java calling convention I'm less convinced that it is desirable. The way I look at it right now is the following: Scala introduces a mismatch between Java / Scala "static" calling convention. Both Scala JVM and Scala.js need to bridge this gap in some fashion; this is where things necessarily diverge, because Scala.js is unable to represent the full Java language spec in its IR (no public fields, no static initializers, IIRC).
> 
> Independently of that, we observe that calling "Java code" will always be a special case in both the Scala JVM and Scala.js compilers. So from that POV, I'm not sure it's a problem if Scala.js and Scala JVM are different, especially if even after this PR they are different for fields.

I don't really have a strong opinion on this. It's just that it has always bothered me that a) we have static methods in the IR but b) we don't use IR static methods to represent Java static methods. It's a bit awkward. Fields we can rationally explain why they are different: because we don't have JVM-style static initializers, and that is explained by the nature of Scala.js, where classes are linked at link-time instead of being dynamically loaded with a run-time VM. But for methods there's no such objective reason.

We can look at it in a completely different way, though. We can say that the user-visible static API of a class really *is* stored in a companion module class. And we can also explain that by the lack of static initialization: a module class is a reification of the needs of static initialization. So we can motivate the status quo like that.

gzm0:
>  We can say that the user-visible static API of a class really is stored in a companion module class. 

I like this more. It is also how the API is written. So another way to look at it is that the class files we compile against are strange, but nothing else.

sjrd:
I thought more about this. There is one qualitative improvement that this PR brings. It definitively removes from the IR contract the notion of *companion* module class linked only by a `$` appended to a *name*. This is the only place in all of the IR contract where names are not black box identifiers.

I think it would be worth removing this name-based link between a class and a module class.

We can even imagine that a later version of Scala itself would change its encoding of module classes. That would require difficult hacks to reconcile with our IR. Although unlikely in the foreseeable future, this scenario highlights that currently our IR contract depends on things that are not under its control, which is not good.

gzm0:
I assume that by IR contract you mean the union of javalib calling contract and module initializer contract?

It is true that the quasi module class calling javalib calling convention is inconsistent with our efforts to make the javalib independent of Scala.

It is definitely the case that not having the magic `$` suffix in the `ModuleInitizalizer` is nicer, although since it is really only on the surface, I thought it is acceptable.

In the interest of taking the discussion forward, let's see if we can resolve more lower level concerns:
- Are we willing to commit that there will *never* be a public static Javalib field named the same as a method (is this even allowed in Java)? If yes, we can worry about setting public static fields if it happens (if we do not have a convention right now, we can define one when we need it).
- I'm not super happy about the algorithm that generates static forwarders, n.b. because it silently ignores things it cannot create forwarders for. AFAICT, this could make us victims to implementation details of scalac, in the sense that some static forwarders suddenly disappear. So this in turn means IIUC, that the only "safe" way to generate them is *exactly* the same way than scalac.

sjrd:
> I assume that by IR contract you mean the union of javalib calling contract and module initializer contract?

Yes, that's right.

> Are we willing to commit that there will _never_ be a public static Javalib field named the same as a method (is this even allowed in Java)? If yes, we can worry about setting public static fields if it happens (if we do not have a convention right now, we can define one when we need it).

It is unfortunately allowed in Java:
```java
// Test.java
public class Test {
  public static int FOO = 5;
  public static int FOO() {
    return 6;
  }

  public static void main(String[] args) {
    System.out.println(FOO);
    System.out.println(FOO());
  }
}
```
```bash
temp$ javac Test.java 
temp$ java -cp . Test
5
6
temp$ javap -cp . Test
Compiled from "Test.java"
public class Test {
  public static int FOO;
  public Test();
  public static int FOO();
  public static void main(java.lang.String[]);
  static {};
}
```
Perhaps we shouldn't be willing to commit to never supporting this. An alternative would be to encode those accessors with a character that is not a valid Java identifier part (but a valid JVM identifier part, so that it is valid in a `SimpleMethodName`). For example, the getter for `static int FOO` could be named `get-FOO`, and we could generalize that to a setter which would be `set-FOO`.

This has a non-negligible cost, though, because we'll have to annotate the JDK lib sources with some relevant annotations so that the back-end knows to emit the getters as `get-FOO()` instead of `FOO()`, as it currently does (for free, because that's how scalac encodes `val`s).

I am not sure that this cost outweighs the quite theoretical possibility that someone would define a public static field and a 0-arg public static method with the same name and type ...

> I'm not super happy about the algorithm that generates static forwarders, n.b. because it silently ignores things it cannot create forwarders for. AFAICT, this could make us victims to implementation details of scalac, in the sense that some static forwarders suddenly disappear. So this in turn means IIUC, that the only "safe" way to generate them is _exactly_ the same way than scalac.

The algorithm that I use is the same as scalac, except that it silently ignores *strictly less* things. In fact, in practice it will never ignore anything, because we do not otherwise define *public static* methods in a class. Perhaps we could report an error instead of silently dropping, to make that clearer?

gzm0:
> I am not sure that this cost outweighs the quite theoretical possibility that someone would define a public static field and a 0-arg public static method with the same name and type ...

Yeah fair enough. I also just realized that we have this problem also with static initializers: If a JDK thing starts to rely on them, we have a problem.

> Perhaps we could report an error instead of silently dropping, to make that clearer?

Yes, I think that would be much better.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSFiles.scala
```diff
@@ -30,31 +30,21 @@ trait GenJSFiles[G <: Global with Singleton] extends SubComponent {
...
-    val outfile = getFileFor(cunit, sym, suffix.getOrElse("") + ".sjsir")
+  def genIRFile(cunit: CompilationUnit, tree: ir.Trees.ClassDef): Unit = {
+    val outfile = getFileFor(cunit, tree.name.name.nameString, ".sjsir")
     val output = outfile.bufferedOutput
     try ir.Serializers.serialize(output, tree)
     finally output.close()
   }
 
-  private def getFileFor(cunit: CompilationUnit, sym: Symbol,
+  private def getFileFor(cunit: CompilationUnit, fullName: String,
```
gzm0: Pass in a `ClassName` here?

ci/checksizes.sh
```diff
@@ -34,20 +34,20 @@ REVERSI_OPT_GZ_SIZE=$(stat '-c%s' "$REVERSI_OPT.gz")
 
 case $FULLVER in
   2.11.12)
-    REVERSI_PREOPT_EXPECTEDSIZE=436000
-    REVERSI_OPT_EXPECTEDSIZE=94000
+    REVERSI_PREOPT_EXPECTEDSIZE=437000
+    REVERSI_OPT_EXPECTEDSIZE=95000
```
gzm0: If the optimizer removes the forwarders, why do we see a bump in code size?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -928,6 +969,128 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+
+      assert(moduleClass.isModuleClass, moduleClass)
+
+      val conflictingNames: Set[MethodName] = {
+        existingMembers.collect {
+          case js.MethodDef(flags, name, _, _, _, _)
+              if flags.namespace == js.MemberNamespace.PublicStatic =>
+            name.name
+        }.toSet
+      }
```
gzm0: So this should be `assert(!exisitngMembers.exists(_.flags.namespace == PublicStatic)`?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -928,6 +969,128 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+
+      def listMembersBasedOnFlags = {
+        // Copy-pasted from BCodeHelpers (it's somewhere else in 2.11.x)
+        val ExcludedForwarderFlags: Long = {
+          import scala.tools.nsc.symtab.Flags._
+          // Don't include DEFERRED but filter afterwards, see comment on `findMembers`
```
gzm0: This comment seems outdated? I can't find a method `findMembers`.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -928,6 +969,128 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+      scala.util.Properties.versionNumberString.startsWith("2.11.")
+
+    /** Adds static forwarders for the methods of a module class to a list of
+     *  existing members.
+     *
+     *  Precondition: `isCandidateForForwarders(moduleClass)` is true
+     */
+    def addStaticFowardersFromModuleClass(existingMembers: List[js.MemberDef],
+        moduleClass: Symbol)(
+        implicit pos: Position): List[js.MemberDef] = {
```
gzm0: I find the way this method works a bit strange: Why does it not simply generate the static forwarders and return them? Returning the first argument as well seems odd.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -4013,9 +4175,10 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
           } else platform.externalEquals
           // scalastyle:on line.size.limit
         }
-        val moduleClass = equalsMethod.owner
-        val instance = genLoadModule(moduleClass)
-        genApplyMethod(instance, equalsMethod, List(lsrc, rsrc))
+        if (BoxesRunTimeClass.isJavaDefined)
```
gzm0: Hum, when does this happen? Isn't BoxesRunTime in the scalalib?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -5931,7 +6094,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
     }
 
     /** Generate access to a static member */
-    private def genStaticMember(sym: Symbol)(implicit pos: Position) = {
+    private def genStaticField(sym: Symbol)(implicit pos: Position): js.Tree = {
       /* Actually, there is no static member in Scala.js. If we come here, that
```
gzm0: this comment seems outdated.

project/Build.scala
```diff
@@ -1344,9 +1344,10 @@ object Build {
              * implementation is in javalib.
              */
             val superMappings = (mappings in packageBin).value
-            val libraryMappings = superMappings.filter(
-                _._2.replace('\\', '/') !=
-                  "scala/scalajs/js/typedarray/TypedArrayBufferBridge$.sjsir")
+            val libraryMappings = superMappings.filter { mapping =>
+              !mapping._2.replace('\\', '/').startsWith(
+                  "scala/scalajs/js/typedarray/TypedArrayBufferBridge")
+            }
```
gzm0: Unrelated change?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
```diff
@@ -40,53 +40,50 @@ object ModuleInitializer {
   private val ArrayOfStringTypeRef =
     ArrayTypeRef(ClassRef(BoxedStringClass), 1)
 
-  /** Makes an [[ModuleInitializer]] that calls a zero-argument method returning
-   *  `Unit` in a top-level `object`.
+  /** Makes an [[ModuleInitializer]] that calls a static zero-argument method
```
gzm0: an -> a while you are here? (also below)

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1359,12 +1359,12 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
     implicit val pos = Position.NoPosition
 
     ModuleInitializerImpl.fromModuleInitializer(moduleInitializer) match {
-      case VoidMainMethod(moduleClassName, mainMethodName) =>
-        js.Apply(genLoadModule(moduleClassName) DOT genName(mainMethodName), Nil)
+      case VoidMainMethod(className, mainMethodName) =>
+        js.Apply(codegenVar("s", className, mainMethodName), Nil)
```
gzm0: IIUC you do not need `TreeDSL` anymore.

ci/checksizes.sh
sjrd: Oh that's because of the following situation. Let's say you have the following IR before the change:
```scala
module class Foo$ extends Object {
  // typically would not be inlined
  def bar(x: int, y: int): int = {
    val z = x * y
    x + y + z
  }
}

def main() {
  mod:Foo$.bar(someInt(), someOtherInt())
}
```
Since the call to `mod:Foo$.bar` is not inlined, its arguments stay as is, their expressions inline in the parameters at the JS level:
```js
main__AT__V() {
  $m_Foo$().bar(someInt(), someOtherInt())
}
```
However, now if we have the IR with the static methods:
```scala
class Foo extends Object {
  static def bar(x: int, y: int): int = {
    mod:Foo$.bar(x, y)
  }
}

def main() {
  Foo::bar(someInt(), someOtherInt())
}
```
The call to `Foo::bar` is now inlined, so its non-pure arguments must be assigned to temporary `val`s. Since `someInt()` and `someOtherInt()` are not known to be pure, the JS code for main ends up being:
```js
main__AT__V() {
  const x$1 = someInt();
  const x$2 = someOtherInt();
  $m_Foo$().bar(x$1, x$2)
}
```
which is longer in terms of character count.

This situation already happens before this PR for all sorts of "fowarder" methods. We could one day improve this in the optimizer by identifying that forwarder methods can be inlined by directly splicing their actual arguments in place of the formal parameters, if all the formal parameters are used exactly once, in the same evaluation order, before any non-pure operation is performed by the body of the method. In particular, that would be easy to prove for any static forwarder if the module has a pure constructor (which is an analysis we already perform).

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: I preferred using a nice `reporter.error` with a position, so that potential bug reports can be more useful. Also I'm doing it only when there is a conflict, since I can imagine a situation, in the future, where scalac would generate private static methods in a class for its own purposes, that would then be made public to be accessible from a different class, but in that case they would have a mangled name, which could not collide with the static forwarders (we do not generate forwarders for methods with mangled names).

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: I removed it. It made sense in the scalac codebase, but here it doesn't. It's enough to know that the code was copied from scalac.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: It is in the scalalib, but upstream it is written [in Java](https://github.com/scala/scala/blob/2.13.x/src/library/scala/runtime/BoxesRunTime.java), so under separate compilation it is Java-defined.

project/Build.scala
sjrd: It's related. Now the object `TypedArrayBufferBridge$` emits a companion class `TypedArrayBufferBridge` with its static accessors, which must be ignored as well here.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
gzm0: > I preferred using a nice reporter.error with a position, so that potential bug reports can be more useful. 

That's fair.

> Also I'm doing it only when there is a conflict [...]

Since this is not necessary right now, personally, I'd feel more comfortable with checking the (more rigid) assumptions we are making right now and adjusting when necessary. (Of course unless there is already something like that in the pipeline and we know it will break soon).

sjrd: Well there's https://github.com/scala/scala/pull/8509 in the pipeline, which, IIUC, will generate public static methods for default getters. Default getters have mangled names, so they shouldn't collide with static forwarders.

sjrd: Although that PR upstream is not going to be relevant any time soon, since it breaks binary compat and therefore would only go to 2.14.x, not 2.13.x.

sjrd: > Since this is not necessary right now, personally, I'd feel more comfortable with checking the (more rigid) assumptions we are making right now and adjusting when necessary.

Done now.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -928,6 +970,131 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+    }
+
+    private lazy val dontUseExitingUncurryForForwarders =
+      scala.util.Properties.versionNumberString.startsWith("2.11.")
+
+    /** Gen the static forwarders for the methods of a module class.
+     *
+     *  Precondition: `isCandidateForForwarders(moduleClass)` is true
+     */
+    def genStaticFowardersFromModuleClass(existingMembers: List[js.MemberDef],
```
gzm0: Typo: Fo_warders

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -928,6 +970,131 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
       }
     }
 
+    // Static forwarders -------------------------------------------------------
+
+    /* This mimics the logic in BCodeHelpers.addForwarders and the code that
+     * calls it, except that we consider as collisions only methods with the
+     * same `MethodName` (including signature) and that are also
+     * `PublicStatic`. Anything else is not a collision as far the IR is
```
gzm0: This explanation seems to be out of date now?

PR #3895, opened 2019-12-09
Make `binaryCross` version same as major for minor milestones.

sjrd:
For example, in version 1.2.0-M1 that emits IR v1.2-M1, the binary cross version is now `1` instead of `1.2-M1`. This makes the behavior between snapshots and milestones uniform.

More importantly, this change allows to always derive the binary cross from the full Scala.js version, using 2 simple rules:

* If the version is `x.y.z-w`, then the binary cross is `x.y-w`.
* If the version is `x.y.z`, then the binary cross is `x`.

Before this change, in a situation where the full Scala.js version is a milestone (e.g., 1.3.0-M1), it was not possible to derive the binary cross version from it. It might be 1.3-M1 if the binary emitted version was 1.3-M1, but it might also be 1 if the binary emitted version was 1.2, for example.

Some build tools do not have access to the binary emitted version; they only have access to the full Scala.js version. It is important for them to be able to derive the binary cross version from the full version only.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -34,31 +34,19 @@ class VersionChecks private[ir] (
 
   private val (binaryMajor, binaryMinor, binaryPreRelease) = parseBinary(binaryEmitted)
 
-  /** The cross binary version
+  /** The cross binary version.
    *
    *  This is the version advertised in artifacts released by Scala.js users.
    *
-   *  For non-pre release versions, this is only the major version, since binary
-   *  minor versions are backwards compatible.
-   *
-   *  For pre-release versions, the story is a bit more complicated:
-   *
-   *  - Any pre-release version with a minor version == 0 is ''before'' the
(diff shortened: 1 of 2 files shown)


sjrd:
I noticed this problem when adding support for Scala.js 1.x in Mill (https://github.com/lihaoyi/mill/pull/738).

ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
```diff
@@ -57,8 +57,7 @@ class VersionChecks private[ir] (
    *    versions. Therefore, the full [[binaryEmitted]] version is used.
    */
   final val binaryCross: String = {
-    val needsFull =
-      binaryPreRelease.fold(false)(_ != "SNAPSHOT" || binaryMinor == 0)
+    val needsFull = binaryPreRelease.isDefined && binaryMinor == 0
```
gzm0: The comment on this method needs to be updated.

sjrd: Indeed. Updated.

PR #3902, opened 2019-12-16
No special casing for inlineable inits

gzm0:


gzm0:
A note about branching: I propose to merge post-v1 work into the `post-v1.0.0` branch and *rebase* this branch on top of master until `1.0.0` is released.

sjrd:
Rebasing a branch that contains merge commits unfortunately flattens out all the merge commits. So I don't think that branching strategy is adequate. IMO we should just do a proper *merge* of the `post-v1.0.0` branch into `master` after the release of 1.0.0.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -1120,7 +1121,7 @@ private[emitter] final class ClassEmitter(jsGen: JSGen) {
             val decodedName = className.nameString.stripSuffix("$")
             val msg = s"Initializer of $decodedName called before completion " +
               "of its super constructor"
-            val obj = js.New(encodeClassVar(UndefinedBehaviorErrorClass), Nil)
+            val obj = genScalaClassNew(UndefinedBehaviorErrorClass, NoArgConstructorName)
             val ctor = obj DOT js.Ident(genName(StringArgConstructorName))
             js.Throw(js.Apply(ctor, js.StringLiteral(msg) :: Nil))
```
sjrd: Consider moving this logic in a helper `$throwModuleInitError(decodedName)` defined once and for all by the core JS lib (only when `semantics.moduleInit` is Fatal). Then here we would just call `genCallHelper("throwModuleInitError", js.StringLiteral(decodedName))`.

This way, we do not need to require a global knowledge for `genModuleAccessor`, and hence we do not need to remove its caching in the emitter. It's better for code size too.

This change could be done in a separate commit that comes before this commit.

gzm0:
Could we use: [`--rebase-merges`][1]? My concern with merging is that the history will be unnecessarily hard to understand.

[1]: https://stackoverflow.com/questions/15915430/what-exactly-does-gits-rebase-preserve-merges-do-and-why

sjrd:
Hum I was not aware of `--rebase-merges`. Perhaps that can work, indeed.

I'm not sure it's really necessary, though. If we have *one* merge of post-v1.0.0 into master, it can't be more complicated than the many merges from 0.6.x into master that we have every now and then.

gzm0:
> I'm not sure it's really necessary, though. If we have one merge of post-v1.0.0 into master, it can't be more complicated than the many merges from 0.6.x into master that we have every now and then.

Yes, if that is the case. What if we need to merge `master` back into `post-v1.0.0` because there are changes on `master`? It just feels like it complicates history unnecessarily. The difference with the `0.6.x` branch being that we actually need the parallel history.

sjrd:
Ah, yes, merges from `master` into `post-v1.0.0` would complicate the history.

Another way to look at it is: instead of merging post-v1.0.0 into master at the end, we'll have post-v1.0.0 *become* master. Or, in other words, we'll *fast-forward* master to post-v1.0.0 after 1.0.0 is released and we've merged `master` into `post-v1.0.0` a last time.

This would produce a history where we have: 0.6.x merges into master; master merges into post-v1.0.0. So something like:
```
post-v1.0.0   master    0.6.x
    |           |         |
    *           |         |
    |\_________ |         |
    *          \|         |
    |\          *         |
    | *         |         |    a commit that only makes it post-v1.0.0
    |/          *         |
    *           |\_______ |
    |           *        \|
    |           |\        *
    |           | *       |    a commit that makes it into v1.0.0
    |           |/        |
    |           *         |
    |           |         *
    |           |         |\
    |           |         | *  a 0.6.x commit
    |           |         |/
    |           |         |
```
This is really the same thing as merge 0.6.x into master, done a second time to merge master into post-v1.0.0. When we're done, we fast-forward master to post-v1.0.0 and the old master branch is closed off.

To make it clearer, we could also keep `master` for what is now `post-v1.0.0`, and use another branch `release-1.0.0` for what is called `master` in the diagram above.

gzm0:
> To make it clearer, we could also keep master for what is now post-v1.0.0, and use another branch release-1.0.0 for what is called master in the diagram above.

Yeah, that sounds like a good idea.

So we FF `master` to `post-v1.0.0` and create `release-1.0.0` at 5fe23fab8?

sjrd:
Yes, let's do that. I'm not on a computer right now, so feel free to do it.

PR #3903, opened 2019-12-16
Global ref handling cleanups

gzm0:


gzm0:
Targeting post-v1 to be conservative.

gzm0:
> Also I believe the target branch for this PR should be post-v1.0.0, no?

yes.

> Consider also using transformGlobalVarIdent instead of referenceGlobalName+js.Ident() in def doAssign, while you're getting rid of explicit calls to referenceGlobalName.

Good point. Will update.

PR #3904, opened 2019-12-16
Move Name generation into its own class

gzm0:
There was no shared logic at all with JSGen. This gives us two
advantages:

- Keep private things more private,
- Keep name caches alive for longer.


sjrd:
I understand that you're targeting `master` for this PR? I'm fine with that, since it's only a refactoring, basically.

gzm0:
> I understand that you're targeting master for this PR? I'm fine with that, since it's only a refactoring, basically.

Unintended. Fixed.

PR #3905, opened 2019-12-17
Global ref handling cleanups

gzm0:


gzm0:
Re-open of #3903 that got closed due to me removing its target branch.

PR #3906, opened 2019-12-17
No special casing for inlineable inits

gzm0:


gzm0:
This is a re-open of #3902 which got closed because I removed its target branch.

Issue #3908, opened 2019-12-18
Bootstrap test broken in 0.6.x with Node.js 13


sjrd:
With Node.js 13.3.0 on the PATH, in 0.6.x, `toolsJS/test` fails one test, namely `StackTraceTest.decode_class_name_and_method_name`:
```
Test run started
Test org.scalajs.testsuite.library.StackTraceTest.decode_class_name_and_method_name started
Test org.scalajs.testsuite.library.StackTraceTest.decode_class_name_and_method_name failed: null, took 0.041000192 sec
Test org.scalajs.testsuite.library.StackTraceTest.decode_class_name_and_method_name finished, took 0.041000192 sec
Test run finished: 1 failed, 0 ignored, 1 total, 0.041000192s
```
This does not happen on Node.js v10, nor in 1.x (for `testSuite2_12/bootstrap:test` in that branch). Upgrading source-map-support to 0.5.7 (same as in 1.x) does not solve the issue.

My first guess would be that it's related to the fact that `toolsJS/test` executes everything in an `eval()`, whereas in 1.x `testSuite2_12/bootstrap:test` uses all the right setups to execute the bootstrapped test suite as a top-level script, just like the non-bootstrapped one.

If I can't nail it down quickly enough, I will first submit a PR to disable that test in the boostrap, to repair the CI.

gzm0:
I also propose to close this one.

PR #3909, opened 2019-12-19
[no-master] Work around #3908: Disable StackTraceTest in bootstrap.

sjrd:
With Node.js 13, the shape of stack traces reported during bootstrap are altered because the execution is within a big `eval`. Stack trace lines that were previously of the shape

    at $c_LSomeClass.someMethod__O__O (/.../file.js:15:17)

are now of the shape

    at $c_LSomeClass.eval [as someMethod__O__O] (/.../file.js:15:17)

Our `StackTrace.scala` analyzer does not understand this shape, which causes the `StackTraceTest` to fail during bootstrap.

This commit simply ignores the test during bootstrap, in order to repair the CI. Further investigation is needed to completely fix the issue.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -744,7 +744,8 @@ object Build {
           val scalaJSEnv = {
             s"""
             {"javaSystemProperties": {
-              "scalajs.scalaVersion": "${scalaVersion.value}"
+              "scalajs.scalaVersion": "${scalaVersion.value}",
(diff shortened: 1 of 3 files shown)


gzm0:
As a side note: I'm OK with leaving this test disabled in bootstrap in 0.6.x and basically admitting that stack trace detection within eval is broken. I think this is acceptable.

PR #3911, opened 2019-12-20
Remove Dummy Declarations for GCC

gzm0:
This makes GCC less special in the Emitter setup.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
@@ -28,14 +28,8 @@ import EmitterNames._
 private[emitter] final class KnowledgeGuardian(config: CommonPhaseConfig) {
   import KnowledgeGuardian._
 
-  private var firstRun: Boolean = true
-
-  private var isParentDataAccessed: Boolean = _
-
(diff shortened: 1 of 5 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -369,6 +360,88 @@ private[emitter] final class KnowledgeGuardian(config: CommonPhaseConfig) {
...
+    }
+
+    /** Call this when we invalidate all caches. */
+    def unregisterAll(): Unit = {
+      methodsInHijackedClassesAskers.clear()
+    }
+  }
+
+  private def invalidateAskers(askers: mutable.Set[Invalidatable]): Unit = {
+    /* Calling `invalidateAndUnregisterFromAll()` will cause the
```
sjrd: There seems to be a mismatch between `invalidateAndUnregisterFromAll` in the comment versus `invalidate` in the actual code. I know it was already like that before the change, but maybe use the opportunity to fix the inconsistency?

gzm0: Good catch.

gzm0:
Updated.

gzm0:
Updated Mima config.

PR #3917, opened 2020-01-07
Expose less things for collection wrappers

gzm0:


gzm0:
Seems this exposes a bunch of baggage we inherited from the collection migration (`WrappedVarArgs`). 

Also, it seems there is an implicit conversion `js.WrappedArray => js.Array` but not for `js.WrappedDictionary => js.Dictionary`.

library/src/main/scala-new-collections/scala/scalajs/js/WrappedArray.scala
```diff
@@ -21,7 +21,7 @@ import scala.scalajs.js
 
 /** Equivalent of scm.WrappedArray for js.Array */
 @inline
-final class WrappedArray[A](val array: js.Array[A])
+final class WrappedArray[A](array: js.Array[A])
```
exoego: I wonder these wrapper can be value class, which alocates less garbage...

sjrd: They can't, because they extend non-universal traits.

sjrd: Apparently we are actually using the access to the underlying `array`, for example at
https://github.com/scala-js/scala-js/blob/043912f8baca36843ea1d31853930c5cd0cfc18a/library/src/main/scala-old-collections/scala/scalajs/js/JSConverters.scala#L64

Is that legitimate? If yes, maybe this should actually stay public.

gzm0: Well, its an optimization and I'm not even 100% sure if its a valid one (w.r.t. mutability). In any case, there is also an implicit conversion back to the array. So it seems this part of the library was a bit neglected :-/

sjrd: > Well, its an optimization and I'm not even 100% sure if its a valid one (w.r.t. mutability).

We discussed this before. It is consistent with what the Scala collections do for mutable wrappers:
```scala
scala> val arr = Array("one", "two")
arr: Array[String] = Array(one, two)

scala> val seq: Seq[String] = arr
seq: Seq[String] = WrappedArray(one, two)

scala> seq.toArray
res3: Array[String] = Array(one, two)

scala> res3 eq arr
res4: Boolean = true
```

gzm0:
Updated:
- Fixed usages of `WrappedArray` in 2.13 `ArrayOps`.
- Added a commit documenting that the mutability behavior of `toJSArray` is consistent with the Scala library and `Array`.

PR #3919, opened 2020-01-09
Fix tests because erased is a new keyword in dotty

molikto:
I am trying to work on make tests of Scala.js pass in Dotty.

This test will not pass in Dotty because erased is a keyword now.

`node_modules/` is added to `.gitignore`, will remove if this is unwanted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ target/
 /.idea/
 /.idea_modules/
 bin/
+node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ target/
 /.idea/
 /.idea_modules/
 bin/
-node_modules/
+/node_modules/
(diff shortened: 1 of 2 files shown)


molikto:
Please check CLA again

gzm0:
May I suggest the following commit title: 'Avoid dotty keyword "erased"'? IMHO it is clearer and shorter.

sjrd:
It's unclear that `erased` should stay a keyword in Dotty, as the whole `erased` feature has been put under a flag in https://github.com/lampepfl/dotty/pull/7906. I have submitted an issue in Dotty at https://github.com/lampepfl/dotty/issues/7944, which now has a work-in-progress PR at https://github.com/lampepfl/dotty/pull/7947.

Let's wait to see how things develop in Dotty before changing anything in Scala.js.

PR #3920, opened 2020-01-11
Fix various tests for Dotty

molikto:
`toXxx` is because I started running them on JDK 13, these should be harmless changes.

Also Dotty removed `any2stringadd` in Prelude. So to make minimal changes to the tests it is defined as a implict value class as before


molikto:
Closed for now because I need to figure out if it is a bug of Dotty

PR #3925, opened 2020-01-16
Fix #3924: Fix Class.isAssignableFrom for boxed classes.

### LINKED ISSUE #3924: isAssignableFrom differences on 1.0.0-RC2
While trying to make shapeless work with java 1.0.0-RC2 I found a case where the JVM and JS 0.6.x behaves the same but 1.0.0-RC2 don't.
for `Character` and `Long` this operations returns `true` in 0.6 and `false` in 1.0
```
    val c: Any = 'c'
    classOf[java.lang.Character].isAssignableFrom(c.getClass)}
```
The following repo contains a bit more code
https://github.com/cquiroz/assignable

/cc @milessabin

sjrd:
There were two issues:

* `Long` and `Void` were comparing a class *data* with a classOf, always yielding `false`.
* `Character` was not tested at all, although it has been needed since cc56fe758fbd1e6505ba05494142eb89b159780e.


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
```diff
@@ -100,13 +112,26 @@ class ClassTest {
...
     assertTrue(classOf[Seq[_]].isAssignableFrom(classOf[List[_]]))
+    assertTrue(classOf[List[_]].isAssignableFrom(classOf[::[_]]))
+    assertTrue(classOf[Seq[_]].isAssignableFrom(classOf[::[_]]))
     assertTrue(classOf[Object].isAssignableFrom(classOf[Array[String]]))
     assertTrue(classOf[Array[Seq[_]]].isAssignableFrom(classOf[Array[List[_]]]))
 
     assertFalse(classOf[String].isAssignableFrom(classOf[Object]))
     assertFalse(classOf[List[_]].isAssignableFrom(classOf[Seq[_]]))
+    assertFalse(classOf[Option[_]].isAssignableFrom(classOf[::[_]]))
+    assertFalse(classOf[Set[_]].isAssignableFrom(classOf[::[_]]))
```
gzm0: I do not understand what these two new cases are adding in terms of coverage.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
```diff
@@ -100,13 +112,26 @@ class ClassTest {
...
     assertFalse(classOf[Array[List[_]]].isAssignableFrom(classOf[Array[Seq[_]]]))
+
+    for {
+      cls <- BoxedClassOfs
+      if cls != classOf[java.lang.Void]
+    } {
+      assertTrue(
+          s"classOf[Comparable[_]].isAssignableFrom($cls) should be true",
+          classOf[Comparable[_]].isAssignableFrom(cls))
+    }
```
gzm0: Maybe add a comment that boxed classes extend `Comparable` and since they are hijacked, it makes sense to test that in addition to subclassing in general?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
sjrd: They're covering cases where the left hand side and/or right hand side is a *class* (not a trait) that is not special from the linker's perspective. Instance tests for classes use `instanceof` now, so it's good to have tests for that.

sjrd:
Updated.

Issue #3921, opened 2020-01-11
Running multiple fastOptJS causes java.lang.NoClassDefFoundError: jdk/internal/reflect/MethodAccessorImpl


xerial:
Symptom:
- MethodAccessorImpl NoClassDefError happens with Scala.js 1.0.0-RC2, when compiling multiple Scala.JS projects. We have no such issue with Scala.js 0.6.x
- After rerunning fastOptJS, the compilation eventually succeeds.
- I'm using sbt 1.3.6 and JDK11. The same error happens with JDK8.

![image](https://user-images.githubusercontent.com/57538/72209440-0051f800-3463-11ea-9251-3c4bb364b8ae.png) 
- CI log: https://github.com/wvlet/airframe/runs/385008133

~It might be an issue of sbt-scalajs plugin and sbt-1.3.x's new classloader.~ (The same problem happens with sbt 1.2.8) I'll work on reproduction of this issue. 

It seems the reflection call of PathOutputFile in the LinkerImpl is failing:
```scala
    private val outputFileMethod =
      loadMethod("PathOutputFile", "atomic", classOf[LinkerOutput.File], classOf[Path])
```
This is used here:
```scala
            val out = LinkerOutput(linkerImpl.outputFile(output.toPath))
              .withSourceMap(linkerImpl.outputFile(sourceMapFile.toPath)) <--------
              .withSourceMapURI(relURI(sourceMapFile.getName))
              .withJSFileURI(relURI(output.getName))
```

A workaround is splitting an aggregated project into a smaller set of projects and running fastOptJS individually.

gzm0:
Thanks for the report.

If I read that stack trace right, the call to `invoke` on the `Method` is failing, not the lookup. Looks like a thread safety issue on first sight, but all my Googling so far says `java.reflect.Method` is thread safe :-/

xerial:
My guess is sbt 1.3.x's layered class loader is unloading some necessary classes before Scala.js compilation finishes. So trying sbt 1.2.x might be necessary to identify the issue.

sjrd:
[...]

PR #3927, opened 2020-01-16
#3921: Workaround for MethodAccessorImpl NoClassDefError

xerial:


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -75,6 +75,7 @@ object LinkerImpl {
     private val parentPrefixes = List(
       "java.",
       "scala.",
+      "sun.",
```
sjrd: We'll need a comment explaining why we have this, with a reference to the issue and possibly also the OpenJDK issue.

xerial: ok. I'll add

sjrd:
Also this PR must target the `release-1.0.0` branch. The `master` branch is now for 1.0.1 and later.

xerial:
ok. I'll retarget the branch. May need to create another PR.

xerial:
Created another PR #3928 to target release-1.0.0 branch.

PR #3928, opened 2020-01-16
Fix #3921: Work around JDK-6265952: MethodAccessorImpl NoClassDefError

### LINKED ISSUE #3921: Running multiple fastOptJS causes java.lang.NoClassDefFoundError: jdk/internal/reflect/MethodAccessorImpl

xerial:


gzm0:
Higher level ([details](https://github.com/scala-js/scala-js/issues/3921#issuecomment-575531438)): I don't think we should release 1.0.0 (or 1.0.0-RC3) with this. So a workaround is not needed as this repo does not suffer from the issue.

Lower level: The reported stack-trace in #3921 suggests that we'll also need to add `"jdk."` for some jdks.

gzm0:
Please see https://github.com/scala-js/scala-js/issues/3921#issuecomment-575627559. tl;dr: I think this workaround is fine.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -75,6 +75,18 @@ object LinkerImpl {
...
+       * This is a workaround for java.lang.NoClassDefFoundError: jdk/internal/reflect/MethodAccessorImpl,
+       * which happened when compiling multiple Scala.js projects:
+       * https://github.com/scala-js/scala-js/issues/3921
+       *
+       * The cause is java.lang.reflect.Method becomes invalid after 14 "invokes".
+       * The default threshold of -Dsun.reflect.inflationThreshold is 15:
+       * https://www.ibm.com/developerworks/community/forums/html/threadTopic?id=77777777-0000-0000-0000-000013894489&ps=25
+       *
+       * So we will delegate the access to sun.* packages to the parent class loader instead of using reflection.
+       */
```
gzm0: Please:

- Put the workaround at the end of the list.
- Adjust the comment as pointed out in https://github.com/scala-js/scala-js/issues/3921#issuecomment-575553465
- Consider only adding the prefix "sun.reflect."
- Also add "jdk.internal.reflect." (judging from the stack trace in #3921) or verify this workaround works for the airframe CI.

sjrd:
Marking "DO NOT MERGE" until #3931 is merged.

xerial:
@gzm0 Fixed the PR as suggested. And confirmed this version worked well for building Airframe's js projects all at once!

xerial:
oops. Forgot to add a trailing comma to "sun.reflect". Fixed.

xerial:
Fixed the formatting issue too (120 column limit)

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
gzm0: So on second reading, I think this comment still needs improvement. Notably, the second section doesn't "connect up" to the first one.

Things that at the moment are unclear from the comment IMHO:
1. Why is `"sun.reflect."` there as well?
2. If `java.lang.reflect.Method` becomes invalid, why does this fix it? Why do we not need to recreate a new `jlr.Method`?
3. Its non obvious that the prefixes are implementation details of the relevant JDKs (this is related to the first point). So if somebody sees a `NoClassDefFoundError: moon.reflect.MethodAccessorImpls` missing, this is probably the same issue.

I think point 1 and 3 can be addressed by making it clear that this is an implementation detail and `jdk.internal.reflect.MethodAccessorImpl` and `sun.reflect.MethodAccessorImpl` are the *currently* known affected classes.

Point 2 is a bit more tricky: I think we need to either explain the thing in full, or omit explanation of *why* this happens and simply describe

- trigger (call `invoke` `sun.reflect.inflationThreshold` times on same `Method` instance)
- symptom (`NoClassDefFoundError`)
- mitigation (add prefix here)

and refer to the bug, letting the reader reconstruct the details by themselves.

xerial:
@gzm0 Thanks for the comment. I still don't have the full context, including why this workaround can fix the issue, and how the current LinkerImpl works, so I think you are the best person for fixing the comment to address these new questions.

Due to the holiday season of US, I'll not be available for 2 or 3 days, so if you are available please fix the comment so that it can be helpful for future readers. 

I can take a look later, but please understand my knowledge around Scala.js compiler is quite limited at this moment.

gzm0:
> if you are available please fix the comment so that it can be helpful for future readers.

I have done so. Thanks for your contribution!

gzm0:
@sjrd PTAL a the comment I have edited.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -77,7 +77,36 @@ object LinkerImpl {
...
+       *
+       * It manifests as a `java.lang.NoClassDefFoundError` being thrown,
+       * claiming the class `MethodAccessorImpl` is missing. The package of the
+       * class is implementation specific. The currently known packages are
+       * listed as prefixes below.
+       *
+       * The bug is triggered when calling `java.lang.Method#invoke` if both of
+       * the following conditions are met:
+       *
+       * - this call 15th (or later) time `invoke` is called on this instance,
```
sjrd: This sentence looks broken.

Otherwise LGTM.

gzm0: :-/ yes. fixed

xerial:
@gzm0 Thanks for the fix!

PR #3934, opened 2020-01-21
Fix #3933: Use fresh names everywhere in GenJSExports.

### LINKED ISSUE #3933: 1.0.0-RC2: using arg as argument name creates IR error
I cannot pinpoint the issue here but this code:

```scala
  val sfc = SFC1[Props] { arg =>
    val cond = arg.summary.facets.headOption.map(f => convertCondition(f.label)).orUndefined
    val props = new ReactWeatherDisplayProps {
      temperature = arg.summary.temp
      currentTemperature = arg.summary.temp
      width = 150
      height = 150
      condition = cond
      currentCondition = cond
    }
    div(new DivProps {
      className = "daily"
      style = layout
    })(
      Label(labelProps)(arg.summary.dateStr),
      ReactWeatherDisplay(props)
    )
  }
```

gives an IR error

```sh
[info] Fast optimizing /home/me/proj/workspace/scalajs-reaction/examples/target/scala-2.13/Scala.js
[error] file:/home/me/proj/workspace/scalajs-reaction/examples/src/main/scala/atmoache/atmoache.scala(134:21:VarDef): Duplicate local variable name arg$1.
[error] There were 1 IR checking errors.
[error] (examples / Compile / fastOptJS) There were 1 IR checking errors.
```

If a change arg to argx it compiles fine.

sjrd:
The logic in `GenJSExports.scala` was always using compiler-generated local names that were not fresh. This was fine before, because the local scope of an exporter method was never mixed with that of user-defined code of even scalac-generated code. However, it became problematic with the introduction of Scala.js-defined classes, because their constructors precisely create such mixes.

We now use proper fresh local names everywhere when creating exporter methods.

---

We will need to merge this into `release-v1.0.0` once this PR is merged. There will be merge conflicts, but no "semantic" conflict AFAICT.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
+++ b/compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
@@ -90,9 +90,9 @@ trait GenJSExports[G <: Global with Singleton] extends SubComponent {
             (exp, ctor) <- namedExports
           } yield {
             implicit val pos = exp.pos
-            ExportedBody(List(JSAnyTpe),
-              genNamedExporterBody(ctor, genFormalArg(1).ref),
-              nme.CONSTRUCTOR.toString, pos)
+            new ExportedBody(List(JSAnyTpe),
+                argsRegistry => genNamedExporterBody(ctor, argsRegistry.genArgRef(0)),
+                nme.CONSTRUCTOR.toString, pos)
           }
 
           val ctors = normalCtors ++ namedCtors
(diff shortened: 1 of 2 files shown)


compiler/src/main/scala/org/scalajs/core/compiler/GenJSExports.scala
```diff
@@ -1058,45 +1061,79 @@ trait GenJSExports[G <: Global with Singleton] extends SubComponent {
...
     }
   }
 
-  private def genRestArgRef()(implicit pos: Position): js.Tree =
-    js.VarRef(js.Ident("arg$rest"))(jstpe.AnyType)
+  private object FormalArgsRegistry {
+    def forGetter(): FormalArgsRegistry = new FormalArgsRegistry(0, false)
+
+    def forSetter(): FormalArgsRegistry = new FormalArgsRegistry(1, false)
+  }
```
gzm0: It feels to me that these are better inlined at their respective call sites. Like that `FormalArgsRegistry` has no knowledge about setters / getters. This means in practice that destructuring of `genFormalArgs` will have to take place when generating a setter. However, IMHO this is not any less clean than asserting in `genSetterFormalArg`.

sjrd: Makes sense. Updated.

PR #3943, opened 2020-01-23
Respect sbt `repositories` when fetching the linker

smarter:
sbt has a ~/.sbt/repositories file that can be used to add extra default
resolvers, this is especially useful for setting up a cache or a proxy
(see https://www.scala-sbt.org/1.x/docs/Proxy-Repositories.html for more
information).

So far, sbt-scalajs did not respect this setting and instead manually
specified a list of resolvers, this commit fixes this. Ideally we would
just use `dependencyResolution.value` but sbt does not define this task
in the global scope.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -198,10 +198,14 @@ object ScalaJSPlugin extends AutoPlugin {
           val s = streams.value
           val log = s.log
           val retrieveDir = s.cacheDirectory / "scalajs-linker" / scalaJSVersion
+          // We can't use `dependencyResolution.value` because it's only defined in
+          // the project scope
           val lm = {
             import sbt.librarymanagement.ivy._
+            val resolvers = Classpaths.appRepositories(appConfiguration.value)
+              .getOrElse(Vector(Resolver.defaultLocal, Resolver.mavenCentral))
             val ivyConfig = InlineIvyConfiguration()
-              .withResolvers(Vector(Resolver.defaultLocal, Resolver.mavenCentral))
+              .withResolvers(resolvers)
               .withLog(log)
             IvyDependencyResolution(ivyConfig)
           }


sjrd:
That looks reasonable to me, but I'll let @gzm0 comment because he authored that code.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -198,10 +198,14 @@ object ScalaJSPlugin extends AutoPlugin {
           val s = streams.value
           val log = s.log
           val retrieveDir = s.cacheDirectory / "scalajs-linker" / scalaJSVersion
+          // We can't use `dependencyResolution.value` because it's only defined in
+          // the project scope
           val lm = {
             import sbt.librarymanagement.ivy._
+            val resolvers = Classpaths.appRepositories(appConfiguration.value)
+              .getOrElse(Vector(Resolver.defaultLocal, Resolver.mavenCentral))
```
sjrd: I assume this is more or less copied from some code in sbt/sbt? If yes, could you add a comment pointing to the origin of the code, so that we can refer to it in the future?

gzm0: It seems to be a combination of
- [Defaults.scala][1]
- [ResolverExtra.scala][2]

[1]: https://github.com/sbt/sbt/blob/v1.0.0/main/src/main/scala/sbt/Defaults.scala#L1776-L1804
[2]: https://github.com/sbt/librarymanagement/blob/v1.0.0/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L191-L212

In any case, it does not seem to reproduce the exact sbt behavior :-/

smarter: Yeah, I used the minimum amount of code needed to get the repositories file taken into account. If it makes you more comfortable, I could instead copy-paste the `appResolvers` task as is: https://github.com/sbt/sbt/blob/816958d8c84a0f82968dbfb8f7f1daa85c3cc7ba/main/src/main/scala/sbt/Defaults.scala#L1784-L1791, or copy-paste `fullResolvers` and its dependencies, but remove the pars which are project-dependent.

sjrd:
I have an alternative at #3947.

smarter:
Your alternative LGTM, feel free to close this PR.

gzm0:
Closing in favor of #3947.

PR #3945, opened 2020-01-24
Remove remaining `private[linker]` from interface

gzm0:
They should not have been there in the first place anymore.


sjrd:
> An alternative would have been to leave the case classes / objects in
Semantics.scala and only move `apply` / `map` to
`RuntimeClassNameMapperImpl`. However, this would have separated
structural representation and mapping unnecessarily.

Have you considered leaving everything in `Semantics.scala` as it is now, and only add
```scala
object RuntimeClassNameMapperImpl {
  def map(mapper: RuntimeClassNameMapper, className: String): String =
    mapper(className)
}
```
(possibly renaming `apply` to `map` or `mapInternal`)

?

gzm0:
> Have you considered leaving everything in Semantics.scala as it is now, and only add

No :) It's much simpler indeed. Updated.

PR #3946, opened 2020-01-24
Store the scalaJSLinkerImpl in a CacheBox.

sjrd:
This ensures that we reuse the same class loader between several sbt commands. Since we do cache instances of classes loaded through the `LinkerImpl`'s class loader across commands, we can leak class loaders if we do not also cache the class loader itself.


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/CacheBox.scala
```diff
@@ -17,18 +17,20 @@ package org.scalajs.sbtplugin
  *  A CacheBox is needed, once the cached result needs to depend on a task,
  *  since then it cannot simply be made a setting anymore.
  */
-private[sbtplugin] final class CacheBox[T] {
+final class CacheBox[T] {
```
gzm0: Can we make only the type publicly accessible (not it's methods, constructor, etc.)?

project/Build.scala
```diff
@@ -101,6 +101,13 @@ object MyScalaJSPlugin extends AutoPlugin {
           NodeJSEnv.Config().withSourceMap(wantSourceMaps.value)),
 
       scalaJSLinkerImpl := {
+        /* !!! This custom implementation does *not* respect the contract of
+         * scalaJSLinkerImpl, because it does not cache the `LinkerImpl` in
+         * scalaJSLinkerImplBox. The reason is that we need a new class loader
+         * to be created every time, in case the linker sources have changed.
```
gzm0: This is not really true: We store the linker *instance* anyways so we need a reload if we want a new linker. So IIUC, there is no reason to not cache the linker impl as well.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
```diff
@@ -194,21 +198,26 @@ object ScalaJSPlugin extends AutoPlugin {
...
+              import sbt.librarymanagement.ivy._
+              val ivyConfig = InlineIvyConfiguration()
+                .withResolvers(Vector(Resolver.defaultLocal, Resolver.mavenCentral))
+                .withLog(log)
+              IvyDependencyResolution(ivyConfig)
+            }
+            lm.retrieve(
+                "org.scala-js" % "scalajs-linker_2.12" % scalaJSVersion,
+                scalaModuleInfo = None, retrieveDir, log)
+              .fold(w => throw w.resolveException, LinkerImpl.default _)
```
gzm0: Actually, have you considered caching this based on the classpath we pass in? That might solve the whole boxing issue.

project/Build.scala
sjrd: Right, I guess that's true, indeed.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/CacheBox.scala
sjrd: Not if we keep the contract that custom `scalaJSLinkerImpl`s must call `ensure` on `scalaJSLinkerImplCacheBox.value`...

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
sjrd: I don't understand how that would help. We would still need to store the cached `LinkerImpl` somewhere. Where do we cache it then? In a `Map[Seq[File], LinkerImpl]`? Where do we store that map? When do we remove entries from that map?

I guess I didn't understand what you are suggesting. Could you explain further?

gzm0: I meant something like `private var defaultLinkerImpl: Option[(Seq[File], LinkerImpl)] = None`. In *our* implementation we check against this cache.

This on its own doesn't solve the problem for clients yet. Maybe for now, we should make `LinkerImpl.default` private and expect clients that want a different linker to just depend on it directly?

Thinking of it, we should also probably do this instead of having the key `scalaJSGlobalIRCacheBox`; because it does not need to be scoped (unlike `scalaJSLinkerCacheBox`).

gzm0:
I have had another idea:

- Define `fullClasspath in scalaJSLinkerImpl` by default as the fetched JAR.
- Define `scalaJSLinkerImpl` to load the linker via reflection from `fullClasspath in scalaJSLinkerImpl` (with boxing).
- Make `LinkerImpl.default` private to the sbt plugin.

Users that want to load the linker via reflection (including the SJS build) can set `fullClasspath in scalaJSLinkerImpl` and have correct class loader caching behavior.

Users that want to load the linker directly, can simply set `scalaJSLinkerImpl` to whatever they want.

sjrd:
The `fullClasspath in scalaJSLinkerImpl` is definitely a good idea. I have pushed a commit that does that.

However, I do not think that either that or your earlier comments reduces the need for `scalaJSLinkerImplBox`, nor for exposing `CacheBox`. In all cases, people who want to directly customize `scalaJSLinkerImpl` *should* return a cached instance. Then we have two choices:

* either we expose `CacheBox` and `scalaJSLinkerImplBox`, with the proper documentation that they should use it, which allows to *easily* write the *correct* thing to do,
* or we don't, and we still have to document that they must cache it if they customize it, but we don't give them an easy tool to do it correctly, so they are more likely to do it wrong.

Of those two options, I firmly believe the former is best.

gzm0:
Hmmm, that argument is fair. My hope was that we could hide the logic to load the linker via reflection (`LinkerImpl.reflect`) and only expose it via our default implementation.

Somebody wanting to customize things could then do:

```scala
scalaJSLinkerImpl ~= { prev =>
  new LinkerImpl.Forwarding(prev) {
    // override whatever, use prev if desired
  }
}
```

Now this will of course go terribly wrong if they try to put this somehow into `scalaJSLinkerImplBox`.

I have also been thinking about the use case where an implementation is directly loaded into the build. Then this would also be OK IIUC:

```scala
scalaJSLinkerImpl := new LinkerImpl {
  // definitions
}
```

IIUC the caching is only a problem if the values returned from `LinkerImpl` hold on to a reference of their creator (which in general is not the case for factories AFAICT).

Maybe we should focus on use cases we already know about (scalajs-bundler?). How would we want something like https://github.com/scalacenter/scalajs-bundler/pull/304 to use this?

sjrd:
If we take scalajs-bundler as a use case, it would need to:

* Declare its own classes in an independent artifact that is published
* Customize `fullClasspath in scalaJSLinkerImpl` to add that artifact to the class loader
* Then also customize `scalaJSLinkerImpl` to use the default `reflect`, but override its `clearableLinker` method to return its own thing.

To do the latter, it needs access to the `ClassLoader` used by `reflect`, which it currently doesn't even have access to at the moment. So in fact it would need to reimplement all of `LinkerImpl.reflect`, including the `FilteringClassLoader` and all that stuff.

Perhaps `LinkerImpl.reflect()` should return a `LinkerImpl.Reflect` (which would public) and `Reflect` should actually *expose* its `ClassLoader`. Then scalajs-bundler would be able to write:
```scala
scalaJSLinkerImpl := {
  val linkerImplClasspath = (fullClasspath in scalaJSLinkerImpl).value
  scalaJSLinkerImplBox.value.ensure {
    val base = LinkerImpl.reflect(Attributed.data(linkerImplClasspath))
    new CustomLinkerImpl(base)
  }
}

private class CustomLinkerImpl(base: LinkerImpl.Reflect)
    extends LinkerImpl.Forwarding(base) {
  val customClearableLinkerMethod = base.classLoader.loadClass(...)...

  override def clearableLinker(config: StandardLinkerConfig): ClearableLinker =
    ... customClearableLinkerMethod.invoke(...)
}
```

gzm0:
But doesn't scalajs-bundler also have to inject its own config (the file where it writes the entry points to). How is it going to do that? (It seems this is a problem with the overall approach, not even with this PR in particular) :-/

sjrd:
I would keep them in memory, associated with the custom instance of `ClearableLinker`, and access that from a downstream task.

But if we need a different type of config, then obviously we would need to customize `scalaJSLinker` itself, instead of just `scalaJSLinkerImpl`. That's unavoidable with any design where we have a separate key for `scalaJSLinkerConfig` of a well-known, concrete type (which we should keep, IMO).

sjrd:
I have added a commit that exposes `LinkerImpl.Reflect` and its `loader` field, then used that to add a scripted test that reproduces the requirements of scalajs-bundler.

gzm0:
OK. It seems that this is indeed the best we can do with the current API. However, I believe it exposes a bit too much of the sbt plugin implementation. Notably, I think we should provide a better API going forward, for which we'll probably need to adjust some of these internals.

In the interest of unblocking 1.0.0, I propose the following:
- We merge this PR in its current form (modulo detailed review).
- We add comments to the following saying we might break compatibility going forward:
  - `CacheBox`
  - `LinkerImpl` (and subclasses)
  - `scalaJSLinkerImpl`
  - all keys of type `CacheBox[_]`

This is similar to the super stable and not-so-stable API we offer for the linker itself: 
- If you just use the sbt plugin, you get a stable API
- If you want to customize the linker, you get a less stable API, but you are already OK with that (because you use the less stable linker API).

This will allow us to:
- Fix the classloader leaking
- Support scalajs-bundler (somewhat)
- Improve the situation in the future.

WDYT?

sjrd:
Yes, that sounds reasonable.

sbt-plugin/src/sbt-test/linker/custom-linker/build.sbt
```diff
@@ -0,0 +1,43 @@
...
+lazy val main = project
+  .enablePlugins(CustomScalaJSLinkerPlugin)
+  .settings(
+    scalaJSUseMainModuleInitializer := true,
+  )
+
+check := {
+  val modules = (scalaJSImportedModules in (main, Compile, fastOptJS)).value
+  val expected = Set("foo.js", "bar.js")
+  assert(modules.size == expected.size && modules.toSet == expected,
```
gzm0: Why the check for size? Two sets that are do not have the same size will not be equal anyways...

sbt-plugin/src/sbt-test/linker/custom-linker/custom-linker/src/main/scala/customlinker/EntryPointAnalyzerBackend.scala
```diff
@@ -0,0 +1,56 @@
...
+      loadSpec match {
+        case Import(module, _)                              => List(module)
+        case ImportWithGlobalFallback(Import(module, _), _) => List(module)
+        case Global(_, _)                                   => Nil
+      }
+    }
+
+    linkingUnit.classDefs
+      .flatMap(_.jsNativeLoadSpec)
+      .flatMap(importedModulesOf(_))
```
gzm0: Don't we use `(importedModulesOf _)` usually?

sbt-plugin/src/sbt-test/linker/custom-linker/build.sbt
sjrd: To make sure there are no duplicates in `modules` (which is a `List`).

sbt-plugin/src/sbt-test/linker/custom-linker/custom-linker/src/main/scala/customlinker/EntryPointAnalyzerBackend.scala
sjrd: We never settled. I tend to use `importedModulesOf(_)` whereas you tend to use `importedModulesOf _`. ;)

sjrd:
Updated.

I think the separation of the commits is meaningful in this case, so I would not squash anything.

PR #3947, opened 2020-01-24
Fix #3948: Use sbt's `dependencyResolution` for the LinkerImpl.

### LINKED ISSUE #3948: Fetching the Scala.js linker does not respect custom resolvers
See #3943.

sjrd:
Based on #3946: only the last commit belongs to this PR. Alternative to #3943.

Instead of always creating a default `IvyDependencyResolution` to resolve the scalajs-linker artifact, used to create the `LinkerImpl`, we now use sbt's built-in `dependencyResolution` task.

However, since that task is only available per project (and not in the Global scope), we fetch it from the `LocalRootProject`. In the rare case that the root project disabled the built-in `IvyPlugin`, we fall back to the default `IvyDependencyResolution` as before.

With these changes, resolution of the scalajs-linker artifact will follow custom configurations, such as the `~/.sbt/repositories` file.

/cc @smarter

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -194,17 +194,44 @@ object ScalaJSPlugin extends AutoPlugin {
 
         scalaJSLinkerConfig := StandardConfig(),
 
-        scalaJSLinkerImpl := {
-          val s = streams.value
-          val log = s.log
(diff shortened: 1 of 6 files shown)


sjrd:
I tested locally that the warning is not emitted for regular projects, but that it is emitted for the added scripted test.

gzm0:
I have created #3948 to track the issue. Please reference in the commit message.

sjrd:
Updated with a reference to the fix, and rebased directly on `release-v1.0.0` without depending on #3946, so that this PR becomes mergeable.

PR #3956, opened 2020-02-10
Fix #3950: Generate static forwarders for static nested objects

### LINKED ISSUE #3950: Static forwarders are not generated for static nested objects
This is a problem for implementations of Java libraries with nested statics:

```scala
object Locale {
  final class Category private (name: String, ordinal: Int) extends Enum[Category](name, ordinal)

  object Category {
    def DISPLAY: Category = new Category("DISPLAY", 0)
  }
}
```

gzm0:
We use the module initializer infrastructure to test this.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -987,17 +987,17 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
      * `def hashCode(): Int` and a `static def hashCode(Int): Int`. The JVM
      * back-end considers them as colliding because they have the same name,
      * but we must not.
+     *
+     * Further, unlike scalac, we emit forwarders for *any* static object, not
+     * just top-level ones. This is important so we can implement static methods
+     * of nested static classes of JDK APIs (see #3950).
      */
 
     /** Is the given Scala class, interface or module class a candidate for
      *  static forwarders?
      */
-    def isCandidateForForwarders(sym: Symbol): Boolean = {
-      // it must be a top level class (name contains no $s)
-      !settings.noForwarders && exitingPickler {
-        !sym.name.containsChar('$') && !sym.isNestedClass
-      }
-    }
+    def isCandidateForForwarders(sym: Symbol): Boolean =
+      !settings.noForwarders && sym.isStatic && !isImplClass(sym)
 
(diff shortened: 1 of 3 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -987,17 +987,17 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
      *  static forwarders?
      */
-    def isCandidateForForwarders(sym: Symbol): Boolean = {
-      // it must be a top level class (name contains no $s)
-      !settings.noForwarders && exitingPickler {
-        !sym.name.containsChar('$') && !sym.isNestedClass
-      }
-    }
+    def isCandidateForForwarders(sym: Symbol): Boolean =
+      !settings.noForwarders && sym.isStatic
```
sjrd: We've been using `isStaticModule(sym)` instead of `sym.isStatic` in GenJSCode. I do not remember why `isStaticModule` was needed, but it seems to me that the same condition should be applied in this case.

gzm0: This is tracable back to d217c7d7a084ddc706ad564d799233c493410094.

IIUC we cannot use `isStaticModule`, because this also needs to work for static classes.

sjrd: Oh wow that's old 😄 You're right about the classes, though. Your latest version looks good.

project/TestSuiteLinkerOptions.scala
```diff
@@ -29,7 +29,9 @@ object TestSuiteLinkerOptions {
     List(
         ModuleInitializer.mainMethod(module, "mainNoArgs"),
         ModuleInitializer.mainMethodWithArgs(module, "mainWithArgs"),
-        ModuleInitializer.mainMethodWithArgs(module, "mainWithArgs", List("foo", "bar"))
+        ModuleInitializer.mainMethodWithArgs(module, "mainWithArgs", List("foo", "bar")),
+        ModuleInitializer.mainMethod(module + "$NoLinkedClass", "main"),
+        ModuleInitializer.mainMethod(module + "$WithLinkedClass", "main"),
```
sjrd: There's a trailing comma here that doesn't compile on 2.11.

Otherwise LGTM.

PR #3960, opened 2020-02-12
Fix #3953: mistyping when using PartialFunction with HKT alias

### LINKED ISSUE #3953: fastOptJS error in scalaz 7.3 with Scala.js 1.0.0
- https://github.com/scalaz/scalaz/commit/382bb66e804ef0ec56ff4a4eb3508faebb916d36
- https://travis-ci.org/scalaz/scalaz/jobs/647661630#L513
- https://github.com/scalaz/scalaz/blob/382bb66e804ef0ec56ff4a4eb3508faebb916d36/tests/src/test/scala/scalaz/StreamTTest.scala#L160-L162

```
[error] https://raw.githubusercontent.com/scalaz/scalaz/382bb66e804ef0ec56ff4a4eb3508faebb916d36/tests/src/test/scala/scalaz/StreamTTest.scala(160:33:MethodDef): The signature of scalaz.StreamTTest$$anonfun$$nestedInanonfun$new$111$1.applyOrElse(scalaz.StreamT$Step,scala.Function1)java.lang.Object, which is (List(AnyType, ClassType(ClassName<scala.Function1>)),AnyType), does not match its name (should be (List(ClassType(ClassName<scalaz.StreamT$Step>), ClassType(ClassName<scala.Function1>)),AnyType)).
[error] There were 1 IR checking errors.
[error] (testsJS / Test / fastOptJS) There were 1 IR checking errors.
```

errikos:
The issue occurs due to a bug in scalac, where Defdef vparamss and the
respective symbol have inconsistent types. For more information, please
see: https://github.com/scala/bug/issues/11884

This commit is a workaround for the aforementioned issue, until the latter
is fixed in the scala compiler.

It consists of a very dirty hack, where, after the IR Tree generation, we go
and patch the types of the locals to correspond to the types found in the
`Defdef`'s symbol (which are the correct ones).

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -1843,7 +1843,22 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
               }
             }
 
-            Some(methodDefWithoutUselessVars)
+            /* See: https://github.com/scala-js/scala-js/issues/3953 */
+            val methodDefWithPatchedTypes = {
+              val symParamTypes =
+                if (sym.paramss.isEmpty) Nil
+                else sym.paramss.head.map(p => toIRType(p.tpe))
+
+              assert(symParamTypes.length == params.length,
+                  "symParams and vparams have different lengths")
+
+              val patchedParamTypes = collection.mutable.Map(
+                  (params.map(encodeLocalSym(_).name) zip symParamTypes): _*)
+
+              patchTypesOfLocals(methodDefWithoutUselessVars, patchedParamTypes)
+            }
+
+            Some(methodDefWithPatchedTypes)
           }
         }
       }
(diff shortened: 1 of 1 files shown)


sjrd:
You should definitely add the reproduction from the issue as a test case. ;)

gzm0:
Have we assessed the feasibility of fixing this in scalac proper? I'm not a fan of simply postprocessing practically everything we generate.

If this is really necessary, could we at least investigate a deeper integration that correctly translates an `Ident` from the start ([in this `case`][1]).

NB to reply to @errikos from #3960 

> Plus, the easiest way to do this is to patch the IR Tree after it has been generated, in order to avoid touching the generation process itself, which would probably introduce far more code changes.

IMHO this is the wrong way to look at it. We should not code in a way that it easy, but in a way that yields the easiest to understand and maintain result. The difficulty of the code and the review are secondary. A good end result is the primary goal.

From the little I can see, we can create a `Map[Symbol, Symbol]`, put it in a scoped variable and use it when we translate an `Ident`. Wouldn't that work and be simpler?

[1]: https://github.com/scala-js/scala-js/blob/v1.0.0/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala#L2295

errikos:
> IMHO this is the wrong way to look at it. We should not code in a way that it easy, but in a way that yields the easiest to understand and maintain result. The difficulty of the code and the review are secondary. A good end result is the primary goal.

Thanks for the feedback! I totally agree that this is far from the ideal way. That's why I said "simple" instead of "better". But, it is the simplest one, since it only requires patching at a single code position and this is what I was going after, given that my familiarity with the Scala.js codebase is minimal.

> From the little I can see, we can create a `Map[Symbol, Symbol]`, put it in a scoped variable and use it when we translate an `Ident`. Wouldn't that work and be simpler?

I will look into this and see if it leads into a more elegant solution. As @sjrd said, ideally we would like that the bug is fixed in scalac.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1843,7 +1843,22 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
               }
             }
 
-            Some(methodDefWithoutUselessVars)
+            /* See: https://github.com/scala-js/scala-js/issues/3953 */
+            val methodDefWithPatchedTypes = {
+              val symParamTypes =
+                if (sym.paramss.isEmpty) Nil
+                else sym.paramss.head.map(p => toIRType(p.tpe))
```
sjrd: I believe you can just use `sym.tpe.params.map(p => toIRType(p.tpe))`.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1906,6 +1921,38 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
           newBody)(methodDef.optimizerHints, None)(methodDef.pos)
     }
 
+    private def patchTypesOfLocals(methodDef: js.MethodDef,
+        patches: mutable.Map[LocalName, ir.Types.Type]) = {
```
sjrd: This should be an immutable `Map`. It's never mutated.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1843,7 +1843,22 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+            val methodDefWithPatchedTypes = {
+              val symParamTypes =
+                if (sym.paramss.isEmpty) Nil
+                else sym.paramss.head.map(p => toIRType(p.tpe))
+
+              assert(symParamTypes.length == params.length,
+                  "symParams and vparams have different lengths")
+
+              val patchedParamTypes = collection.mutable.Map(
+                  (params.map(encodeLocalSym(_).name) zip symParamTypes): _*)
```
sjrd: You can use `methodDefWithoutUselessVars.args` instead of `params` here. That's better because we already have their `LocalName`, hence we don't need to re-encode the local syms.

Moreover, and more importantly, we should filter out the parameters for which the IR type is already correct (i.e., the `(arg: js.ParamDef).ptpe == symParamType)`. If the resulting list is empty, which is the overwhelmingly common case, we should not call `patchTypesOfLocals` at all!

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: Also, it's usually better to explicitly specify the result type of a function that is a bit long, even when it is private. It helps the reader.

sjrd:
Superseded by #4533.

PR #3961, opened 2020-02-14
Set jsEnvInput without configuration in project scope

gzm0:
This allows users to simply write

    jsEnvInput += Input.Script(myJSLibrary)

in their build.

Discovered while writing the tutorial:
https://github.com/scala-js/scala-js-website/pull/478#discussion_r379045183


sjrd:
The CI failures are genuine. At first they didn't make sense to me. It turns out that, because of the sbt inherits configurations, `Test / jsEnvInput` actually defaults to `Compile / jsEnvInput`, so this change cause `Test / jsEnvInput` to also depend to `Compile / jsEnvInput`! To avoid that, we need to add the following (above the `jsEnvInput +=` setting):
```scala
      /* Do not inherit jsEnvInput from the parent configuration.
       * Instead, always derive it straight from the Zero configuration scope.
       */
      jsEnvInput := (jsEnvInput in (This, Zero, This)).value,
```
We have precedent with `scalaJSModuleInitializers`.

gzm0:
Should wait for #3964.

PR #3962, opened 2020-02-14
Remove the override of `Throwable` in the minilib.

sjrd:
The only blocker was calling `Utils.isUndefined`, which we manually inline instead.

The minilib is now free of any override. It is strictly a subset of the standard java lib. Therefore, it does not need its own project anymore, and is turned into just a custom task of the library.


project/Build.scala
```diff
@@ -773,7 +775,7 @@ object Build {
 
   lazy val linker: MultiScalaProject = MultiScalaProject(
       id = "linker", base = file("linker/jvm")
-  ).zippedSettings("minilib", "library")(
```
gzm0: I believe this overload of `zippedSettings` is now unused and can be removed.

project/Build.scala
```diff
@@ -1354,35 +1356,20 @@ object Build {
...
-      scalaJSExternalCompileSettings,
-  ).zippedSettings(library)(library =>
-      inConfig(Compile)(Seq(
-          mappings in packageBin := {
-            val superMappings = (mappings in packageBin).value
-            val libraryMappings = (mappings in (library, packageBin)).value
+          },
 
-            val whitelisted = libraryMappings.filter { mapping =>
+          packageMinilib := {
```
gzm0: Have you considered using `Defaults.packageTaskSettings`? (I'm not sure it'll turn out simpler, just wondering if you have looked at it).

sjrd: I have looked at it, and it gets awfully more complicated ;)

project/Build.scala
sjrd: Good catch! Removed.

project/Build.scala
gzm0: That's was my hunch as well :)

Issue #2681, opened 2016-12-09
Shared modules/progressive apps support


scf37:
Currently ScalaJS is not that suitable for mobile development because of monolithic `xxx-opt.js`.
While loading hundreds Ks of javascript on typing site URL is perfectly appropriate for many situations (B2B, desktop), mobile users won't be such as happy given their slow connections.

Google Closure Compiler allows splitting codebase into multiple modules  via `--module` and `--js` arguments:
http://stackoverflow.com/questions/10395810/how-do-i-split-my-javascript-into-modules-using-googles-closure-compiler/10401030#10401030

It could be really great if ScalaJS allow to split generated javascript to multiple modules to avoid waiting for too long for first page to show. For example:
- common (scala.* and libraries)
- landing page
- main page
- settings
- etc etc etc.

Splitting could be done by packages, by subprojects, by deps etc.

sjrd:
I have the feeling that we already have this idea filed somewhere, but I can't find it. Does someone remember where we discussed this before?

glmars:
This one: #1787 ?

sjrd:
@scf37 Some part of that discussion was relevant, indeed, but not all of it. The present issue is a neat summary of one particular core feature, so I'll keep it. The other features necessary for #1787 turned out to be already implemented. So I closed #1787 in favor of the present issue + these other implemented features.

ramnivas:
One more concern (besides the desired quick loading of the first page) is that using Scala.js commonjs support with tools such as Webpack (which we use currently) is basically unusable during development. When implemented with typical settings (enable source maps, for example), each time (even with a tiny change), our development builds take 40-45 seconds (3-5 second for the scala.js part and the rest for the Webpack part). This is mostly due to Webpack parsing a huge -fastopt.js for every change. We got away, for now,  using the noParse option set for -fastopt.js, but that makes us lose the sourcemaps support. We could try scalajs-bundler, but I highly doubt if it is going to do any better, since at the end of the day, it too runs Webpack.

These tools are optimized for the situations where only few small source files (js, css, etc) change at a time during development. Scala.js in its current form doesn't meet this assumption (it spits out a large "source file").

This feature could the use of promote import/export of many small modules within an app. Then each change should touch only the relevant -fastopt.js file and lower the parsing load on Webpack and other bundler tools.

cuzfrog:
Agreed. @ramnivas  scalajs-bundler seems like a workaround, not a solution. 

[...]

PR #3965, opened 2020-02-18
DO NOT MERGE RFC Interface for module splitting

gzm0:
If you come here as an sbt user, in this design, `ModuleOutputSpec` is what you'd have to provide in your build (`ModuleLinkerOutput` would probably be calculated automatically).

### DIFF
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
@@ -23,8 +23,21 @@ import org.scalajs.logging.Logger
  *  .js file.
  */
 abstract class Linker private[interface] () {
-  def link(irFiles: Seq[IRFile],
+  final def link(irFiles: Seq[IRFile],
       moduleInitializers: Seq[ModuleInitializer],
       output: LinkerOutput, logger: Logger)(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    val module = ModuleOutputSpec.Module("core")
+      .withInitializers(moduleInitializers: _*)
+
+    val spec = ModuleOutputSpec()
+      .addModule(ModuleOutputSpec.Selector.default, module)
+
+    val moduleOutput = ModuleLinkerOutput(module.id, output)
+
+    link(irFiles, spec, Seq(moduleOutput), logger)
+  }
+
+  def link(irFiles: Seq[IRFile], moduleSpec: ModuleOutputSpec, outputs: Seq[ModuleLinkerOutput], logger: Logger)(
       implicit ec: ExecutionContext): Future[Unit]
 }
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
@@ -1,0 +1,54 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+/** A single Module output specification for a linker run.
+ *
+ *  The fields [[moduleID]] and [[isInternal]] uniquely identify this output in
+ *  a [[Linker]] run.
+ *
+ *  @param moduleID The ID of the module this linker output represents.
+ *
+ *  @param isInternal Whether this is an internal (or an external) module.
+ *
+ *  @param output The [[LinkerOutput]] for this module.
+ *
+ *  @param moduleName Name of this module for import. A [[Linker]] will use
+ *      this name to include this module from one of its dependees. Must be set,
+ *      if this module is internal.
+ */
+final class ModuleLinkerOutput private (
+    val moduleID: ModuleOutputSpec.ModuleID,
+    val isInternal: Boolean,
+    val output: LinkerOutput,
+    val moduleName: Option[String]
+) {
+  private def this(moduleID: ModuleOutputSpec.ModuleID, output: LinkerOutput) =
+    this(moduleID, false, output, None)
+
+  def withInternal: ModuleLinkerOutput =
+    copy(isInternal = true)
+
+  def withModuleName(moduleName: String): ModuleLinkerOutput =
+    copy(moduleName = Some(moduleName))
+
+  private def copy(isInternal: Boolean = isInternal,
+      moduleName: Option[String] = moduleName): ModuleLinkerOutput = {
+    new ModuleLinkerOutput(moduleID, isInternal, output, moduleName)
+  }
+}
+
+object ModuleLinkerOutput {
+  def apply(moduleID: ModuleOutputSpec.ModuleID, output: LinkerOutput) =
+    new ModuleLinkerOutput(moduleID, output)
+}
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
@@ -1,0 +1,68 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+/** Module output specification. */
+final class ModuleOutputSpec private (modules: Seq[(ModuleOutputSpec.Selector, ModuleOutputSpec.Module)]) {
+  private def this() = this(Nil)
+
+  def addModule(selector: ModuleOutputSpec.Selector, module: ModuleOutputSpec.Module): ModuleOutputSpec = {
+    // TODO ensure selectors don't conflict.
+    ???
+    new ModuleOutputSpec(modules :+ (selector, module))
+  }
+}
+
+object ModuleOutputSpec {
+  def apply(): ModuleOutputSpec = new ModuleOutputSpec()
+
+  final class ModuleID private[ModuleOutputSpec] (val id: String) extends AnyVal
+
+  final class Module private (
+      val id: ModuleID,
+      initializers: Seq[ModuleInitializer],
+      deps: Seq[ModuleID]
+  ) {
+    private def this(id: ModuleID) = this(id, Nil, Nil)
+
+    def withInitializers(initializers: ModuleInitializer*): Module =
+      copy(initializers = initializers)
+
+    def withDeps(deps: ModuleID*): Module =
+      copy(deps = deps)
+
+    private def copy(
+        initializers: Seq[ModuleInitializer] = initializers,
+        deps: Seq[ModuleID] = deps): Module = {
+      new Module(id, initializers, deps)
+    }
+  }
+
+  object Module {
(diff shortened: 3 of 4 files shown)


linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
```diff
@@ -23,8 +23,21 @@ import org.scalajs.logging.Logger
...
       moduleInitializers: Seq[ModuleInitializer],
       output: LinkerOutput, logger: Logger)(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    val module = ModuleOutputSpec.Module("core")
+      .withInitializers(moduleInitializers: _*)
+
+    val spec = ModuleOutputSpec()
+      .addModule(ModuleOutputSpec.Selector.default, module)
+
+    val moduleOutput = ModuleLinkerOutput(module.id, output)
```
gzm0: Note: By implicitly defining that an external module is enough, if it does not have any dependencies, we can reduce the base case to a module case.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
```diff
@@ -23,8 +23,21 @@ import org.scalajs.logging.Logger
...
+
+    link(irFiles, spec, Seq(moduleOutput), logger)
+  }
+
+  def link(irFiles: Seq[IRFile], moduleSpec: ModuleOutputSpec, outputs: Seq[ModuleLinkerOutput], logger: Logger)(
```
gzm0: Note: The reason `moduleSpec` here is separate from `outputs` is that I anticipate most linker usages to be in a build system. In such a system, `moduleSpec` is provided by the user, but `outputs` is provided by the build system (probably dependent on `moduleSpec`).

This is less nice, because it will cause runtime failures, if the outputs are not in sync with what is in the `moduleSpec`.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
```diff
@@ -0,0 +1,54 @@
...
+ *
+ *  @param isInternal Whether this is an internal (or an external) module.
+ *
+ *  @param output The [[LinkerOutput]] for this module.
+ *
+ *  @param moduleName Name of this module for import. A [[Linker]] will use
+ *      this name to include this module from one of its dependees. Must be set,
+ *      if this module is internal.
+ */
+final class ModuleLinkerOutput private (
```
gzm0: In another design, I have merged these fields into `LinkerOuput`. I'm not 100% sure what is better.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+/** Module output specification. */
+final class ModuleOutputSpec private (modules: Seq[(ModuleOutputSpec.Selector, ModuleOutputSpec.Module)]) {
```
gzm0: We need to finely balance what is accessible in this interface. The use case of a build tool would require access to the list of modules to prepare the relevant output files.

However, `deps` and `initializers` should probably not be accessible. (if we are to make `deps` accessible, we should make the verified and potentially inferred deps accessible).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+      val id: ModuleID,
+      initializers: Seq[ModuleInitializer],
+      deps: Seq[ModuleID]
+  ) {
+    private def this(id: ModuleID) = this(id, Nil, Nil)
+
+    def withInitializers(initializers: ModuleInitializer*): Module =
+      copy(initializers = initializers)
+
+    def withDeps(deps: ModuleID*): Module =
```
gzm0: The idea here is that in a future version we can add `withAutoDeps` if this is a feature that is needed.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+/** Module output specification. */
+final class ModuleOutputSpec private (modules: Seq[(ModuleOutputSpec.Selector, ModuleOutputSpec.Module)]) {
+  private def this() = this(Nil)
+
+  def addModule(selector: ModuleOutputSpec.Selector, module: ModuleOutputSpec.Module): ModuleOutputSpec = {
```
gzm0: The idea here is that in a future version we can add: `addAutoClassifyModule(selector)` if this feature is needed.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
```diff
@@ -0,0 +1,54 @@
...
+ *
+ *  @param moduleID The ID of the module this linker output represents.
+ *
+ *  @param isInternal Whether this is an internal (or an external) module.
+ *
+ *  @param output The [[LinkerOutput]] for this module.
+ *
+ *  @param moduleName Name of this module for import. A [[Linker]] will use
+ *      this name to include this module from one of its dependees. Must be set,
+ *      if this module is internal.
```
sjrd: I do not really understand what `moduleName` does. Is this basically the relative path of this module as seen from the modules that will depend on it? I guess that assumes that all modules are in the same directory?

What is the meaning of `moduleName` for an external module?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
```diff
@@ -0,0 +1,54 @@
...
+final class ModuleLinkerOutput private (
+    val moduleID: ModuleOutputSpec.ModuleID,
+    val isInternal: Boolean,
+    val output: LinkerOutput,
+    val moduleName: Option[String]
+) {
+  private def this(moduleID: ModuleOutputSpec.ModuleID, output: LinkerOutput) =
+    this(moduleID, false, output, None)
+
+  def withInternal: ModuleLinkerOutput =
```
sjrd: If an internal module must have a module name, perhaps `withInternal` should take `moduleName` as parameter?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleOutputSpec.scala
sjrd: If dependencies must/can be explicitly specified, what happens if they contradict the reality? For example, if I forget to declare a dependency from A to B, but in practice a class in A references a class in B?

If we have to detect those issues, why can't we *always* detect dependencies, making this explicitly configuration redundant?

gzm0: > How does the caller of withDeps create ModuleIDs? It seems there is no public way to construct one.

`Module#apply` creates one from a string.

> If dependencies must/can be explicitly specified, what happens if they contradict the reality?

Does this answer your question: https://github.com/scala-js/scala-js/issues/2681#issuecomment-581153171?

tl;dr IMHO it is important to make sure there are no *unwanted* dependencies. Maybe we should call this `withAllowedStaticDependencies` or something like that.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
gzm0: Right... There is actually the additional detail that a module does not need an internal module if no other modules depend on it. This constraint seems to not be easily mapped into the type system. So we need to validate outputs anyways.

That being said, the more type safety the better is of course also an argument :S

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
```diff
@@ -0,0 +1,54 @@
...
+package org.scalajs.linker.interface
+
+/** A single Module output specification for a linker run.
+ *
+ *  The fields [[moduleID]] and [[isInternal]] uniquely identify this output in
+ *  a [[Linker]] run.
+ *
+ *  @param moduleID The ID of the module this linker output represents.
+ *
+ *  @param isInternal Whether this is an internal (or an external) module.
```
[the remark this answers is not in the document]
gzm0: Ah. An internal module is a module that is to be imported by other SJS emitted modules. An external module is a module that contains definitions exported from SJS.

external modules are always leafs from SJS's POV.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
gzm0: `moduleName` is the `"module-name"` of the import statement ([MDN][1]). It is used in the modules depending on this one to import it. This is why it is required for internal modules but not external modules. For external modules, this would be ignored but not harmful (thought is that it might simplify upstream code if we allow setting it all the time).

This is similar to the `URI`s in `LinkerOutput`: We allow to set them partially even if it is "typically meaningless".

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
gzm0: Probably a better design choice indeed. Although see my previous comment about `moduleName` being meaningful all the time and this maybe being unnecessary burden on upstream code.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
sjrd: It may not need a module-name, but it still needs an internal module, just so you know what classes to put in what file.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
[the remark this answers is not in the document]
sjrd: Is it then the case that a module is external iff one of the classes it selects contains a top-level export?

What if I have a module that contains a top-level export *and* declares a Scala class that is used by another module? Should that be ruled out?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
sjrd: OK, I see.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleLinkerOutput.scala
sjrd: I'm not sure it helps unless `withInternal` takes a boolean. Because it's a choice between (currently):
```scala
val moduleOutput0 = ModuleLinkerOutput(???)
  .withModuleName(someModuleName)
val moduleOutput =
  if (isInternal) moduleOutput0.withInternal
  else moduleOutput0
```
and (with `withModule` taking the `moduleName`):
```scala
val moduleOutput0 = ModuleLinkerOutput(???)
val moduleOutput =
  if (isInternal) moduleOutput0.withInternal(someModuleName)
  else moduleOutput0
```
It would be simplified if `withInternal` took a boolean instead, as that would allow:
```scala
val moduleOutput = ModuleLinkerOutput(???)
  .withModuleName(someModuleName)
  .withInternal(isInternal)
```

PR #3966, opened 2020-02-20
Inline keepOnlyTrackedGlobalRefs in FunctionEmitter

gzm0:
It was only used there.


gzm0:
Given the outcome of #3967, I'll need to rethink some parts of the dependency tracking, so I'll close this for now.

PR #3967, opened 2020-02-20
Make ModuleInitializers a backend only concept

gzm0:
This simplifies symbol requirement handling and LinkingUnit transformation.

Downsides:
- Impossible to optimize module initializers
- LinkingUnit does not contain "all code" anymore

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -17,7 +17,7 @@ import java.util.concurrent.ConcurrentHashMap
 import scala.util.matching.Regex
 
 object ScalaJSVersions extends VersionChecks(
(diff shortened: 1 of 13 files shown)


gzm0:
@sjrd WDYT overall? This kind of paves the path to make modules a backend only concept (which for now would work to implement #2681). However, it might have implications I'm unaware about. So to some extent I'm sending this early to probe you about it :P

sjrd:
TBH I don't like this. To me, it is pretty fundamental that the module initializers are part of the `LinkingUnit` and must be known by the frontend. The reason is that they are an integral part of the *semantics* of the whole program (it's also why the `CoreSpec` are in linking units). The frontend should know about the semantics of the whole program.

For the same reason, to the extent that module splitting affects the visible semantics of a program, I would *expect* the frontend to know about them, and for their spec to be stored in `LinkingUnit`.

gzm0:
That's what I thought :)

PR #3968, opened 2020-02-24
Fix #3939: Ignore methods in jl.Object in non-native JS classes

### LINKED ISSUE #3939: Compile error on a method with @JSName("finalize") in a non-native JS class
```scala
// based on https://github.com/scala-js/scalajs-tutorial/blob/0.6.x/src/main/scala/tutorial/webapp/TutorialApp.scala
package tutorial.webapp

import scala.scalajs.js

@js.annotation.ScalaJSDefined
class Test extends js.Object {
  @js.annotation.JSName("finalize")
  def foo(): Unit = {
    println("success")
  }
}

object TutorialApp {
    new Test().foo()
    println("hoge")
  }
}
```

This code should print "success" and "hoge", but it cannot be compiled:

Scala 2.12.10 / ScalaJS 0.6.31 (and 1.0.0-RC2)

```
sbt:Scala.js Tutorial> run
[info] Compiling 1 Scala source to D:\hoge\target\scala-2.12\classes ...
[warn] D:\hoge\src\main\scala\tutorial\webapp\TutorialApp.scala:5:2: @ScalaJSDefined is deprecated: add `-P:scalajs:sjsDefinedByDefault` to your scalac options and simply remove `@ScalaJSDefined`
[warn] @js.annotation.ScalaJSDefined
[warn]  ^
[error] D:\hoge\src\main\scala\tutorial\webapp\TutorialApp.scala:8:7: Cannot disambiguate overloads for method finalize with types
[error]   ()Unit
[error]   ()Unit
[error]   def foo(): Unit = {
[error]       ^
[warn] one warning found
[error] one error found
[error] Total time: 1 s, completed 2020/01/22 22:03:20
```

In ScalaFiddle, it can be compiled but it only prints "hoge": https://scalafiddle.io/sf/rPmTeX6/0

gzm0:


gzm0:
Seems like the `has-pr` label died :)

sjrd:
> Seems like the `has-pr` label died :)

Hum? What do you mean?

gzm0:
> Hum? What do you mean?

There seems to be native support now via "Linked issues" / "Linked pull requests". So there's no need for a label anymore.

sjrd:
Oh, I see. And yes, it's even shown in the issue list. So indeed, that sounds like this label is not useful anymore :)

compiler/src/test/scala/org/scalajs/core/compiler/test/ScalaJSDefinedTest.scala
```diff
@@ -691,4 +691,39 @@ class ScalaJSDefinedTest extends DirectTest with TestHelpers {
...
+  def overrideJavaLangObjectMethod: Unit = {
+    val objs = List("class", "object")
+
+    for {
+      obj <- objs
+    } {
+      s"""
+      $obj A extends js.Object {
+        override def clone(): Object = null
+        override def finalize(): Unit = ()
```
sjrd: I wonder if we shouldn't allow `clone()` and `finalize()` nevertheless. Since they are `protected`, they can never be called on an `Object`. They can only be called on a subclass that redefines them to be public. So in practice the "dangerous" behavior of an `override` not actually overriding some behavior cannot be observed. Therefore, it doesn't seem so dangerous.

The only real danger is overriding `hashCode()` and `equals(Any)`, because there the false overriding relationship can actually be observed. Perhaps we should target the warning at those two methods, and specifically mention them in the warning message?

compiler/src/test/scala/org/scalajs/core/compiler/test/ScalaJSDefinedTest.scala
```diff
@@ -691,4 +691,39 @@ class ScalaJSDefinedTest extends DirectTest with TestHelpers {
     """
   }
 
+  @Test
+  def overrideJavaLangObjectMethod: Unit = {
+    val objs = List("class", "object")
```
sjrd: We should also have a test for what we want to happen in a `trait` (with abstract methods, obviously). We probably also want to warn in that case.

sjrd: We should also decide whether we want to warn for such definitions in `@js.native` stuff.

gzm0:
Updated. I have split the warning into a separate issue.

PR #3970, opened 2020-02-24
Fix #3969: Warn on misleading overrides of jl.Object in JS classes

### LINKED ISSUE #3969: Warn on misleading overrides of jl.Object in JS classes
I wonder if we shouldn't allow `clone()` and `finalize()` nevertheless. Since they are `protected`, they can never be called on an `Object`. They can only be called on a subclass that redefines them to be public. So in practice the "dangerous" behavior of an `override` not actually overriding some behavior cannot be observed. Therefore, it doesn't seem so dangerous.

The only real danger is overriding `hashCode()` and `equals(Any)`, because there the false overriding relationship can actually be observed. Perhaps we should target the warning at those two methods, and specifically mention them in the warning message?

_Originally posted by @sjrd in https://github.com/scala-js/scala-js/pull/3968_

gzm0:


compiler/src/test/scala/org/scalajs/core/compiler/test/JSInteropTest.scala
```diff
@@ -2946,4 +2946,91 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+    }
+    """ hasWarns
+    """
+       |newSource1.scala:7: warning: Overriding hashCode in a JS class does not change its hash code. To silence this warning, change the name of the method and optionally add @JSName("hashCode").
+       |      override def hashCode(): Int
+       |                   ^
+       |newSource1.scala:8: warning: Overriding equals in a JS class does not change how it is compared. To silence this warning, change the name of the method and optionally add @JSName("equals").
+       |      override def equals(obj: Any): Boolean
+       |                   ^
+    """
```
gzm0: @sjrd this test case fails. Am I using the overrides cursor wrong?

PR #3971, opened 2020-02-26
Fix #3853: Split Emitter.symbolRequirements

### LINKED ISSUE #3853: Split Emitter.symbolRequirements
We should split them into:

- Implementation details of the Emitter (the true Emitter.symbolRequirements).
- Things that should be reached by the Analyzer.

See https://github.com/scala-js/scala-js/pull/3817#discussion_r344515580 for details.

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -24,8 +24,6 @@ import org.scalajs.linker.backend.emitter.Transients._
 
 object Infos {
 
-  private val ArithmeticExceptionClass = ClassName("java.lang.ArithmeticException")
-
   private val StringArgConstructorName =
     MethodName.constructor(List(ClassRef(BoxedStringClass)))
```
gzm0: This is still a wart: Only the `Emitter` should know about this.

ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -460,6 +460,18 @@ object Names {
   /** The class of things returned by `ClassOf` and `GetClass`. */
   val ClassClass: ClassName = ClassName("java.lang.Class")
 
+  /** The exception thrown by a division by 0. */
+  val ArithmeticExceptionClass: ClassName =
+    ClassName("java.lang.ArithmeticException")
+
+  /** The exception thrown by an `ArraySelect` that is out of bound. */
```
sjrd:
```suggestion
  /** The exception thrown by an `ArraySelect` that is out of bounds. */
```

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -544,9 +544,18 @@ object Infos {
 
             case IsInstanceOf(_, testType) =>
               builder.maybeAddUsedInstanceTest(testType)
+
             case AsInstanceOf(_, tpe) =>
               builder.maybeAddUsedInstanceTest(tpe)
 
+              /* In theory, we'd need to reach ClassCastException
```
sjrd: This reads as if it applies to nodes *below* the comment, despite the indentation.
Consider moving that right after the line `case AsInstanceOf(_, tpe) =>`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -572,11 +581,26 @@ object Infos {
...
+               * Instead, the Emitter simply requests the exception in its
+               * symbol requirements.
+               */
+
             case ClassOf(cls) =>
               builder.maybeAddAccessedClassData(cls)
 
+            case GetClass(_) =>
+              // `GetClass` instantiates a java.lang.Class.
+              builder.addInstantiatedClass(ClassClass, ObjectArgConstructorName)
```
sjrd: This should also be added to `ClassOf`, I believe.

(Not that it makes an actual difference, because there will *always* be a `GetClass` node in `java.lang.Object.toString()`, which is always reachable because `jl.Object` is `instantiated()` and `toString()` is exported.)

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -545,7 +545,7 @@ private[emitter] object CoreJSLib {
             paramList((instance :: args): _*), body)
       }
 
-      // toString()java.lang.String
+      // toString()java.lang.String is special (it is always exported).
```
sjrd: It's not because it is always exported. In fact, the IR spec doesn't specify that `j.l.Object.toString()` must be exported (there is an actual export definition in the hand-written IR of `j.l.Object`).

`toString()` is special here because the IR spec says so: see [here](https://lampwww.epfl.ch/~doeraene/sjsir-semantics/#sec-sjsir-apply-runtime-semantics-evaluation), step 8.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -554,49 +554,18 @@ private[emitter] object CoreJSLib {
...
-      }
-
-      def defineStandardDispatcher(methodName: MethodName, args: List[VarRef],
-          implementationInObject: Option[Tree],
-          maybeImplementingHijackedClasses: List[ClassName]): Unit = {
+      /* A standard dispatcher performs a type test on the instance and then
+       * calls the relevant implementation which is either of:
+       *
+       * - A normal method call if the instance is a normal scala class.
+       * - A default method the relevant hijacked class.
```
sjrd:
```suggestion
       * - A method in the relevant hijacked class.
```
Methods in hijacked classes are not *default* methods. They're just ... methods in hijacked classes.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -609,26 +578,39 @@ private[emitter] object CoreJSLib {
...
-            allButNormal
+          if (implementedInObject) {
+            val staticObjectCall: Tree = {
+              val fun = encodeClassVar(ObjectClass).prototype DOT genName(methodName)
+              Apply(fun DOT "call", instance :: args)
+            }
+
+            If(genIsScalaJSObjectOrNull(instance),
+              Return(normalCall),
+              hijackedDispatch(staticObjectCall))
```
sjrd:
```suggestion
            If(genIsScalaJSObjectOrNull(instance),
                Return(normalCall),
                hijackedDispatch(staticObjectCall))
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -645,46 +627,14 @@ private[emitter] object CoreJSLib {
...
-
-      defineStandardDispatcher(subSequenceMethodName,
-          varRef("start") :: varRef("end") :: Nil, None,
-          List(BoxedStringClass))
-
       for {
-        methodName <- List(byteValueMethodName, shortValueMethodName,
-            intValueMethodName, longValueMethodName, floatValueMethodName,
-            doubleValueMethodName)
+        (methodName, representativeClasses) <- globalKnowledge.getRepresentativeClassMethods()
```
sjrd: That's neat, but it's a bit *too* neat. There are a number of things that should be pruned from that list:

* Methods that are only defined in hijacked classes themselves, but none of their super types. In particular, most methods of `String`. Those will never be called through a dispatcher, because the static type is needed to call them, and hence `FunctionEmitter` can generate a direct call to the correct implementation. (This is an optimization.)
* `Byte`, `Short`, `Integer` and `Float` must be filtered out, because `Double` takes care of them. This is required for correctness, because for example `Integer.compareTo(Object)` casts its argument to `int`, which fails if the left-hand-side is `5.0` and the rhs is `5.5`, for example.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
gzm0: Then why are they on the "f" env field?

gzm0:
System.identityHashCode of a JS class causes a stack overflow because `Object#hashCode` and `System.identityHashCode` keep delegating to each other. Why do we have an intrinsic for `identityHashCode` in the first place? It doesn't seem to do anything different (except for bigints, but they could simply be treated in the branch dealing with `scala.Double` as well).

gzm0:
Ok, so it seems we have navigate ourselves a bit into a corner here: The stdlib's implementation of `System.identityHashCode` assumes support from the linker. Also, it seems otherwise it exploits UB if `WeakMap` is not available (setting an unexported field on a Scala class).

My take on this is the best course of action at this point is to introduce a new IR node `IdentityHashCode` (or similar) and make this a functionality provided by the IR itself. Subsequently,  `System.identityHashCode` can be made a compiler primitive.

@sjrd WDYT?

gzm0:
I have just realized that introducing a new IR node for ID hash code does not address the compatibility problem we have created here. My current understanding of the situation is as follows:

1. The Scala.js linker has a bug: It does not respect [`SJSIR 9.25.2`][1] but makes assumptions about the relevant implementations (leaving intrinsics aside for a moment).
2. The Scala.js library has a bug: It relies on the bug in the linker (1) for the implementation of `identityHashCode` to be correct.

Fixing just one of these two bugs will cause a user visible behavior change. Not fixing these bugs, will lower the value of the IR, because it does not behave as specified.

We seem to have the following options:
1. Fix the bugs in 1.0.x
1. Fix the bugs in 2.0.x
1. Change the IR spec (but no code).

IMHO our best course of action here is to do 1, with a bump to the IR version and the linker logging a warning if it is linking against 1.0 IR. Most users will be unaffected, because they just use the sbt plugin, which bumps the library version anyways. More advanced users might be affected, but at this point also should care, since they potentially "see" the bug pair described above.

[1]: http://lampwww.epfl.ch/~doeraene/thesis/sjsir-semantics/#sec-sjsir-apply-runtime-semantics-evaluation

sjrd:
That is a pretty bad situation, indeed. At this point the issue with `hashCode` is clearly beyond the scope of this PR. I suggest opening an issue with this specific problem, and in this PR keep the dispatcher for `hashCode` as special, the way it was before, with a comment referring to the issue. We can further discuss in the issue what's the best path forward beyond that.

gzm0:
Blocked on #3976.

PR #3972, opened 2020-02-26
Fix #3914: Disable IR checker in fastOptJS

### LINKED ISSUE #3914: Disable IR checker for fastOptJS
This is a follow-up of #3072 where we always enabled the IR checker in 1.0.x. Unclear if we should do this in 1.0.0 or in 1.0.1.

gzm0:
To do this, we switch the default to false again.

### DIFF
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
@@ -62,7 +62,7 @@ final class StandardConfig private (
         semantics = Semantics.Defaults,
         moduleKind = ModuleKind.NoModule,
         esFeatures = ESFeatures.Defaults,
-        checkIR = true,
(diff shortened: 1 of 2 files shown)


gzm0:
@sjrd WDYT? I'm not sure when the right time is for this. But we also don't want everybody to pay the performance penalty.

sjrd:
I think we should wait until a significant set of libraries have upgraded, so that we are reasonably sure that we caught the remaining issues, and that we're not publishing jars with invalid IR.

gzm0:
@sjrd I propose to merge this post the v1.1.0 cut. WDYT?

sjrd:
Yes, that will probably be a good time. :)

gzm0:
Rebased.

I have thought about two alternatives to this:
- Leave the default to `true` only set it to false in `fastOptJS`.
- Always IR check in the first run of a linker.

IMO this PR is still the best option.

PR #3973, opened 2020-02-26
Implement java.lang.Iterable.forEach.

derrickoswald:
- adds the forEach method, but not the spliterator method
- added IterableTest and shoe-horned it under CollectionTest in the test class hierarchy

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Iterable.scala
+++ b/javalanglib/src/main/scala/java/lang/Iterable.scala
@@ -13,7 +13,17 @@
 package java.lang
 
 import java.util.Iterator
+import java.util.function.Consumer
+
+import scala.scalajs.js.annotation.JavaDefaultMethod
 
 trait Iterable[T] {
   def iterator(): Iterator[T]
+
+  @JavaDefaultMethod
+  def forEach(action: Consumer[_ >: T]): Unit = {
+    val iter = iterator()
+    while (iter.hasNext())
+      action.accept(iter.next())
+  }
 }
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
@@ -1,0 +1,86 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.javalib.lang
+
+import java.lang.{Iterable => JIterable}
+import java.{util => ju}
+import java.util.function.Consumer
+
+import scala.reflect.ClassTag
+
+import org.junit.Test
+import org.junit.Assert._
+
+/** Tests the implementation of the java standard library Iterable
+ */
+trait IterableTest {
+
+  def factory: IterableFactory
+
+  @Test def empty(): Unit = {
+    val iter = factory.fromElements[Int]()
+    var hit = false
+    iter.forEach(new Consumer[Int] {
+      def accept(x: Int): Unit = hit = true
+    })
+    assertTrue(!hit)
+  }
+
+  @Test def simpleSum(): Unit = {
+    val iter = factory.fromElements[Int](42, 50, 12, 0, -45, 102, 32, 75)
+    var sum = 0
+    iter.forEach(new Consumer[Int] {
+      def accept(x: Int): Unit = sum = sum + x
+    })
+    assertEquals(268, sum)
+  }
+}
+
+class IterableDefaultTest extends IterableTest {
+  def factory: IterableFactory = new IterableFactory {
+    override def implementationName: String = "java.lang.Iterable"
+
+    def empty[E: ClassTag]: JIterable[E] = {
+      new JIterable[E] {
+        override def iterator(): ju.Iterator[E] = {
+          new ju.Iterator[E] {
+            override def hasNext(): Boolean = false
+            override def next(): E = throw new NoSuchElementException()
+            override def remove(): Unit = throw new NoSuchElementException()
+          }
+        }
+      }
+    }
+
+    def fromElements[E: ClassTag](elems: E*): JIterable[E] = {
+      new JIterable[E] {
+        override def iterator(): ju.Iterator[E] = {
+          val l: Iterator[E] = elems.toIterator
+          new ju.Iterator[E] {
+            override def hasNext(): Boolean = l.hasNext
+            override def next(): E = l.next
+            override def remove(): Unit = throw new NoSuchElementException()
+          }
+        }
+      }
+    }
+  }
+}
+
+trait IterableFactory {
+  def implementationName: String
+
+  def empty[E: ClassTag]: JIterable[E]
+
+  def fromElements[E: ClassTag](elems: E*): JIterable[E]
+}
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/CollectionTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/CollectionTest.scala
@@ -17,13 +17,15 @@ import java.{util => ju, lang => jl}
 import org.junit.Test
 import org.junit.Assert._
 
+import org.scalajs.testsuite.javalib.lang.IterableFactory
+import org.scalajs.testsuite.javalib.lang.IterableTest
 import org.scalajs.testsuite.utils.AssertThrows._
 
 import scala.reflect.ClassTag
 
 import Utils._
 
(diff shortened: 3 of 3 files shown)


javalanglib/src/main/scala/java/lang/Iterable.scala
```diff
@@ -13,7 +13,14 @@
 package java.lang
 
 import java.util.Iterator
+import java.util.function.Consumer
 
 trait Iterable[T] {
   def iterator(): Iterator[T]
+
+  def forEach(action: Consumer[T]): Unit = {
```
sjrd: According to the JavaDoc, the argument should be a `Consumer[_ >: T]` instead of `Consumer[T]`.

javalanglib/src/main/scala/java/lang/Iterable.scala
```diff
@@ -13,7 +13,14 @@
...
+    val iter = iterator()
+    while (iter.hasNext)
```
sjrd: Our coding style asks that we always use `()` when calling Java methods. So this should be `.hasNext()`.

javalanglib/src/main/scala/java/lang/Iterable.scala
```diff
@@ -13,7 +13,14 @@
...
+      action.accept (iter.next())
```
sjrd: Spurious space before the `(`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.javalib.lang
+
+import java.lang.{Iterable => JIterable}
+import java.util
```
sjrd: Prefer using
```scala
import java.{util => ju}
```
then `ju.Whatever` of you need this. Importing `util` is confusing, because `scala.util` also exists.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    var sum = 0
+    iter.forEach(new Consumer[Int] {
+      def accept(x: Int): Unit = sum = sum + x
+    })
+    assertEquals(268, sum)
+  }
+}
+
+class IterableDefaultTest extends IterableTest
+{
```
sjrd: This `{` should be at the end of the previous line.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+  def factory: IterableFactory = new IterableFactory {
+    override def implementationName: String = "testIterable"
```
sjrd: The value here should be `"java.lang.Iterable"`, since this factory tests the implementation in `java.lang.Iterable`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    def empty[E: ClassTag]: JIterable[E] = {
+      new JIterable[E] {
+        override def iterator(): util.Iterator[E] = {
+          new util.Iterator[E] {
+            override def hasNext: Boolean = false
+            override def next(): E = throw new NoSuchElementException()
+            override def remove(): Unit = throw new NoSuchElementException()
+            override def forEachRemaining(action: Consumer[_ >: E]): Unit = {
```
sjrd: I believe defining this method is not necessary.

Same thing below in `fromElements`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+class IterableDefaultTest extends IterableTest
+{
+  def factory: IterableFactory = new IterableFactory {
+    override def implementationName: String = "testIterable"
+
+    def empty[E: ClassTag]: JIterable[E] = {
+      new JIterable[E] {
+        override def iterator(): util.Iterator[E] = {
+          new util.Iterator[E] {
+            override def hasNext: Boolean = false
```
sjrd: Our coding style asks that we always define Java methods with `()`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+            }
+          }
+        }
+      }
+    }
+
+    def fromElements[E: ClassTag](elems: E*): JIterable[E] = {
+      val l = elems.toIterator
+      new JIterable[E] {
+        override def iterator (): util.Iterator[E] = {
```
sjrd: Spurious space before `()`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+          new util.Iterator[E] {
+            override def hasNext: Boolean = l.hasNext
+            override def next (): E = l.next
```
sjrd: Spurious space before ().

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,95 @@
...
+              while (hasNext())
+                action.accept(next())
+            }
+          }
+        }
+      }
+    }
+
+    def fromElements[E: ClassTag](elems: E*): JIterable[E] = {
+      val l = elems.toIterator
```
sjrd: Putting this here is incorrect. Calling `iterator()` twice will return two iterators that are linked, instead of independent.

You should put that *inside* `def iterator()`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
derrickoswald: I can't see where "The value" you are mentioning is in the code.
JIterable is already `java.lang.Iterable`:
`import java.lang.{Iterable => JIterable}`
or is there something about the line immediately above:
` override def implementationName: String = "testIterable"

Maybe I just don't understand this commenting system. My bad.`

sjrd: I mean this:
```suggestion
    override def implementationName: String = "java.lang.Iterable"
```
;)

javalanglib/src/main/scala/java/lang/Iterable.scala
```diff
@@ -13,7 +13,14 @@
 package java.lang
 
 import java.util.Iterator
+import java.util.function.Consumer
 
 trait Iterable[T] {
   def iterator(): Iterator[T]
+
+  def forEach(action: Consumer[_ >: T]): Unit = {
```
sjrd: You'll need to add the `@JavaDefaultMethod` annotation for tests to pass on 2.10 and 2.11.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/IterableTest.scala
```diff
@@ -0,0 +1,86 @@
...
+            override def next(): E = throw new NoSuchElementException()
+            override def remove(): Unit = throw new NoSuchElementException()
+          }
+        }
+      }
+    }
+
+    def fromElements[E: ClassTag](elems: E*): JIterable[E] = {
+      new JIterable[E] {
+        val l: Iterator[E] = elems.toIterator
```
sjrd: This val still needs to go inside the `def iterator()`.

It could also have a better name, like `iter` or even `elemsIter`.

derrickoswald:
@sjrd 
I've rebased with squash.
If this isn't what is expected my git foo probably isn't up to the task.

gzm0:
Uh, oh. Seems something funky happened. At this point the best is probably to completely rewrite history:

```sh
# checkout the branch
git checkout add_foreach_to_iterable

# obliterate history of the branch (!), but not the changes
# this assumes that 0.6.x doesn't have any new changes.
git reset 0.6.x

# check the diff
git diff

# make a new single commit
git commit -m 'Implement java.lang.Iterable.forEach.'

# force push the branch
git push -f add_foreach_to_iterable
```

HTH

sjrd:
I took the liberty to apply the git magic to fix the branch.

sjrd:
Thanks @derrickoswald, and congrats on your first contribution to Scala.js.

PR #3977, opened 2020-03-02
Fix #3976: Apply dispatch for Object.hashCode is not according to spec

### LINKED ISSUE #3976: Apply dispatch for Object.hashCode is not according to spec
I have just realized that introducing a new IR node for ID hash code does not address the compatibility problem we have created here. My current understanding of the situation is as follows:

1. The Scala.js linker has a bug: It does not respect [`SJSIR 9.25.2`][1] but makes assumptions about the relevant implementations (leaving intrinsics aside for a moment).
2. The Scala.js library has a bug: It relies on the bug in the linker (1) for the implementation of `identityHashCode` to be correct.

Fixing just one of these two bugs will cause a user visible behavior change. Not fixing these bugs, will lower the value of the IR, because it does not behave as specified.

We seem to have the following options:
1. Fix the bugs in 1.0.x
1. Fix the bugs in 2.0.x
1. Change the IR spec (but no code).

IMHO our best course of action here is to do 1, with a bump to the IR version and the linker logging a warning if it is linking against 1.0 IR. Most users will be unaffected, because they just use the sbt plugin, which bumps the library version anyways. More advanced users might be affected, but at this point also should care, since they potentially "see" the bug pair described above.

[1]: http://lampwww.epfl.ch/~doeraene/thesis/sjsir-semantics/#sec-sjsir-apply-runtime-semantics-evaluation

_Originally posted by @gzm0 in https://github.com/scala-js/scala-js/pull/3971#issuecomment-593236225_

gzm0:


sjrd:
LGTM, but I think we should release 1.0.1 before we merge this, to address #3950 in the wild.

gzm0:
> LGTM, but I think we should release 1.0.1 before we merge this, to address #3950 in the wild.

Right. However, I think we should not include all the changes in master in `1.0.1`. IMHO the following two warrant a minor version increase anyways:

- 159a6260c Upgrade Google Closure Compiler
- 3fbbee648 Remove JSBuilders and simply return trees from Emitter

Since there seems to be nothing else that is user visible, maybe we just cherry-pick 995c51950?

gzm0:
Hah, seems like the IR cleaner has broken our usual approach of testing IR changes :-/

Unclear how to fix that. Adding a `disableHacks` parameter to `Serializers.deserialize` seems really bad :(

sjrd:
> 
> 
> > LGTM, but I think we should release 1.0.1 before we merge this, to address #3950 in the wild.
> 
> Right. However, I think we should not include all the changes in master in `1.0.1`. IMHO the following two warrant a minor version increase anyways:
> 
>     * [159a626](https://github.com/scala-js/scala-js/commit/159a6260cc20ace008371428c7b7fa20177c6e6f) Upgrade Google Closure Compiler
> 
>     * [3fbbee6](https://github.com/scala-js/scala-js/commit/3fbbee648563fb523b596a655cf2fa1cf9680dc8) Remove JSBuilders and simply return trees from Emitter
> 
> 
> Since there seems to be nothing else that is user visible, maybe we just cherry-pick [995c519](https://github.com/scala-js/scala-js/commit/995c5195002636e0f1167537e5f1afb533bba959)?

I guess we need to precisely define what warrants a minor version bump. I filed #3980 to discuss this.

sjrd:
> 
> 
> Hah, seems like the IR cleaner has broken our usual approach of testing IR changes :-/
> 
> Unclear how to fix that. Adding a `disableHacks` parameter to `Serializers.deserialize` seems really bad :(

Well ... we can argue that if the IR cleaner fixed something automatically by deserialization, then the linker would also correctly fix it on released code. So all we need is to allow serialization of `IdentityHashCode` in the intermediate commit.

sjrd:
Also, now that I think about it, in the past we used to do the following steps when introducing hacks:

1. Bump the IR version
2. Introduce the new IR nodes, implement them entirely, and introduce the deserialization hack so that it is always applied (regardless of the IR version read in the binary)
3. Change the compiler, and restrict the hack to only apply base the IR version read in the binary

The steps that are currently in this PR cannot be generalized to 2 hacks in the same IR version.

gzm0:
Based on the discussion in #3980, I agree we should release a 1.0.1 with what we have (probably post #3982).

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -181,83 +181,8 @@ object System {
...
-              // Otherwise, we unfortunately have to return a constant.
-              42
-            }
-          }
-      }
-    }
-  }
-
-  // Intrinsic
   def identityHashCode(x: Object): scala.Int =
```
sjrd: This method should be annotated with `@inline`, unless we enhance the optimizer's heuristics to consider `IdentityHashCode(arg)` to be inlinable by default. Otherwise the reachability analysis won't be able to get rid of `j.l.System$` in any codebase.

gzm0:
Updated:
- Make replacement conditional on namespace (otherwise we'll replace the static forwarder).
- Always set `@inline` on replacement (otherwise we'll blow up code form 1.0 for no reason).

PR #3981, opened 2020-03-04
Fix #3979: Duplicate 'use strict'; in fullOptJS

### LINKED ISSUE #3979: fullOpt contains 'use strict'; twice
It seems that GCC also emits a `'use strict';`.

gzm0:
GCC already emits a 'use strict' clause itself. We therefore only emit
one in the basic backend.

Note that putting the clause pre-GCC pass will also cause GCC to emit
it.

I have manually tested that both fastOptJS and fullOptJS contain
exactly one 'use strict' clause.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/BasicLinkerBackend.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/BasicLinkerBackend.scala
@@ -59,6 +59,7 @@ final class BasicLinkerBackend(config: LinkerBackendImpl.Config)
         val code = withWriter { writer =>
           val printer = new Printers.JSTreePrinter(writer)
           writer.write(emitterResult.header)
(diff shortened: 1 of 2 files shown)


gzm0:
@sjrd any idea how to test this?

sjrd:
Meh ... I'm not sure I want to fix this if it means that the `LinkerBackend` needs to be aware of `"use strict;"`. We had significantly improved the separation of responsibility about this kind of thing in 3fbbee648563fb523b596a655cf2fa1cf9680dc8. Now it seems we're making this a bit worse.

gzm0:
> We had significantly improved the separation of responsibility about this kind of thing in 3fbbee6.

Well, apparently not. Otherwise closure / emitter would agree on who is responsible for emitting `'use strict';`.

sjrd:
Hum, I guess that's true. :s

PR #3982, opened 2020-03-04
Optimize `x.getClass().getName()` not to instantiate `j.l.Class`.

sjrd:
There are a number of use cases for `x.getClass().getName()` in core code, including in `Object.toString()`, `Throwable.toString()` and `MatchError.getMessage()`. These normally require instantiating a `java.lang.Class` from its class data, then to fetch the name from the data.

We make this more efficient by intrinsifying `Class.getName()` with a more direct implementation if the receiver is a `GetClass(x)` or a `ClassOf(C)`.

With this optimization, it becomes possible for a codebase not to have any `ClassOf` or `GetClass` node at all, and hence for `java.lang.Class` not to be needed at all. In that case, the core JS lib was leaving an erroneous reference to it. To solve this, and optimize a bit more in the process, we add checks to remove `j.l.Class` support from the core JS lib if `j.l.Class` is never instantiated.

The above scenario actually happens for the hello world example.


gzm0:
This is so cool!!

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -2115,6 +2116,64 @@ private[optimizer] abstract class OptimizerCore(config: CommonPhaseConfig) {
...
+                  case CharType    => "C"
+                  case ByteType    => "B"
+                  case ShortType   => "S"
+                  case IntType     => "I"
+                  case LongType    => "J"
+                  case FloatType   => "F"
+                  case DoubleType  => "D"
+                  case NothingType => "E" // invalid
+                  case NullType    => "N" // invalid
+                }
```
gzm0: Could we re-use `charCodeOfPrimRef` in the `CoreJSLib` here? We might even move it to `ir`....

If yes, maybe the info in `definePrimitiveTypeDatas` could also be consolidated.

OTOH, it's probably unavoidable that the optimizer has some duplicate code :-/

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassTest.scala
```diff
@@ -44,15 +46,65 @@ class ClassTest {
...
+
+    test("void", classOf[Unit])
+    test("boolean", classOf[Boolean])
+    test("char", classOf[Char])
+    test("byte", classOf[Byte])
+    test("short", classOf[Short])
+    test("int", classOf[Int])
+    test("long", classOf[Long])
+    test("float", classOf[Float])
+    test("double", classOf[Double])
```
gzm0: We seem to be not testing that `classOf[Some[_]].getName` yield the right result in both optimized and non-optimized scenario. (there is `java_lang_Class_getName_under_normal_circumstances` above, but it does not have the inline / non-inline treatment).

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/OptimizerCore.scala
sjrd: I've added a `val charCode` in `PrimRef`, and used the existing `displayName` for the other match above.

gzm0:
> This is very hard to test, but easy to reason about:

What I do not understand is what prevents us from using, say `checkCast` directly on a `TypeData` from emitted code? (We might already be doing this, I don't even know an easy way to review this).

All our test would then happily pass (because we generate `checkCast`), but we might totally break something. (or am I misunderstanding the situation here).

sjrd:
That is possible, in theory. To make sure that that wasn't currently the case, I audited the codebase with the following commands:
```
$ git grep '"isAssignableFrom"'
.../linker/backend/emitter/CoreJSLib.scala:        MethodDef(static = false, StringLiteral("isAssignableFrom"),
$ git grep '"checkCast"'
.../linker/backend/emitter/CoreJSLib.scala:        MethodDef(static = false, StringLiteral("checkCast"), paramList(obj),
$ git grep '"getSuperclass"'
.../linker/analyzer/Analyzer.scala:    MethodName("getSuperclass", Nil, ClassRef(ClassClass))
.../linker/backend/emitter/CoreJSLib.scala:        MethodDef(static = false, StringLiteral("getSuperclass"), Nil, {
.../linker/backend/emitter/EmitterNames.scala:  val getSuperclassMethodName = MethodName("getSuperclass", Nil, ClassRef(ClassClass))
$ git grep '"getComponentType"'
.../linker/backend/emitter/CoreJSLib.scala:        MethodDef(static = false, StringLiteral("getComponentType"), Nil, {
.../linker/frontend/optimizer/OptimizerCore.scala:            m("getComponentType", Nil, ClassClassRef) -> ClassGetComponentType,
$ git grep '"newArrayOfThisClass"'
.../linker/backend/emitter/CoreJSLib.scala:        MethodDef(static = false, StringLiteral("newArrayOfThisClass"),
```
This reveals that there are only
* The *definitions* of those methods in `CoreJSLib`
* Uses of those strings in `MethodName`s (so not related to the JS names)

I am not sure how to *enforce* this going forward, though, short of, as I said above, creating a dedicated object within `getClassOf()` to be passed to `j.l.Class.<init>()`, instead of adding the functionality directly on the `$TypeData`.

gzm0:
This is really not ideal :( I have just been thinking that, for example, if in the future, we decide to implement `objectClassName` via `objectGetClass` if it is available, this problem gets worse, in the sense that we wont even test the alternative implementation of `objectClassName`.

For now, let's proceed with this then, but we should come up with a better strategy for these things going forward. (`getSuperClass` is in the same category IIUC).

PR #3985, opened 2020-03-05
Fix #3980: Document versioning / compatiblity guarantees

### LINKED ISSUE #3980: Define the compatibility guarantees of minor/patch versions
It's pretty clear by now what constitutes a major version bump. Any of the following would trigger a major version bump:

* Backward incompatible changes in the IR
* Backward binary incompatible changes in the standard library
* Backward incompatible changes in the contract for calling JDK APIs

However, we have not precisely defined what would cause a minor version bump over a patch version bump.

gzm0:


gzm0:
I have tried to summarize the discussion in the referenced issue. PTAL

sjrd:
We should probably accompany this developer-facing documentation with some user-facing documentation on the website, at some point.

PR #3987, opened 2020-03-07
Fix #3283: Fix incremental linking on configuration changes

### LINKED ISSUE #3283: fastOptJS should invalidate when relevant build.sbt settings change
cross-/re-posting https://github.com/scalacenter/scalajs-bundler/issues/206#issuecomment-360835179

1. run `fastOptJS`
1. change `scalaJSUseMainModuleInitializer`
1. `reload` in sbt console
1. re-run `fastOptJS`
1. observe that JS file is not modified
1. run sbt `clean`
1. re-run `fastOptJS`
1. new JS file reflects change from 2.

perhaps other settings have this issue as well?

MaximeKjaer:
This PR fixes #3283 by adding a JSON serializer and deserializer for `scalaJSModuleInitializers` and `scalaJSLinkerConfig`, so that the previous value of these settings can be compared to the current value. If the value has changed, linking is re-done.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -298,7 +298,7 @@ def Tasks = [
     sbt ++$scala irJS$v/test linkerJS$v/test &&
     sbt 'set scalaJSStage in Global := FullOptStage' \
         'set scalaJSStage in testSuite.v$v := FastOptStage' \
-        ++$scala irJS$v/test linkerJS$v/test &&
+        ++$scala irJS$v/test linkerJS$v/test linkerInterfaceJS$v/test &&
     sbt ++$scala testSuite$v/bootstrap:test &&
     sbt 'set scalaJSStage in Global := FullOptStage' \
         'set scalaJSStage in testSuite.v$v := FastOptStage' \
@@ -311,7 +311,7 @@ def Tasks = [
   "tools": '''
     setJavaVersion $java
     npm install &&
-    sbt ++$scala ir$v/test logging$v/compile linkerInterface$v/compile \
+    sbt ++$scala ir$v/test logging$v/compile linkerInterface$v/test \
         linker$v/compile jsEnvs$v/test nodeJSEnv$v/test testAdapter$v/test \
         ir$v/mimaReportBinaryIssues logging$v/mimaReportBinaryIssues \
         linkerInterface$v/mimaReportBinaryIssues linker$v/mimaReportBinaryIssues \
@@ -341,8 +341,7 @@ def Tasks = [
         ir$v/scalastyle compiler$v/scalastyle \
         compiler$v/test:scalastyle \
         logging$v/scalastyle logging$v/test:scalastyle \
-        linkerInterface$v/scalastyle \
-        linkerInterface$v/scalastyle \
+        linkerInterface$v/scalastyle linkerInterface$v/test:scalastyle \
         linker$v/scalastyle linker$v/test:scalastyle \
         jsEnvs$v/scalastyle jsEnvsTestKit$v/scalastyle nodeJSEnv$v/scalastyle \
         jsEnvs$v/test:scalastyle nodeJSEnv$v/test:scalastyle testAdapter$v/scalastyle \
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/CheckedBehavior.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/CheckedBehavior.scala
@@ -24,4 +24,16 @@ object CheckedBehavior {
   case object Compliant extends CheckedBehavior
   case object Fatal extends CheckedBehavior
   case object Unchecked extends CheckedBehavior
+
+  private[interface] implicit object CheckedBehaviorFingerprint
+      extends Fingerprint[CheckedBehavior] {
+
+    override def fingerprint(behavior: CheckedBehavior): String = {
+      behavior match {
+        case Compliant => "Compliant"
+        case Fatal     => "Fatal"
+        case Unchecked => "Unchecked"
+      }
+    }
+  }
 }
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
@@ -1,0 +1,75 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+import scala.collection.mutable
+
+private[interface] trait Fingerprint[T] {
+
+  /** Generate a fingerprint of an object.
+   *
+   * A fingerprint is an injective one-way serialization representing the
+   * object.
+   *
+   * @param obj Object to fingerprint
+   * @return A fingerprint of the object
+   */
+  def fingerprint(obj: T): String
+}
+
+private[interface] object Fingerprint {
+  def fingerprint[T: Fingerprint](obj: T): String =
+    implicitly[Fingerprint[T]].fingerprint(obj)
+
+  final class FingerprintBuilder(name: String) {
+    private val fields = new mutable.ListBuffer[String]
+
+    def addField[T: Fingerprint](name: String, value: T): FingerprintBuilder = {
+      fields.append(s"$name=${fingerprint(value)}")
+      this
+    }
+
+    def build(): String = fields.mkString(s"$name(", ",", ")")
+  }
+
+  implicit object BooleanFingerprint extends Fingerprint[Boolean] {
+    override def fingerprint(bool: Boolean): String = bool.toString
+  }
+
+  implicit object IntFingerprint extends Fingerprint[Int] {
+    override def fingerprint(int: Int): String = int.toString
+  }
+
+  implicit object StringFingerprint extends Fingerprint[String] {
+    override def fingerprint(str: String): String = str
+  }
+
+  implicit def optionFingerprint[T: Fingerprint]: Fingerprint[Option[T]] = {
+    new Fingerprint[Option[T]] {
+      override def fingerprint(opt: Option[T]): String = opt match {
+        case Some(x) => s"Some(${implicitly[Fingerprint[T]].fingerprint(x)})"
+        case None    => "None"
+      }
+    }
+  }
+
+  implicit def listFingerprint[T: Fingerprint]: Fingerprint[List[T]] = {
+    new Fingerprint[List[T]] {
+      override def fingerprint(list: List[T]): String = {
+        list
+          .map(implicitly[Fingerprint[T]].fingerprint)
+          .mkString("List(", ",", ")")
+      }
+    }
+  }
+}
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
@@ -17,6 +17,8 @@ import org.scalajs.ir.Types._
 
 import org.scalajs.linker.interface.unstable.ModuleInitializerImpl
 
+import Fingerprint.FingerprintBuilder
+
 /** A module initializer for a Scala.js application.
  *
  *  When linking a Scala.js application, a sequence of `ModuleInitializer`s can
@@ -87,4 +89,39 @@ object ModuleInitializer {
         MethodName(mainMethodName, ArrayOfStringTypeRef :: Nil, VoidRef),
         args)
   }
+
+  private implicit object MethodNameFingerprint
+      extends Fingerprint[MethodName] {
+
+    override def fingerprint(methodName: MethodName): String =
+      methodName.nameString
+  }
+
+  private implicit object ClassNameFingerprint extends Fingerprint[ClassName] {
+    override def fingerprint(className: ClassName): String =
+      className.nameString
+  }
+
+  private implicit object ModuleInitializerFingerprint
+      extends Fingerprint[ModuleInitializer] {
+
+    override def fingerprint(moduleInitializer: ModuleInitializer): String =
+      moduleInitializer.impl match {
+        case VoidMainMethod(className, encodedMainMethodName) =>
+          new FingerprintBuilder("VoidMainMethod")
+            .addField("className", className)
+            .addField("encodedMainMethodName", encodedMainMethodName)
+            .build()
+
+        case MainMethodWithArgs(className, encodedMainMethodName, args) =>
+          new FingerprintBuilder("MainMethodWithArgs")
+            .addField("className", className)
+            .addField("encodedMainMethodName", encodedMainMethodName)
+            .addField("args", args)
+            .build()
+      }
+  }
+
+  def fingerprint(moduleInitializer: ModuleInitializer): String =
+    Fingerprint.fingerprint(moduleInitializer)
 }
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleKind.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleKind.scala
@@ -49,4 +49,15 @@ object ModuleKind {
    */
   case object CommonJSModule extends ModuleKind
 
+  private[interface] implicit object ModuleKindFingerprint
+      extends Fingerprint[ModuleKind] {
+
+    override def fingerprint(moduleKind: ModuleKind): String = {
+      moduleKind match {
+        case NoModule       => "NoModule"
+        case ESModule       => "ESModule"
+        case CommonJSModule => "CommonJSModule"
+      }
+    }
+  }
 }
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
@@ -13,6 +13,7 @@
 package org.scalajs.linker.interface
 
 import CheckedBehavior._
+import Fingerprint.FingerprintBuilder
 
 final class Semantics private (
     val asInstanceOfs: CheckedBehavior,
@@ -177,6 +178,45 @@ object Semantics {
     def regexReplace(regex: scala.util.matching.Regex,
         replacement: String): RuntimeClassNameMapper = {
       regexReplace(regex.pattern, replacement)
+    }
+
+    private[interface] implicit object RuntimeClassNameMapperFingerprint
+        extends Fingerprint[RuntimeClassNameMapper] {
+
+      override def fingerprint(mapper: RuntimeClassNameMapper): String = {
+        mapper match {
+          case KeepAll    => "KeepAll"
+          case DiscardAll => "DiscardAll"
+
+          case RegexReplace(pattern, flags, replacement) =>
+            new FingerprintBuilder("RegexReplace")
+              .addField("pattern", pattern)
+              .addField("flags", flags)
+              .addField("replacement", replacement)
+              .build()
+
+          case AndThen(first, second) =>
+            new FingerprintBuilder("AndThen")
+              .addField("first", fingerprint(first))
+              .addField("second", fingerprint(second))
+              .build()
+        }
+      }
+    }
+  }
+
+  private[interface] implicit object SemanticsFingerprint
+      extends Fingerprint[Semantics] {
+
+    override def fingerprint(semantics: Semantics): String = {
+      new FingerprintBuilder("Semantics")
+        .addField("asInstanceOfs", semantics.asInstanceOfs)
+        .addField("arrayIndexOutOfBounds", semantics.arrayIndexOutOfBounds)
+        .addField("moduleInit", semantics.moduleInit)
+        .addField("strictFloats", semantics.strictFloats)
+        .addField("productionMode", semantics.productionMode)
+        .addField("runtimeClassNameMapper", semantics.runtimeClassNameMapper)
+        .build()
     }
   }
 
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
@@ -16,6 +16,8 @@ import scala.language.implicitConversions
 
 import java.net.URI
 
+import Fingerprint.FingerprintBuilder
+
 /** Configuration of a standard linker. */
 final class StandardConfig private (
     /** Scala.js semantics. */
@@ -157,6 +159,31 @@ final class StandardConfig private (
 
 object StandardConfig {
   import StandardConfigPlatformExtensions.ConfigExt
+
+  private implicit object StandardConfigFingerprint
+      extends Fingerprint[StandardConfig] {
+
+    override def fingerprint(config: StandardConfig): String = {
+      new FingerprintBuilder("StandardConfig")
+        .addField("semantics", config.semantics)
+        .addField("moduleKind", config.moduleKind)
+        .addField("esFeatures", config.esFeatures)
+        .addField("checkIR", config.checkIR)
+        .addField("optimizer", config.optimizer)
+        .addField("parallel", config.parallel)
+        .addField("sourceMap", config.sourceMap)
+        .addField("relativizeSourceMapBase",
+            config.relativizeSourceMapBase.map(_.toASCIIString()))
+        .addField("closureCompilerIfAvailable",
+            config.closureCompilerIfAvailable)
+        .addField("prettyPrint", config.prettyPrint)
+        .addField("batchMode", config.batchMode)
+        .build()
+    }
+  }
+
+  def fingerprint(config: StandardConfig): String =
+    Fingerprint.fingerprint(config)
 
   /** Returns the default [[StandardConfig]].
    *
--- a/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
+++ b/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
@@ -1,0 +1,47 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+import org.junit.Test
+import org.junit.Assert._
+
+import Fingerprint._
+
+class FingerprintBuilderTest {
+  @Test
+  def noFields(): Unit = {
+    val fingerprint = new FingerprintBuilder("Test0").build()
+    assertEquals("Test0()", fingerprint)
+  }
+
+  @Test
+  def oneField(): Unit = {
+    val fingerprint = new FingerprintBuilder("Test1")
+      .addField("test1", true)
+      .build()
+    assertEquals("Test1(test1=true)", fingerprint)
+  }
+
+  @Test
+  def multipleFields(): Unit = {
+    val fingerprint = new FingerprintBuilder("Test2")
+      .addField("test1", Some(1): Option[Int])
+      .addField("test2", None: Option[Int])
+      .addField("test3", List.empty[Int])
+      .addField("test4", List(1, 2, 3))
+      .build()
+    assertEquals(
+        "Test2(test1=Some(1),test2=None,test3=List(),test4=List(1,2,3))",
+        fingerprint)
+  }
+}
--- a/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/ModuleInitializerFingerprintTest.scala
+++ b/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/ModuleInitializerFingerprintTest.scala
@@ -1,0 +1,51 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+import org.junit.Test
+import org.junit.Assert._
+
+class ModuleInitializerFingerprintTest {
+  def assertFingerprintsNotEquals(mi1: ModuleInitializer, mi2: ModuleInitializer): Unit = {
+    assertNotEquals(ModuleInitializer.fingerprint(mi1),
+        ModuleInitializer.fingerprint(mi2))
+  }
+
+  @Test
+  def noFingerprintCollisionClassName(): Unit = {
+    val mi1 = ModuleInitializer.mainMethod("Test1", "main1")
+    val mi2 = ModuleInitializer.mainMethod("Test2", "main1")
+    assertFingerprintsNotEquals(mi1, mi2)
+  }
+
+  @Test
+  def noFingerprintCollisionMainMethodName(): Unit = {
+    val mi1 = ModuleInitializer.mainMethod("Test1", "main1")
+    val mi2 = ModuleInitializer.mainMethod("Test1", "main2")
+    assertFingerprintsNotEquals(mi1, mi2)
+  }
+
+  @Test
+  def noFingerprintCollisionType(): Unit = {
+    val mi1 = ModuleInitializer.mainMethod("Test1", "main1")
+    val mi2 = ModuleInitializer.mainMethodWithArgs("Test1", "main1", List())
+    assertFingerprintsNotEquals(mi1, mi2)
+  }
+
+  @Test
+  def noFingerprintCollisionArgs(): Unit = {
+    val mi1 = ModuleInitializer.mainMethodWithArgs("Test1", "main1", List())
+    val mi2 = ModuleInitializer.mainMethodWithArgs("Test1", "main1", List("x"))
+    assertFingerprintsNotEquals(mi1, mi2)
+  }
+}
--- a/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
+++ b/linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
@@ -1,0 +1,57 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+import java.net.URI
+
+import org.junit.Test
+import org.junit.Assert._
+
+import Semantics.RuntimeClassNameMapper._
+
+class StandardConfigFingerprintTest {
+  def assertFingerprintsNotEquals(sc1: StandardConfig, sc2: StandardConfig): Unit = {
+    assertNotEquals(StandardConfig.fingerprint(sc1),
+        StandardConfig.fingerprint(sc2))
+  }
+
+  @Test
+  def noFingerprintCollisionCheckIR(): Unit = {
+    val sc1 = StandardConfig().withCheckIR(false)
+    val sc2 = StandardConfig().withCheckIR(true)
+    assertFingerprintsNotEquals(sc1, sc2)
+  }
+
+  @Test
+  def noFingerprintCollisionRelativizeSourceMapBase(): Unit = {
+    val sc1 = StandardConfig().withRelativizeSourceMapBase(None)
+    val sc2 = StandardConfig().withRelativizeSourceMapBase(Some(new URI("a")))
+    assertFingerprintsNotEquals(sc1, sc2)
+  }
+
+  @Test
+  def noFingerprintCollisionESFeatures(): Unit = {
+    val sc1 = StandardConfig().withESFeatures(_.withUseECMAScript2015(true))
+    val sc2 = StandardConfig().withESFeatures(_.withUseECMAScript2015(false))
+    assertFingerprintsNotEquals(sc1, sc2)
+  }
+
+  @Test
+  def noFingerprintCollisionRuntimeClassNameMapper(): Unit = {
+    val sc1 = StandardConfig().withSemantics(_.withRuntimeClassNameMapper(
+        keepAll() andThen discardAll()))
+    val sc2 = StandardConfig().withSemantics(_.withRuntimeClassNameMapper(
+        regexReplace("""\d+""".r, "0")))
+    assertFingerprintsNotEquals(sc1, sc2)
+  }
+}
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -716,6 +716,7 @@ object Build {
       id = "linkerInterface", base = file("linker-interface/jvm")
   ).settings(
       commonLinkerInterfaceSettings,
+      libraryDependencies += "com.novocode" % "junit-interface" % "0.11" % "test",
   ).dependsOn(irProject, logging)
 
   lazy val linkerInterfaceJS: MultiScalaProject = MultiScalaProject(
@@ -724,8 +725,8 @@ object Build {
       MyScalaJSPlugin
   ).settings(
       commonLinkerInterfaceSettings,
-  ).withScalaJSCompiler.dependsOn(
-      library, irProjectJS, loggingJS,
+  ).withScalaJSCompiler.withScalaJSJUnitPlugin.dependsOn(
+      library, irProjectJS, loggingJS, jUnitRuntime % "test", testBridge % "test", jUnitAsyncJS % "test",
   )
 
   lazy val linkerPrivateLibrary: Project = (project in file("linker-private-library")).enablePlugins(
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -41,7 +41,6 @@ import org.scalajs.testing.adapter.{TestAdapter, HTMLRunnerBuilder, TestAdapterI
 import Loggers._
 
 import sjsonnew.BasicJsonProtocol._
-import sjsonnew.BasicJsonProtocol.seqFormat
 
 /** Implementation details of `ScalaJSPlugin`. */
 private[sbtplugin] object ScalaJSPluginInternal {
@@ -158,6 +157,12 @@ private[sbtplugin] object ScalaJSPluginInternal {
       concurrentRestrictions in Global +=
         Tags.limit((usesScalaJSLinkerTag in key).value, 1),
 
+      scalaJSModuleInitializersFingerprints in key :=
+        scalaJSModuleInitializers.value.map(ModuleInitializer.fingerprint),
+
+      scalaJSLinkerConfigFingerprint in key :=
+        StandardConfig.fingerprint((scalaJSLinkerConfig in key).value),
+
       key := Def.taskDyn {
         /* It is very important that we evaluate all of those `.value`s from
          * here, and not from within the `Def.task { ... }`, otherwise the
@@ -179,6 +184,19 @@ private[sbtplugin] object ScalaJSPluginInternal {
           val log = s.log
           val realFiles = irInfo.get(scalaJSSourceFiles).get
           val ir = irInfo.data
+
+          def moduleInitializersChanged = (scalaJSModuleInitializersFingerprints in key)
+            .previous
+            .exists(_ != (scalaJSModuleInitializersFingerprints in key).value)
+
+          def linkerConfigChanged = (scalaJSLinkerConfigFingerprint in key)
+            .previous
+            .exists(_ != (scalaJSLinkerConfigFingerprint in key).value)
+
+          val configChanged = moduleInitializersChanged || linkerConfigChanged
+          if (configChanged && output.exists()) {
+            output.delete() // triggers re-linking through FileFunction.cached
+          }
 
           FileFunction.cached(s.cacheDirectory, FilesInfo.lastModified,
               FilesInfo.exists) { _ => // We don't need the files
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
@@ -12,6 +12,8 @@
 
 package org.scalajs.linker.interface
 
+import Fingerprint.FingerprintBuilder
+
 /** ECMAScript features to use when linking to JavaScript.
  *
  *  The options in `ESFeatures` specify what features of modern versions of
@@ -94,4 +96,15 @@ object ESFeatures {
    *  - `allowBigIntsForLongs`: false
    */
   val Defaults: ESFeatures = new ESFeatures()
+
+  private[interface] implicit object ESFeaturesFingerprint
+      extends Fingerprint[ESFeatures] {
+
+    override def fingerprint(esFeatures: ESFeatures): String = {
+      new FingerprintBuilder("ESFeatures")
+        .addField("useECMAScript2015", esFeatures.useECMAScript2015)
+        .addField("allowBigIntsForLongs", esFeatures.allowBigIntsForLongs)
+        .build()
+    }
+  }
 }
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala
@@ -176,6 +176,10 @@ object ScalaJSPlugin extends AutoPlugin {
         "Module initializers of the Scala.js application, to be called when it starts.",
         AMinusTask)
 
+    val scalaJSModuleInitializersFingerprints = TaskKey[Seq[String]]("scalaJSModuleInitializersFingerprints",
+        "An internal task used to track changes to the `scalaJSModuleInitializers` setting",
+        KeyRanks.Invisible)
+
     val scalaJSUseMainModuleInitializer = SettingKey[Boolean]("scalaJSUseMainModuleInitializer",
         "If true, adds the `mainClass` as a module initializer of the Scala.js module",
         APlusSetting)
@@ -194,6 +198,10 @@ object ScalaJSPlugin extends AutoPlugin {
         "scalaJSLinkerConfig",
         "Configuration of the Scala.js linker",
         BPlusSetting)
+
+    val scalaJSLinkerConfigFingerprint = TaskKey[String]("scalaJSLinkerConfigFingerprint",
+        "An internal task used to track changes to the `scalaJSLinkerConfig` setting",
+        KeyRanks.Invisible)
 
     val scalaJSStage = SettingKey[Stage]("scalaJSStage",
         "The optimization stage at which run and test are executed", APlusSetting)
--- a/sbt-plugin/src/sbt-test/incremental/change-config-and-source/Main.scala.first
+++ b/sbt-plugin/src/sbt-test/incremental/change-config-and-source/Main.scala.first
@@ -1,0 +1,7 @@
+package org.scalajs.sbtplugin.test
+
+object Main {
+  def main(args: Array[String]): Unit = {
+    println("Hello world")
+  }
+}
--- a/sbt-plugin/src/sbt-test/incremental/change-config-and-source/Main.scala.second
+++ b/sbt-plugin/src/sbt-test/incremental/change-config-and-source/Main.scala.second
@@ -1,0 +1,8 @@
+package org.scalajs.sbtplugin.test
+
+object Main {
+  def main(args: Array[String]): Unit = {
+    // This file changes the printed message:
+    println("Hi planet")
+  }
+}
--- a/sbt-plugin/src/sbt-test/incremental/change-config-and-source/build.sbt
+++ b/sbt-plugin/src/sbt-test/incremental/change-config-and-source/build.sbt
@@ -1,0 +1,5 @@
+scalaVersion := "2.12.10"
+
+enablePlugins(ScalaJSPlugin)
+
+scalaJSUseMainModuleInitializer := false
--- a/sbt-plugin/src/sbt-test/incremental/change-config-and-source/project/build.sbt
+++ b/sbt-plugin/src/sbt-test/incremental/change-config-and-source/project/build.sbt
@@ -1,0 +1,1 @@
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % sys.props("plugin.version"))
--- a/sbt-plugin/src/sbt-test/incremental/change-config-and-source/test
+++ b/sbt-plugin/src/sbt-test/incremental/change-config-and-source/test
@@ -1,0 +1,16 @@
+$ copy-file Main.scala.first Main.scala
+> fastOptJS
+$ copy-file target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-1.js
+
+# When the linker config and source both change, re-running fastOptJS should re-link:
+> set scalaJSUseMainModuleInitializer := true
+$ copy-file Main.scala.second Main.scala
+> fastOptJS
+-$ must-mirror target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-1.js
+$ newer target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-1.js
+$ copy-file target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-2.js
+
+# However, this re-linking should not happen more than once:
+> fastOptJS
+$ must-mirror target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-2.js
+-$ newer target/scala-2.12/change-config-and-source-fastopt.js target/scala-2.12/output-2.js
--- a/sbt-plugin/src/sbt-test/incremental/change-config/Main.scala
+++ b/sbt-plugin/src/sbt-test/incremental/change-config/Main.scala
@@ -1,0 +1,7 @@
+package org.scalajs.sbtplugin.test
+
+object Main {
+  def main(args: Array[String]): Unit = {
+    println("Hello world")
+  }
+}
--- a/sbt-plugin/src/sbt-test/incremental/change-config/build.sbt
+++ b/sbt-plugin/src/sbt-test/incremental/change-config/build.sbt
@@ -1,0 +1,5 @@
+scalaVersion := "2.12.10"
+
+enablePlugins(ScalaJSPlugin)
+
+scalaJSUseMainModuleInitializer := false
--- a/sbt-plugin/src/sbt-test/incremental/change-config/project/build.sbt
+++ b/sbt-plugin/src/sbt-test/incremental/change-config/project/build.sbt
@@ -1,0 +1,1 @@
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % sys.props("plugin.version"))
--- a/sbt-plugin/src/sbt-test/incremental/change-config/test
+++ b/sbt-plugin/src/sbt-test/incremental/change-config/test
@@ -1,0 +1,14 @@
+> fastOptJS
+$ copy-file target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-1.js
+
+# When the linker config changes, re-running fastOptJS should re-link:
+> set scalaJSUseMainModuleInitializer := true
+> fastOptJS
+-$ must-mirror target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-1.js
+$ newer target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-1.js
+$ copy-file target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-2.js
+
+# However, this re-linking should not happen more than once:
+> fastOptJS
+$ must-mirror target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-2.js
+-$ newer target/scala-2.12/change-config-fastopt.js target/scala-2.12/output-2.js


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSConfigJSONProtocol.scala
```diff
@@ -0,0 +1,214 @@
...
+      ("esFeatures", conf.esFeatures) :*:
+      ("checkIR", conf.checkIR) :*:
+      ("optimizer", conf.optimizer) :*:
+      ("parallel", conf.parallel) :*:
+      ("sourceMap", conf.sourceMap) :*:
+      ("relativizeSourceMapBase", conf.relativizeSourceMapBase) :*:
+      ("closureCompilerIfAvailable", conf.closureCompilerIfAvailable) :*:
+      ("prettyPrint", conf.prettyPrint) :*:
+      ("batchMode", conf.batchMode) :*:
+      ("semantics", conf.semantics) :*:
```
MaximeKjaer: There's a bug in sjson-new where if I place the `semantics` object in the beginning, the deserializer crashes. I haven't been able to to a minimal bug repro of this, but placing the object at the end seems to solve the problem :man_shrugging:

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -210,7 +214,28 @@ private[sbtplugin] object ScalaJSPluginInternal {
             }
 
             Set(output, sourceMapFile)
-          } (realFiles.toSet)
+          }
+
+          val useChanged = (scalaJSUseMainModuleInitializerTask in key)
+            .previous
+            .exists(_ != (scalaJSUseMainModuleInitializerTask in key).value)
```
MaximeKjaer: In theory, this check should not be necessary, as a change in `scalaJSUseMainModuleInitializer` implies a change in `scalaJSModuleInitializers`. 

However, in practice, if I remove this call, the `previous` value of `scalaJSModuleInitializers` is not written correctly to the `previous` file. It's as if looking at the `.previous` of `scalaJSUseMainModuleInitializerTask` has a side-effect that makes `scalaJSModuleInitializers` be written correctly.

This is very weird, and I don't know how else to solve the problem.

sbt-plugin/src/test/scala/org/scalajs/sbtplugin/ScalaJSConfigJSONProtocolTest.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.Assert._
+
+import ScalaJSConfigJSONProtocol._
+import org.scalajs.linker.interface._
+
+import sjsonnew.JsonFormat
+import sjsonnew.BasicJsonProtocol._
+import sjsonnew.support.scalajson.unsafe.Converter.{fromJsonUnsafe, toJsonUnsafe}
+
+class ScalaJSConfigJSONProtocolTest {
```
MaximeKjaer: Is this the best place to put this test? How can I add it to the CI?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/unstable/RuntimeClassNameMapperImpl.scala
```diff
@@ -15,6 +15,38 @@ package org.scalajs.linker.interface.unstable
...
+    * equality.
+    */
+  final case class RegexReplace(pattern: String, flags: Int,
+      replacement: String)(
+      val compiledPattern: java.util.regex.Pattern)
+      extends RuntimeClassNameMapper
+
+  final case class AndThen(first: RuntimeClassNameMapper,
+      second: RuntimeClassNameMapper)
+      extends RuntimeClassNameMapper
```
MaximeKjaer: In order to serialize different `RuntimeClassNameMapper`s, I needed to be able to know which kind of mapper I am serializing. However, this breaks the previous design that strongly hid the implementation details, and just passed a `RuntimeClassNameMapper` to the user as a black box that they could not peer into.

As discussed with @sjrd, this solution is not fantastic. I'm open to better solutions.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Have you considered dumping all this in a file instead? Then it will also work across sbt invocations.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/unstable/RuntimeClassNameMapperImpl.scala
gzm0: An alternative would be to provide the serialization only (and no deserialization for that matter). Because all we seem to care about is to check that it is unchanged. So we can compare serialized forms. (I realize this would mean not relying on `sjsjsonnew`).

sbt-plugin/src/test/scala/org/scalajs/sbtplugin/ScalaJSConfigJSONProtocolTest.scala
gzm0: Add `sbtPlugin/test` here: https://github.com/scala-js/scala-js/blob/master/Jenkinsfile#L364

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
MaximeKjaer: Do you mean manually writing to a file?

Sbt's `.previous` already works across invocations, it reads the serialized JSON of the previous value from a file called  `target/streams/<key>/<value>/previous`.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/unstable/RuntimeClassNameMapperImpl.scala
MaximeKjaer: Yes, that sounds like the right way to go. Moving these caused the tests to fail because of broken binary compatibility.

I could offer a serialization to a string of the whole linker config, store the serialized representation in a task instead of storing the object itself in the task, and then still rely on sbt's `.previous`. That does mean deleting the JSON format typeclasses for `StandardConfig`.

sjrd: It's for *serialization* that we need access to those things. *Deserialization* does not require those changes. So the alternative to only provide serialization does not make things any easier.

gzm0: Ah, that was maybe not clear from my comment. I meant to provide serialization in the linker interface directly. As such it would solve the problem of not exposing these things.

sjrd: Ah ! I hadn't thought about that. That's a funny way to look at the problem, but indeed, it seems to be a good approach.

gzm0: Thinking about this some more, it feels to me like this is the *only* correct approach for the entire `StandardConfig`. Otherwise adding a field to the `StandardConfig` becomes an incompatible change (because the serialization in sbt would not know about it and ignore it).

MaximeKjaer:
Making these changes brings the PR from 400 additions to 63, so I'm all in favor of this simpler solution.

I simply used `toString` as a serialization mechanism. Do you think that it would be cleaner to use a dedicated serialization method? In the end, that method would basically duplicate the functionality of `toString`.

gzm0:
> Do you think that it would be cleaner to use a dedicated serialization method?

Yes, absolutely. `toString` is for humans. Quoting JDK doc:

> The result should be a concise but informative representation that is easy for a person to read.

We might want to consider re-using some of the serialization stuff already in place for test-adapter / test-interface.

MaximeKjaer:
I refactored it to use serializers defined in `sbt-plugin`. I only changed `linker-interface` for the `RuntimeClassNameMapper`, by adding a `serialized` method that returns a `String` of the object, which the `sbt-plugin` serializer can use.

I wasn't sure if I should add `.serialized` methods on all config objects in `linker-interface`, or if I should define serializers in `sbt-plugin`. I opted for the latter: my thinking is that the serialization is specific to sbt, and the `linker-interface` should be modified as little as possible.

I ended up rolling my own serialization stuff, as I didn't really find what you referred to in `test-adapter` / `test-interface`. If you think it would be better to use something else and you can point me to it, please let me know.

sjrd:
As @gzm0 said at https://github.com/scala-js/scala-js/pull/3987#discussion_r391537960, the only correct solution is to write the entire serialization mechanism inside the `linker-interface`. The sbt plugin could be used to manipulate configuration from a more recent linker-interface (due to dependency eviction), in which case it would miss new fields if the serialization is in the sbt plugin.

This more critical for other build tools, which will likely also want to have that feature at some point. They shouldn't have to reimplement serialization. Plus other tools are much more prone to the version mismatch problem.

MaximeKjaer:
I hadn't thought of other build tools needing this, but these are all good point. I've fixed that in the latest commit.

Should I add some documentation for these `.serialized` methods and their intended purpose somewhere?

MaximeKjaer:
The tests fail because of the following error:

```
[info] Scala.js linker interface: found 1 potential binary incompatibilities while checking against org.scala-js:scalajs-linker-interface_sjs1_2.12:1.0.1 
[error]  * abstract method serialized()java.lang.String in class org.scalajs.linker.interface.ModuleKind is present only in current version
```

This is expected -- this PR adds to the existing interface.

sjrd:
This is OK because `ModuleKind` is effectively sealed (it has a private constructor, and all subclasses are objects).

You can fix this in two ways:
* Either you declare `def serialize` as *concrete` in `ModuleKind`, and use a pattern-match on `this`
* Or you filter out that MiMa error in `project/BinaryIncompatibilities.scala` (see https://github.com/scala-js/scala-js/blob/v1.0.1/project/BinaryIncompatibilities.scala for the format in which we write those)

I would recommend the former in this case, because `ModuleKind`s are typically pattern-matched over.

MaximeKjaer:
Ah, I see. The thinking behind the previous approach was that the pattern match wouldn't have exhaustivitiy checks, so I tried to avoid it. But I also see how adding an abstract method breaks the binary compat check.

Thank you for the pointer.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -110,6 +110,22 @@ final class StandardConfig private (
   def withBatchMode(batchMode: Boolean): StandardConfig =
     copy(batchMode = batchMode)
 
+  def serialized: String =
```
gzm0: Consider:

- Calling this `fingerprint`? (`serialized` kind of does imply that we can deserialize it...).
- Return an `Array[Byte]` (not 100% sure about this, but something to think about).
- Putting this on the companion (feels cleaner to me, probably because it separates the namespace of config values and operations).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -75,13 +75,23 @@ final class Semantics private (
     finalizeHash(acc, 6)
   }
 
+  def serialized: String =
```
gzm0: I would not expose these, but just make them private to `interface` (maybe also on the companion).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
MaximeKjaer: - I like `fingerprint` a lot. I've renamed all instances to that.
- I'm not convinced by `Array[Byte]` just yet. I have a few concerns that I'll address in the top-level conversation below.
- I agree. Done :)

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
MaximeKjaer: Done :+1:

MaximeKjaer:
Thank you for the review, and I'm sorry it took me so long to get back to you. I've renamed the methods to `fingerprint`, moved them to the companion objects, and made them `private[interface]` whenever possible.

About `Array[Byte]`, I can implement it if you'd like, but I'd like to just raise these few points first:

- A `String` is already convertible to an `Array[Byte]` with `.getBytes()`, so consumers of the `fingerprint` method do not gain from receiving an `Array[Byte]` over a `String`.
- The implementation of `Array[Byte]` is actually a little tricky if we want to avoid two different configs having colliding fingerprints, which would be a very annoying bug if it ever happens.
  
  For instance, if we have a class `Config(foo: Foo, bar: Bar)`, we'd like to handle the scenario where `fingerprint(foo) == [0]` and `fingerprint(bar) = [1, 2]` in one case, and `fingerprint(foo) == [0, 1]` and `fingerprint(bar) = [2]` in another case; the combined fingerprint should not be `[0, 1, 2]`.

  I think the solution would be to prepend the fingerprint of every member of a config class with its length, or with the name of the parameter. That might be a bit of boilerplate.

Let me know what you think -- again, I'm willing to implement this if you think it's worthwhile, but just wanted to discuss these points first.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -210,7 +216,31 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
+
+          def linkerConfigChanged = (scalaJSLinkerConfigFingerprint in key)
+            .previous
+            .exists(_ != (scalaJSLinkerConfigFingerprint in key).value)
+
+          val configChanged =
+            useMainChanged || moduleInitializersChanged || linkerConfigChanged
+
+          if (configChanged) {
+            link()
```
gzm0, line 237: Will the next run that uses `FileFunction.cached` work properly if you do not always use it?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: > Sbt's .previous already works across invocations, it reads the serialized JSON of the previous value from a file called target/streams/<key>/<value>/previous.

Ah, OK. In that case I guess we can use this. Although the additional keys are a bit hairy IMHO.

I would like you to investigate why `scalaJSUseMainModuleInitializer` needs to be handled explicitly. It should not be special. If that breaks, an custom setup to generate `scalaJSModuleInitializers` might also break.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -158,6 +158,22 @@ final class StandardConfig private (
 object StandardConfig {
   import StandardConfigPlatformExtensions.ConfigExt
 
+  def fingerprint(config: StandardConfig): String =
```
gzm0, line 161: I think this is a good API at this point.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -158,6 +158,22 @@ final class StandardConfig private (
...
+       |optimizer=${config.optimizer},
+       |parallel=${config.parallel},
+       |sourceMap=${config.sourceMap},
+       |relativizeSourceMapBase=
+       |${config.relativizeSourceMapBase.map(_.toASCIIString())},
+       |closureCompilerIfAvailable=${config.closureCompilerIfAvailable},
+       |prettyPrint=${config.prettyPrint},
+       |batchMode=${config.batchMode}
+       |)""".stripMargin.replaceAll("\n", "")
+
```
gzm0, line 176: I'd like us to explore other solutions to generate fingerprints. The main problem I have with the current approach is that it is difficult to see if the strings in a fingerprint are properly derived. For example, is `config.checkIR` ok? I need to go and look at the type of `checkIR` and then at its `toString` method to know. That's not ideal.
There is of course also the issue that one has to manually check all fields are there, but I don't think this can be avoided without macros (which we want to avoid more).

Did you have a look at the serializer API in test-common? If yes, is it infeasible to use it?

sjrd: > Did you have a look at the serializer API in test-common? If yes, is it infeasible to use it?

The linker interface should definitely not depend on test-common. ;)

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
sjrd, line 161: This method should have `{}` to surround its body, since it spans more than 1 line.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: > 
> 
> In theory, this check should not be necessary, as a change in `scalaJSUseMainModuleInitializer` implies a change in `scalaJSModuleInitializers`.
> 
> However, in practice, if I remove this call, the `previous` value of `scalaJSModuleInitializers` is not written correctly to the `previous` file. It's as if looking at the `.previous` of `scalaJSUseMainModuleInitializerTask` has a side-effect that makes `scalaJSModuleInitializers` be written correctly.
> 
> This is very weird, and I don't know how else to solve the problem.

Is this still true, even though everything was now moved to separate tasks with the fingerprints?

sjrd: What if you always store the `.value` of the fingerprints in `val`s, and the `.previous` in `val`s as well? Hiding them in local `def`s and in the lambdas passed to `.exists` might confuse sbt's macros that analyze the dependencies. In fact I am surprised it doesn't flag those usages as downright illegal.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
gzm0: Well yes, but we can move the code to somewhere else or source include it. It's just an implementation detail here, isn't it?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: Hum, that is potentially an issue. I'm not sure how `FileFunction` is supposed to be used to get around that.

Perhaps one way to workaround this difficulty is to always use `FileFunction.cached()`, but before that do
```scala
if (configChanged && output.exists())
  output.delete()
```
?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
sjrd: That looks like overengineering to me. Sourcing-in is not acceptable here because they would end up twice on classpaths that depend both on the linker interface and the test adapter (including our own sbt plugin!). And creating yet another artifact for this looks overkill, when simply using `toString`s and `fingerprint`s is basically sufficient.

Yes, there exists a possibility that we're going to use an implicit `toString()` instead of a `fingerprint()` by mistake. Is preventing ourselves from that mistake worth so much infrastructure? I don't think so.

gzm0: OK. Fair enough. But can we at least find a solution that makes it reasonably easy to visually inspect that a `fingerprint` is correct? With the current code, it is very hard and that, combined with it being hard to test, will set us up for likely bugs.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -81,7 +81,7 @@ final class Semantics private (
        |  arrayIndexOutOfBounds = $arrayIndexOutOfBounds,
        |  moduleInit            = $moduleInit,
        |  strictFloats          = $strictFloats,
-       |  productionMode        = $productionMode
+       |  productionMode        = $productionMode,
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 8080c2fef7ef0fa7c113bd5af619a10dc9557a50

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -132,20 +146,20 @@ object Semantics {
   }
 
   object RuntimeClassNameMapper {
-    private case object KeepAll extends RuntimeClassNameMapper
+    case object KeepAll extends RuntimeClassNameMapper
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 8080c2fef7ef0fa7c113bd5af619a10dc9557a50

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -132,20 +146,20 @@ object Semantics {
...
+    case object DiscardAll extends RuntimeClassNameMapper
 
     /* We use `pattern` and `flags` in the case parameters, rather than the
-     * `j.u.regex.Pattern`, because the latter does not have meaningful
-     * equality.
-     */
-    private final case class RegexReplace(pattern: String, flags: Int,
+      * `j.u.regex.Pattern`, because the latter does not have meaningful
+      * equality.
+      */
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 8080c2fef7ef0fa7c113bd5af619a10dc9557a50

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
MaximeKjaer: Done in 87b8de2b19bdc8b9e1518365a1a14733e87b6299

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
MaximeKjaer: > I would like you to investigate why `scalaJSUseMainModuleInitializer` needs to be handled explicitly. It should not be special.

I've looked into it,  and removing the explicit handling of the setting works now. I don't know why it didn't work before. I'm unable to replicate any of the weird behavior I had before, so I think I must've done something wrong in my testing setup back then.

In any case, this is fixed as of 49f4a0280d58ccbed98a5ad688eb223f4b2536ed. Changes in `scalaJSUseMainModuleInitializer` are indeed captured in `scalaJSModuleInitializers`, as expected.

MaximeKjaer: > Hiding them in local defs and in the lambdas passed to .exists might confuse sbt's macros that analyze the dependencies.

My initial implementation actually stored the comparison (with `.exists`) in a `val`. I eventually changed it to `def` so that the short-circuiting `||` operator could save us a file read in some scenarios.

Based on all my testing, it seems like it works, and sbt doesn't complain :man_shrugging:

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
MaximeKjaer: Ah, I see the problem. Currently, if I change a config setting *and* change some code, run `reload`, `fastOptJS` and `fastOptJS` again, it links twice: once because of the config change, and once because the cached file has changed. The compilation output stays correct, but we don't need to link the second time around, and can save some effort there.

I'll experiment with this and report back.

MaximeKjaer: The solution you suggested works fine, and is what I've implemented in 818ce79a384408cd7cd0ec51481b4966e26957ff

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
MaximeKjaer: What do you think of the following approach?

We could add a `FingerprintBuilder` with overloaded methods for each "fingerprintable" type:

```scala
class FingerprintBuilder(name: String) {
  def addField(name: String, value: Boolean): FingerprintBuilder = ???
  def addField(name: String, value: String): FingerprintBuilder = ???
  def build(): String = ???
}
```

This would then be called as follows:

```scala
object StandardConfig {
  def fingerprint(config: StandardConfig): String = {
    new FingerprintBuilder("StandardConfig")
      .addField("semantics", Semantics.fingerprint(config.semantics))
      // ...
      .addField("batchMode", config.batchMode)
      .build()
  }
}
```

There are perhaps more fancy variants of this idea, with `Fingerprint` wrappers or tagged types, or `Fingerprint[T]` typeclasses and the like, but I think the above might be good enough. 

The `FingerprintBuilder` could be tested with Scalacheck or something like that. Centralizing the fingerprinting logic also allows us to more easily change the type of a fingerprint `Array[Byte]` or something else if we choose to.

MaximeKjaer:
I've implemented most the feedback above, and have suggested (but not implemented) an approach for making it clearer that `fingerprint` is correct.

Some of the above commits fail the CI because the server used by the CI to install sbt was down for a bit today.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
gzm0: Oh, I like this approach. Will prevent most of the mishaps (implicit to string, reasonably OK to inspect for missing fields). @sjrd WDYT?

Just one thing, your comment says "with overloaded methods for each "fingerprintable" type:". However, the code suggests that a `FingerprintBuilder` only accepts primitives (hence the explicit call to `Semantics.fingerprint`). I think how the code currently is would be better. If `FingerprintBuilder` knows about all the fingerprintable types, that'd be kind of strange.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -180,37 +185,49 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
               }
-            } catch {
-              case e: LinkingException =>
-                throw new MessageOnlyException(e.getMessage)
-            }
 
-            Set(output, sourceMapFile)
-          } (realFiles.toSet)
+              Set(output, sourceMapFile)
+            } (realFiles.toSet)
```
gzm0, line 230: I think the indentation is all broken here, making the diff much larger than it actually is.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
MaximeKjaer: Oh, that's what I meant, but I see that how I wrote it isn't super clear. Indeed, the `FingerprintBuilder` shouldn't know about the linker types, but should only accept primitives. If needed, the implementations of `fingerprint` should explicitly call other `fingerprint` methods to convert members to primitives that `FingerprintBuilder` can use.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
MaximeKjaer: True, my bad. Fixed in 45db847395e8e303891e73da7fc50aaf4b15c930

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
MaximeKjaer: I spoke with @sjrd and he said he preferred an approach based on typeclasses, which is what I've implemented in 328801eabde719e7ac1b57593d3b541ec04612fa.

The problem with the approach I had previously suggested is that it exposes data in an intermediary form (string or boolean), which he said felt less clean than encapsulating the transformation to fingerprints in a typeclass (let me know if I've oversimplified the argument, @sjrd).

This new approach gets rid of all `private[interface] def fingerprint`, except for one in `RuntimeClassNameMapper`, which is needed to expose the private subtypes of `RuntimeClassNameMapper` as fingerprints.

gzm0: That is of course an even more luxurious approach. I'm absolutely OK with that. Some less-higher level comments:

- Have you considered moving `Fingerprint` to `unstable` (but still make it `private[interface]`)? (I'm genuinely not sure what the right place is her).
- It feels strange to me to separate the fingerprinting of an individual class from it's definition. Have you considered defining the typeclasses (except for primitive types) in the companions of the respective classes?
- Have you considered using `implicit object` instead of `implicit val` + anonymous class?

MaximeKjaer: - I don't think `Fingerprint` in its current form needs to go in `unstable`, as it's `private[interface]` anyway. What could however be marked "unstable" are the public parts of fingerprinting, namely the two public `fingerprint` methods (on `StandardConfig` and `ModuleInitializer`). I could mark those as unstable by:
  1. Moving `Fingerprint` to `unstable`
  2. Making `Fingerprint` public and making its members `private[interface]`
  3. Moving the two public  fingerprinting methods (on `StandardConfig` and `ModuleInitializer`) to `Fingerprint` and making them public.

  What do you think of that approach?
- I'm okay with that. Changed in 2e52230891c0352c29325346112ae1fd89e911e0
- That's a good idea, it's terser and probably a little simpler. Changed in 2e52230891c0352c29325346112ae1fd89e911e0

gzm0: > What do you think of that approach?

Ah, no the `fingerprint` higher level method needs to be in `interface`. It's a stable API we want to commit to.

So this approach is good with me :) I'll make some more detailed comments.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/CheckedBehavior.scala
```diff
@@ -24,4 +24,14 @@ object CheckedBehavior {
   case object Compliant extends CheckedBehavior
   case object Fatal extends CheckedBehavior
   case object Unchecked extends CheckedBehavior
+
+  private[interface] implicit object CheckedBehaviorFingerprint
+    extends Fingerprint[CheckedBehavior] {
+      override def fingerprint(behavior: CheckedBehavior): String =
```
gzm0, line 30: I believe the indentation is wrong here:
https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#class-declaration

`extends` should have 4 spaces, members only 2.

This applies elsewhere as well.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  implicit object StringFingerprint extends Fingerprint[String] {
+    override def fingerprint(str: String): String = str
+  }
+
+  implicit def optionFingerprint[T](implicit itemFingerprint: Fingerprint[T]): Fingerprint[Option[T]] =
+    new Fingerprint[Option[T]] {
+      override def fingerprint(opt: Option[T]): String =
+        opt.map(itemFingerprint.fingerprint).map(f => s"Some($f)").getOrElse("None")
```
gzm0, line 50: Use `fold` instead of `map` followed by `getOrElse`.

MaximeKjaer:
Thank you for reviewing my changes, I'll get those coding style issues fixed. 

I've pushed some tests for fingerprinting. If the approach for testing the fingerprinting works for you, I'll add tests for `StandardConfig` as well.

The idea behind only generating two choices for each field is to augment the chances of objects differing by only a single field that could've been forgotten in the fingerprint. I'm not so much testing that fingerprints can handle a breadth of inputs, but more that two different objects don't have the same fingerprint.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -13,6 +13,7 @@
 package org.scalajs.linker.interface
 
 import CheckedBehavior._
+import org.scalajs.linker.interface.Fingerprint.FingerprintBuilder
```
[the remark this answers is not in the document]
MaximeKjaer: I may have found such a creative way to import things that the code style doesn't cover it :) This is what my editor generated, but I think the unwritten convention is to only fully qualify an import path if necessary; imports from the same package do not need to be qualified.

Fixed in 5628647d30f13ee91950c380e1bbf5ea597acd01

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -13,9 +13,10 @@
 package org.scalajs.linker.interface
 
 import scala.language.implicitConversions
-
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 5628647d30f13ee91950c380e1bbf5ea597acd01

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
MaximeKjaer: Fixed in 5628647d30f13ee91950c380e1bbf5ea597acd01

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/CheckedBehavior.scala
MaximeKjaer: Fixed in 5628647d30f13ee91950c380e1bbf5ea597acd01

project/Build.scala
```diff
@@ -709,6 +709,7 @@ object Build {
       mimaBinaryIssueFilters ++= BinaryIncompatibilities.LinkerInterface,
       exportJars := true, // required so ScalaDoc linking works
 
+      libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.14.1" % "test",
```
gzm0, line 712: We cannot depend on scalacheck. Please use jUnit like all other test we have.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
```diff
@@ -0,0 +1,134 @@
...
+        keepAll() andThen discardAll()
+      )
+
+    lazy val checkedBehaviorStream: Stream[CheckedBehavior] = Stream(
+      CheckedBehavior.Compliant,
+      CheckedBehavior.Fatal,
+      CheckedBehavior.Unchecked
+    )
+
+    lazy val semanticsStream: Stream[Semantics] = {
```
gzm0, line 50: Couldn't this simply be written as:

```scala
lazy val semanticsStream = for {
  asInstanceOfs <- checkedBehaviorStream
  arrayIndexOutOfBounds <- checkedBehaviorStream
  ...
} yield {
  Semantics.Defaults
    .withAsInstanceOfs(asInstanceOfs)
    .withArrayIndexOutOfBounds(arrayIndexOutOfBounds)
    ...
}
```

Seems simpler to me (and I think actually gives us the full cross product, unlike the current code). Then again, less logic in tests is also good. So maybe just a bunch of example cases would be enough.

This comment applies in general to the tests.

sbt-plugin/src/sbt-test/incremental/change-config/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.3.10
```
gzm0, line 1: Is this deliberately different from other versions in `build.properties`?

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
MaximeKjaer: I tried the full cross product, but it takes a minute to run, which seemed a little too long. Would it be acceptable to have such a long-running test?

sbt-plugin/src/sbt-test/incremental/change-config/project/build.properties
MaximeKjaer: No, there's actually no reason for this file, it's an artifact from having tried to run sbt manually on the command line (which doesn't work). Fixed in 05d8a6791c30eeeb4c4a3f4a01447e71b6ac387a.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
gzm0: No, I don't think so. Especially for such a simple thing.

TBH, I'm not sure it is worth trying to make the tests cover every field. Mainly because we cannot write a more general test for this. What I mean by that is: If we add a new field somewhere and we forget the put it in the fingerprint, we'll very likely also forget to put it in the test. So from this POV, exhaustive field testing seems overkill anyways.

What we should test is that the `FingerprintBuilder` works properly (and is wired somewhat reasonably). But that we can test with 2-3 examples. This might be better than complicated logic.

Since it seems that we are sharing our internal documentation, this explains why:
https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

MaximeKjaer: Thank you, that was a fun read :) I've simplified the `StandardConfig` fingerprint tests to just run a few examples in 27a213e1c7ee420d3bd49913d45a2bbef425cdd3, and I have also added a few tests directly for the `FingerprintBuilder` in d0ad254ba744ba53f654e16763461483aed53822.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
```diff
@@ -87,4 +89,38 @@ object ModuleInitializer {
         MethodName(mainMethodName, ArrayOfStringTypeRef :: Nil, VoidRef),
         args)
   }
+
+  private implicit object MethodNameFingerprint
+      extends Fingerprint[MethodName] {
+
+    override def fingerprint(methodName: MethodName): String =
+      methodName.simpleName.encoded.toString()
```
gzm0, line 97: Hum, this is plain wrong. A `MethodName` is not identified just by its simple name but by all of its members.

All of these should use `nameString` (which is defined and correct for both `ClassName` and `MethodName`).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -178,6 +179,43 @@ object Semantics {
...
+        mapper match {
+          case KeepAll => "KeepAll"
+          case DiscardAll => "DiscardAll"
+          case RegexReplace(pattern, flags, replacement) =>
+            new FingerprintBuilder("RegexReplace")
+              .addField("pattern", pattern)
+              .addField("flags", flags)
+              .addField("replacement", replacement)
+              .build()
+          case AndThen(first, second) =>
```
gzm0, line 196: I think there needs to be a blank line before this. Also in other spots.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -158,6 +160,35 @@ final class StandardConfig private (
...
+      new FingerprintBuilder("StandardConfig")
+        .addField("semantics", config.semantics)
+        .addField("moduleKind", config.moduleKind)
+        .addField("esFeatures", config.esFeatures)
+        .addField("checkIR", config.checkIR)
+        .addField("optimizer", config.optimizer)
+        .addField("parallel", config.parallel)
+        .addField("sourceMap", config.sourceMap)
+        .addField(
+          "relativizeSourceMapBase",
```
gzm0, line 176: 4 spaces indent, probably put string on same line as `addField`.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
```diff
@@ -0,0 +1,50 @@
...
+    val fingerprint = new FingerprintBuilder("Test1")
+      .addField("test1", true)
+      .build()
+    assertEquals("Test1(test1=true)", fingerprint)
+  }
+
+  @Test
+  def multipleFields(): Unit = {
+    val test1: Option[Int] = Some(1)
+    val test2: Option[Int] = None
```
gzm0, line 38: I think it is easier to read if you type ascribe these at call site (if it is necessary to make implicit resolution work). E.g. `addField("test1", Some(1): Option[Int])`.

Otherwise this leaves me wondering why these are vals.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
```diff
@@ -0,0 +1,50 @@
...
+    val fingerprint = new FingerprintBuilder("Test2")
+      .addField("test1", test1)
+      .addField("test2", test2)
+      .addField("test3", List.empty[Int])
+      .addField("test4", List(1, 2, 3))
+      .build()
+    assertEquals(
+      "Test2(test1=Some(1),test2=None,test3=List(),test4=List(1,2,3))",
```
gzm0, line 46: Indent 4.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
```diff
@@ -0,0 +1,66 @@
...
+
+class StandardConfigFingerprintTest {
+  @Test
+  def noFingerprintCollisionCheckIR(): Unit = {
+    val sc1 = StandardConfig().withCheckIR(false)
+    val sc2 = StandardConfig().withCheckIR(true)
+    assertNotEquals(
+      StandardConfig.fingerprint(sc1),
+      StandardConfig.fingerprint(sc2)
+    )
```
gzm0, line 30: Consider a helper for this. Also indent 4 otherwise.

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Test
+  def noFingerprintCollisionRuntimeClassNameMapper(): Unit = {
+    val sc1 = StandardConfig().withSemantics(
+      _.withRuntimeClassNameMapper(keepAll() andThen discardAll())
+    )
```
gzm0, line 57: Indent 4, closing parens on previous line.

project/Build.scala
```diff
@@ -709,6 +709,7 @@ object Build {
       mimaBinaryIssueFilters ++= BinaryIncompatibilities.LinkerInterface,
       exportJars := true, // required so ScalaDoc linking works
 
+      libraryDependencies += "com.novocode" % "junit-interface" % "0.11" % "test",
```
gzm0, line 712: That seems wrong. This is a JVM dependency but it goes into the common settings.

Jenkinsfile
```diff
@@ -311,7 +311,7 @@ def Tasks = [
   "tools": '''
     setJavaVersion $java
     npm install &&
-    sbt ++$scala ir$v/test logging$v/compile linkerInterface$v/compile \
+    sbt ++$scala ir$v/test logging$v/compile linkerInterface$v/test \
```
gzm0, line 314: The linkerInterface tests also need to run in JS. (look for `linkerInterfaceJS$v`)

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/ModuleInitializerFingerprintTest.scala
```diff
@@ -0,0 +1,41 @@
...
+    genMainMethod ++ genMainMethodWithArgs
+  }
+
+  @Test
+  def noFingerprintCollision(): Unit = {
+    for ((a, b) <- moduleInitializers.zip(moduleInitializers.tail)) {
+      assertNotEquals(ModuleInitializer.fingerprint(a), ModuleInitializer.fingerprint(b))
+    }
+  }
+}
```
gzm0, line 41: Consider simplifying this test as well for consistency.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
```diff
@@ -0,0 +1,67 @@
...
+
+private[interface] trait Fingerprint[T] {
+  def fingerprint(obj: T): String
+}
+
+private[interface] object Fingerprint {
+  def fingerprint[T: Fingerprint](obj: T): String =
+    implicitly[Fingerprint[T]].fingerprint(obj)
+
+  class FingerprintBuilder(name: String) {
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 22126ae0a189b4e3b1997b36365a266f418ebbec

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
MaximeKjaer: Fixed in dcb1a8ac1fa008d993c2fc107bbef72a1230688d

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
MaximeKjaer: Fixed in eb6e308689224c3bbcc9c059e10fa1f02bee3acb

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
MaximeKjaer: Yes, those are indeed needed for implicit resolution. Fixed in eb6e308689224c3bbcc9c059e10fa1f02bee3acb

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
MaximeKjaer: Fixed in eb6e308689224c3bbcc9c059e10fa1f02bee3acb

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
MaximeKjaer: Fixed in eb6e308689224c3bbcc9c059e10fa1f02bee3acb

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
MaximeKjaer: Indentation is fixed in eb6e308689224c3bbcc9c059e10fa1f02bee3acb.

I haven't fixed the closing parens yet, because I'm a little unclear on what the coding style guideline is here. Is it that things with varargs are allowed to have a closing parens on a single line, but regular functions are not?

```scala
// This happens often in the Scala.js codebase:
val x = Seq(
    thingNumberOne,
    thingNumberTwo,
    // ...
    thingNumberN
)
```

The code style guide is quite strict on where to break.

> A line can be broken after either a `,` or a `(`, or possibly after a binary operator in a long expression.

This wouldn't really allow the above snippet, or long chained calls (which are usually broken on `)`, before `.`).

project/Build.scala
MaximeKjaer: Fixed in b58a8673c6b5bd196a2e9af6d86353b59a91ec12. Is this the correct way to do it?

Jenkinsfile
MaximeKjaer: Fixed in b58a8673c6b5bd196a2e9af6d86353b59a91ec12

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/ModuleInitializerFingerprintTest.scala
MaximeKjaer: Fixed in 2749f3ee1fd7cf9124217651c7a26130739fcd1d

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
sjrd: Yes, it's allowed for "collection-like" constructors like that. Typically things with varargs, indeed, can have the closing paren in a separate line (in that case, the first argument should already be on a separate line than the opening `(`.

Jenkinsfile
```diff
@@ -311,7 +311,8 @@ def Tasks = [
   "tools": '''
     setJavaVersion $java
     npm install &&
-    sbt ++$scala ir$v/test logging$v/compile linkerInterface$v/compile \
+    sbt ++$scala ir$v/test logging$v/compile \
+        linkerInterface$v/test linkerInterfaceJS$v/test \
```
gzm0, line 315: This should not be here. But in the `bootstrap` task. (the `testInterfaceJS$v/compile` was implicit in the mima check there, sorry; didn't realize that).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
```diff
@@ -0,0 +1,67 @@
...
+          .fold("None")(f => s"Some($f)")
+      }
+    }
+  }
+
+  implicit def listFingerprint[T: Fingerprint]: Fingerprint[List[T]] = {
+    new Fingerprint[List[T]] {
+      override def fingerprint(list: List[T]): String = {
+        list
+          .map(implicitly[Fingerprint[T]].fingerprint)
```
gzm0, line 62: Just saw this: Simply `.map(fingerprint(_))`?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
```diff
@@ -0,0 +1,67 @@
...
+  implicit object StringFingerprint extends Fingerprint[String] {
+    override def fingerprint(str: String): String = str
+  }
+
+  implicit def optionFingerprint[T: Fingerprint]: Fingerprint[Option[T]] = {
+    new Fingerprint[Option[T]] {
+      override def fingerprint(opt: Option[T]): String = {
+        opt
+          .map(implicitly[Fingerprint[T]].fingerprint)
+          .fold("None")(f => s"Some($f)")
```
gzm0, line 53: Also just saw this: probably put everything in the `fold`? (or even use a "top-level" match):

```scala
def ...  = opt match {
  case Some(f) => s"Some(${fingerprint(f)})"
  case None    => "None"
}
```

gzm0:
I'll leave it up to @sjrd to answer the style questions :) He's the guardian of the style after all. Afterwards, good to squash for me.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
MaximeKjaer: I like the top-level match. Fixed in d3afbdc3ea834aaf3bc6b604ff5f9a99789d1970

Jenkinsfile
MaximeKjaer: Ah, I see, all JS tasks go in `bootstrap` and JVM tasks in `tools`. Fixed in 0c9e543c35dc32a34681040579c1afd36fa18e9b

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
MaximeKjaer: I'm not sure I follow. How would this type-check? The `fingerprint` I need to apply to list items is the one for `T`, not `List[T]`. 

Would you prefer the alternative notation that takes an `implicit itemFingerprint: Fingerprint[T]` instead of a context bound `[T: Fingerprint]` and `implicitly[Fingerprint[T]]`?

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/StandardConfigFingerprintTest.scala
MaximeKjaer: Okay, I see. Thank you for clearing this up :)

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Fingerprint.scala
gzm0: Ah, ah, I see the problem: I was referring to the top-level fingerprint that merely calls `fingerprint` on its own implicit parameter. But because of the same name, that of course doesn't work, because `fingerprint` refers to the `def` itself. It's just a nuisance anyways. So maybe fine to leave it as such.

MaximeKjaer:
I suggest the following commit message:

> **Fix incremental linking on configuration changes**
> 
> The output of the linker depends on both source code and linker 
> settings; a change in either should cause linking to happen again.
> 
> Previously, the sbt plugin would only do an incremental re-linking when
> the source had changed; it would not re-link if only the config had
> changed.
> 
> This commit adds methods to the linker interface to "fingerprint"
> linker configurations. A fingerprint is a one-way, injective
> serialization. Two fingerprints can be compared by equality to infer
> the equality comparison between the two fingerprinted objects.
> 
> The sbt plugin writes fingerprints of linker configuration keys to disk
> at every run of `fastOptJS` and `fullOptJS`. It then compares the
> fingerprints of the current configuration objects to those saved to
> disk on the previous run in order to determine whether it should re-run
> linking.
> 
> If both source and linker config have changed, the re-linking only
> happens once.
> 
> Fixes #3283

If you're okay with this, I can rename the PR to `Fix incremental linking on configuration changes (#3283)`.

Would you like me to squash the commits, or will you do that when merging?

sjrd:
Very good commit message. Just one change to follow our format: include `Fix #3283: ` in the commit title, instead of as a last paragraph of the description. So the commit (and PR) title could be

    Fix #3283: Fix incremental linking on configuration changes.

gzm0:
> If you're okay with this, I can rename the PR to Fix incremental linking on configuration changes (#3283).

Have a look [here][1] (just when you thought we're done being pedantic :P 

> Would you like me to squash the commits, or will you do that when merging?

Please squash them, but maybe wait for @sjrd to review so he has history (if he cares).

> A fingerprint is a one-way, injective serialization. Two fingerprints can be compared by equality to infer the equality comparison between the two fingerprinted objects.

While correct for a single version of Scala.js. I do not think we should commit to this in a backwards compatible fashion. This point is rather subtle. What I mean is that a previously generated fingerprint *may* be different even for the (semantically) same config. This will happen in practice if we add a new field and could happen if we chose to change the representation (for whatever reason).

It might make sense to document that on the `fingerprint` method itself (it's a property of the method, not of the change, if you want so).

[1]: https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages

sjrd:
> Please squash them, but maybe wait for @sjrd to review so he has history (if he cares).

I've been quietly following the changes. No need to wait for me to squash.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleInitializer.scala
```diff
@@ -87,4 +89,39 @@ object ModuleInitializer {
...
+              .addField("className", className)
+              .addField("encodedMainMethodName", encodedMainMethodName)
+              .build()
+
+        case MainMethodWithArgs(className, encodedMainMethodName, args) =>
+          new FingerprintBuilder("MainMethodWithArgs")
+              .addField("className", className)
+              .addField("encodedMainMethodName", encodedMainMethodName)
+              .addField("args", args)
+              .build()
```
sjrd, line 121:
```suggestion
          new FingerprintBuilder("MainMethodWithArgs")
            .addField("className", className)
            .addField("encodedMainMethodName", encodedMainMethodName)
            .addField("args", args)
            .build()
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleKind.scala
```diff
@@ -49,4 +49,15 @@ object ModuleKind {
...
   case object CommonJSModule extends ModuleKind
 
+  private[interface] implicit object ModuleKindFingerprint
+      extends Fingerprint[ModuleKind] {
+
+    override def fingerprint(moduleKind: ModuleKind): String = {
+      moduleKind match {
+        case NoModule => "NoModule"
+        case ESModule => "ESModule"
+        case CommonJSModule => "CommonJSModule"
```
sjrd, line 59:
```suggestion
        case NoModule       => "NoModule"
        case ESModule       => "ESModule"
        case CommonJSModule => "CommonJSModule"
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -178,6 +179,45 @@ object Semantics {
...
+        mapper match {
+          case KeepAll => "KeepAll"
+          case DiscardAll => "DiscardAll"
+
+          case RegexReplace(pattern, flags, replacement) =>
+            new FingerprintBuilder("RegexReplace")
+                .addField("pattern", pattern)
+                .addField("flags", flags)
+                .addField("replacement", replacement)
+                .build()
```
sjrd, line 196:
```suggestion
            new FingerprintBuilder("RegexReplace")
              .addField("pattern", pattern)
              .addField("flags", flags)
              .addField("replacement", replacement)
              .build()
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -178,6 +179,45 @@ object Semantics {
...
       regexReplace(regex.pattern, replacement)
     }
+
+    private[interface] implicit object RuntimeClassNameMapperFingerprint
+        extends Fingerprint[RuntimeClassNameMapper] {
+
+      override def fingerprint(mapper: RuntimeClassNameMapper): String = {
+        mapper match {
+          case KeepAll => "KeepAll"
+          case DiscardAll => "DiscardAll"
```
sjrd, line 189:
```suggestion
          case KeepAll    => "KeepAll"
          case DiscardAll => "DiscardAll"
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -178,6 +179,45 @@ object Semantics {
...
+                .addField("pattern", pattern)
+                .addField("flags", flags)
+                .addField("replacement", replacement)
+                .build()
+
+          case AndThen(first, second) =>
+            new FingerprintBuilder("AndThen")
+                .addField("first", fingerprint(first))
+                .addField("second", fingerprint(second))
+                .build()
```
sjrd, line 202:
```suggestion
            new FingerprintBuilder("AndThen")
              .addField("first", fingerprint(first))
              .addField("second", fingerprint(second))
              .build()
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala
```diff
@@ -178,6 +179,45 @@ object Semantics {
...
+
+    override def fingerprint(semantics: Semantics): String = {
+      new FingerprintBuilder("Semantics")
+          .addField("asInstanceOfs", semantics.asInstanceOfs)
+          .addField("arrayIndexOutOfBounds", semantics.arrayIndexOutOfBounds)
+          .addField("moduleInit", semantics.moduleInit)
+          .addField("strictFloats", semantics.strictFloats)
+          .addField("productionMode", semantics.productionMode)
+          .addField("runtimeClassNameMapper", semantics.runtimeClassNameMapper)
+          .build()
```
sjrd, line 219:
```suggestion
      new FingerprintBuilder("Semantics")
        .addField("asInstanceOfs", semantics.asInstanceOfs)
        .addField("arrayIndexOutOfBounds", semantics.arrayIndexOutOfBounds)
        .addField("moduleInit", semantics.moduleInit)
        .addField("strictFloats", semantics.strictFloats)
        .addField("productionMode", semantics.productionMode)
        .addField("runtimeClassNameMapper", semantics.runtimeClassNameMapper)
        .build()
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala
```diff
@@ -158,6 +160,31 @@ final class StandardConfig private (
...
+          .addField("optimizer", config.optimizer)
+          .addField("parallel", config.parallel)
+          .addField("sourceMap", config.sourceMap)
+          .addField("relativizeSourceMapBase",
+              config.relativizeSourceMapBase.map(_.toASCIIString()))
+          .addField("closureCompilerIfAvailable",
+              config.closureCompilerIfAvailable)
+          .addField("prettyPrint", config.prettyPrint)
+          .addField("batchMode", config.batchMode)
+          .build()
```
sjrd, line 181:
```suggestion
      new FingerprintBuilder("StandardConfig")
        .addField("semantics", config.semantics)
        .addField("moduleKind", config.moduleKind)
        .addField("esFeatures", config.esFeatures)
        .addField("checkIR", config.checkIR)
        .addField("optimizer", config.optimizer)
        .addField("parallel", config.parallel)
        .addField("sourceMap", config.sourceMap)
        .addField("relativizeSourceMapBase",
            config.relativizeSourceMapBase.map(_.toASCIIString()))
        .addField("closureCompilerIfAvailable",
            config.closureCompilerIfAvailable)
        .addField("prettyPrint", config.prettyPrint)
        .addField("batchMode", config.batchMode)
        .build()
```

linker-interface/shared/src/test/scala/org/scalajs/linker/interface/FingerprintBuilderTest.scala
```diff
@@ -0,0 +1,47 @@
...
+  def noFields(): Unit = {
+    val fingerprint = new FingerprintBuilder("Test0").build()
+    assertEquals("Test0()", fingerprint)
+  }
+
+  @Test
+  def oneField(): Unit = {
+    val fingerprint = new FingerprintBuilder("Test1")
+        .addField("test1", true)
+        .build()
```
sjrd, line 31: More like that ;)

MaximeKjaer:
Thank you! I've amended the commit to fix these last stylistic issues, and have removed the extra reference to the issue in the commit body :)

sjrd:
`incremental / change-config` failed in the CI. Any idea why?

MaximeKjaer:
I've looked into it, but I don't know why that happened.

Actually, [run 24 on Jenkins](https://scala-webapps.epfl.ch/sjsjenkins/blue/organizations/jenkins/scalajs-main/detail/PR-3987/24/pipeline) should not have passed, because I had broken the `FingerprintBuilderTest` in d3afbdc3ea834aaf3bc6b604ff5f9a99789d1970 by forgetting to add a closing paren to the `Some` case. But [`linkerInterface$v/test` was in the Jenkinsfile at that point](https://github.com/scala-js/scala-js/blob/d3afbdc3ea834aaf3bc6b604ff5f9a99789d1970/Jenkinsfile#L315), so the test should have run.

In any case, I've fixed that issue in ffbf9ab9b3e238b6cb36f7d86d0a49ee59271b6f. I'm not able to reproduce the `NullPointerException` that happened on the CI, so hopefully this commit will pass.

If it doesn't, there may be something wrong with the `Jenkinsfile`.

MaximeKjaer:
> Thank you @MaximeKjaer for sticking with this! It took a while, but this will be very useful! :tada:

Thank you to both of you for reviewing the many iterations of this PR!

sjrd:
The job on Jenkins is green, but the status was not reported to GitHub. Merging now :)

gzm0:
Yeah!!!

PR #3990, opened 2020-03-10
Fix #3670: Remove workaround for GCC bug creating duplicate fields

### LINKED ISSUE #3670: Investigate whether GCC still recreates object literals with duplicate key names
See https://github.com/scala-js/scala-js/pull/3579#discussion_r287554302

gzm0:
I have manually verified that there are no bad object literals in the output via the generated code of `DynamicTest.should_allow_object_literals_to_have_duplicate_keys_issue_1595`. There are only `{bar:5,foo:6}` now.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
@@ -1743,23 +1743,12 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
             val objVarDef =
               genLet(objVarIdent, mutable = false, js.ObjectConstr(Nil))
 
-            val assignFields = fields.foldRight((Set.empty[String], List.empty[Tree])) {
-              case ((key, value), (namesSeen, statsAcc)) =>
-                implicit val pos = value.pos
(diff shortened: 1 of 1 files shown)


sjrd:
Did you manually check that GCC is not recreating a bogus object, especially with ECMAScript 5.1? (with ES 2015 it's valid to have duplicate fields in an object literal)

gzm0:
Isn't this test enough for that: https://github.com/scala-js/scala-js/blob/v1.0.1/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/DynamicTest.scala#L291-L317?

gzm0:
I have verified the output of fullOpt manually.

sjrd:
> 
> 
> Isn't this test enough for that: https://github.com/scala-js/scala-js/blob/v1.0.1/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/DynamicTest.scala#L291-L317?

It's enough if we manage to run it on a JS engine that strictly implements ES 5.1. But this is a very weird thing because it wasn't valid in ES 5.1 but became valid in ES 2015. So when we test on a non-obsolete Node.js version, we wouldn't detect a regression in that particular aspect, unfortunately.

PR #3992, opened 2020-03-17
add: UncaughtExceptionHandler interface to java.lang.Thread and more changes to Thread

andreaTP:
This is due to the more stringent linker checks.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Thread.scala
+++ b/javalanglib/src/main/scala/java/lang/Thread.scala
@@ -12,12 +12,10 @@
 
 package java.lang
 
-/* We need a constructor to create SingleThread in the companion object, but
- * we don't want user code doing a 'new Thread()' to link, because that could
- * be confusing.
- * So we use a binary signature that no Java source file can ever produce.
+/* We need a constructor to create SingleThread in the companion object, but:
+ * PLEASE DO NOT USE THIS CLASS IN USER CODE!
  */
-class Thread private (dummy: Unit) extends Runnable {
+class Thread extends Runnable {
   private var interruptedState = false
   private[this] var name: String = "main" // default name of the main thread
 
@@ -39,10 +37,17 @@ class Thread private (dummy: Unit) extends Runnable {
     StackTrace.getCurrentStackTrace()
 
   def getId(): scala.Long = 1
+
+  // Javadocs:
+  // Returns the handler invoked when this thread abruptly terminates due to an uncaught exception. If this thread has not had an uncaught exception handler explicitly set then this thread's ThreadGroup object is returned, unless this thread has terminated, in which case null is returned.
+  // Therefore:
+  // JS VMs are running on a single thread, if this thread has been terminated it has to be this one, so, since it's terminated, we return null
+  def getUncaughtExceptionHandler(): Thread.UncaughtExceptionHandler = null
+
 }
 
 object Thread {
-  private[this] val SingleThread = new Thread(())
+  private[this] val SingleThread = new Thread()
 
   def currentThread(): Thread = SingleThread
 
--- a/javalanglib/src/main/scala/java/lang/Thread.scala
+++ b/javalanglib/src/main/scala/java/lang/Thread.scala
@@ -51,4 +51,8 @@ object Thread {
     currentThread.interruptedState = false
     ret
   }
+
+  trait UncaughtExceptionHandler {
+    def uncaughtException(t: Thread, e: Throwable): Unit
+  }
 }


sjrd:
Also could you make the commit message follow our usual commit messages for new JDK APIs, like this:

    Implement java.lang.Thread.UncaughtExceptionHandler.

Thank you!

andreaTP:
Make sense, I'll do that.
Unfortunately I have had to do some more modifications to `Thread` and I understand that they can be "opinionated" ... 

This is my use-case:
https://github.com/akka-js/akka.js/blob/441c60a978227916d593880de9cd89c53e9b1bb3/akka-js-actor/js/src/main/scala/akka/dispatch/ThreadPoolBuilder.scala#L45

Basically to provide an alternative `ThreadFactory` I need to spawn `Thread` that are mocked by simple `Runnable` at the end of the day.
This enable me to keep the original signatures and comply with the interfaces.

andreaTP:
On the bright side the heavy lifting is done and all the test suites are already passing:
https://travis-ci.org/github/akka-js/akka.js/jobs/663647282
with Akka.Js on Scala.Js 1.0.1
Amazing work, the transition was smooth even for a project full of hacks and edge cases like Akka.Js.

javalanglib/src/main/scala/java/lang/Thread.scala
```diff
@@ -39,10 +37,17 @@ class Thread private (dummy: Unit) extends Runnable {
     StackTrace.getCurrentStackTrace()
 
   def getId(): scala.Long = 1
+
+  // Javadocs:
+  // Returns the handler invoked when this thread abruptly terminates due to an uncaught exception. If this thread has not had an uncaught exception handler explicitly set then this thread's ThreadGroup object is returned, unless this thread has terminated, in which case null is returned.
+  // Therefore:
+  // JS VMs are running on a single thread, if this thread has been terminated it has to be this one, so, since it's terminated, we return null
+  def getUncaughtExceptionHandler(): Thread.UncaughtExceptionHandler = null
```
sjrd: I don't think this is correct. Even with a single thread, while the thread is running, `getUncaughtExceptionHandler()` already returns an instance of `ThreadGroup`:
```scala
>scala
Welcome to Scala 2.13.1 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60).
Type in expressions for evaluation. Or try :help.

scala> val t = Thread.currentThread
t: Thread = Thread[main,5,main]

scala> t.getUncaughtExceptionHandler
res6: java.lang.Thread.UncaughtExceptionHandler = java.lang.ThreadGroup[name=main,maxpri=10]
```

javalanglib/src/main/scala/java/lang/Thread.scala
```diff
@@ -12,12 +12,10 @@
...
 
-/* We need a constructor to create SingleThread in the companion object, but
- * we don't want user code doing a 'new Thread()' to link, because that could
- * be confusing.
- * So we use a binary signature that no Java source file can ever produce.
+/* We need a constructor to create SingleThread in the companion object, but:
+ * PLEASE DO NOT USE THIS CLASS IN USER CODE!
  */
-class Thread private (dummy: Unit) extends Runnable {
+class Thread extends Runnable {
```
sjrd: I'm afraid this change does not fit in our global policy of "if it links, it is correct". Therefore we cannot accept this change in the core repo.

andreaTP: I understand your reason, in defense of this change I would like to say that the current signature is not respecting the contract in the JavaDoc (hence the comment above):
https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#Thread()
that allow the construction of the Thread object.

A program that implements `Thread` in a way that is supported by the runtime *is* correct in my opinion, as an example a user can leverage GraalVM and provide support for multi-threading for the specific target; I think that being forced to use alternative classes is a stretch.

This said I "can" keep this file in the Akka.Js codebase (despite the fact that I'm going to provide the users a bad experience in this sense e.g. linking will succeed or fail depending on your dependencies) but I'm not able to access sibling classes therefore this hack:
https://github.com/akka-js/akka.js/blob/sjs1Support/akka-js-actor/js/src/main/scala/java/lang/Thread.scala#L38-L40

Is possibly going to have deeper consequences, can you have a look @sjrd ?

andreaTP: Another approach that I would like to explore with you is to make the `Thread` constructor package `protected`.
This way:
 - normal users are not allowed to instantiate it
 - edge cases and/or ppl that are aware of the limitations etc. etc. are able to workaround

sjrd: The problem with all these solutions is always the same: it makes it possible to link a program that *doesn't* use any weird stuff, make it link, and then expect it to work but it won't! This goes against our policy. It doesn't matter that *if* you carefully override all the things to make it work *then* you have a correct program.

The existing constructor of `java.lang.Thread` is `private` and intentionally uses a different signature *so that it won't link*, because it is not correct in general. It is only correct to the main and only thread of a JS environment.

I am afraid that, by trying too hard to reuse `java.lang.Thread` as an interface, you *are* looking for trouble because you try to repurpose JDK APIs with different specifications. This goes against the policy of Scala.js and our implementation of the JDK. Instead, you should abstract these things away as a separate interface, so that there are contractually valid implementations of *that* interface on both platforms.

andreaTP: I double checked and I should be able to keep this class in akka.js codebase.
Just need to confirm with you that: `thread.getStackTrace` is not used for functional reasons by Scala.Js itself (at a first look it doesn't looks like).

sjrd: It's only used by `thread.printStackTrace`.

andreaTP: Fine, so I decline this PR

PR #3993, opened 2020-03-17
Implement java.util.concurrent.ExecutorService

andreaTP:


andreaTP:
It's a mistake indeed, basically the incremental compiler got confused and allowed me to compile by fixing only part of the required methods, here you have the complete interface and I will follow up with a test.

andreaTP:
This should be ready for final review (despite the laziness in the tests we are guaranteeing the functionality of the linker).

Side note:
Regarding the strange behavior I noticed from the linker, `ExecutorService` defined here:
https://github.com/akka-js/akka.js/tree/sjs1Support/akka-js-actor/js/src/main/scala/java/util/concurrent
links without having `java.util.concurrent.Future`
reproducible in CI:
https://travis-ci.org/github/akka-js/akka.js/builds/663647262

Should I file a separate issue?

sjrd:
Ouch, I'm sorry, I hadn't noticed before that `ExecutorService` defines methods that return `java.util.concurrent.Future`s. 😕 That's very problematic because the API of `Future` is inherently blocking, which means that no one can ever correctly implement that API, and consequently no one can correctly implement `ExecutorService`.

So there seems to be no point in adding this to the standard JDK that we implement. It's just noise.

I guess you're going to tell me that you made an implementation that throws or immediately returns or something like that, but again, that's not correct. It's a hack and we don't add support for hacks.

---

Regarding the unexpected successful linking, that can happen if you do not actually call any of the methods that mention `Future` in their signature. In that case, they will be eliminated before there is any need to load the definition of `Future`, and therefore linking succeeds.

This is not a bug. In fact it is like that by design, otherwise we wouldn't be able to link the collections of Scala 2.13, given that they mention Java 8 `Stream` APIs in some of their methods.

andreaTP:
Turns out that I can comment all the methods that have `Fututre` in the signature.
Would you accept a PR with all the rest?

sjrd:
I don't think that will work. You'll still have to implement them in any implementation of the trait, because the implementation will be validated against the JDK signatures. That's why abstract methods must always all be defined. There's no way not to define them.

andreaTP:
This means that no one will ever contribute those interfaces to scala-js core, hence I don't risk conflicts  from Akka.Js

PR #3994, opened 2020-03-17
add: Minimal implementation j.u.c.CompletionStage

andreaTP:


andreaTP:
The only used method in Akka is `whenComplete` but if you feel we should add all of the declared methods I can do it ...

sjrd:
Yes, I'm afraid it must be all or nothing for abstract methods. Otherwise other users of that interface could get bogus behavior.

It's OK to motivate the addition of a JDK interface because Akka needs it, but that's no excuse for lowering our standards of quality, which are "if it links, it's correct".

andreaTP:
Too much work for so little gain, sorry I'm not able to tackle this at this point.

PR #3995, opened 2020-03-21
Only export relevant projects to IDEs.

sjrd:
We detect whether bloop is on the classpath (which will be the case during import in Metals) and if yes, we deactivate the bloop export for irrelevant projects.

Notably, we deactivate all non-2.12 projects in `MultiScalaProject`, since it is enough to have IDE support for the default Scala version.

### DIFF
--- a/project/MultiScalaProject.scala
+++ b/project/MultiScalaProject.scala
@@ -83,15 +83,22 @@ object MultiScalaProject {
     "2.13" -> Seq("2.13.0", "2.13.1"),
   )
 
+  private final val ideVersion = "2.12"
+
   private def projectID(id: String, major: String) = id + major.replace('.', '_')
 
   def apply(id: String, base: File): MultiScalaProject = {
     val projects = for {
       (major, minors) <- versions
     } yield {
+      val noIDEExportSettings =
+        if (major == ideVersion) Nil
+        else NoIDEExport.noIDEExportSettings
+
       major -> Project(id = projectID(id, major), base = new File(base, "." + major)).settings(
         scalaVersion := minors.last,
         crossScalaVersions := minors,
+        noIDEExportSettings,
       )
     }
 
--- a/project/NoIDEExport.scala
+++ b/project/NoIDEExport.scala
@@ -1,0 +1,41 @@
+package build
+
+import sbt._
+import Keys._
+
+/** Settings to prevent projects from being exported to IDEs. */
+object NoIDEExport {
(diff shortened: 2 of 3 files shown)


project/MultiScalaProject.scala
```diff
@@ -89,9 +89,14 @@ object MultiScalaProject {
     val projects = for {
       (major, minors) <- versions
     } yield {
+      val noIDEExportSettings =
+        if (major == "2.12") Nil
```
gzm0: Make this a constant `ideVersion` next to `versions`?

project/NoIDEExport.scala
```diff
@@ -0,0 +1,31 @@
+package build
+
+import sbt._
+import Keys._
+
+object NoIDEExport {
+  lazy val bloopGenerateKey: Option[TaskKey[Option[File]]] = {
```
gzm0: private?

project/NoIDEExport.scala
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    optBloopKeysClass.map { bloopKeysClass =>
+      val bloopGenerateGetter = bloopKeysClass.getMethod("bloopGenerate")
+      bloopGenerateGetter.invoke(null).asInstanceOf[TaskKey[Option[File]]]
+    }
+  }
+
+  lazy val noIDEExportSettings: Seq[Setting[_]] = {
+    bloopGenerateKey match {
```
gzm0: `.toSeq.flatMap { key => ` ...?

gzm0: I think this deserves a comment that bloop is metals (pretty much the comment in the commit message will do IMO).

project/NoIDEExport.scala
```diff
@@ -0,0 +1,31 @@
...
+      case _: ClassNotFoundException => None
+    }
+
+    optBloopKeysClass.map { bloopKeysClass =>
+      val bloopGenerateGetter = bloopKeysClass.getMethod("bloopGenerate")
+      bloopGenerateGetter.invoke(null).asInstanceOf[TaskKey[Option[File]]]
+    }
+  }
+
+  lazy val noIDEExportSettings: Seq[Setting[_]] = {
```
gzm0: Consider giving this a short doc description. As in: "Settings to prevent exporting a project to (supported) IDEs".

project/NoIDEExport.scala
sjrd: I don't find that easier to read, TBH.

PR #3996, opened 2020-03-22
Introduced support SecureRandom for Scala.js

catap:
This commit implemented a simple wrapper around `Crypto.getRandomValues()` and `java.util.Random` that covers all API for jdk13.

### DIFF
--- a/javalib/src/main/scala/java/security/Provider.scala
+++ b/javalib/src/main/scala/java/security/Provider.scala
@@ -1,0 +1,15 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package java.security
+
+trait Provider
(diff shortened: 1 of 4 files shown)


catap:
Signed CLA

sjrd:
Hello @catap. Thank you for your interest in contributing to Scala.js!

Unfortunately, your approach in implementing `SecureRandom`, while it might be done in a separate repo/library, cannot be used in the core Scala.js repository.

The reason is that it requires some browser-specific APIs. The core Scala.js repo is committed to implement everything in a way that is portable across all JavaScript engines supporting ES 5.1 strict or ES 2015. This is much broader than browsers, so we can never use browser-specific APIs.

catap:
@sjrd I see your point and I've agree with it.

I can try to use approach similar to GPG4Browsers / OpenPGP.js https://github.com/openpgpjs/openpgpjs/blob/master/src/crypto/random.js but I feel that it isn't good :(

sjrd:
Unfortunately, even that alternative approach cannot be used here, because it would only work in browsers and Node.js. There are other JS environments out there ;)

Basically, anything that relies on an API that is not specified in ECMAScript itself will be off-limit.

That does not mean that it is a complete dead end. It would be acceptable to publish a separate, non-`org.scala-js`-branded library, that would implement those APIs on a limited set of platforms.

catap:
@sjrd their approach has downstream to just a random buffer, but I agree that it isn't secure.

PR #3999, opened 2020-03-24
Fix #3998: Allow self types in JS types

### LINKED ISSUE #3998: Self-types in non-native JS traits cause confusing error message
```scala
trait Foo extends js.Object { self => }
```

Causes the following error (without position!).

```
 Members of non-native JS traits must either be abstract, or their right-hand-side must be `js.undefined`.
```

gzm0:


gzm0:
I'm not sure this is the best way of tackling this... But it seems to work :-/

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ScalaJSDefinedTest.scala
```diff
@@ -1977,4 +1977,5 @@ object ScalaJSDefinedTest {
     final val MethodName = "myMethod"
   }
 
+  trait SelfTypeTest extends js.Object { self => }
```
sjrd: Perhaps consider the following alternatives as well:
```scala
trait SelfTypeTest2 extends js.Object { self: js.Date => }
trait SelfTypeTest3 extends js.Object { this: js.Date => }
```

PR #4003, opened 2020-03-30
Remove the override for RedBlackTree.scala in 2.12.2+.

sjrd:
It hasn't been necessary since 2.12.2, because our fix was merged upstream in that version.

This commit is an update of 36ea85c34e647a18894154df852ac774976daada which had already removed the override for 2.13.x.

This change is necessary now because 2.12.12+ has a completely different implementation, backported from 2.13.x in the upstream pull requests https://github.com/scala/scala/pull/8749, which is not compatible with the classes that use `RedBlackTree`.

---

This was discovered by the nightly-against-nightly build.


sjrd:
Fixed the commit message. I launched a `full` build for the PR instead of `auto`, so that we can check that everything goes well in all minor versions of 2.12.x.

sjrd:
There was only one 2.10.2 job that timed out. For a full build, that's green enough.

PR #4008, opened 2020-04-08
Move initialization of L0 to CoreJSLib

gzm0:
This brings it closer to its counterpart and allows us to leverage existing CoreJSLib infrastructure for tree generation.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
@@ -29,8 +29,22 @@ import EmitterNames._
 
 private[emitter] object CoreJSLib {
 
-  def build(jsGen: JSGen, globalKnowledge: GlobalKnowledge): WithGlobals[Tree] =
+  def build(jsGen: JSGen, globalKnowledge: GlobalKnowledge): WithGlobals[Lib] =
     new CoreJSLibBuilder(jsGen)(globalKnowledge).build()
+
+  /** A fully built CoreJSLib
+   *
+   *  @param definitions The bulk of the CoreJSLib.
+   *      Definitions that do not depend on any other Scala.js emitted code
+   *      (notably RuntimeLong). These must be available to all Scala.js emitted
+   *      code.
+   *
+   *  @param initialization Things that depend on Scala.js generated classes.
+   *      These must have class definitions (but not static fields) available.
+   */
+  final class Lib private[CoreJSLib] (
+      val definitions: Tree,
+      val initialization: Tree)
 
   private class CoreJSLibBuilder(jsGen: JSGen)(
       implicit globalKnowledge: GlobalKnowledge) {
@@ -71,7 +85,12 @@ private[emitter] object CoreJSLib {
           FloatRef, DoubleRef)
     }
 
-    def build(): WithGlobals[Tree] = {
+    def build(): WithGlobals[Lib] = {
+      val lib = new Lib(buildDefinitions(), buildInitializations())
+      WithGlobals(lib, trackedGlobalRefs)
+    }
+
+    private def buildDefinitions(): Tree = {
       defineLinkingInfo()
       defineJSBuiltinsSnapshotsAndPolyfills()
       declareCachedL0()
@@ -91,7 +110,11 @@ private[emitter] object CoreJSLib {
       defineAsArrayOfPrimitiveFunctions()
       definePrimitiveTypeDatas()
 
-      WithGlobals(Block(buf.result()), trackedGlobalRefs)
+      Block(buf.result())
+    }
+
+    private def buildInitializations(): Tree = {
+      assignCachedL0()
     }
 
     private def defineLinkingInfo(): Unit = {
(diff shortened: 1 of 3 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -29,9 +29,23 @@ import EmitterNames._
...
 
-  def build(jsGen: JSGen, globalKnowledge: GlobalKnowledge): WithGlobals[Tree] =
+  def build(jsGen: JSGen, globalKnowledge: GlobalKnowledge): WithGlobals[Lib] =
     new CoreJSLibBuilder(jsGen)(globalKnowledge).build()
 
+  /** A fully built CoreJSLib
+   *
+   *  @param definitions The bulk of the CoreJSLib.
+   *      Definitions that do not depend on any other Scala.js emitted code
+   *      (n.b. RuntimeLong). These must be available to all Scala.js emitted
```
sjrd, line 39: Did you mean `e.g.` instead of `n.b.`? Or did you mean "notably"?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -29,9 +29,23 @@ import EmitterNames._
...
+   *      code.
+   *
+   *  @param initializations Things that depend on Scala.js generated classes.
```
sjrd, line 42:
```suggestion
   *  @param initialization Things that depend on Scala.js generated classes.
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -146,57 +146,49 @@ final class Emitter private (config: CommonPhaseConfig,
...
-          trees ++= generatedClass.topLevelExports
-
-        for (moduleInitializer <- unit.moduleInitializers)
-          trees += classEmitter.genModuleInitializer(moduleInitializer)
-
-        WithGlobals(trees.result(), trackedGlobalRefs ++ coreJSLibTrackedGlobalRefs)
+        def classIter = generatedClasses.iterator
+
+        val treesIter: Iterator[js.Tree] = (
+            Iterator(coreJSLib.definitions) ++
```
sjrd, line 156:
```suggestion
            Iterator.single(coreJSLib.definitions) ++
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -146,57 +146,49 @@ final class Emitter private (config: CommonPhaseConfig,
...
+             *    of type Long which need $L0.
+             * 4. All static initializers, which in the worst case can observe some
+             *    "zero" state of other static field definitions, but must not
+             *    observe a *non-initialized* (undefined) state.
+             * 5. All the exports, during which some JS class creation can happen,
+             *    causing JS static initializers to run. Those also must not observe
+             *    a non-initialized state of other static fields.
+             */
+            classIter.flatMap(_.main) ++
+            Iterator(coreJSLib.initialization) ++
```
sjrd, line 175:
```suggestion
            Iterator.single(coreJSLib.initialization) ++
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -153,14 +149,41 @@ final class Emitter private (config: CommonPhaseConfig,
...
+        val WithGlobals(coreJSLib, coreJSLibTrackedGlobalRefs) =
+          state.coreJSLibCache.lib
 
-        trees += coreJSLibTree
+        trees += coreJSLib.definitions
 
         emitModuleImports(orderedClasses, trees, logger)
 
-        emitGeneratedClasses(trees, generatedClasses)
+        /* Emit all the classes, in the appropriate order:
```
sjrd: If you inline the contents of `emitGeneratedClasses`, perhaps it would be good to move this big comment at the beginning (before `trees += coreJSLib.definitions`), and explicitly list the steps of 1) emit the definitions of the core JS lib and 2) emit the module imports.

That would also be nicer in the next commit, as then the comment would apply to the entire iterator concatenation, rather than being in the middle of it.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -46,10 +46,6 @@ final class Emitter private (config: CommonPhaseConfig,
 
   private val nameGen: NameGen = new NameGen
 
-  /** Dummy KnowledgeAccessor to generate uncached trees. */
-  private val uncachedKnowledgeAccessor =
-    new knowledgeGuardian.KnowledgeAccessor {}
```
sjrd, line 51: That's nice :)

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
gzm0: I did mean notably / nota bene.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
gzm0: Good point. I have even split the comment over the multiple lines. Like this, it is much less likely to get out of sync.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
sjrd: Hum but "nota bene" doesn't mean notably. It means "Pay attention to:", whereas "notably" can be rephrased as "in particular".

gzm0: Ah, ah :) So I guess "notably" is most appropriate here? (IMHO it's going a bit far to call `RuntimeLong` an example :P).

gzm0:
Updated:
- Replaced n.b. with notably.
- Fixed comments in intermediate commit.

PR #4009, opened 2020-04-08
Fix #4001: Only import JS classes that are needed

### LINKED ISSUE #4001: Scala.js 1.0.0 unnecessary generates imports for parent classes
As mentioned in https://github.com/ScalablyTyped/Converter/issues/131 Scala.js 1.0.0 ends up including imports of superclasses unnecessarily when referencing a class. I'll copy in the example:

```scala
object Main {
  @JSImport("rsocket-websocket-client/RSocketWebSocketClient", JSImport.Default)
  @js.native
  class Parent extends js.Object

  @JSImport("rsocket-websocket-client", JSImport.Default)
  @js.native
  class Child extends Parent

  def main(args: Array[String]): Unit = 
    println(new Child)
}
```

With Scala.js 0.6, this would only import from the module `rsocket-websocket-client`, while Scala.js 1.0.0 will also import from `rsocket-websocket-client/RSocketWebSocketClient`.

This is a problem because the latter doesn't exist, so bundling won't succeed.

There are a range of reasons why Typescript definitions may end up mentioning module names which don't have a corresponding .js file at bundle-time. For this particular example the third party typings are imprecise.

It's not impossible for me to find a different encoding for re-exporting classes such that this would be safe, but it would end up quite a bit uglier.

Any chance we can revert this behaviour?

gzm0:


gzm0:
This fails locally because the native JS modules are not imported anymore (the classes are). I do not understand why.

sjrd:
Probably because `def instantiated()` does nothing when `kind == NativeJSModule`. We should add an additional case when computing `val isNativeJSClass`.

gzm0:
Hah, I missed that. Is was right there :P

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -118,4 +122,13 @@ object ModulesTest {
...
   }
+
+  // #4001 - Test that unused super-classes are not imported.
+  @js.native
+  @JSImport("non-existent", "Foo")
+  class NonExistentSuperClass extends js.Object
+
+  @js.native
+  @JSImport("string_decoder", "StringDecoder")
+  class ExistentSubClass extends NonExistentSuperClass
```
sjrd, line 133: Should we also add a test case with an `object ExistentObject extends NonExistentSuperClass`?

PR #4010, opened 2020-04-09
Prototype for Module Splitting

gzm0:
This is a relatively quick and dirty way to get module splitting working. It is intended to serve as an evaluation point to determine whether splitting in minimal files (which this setup does) is acceptable performance wise.

With this PR the test suite for 2.12 is able to run (not pass), so we should have a reasonably large code base to test on. It currently generates ~2300 modules.


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -385,7 +385,80 @@ private[emitter] final class JSGen(val semantics: Semantics,
   def genClassDataOf(className: ClassName)(implicit pos: Position): Tree =
     genClassDataOf(ClassRef(className))
 
-  def envModuleField(module: String)(implicit pos: Position): VarRef = {
+  // Compiler generated identifiers --------------------------------------------
+
+  /* We distinguish three types of compiler generated identifiers:
+   *
+   * - envVar: Things pertaining to a given class. Accessible from every chunk
```
sjrd, line 403: Should this be called `classVar` if it is associated with a class?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -385,7 +385,80 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+
+  def envVar(field: String, className: ClassName, methodName: MethodName,
+      origName: OriginalName)(
+      implicit pos: Position): VarRef = {
+    fileLevelVar(field, genName(className) + "__" + genName(methodName), origName)
+  }
+
+  def envVar(field: String, typeRef: NonArrayTypeRef)(
+      implicit pos: Position): VarRef = {
+    // The mapping in this function is an implementation detail of the emitter
```
sjrd: This comment seems obsolete. It is covered by the one in `coreJSLibVarIdent`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -385,7 +385,80 @@ private[emitter] final class JSGen(val semantics: Semantics,
...
+   *
+   * - envVar: Things pertaining to a given class. Accessible from every chunk
+   *   of the Scala.js output by importing its module.
+   * - coreJSLibVar: Things defined by the CoreJSLib. Accessible from every
+   *   chunk by importing the root module.
+   * - fileLevelVar: Things that are local to an individual file.
+   */
+
+  def envVar(field: String, className: ClassName)(implicit pos: Position): VarRef =
+    fileLevelVar(field, genName(className))
```
sjrd: It's confusing that `envVar` and `coreJSLibVar` delegate to `fileLevelVar`, although those three appear to be on par with each other. Consider having all three delegate to a "generic" implementation, that is not directly associated with one of them.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -127,11 +127,6 @@ private[emitter] final class JSGen(val semantics: Semantics,
   private def genFieldJSName(className: ClassName, field: irt.FieldIdent): String =
     genName(className) + "__f_" + genName(field.name)
 
-  def genSelectStatic(className: ClassName, item: irt.FieldIdent)(
```
sjrd, line 130: Why was this abstraction removed?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -262,15 +262,12 @@ private[emitter] final class JSGen(val semantics: Semantics,
     Apply(coreJSLibVar(helperName), args.toList)
   }
 
-  def encodeClassVar(className: ClassName)(implicit pos: Position): Tree =
```
sjrd: I'm not sure about this removal. Yes, it's a simple forwarder, but it's used quite a lot.

gzm0:
I have measured runtime of `testSuite2_12/test` on node v13.12.0 with the following pre-amble.

```
testSuite2_12/clean
set scalaJSLinkerConfig in testSuite.v2_12  ~= (_.withModuleKind(ModuleKind.ESModule))
set artifactPath in (testSuite.v2_12, Test, fastOptJS) := (crossTarget in testSuite.v2_12).value / "scalajs-test-suite-test-fastopt.mjs"
testSuite2_12/test:fastOptJS
```

Runtime (simply the time sbt shows, mod = this PR, base = merge base):

time

Summary:

```
$` mod`
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
     26      28      29      30      30      37 

$base
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
   19.0    22.0    24.0    23.6    25.0    28.0 
```

Raw data:

```
branch, time (s)
 mod, 29
 mod, 37
 mod, 26
 mod, 28
 mod, 30
base, 25
base, 28
base, 22
base, 19
base, 24
```

gzm0:
To me, this looks acceptable. We see some slowdown but nothing prohibitive.

@sjrd WDYT? Could you do similar measurements on windows to see if the slowdown is significantly different?

gzm0:
> A few comments about the preparatory commits (until "Wip modules", excluded).

Thanks. A lot of these things are contingent on whether we want to run the module analyzer post tree generation. IMHO this approach has a lot of benefits (e.g. determining if a class is "empty" is trivial). In any case, I'll have to measure performance to see how much it costs us to track all these fields.

sjrd:
I'll try on Windows. How exactly are you measuring? If it's the time reported in [success] by sbt, after a `testSuite2_12/clean`, that will include the compilation + linking time from scratch of the whole test suite.

gzm0:
I run `testSuite2_12/test:fastOptJS` before. And then in the same session `testSuite2_12/test` multiple times.

gzm0:
I also ran linker performance analysis (`split` is this PR, `master` is baseline).

speed

Observations:
- Refiner Analysis is slightly slower, this expected, due to the tracking of class dependencies (which is unused, but used in an earlier version).
- Class generation is on par. This indicates that field dependency tracking is not a significant overhead. Caveat: We do not sort classes anymore in the class generation step, but later. This might be the speed improvement we see.
- Module analysis is relatively slow (slower that a refiner analysis pass).
- Emitter tree assembly is slower. This might be due to grouping and sorting in this pass.
- File writing is slower. This is expected, we write significantly more files.

Conclusion:
- The current approach seems viable performance wise.
- Module Analyzer performance can probably be improved.

@sjrd other observations?

sjrd:
That seems very reasonable, especially since (I assume) the additional analyses are only performed if we request things to be split in several modules?

The module analysis should definitely not be slower than a `Refiner` pass, so there's certainly improvements to be made there, indeed.

gzm0:
I'd like to discuss when to run the module analysis.

The current proposal runs the module analysis *after* we have generated the JS trees. This works well, if the generated trees do not depend (locally) on the module structure. The advantage is, that needed fields get collected automatically, class dependencies are collected implicitly and empty modules are easy to avoid.

However, once we'll start introducing async module loading, the generated trees will depend on the module structure (unless we are willing to hoist a function calling `import` for all used fields).

Per consequence, it seems like we should run module analysis *before* JS tree generation. Resulting in the following challenges:

1. Determine the fields we need from a module.
1. Collect class dependencies.
1. Prune empty modules.

(2) can be done in the Analyzer (this PR has some code for this already).
(3) can be done in the Module Analyzer with a bit more of a rigorous approach to what all the fields in `LinkedClass` mean (probably mostly documentation, maybe some validation).

(1) Would require us to either:
- Run field collection anyways when emitting the trees.
- Provide more analysis results downstream (basically almost a full Info dependency graph).

However, this is only required if we want to avoid namespace imports. I would like to make a case, that we evaluate this decision more rigorously:

- Both @sjrd's comment on https://github.com/scala-js/scala-js/issues/3743#issuecomment-522293790 and this [SO answer](https://stackoverflow.com/a/45746950/1149944) hint that there is little to no benefit to using named imports.
- In my understanding, it goes against SJS philosophy to optimize for something a standard JavaScript optimizer can optimize (especially if it is not easy for us to do).

I think more evidence is needed that we actually gain any benefit from not using namespace imports. Next steps:

- @sjrd I'd like to hear your thoughts on this (also /cc @vhiairrassary who opened #3743).
- I'll go and check if webpack is able to prune unused module dependencies from a namespace import, if yes, we have strong case that namespace imports are fine. If no, we need to dig more.

gzm0:
As a post-scriptum: Dynamic imports and CommonJS imports both are much more like namespace imports. As a result, I expect simpler/less code to support all of these if we allow ourselves to use namespace imports.

gzm0:
I have tried out the [tree shaking example][1] from webpack. There was no difference in size (or output whatsoever) when switching a named import (like in the example) to a namespace import.

[1]: https://webpack.js.org/guides/tree-shaking/

sjrd:
My measurements on Windows are pretty damning.

Merge base:
```
60
49
32
35
28
34
```
This branch:
```
100
62
64
61
54
50
```
I don't put too much trust in the first measurement in each case, because it seems it did retrigger "fast optimizing" once, at least in "this branch". However, that's still a 35-ish versus 58-ish seconds.

Visually, I can see that a lot of time is lost while *loading*, and in particular between the display of "Module support is experimental" by Node.js and the first log output from the testing framework. As in 5-10s. In the merge base case, the delay exists but is well below 1s.

gzm0:
Plot of your data (including suspected outliers):

speed

Excluding suspected outliers:

speed

gzm0:
So it seems that splitting things into very small modules on Windows roughly doubles bare-bone execution time. Is this a deal-breaker?

My understanding is that the main reason we try to build this is support better incremental use cases for downstream systems. As a result, the bare-bone benchmarks we have are but an approximation. From what I see, I do not think we can conclude that this will definitely be too slow on windows (in an incremental setting). However, I don't think we can conclude that it will work fine on windows either.

Hence, for me the question is; shall we:
- Continue with this approach, hoping it will be fine.
- Benchmark more realistic use cases (i.e. hook up webpack, do incremental runs).
- Investigate other ways of splitting modules (I intend to build a "largest possible modules" mode in any case, so "other ways" here would be something in between).

Personally, I believe we should go ahead with the following features:

- Split entry points into multiple user-importable modules (not implemented here).
- Split "internal modules" into both "smallest" (this PR), and "biggest" (on a per user-importable module basis with shared modules).
- Later: Provide dynamic loading and relevant splitting.

@sjrd WDYT? Would this be OK with you, or do you think we need to investigate some of the loading performance more?

sjrd:
> So it seems that splitting things into very small modules on Windows roughly doubles bare-bone execution time. Is this a deal-breaker?

Maybe not a deal-breaker. We should document the potential slowdown, but if it enables better incremental runs in downstream tools, it might be worth it anyway.

> Personally, I believe we should go ahead with the following features:
> 
> * Split entry points into multiple user-importable modules (not implemented here).
> * Split "internal modules" into both "smallest" (this PR), and "biggest" (on a per user-importable module basis with shared modules).
> * Later: Provide dynamic loading and relevant splitting.

I agree. That sounds like what we want, indeed.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
gzm0: We currently call `envVar("c")` `classVar` (through `encodeClassVar`). I think `classVar` *is* a better name than `envVar`, but then we definitely need to rename `encodeClassVar` (or remove it).

ritschwumm:
it might be interesting to find out what exact filesystem operations make this so slow on windows - maybe there's a cheap workaround like just avoiding to enumerate files in a large directory or something like that...

godenji:
> My understanding is that the main reason we try to build this is support better incremental use cases for downstream systems.

My primary interest is:
> Later: Provide dynamic loading and relevant splitting.

Lack of lazy loading/code splitting keeps me tied to TypeScript and related frameworks that provide this functionality, but am very much looking forward to returning to Scala.js when this feature lands!

Glad to see progress here, thanks for getting the ball rolling.

PR #4012, opened 2020-04-14
Upgrade sbt to 1.3.9.

hepin1989:


sjrd:
Hi. Thanks for your contribution. Unfortunately we don't take PRs that upgrade sbt. Using `crossSbtVersions` is not nearly enough to get test coverage, because for most of our test suite, we source-in the sbt plugin inside the build, and we test the sbt plugin within the build. So we cannot ensure that our sbt plugin works correctly with sbt 1.2.x if we upgrade the build's sbt version to 1.3.x.

hepin1989:
Thanks for that infor

PR #4014, opened 2020-04-15
Fix #3500: Implement @js.native vals and defs.

### LINKED ISSUE #3500: Proposal: `@js.native` vals and defs
Currently, we can use `@js.native` only on `class`es and `object`s (and `trait`s, but with a different story). We can do that if they are at the top-level of a package, or within a static Scala `object`. For example:
```scala
object Container {
  @js.native
  @JSImport("foo.js", "bar")
  object Bar extends js.Object
}
```
This allows to easily import (or fetch from the global scope) *classes* and *objects* from JavaScript. However, for values and functions, we have to use an artificial enclosing object, and then use members of that object. For example:
```scala
@js.native
@JSImport("foo.js", JSImport.Namespace)
object Foo extends js.Any {
  val bar: Int = js.native
  def baz: String = js.native
  def add(x: Int, y: Int): Int = js.native
}
```
This is an inconsistent limitation, especially when contrasted with:
* `@JSExportTopLevel`, which can be used on any static term definition, and
* the fact that in Scala, every level of scoping obeys the same rules.

I propose that we allow `@js.native` (together with `@JSGlobal` or `@JSImport`) on `val`s, getter `def`s and method `def`s (not on setter `def`s) enclosed in a static Scala object (just like we do for `class`es and `object`s). And if Scala ever supports top-level `val`s and `def`s, we would extend `@js.native` to those as well. The right-hand side must in that case be `= js.native`. For example:
```scala
object Container {
  @js.native
  @JSImport("foo.js", "bar")
  val bar: Int = js.native

  @js.native
  @JSImport("foo.js", "baz")
  def baz: String = js.native

  @js.native
  @JSImport("foo.js", "add")
  def add(x: Int, y: Int): Int = js.native
}
```
In addition to the consistency argument, the above definition, while more verbose given the multiple imports, will allow to emit (and guarantee) better code for ES modules. Indeed, we will be able to avoid namespace imports, which have poor properties in terms of cross-module dead-code-elimination, something that is supposed to be one of the flagship advantages of ES modules.

Such definitions would naturally also be allowed in `package object`s, which will compensate the fact that we forbade package objects from extending `js.Any`.

sjrd:
We now allow `val`s and `def`s (but not `lazy val`s, `var`s or setter `def`s) in Scala static objects to be annotated with `@js.native`. For example:
```scala
object QueryStringWithNativeDef {
  @js.native
  @JSImport("querystring", "stringify")
  def stringify(obj: js.Dictionary[String], sep: String = "&",
      eq: String = "="): String = js.native
}

object OSWithNativeVal {
  @js.native
  @JSImport("os", "EOL")
  val EOL: String = js.native

  @js.native
  @JSImport("os", "EOL")
  def EOLAsDef: String = js.native
}
```

The rhs of such members must be `= js.native`, and they must have a native JS load spec annotation (`@JSGlobal` or `@JSImport`).

When accessing such a member, its value is looked up according to its native load spec, and for methods it is then called as a JS function.

---

Implementation-wise, these changes require that we can attach a JS native load spec to *members* of classes, whereas before only `ClassDef`s could have native load specs. We choose to add a new kind of `MemberDef` for this purpose:
```scala
JSNativeMemberDef(flags: MemberFlags, name: MethodIdent,
    jsNativeLoadSpec: JSNativeLoadSpec)
```
along with a term node to read their values:
```scala
SelectJSNativeMember(className: ClassName, member: MethodIdent)
```
The call sites identify `@js.native` members and, instead of compiling them to Scala `Apply` nodes, use `SelectJSNativeMember` (wrapped in a `JSFunctionApply` for methods).

Note that those members have *method* names, even though they are simple values. We will come back to this below.

Needed native JS members are tracked by the analyzer, using a dedicated "namespace".

The emitter emits native JS members exactly like existing classes and modules.

---

Several alternatives were considered and rejected, which led to the present design. Here is a complete history of ideas:

First, I thought we would make the public binary API use normal Scala methods, whose *body* would encode, in an implementation-specific way, the loading of the value and the function call. This would have allowed native members to implement and override inherited members. However, the idea was killed by the requirements for default parameters: it is not possible, inside the body of the method, to retroactively decide what parameters must be left out or not.

Once it was clear that the interpretation of the parameters had to be handled at call site, the problem of *overloading* appeared. Short of forbidding overloads of JS native defs, which would not have been consistent with methods in JS types, the binary API had to preserve JVM overloading semantics. This forced the unintuitive design that native JS members have *method names*, as opposed to *field names*, although they behave more like fields.

Another alternative that would have avoided the need for method names would have been to forbid overloads *that have different native load specs*. However, that would also have been inconsistent with methods of JS types (who can have different `@JSName`s), and would still have required an awkward binary API.

In the end, the design therefore settles on method names for dedicated node types.

The particular choice of `MethodName` invalidates a previously nice aspect of the IR, which was that each kind of `Name` was used for one purpose only. Now, `MethodName`s are used for both Scala methods and native JS fields. We could have used a different kind of name, isomorphic to `MethodName` (either by wrapping or replicating the structure) to avoid that. However, this was deemed unnecessary.


sjrd:
@gzm0 This is now ready for review 😃

gzm0:
Thank you for the extensive description of how this PR got into its current state. I think it'll speed up my review a lot.

So just to clarify: Using the same (IR) calling (but not definition) convention for non-native and native members is infeasible due to default parameters as well? (This leads to the unfortunate inconsistency that changing a class / object's native status is binary compatible, whereas a member's is not :().

sjrd:
> So just to clarify: Using the same (IR) calling (but not definition) convention for non-native and native members is infeasible due to default parameters as well?

Yes. In fact in the "First, I though" idea from the PR description, I actually implemented the whole thing based on using the same IR calling convention (using `Apply` nodes), until I got stuck with the default parameters. When I hit that road block, I was quite disappointed, although I realized "oh, of course, that cannot work: the calling convention for JS stuff and Scala are fundamentally different".



> This leads to the unfortunate inconsistency that changing a class / object's native status is binary compatible, whereas a member's is not :(

The fundamental difference here is that marking a class as `@js.native` or not does not change its status of being *a JS type*. That's why it is binary compatible. For vals and defs, adding `@js.native` turns them into JS members. That is fundamentally incompatible with them being Scala members. If you want to compare apples with apples in terms of class/object, it is the different between extending `js.Any` or not, and that is clearly not binary compatible.

So there might be an inconsistency in the full meaning of `@js.native`. But at a deeper level there is no inconsistency I think. Fixing the first inconsistency would mean adding a second annotation `@JSMember` in addition to `@js.native` (and in addition of the load spec annotation). But since there's no point in having a JS member that is *not* native, that seems completely redundant.

gzm0:
Yes I agree. Fundamentally, changing "JS status" is non-binary compatible.

We could think of exports as non-native JS members on non-JS classes, but of course the Scala.js calling convention does not match there (as in: there is no easy way to call an exported member with JS calling convention).

So I think, yes this is the best option we have at this point.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -253,6 +264,18 @@ private[emitter] final class KnowledgeGuardian(config: CommonPhaseConfig) {
...
 
+    private def computeJSNativeMemberLoadSpecs(
+        linkedClass: LinkedClass): Map[MethodName, JSNativeLoadSpec] = {
+      if (linkedClass.jsNativeMembers.isEmpty) {
+        // Fast path
+        Map.empty
+      } else {
+        (for (nativeMember <- linkedClass.jsNativeMembers) yield {
+          nativeMember.name.name -> nativeMember.jsNativeLoadSpec
+        }).toMap
```
gzm0, line 275: Consider:

```scala
linkedClass.jsNativeMembers
  .map(nm => nm.name.name -> nm.jsNativeLoadSpec)
  .toMap
```

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1327,14 +1333,18 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
           classDef.ancestors.toSet,
           classDef.hasInstances,
           classDef.jsNativeLoadSpec,
-          CheckedClass.checkedFieldsOf(classDef))
+          CheckedClass.checkedFieldsOf(classDef),
+          classDef.jsNativeMembers.map(_.name.name).toSet)
```
gzm0, line 1337: Is it intended that this basically allows `JSNativeMemberDefs` on any type of class?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/InteroperabilityTest.scala
```diff
@@ -392,6 +392,35 @@ class InteroperabilityTest {
     assertEquals(654, obj.bar)
   }
 
+  @Test def should_acces_top_level_JS_vars_and_functions_via_Scala_object_with_native_vals_and_defs(): Unit = {
```
gzm0, line 395: nit: access

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -204,25 +204,32 @@ final class Emitter private (config: CommonPhaseConfig,
...
+            case JSNativeLoadSpec.ImportWithGlobalFallback(
+                JSNativeLoadSpec.Import(module, _), _) =>
+              addModuleRef(module)
+          }
         }
+
+        if (classDef.hasInstances)
+          classDef.jsNativeLoadSpec.foreach(addModuleRefFor(_))
+
+        if (classDef.jsNativeMembers.nonEmpty) // likely false
```
gzm0, line 231: I somehow doubt that this will make anything faster, but OK :P

project/BinaryIncompatibilities.scala
```diff
@@ -20,6 +20,18 @@ object BinaryIncompatibilities {
   )
 
   val Linker = Seq(
+      // !! SemVer-Minor Breaking
```
gzm0, line 23: Maybe adjust this to make it clear that this is according to our compatibility guarantees (so OK).

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -102,26 +122,45 @@ object Infos {
...
           staticFieldsWritten, methodsCalled, methodsCalledStatically,
-          instantiatedClasses, accessedModules, usedInstanceTests,
-          accessedClassData, referencedClasses)
+          jsNativeMembersUsed, instantiatedClasses, accessedModules,
+          usedInstanceTests, accessedClassData, referencedClasses)
+    }
+
+    @deprecated("Use a ReachabilityInfoBuilder to construct ReachabilityInfo",
+        "1.1.0")
+    def apply(
```
gzm0, line 149: Simply remove this?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -102,26 +122,45 @@ object Infos {
...
   object ReachabilityInfo {
     val Empty: ReachabilityInfo = {
-      ReachabilityInfo(
-          Map.empty, Map.empty, Map.empty, Map.empty, Map.empty,
+      make(Map.empty, Map.empty, Map.empty, Map.empty, Map.empty, Map.empty,
           Nil, Nil, Nil, Nil, Nil)
     }
 
-    def apply(
+    private[Infos] def make(
```
gzm0, line 129: Feels a bit like overkill. Maybe make the constructor `private[Infos]`?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -59,8 +78,8 @@ object Infos {
         exportedMembers: List[ReachabilityInfo],
         topLevelExportedMembers: List[ReachabilityInfo],
         topLevelExportNames: List[String]): ClassInfo = {
-      new ClassInfo(className, isExported, kind, superClass,
-          interfaces, referencedFieldClasses, methods, exportedMembers,
+      make(className, isExported, kind, superClass, interfaces,
+          referencedFieldClasses, methods, Nil, exportedMembers,
           topLevelExportedMembers, topLevelExportNames)
```
gzm0, line 83: Same as below. Maybe just make `ClassInfo`s constructor `private[Infos]`?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -1157,44 +1156,39 @@ private final class Analyzer(config: CommonPhaseConfig,
           classInfo.callMethodStatically(methodName)
       }
     }
+
+    val jsNativeMembersUsedIterator = data.jsNativeMembersUsed.iterator
+    while (jsNativeMembersUsedIterator.hasNext) {
+      val (className, members) = jsNativeMembersUsedIterator.next()
+      lookupClass(className) { classInfo =>
+        for (memberName <- members)
+          classInfo.jsNativeMembersUsed += memberName
```
gzm0, line 1165: `classInfo.jsNativeMembersUsed ++= members`?

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -101,15 +105,51 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+      def d_=(v: Int): Unit = ()
+
+      @JSName("foo")
+      def e(x: Int): Int = x + 1
+    }
+    """ hasErrors
+    """
+      |newSource1.scala:6: error: Non JS-native vals may not have an @JSName annotation.
+      |      @JSName("foo")
+      |       ^
```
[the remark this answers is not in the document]
sjrd: Fixed.

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -405,28 +628,67 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+        """.stripMargin hasErrors s"""
+          |newSource1.scala:7: error: Native JS objects must have exactly one annotation among @JSGlobal, @JSImport and @JSGlobalScope.
+          |$secondAnnot
+          | ^
+          |newSource1.scala:12: error: Native JS classes must have exactly one annotation among @JSGlobal, @JSImport and @JSGlobalScope.
           |$secondAnnot
           | ^
         """
+
+        if (firstAnnot != "@JSGlobal" && secondAnnot != "@JSGlobal") {
```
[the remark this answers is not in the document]
sjrd: That's not allowed. I have added a test to make sure it is rejected (that was indeed missing).

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -400,33 +623,125 @@ class JSInteropTest extends DirectTest with TestHelpers {
...
+      |newSource1.scala:12: error: Only native JS objects can have an @JSGlobalScope annotation.
+      |      @js.native @JSGlobalScope
+      |                  ^
+      |newSource1.scala:15: error: Traits may not have an @JSGlobalScope annotation.
+      |      @js.native @JSGlobalScope
+      |                  ^
+      |newSource1.scala:18: error: Only native JS objects can have an @JSGlobalScope annotation.
+      |      @js.native @JSGlobalScope
+      |                  ^
+      |newSource1.scala:19: error: Native JS vals must have exactly one annotation among @JSGlobal, @JSImport and @JSGlobalScope.
```
gzm0, line 667: This error message is wrong :) It says `@JSGlobalScope` is OK. It's not :)

sjrd: Fixed. I also fixed it for classes, which were already wrong. Only `object`s can have `@JSGlobalScope`.

gzm0:
Thanks for the updates. LG so far.

I also just realized that there is no test for a `@js.native` member in a package object. Since this seems to be an explicitly advertised use case, maybe we should add one?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1125,7 +1134,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
       (for {
         f <- classSym.info.decls
         if !f.isMethod && f.isTerm && !f.isModule
-        if !f.hasAnnotation(JSOptionalAnnotation)
+        if !f.hasAnnotation(JSOptionalAnnotation) && !fieldSymIsBackingFieldForJSNativeMember(f)
```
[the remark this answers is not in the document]
sjrd: OK, I have just inlined the `_.hasAnnotation(JSNativeAnnotation)` everywhere.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -2386,6 +2406,14 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
               if (isNonNativeJSClass(sym.owner)) {
                 js.Assign(genNonNativeJSClassSelectAsBoxed(genQual, sym),
                     genBoxedRhs)
+              } else if (fieldSymIsBackingFieldForJSNativeMember(sym)) {
+                /* This is an assignment to a @js.native field. Since we reject
+                 * `@js.native var`s as compile errors, this can only happen in
```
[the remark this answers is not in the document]
sjrd: It's true that we could have `@JSGlobal`-annotated `var`s in the future. It's not possible with `@JSImport`, however. We can add support for that in the future, if there is demand for it.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -1327,14 +1333,18 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
     }
 
     def lookupField(name: FieldName): Option[CheckedField] =
       fields.get(name)
 
     def lookupStaticField(name: FieldName): Option[CheckedField] =
       staticFields.get(name)
+
+    def hasJSNativeMember(name: MethodName): Boolean =
+      jsNativeMembers.contains(name)
```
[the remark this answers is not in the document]
sjrd: Added checks for duplicate `jsNativeMembers`.

sjrd:
> I also just realized that there is no test for a `@js.native` member in a package object. Since this seems to be an explicitly advertised use case, maybe we should add one?

Added.

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -191,6 +191,17 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
           i"Duplicate definition of ${namespace.prefixString}method " +
           i"'$name' in class '${classDef.className}'")
     }
+
+    // JS native members
+    for {
+      (name, dupes) <- classDef.jsNativeMembers.groupBy(m => m.name.name)
```
gzm0, line 197: Consider dropping a comment that `jsNativeMembers` may only be in the public static namespace and that this is checked by assertion.

sjrd: Added :+1:

sjrd:
Rebased because of conflicts with #4028.

PR #4022, opened 2020-04-27
Fix #4021: Filter fields in Refiner exactly like BaseLinker.

### LINKED ISSUE #4021: The Refiner eliminates needed static fields in non-instantiated classes
The following `OptimizerTest` test case fails:
```scala
  @Test
  def testOptimizerDoesNotEliminateRequiredStaticField(): AsyncResult = await {
    val StringType = ClassType(BoxedStringClass)
    val fooGetter = m("foo", Nil, T)
    val classDefs = Seq(
        classDef(
            MainTestClassName,
            kind = ClassKind.ModuleClass,
            superClass = Some(ObjectClass),
            memberDefs = List(
                trivialCtor(MainTestClassName),
                FieldDef(EMF.withNamespace(PublicStatic), "foo", NON, StringType),
                MethodDef(EMF.withNamespace(MemberNamespace.PublicStatic),
                    fooGetter, NON, Nil, StringType, Some({
                      SelectStatic(MainTestClassName, "foo")(StringType)
                    }))(EOH, None),
                mainMethodDef({
                  predefPrintln(ApplyStatic(EAF, MainTestClassName, fooGetter, Nil)(StringType))
                })
            )
        )
    )

    for (linkingUnit <- linkToLinkingUnit(classDefs, MainTestModuleInitializers, TestIRRepo.fulllib)) yield {
      val mainClassDef = linkingUnit.classDefs.find(_.className == MainTestClassName).get
      println(mainClassDef.fields)
      for (m <- mainClassDef.methods)
        println(m.value.show)
      assertTrue(mainClassDef.fields.exists {
        case FieldDef(_, FieldIdent(name), _, _) => name == FieldName("foo")
        case _                                   => false
      })
    }
  }
```
The test passes if we disable the optimizer.

The issue is that the `Refiner` eliminates *all* fields of classes that are not instantiated. If we have a static-only class, which is never instantiates, this will also remove the static fields, although they are needed.

The `BaseLinker` has been doing the right thing since https://github.com/scala-js/scala-js/commit/9328d629294defc6eb75a7ffd0907068ce02f984#diff-f0f026500dbe763b128f16f8c1a07cde, but we did not transfer that logic in the `Refiner`.

This issue is "theoretical" as far as Scala.js is concerned, because we only create static fields to back `@JSExportTopLevel val`s, which always cause their enclosing class to be instantiated. So we cannot reproduce this issue with Scala.js code. However, it causes issues if we compile another language to the IR. In this case this was discovered by @arthanzel when compiling Java to the IR.

sjrd:
Previously, the Refiner would eliminate *all* fields in non-instantiated classes. This would remove static fields even if they are needed, which causes invalid .js files to be generated.

The BaseLinker has been doing the right thing since the commit 9328d629294defc6eb75a7ffd0907068ce02f984, but the Refiner had not been updated. We now make them both do the same thing, which solves the issue.

Since the issue is not reproducible with Scala.js source code, we test this with a linker test.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
@@ -149,17 +149,8 @@ final class BaseLinker(config: CommonPhaseConfig) {
     }
 
     classDef.memberDefs.foreach {
-      case field: FieldDef =>
-        val isNeeded = {
-          if (field.flags.namespace.isStatic) analyzerInfo.isAnyStaticFieldUsed
-          else if (classDef.kind.isJSType) analyzerInfo.isAnyPrivateJSFieldUsed
-          else analyzerInfo.isAnySubclassInstantiated
-        }
-        if (isNeeded)
-          fields += field
-
-      case field: JSFieldDef =>
-        if (analyzerInfo.isAnySubclassInstantiated)
+      case field: AnyFieldDef =>
+        if (isFieldDefNeeded(analyzerInfo, field))
           fields += field
 
       case m: MethodDef =>
@@ -217,7 +208,22 @@ final class BaseLinker(config: CommonPhaseConfig) {
   }
 }
 
-private object BaseLinker {
+private[frontend] object BaseLinker {
+  private[frontend] def isFieldDefNeeded(classInfo: ClassInfo,
+      field: ir.Trees.AnyFieldDef): Boolean = {
+    import ir.Trees._
+
+    field match {
+      case field: FieldDef =>
+        if (field.flags.namespace.isStatic) classInfo.isAnyStaticFieldUsed
+        else if (classInfo.kind.isJSType) classInfo.isAnyPrivateJSFieldUsed
+        else classInfo.isAnySubclassInstantiated
+
+      case field: JSFieldDef =>
+        classInfo.isAnySubclassInstantiated
+    }
+  }
+
   private class InputProvider extends Analyzer.InputProvider with MethodSynthesizer.InputProvider {
     private var classNameToFile: collection.Map[ClassName, IRFileImpl] = _
     private var entryPoints: collection.Set[ClassName] = _
--- a/linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
+++ b/linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
@@ -21,6 +21,7 @@ import org.scalajs.ir.ClassKind
 import org.scalajs.ir.EntryPointsInfo
 import org.scalajs.ir.Names._
 import org.scalajs.ir.Trees._
+import org.scalajs.ir.Trees.MemberNamespace._
 import org.scalajs.ir.Types._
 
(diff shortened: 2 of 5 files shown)


linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
```diff
@@ -217,7 +213,22 @@ final class BaseLinker(config: CommonPhaseConfig) {
   }
 }
 
-private object BaseLinker {
+private[frontend] object BaseLinker {
+  private[frontend] def isFieldDefNeeded(classInfo: ClassInfo,
+      classKind: ir.ClassKind, field: ir.Trees.AnyFieldDef): Boolean = {
```
gzm0, line 218: `classKind` is also on `classInfo`, right? Just take it from there?

linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
```diff
@@ -149,17 +149,13 @@ final class BaseLinker(config: CommonPhaseConfig) {
     }
 
     classDef.memberDefs.foreach {
-      case field: FieldDef =>
+      case field: AnyFieldDef =>
         val isNeeded = {
           if (field.flags.namespace.isStatic) analyzerInfo.isAnyStaticFieldUsed
           else if (classDef.kind.isJSType) analyzerInfo.isAnyPrivateJSFieldUsed
           else analyzerInfo.isAnySubclassInstantiated
         }
```
gzm0, line 157: I think this val should go?

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
```diff
@@ -168,6 +169,42 @@ class OptimizerTest {
...
+                    }))(EOH, None),
+                // static def main(args: String[]) { println(Test::foo()) }
+                mainMethodDef({
+                  predefPrintln(ApplyStatic(EAF, MainTestClassName, fooGetter, Nil)(StringType))
+                })
+            )
+        )
+    )
+
+    for (linkingUnit <- linkToLinkingUnit(classDefs, MainTestModuleInitializers, TestIRRepo.fulllib)) yield {
```
gzm0, line 199: Do we need the `fullib` for this test? IIUC this test should work with the `minilib` (in other words: `println` is not required).

sjrd: I replaced by a `console.log` to get rid of the full lib. It's useful to send it to some black-hole like `console.log` so that the optimizer can never eliminate the whole thing as dead code.

sjrd:
Updated.

linker/shared/src/test/scala/org/scalajs/linker/OptimizerTest.scala
gzm0: Ah, yes of course. I meant `println` specifically is not required (as opposed to `console.log` or something similar). I take it my comment was skewed from me looking at the classclass test, where we test the reachability chain of `println` specifically. In that sense `println` is "more" required there.

PR #4023, opened 2020-04-29
Upgrade to Scala 2.12.11 and 2.13.2.

sjrd:
The huge bumps in checksizes are really not nice. They are related to changes in the stdlib, so there's not much we can do about them.

I wonder how we could provide quicker feedback for these bumps to upstream Scala, so that they could avoid making Scala.js bigger all the time...


gzm0:
CI says:

```
[error] tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/BasicLinkerBackend.scala:38: procedure syntax is deprecated for constructors: add `=`, as in method definition
[error]       config: LinkerBackend.Config) {
[error]                                    ^
[error] tools/shared/src/main/scala/org/scalajs/core/tools/linker/backend/LinkerBackend.scala:40: procedure syntax is deprecated for constructors: add `=`, as in method definition
[error]       config: LinkerBackend.Config) {
[error]                                    ^
[error] two errors found
```

gzm0:
Otherwise LG

sjrd:
Added a commit to fix the procedure syntax deprecation errors.

PR #4024, opened 2020-04-30
Gitignore node_modules

MaximeKjaer:
The first step of `DEVELOPING.md` is to run `npm install`, which fills up `node_modules` with dependency artifacts that shouldn't be tracked by git.

PR #4025, opened 2020-04-30
Gitignore node_modules

MaximeKjaer:
The first step of `DEVELOPING.md` is to run `npm install`, which fills up `node_modules` with dependency artifacts that shouldn't be tracked by git.

Same PR as #4024 but targeting branch `0.6.x` instead of `master`.


.gitignore
```diff
@@ -9,3 +9,4 @@ target/
 /.idea/
 /.idea_modules/
 bin/
+node_modules/
```
gzm0, line 12: Actually, shouldn't this be `/node_modules/`?

MaximeKjaer: I suppose there wouldn't be `node_modules` anywhere else than at the root, so that makes sense, yes.

MaximeKjaer:
Okay, good to know -- I'll do that going forward. I've squashed the two commits into one.

PR #4028, opened 2020-05-02
Minor fixes for top level exports

gzm0:


sjrd:
Bootstrap tests don't pass. There's a compile error.
```
[error] .../linker/shared/src/test/scala/org/scalajs/linker/testutils/IRAssertions.scala:93:57: value value is not a member of org.scalajs.ir.Trees.TopLevelExportDef
[error]             traverseTopLevelExportDef(topLevelExportDef.value)
[error]                                                         ^
[error] one error found
```

PR #4032, opened 2020-05-04
Always use ES6 as intermediary closure input

gzm0:


gzm0:
@sjrd given https://github.com/scala-js/scala-js/issues/2675#issuecomment-623382097, we should close this. Do you think there is value in the first commit on its own?

gzm0:
Seems like your concerns were justified :P The ES5 test suite seems to fail :P

sjrd:
> Seems like your concerns were justified :P The ES5 test suite seems to fail :P

😅 

> Do you think there is value in the first commit on its own?

Yes, the first commit looks good to have.

gzm0:
Closing. First commit is in #4033.

PR #4033, opened 2020-05-04
Give the Emitter its own config

gzm0:
This removes the InternalOptions construct and replaces it with a
standard construct. The relevant options can still be removed at any
point in time, due to our loose compatibility guarantees for this
package.


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -388,7 +389,7 @@ private[emitter] class FunctionEmitter(jsGen: JSGen) {
          * slower, but running GCC will take most of the time anyway in that case.
          */
         val globalRefs =
-          if (trackAllGlobalRefs) globalVarNames.toSet
+          if (config.trackAllGlobalRefs) globalVarNames.toSet
```
sjrd, line 392: Is this change necessary? There's an `import config._` at the top.

gzm0: No, it's not. I have found a bunch of others where the `import config._` makes the code shorter.

gzm0:
Rebased, resolved trivial conflicts in `BinaryIncompatibilities.scala`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
sjrd: > No, it's not

Consider reverting that change, then.

gzm0: Ugh... I thought I did... Updated.

PR #4040, opened 2020-05-11
Fix #4034: Encode -x as -1.0 * x instead of 0.0 - x.

### LINKED ISSUE #4034: Incorrect handling of negative 0
The following snippet produces different results on Scala.js and on the JVM:

```scala
val float = 0.0f
println(String.format("%.1f", -float))     // prints 0.0 on Scala.js, -0.0 on JVM
println(String.format("%.1f", -1 * float)) // prints -0.0 on both Scala.js and JVM
```

From the JS output, it looks like the first print is being constant folded incorrectly to `0.0` instead of `-0.0` in Scala.js. The second print is being constant folded correctly to `-0.0`:

```js
this$9.java$lang$JSConsoleBasedPrintStream$$printString__T__V((x$1 + "\n"));
const args$1 = $makeNativeArrayWrapper($d_O.getArrayOf(), [0.0]);
const x$2 = $m_jl_String$().format__T__AO__T("%.1f", args$1);
const this$11 = $m_s_Console$();
const this$12 = this$11.out__Ljava_io_PrintStream();
this$12.java$lang$JSConsoleBasedPrintStream$$printString__T__V((x$2 + "\n"));
const args$2 = $makeNativeArrayWrapper($d_O.getArrayOf(), [(-0)]);
const x$3 = $m_jl_String$().format__T__AO__T("%.1f", args$2);
```

I can try to prevent constant folding from taking place by placing the negation in a function:

```scala
def negate1(x: Float): Float = -x
def negate2(x: Float): Float = -1 * x
```

However, this also results in an incorrect translation. Both `-x` and `-1 * x` are translated to `Math.fround(0.0 - x)`, which produces an incorrect result when `x` is `0.0`.

```js
const $p_Lcom_test_Test__negate1$1__F__F = (function($thiz, x) {
  return $fround((0.0 - x))
});
const $p_Lcom_test_Test__negate2$1__F__F = (function($thiz, x) {
  return $fround((0.0 - x))
});
```

I am using Scala.js 1.0.1 on Scala 2.13.2.

sjrd:
The result of `-x` is not the same as `0.0 - x` when `x` is `0.0`: it is -0.0 instead of +0.0. Therefore, encoding `-x` as `0.0 - x` was incorrect, as well as link-time optimizations that interpreted `0.0 - x` as `-x`.

We now encode `-x` as `-1.0 * x` instead, which always give the correct result. We adapt the rules in the optimizer and in the emitter to recognize `-1.0 * x` as `-x`, instead of `0.0 - x`.

In addition, we remove some wrong simplifications in the optimizer, that were assuming that `+` is associative for doubles, whereas it is not.

All of the above also applies to floats.

Only code compiled with the newer version of the compiler will be fixed.

**Attention!** Code doing `-x` in source code and compiled with an earlier version of the compiler will be *broken* by this change, since it was encoded as `0.0 - x` in the IR, which used to be emitted as `-x` (fixing the wrong encoding) and now emitted as `0.0 - x` (which is correct given the IR, but incorrect given the original source code). An alternative would be to add a deserialization hack to deserialize `0.0 - x` as `-1.0 * x`.

### DIFF
--- a/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
+++ b/tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
@@ -3451,40 +3451,6 @@ private[optimizer] abstract class OptimizerCore(
           case _ => default
         }
 
-      case Float_+ =>
-        (lhs, rhs) match {
-          case (PreTransLit(FloatLiteral(0)), _) =>
-            rhs
-          case (_, PreTransLit(FloatLiteral(_))) =>
-            foldBinaryOp(Float_+, rhs, lhs)
-
-          case (PreTransLit(FloatLiteral(x)),
-              PreTransBinaryOp(innerOp @ (Float_+ | Float_-),
-                  PreTransLit(FloatLiteral(y)), z)) =>
-            foldBinaryOp(innerOp, PreTransLit(FloatLiteral(x + y)), z)
-
-          case _ => default
-        }
-
-      case Float_- =>
-        (lhs, rhs) match {
-          case (_, PreTransLit(FloatLiteral(r))) =>
-            foldBinaryOp(Float_+, lhs, PreTransLit(FloatLiteral(-r)))
-
-          case (PreTransLit(FloatLiteral(x)),
-              PreTransBinaryOp(Float_+, PreTransLit(FloatLiteral(y)), z)) =>
-            foldBinaryOp(Float_-, PreTransLit(FloatLiteral(x - y)), z)
-          case (PreTransLit(FloatLiteral(x)),
-              PreTransBinaryOp(Float_-, PreTransLit(FloatLiteral(y)), z)) =>
-            foldBinaryOp(Float_+, PreTransLit(FloatLiteral(x - y)), z)
-
-          case (_, PreTransBinaryOp(BinaryOp.Float_-,
-              PreTransLit(FloatLiteral(0)), x)) =>
-            foldBinaryOp(Float_+, lhs, x)
-
-          case _ => default
-        }
-
       case Float_* =>
         (lhs, rhs) match {
           case (_, PreTransLit(FloatLiteral(_))) =>
@@ -3492,8 +3458,9 @@ private[optimizer] abstract class OptimizerCore(
 
           case (PreTransLit(FloatLiteral(1)), _) =>
             rhs
-          case (PreTransLit(FloatLiteral(-1)), _) =>
-            foldBinaryOp(Float_-, PreTransLit(FloatLiteral(0)), rhs)
+          case (PreTransLit(FloatLiteral(-1)),
+              PreTransBinaryOp(Float_*, PreTransLit(FloatLiteral(-1)), z)) =>
+            z
 
           case _ => default
         }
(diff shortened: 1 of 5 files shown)


sjrd:
@gzm0 Please pay particular attention to the last paragraph of the PR description. WDYT, should we use the alternative instead? (that would break code doing `0.0 - x` in source code instead of breaking code doing `-x` in source code)

tools/shared/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/OptimizerCore.scala
```diff
@@ -3451,49 +3451,16 @@ private[optimizer] abstract class OptimizerCore(
           case _ => default
         }
 
-      case Float_+ =>
-        (lhs, rhs) match {
-          case (PreTransLit(FloatLiteral(0)), _) =>
-            rhs
-          case (_, PreTransLit(FloatLiteral(_))) =>
-            foldBinaryOp(Float_+, rhs, lhs)
```
gzm0, line 3459: These two are also incorrect? Or just not worth it anymore? (similar comments below).

sjrd: The first one is incorrect : `+0.0f + -0.0f` gives `+0.0f`, not `rhs` which is `-0.0f`.

The second one is not worth it anymore.

sjrd:
> My understanding is that because the linker (optimizer and emitter) freely translated between `0.0 - x` and `-x` likely we can find instances of both being wrong with SJS 1.0.1 and maybe even the exact semantic depending on whether the optimizer was on or not. So breaking both of them differently in 1.1.0 is IMO not a problem.

AFAICT, the previous linker always treated `0.0 - x` with the semantics of `-x` (reliably), including whether or not the optimizer was enabled. I can't think of a scenario where even the optimizer "disconnects" `0.0 - x` into something else, other than by treating it as `-x`, and it was the *emitter* that compiled `0.0 - x` down to `-x` in JS.

So at least I'm pretty sure that all uses of `-x` in the source code of 0.6.32/1.0.1 currently do the right thing.

gzm0:
> So at least I'm pretty sure that all uses of -x in the source code of 0.6.32/1.0.1 currently do the right thing.

Hmmm... But then why is it observable (as stated in the bug report)? At least when constant folding, it seems the linker didn't treat `0.0 - x` with the semantics of `-x` (i.e. if `x` was a known constant).

Or am I misunderstanding what is happening here completely?

sjrd:
Oh, you're right. The optimizer could indeed incorrectly *constant-fold* `-x`. Because if `x` is the *constant* `0.0`, then the optimizer would see a `0.0 - 0.0` and constant-fold it on the JVM to `0.0`, instead of the correct result `-0.0`. The constant-folding indeed bypasses the identification as `-x` and its semantics.

So yes, you're right, the optimizer has an impact on the validity of `-x` in 0.6.32/1.0.1. So indeed it's probably best to stick to the simplest fix, which is this PR as you said.

sjrd:
Added an `assumeFalse(executingInRhino)`, because Rhino computes completely wrong results.

plokhotnyuk:
Why `-1.0 * x`? Is `-0 - x` not an option?
```
$ node
Welcome to Node.js v14.2.0.
Type ".help" for more information.
> -0 - (-0)
0
> -0 - 0
-0
> -0 - (-1)
1
> -0 - 1
-1
```

sjrd:
Well, that would work too, apparently. It's clearly non-obvious, though (much less so than `-1.0 * x`), so *why* `-0.0 - x`? (instead of "why not?")

plokhotnyuk:
Usually ADD/SUB operations have fewer latency cycles and greater throughput than MUL on contemporary CPUs: https://travisdowns.github.io/blog/2019/06/11/speed-limits.html#portexecution-unit-limits

plokhotnyuk:
Usually, FADD/FSUB have fewer latency on contemporary CPUs than FMUL: https://www.agner.org/optimize/instruction_tables.pdf

sjrd:
There's no need to care about that. It is compiled into `-x` in JS at the end of the day (see the changes in `FunctionEmitter` in this PR). The efficiency of the encoding *if it was compiled as is* is completely irrelevant, because it's not compiled as is.

PR #4049, opened 2020-05-14
IRCheck ClassDef.jsSuperClass

gzm0:
It has relatively involved constraints which are documented but were not checked.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
@@ -146,7 +146,7 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
         if (superClass.jsClassCaptures.isDefined)
           reportError(i"super class ${superClass.name} may not have jsClassCaptures")
       } { tree =>
-        val env = Env.fromSignature(NoType, classDef.jsClassCaptures, Nil, AnyType)
+        val env = Env.fromSignature(NoType, classDef.jsClassCaptures, Nil)
         typecheckExpect(tree, env, AnyType)
       }
     } else {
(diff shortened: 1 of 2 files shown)


linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -135,6 +136,24 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+
+    val hasJSSuper = classDef.jsSuperClass.isDefined
+
+    if (classDef.kind.isJSClass) {
+      if (!hasJSSuper) {
+        // .get is OK: the Analyzer checks that a super class is present.
+        val superClass = lookupClass(classDef.superClass.get.name)
+        if (superClass.jsClassCaptures.isDefined)
+          reportError(i"super class ${superClass.name} may not have jsClassCaptures")
+      }
```
sjrd, line 150: Shouldn't we have an `else` branch that typechecks the `Tree` in the `jsSuperClass`, while we're here?

linker/shared/src/main/scala/org/scalajs/linker/checker/IRChecker.scala
```diff
@@ -135,6 +136,25 @@ private final class IRChecker(unit: LinkingUnit, logger: Logger) {
...
+    implicit val ctx = ErrorContext(classDef)
+
+    if (classDef.kind.isJSClass) {
+      classDef.jsSuperClass.fold {
+        // .get is OK: the Analyzer checks that a super class is present.
+        val superClass = lookupClass(classDef.superClass.get.name)
+        if (superClass.jsClassCaptures.isDefined)
+          reportError(i"super class ${superClass.name} may not have jsClassCaptures")
+      } { tree =>
+        val env = Env.fromSignature(AnyType, classDef.jsClassCaptures, Nil, AnyType)
```
sjrd, line 149: I think `thisType` should be `NoType` instead of `AnyType`, shouldn't it?
```suggestion
        val env = Env.fromSignature(NoType, classDef.jsClassCaptures, Nil, AnyType)
```

gzm0: Ah, yes, of cousre.

sjrd:
@gzm0 There were MiMa errors :(

PR #4050, opened 2020-05-14
RFC Split TopLevelExportDefs away from LinkedClass

gzm0:
This will help with organizing modules.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
@@ -110,21 +110,28 @@ final class BaseLinker(config: CommonPhaseConfig) {
   private def assemble(moduleInitializers: Seq[ModuleInitializer],
       analysis: Analysis)(implicit ec: ExecutionContext): Future[LinkingUnit] = {
     def assembleClass(info: ClassInfo) = {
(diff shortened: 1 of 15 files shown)


gzm0:
Not to be merged, just to validate the overall approach.

linker/shared/src/main/scala/org/scalajs/linker/frontend/BaseLinker.scala
```diff
@@ -110,21 +110,28 @@ final class BaseLinker(config: CommonPhaseConfig) {
...
 
     for {
-      linkedClassDefs <- Future.traverse(analysis.classInfos.values)(assembleClass)
+      assembled <- Future.traverse(analysis.classInfos.values)(assembleClass)
     } yield {
+      val (linkedClassDefs, linkedTopLevelExports) = assembled.unzip
+
       new LinkingUnit(config.coreSpec, linkedClassDefs.toList,
+          linkedTopLevelExports.flatten.toList,
           moduleInitializers.toList)
```
gzm0, line 135: @sjrd do you know of a better way of doing this? I'd much rather accumulate with local mutable collections. But the async nature of all this stuff makes this hard :S

sjrd: I'm afraid I don't have any better idea.

PR #4056, opened 2020-05-19
Overhaul PrepJSInterop to be simpler, more uniform, with less code paths.

sjrd:
`PrepJSInterop` was a huge bag of spaghetti code, with more checks added on top of the existing checks as the years passed, without ever reconsidering the code paths.

This commit is a major overhaul of that phase, which several advantages.

First, we clearly separate the handling of a) member defs, b) templates and c) statements/expressions.

Second, checks are structured hierarchically so that each check is performed at exactly one place, rather than being repeated for every kind of member def that requires the checks. This makes it much easier not to forget obscure combinations of features that would be missing checks. As a "proof" of that claim, each non-helper method is called from exactly one call site. (A helper method only inspects some properties of a symbol, as opposed to performing real checks, reporting errors, mutating the symbol table and/or transforming the tree.)

Exceptions are:

* `validateJSConstructorOf`, called both from the transformation of `js.constructorOf[T]` and that of `js.ConstructorTag.materialize[T]`. This could be avoided by re-`transform`ing the tree generated by the materialization, but that would perform useless work.
* `checkSetterSignature`, called both for setters in JS types and exported setters in Scala types.
* `checkRHSCallsJSNative`, called both for concrete members in native JS types and for native JS vals and defs.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
@@ -558,18 +558,13 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
 
       // Check the parents
       for (parent <- sym.info.parents) {
-        def strKind =
-          if (sym.isTrait) "trait"
-          else if (sym.isModuleClass) "object"
-          else "class"
-
         parent.typeSymbol match {
           case AnyRefClass | ObjectClass =>
             // AnyRef is valid, except for non-native JS traits
             if (!isJSNative && !sym.isTrait) {
               reporter.error(implDef.pos,
-                  s"A non-native JS $strKind cannot directly extend AnyRef. " +
-                  "It must extend a JS class (native or not).")
+                  "Non-native JS classes and objects cannot directly extend " +
+                  "AnyRef. They must extend a JS class (native or not).")
             }
           case parentSym if isJSAny(parentSym) =>
             // A non-native JS type cannot extend a native JS trait
@@ -577,8 +572,8 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
             if (!isJSNative && parentSym.isTrait &&
                 parentSym.hasAnnotation(JSNativeAnnotation)) {
               reporter.error(implDef.pos,
-                  s"A non-native JS $strKind cannot directly extend a "+
-                  "native JS trait.")
+                  "Non-native JS types cannot directly extend native JS " +
+                  "traits.")
             }
           case DynamicClass =>
             /* We have to allow scala.Dynamic to be able to define js.Dynamic
@@ -801,7 +796,7 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
           case Some(annot) if annot.symbol == JSGlobalScopeAnnotation =>
             if (!sym.isModuleClass) {
               reporter.error(annot.pos,
-                  "Only native JS objects can have an @JSGlobalScope annotation.")
+                  "@JSGlobalScope can only be used on native JS objects (with @js.native).")
             }
             None
 
@@ -852,8 +847,6 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
     private def transformJSNativeValOrDefDef(tree: ValOrDefDef): ValOrDefDef = {
       val sym = tree.symbol
 
-      def kind = kindStrFor(sym)
-
       if (sym.isLazy || jsInterop.isJSSetter(sym)) {
         reporter.error(tree.pos,
             "@js.native is not allowed on vars, lazy vals and setter defs")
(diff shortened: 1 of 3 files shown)


sjrd:
It's worth pointing out that, due to correctly centralizing everything, there is a possibility that, after this commit, errors will now be reported for incorrect code that was slipping through the cracks before. According to our versioning policy, this is supposed to warrant a **minor** release.

gzm0:
> It's worth pointing out that, due to correctly centralizing everything, there is a possibility that, after this commit, errors will now be reported for incorrect code that was slipping through the cracks before. According to our versioning policy, this is supposed to warrant a minor release.

IMO that just means we fixed bugs. Especially if the definition of incorrect code didn't change.

But it is true that the versioning policy does not distinguish these two things. Personally, I think it crucial it does. Further, we seem to have applied this reasoning, the fix to #3976 being a recent example.

sjrd:
I don't see the link with #3976. There was no code that would compile before its fix and wouldn't compile anymore after the fix.

> IMO that just means we fixed bugs. Especially if the definition of incorrect code didn't change.

Yes, in the case of this PR it would only be fixing bugs. Yet despite that, in the 0.6.x series we used to introduce *warnings* instead of errors when we deliberately added checks that we had not thought about before.

gzm0:
> I don't see the link with #3976. There was no code that would compile before its fix and wouldn't compile anymore after the fix.

The link is in the wording of `VERSIONING.md`: #3976 would be considered "a backward incompatible change" if "backward incompatible" would include bugs. This is because an alternative implementation of `System.identityHashCode` would be replaced.

In that case, this change would be a

> Backward source incompatible change at the language level or at the standard library level (including any addition of public API in the stdlib)

indeed warranting a minor version bump.

> Yet despite that, in the 0.6.x series we used to introduce warnings instead of errors when we deliberately added checks that we had not thought about before.

Fair enough. I'm willing to attribute this to the ambiguity of specification itself.

In any case, all things considered, I do not think we should bump minor for this PR. However, I do not feel very strongly about it.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -521,25 +546,53 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
+          else "class"
+
+        parent.typeSymbol match {
+          case AnyRefClass | ObjectClass =>
+            // AnyRef is valid, except for non-native JS traits
+            if (!isJSNative && !sym.isTrait) {
+              reporter.error(implDef.pos,
+                  s"A non-native JS $strKind cannot directly extend AnyRef. " +
+                  "It must extend a JS class (native or not).")
+            }
```
gzm0, line 573: Don't we need a newline here?

sjrd:
> In any case, all things considered, I do not think we should bump minor for this PR. However, I do not feel very strongly about it.

OK, let's keep it as is, as a patch version.

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -2115,6 +2138,11 @@ class JSInteropTest extends DirectTest with TestHelpers {
   def noJSGlobalOnMembersOfClassesAndTraits: Unit = {
 
     for (outer <- Seq("class", "trait")) {
+      // There is a bug in kindStrFor() for vars in traits in Scala 2.12+
+      val varKind =
+        if (outer == "trait" && !scala.util.Properties.versionNumberString.startsWith("2.11.")) "vals"
+        else "vars"
```
sjrd: This problem disappears in the second commit by construction, so after ~10 minutes of trying to make it work in the intermediary commit, I abandoned.

Same thing in `noJSImportOnMembersOfClassesAndTraits`.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
gzm0: Ah :)

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -81,10 +81,10 @@ class JSInteropTest extends DirectTest with TestHelpers {
       $obj B extends js.Object
       """ hasErrors
       s"""
-        |newSource1.scala:5: error: Non JS-native ${pluralOfKind(obj)} may not have an @JSName annotation.
```
gzm0, line 84: `pluralOfKind` seems to be unused now.

compiler/src/test/scala/org/scalajs/nscplugin/test/JSInteropTest.scala
```diff
@@ -2285,13 +2275,13 @@ class JSInteropTest extends DirectTest with TestHelpers {
       }
       """ hasErrors
       s"""
```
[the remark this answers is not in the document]
sjrd: That one is necessary (see `$fallbackStr`)

PR #4060, opened 2020-05-25
Mitigate #4054: Refuse to run TestAdapter with empty input

### LINKED ISSUE #4054: Empty jsEnvInput causes freeze
How on earth to do configure SBT and/or Scala.JS to allow a JS module without tests?

I have no idea what the default `jsEnv` is or how it works but a new module with basically just `enablePlugins(ScalaJSPlugin, JSDependenciesPlugin)` and not much else, running `test` throws an exception:

```
[error] stack trace is suppressed; run last demo / Test / jsEnvInput for the full output
[error] (demo / Test / jsEnvInput) java.nio.file.NoSuchFileException: require-file:/tmp/tmp-12228659344646593601react.development.js
```

I'm trying a few things but to no avail. One of the things I've tried is `Test / jsEnvInput := Nil` but when I run `test` it just hangs. Presumably it simply doesn't handle this case and sits around waiting for the jsEnv to signal completion.

If `Test / jsEnvInput := Nil` I think either one of two things should happen
* either there should probably be an assertion when the tests start to say "hey, it can't be empty!"
* or it should accept that empty means nothing to do, and just return early with success or completion

gzm0:


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -487,9 +488,16 @@ private[sbtplugin] object ScalaJSPluginInternal {
               s"$configName := true`")
         }
 
+        if (input.isEmpty) {
+          throw new MessageOnlyException(
+              s"`jsEnvInput in $configName` is empty but `test in $configName` got called. " +
+              "This is not allowed, since running tests requires the generated Scala.js code. " +
+              s"If you want to call `test in $configName` but not have it do anything, " +
+              s"set `test in $configName` := {}`.")
```
sjrd, line 496: It's probably time to write scoped tasks using the slash syntax of newer sbt versions, at least in error messages. sbt users tend to only use the slash syntax now, so they might not understand the `in` variants.
```suggestion
              s"`$configName / jsEnvInput` is empty but `$configName / test` got called. " +
              "This is not allowed, since running tests requires the generated Scala.js code. " +
              s"If you want to call `$configName / test` but not have it do anything, " +
              s"set `$configName / test` := {}`.")
```
We should probably swipe through the file to change all the existing error messages at some point, but at least for new messages I think we should use the new syntax.

sjrd: Also consider inverting the two subsentences of the first sentence:
```scala
              s"`$configName / test` got called but `$configName / jsEnvInput` is empty. " +
```
It seems more comprehensible to me, but YMMV.

PR #4071, opened 2020-06-08
Syntax change to make it compatible with dotty

FabioPinheiro:
The test ArrayBuilderTest does not compile on Dotty but the change is not related with ScalaJs
@smarter mentioned that dotty is stricter about auto-tupling
https://gitter.im/lampepfl/dotty?at=5ede2cb6a91f120a6cee7c6f

Minor change on code syntax so the test can be enable:
https://github.com/lampepfl/dotty/blob/ff7673c1aed82cdc7293ddd71f43d2176160eb9a/project/Build.scala#L1061

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/ArrayBuilderTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/ArrayBuilderTest.scala
@@ -193,7 +193,7 @@ class ArrayBuilderTest {
 
   @Test def Unit_normal_case_inline(): Unit = {
     val b = ArrayBuilder.make[Unit]
-    b += ()
+    b += (())
     val a = b.result()
 
     assertSame(classOf[Array[Unit]], a.getClass)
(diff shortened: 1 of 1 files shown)


sjrd:
From the discussion you link, it is not clear to me that the behavior of Dotty is desirable. I'd like to see an explicit decision that it is indeed desirable. To me Scala 2.13.2 is correct:
```scala
$ ~/opt/scala-2.13.2/bin/scala -feature -deprecation
Welcome to Scala 2.13.2 (OpenJDK 64-Bit Server VM, Java 1.8.0_252).
Type in expressions for evaluation. Or try :help.

scala> val b = scala.collection.mutable.ArrayBuilder.make[Unit]
val b: scala.collection.mutable.ArrayBuilder[Unit] = ArrayBuilder.ofUnit

scala> b += ()
val res0: b.type = ArrayBuilder.ofUnit

scala> b.+=()
           ^
       warning: adaptation of an empty argument list by inserting () is deprecated: this is unlikely to be what you want
               signature: Growable.+=(elem: A): Growable.this.type
         given arguments: <none>
        after adaptation: Growable.+=((): Unit)
val res1: b.type = ArrayBuilder.ofUnit

scala> b.+=(())
val res2: b.type = ArrayBuilder.ofUnit
```
According to Scala 2.13.2, `b += ()` is not doing any auto-tupling. It is precisely the legit use case of using the value `()` as the rhs of the operator `+=`.

sjrd:
This is being fixed in dotty: https://github.com/lampepfl/dotty/pull/9137

PR #4073, opened 2020-06-09
Optimize JSIdentifier by eliminating mutation and increasing readability

ayushhooda:


ayushhooda:
I did sign the CLA

sjrd:
Hello @ayushhooda. Thanks for your interest in contributing to Scala.js.

I am not sure what you are optimizing for? Is it speed? Intuitively it seems that it would be slower, because at least the tests for `nameLength == 0` and `i == 0` are done at every iteration of the loop, rather than only once at the beginning before. Do you have any evidence that your suggestion is faster than the existing code?

ayushhooda:
@sjrd Yes agree to your point, but rest conditions can be iterated using tail recursion instead of while loop. I can update the PR if you agree

sjrd:
I'm still not sure what exactly you are looking to improve. From your last comment, it seems you are mostly after making the implementation of that method functional inside instead of using `var`s? If it's just that, then that is really not a priority in this codebase. We often use mutations locally within functions, especially for methods that are performance-sensitive (and for this method, there is specifically a comment that says it should be fast). We don't consider the pure functional style to be necessarily better in the small.

So unless there is an advantage in terms of performance, I'm afraid we're not going to accept a PR that just makes things more functional for the sake it.

sjrd:
Closing this because of lack of activity to answer my latest concerns.

PR #4075, opened 2020-06-10
Restructure Emitted Class generation flow

gzm0:


sjrd:
Also I'm not convinced by the third commit. There is an important difference between the timing of tasks subject to caching ("generate trees") and the timings of tasks that are done from scratch every time ("write trees").

PR #4076, opened 2020-06-10
Add an IDE stub for `emitter.PrivateLibHolder` in the linkerJS.

sjrd:
In the JVM version of `linker`, `PrivateLibHolder` is a normal source file, but in JS it is generated by the build. Depending on whether Metals chose to associate `Emitter.scala` to the JVM or JS project, we could get spurious compile errors in the IDE because it would not find `PrivateLibHolder` in the JS project.

We fix this by adding an IDE stub for `PrivateLibHolder` in the JS version of the linker, similarly to what we do for other generated sources.

---

I've been puzzled for a while that Metals would sometimes fail to compile the `linker` files, and sometimes not. I finally figured it out.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -810,48 +810,53 @@ object Build {
   ).zippedSettings("library")(
       commonLinkerSettings _
   ).settings(
-      sourceGenerators in Compile += Def.task {
-        val dir = (sourceManaged in Compile).value
-        val privateLibProducts = (products in (linkerPrivateLibrary, Compile)).value
+      if (isGeneratingForIDE) {
+        unmanagedSourceDirectories in Compile +=
+          baseDirectory.value.getParentFile.getParentFile / "js/src/main/scala-ide-stubs"
+      } else {
+        sourceGenerators in Compile += Def.task {
+          val dir = (sourceManaged in Compile).value
+          val privateLibProducts = (products in (linkerPrivateLibrary, Compile)).value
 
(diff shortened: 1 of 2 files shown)


project/Build.scala
```diff
@@ -810,48 +810,53 @@ object Build {
...
       commonLinkerSettings _
   ).settings(
-      sourceGenerators in Compile += Def.task {
-        val dir = (sourceManaged in Compile).value
-        val privateLibProducts = (products in (linkerPrivateLibrary, Compile)).value
+      if (isGeneratingForIDE) {
+        unmanagedSourceDirectories in Compile +=
+          baseDirectory.value.getParentFile.getParentFile / "js/src/main/scala-ide-stubs"
+      } else {
+        sourceGenerators in Compile += Def.task {
```
gzm0, line 817: Have you considered introducing a helper `sourceGeneratorOrStub` or similar?

This might make it easier for us to spot sites where we use source generators without IDE support.

sjrd: I'm not sure the abstraction is worth it TBH. Spotting sites with source generators without IDE support is easy enough: grep for `sourceGenerators`, and if it's not enclosed in an `if (generatingForIDE)` test, it doesn't have an IDE stub.

There is actually another one like that for the `SourceMapTest`. But since that one isn't used by other code (it's just a test), it doesn't need any IDE stub.

PR #4077, opened 2020-06-10
Fix #4061: Report a linking error on export of non-JS ident in a Script.

### LINKED ISSUE #4061: @JSExportTopLevel("default") generates invalid JS code on Scala.js 1.0
Also reported in https://github.com/scalameta/scalameta/issues/2051

scalameta 4.3.12, published for Scala.js 1.0.1, seems broken on Scala.js 1.x project while running test:

See https://travis-ci.com/github/exoego/scalajs-util-types/jobs/339579547#L441
```
/home/travis/build/exoego/scalajs-util-types/macros/target/scala-2.12/macros-test-fastopt.js:1
let parseSource,default,parseStat;
                ^^^^^^^
```

Scalameta uses `@JSExportTopLevel("default")`, which was fine with Scala.js 0.6, in https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala#L111-

But it seems that Scala.js 1.0 generates JavaScript with a variable named `default` , which is [keyword in ECMAScript 2015](https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Statements/default).

sjrd:
When emitting a Script (`NoModule`), we cannot export on the top-level under a name that is not a valid JavaScript identifier. Such exports must be emitted as `var`s or `let`s, which is only possible for valid JS identifiers.

Previously, attempts at such invalid exports would silently result in invalid .js code, producing a `SyntaxError` at run-time. In this commit, we preemptively report a linking error instead (as a thrown `LinkingException`).

It can be argued that we might want fallback behaviors for these cases, as discussed in #4061. However, this commit is a definitive improvement over the status quo, and leaves the door open for fallbacks in the future, if they prove desirable/necessary.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
@@ -77,9 +77,25 @@ final class Emitter(config: Emitter.Config) {
 
     moduleKind match {
       case ModuleKind.NoModule =>
-        val topLevelVars = unit.classDefs
-          .flatMap(_.topLevelExports)
-          .map(_.topLevelExportName)
+        // #4061 Detect invalid top-level export names for Scripts
+        var hasInvalidTopLevelVar: Boolean = false
+        val topLevelVars = unit.classDefs.flatMap(_.topLevelExports).map {
+          topLevelExport =>
+            val exportName = topLevelExport.topLevelExportName
+            if (!js.ExportName.isValidExportNameInScript(exportName)) {
+              hasInvalidTopLevelVar = true
+              val pos = topLevelExport.pos
+              logger.error(
+                  s"${pos.source}(${pos.line+1}:${pos.column+1}): " +
+                  s"Invalid export as '$exportName' in a script (NoModule) " +
+                  "because it is not a valid JavaScript identifier. " +
+                  "Did you want to emit a module instead?")
+            }
+            exportName
+        }
+        if (hasInvalidTopLevelVar) {
+          throw new LinkingException("There were invalid exports in a script.")
+        }
 
         val header = {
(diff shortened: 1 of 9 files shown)


sjrd:
> I wonder if this is the right spot to check this.
>
> I realize there is precedent with imports.
> 
> It's just unclear to me which checks we perform where. E.g. export collisions, we check in the Analyzer.
>
> My general concern is that we have no error collection infrastructure in the Emitter (or elsewhere, really). As such, we have to keep things together to report "all" errors.

Yes, I was also a bit uneasy about this. As you noticed, we had precedent for imports, so it seemed to be the best place given the status quo.

Rationalizing the status quo, it seems that we report in the Analyzer errors that are independent of the module kind (or other backend concerns), and in the Emitter for things that are module kind-specific. This might be a remnant from the time where we segregated much more which Config went to the frontend and to the backend.

It seems like we should probably report all expected linking errors in the Analyzer.

sjrd:
Updated to resolve the wrong commits.

gzm0:
> It seems like we should probably report all expected linking errors in the Analyzer.

Yes, I feel so too. Module splitting will make the need for this even bigger.

So the options we have here are:
1. Accept this PR as-is, fix it later (I'll probably do so soon).
1. Do it now, rewrite this PR in the new system

Since this PR is quite infrastructure heavy, I'm inclined to go with option 2 (some of the infra changes will become unnecessary IIUC). OTOH, it is basically ready (I haven't reviewed the changes yet, but I really don't expect any problems).

WDYT?

sjrd:
Yes, option 2 is better.

sjrd:
Half of option 2 (fixing #4061, but not moving the existing check from the Emitter to the Analyzer) available in #4081.

gzm0:
I take it we can close this PR?

sjrd:
Yes, indeed.

PR #4079, opened 2020-06-10
Adapt SymbolTest to Dotty & Scala2 common syntax

FabioPinheiro:
Dotty longer supported symbol literal.
Adapt SymbolTest to use application Symbol instead of symbol literal.
Add test cases using symbol literal on SymbolTestScala2.

### DIFF
--- a/test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/scalalib/SymbolTestScala2.scala
+++ b/test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/scalalib/SymbolTestScala2.scala
@@ -1,0 +1,58 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.scalalib
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.testsuite.utils.Platform.scalaVersion
+
+class SymbolTestScala2 {
+
+  /**
+   * This is a Scala 2.x only test because:
+   * Dotty no longer supports symbol literal.
+   */
+  @Test def should_support_symbol_literal(): Unit = {
+    val scalajs = 'ScalaJS
+
+    assertEquals(Symbol("ScalaJS"), scalajs)
+    assertEquals(Symbol("$"), '$)
+    assertEquals(Symbol("$$"), '$$)
+    assertEquals(Symbol("-"), '-)
+    assertEquals(Symbol("*"), '*)
+  }
+
+  /**
+   * This test is similar to the one found in SymbolTest with the same name.
+   * But it uses symbol literals that are not supported on Dotty.
+   */
+  @Test def should_ensure_unique_identity(): Unit = {
+    def expectEqual(sym1: Symbol, sym2: Symbol): Unit = {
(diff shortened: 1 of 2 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/scalalib/SymbolTest.scala
```diff
@@ -23,60 +23,64 @@ class SymbolTest {
...
     val toStringUsesQuoteSyntax = {
       scalaVersion.startsWith("2.10.") ||
       scalaVersion.startsWith("2.11.") ||
       scalaVersion.startsWith("2.12.") ||
       scalaVersion == "2.13.0" ||
       scalaVersion == "2.13.1" ||
-      scalaVersion == "2.13.2"
+      scalaVersion == "2.13.2" ||
+      scalaVersion.startsWith("0.") || // Dotty version start with "0." at the moment
+      scalaVersion.startsWith("3.") // Future scala 3
```
sjrd, line 67: This will become invalid as soon as Dotty upgrades its standard library to 2.13.3. Therefore I think it's better to put this PR on hold until that happens, so we don't introduce more need to change stuff here.

FabioPinheiro: Ok. I saw now that it will change on the next version of standard library.
I'm go remove those to two lines.

test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/scalalib/SymbolTestScala2.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.testsuite.utils.Platform.scalaVersion
+
+class SymbolTestScala2 {
+
+  /**
+   * This is a Scala 2.x only test because:
+   * Dotty longer supported symbol literal.
```
sjrd, line 24:
```suggestion
   * Dotty no longer supports symbol literal.
```

test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/scalalib/SymbolTestScala2.scala
```diff
@@ -0,0 +1,58 @@
...
+    assertEquals(Symbol("ScalaJS"), scalajs)
+    assertEquals(Symbol("$"), '$)
+    assertEquals(Symbol("$$"), '$$)
+    assertEquals(Symbol("-"), '-)
+    assertEquals(Symbol("*"), '*)
+  }
+
+  /**
+   * This test is similar to the one found in SymbolTest with the same name.
+   * But it uses symbol literal that is not supported on Dotty.
```
sjrd, line 38:
```suggestion
   * But it uses symbol literals that are not supported on Dotty.
```

FabioPinheiro:
@sjrd there is an open question about the list of versions in the last test `should_support_toString`
Inicial I am proposing to add this:
```
 scalaVersion.startsWith("0.") || // Dotty version start with "0." at the moment
 scalaVersion.startsWith("3.") // Future scala 3
```

sjrd:
The list of versions is good as is. Once dotty upgrades to 2.13.3 (if that's not already done?), it will correctly answer `false` for `toStringUsesQuoteSyntax`, and will therefore test the `"Symbol(foo)"` output, which is what 2.13.3 implements. So there's nothing to change there.

There's still the typo in the commit message. Once it's fixed, we can merge.

FabioPinheiro:
>Once dotty upgrades to 2.13.3 (if that's not already done?)

I assume not, because that test is still failing. But LGTM

Thanks for the typos fixes!

PR #4081, opened 2020-06-13
Fix #4061: Report a linking error on export of non-JS ident in a Script.

### LINKED ISSUE #4061: @JSExportTopLevel("default") generates invalid JS code on Scala.js 1.0

sjrd:
When emitting a Script (`NoModule`), we cannot export on the top-level under a name that is not a valid JavaScript identifier. Such exports must be emitted as `var`s or `let`s, which is only possible for valid JS identifiers.

Previously, attempts at such invalid exports would silently result in invalid .js code, producing a `SyntaxError` at run-time. In this commit, we preemptively report a linking error instead (as a thrown `LinkingException`).

It can be argued that we might want fallback behaviors for these cases, as discussed in #4061. However, this commit is a definitive improvement over the status quo, and leaves the door open for fallbacks in the future, if they prove desirable/necessary.

### DIFF
--- a/linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala
+++ b/linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala
@@ -29,7 +29,7 @@ import org.scalajs.logging.NullLogger
 import org.scalajs.linker._
 import org.scalajs.linker.analyzer._
 import org.scalajs.linker.frontend.IRLoader
-import org.scalajs.linker.interface.IRFile
+import org.scalajs.linker.interface._
 import org.scalajs.linker.standard._
 
 import Analysis._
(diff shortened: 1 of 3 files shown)


sjrd:
Alternative to #4077 that implements the check in the Analyzer.

This PR does *not* move the existing check for Imports without fallbacks, nor does it add tests for that. So it does not entirely supersede #4077. Doing that is left for another PR, since it was now unrelated to fixing #4061, and not trivial to add due to the fact that Infos do not know about Imports yet.

linker/shared/src/test/scala/org/scalajs/linker/AnalyzerTest.scala
```diff
@@ -537,12 +576,13 @@ object AnalyzerTest {
 
   private def computeAnalysis(classDefs: Seq[ClassDef],
       symbolRequirements: SymbolRequirement = reqsFactory.none(),
+      config: CommonPhaseConfig = CommonPhaseConfig(),
```
gzm0, line 579: Consider taking a `StandardConfig` here. This hopefully makes the `config` helper above unnecessary.

sjrd: Good point. Done.

PR #4083, opened 2020-06-16
Use the published artifacts of scalajs-{logging,js-envs,env-nodejs}.

sjrd:
Now that they are independently published, we simply depend on them through regular `libraryDependencies`.

The sources of `logging` are moved to the `js/` sources of the `linker-interface`, because there is no published version of `scalajs-logging` for Scala.js. We simply keep a copy of the sources rather than fetching them through the source jars because it has a very small footprint and is unlikely to change.

The sources of `js-envs`, `js-envs-test-kit` and `env-nodejs` are deleted.


project/Build.scala
```diff
@@ -943,6 +875,7 @@ object Build {
       mimaBinaryIssueFilters ++= BinaryIncompatibilities.SbtPlugin,
 
       addSbtPlugin("org.portable-scala" % "sbt-platform-deps" % "1.0.0"),
+      libraryDependencies += "org.scala-js" %% "scalajs-env-nodejs" % "1.1.1",
```
gzm0, line 878: Add `"org.scala-js" %% "scalajs-js-envs" % "1.1.1"` as well? Better not to rely on transitive dependencies...

gzm0:
> We simply keep a copy of the sources rather than fetching them through the source jars because it has a very small footprint and is unlikely to change.

I'm not a fan. Mostly not because I'm concerned about changes, but there's nothing that makes it clear / documents that these need to be the same.

sjrd:
Updated. I have added a commit that fetches the sources of `scalajs-logging` instead of keeping a copy of them in this repo.

project/Build.scala
```diff
@@ -699,6 +699,42 @@ object Build {
...
+        val fileSet = FileFunction.cached(cacheDir, FilesInfo.lastModified, FilesInfo.exists) { _ =>
+          s.log.info(s"Unpacking scalajs-logging sources to $targetDir...")
+          if (targetDir.exists)
+            IO.delete(targetDir)
+          IO.createDirectory(targetDir)
+          IO.unzip(jar, targetDir)
+        } (Set(jar))
+
+        fileSet.toSeq.filter(_.getPath().endsWith(".scala"))
+      }.taskValue,
```
gzm0, line 737: Is there a reason this is not the same code than the one to fetch scala sources?

sjrd: * Using `dependencyResolution` directly makes sure that the dependency does not leak in the POM files in any way (putting anything in `libraryDependencies` leaks into the POM). This is not a problem for scalalib because we don't publish that artifact; we only take its .sjsir files and stuff them in another artifact with its own POM.
* It allows the IDEs to understand what's going on. With the way scalalib does it, they wouldn't (but we don't need to navigate in the scalalib so it's fine there).

gzm0: I see. Could we use this approach for the scalalib as well instead? (I understand there are some considerations for the community build there).

sjrd: We could probably use that for the scalalib as well. We're not constrained by the community build anymore, since this repo is not part of it anymore. We have our own nightly-against-nightly at https://github.com/scala-js/scala-js-test-scala-nightly. As long as that works, we're good.

PR #4084, opened 2020-06-16
Move classVar definitions to VarGen

gzm0:
This will become important for module support, since all importing and exporting can be handled in VarGen.


gzm0:
@sjrd not sure you are comfortable with merging this independently.

The [`split-module` branch ][1] shows how this will be used.

[1]: https://github.com/scala-js/scala-js/compare/master...gzm0:split-module?expand=1

sjrd:
> 
> 
> @sjrd not sure you are comfortable with merging this independently.
> 
> The [`split-module` branch ](https://github.com/scala-js/scala-js/compare/master...gzm0:split-module?expand=1) shows how this will be used.

Thanks. I see how it fits. I think it's fine to merge this PR in isolation.

PR #4090, opened 2020-06-18
Fix #4089: Do not `addAnnotation` on `NoSymbol`.

### LINKED ISSUE #4089: Scala.js adds the ExposedJSMember annotation to NoSymbol
As seen in https://github.com/sbt/zinc/issues/718#issuecomment-645841132 and worked around in Zinc in https://github.com/sbt/zinc/pull/798.

Calls to `sym.addAnnnotation` should be guarded with `sym != NoSymbol`. Ideally the compiler would drop the added annotations or give a hard error, but right now it leaves all this up to the caller.

sjrd:
We were blindly doing `sym.accessed.addAnnotation()` in one spot, without first checking that `sym.accessed != NoSymbol`. This was wrong in Scala 2.12+ because fields are created later than the `PrepJSInterop` phase, leading to adding an annotation on `NoSymbol`.

---

I manually checked that this fixes the overcompilation reported in the original ticket https://github.com/sbt/zinc/issues/718, after having upgrade the reproduction to 1.1.1-SNAPSHOT (and reproducing the overcompilation on the latest master).

/cc @retronym

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
@@ -1203,8 +1203,11 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
         sym.addAnnotation(ExposedJSMemberAnnot)
         /* For accessors, the field being accessed must also be exposed,
          * although it is private.
+         *
+         * #4089 Don't do this if `sym.accessed == NoSymbol`. This happens in
+         * 2.12+, where fields are created later than this phase.
          */
-        if (sym.isAccessor)
+        if (sym.isAccessor && sym.accessed != NoSymbol)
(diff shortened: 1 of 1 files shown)


sjrd:
I don't know what a relevant test case would be within this repo.
* We have no infrastructure to test incremental compilation, let alone *over*-compilation managed by Zinc.
* I thought about using a private helper for `addAnnotation` that always asserts that the symbol is not `NoSymbol`, but that wouldn't help if we introduce a new call to `addAnnotation` and forget that we have this helper, so that's basically useless.

gzm0:
> I thought about using a private helper for addAnnotation that always asserts that the symbol is not NoSymbol, but that wouldn't help if we introduce a new call to addAnnotation and forget that we have this helper, so that's basically useless.

Consider contributing this upstream. Putting an annotation on NoSymbol hopefully never makes sense.

PR #4091, opened 2020-06-18
Fix #4088: Avoid an Int overflow in BigDecimal.toString().

### LINKED ISSUE #4088: BigDecimal(1, -2147483648).toString returns "1E-2147483648"
1. Make sure it is a bug for this repo
  * Can you reproduce the bug with Scala/JVM? 
No, I cannot
  * Make sure it is not one of the intended semantic differences: https://www.scala-js.org/doc/semantics.html
It is not a case of semantic differences for sure
  * Does the bug involve macros?
No, it doesn't
  * Is it about the website?
No, it isn't
  * If in doubt, ask on Gitter first: https://gitter.im/scala-js/scala-js
It is a bug, no doubts

2. Make sure we can understand and reproduce the bug
  * Give code *and* reproducible steps to provoke the bug
```
BigDecimal(1, -2147483648).toString
```
  * Give the version of Scala and Scala.js you are using (if relevant, sbt as well)
Reproduced with Scala 2.13.2 + Scala.js 1.1.0 + sbt 1.3.12 
  * What did you expect to happen?
Should return "1E+2147483648"
  * What actually happened?
Actually returns "1E-2147483648"

sjrd:


javalib/src/main/scala/java/math/Conversion.scala
```diff
@@ -266,17 +266,16 @@ private[math] object Conversion {
           }
           result = "0." + result
         }
-      } else if (scale !=0) {
```
sjrd, line 269: Starting from here in this file, the changes are just being a Boy Scout.

Issue #4094, opened 2020-06-20
JSON.stringify does not allow returning undefined


kryptt:
Calling JSON.stringify with undefined, or a function returns undefined in most contexts (browsers)

```
val x = scala.scalajs.js.JSON.stringify(())

println(s"found: $x" )
```
see: https://scalafiddle.io/sf/oFq7bkS/0
to reproduce.

I believe there are two directions to go, either change the return type to UndefOr[String] or document that this function may throw an exception.

ritschwumm:
if it returns undefined in JS, then it should return undefined in scala-js.

standard apis are always described, never extended or changed in behaviour, which is a good thing (tm).

kryptt:
You will notice if you run the fiddle, it does not just return undefined. It will throw a ClassCastException. 

In fact I realized the behavior should be different betwee fastOptJS and fullOptJS. For fastOptJS it actually gets wrapped: `$as_T(JSON.stringify(value))`.

This should probably be treated as a bug.

gzm0:
> if it returns undefined in JS, then it should return undefined in scala-js.

@ritschwumm It you refer to the facade definition, this is not necessarily true. For example `js.Array[T]`'s index access (i.e. `a[2]`) specifies a return type of `T`. However, if you access a JavaScript array out of bounds, it'll return `undefined`.

It all boils down to defining what "valid" usages of a certain API are. [MDN says about `JSON.stringify`][1]:

> undefined, Functions, and Symbols are not valid JSON values. If any such values are encountered during conversion they are either omitted (when found in an object) or changed to null (when found in an array). JSON.stringify() can return undefined when passing in "pure" values like JSON.stringify(function(){}) or JSON.stringify(undefined).

To me, this clearly indicates that the caller is responsible to ensure an allowed value is passed to `JSON.stringify`.

As such, IMHO the only thing to do here is to update the documentation.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

kryptt:
@gzm0  How do you normally handle discrepancies between fastOptJS and fullOptJS ?

gzm0:
@kryptt Have a look at the section "Undefined Behavior" here: https://www.scala-js.org/doc/semantics.html

tl;dr: You do not (after the fact). If it triggers, your code has a bug you need to fix.

[...]

Issue #4085, opened 2020-06-16
Failed to deserialize sjsir files on large project on scalajs 1.1.0


ngbinh:
I see this error on a very large project at work when upgrading from scala.js 0.6.32 to 1.1.0

```scala
[info] Fast optimizing .../target/scala-2.13/scalajs-bundler/main/gondorappclient-fastopt.js
[debug] Linker: 57836 us
[error] java.io.IOException: Failed to deserialize /path-to/target/scala-2.13/classes/com/anduin/stargazer/fundsub/module/common/
FundSubZenDeskWidget$State.sjsir
[error]         at org.scalajs.linker.interface.unstable.IRFileImpl$$anonfun$withPathExceptionContext$1.applyOrElse(IRFileImpl.scala:65)
[error]         at org.scalajs.linker.interface.unstable.IRFileImpl$$anonfun$withPathExceptionContext$1.applyOrElse(IRFileImpl.scala:58)
[error]         at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:38)
[error]         at scala.util.Failure.recover(Try.scala:234)
[error]         at scala.concurrent.Future.$anonfun$recover$1(Future.scala:395)
[error]         at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
[error]         at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
[error]         at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
[error]         at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1425)
[error]         at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290) 
[error]         at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1016)
[error]         at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)
[error]         at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)
[error]         at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
[error] Caused by: java.nio.file.FileSystemException: path-to-dir/FundSubZenDeskWidget$State.sjsir: Too many open files
[...]

PR #4095, opened 2020-06-22
Improve documentation of how JSON.stringify serializes things

gzm0:
This is in response to #4094.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/JSON.scala
+++ b/library/src/main/scala/scala/scalajs/js/JSON.scala
@@ -41,10 +41,32 @@ object JSON extends js.Object {
   def parse(text: String,
       reviver: js.Function2[js.Any, js.Any, js.Any] = ???): js.Dynamic = js.native
 
+  // scalastyle:off line.size.limit
   /**
    * Convert a value to JSON, optionally replacing values if a replacer function
    * is specified, or optionally including only the specified properties if a
    * replacer array is specified.
+   *
+   *   - If the value has a
+   *     [[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior toJSON()]]
+   *     method, it is responsible to define what data will be serialized.
+   *   - [[Boolean]], [[Double Number]], and [[java.lang.String String]] objects are converted to
+   *     the corresponding primitive values during stringification, in accord
+   *     with the traditional conversion semantics.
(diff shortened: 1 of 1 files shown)


gzm0:
@sjrd it might be a good idea to hold off on this until `1.1.1` is out? (I'm working on a fix on #4085, but the testing is a bit involved :)).

library/src/main/scala/scala/scalajs/js/JSON.scala
```diff
@@ -46,6 +46,26 @@ object JSON extends js.Object {
...
+   *     found in an [[Array]]).
+   *   - All [[Symbol]]-keyed properties will be completely ignored, even when
+   *     using the replacer function.
+   *   - The instances of [[Date]] implement the [[Date#toJSON()*]] function by
+   *     returning a string (the same as [[Date#toISOString]]). Thus, they are
+   *     treated as strings.
+   *   - The numbers Infinity and NaN, as well as the value null, are all
+   *     considered null.
+   *   - All the other [[Object]] instances (including Map, Set, WeakMap, and
+   *     WeakSet) will have only their enumerable properties serialized.
```
sjrd, line 67: We should add a paragraph that explains what happens when the top-level argument is a function or `undefined`. Basically a replacement for the MDN doc that says:

> JSON.stringify() can return undefined when passing in "pure" values like JSON.stringify(function(){}) or JSON.stringify(undefined).

In our case we should say that those calls are invalid and will result in Undefined Behavior.

library/src/main/scala/scala/scalajs/js/JSON.scala
```diff
@@ -46,6 +46,26 @@ object JSON extends js.Object {
    * is specified, or optionally including only the specified properties if a
    * replacer array is specified.
    *
+   *   - If the value has a
+   *     [[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior toJSON()]]
+   *     method, it's responsible to define what data will be serialized.
```
sjrd, line 51:
```suggestion
   *     method, it is responsible to define what data will be serialized.
```

sjrd:
> @sjrd it might be a good idea to hold off on this until `1.1.1` is out? (I'm working on a fix on #4085, but the testing is a bit involved :)).

Yes, I agree that we should release a 1.1.1 before merging this PR.

gzm0:
I have isolated the commit that just improves the doc so we can merge it now.

gzm0:
@sjrd any further comments on this? Some of the details changed to make CI happy.

PR #4096, opened 2020-06-22
Restrict type of JSON.stringify.space and improve doc

gzm0:
This is a potentially source incompatible change. It is binary compatible, since native types do not appear at the binary level.


gzm0:
This is a spin-off from #4095 with the changes that should go in after `1.1.1` is released only.

gzm0:
@sjrd is this good to merge?

PR #4097, opened 2020-06-23
Fix #4085: Throttle concurrent file reads

gzm0:

PR #4101, opened 2020-06-26
Expose the IRFileCacheConfig for the global cache in the sbt plugin

gzm0:
Cleanups post #4097 that need a minor version bump.

### DIFF
--- a/project/BinaryIncompatibilities.scala
+++ b/project/BinaryIncompatibilities.scala
@@ -14,6 +14,13 @@ object BinaryIncompatibilities {
   )
 
   val SbtPlugin = Seq(
+      // Changes in LinkerImpl, which is declared that we can break it.
+      exclude[ReversedMissingMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl.irFileCache"),
+      exclude[FinalMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl#Reflect.irFileCache"),
+      exclude[FinalMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl#Forwarding.irFileCache"),
   )
 
   val TestCommon = Seq(
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
@@ -30,7 +30,9 @@ import org.scalajs.linker.interface._
 trait LinkerImpl {
   def clearableLinker(cfg: StandardConfig): ClearableLinker
 
-  def irFileCache(): IRFileCache
+  final def irFileCache(): IRFileCache = irFileCache(IRFileCacheConfig())
+
+  def irFileCache(cfg: IRFileCacheConfig): IRFileCache
 
   def irContainers(classpath: Seq[Path])(
       implicit ec: ExecutionContext): Future[(Seq[IRContainer], Seq[Path])]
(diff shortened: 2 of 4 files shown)


gzm0:
FYI @sjrd this is still open for review, in case you have missed it.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
```diff
@@ -30,7 +30,7 @@ import org.scalajs.linker.interface._
 trait LinkerImpl {
   def clearableLinker(cfg: StandardConfig): ClearableLinker
 
-  def irFileCache(): IRFileCache
+  def irFileCache(cfg: IRFileCacheConfig): IRFileCache
```
sjrd, line 33: This is debatable, but IMO we should preserve existing methods when possible, even in this API. It is an unstable API because we need the ability to *add* methods (or overloads) even though it is an open trait, so we allow ourselves to break classes that *extend* `LinkerImpl`. However, we should still make a decent effort at not breaking *users* of a `LinkerImpl`, so we should not *remove* methods unless necessary.

Therefore, I suggest adding this new overload without removing the 0-arg one.

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardIRFileCache.scala
```diff
@@ -33,8 +33,6 @@ final class StandardIRFileCache(config: IRFileCacheConfig) extends IRFileCacheIm
    * that paying the cost for synchronization is lower than I/O.
    */
 
-  def this() = this(IRFileCacheConfig())
```
sjrd, line 36: This seems like a useful shorthand to keep around. Usually we offer a no-arg constructor as a forwarder for a config-base constructor with the default config. For example all JS envs have a constructor like that.

project/BinaryIncompatibilities.scala
```diff
@@ -8,12 +8,24 @@ object BinaryIncompatibilities {
...
   val SbtPlugin = Seq(
+      // Changes in LinkerImpl, which is declared that we can break it.
+      exclude[DirectMissingMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl.irFileCache"),
+      exclude[ReversedMissingMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl.irFileCache"),
+      exclude[DirectMissingMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl#Forwarding.irFileCache"),
+      exclude[DirectMissingMethodProblem](
+          "org.scalajs.sbtplugin.LinkerImpl#Reflect.irFileCache"),
```
sjrd, line 28: Those changes are in the wrong commit.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
gzm0: Hmmm... But then we should make it a final default method passing a default config?

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardIRFileCache.scala
gzm0: Do you also want to not deprecate the convenience method in `StandardImpl` then?

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/LinkerImpl.scala
sjrd: Yes, it's probably necessary to make it `final` to prevent existing subclasses that overrode it from still linking, since otherwise their override would never be called anymore.

linker/shared/src/main/scala/org/scalajs/linker/standard/StandardIRFileCache.scala
sjrd: Yes, that would make sense.

PR #4103, opened 2020-06-27
Avoid Option(null) from UndefOrOps#toOption

AlexITC:
The `UndefOrOps#toOption` method may return `Option(null)` which is unlikely what users expect, now, it returns `None` instead.


AlexITC:
I have signed the agreement.

sjrd:
As replied on Gitter, this existing behavior is as designed. `null` is not a special value. For `UndefOr`, `null` is a present value, so it is converted to `Some` which is the present value of `Option`.

PR #4104, opened 2020-06-28
Use a js.Block as result from Emitter

gzm0:
js.Blocks are a specialized way of returning a list of trees. As such,
using List[js.Tree] as a way to return multiple trees is inconsistent.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/Trees.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/Trees.scala
@@ -127,21 +127,35 @@ object Trees {
   }
 
   object Block {
-    def apply(stats: List[Tree])(implicit pos: Position): Tree = {
-      val flattenedStats = stats flatMap {
-        case Skip() => Nil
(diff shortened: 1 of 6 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/javascript/Trees.scala
```diff
@@ -127,16 +127,19 @@ object Trees {
...
 
   object Block {
-    def apply(stats: List[Tree])(implicit pos: Position): Tree = {
-      val flattenedStats = stats flatMap {
-        case Skip() => Nil
+    def apply(stats: List[Tree])(implicit pos: Position): Tree =
+      apply(stats.iterator)
+
+    def apply(stats: Iterator[Tree])(implicit pos: Position): Tree = {
+      val flattenedStats = stats.flatMap {
```
sjrd, line 134: The implementation of `flatMap` in `Iterator` is significantly more complex than in `List`. I don't think we should lose the fast implementation when the input is known to be a `List`.

gzm0:
Updated

PR #4106, opened 2020-06-30
Store isParentDataAccessed in LinkingUnit

gzm0:
It is linking information just like hasRuntimeTypeInfo. As such, we
should not recalculate it, but simply store it.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
@@ -33,6 +33,7 @@ trait Analysis {
   import Analysis._
 
   def classInfos: scala.collection.Map[ClassName, ClassInfo]
+  def isParentDataAccessed: Boolean
   def errors: scala.collection.Seq[Error]
 }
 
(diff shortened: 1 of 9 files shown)


sjrd:
I don't know about this. Do we really want to expose the special tracking of `isParentDataAccessed` to users of the `linker.standard.*` API?

If yes, at least we should generalize this boolean flag as a data class `SpecialLinkingInfo`, in case we need to add more flags there in the future.

But I'm not sure this is worth it. Does this help with module splitting?

gzm0:
> Does this help with module splitting?

Yes. In fact, it determines, to some extent, the semantics of the `superClass` field: If `isParentDataAccessed`, the `superClass` is required even if it is native; otherwise it is not.

> I don't know about this. Do we really want to expose the special tracking of isParentDataAccessed to users of the linker.standard.* API?

Well, we probably need to talk about this anyways :) For module splitting, I need to pass on the (pre analyzer) infos for each class in the linking unit.

If you want, I'll just put this on my split-module branch for now as a base commit.

gzm0:
Closing for now. We can continue this discussion on #4111.

PR #4111, opened 2020-07-01
Prototype v2 for Module Splitting

gzm0:
Since the previous PR (#4010), the following has improved:

- The overall architecture is ready for review.
- The test suite passes in CommonJS.


linker-interface/shared/src/main/scala/org/scalajs/linker/interface/LinkerCompat.scala
```diff
@@ -0,0 +1,68 @@
...
+package org.scalajs.linker.interface
+
+import scala.concurrent._
+
+import java.nio.ByteBuffer
+
+import org.scalajs.linker.interface.unstable.{OutputFileImpl, OutputDirectoryImpl}
+
+private[linker] object LinkerCompat {
+  def linkerOutputToOutputSpec(
```
gzm0, line 22: Is this adaptation sufficiently backwards compatible? (it is not 100%).

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ModuleSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+import org.scalajs.linker.interface.unstable.ModuleSpecImpl.SelectorImpl
+
+/** Module output specification. */
+final class ModuleSpec (
```
gzm0, line 18: I have also considered to make a `ModuleSpec` a list inherently (like in an earlier proposal). However, it seems that with the fully automatic module tracking, the input `ModuleSpec` is unlikely to turn into something more automatic.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/OutputSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker.interface
+
+final class OutputSpec private (
+    val directory: OutputDirectory,
+    val jsName: ModuleSpec.ModuleID => String,
+    val sourceMapName: ModuleSpec.ModuleID => String,
```
gzm0, line 18: I'm unsure where to put `jsName` / `sourceMapName`. If I put it here, it is nice for the sbt plugin, since it keeps full control. OTOH, we need more machinery in the sbt plugin to support the "mjs" suffix for node.

Maybe putting these into the config (and then maybe just make them suffix strings due to serializability) and instead return a `LinkingReport` containing the files that were written might be a better idea.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/DependencyAnalyzer.scala
```diff
@@ -0,0 +1,155 @@
...
+    deps ++= coreDeps
+
+    if (clazz.hasRuntimeTypeInfo && unit.isParentDataAccessed)
+      deps ++= classInfos.superClass
+    else if (clazz.hasInstances)
+      deps ++= classInfos.superClass.filter(isNonNative(_))
+        
+      //TODO: This dependency chain fails with this:
+      // java.io.Flushable: java.lang.Object: org.scalajs.linker.runtime.UndefinedBehaviorError: java.lang.Throwable: java.lang.StackTraceElement: key not found: ClassName<java.lang.StackTraceElement>
+      //deps ++= classInfos.referencedFieldClasses
```
gzm0, line 53: Any idea what is happening here?

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MinModuleSplitter.scala
```diff
@@ -0,0 +1,206 @@
...
+
+        staticDeps
+          .map(className => internalModules(visitedNodes(className).compId).id)
+          .toSet
+      }
+
+      val externalDeps =
+        topLevelExports.flatMap(externalDependencies(_)).distinct.toList
+
+      // TODO: Filter empty modules.
```
gzm0, line 113: Unsure what to do here. Should an empty module cause a failure? Should it cause an empty file? Should it cause no file?

linker/shared/src/main/scala/org/scalajs/linker/standard/ModuleSet.scala
```diff
@@ -0,0 +1,45 @@
...
+
+package org.scalajs.linker.standard
+
+import org.scalajs.ir.Names.ClassName
+import org.scalajs.ir.Trees.TopLevelExportDef
+
+import org.scalajs.linker.interface.ModuleInitializer
+import org.scalajs.linker.interface.ModuleSpec.ModuleID
+
+final class ModuleSet private[linker] (
```
gzm0, line 21: An alternative is to evolve the `LinkingUnit` into this class instead. What is unclear is how the `Refiner` would deal with a non-single module `LinkingUnit`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -207,31 +207,32 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
-            val out = LinkerOutput(linkerImpl.outputFile(output.toPath))
-              .withSourceMap(linkerImpl.outputFile(sourceMapFile.toPath))
-              .withSourceMapURI(relURI(sourceMapFile.getName))
-              .withJSFileURI(relURI(output.getName))
+            val out = OutputSpec(linkerImpl.outputDirectory(outputDir.toPath))
+            val moduleSpec = List(
+              ModuleSpec(modName)
+                .withInitializers(moduleInitializers)
+                .withSelector(ModuleSpec.Selector.default)
+            )
```
gzm0, line 219: sbt plugin needs a lot more work. It's not 100% clear to me ATM how we can reasonably support existing builds that assume single-file outputs without introducing a lot of double logic.

gzm0:
Rebased on top of #4120 and addressed the comments in the first commit (@sjrd WDYT about merging that one on its own?).

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
```diff
@@ -47,6 +49,7 @@ object Analysis {
    */
   trait ClassInfo {
     def className: ClassName
+    def data: Infos.ClassInfo
```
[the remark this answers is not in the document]
gzm0: Hmmm... But what is the alternative? We need an info-like thing to determine module dependencies. Are you suggesting we re-calculate this if necessary with another tree traverse?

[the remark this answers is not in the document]
sjrd: We can potentially extend the `Analysis`, but now I see that there is a bigger problem: `LinkedClass`, which is in `standard`, contains references to `Infos` as well. That goes beyond intention, and directly into stabler-references-less-stable territory, which is not good at all.

From what I see, those Infos are needed in `ModuleAnalyzer.staticDependencies`. IIUC, this computes a `dependencies: Set[ClassName]`. Could we compute these `dependencies` eagerly in the `Infos`/`Analyzer`? Then we could expose that `dependencies` in `Analysis` (where it would make sense) and further in `LinkedClass` (which would therefore not depend on a type less stable than itself).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -232,6 +268,68 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
+            }
+
+            // Options passed to the defineProperty method
+            val descriptor = ObjectConstr(
+              getterDef ::
+                (StringLiteral("configurable") -> BooleanLiteral(true)) ::
+                Nil
+            )
+
+            Block(tree, Apply(defProp, List(exportsVarRef, name, descriptor)))
```
sjrd, line 324: Consider refactoring as:
```suggestion
            val export = Apply(defProp, List(exportsVarRef, name, descriptor))
            Block(tree, export)
```
and then you can factor out `val export =` and `Block(tree, export)` outside of the `if (mutable)`.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MinModuleSplitter.scala
sjrd: It seems to me that the most reasonable outcome would be an empty file. If a module has been requested by the user, and it ends up being empty, why should we get rid of it or consider that an error?

linker/shared/src/main/scala/org/scalajs/linker/standard/ModuleSet.scala
sjrd: No, I think what you have now is better. It's better if front-end phases (the optimizer, but also future custom front-end phases) don't have to worry about modules.

project/Build.scala
```diff
@@ -1512,7 +1512,19 @@ object Build {
...
+
+      // HACK: Test Module support.
+      scalaJSLinkerConfig ~= { _
+        .withModuleKind(ModuleKind.ESModule)
+        .withModuleSplitStyle(ModuleSplitStyle.SmallestModules)
+      },
+      jsEnv in Test := {
+        val out = (artifactPath in Test in fastOptJS).value
+        val config = NodeJSEnv.Config()
+          .withEnv(Map("NODE_PATH" -> out.toString))
```
sjrd, line 1525: That seems fishy. Should we emit references to other modules as `./foo.js` instead of `foo.js` to avoid the need for `NODE_PATH`?

linker/jvm/src/main/scala/org/scalajs/linker/backend/closure/ClosureLinkerBackend.scala
```diff
@@ -73,18 +73,26 @@ final class ClosureLinkerBackend(config: LinkerBackendImpl.Config)
...
    *  @param output File to write to
    */
-  def emit(unit: LinkingUnit, output: LinkerOutput, logger: Logger)(
+  def emit(moduleSet: ModuleSet, outputSpec: OutputSpec, logger: Logger)(
       implicit ec: ExecutionContext): Future[Unit] = {
-    verifyUnit(unit)
+    verifyModuleSet(moduleSet)
+
+    require(moduleSet.modules.size == 1,
+        "Cannot user multiple modules with the Closure Compiler")
```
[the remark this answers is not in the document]
gzm0: No. The renaming across the modules needs to be consistent. So it might be possible to make this work, but it seems the main use case we want to address here is to feed these things downstream to webpack or other. So IMO we should not try to support this in a first iteration (probably provide better error messages though :)).

project/Build.scala
gzm0: Maybe. I didn't understand that standard well enough. Is that required to work in every engine? (if yes, we should probably indeed do that).

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MinModuleSplitter.scala
gzm0: > why should we get rid of it or consider that an error?

It might hint at a configuration error. But yes, probably an empty module is OK.

project/Build.scala
sjrd: Nothing is required to work in every engine, if you only read the standard. But if an engine doesn't support explicit relative paths like that, it's unlikely that it supports any kind of reasonable module resolution anyway.

gzm0: Hmmm... But then haven't we taken the stance that we do not want to rely on specifics of engines? Maybe we need to have an additional `moduleName` function (just like `jsNameWithSourceMap`, or have that one return all of the things we need).

gzm0: I think I like this idea much more. We should maybe even do this for the references between source maps and JS file. That way:
- We can provide full backwards compatibility.
- We have a single, simple mechanism to deal with all the oddities of how things reference each other in a given environment.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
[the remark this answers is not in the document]
gzm0: > Could we compute these dependencies eagerly in the Infos/Analyzer?

Yes. The first prototype did exactly that. Downsides:

- It makes the Analyzer even more complicated.
- The Analyzer now contains specifics to modules.
- We calculate it twice but only need it once (if we do not add yet another flag to the `Analyzer`).

I feel like the `dependencies` we calculate in `ModuleAnalyzer` are much more private than the Infos. Notably, they are not independent of other classes and depend on `symbolRequirements`. The Infos on the other hand are a mere summary of the trees of a class.

An alternative to what we have right now, might be to make Info caching more central. As such, both the `Refiner` and the `ModuleAnalyzer` could access Infos (without exposing them in the `LinkedClass`). It is unclear how to properly identify the `LinkedClass` in this case though.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/OutputSpec.scala
gzm0: I'm less and less convinced this is the right spot TBH. I think we should have patterns for the following:

- JS file name
- Source map name
- Source map URI
- JS file URI
- module name (for include)

At that point, it really doesn't feel like this belongs to `OutputSpec`, but rather in a `ModulePatterns` option class in `StandardConfig`. WDYT?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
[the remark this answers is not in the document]
sjrd: If we break down the things that `staticDependencies` computes, there are:

* Things read from `ReachabilityInfo`, among which
  * `instantiatedClasses` and `accessedModules`, which depend on `isNonNative`. Why do we need to know whether it is non-native? Could we get away without this knowledge? If not, we could store them in a separate set that the `ModuleAnalyzer` will filter for non-native classes.
  * Everything else, which do not depend on other classes.
* Things read directly from the `classInfos`, which can also be read from existing fields of `LinkedClass`, AFAICT
* Things depending on the `symbolRequirements`, which are only added to `j.l.Object`, and that the `ModuleInitializer` could patch up when reading the dependencies of `j.l.Object` (after all, that *is* clearly a private implementation detail of the ModuleInitializer).
* The artificial dependency from everything on `j.l.Object`, which the ModuleInitializer can also patch up when reading the dependencies from the `LinkedClass`.

That would remove all the dependencies on other classes and on `symbolRequirements`.

The Infos might be less private in terms of exposed information, but as an API they are very unstable, and clearly an implementation detail of the `Analyzer`.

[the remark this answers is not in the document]
gzm0: > Why do we need to know whether it is non-native?

Because instantiating a native class does not actually need the SJS module that class is in. But the native module that class is in.

This is why we need the distinction between, for example, `instantiatedClasses` and `methodsCalledStatically.keys`. A static method of a native class is still an SJS thing. However, a new instance is not.

> Could we get away without this knowledge?

No.

So in terms of `ReachabilityInfo` this establishes that we need finer granularity than simply a `Set[ClassName]`. The alternative I see here is that we make a difference between a static and an external dependency in the Analyzer. This is not really a breach of abstraction, since it follows from the IR spec (the distinction, what exactly constitutes a static dependency is, AFAICT an implementation detail of the Emitter, so we probably have a breach of abstraction either way).

The upside of this would be:
- No exposed infos.
- No exposed `isParentDataAccessed`

The downside of this would be:
- More complexity in the Analyzer.
- static and external dependencies are calculated pre optimizer for no reason (if it is enabled).
- static and external dependencies are calculated even without module splitting.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/OutputSpec.scala
gzm0: Hmmm... Turns out like this we cannot easily provide a backwards compatible method in `Linker` (because `config` cannot be change at that point).

I'm unsure how to proceed. IMHO as soon as things modify the *contents* of the files, it definitely should go into `config` :-/

sjrd: Perhaps one way to answer those questions: `StandardConfig` is just *one* way of configuring a linker, i.e., a standard linker. Does it make sense for another linker to be configured in a different way for file names? Would the sbt plugin be able to survive a different `scalaJSLinker` implementation that used a completely different scheme for file name configuration?

Today, the sbt plugin considers `scalaJSLinkerConfig` *almost* entirely as a blackbox, only used to create the default `scalaJSLinkerConfig`. The "almost" is because we actually dig into its `moduleKind` field to build `jsEnv` (and having some weird warning that if `moduleKind` differs `in fast/fullOptJS` than just in the configuration, things will break). To me this suggests that `moduleKind` already has a very specific status, and maybe shouldn't have been part of the `scalaJSLinkerConfig` to begin with. If these module patterns would equally affect the rest of the sbt plugin, perhaps they should stay out of `StandardConfig`?

gzm0: Hah, I think this is an excellent way of looking at it.

So what we observe is that the sbt plugin is peeking at the `scalaJSLinkerConfig` to find out which `moduleKind` the output is. It does this without taking the `scalaJSStage` into account, requiring the weird warning.

IMHO a better way to deal with this is a `LinkingReport` providing this information. At this point, the sbt plugin does not need to know/care anymore *why* a certain module type was returned (we could imagine a custom linker that only supports a specific type). It simply knows *which* module type was returned.

A similar strategy can then be applied to output files: The `LinkingReport` will simply list all the file names (and source maps) that have been produced.

This partially also addresses the backwards compatibility issue: We need not care about the config for file names but can
- write to a in memory directory
- retrieve the single output from the `LinkingReport` and write to the actual output

It does not address how to deal with `sourceMapURI` / `jsFileURI`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analysis.scala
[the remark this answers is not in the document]
gzm0: @sjrd would you like me to try this? How I expect this to look is that `LinkedClass` would get a `staticDependencies: Set[ClassName]` and a `externalDependencies: Set[String]` field. (but we should probably introduce a more specific type for `externalDependencies`).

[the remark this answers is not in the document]
sjrd: Yes, I believe that plan sounds good :)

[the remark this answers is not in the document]
gzm0: I have added a new commit with this. I think it is better.

The only thing I do not like is how we handle module initializers. Maybe we should introduce a `LinkedModuleInitializer`? It is still not super nice, that the `Refiner` is re-classifying entry points into modules.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -232,6 +268,71 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
+      tree
+    } else {
+      config.moduleKind match {
+        case ModuleKind.NoModule =>
+          throw new AssertionError("non-leaf module in NoModule mode")
+
+        case ModuleKind.ESModule =>
+          // TODO: Mutable export!
+          val export = Export((ident -> ExportName(ident.name)) :: Nil)
+          Block(tree, export)
```
gzm0, line 294: @sjrd this is another thing we should discuss: Currently, we theoretically allow assigning to any static field from anywhere. Nothing I know of uses this. It is more difficult to do across ES6 modules because we need a dedicated setter function (and hence the trees look entirely different).

We can of course build it, but maybe only allowing to write to a static field from the context of the class it is defined in would be an acceptable squint-with-the-eyes IR change?

Otherwise, how do we test this :)

gzm0:
Open questions at this point:

- Stick with dependency analysis in `Analyzer`? (I vote yes)
- Introduce `LinkedModuleInitializer`? (I vote yes)
- What to do with filename patterns / `OutputSpec`?
- What to do with writing of static fields?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
sjrd: Under optimizations, this is very difficult to ensure. You could have a method that modifies the static field in its class, but if that method is inlined in another class, the assignment will be moved to a different class, and there's not much you can do about it.

> Otherwise, how do we test this :)

With an `@inline def` :p

gzm0: Ah, right. Fair enough :)

gzm0:
I have added a new commit that adds the `moduleSelectors` to the `LinkingUnit` (which is now private to `frontend`). This is IMO a more or less reasonable structure. So no need to introduce a `LinkedModuleAnalyzer`.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -884,7 +901,11 @@ private final class Analyzer(config: CommonPhaseConfig,
          * Note that the classes referenced by static fields are reached
          * implicitly by the call-sites that read or write the field: the
          * SelectStatic expression has the same type as the field.
+         *
+         * We do not need to add this to staticDependencies: The definition
+         * site will not reference the classes in the final JS code.
          */
+        // TODO: Why is this not in subclassInstantiated()?
```
gzm0, line 908: @sjrd know the answer to this?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -849,8 +861,13 @@ private final class Analyzer(config: CommonPhaseConfig,
...
 
       tryLookupStaticLikeMethod(MemberNamespace.StaticConstructor,
-          StaticInitializerName).foreach {
-        _.reachStatic()(fromAnalyzer)
+          StaticInitializerName).foreach { initializer =>
+        initializer.reachStatic()(fromAnalyzer)
+        staticInitializerModule =
+          moduleSelectors.classifyClass(className.nameString)
+        if (staticInitializerModule.isEmpty) {
+          _errors += ???
```
gzm0, line 869: This is problematic for the following:

```scala
object Foo {
  @JSExportTopLevel("myField")
  val x = 1
}

val moduleSpec = ModuleSpec(modName).withSelector(ModuleSpec.Selector.exportName("myField"))
```

This will trigger an error here, because `Foo`'s static initializer (created for the static field) is not classified into a leaf module. In this specific example, this is not a problem, because static initializers are always emitted with their classes and the top level export depends on it.

However, when static initializers are used for reflection, they need to be referenced from some leaf module. Otherwise they never register themselves.

This is a specification problem for module splitting in the presence of static initializers. The easiest fix is to simply disallow classifying exports differently based on their name (but only based on class name). But that seems somewhat arbitrary.

@sjrd WDYT?

gzm0:
Status:

Awaiting input from @sjrd:

- Stick with dependency analysis in Analyzer? (I vote yes)
- How to classify static initializers?

TODO @gzm0:

1. Figure out what to do with filename patterns / OutputSpec?
1. Implement writing of static fields across modules.
1. Make Jenkins pass (including new tests).
1. Ensure incremental writing (i.e. only write file if changed).
1. Ensure stability of module IDs (in `MinModuleAnalyzer`).
1. Complete and cleanup module analyzers.
1. Complete and test sbt integration.
1. Full-self review.

sjrd:
> Stick with dependency analysis in Analyzer? (I vote yes)

Yes, this looks much better :)

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
sjrd: This is much more problematic than that. From first principles, our spec today for static initializers and they are (all) executed when "the module" is loaded. In theory, that would mean that they should all be executed at startup, whatever leaf module is used, and that consequently they become dependencies of *all leaf modules*. (They can be stored in their class' module, that's alright, as long as all leaf modules make sure to depend on them.)

Those semantics are IMO the only correct semantic for reflection, since we cannot know who will want what when. Any other semantics risks creating non-deterministic results for `Reflect` uses. (An alternative, semantically correct as well, would be that the module that holds `Reflect` itself depends on all those static initializers...)

For the module initializer of exports, however, that would kill the separation of top-level exports in several modules, as far as transitive dependencies are concerned (particularly for the "biggest modules" splitting strategy). For those we clearly only want them depended on by their exports.

Perhaps we'll need to split the semantics of our static initializers :s Or rethink how they work entirely.

I'm going to keep thinking about this.

gzm0: I've been thinking about this as well and have been wondering: What if we put the burden of deciding which module something is in on the compiler?

Consequences:

- We'd need a "default module" for backwards compatibility.
- For top-level exports, that would just be another annotation parameter.
- For static initializers, the compiler can make a more informed decision, as it has semantic information about the use of static initializers. (to solve the `Reflect` use case properly, we'd need some form of language to express the modules that need to reach the static initializers... not sure if that works).

sjrd: That could work.

For exports, is there ever a valid use case for something different than where the enclosing class goes? The exports themselves and the static initializer should be in the same module, and the static initializer depends on the class' module accessor. So they'll always end up together anyway.

sjrd: For `Reflect`, to me the natural, semantically correct choice is to go in the same module as `Reflect` itself. Because you need them to run if and only if you use one of the methods of `Reflect`.

gzm0: > For exports, is there ever a valid use case for something different than where the enclosing class goes?

No, I don't think so.

> The exports themselves and the static initializer should be in the same module.

This is no correct.  Splitting exports away from the classes they are defined in is necessary, since other modules might also depend on the class, so the class itself needs to be in an "internal" (i.e. non-leaf) module.

As a consequence, the current scheme allows exports of the same class to go into different leaf modules, because it's "free".

> For Reflect, to me the natural, semantically correct choice is to go in the same module as Reflect itself.

Yes. The question there is: Do we really want to allow a spec of the form: "Wherever X is"? If the answer is yes, it's unclear to me how we would solve backwards compatibility here :-/

On a related note: Your wording suggests that we put the static initializers themselves in the relevant modules. The current design always leaves them with the class, it just force-imports the class form things that "reach" the static initializer. This ensures that a class can never be used without the static initializer running. I think we probably want to keep that. This also avoids the problem of having multiple static initializers (they just might be force-imported form multiple spots).

sjrd: > The current design always leaves them with the class, it just force-imports the class form things that "reach" the static initializer. This ensures that a class can never be used without the static initializer running. I think we probably want to keep that.

Yes, absolutely. I wasn't very clear indeed. The current design is good.

> The question there is: Do we really want to allow a spec of the form: "Wherever X is"? If the answer is yes, it's unclear to me how we would solve backwards compatibility here :-/

Our existing static initializers follow a simple pattern. We can have a simple, if gross, hack to transition them.

gzm0: So that would basically amount to introducing a "reverse dependency". Where a class arbitrarily could specify that another one "depends" on it.

I'm wondering if we should introduce "registration" as a paradigm. So instead of baking this into the static initializer, we could have a "registerTo" thingy. This would, in the future, allow us to even eliminate these dependencies if they registry isn't reachable.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/OutputSpec.scala
gzm0: How about something like this?

```scala
final class OutputPatterns private (
    val jsFile: String,
    val sourceMapFile: String,
    val moduleName: String,
    val jsFileURI: String,
    val sourceMapURI: String
) {
  def this() = {
    this(
        jsFile = "[module-id].js",
        smFile = "[js-file].map",
        moduleName = "./[js-file]",
        jsFileURI = "[js-file]",
        smFileURI = "[sm-file]"
    )
  }

  // snip
}
```

This is inspired by webpack's placeholders. We could make this fully backwards compatible except:
- There would always be a link between source-map / js file.
- `jsFileURI` and `sourceMapURI` of a given `LinkerOutput` are set to the same (non-empty) value.

We'd have to combine this with a `LinkingReport` such that the sbt-plugin can recover root files and types.

@sjrd WDYT?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
sjrd: It's difficult to imagine what a registration feature should look like without having at least a second use case. But that seems like an idea to explore.

gzm0: Java's SPI comes to mind: https://docs.oracle.com/javase/tutorial/ext/basics/spi.html (also see https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)

gzm0:
I have rebased this on top of #4139 and #4141 and added a commit that adds proper global ref tracking to `VarGen`.

Hopefully this clarifies some of the changes in the base PRs.

gzm0:
@sjrd I have added a new commit with an additional parameter to `JSExportTopLevel` and a "reverse dependency" feature. I think this is a much neater approach. Since it changes the story on backwards compatibility w.r.t. to the output patterns, I wanted to get your input first, before proceeding with that one.

gzm0:
Regarding the commit "Remove ModuleSpec": I think it is definitely an improvement, mostly because the user interface is much smaller without less functionality. Things I'd like to talk about:

# ModuleInitializers API

Is the `API` for `moduleInitializers` good? The alternative is to "bake" the module into the initializer itself. This gives us a more opaque API, but somehow fuses two concepts (the code and where it goes). What I do not like about the current API, is that one can pass `Map("my_module" -> Nil)`, which doesn't make any sense IMO.

The options are:
1. Keep as is.
1. Add `withModule` to `ModuleInitializer`.
1. Add a wrapper type around `ModuleInitializer`.

Option 2 seems to be the pragmatic one; it'll also cause less trouble in the sbt plugin.

# InjectedReverseDependencies

`injectedReverseDependencies` is the minimal mechanism that allows us to address the `Reflect` use case of static initializers. It would even allow us to DCE things if `Reflect` itself isn't reached.

However, the abstraction doesn't compose: If a class that has a static initializer due to a field export also happens to register with `Reflect`, we must force-reach `Reflect` (for no reason).

Further, it causes `Reflect` and the thing it instantiates to mutually depend on each other for no reason. This in turn bundles all things into one huge single module. This might be a problem (think: all actors *must* be in the same module).

We could fix this by introducing something like a `foreignStaticInitializer`: A static initializer attached to *another* class by a class. Post-loading, this would be completely independent of the original class. As a result, `Reflect` would have one big static initializer that registers all its classes. A class merely reachable from `Reflect` would not depend on `Reflect` itself anymore.

This primitive would also allow us to use dynamic loading for reflect (with an appropriate user API of course). OTOH, it probably quite complicates the incremental pipeline :-/

# ModuleName

The name `ModuleName` feels misleading since it is *not* an ES module-name, but an internal ID we are able to convert to a module-name. OTOH, I felt that we need a strongly typed abstraction for it and the `Name` infrastructure seemed to fit the bill. Maybe we should introduce both `ModuleID` and `ModuleName` and actually use them as such (also in the trees related to `import`).

gzm0:
I have added a new commit with the "baked-in" ModuleInitializer.

gzm0:
PR #4146 contains an alternative proposal to `injectedReverseDependencies`. It does not have the shortcomings discussed in the previous comment.

sjrd:
> Further, it causes `Reflect` and the thing it instantiates to mutually depend on each other for no reason. This in turn bundles all things into one huge single module. This might be a problem (think: all actors _must_ be in the same module).

This has been bothering me for a few days. I've come to the conclusion that it's worse than what you describe. If we take the Actors use case, the problem is that *by construction*, most Actors (i.e., all the actors that spawn other actors) will transitively depend on `Reflect` no matter what we do about static initializers. That's because they will call methods that eventually call `Reflect` to instantiate the spawned Actor. There is no way around that.

Therefore, if we add a dependency from `Reflect` to all Actors (either through `injectedReversedDependencies` or your proposal for foreign static initializers), we will have cycles that force all non-leaf Actors to be in the same module as `Reflect`. Only leaf actors (those that never spawn other actors) can be in other modules that `Reflect` depends on.

Based on that, I think we have to throw away solutions that create a dependency from `Reflect` to the things that are reflectively instantiable.

---

Now if we accept that, then I propose the following: have some notion of "triggered" static initializer. Similarly to foreign static initializers, they mention a `className` that they are "attached to". However, whereas foreign static initializers must be injected in the same module as the target, a triggered static initializer would tell the linker this: "for every user-visible module that transitively depends on the target's module, that [user-visible] module must also transitively depend on (call) this static initializer".

For example, an actor `A` declares that its static initializer `<stinit>` to be triggered on target class `Reflect` (possibly among others). `A` ends up in module `a.js`, while `Reflect` ends up in module `reflect.js`. Some user-visible (leaf/root, I never remember which is which) module `user.js` ends up in the graph in a way that it transitively depend on `reflect.js` -> that causes the emitter to add an `import 'a.js'` in `user.js`, so that loading `user.js` will necessarily execute the static initializer of `A`.

This breaks the previous cycle, since `reflect.js` does not depend on `a.js`, while ensuring that any code that reaches `Reflect` can only be actually executed after the static initializer of `A` has been run.

Such triggers would be atomic (containing no `Tree`s): they would simply be a `triggeredBy: List[ClassName]` in the `ClassDef`. In the above example, the `ClassDef` for `A` would include `Reflect` in its `triggeredBy`. That is a very simple design, I think, which should introduce virtually no complexity.

gzm0:
I think `triggeredBy` absolutely makes sense (even if it will complicate the `ModuleAnalyzer` a bit, but I think we can live with that). What I'm still unsure about, is whether we should do this for "static initializer chunks" or for a single static initializer.

Not having chunks will lead to things like: A reflectively instantiatable class with a top level field export will reach `Reflect` which in turn will trigger all reflectively instantiatable classes without any reflection happening.

I'll adjust this PR to `triggeredBy` and then we can see if we want to try to adjust #4146.

sjrd:
I'm not sure the reflectively instantiatable class with top level field export is such a realistic example that we have to optimize for it.

That said, we could prevent the issue by specializing `triggeredBy` a bit more, saying that it's not just a `ClassName`, but a `(ClassName, MemberNamespace, MethodName)`, i.e., that it represents a specific method rather than a class. That way a static initializer for a reflectively instantiatable class would not trigger all other reflectively instantiatable classes (it reaches `Reflect.register...`, but only `Reflect.lookup...` would trigger the others).

This also solves a fragility I was thinking about last night: what if the optimizer is so good that it ends up removing a specific target class, while in fact leaving some code relying on the capability that this class needs triggers for? In that case we might miss static initializers that are required.

There exists a similar problem with `java.lang.Class.getSuperclass`. For that one, we marked that method as `@noinline` to prevent the issue from happening. I think we could do the same here in an automatic fashion: automatically mark all methods that are the target of a triggered by as not to be inlined. We couldn't do that if a class as a whole is the trigger.

sjrd:
Or, instead of hard-coding a specific method as trigger (which can prevent refactorings or additions of capabilities in the target class, e.g., `Reflect`), we could add an indirection: have the methods in `Reflect` declare that require a certain "capability" (an identifying string), and have the static initializers of reflectively instantiatable classes declare that they are triggered by that capability.

For example, `Reflect.lookupInstantiatableClass` would declare a dependency on the capability `"lookup-instantiatable-classes"`, and the static initializers of classes (not module classes) would declare that they are triggered by that same capability.

We could use the same mechanism for `getSuperclass`, with the special case that it is a linker-provided capability.

We could even lift this in user-space with two annotations (this could be added in the future; it doesn't have to be now):
```scala
object Foo { // or could be class
  @requiresCapability("my-capability")
  def useMyCapability(...) = ...

  def registerForMyCapability(cls: Class[_]): Unit = ...
}

object Bar { // must be object so that the method is static
  @triggeredByCapability("my-capability")
  private def registerAsMyCapability(): Unit =
    Foo.registerForMyCapability(this.getClass())
}
```

gzm0:
So on first thought, I really liked your capabilities proposal. It seemed to unify some concepts nicely. However, I realized that the propagation semantics along the module graph are vastly different. This makes me unsure.

For now, I propose the following to unblock this: Require that all static initializers be reached by all leaf modules. This is:
- Simple
- Correct
- Probably sufficient

There is the problem of top level field exports "leaking" into other leaf modules. But for now, we might just want to call this a known issue with the workaround to export an accessor instead.

Note that by doing so, we are not changing or committing to stronger user visible and/or IR semantics. As such we are not making the situation worse (or in other words, we can fix it later at the same cost).

WDYT?

sjrd:
That sounds like a great solution for now, actually. I had not thought of that.

I think ultimately we'll want to address the issue more profoundly, especially for static field exports, but as you said we can do that later without breaking anything.

:+1:

gzm0:
Status:

Requesting input from @sjrd:

- `ModuleInitializer` design: I don't like that the top-level class is now more of a config class. I was hoping to use `ModuleID` internally.
- `ModuleID` design: Is it OK to use `Name` for this?
- `OutputPatterns` design: I don't like how verbose it is. Currently setting `OutputPatterns` for an `.mjs` extension is terrible.
- `Linker.Report` design: I like how this allows the sbt plugin to treat the `scalaJSLinkerOptions` as pretty much opaque.
- Incremental writing design: Is it done at the right level? Is it OK to not use sbt's `FileFunction` anymore?
- With `TopLevelExports` we now have a choice to allow the same name in multiple modules. I think this is problematic because we don't know what to do in `NoModule` mode anymore. So I vote to disallow this.

TODO @gzm0:

- Implement writing of static fields across modules.
- Make Jenkins pass (including new tests).
- Ensure stability/uniqueness of module IDs (in MinModuleAnalyzer).
- Complete and cleanup module analyzers.
- Complete and cleanup compiler (TopLevel export machinery).
- Complete and test sbt integration.
- Full-self review.

gzm0:
> If we follow that spirit, I believe that what is now ModuleID should simply be a String.

I agree with the reasoning. Two concerns:

- some of the linker code will be too stringly typed in this case.
- where should the `DefaultModuleID` go.

First point could be addressed by putting `ModuleID` into `ModuleSet` (as `AnyVal`). Second point would need it to be part of  `ir` since it's the only artifact available to both the compiler and the linker. Alternatively, we could just define it as "convention" and not define a constant (not a fan).

> But perhaps with my above comment that ModuleID should probably just be String, that makes this point moot?

Yes. Or at least dependent on it.

> Should there be a notion of a "main" module that is explicitly part of the report?

No, I don't think so. The "main" module should probably remain a concept private to sbt.

> Disallowing that would impose a restriction that users might not be able to work around if it is what they need. I think it should only be an error when we are in NoModule, the same way that having an @JSImport without global fallback is only an error in NoModule.

Fair enough.

> Incremental writing

I'll follow up later.

sjrd:
> where should the `DefaultModuleID` go.

Currently there doesn't seem to be a `DefaultModuleID` in the PR. Where will we need one and for what reason? Is it for `@JSExportTopLevel` without a second argument (and for deserializing existing `TopLevelExportDef`s)? If yes, should we give delayed semantics to "no module was specified in `@JSExportTopLevel`", i.e., specify what module that is at link time? If that is a viable option, we wouldn't need a `DefaultModuleID` in `ir/`.

gzm0:
> Currently there doesn't seem to be a DefaultModuleID in the PR

Oh, sorry. That is still in a WIP branch.

>  Is it for @JSExportTopLevel without a second argument (and for deserializing existing TopLevelExportDefs)?

Yes. Also for the default value of `ModuleInitializer#module`.

> If yes, should we give delayed semantics

That is tricky if `JSExportTopLevel`s are overloaded. For example, what does this mean:

```scala
object Foo {
  @JSExportTopLevel("foo", "main")
  def foo: Int = ???

  @JSExportTopLevel("foo")
  def foo(x: Int): Int = ???
}
```

With late semantics, this would fail at link time, because `("foo", "main")` is defined twice. With eager semantics, this will overload `("foo", "main")`.

sjrd:
What's the relationship between a `ModuleID` and the ultimate file name being produced? Does `main` always correspond to `main.js` (or `main.mjs`)? If yes, does `main` even make sense as a default, and how do we reconcile that with the default file name for Scala.js codebases, which is `nameoftheproject-fastopt.js`?

gzm0:
> What's the relationship between a ModuleID and the ultimate file name being produced?

The `ModuleID` is applied to the `jsFile` pattern. That will be the file name.

>  If yes, does main even make sense as a default,

Not sure. But *not* having a constant for this will probably complicate things a lot.

> and how do we reconcile that with the default file name for Scala.js codebases, which is nameoftheproject-fastopt.js

I would not. A "new-style" linker will write to a directory, not a single file. I was going to name the directory `nameoftheproject-fastopt/` (like in this PR). If full backwards compatibility is needed, the `fastOptJS` / `fullOptJS` tasks could copy the relevant file iff there is only a single output module. (as discussed previously, we'll probably need two new tasks for invoking the linker).

gzm0:
> How do you even create a non-default OutputPatterns?

The missing `with` methods were an oversight. I have added a factory method as you suggested.

gzm0:
> I believe the emitter should know in advance whether it's going to produce the same result as last time for a particular file, and in that case it should avoid emitting it entirely. I'm not sure how realistic that is, in particular across different instances of the Linker (e.g., across two sbt sessions).

I don't think this is realistic. Especially if we consider different implementations of the linker (think post upgrade).

I don't see what is wrong with simply comparing the file we're about to write. The reason this is currently in the file abstraction itself is because that one also takes care of atomic moves. That is not set in stone (but it would increase the API surface and potentially make the JS implementation trickier).

> Getting rid of the FileFunction is dangerous. Currently if nothing has changed in the inputs, we completely bypass the Linker. This is very important, because fastOptJS is often a dependency of larger task graphs, and if you only change your server, you shouldn't have to go through the entire Linker pipeline only to discover that all the files to be emitted are the same. So we need a fast path to avoid everything if no output has changed.

Good point. I didn't think about the scenario of where the linker is restarted entirely, but nothing has changed.

The major challenge here IMO is that we need to serialize the report. This is not easy once we consider compatibility between different linker versions.

> Alternatively, we could emit an additional report file, that is there only to be the one output file that we use for timestamping. In that case, we would be able to keep using FileFunction with that report file being the only output file that FileFunction looks at. IMO that seems like the easiest path.

We need the report file anyways. Otherwise the key's result type will be `Optional[Linker.Report]` which is neither nice, nor feasible for downstream stuff (e.g. `jsEnvInput`, that needs to know the module).

gzm0:
@sjrd would it be OK if I squash the commits in this PR? Rebasing is becoming a pain, and I'm not sure they are still meaningful. (If they are to you, I'll happily leave them)

sjrd:
No, it's OK. Feel free to squash them.

sjrd:
For the default module, with your latest comments, I guess `main` is good. I think it would be fine to have it as constant `String` in the `ir` project. It might have to be wrapped in `ModuleID` type in the linker if that simplifies things.



> I don't see what is wrong with simply comparing the file we're about to write. The reason this is currently in the file abstraction itself is because that one also takes care of atomic moves. That is not set in stone (but it would increase the API surface and potentially make the JS implementation trickier).

Thinking more about this, I believe you're right. It's fine to have this in the file abstraction.

> Good point. I didn't think about the scenario of where the linker is restarted entirely, but nothing has changed.
> 
> The major challenge here IMO is that we need to serialize the report. This is not easy once we consider compatibility between different linker versions.

The issue between different linker versions can be solved easily, I think: serialize the version number first, and when we deserialize, if the version number is not exactly the same as the current version, return `None`, saying that the report cannot be deserialized. When we get `None`, we force relinking. This makes sense because, if I change the version of the linker, I can end up with a different .js file anyway (more optimizations, different encoding, etc.).

> We need the report file anyways. Otherwise the key's result type will be `Optional[Linker.Report]` which is neither nice, nor feasible for downstream stuff (e.g. `jsEnvInput`, that needs to know the module).

Indeed.

gzm0:
Status:

Requesting input from @sjrd:

- OutputPatterns: Should we make accessing the fields part of the unstable API (or even hide them)? This would allow for more complex implementations later if necessary.

TODO @gzm0:

- Make `ModuleID` AnyVal, internal to linker, use strings elsewhere.
- Incremental writing: Bring back FileFunction, serialize report.
- Implement writing of static fields across modules.
- Allow TopLevelExports with same names in different modules.
- Make Jenkins pass (including new tests).
- Ensure stability/uniqueness of module IDs (in MinModuleAnalyzer).
- Complete and cleanup module analyzers.
- Complete and cleanup compiler (TopLevel export machinery).
- Complete and test sbt integration.
- Full-self review.

sjrd:
> OutputPatterns: Should we make accessing the fields part of the unstable API (or even hide them)? This would allow for more complex implementations later if necessary.

If we can, I guess the more hidden, the better. It's easy to expose later if we need it. It's harder to hide/remove stuff that we exposed previously.

gzm0:
TODO @gzm0:

- OutputPatterns: Make accessing the fields part of the unstable API (or even hide them)
- Restrict what a valid module ID is ("   " probably shouldn't be one)
- Incremental writing: Bring back FileFunction, serialize report.
- Implement writing of static fields across modules.
- Allow TopLevelExports with same names in different modules.
- Make Jenkins pass (including new tests).
- Ensure stability/uniqueness of module IDs (in MinModuleAnalyzer).
- Complete and cleanup module analyzers.
- Complete and cleanup compiler (TopLevel export machinery).
- Complete and test sbt integration.
- Full-self review.

gzm0:
@sjrd I'm unsure about the names for the new sbt tasks. Ideally we don't change the stage names (just to minimize changes). But that makes it hard to find new names :-/ Ideas? (see last commit of this PR for a proposal I don't really like).

sjrd:
Perhaps `fastLinkJS` and `fullLinkJS`? I'll keep thinking.

gzm0:
Closing in favor of #4198.

PR #4112, opened 2020-07-01
FF ONLY Towards 1.1.2.

sjrd:


sjrd:
I'll push this now because I won't have access to a computer for the next 4 days, and it's annoying to stay with a version commit on the tip of a branch for so long.

PR #4113, opened 2020-07-02
[WIP] Move some ClassTagTest checks to scala 2.x only

FabioPinheiro:


FabioPinheiro:
I assume that `classOf[Nothing]` and `classOf[Null]` was disabled on dotty. But that is not true
@smarter mentioned it may be disabled in the future but it still enabled.

ClassTagTest was running fine on my local branch. Now after upgrading to scalajs 1.1.1 and rebasing dotty the test no longer compiles. I will have another look to try to figure out why

sjrd:
This test class passes on dotty's master now, so I think we can close this.

PR #4115, opened 2020-07-03
Fix #4114: Correctly identify doubly nested JS objects

### LINKED ISSUE #4114: class containing nested objects yield compiler exception
ScalaJS v 1.1.1

```
@JSExportTopLevel("OuterClass")
class Outer extends js.Object {
  object middle extends js.Object {
    object innerMost extends js.Object {
    }
  }
}
```

throws

```
error] ## Exception when compiling 1 sources to /home/swachter/projects/eigene/scala-ts/explore/target/scala-2.13/classes
[error] java.lang.AssertionError: assertion failed: 
[error]   object Outer$middle$innerMost at source-/home/swachter/projects/eigene/scala-ts/explore/src/main/scala/scalajs.scala,line-304,offset=7603: jsClassValue.isDefined = true but isInnerNonNativeJSClass = false
[error]      while compiling: /home/swachter/projects/eigene/scala-ts/explore/src/main/scala/scalajs.scala
[error]         during phase: jscode
[error]      library version: version 2.13.2
[error]     compiler version: version 2.13.2
[error]   reconstructed args: ... omitted ...
[error] 
[error]   last tree to typer: TypeTree(object Outer$middle$innerMost)
[error]        tree position: line 304 of /home/swachter/projects/eigene/scala-ts/explore/src/main/scala/scalajs.scala
[error]             tree tpe: Outer$middle$innerMost.type
[error]               symbol: object Outer$middle$innerMost
[error]    symbol definition: class Outer$middle$innerMost extends Object (a ModuleClassSymbol)
[error]       symbol package: <empty>
[error]        symbol owners: object Outer$middle$innerMost
[error]            call site: object Outer$middle$innerMost in package <empty>
[error] 
[error] == Source file context for tree position ==
[error] 
[error]    301 @JSExportTopLevel("OuterClass")
[error]    302 class Outer extends js.Object {
[error]    303   object middle extends js.Object {
[error]    304     object innerMost extends js.Object {
[error]    305     }
[error]    306   }
[error]    307 }
```

If `innerMost` object is removed then it compiles without exception.

gzm0:


compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -6412,8 +6412,10 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
   def isNonNativeJSClass(sym: Symbol): Boolean =
     !sym.isTrait && isJSType(sym) && !sym.hasAnnotation(JSNativeAnnotation)
 
-  def isNestedJSClass(sym: Symbol): Boolean =
-    sym.isLifted && !sym.originalOwner.isModuleClass && isJSType(sym)
+  def isNestedJSClass(sym: Symbol): Boolean = {
+    sym.isLifted && isJSType(sym) && (
+        !sym.originalOwner.isModuleClass || sym.originalOwner.isLifted)
```
sjrd, line 6417: Is that third condition even necessary? Could we stop at
```scala
sym.isLifted && isJSType(sym)
```
?

gzm0: IDK. More precisely, I don't know what "lifted" means. For example:

```scala
object A {
  object B
}
```

Is `B` going to be "lifted"?

gzm0: Answer: Yes it is needed. For example, this would cause an assertion error otherwise:

```
object A {
  class Foo extends js.Object
}
```

`Foo` is considered "lifted" but not "nested".

gzm0: By that logic, the following should fail, but doesn't:

```scala
  object TriplyNestedInnerObject_Issue4114 extends js.Object {
    object middle extends js.Object {
      object inner extends js.Object {
        var x = 1
      }
    }
  }
```

gzm0: I have added a test case for this. Full disclosure: I'm mostly experimental coding at this point.

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/NestedJSClassTest.scala
```diff
@@ -423,6 +423,29 @@ class NestedJSClassTest {
...
+
+    outer2.middle.inner.x = 10
+
+    assertEquals("object", js.typeOf(outer1.middle))
+    assertEquals(1, outer1.middle.inner.x)
+    assertEquals(10, outer2.middle.inner.x)
+  }
+
+  @Test def triplyNestedObject_issue4114(): Unit = {
+    val obj = TriplyNestedObject_Issue4114
```
sjrd, line 438:
```suggestion
    val obj = TriplyNestedObject_Issue4114.asInstanceOf[js.Dynamic]
```
Otherwise I think you're mostly testing the Scala typechecker, rather than our own code.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: Perhaps the `TriplyNestedInnerObject_Issue4114` case doesn't fall because we don't even get to the code path that does that assertion. In that case it's a static module, so there is no `New` for it anywhere.

gzm0: Yes, I guess that is why. Note that this is a real problem, since we are using `isNestedJSClass` to determine how to invoke super calls. (we also use it for constructor exports, but IIUC, that is only an assert/fast-path and could be replaced by looking at the actual parameters).

So I expect code like this to break:

```scala
object A extends js.Object {
  object B extends js.Object {
    class C extends js.Object {
      def foo = 1
    }
  }
}

class D extends A.B.C {
  def foo = super.foo + 1
}
```

gzm0: Strangely, it doesn't fail :S

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/NestedJSClassTest.scala
```diff
@@ -628,6 +656,29 @@ object NestedJSClassTest {
...
+    }
+  }
+
+  object TriplyNestedObject_Issue4114 extends js.Object {
+    object middle extends js.Object {
+      object inner extends js.Object {
+        var x = 1
+      }
+
+      class innerClass extends js.Object {
```
sjrd, line 673: Naming convention nitpick:
```suggestion
      class InnerClass extends js.Object {
```

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
gzm0: What I do not understand is why this line doesn't take the true branch in the test case:
https://github.com/scala-js/scala-js/blob/v1.1.1/compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala#L675

gzm0: Hmm... It seems that an object inside an object is not lifted whereas a class inside an object is lifted...

gzm0:
Turns out we had the appropriate helper for this already. PTAL

gzm0:
> Oh that makes a lot of sense!

Yes. I'm happy I found it. I prefer it a lot to: "This fixes the problem, we don't know why, but we should probably merge it because it fixes the problem...".

PR #4119, opened 2020-07-10
Support ESModules in testHTML

gzm0:
As a side effect, we avoid tmp files and relativize all paths. This is
to make it easier to serve the HTML page through a webserver.

### DIFF
--- a/test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
+++ b/test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
@@ -12,10 +12,11 @@
 
 package org.scalajs.testing.adapter
 
-import java.io.{File, InputStream}
-import java.net.URI
+import java.io.{File, IOException}
+import java.nio.CharBuffer
 import java.nio.charset.StandardCharsets.UTF_8
-import java.nio.file.{Files, StandardCopyOption}
+import java.nio.file._
+import java.nio.file.attribute.BasicFileAttributes
 
 import sbt.testing.{Framework, TaskDef}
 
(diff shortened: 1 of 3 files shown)


test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
```diff
@@ -114,4 +99,20 @@ object HTMLRunnerBuilder {
...
+      while (in.read(buf) != -1) {
+        buf.flip()
+        builder.append(buf)
+        buf.clear()
+      }
+      builder.toString
+    } finally {
+      in.close()
+    }
+  }
```
gzm0, line 117: I really do not like this verbose, low-level thing here. The only alternative I found was to write to a jimfs file first (and read back). That would make use depend on jimfs.

sjrd: In Java 9 that would be
```scala
val in = getClass.getResourceAsStream("test-runner.css")
try new String(in.readAllBytes(), UTF_8)
finally in.close()
```
but sticking to JDK 8 APIs, I believe the existing version is best.

test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
```diff
@@ -26,85 +26,70 @@ import org.scalajs.testing.common._
...
-      tmpFile(name, getClass.getResourceAsStream(name))
+    def scriptTag(tpe: String, content: Path) = {
+      tryRelativize(content).fold {
+        s"""
+        <script type="$tpe">
+          ${new String(Files.readAllBytes(content), UTF_8)}
+        </script>
+        """
+      } { relPath =>
+        s"""<script type="$tpe" src="${htmlEscaped(relPath.toString)}"></script>"""
```
sjrd, line 55: Unfortunately, this simple strategy will not honor the guarantee of executing all the `Input`s in the specified order. The problem is that ES modules are executed in `defer` mode by default (and this cannot be changed, other than to `async`) whereas inline scripts and non-module src scripts are executed in synchronous mode by default.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

So if the input is like
```
List(ESModule(file1), Script(file2))
```
with the above strategy, `file2` will be executed first, and `file1` afterwards.

We can force non-module src scripts to execute in `defer` mode with the `defer` attribute of the `<script>` tag. So that would easily solve the problem as long as there are no inline script.

However, inline scripts cannot be executed in `defer` mode. AFAICT, if the input is like my example and `file2` is emitted as an inline script, there does not exist any combination of option that will force execution of `file1` before `file2`.

So, unfortunately, I think that prevents us from using inline scripts in the general case. We have to fall back to emitting temp files.

test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
```diff
@@ -26,85 +26,70 @@ import org.scalajs.testing.common._
...
-    val ser = Serializer.serialize[TestBridgeMode](mode)
-    s"""var __ScalaJSTestBridgeMode = "${escapeJS(ser)}";"""
+    val htmlContent = s"""
+      <!DOCTYPE html>
+      <html>
+        <head>
+          <title>${htmlEscaped(title)}</title>
+          <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+          <style>
+            ${loadCSS()}
```
sjrd, line 81: :+1: for inlining the style sheet.

test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
gzm0: > We have to fall back to emitting temp files.

Hmm... But then we have the problem of where to put them :-/ If modules are used, everything needs to be served via HTTP, otherwise the browser simply disallows it.

So either we move everything to tmp, maybe even serving it ourselves from sbt.

Maybe the best option is to require a target directory for the HTML runner. That would then be the output with all the files in it (probably not even tmp at this point).

sjrd: > Maybe the best option is to require a target directory for the HTML runner. That would then be the output with all the files in it (probably not even tmp at this point).

Yes, that's probably the best option, indeed.

test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
```diff
@@ -26,85 +27,87 @@ import org.scalajs.testing.common._
...
-  def writeToFile(output: File, title: String, input: Seq[Input],
+  def write(output: Path, artifactsDir: Path, title: String, input: Seq[Input],
       frameworkImplClassNames: List[List[String]],
       taskDefs: List[TaskDef]): Unit = {
 
-    val jsFiles = input.map {
-      case Input.Script(script) => script
+    def artifactPath(name: String): (String, Path) = {
+      val path = artifactsDir.resolve(name)
+      (output.getParent().relativize(path).toString(), path)
```
sjrd, line 61: On Windows, this produces `\`s in the relativized path, which are not really valid in HTML. We can avoid this issue by relativize the paths' respective URIs:
```suggestion
      (output.getParent().toUri.relativize(path.toUri).toString(), path)
```

gzm0: Ugh... Right. That approach has two problems:

- It doesn't fail if relativization doesn't work (`Path#relativize` OTOH does).
- It doesn't relativize to the parent directory (which is likely the primary use case if somebody changed `artifactPath in testHtml` manually).

I couldn't figure out a way to fix this on the quick :-/ Need to think more.

sjrd: I you want to keep using `Path.relativize`, then you can simply do `.replace('\\', '/')`.

gzm0: Updated.

PR #4123, opened 2020-07-13
Introduce the distinction between static and class initializers.

sjrd:
Previously, a unique concept of static initializer was used for two different purposes:

* In Scala classes, code that is executed when initializing the module, which is always reachable.
* In JS classes, code that is executed when the class value is created, which happens lazily the first time it is accessed.

We now separate those two behaviors in two different concepts:

* Static initializers are executed when initializing the module, and are always reachable. They are valid in all kinds of classes.
* Class initializers are only valid for JS classes, and are executed when the class is created.

The encoded name `<clinit>`, which was used for all static initializers before, is now used for class initializers. To preserve backward binary compatibility, a deserialization hack turns class initializers found in Scala classes back into static initializers.

In this commit, the compiler always emits *class initializers*, to emulate the behavior of 1.1.x, and the deserialization is always applied.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/Names.scala
+++ b/ir/src/main/scala/org/scalajs/ir/Names.scala
@@ -21,6 +21,9 @@ object Names {
     UTF8String("<init>")
 
   private final val StaticInitializerSimpleEncodedName: UTF8String =
+    UTF8String("<stinit>")
+
+  private final val ClassInitializerSimpleEncodedName: UTF8String =
     UTF8String("<clinit>")
 
   // scalastyle:off equals.hash.code
@@ -181,7 +184,7 @@ object Names {
    *
    *  Simple names must be non-empty, and can contain any Unicode code point
    *  except `/ . ; [`. In addition, they must not contain the code point `<`
-   *  unless they are the string `<init>` or the string `<clinit>`.
+   *  unless they are one of `<init>`, `<stinit>` or `<clinit>`.
    */
   final class SimpleMethodName private (encoded: UTF8String)
       extends Name(encoded) with Comparable[SimpleMethodName] {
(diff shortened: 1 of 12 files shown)


ir/src/main/scala/org/scalajs/ir/Names.scala
```diff
@@ -351,8 +369,8 @@ object Names {
 
     def apply(simpleName: SimpleMethodName, paramTypeRefs: List[TypeRef],
         resultTypeRef: TypeRef, isReflectiveProxy: Boolean): MethodName = {
-      if ((simpleName.isConstructor || simpleName.isStaticInitializer) &&
-          resultTypeRef != VoidRef) {
+      if ((simpleName.isConstructor || simpleName.isStaticInitializer ||
+          simpleName.isClassInitializer) && resultTypeRef != VoidRef) {
```
gzm0, line 373: Somewhat unrelated: But if we enforce the result type, why not the params?

sjrd: That said, if we change something I would go in the other direction, and *not* enforce the result type in `Names` at all. We can do that in the IR checker instead, which would speed things up a bit since we wouldn't have to check that every time we build a constructor name. There's nothing fundamental in Names that requires that a constructor has a `void` result type; it's more a question of IR validity.

WDYT?

gzm0: Yes, sounds reasonably. In fact, it seems that right now, nothing is checking that :-/

gzm0: Same for the reflexive proxy result type for that matter.

gzm0:
Hmmm... Actually I now realize: `NameGen` is special casing static initializers to not emit their names. Is it intentional that this behavior remains?

sjrd:
> Hmmm... Actually I now realize: `NameGen` is special casing static initializers to not emit their names. Is it intentional that this behavior remains?

I had not thought about that. It's a good question. We need static initializers and class initializers to have different names, now. The current behavior of `NameGen` special-cases static initializers but not class initializers, so they will have different names; it's not problematic. But perhaps it doesn't make *sense* anymore, and we should keep the names both of static initializers and class initialiers, indeed.

sjrd:
Oh actually that's bad, because `<clinit>` needs to be encoded with unicode characters to avoid the `<` and `>`. So I need to change this.

Good catch! 🤯

sjrd:
> Hmmm... Actually I now realize: `NameGen` is special casing static initializers to not emit their names. Is it intentional that this behavior remains?

Fixed now.

PR #4125, opened 2020-07-15
Fix #4117: Allow to export abstract JS classes.

### LINKED ISSUE #4117: Allow to export abstract JS classes
JS classes can be extended in JavaScript. Therefore it would be nice if abstract JS classes could be exported:

```
@JSExportTopLevel("AbstractBase")
abstract class AbstractBase extends js.Object
```

sjrd:
There was no reason to forbid this, and in fact the entire pipeline was already prepared to deal with it.

### DIFF
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
@@ -811,6 +811,51 @@ class ExportsTest {
     assertSame(constr, js.constructorOf[SJSDefinedTopLevelExportedClass])
   }
 
+  @Test def toplevel_exports_for_abstract_JS_classes_issue4117(): Unit = {
+    val constr =
+      if (isNoModule) global.TopLevelExportedAbstractJSClass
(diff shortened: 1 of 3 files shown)


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -811,6 +811,36 @@ class ExportsTest {
...
+  @Test def toplevel_exports_for_abstract_JS_classes_issue4117(): Unit = {
+    val constr =
+      if (isNoModule) global.TopLevelExportedAbstractJSClass
+      else exportsNamespace.TopLevelExportedAbstractJSClass
+
+    assertEquals("function", js.typeOf(constr))
+
+    val subclassFun = new js.Function("constr",
+       """
+       class SubClass extends constr {
```
gzm0, line 823: This needs an `assumeTrue(<es6>)`, no? I'm not sure it's a good idea for the test suite to rely on things we happen to provide.

sjrd: Good point. I have adapted the test so that it uses a function+prototype when we are not assuming ES2015.

PR #4126, opened 2020-07-16
Automate testHtml testing

gzm0:


gzm0:
@sjrd do you think it is worth making this work? If yes, what's the best way of getting the relevant things installed on the jenkins machines?

sjrd:
I don't know. Selenium tests are probably less stable than the code being tested (I've never ran into an issue with `testHtml`, but I regularly see the nightly builds of scalajs-env-selenium failing and then work again the following day), so the cost/benefit ratio will probably not be in favor of adding the test.

Could we implement this test with jsdom instead? (In its "regular" use of having a JS script that calls it to execute tests on an .html page; not as a JS Env to run Scala.js code.)

gzm0:
> Could we implement this test with jsdom instead?

Excellent idea. That seems to work; I'm almost done ;)

gzm0:
Prototype with jsdom. TODO:

- Add `testSuite` to Jenkinsfile
- Add tests with ESModules to Jenkinsfile
- Update `TESTING`

sjrd:
> Add tests with ESModules to Jenkinsfile

Ach, I just remembered that jsdom doesn't support `<script type="module">` yet (https://github.com/jsdom/jsdom/issues/2475) :s From the issue, it's unclear what's blocking them at this point.

gzm0:
> Ach, I just remembered that jsdom doesn't support [...]

Oh well. I think we should still go with this and maybe add a split module test to the manual tests if necessary. WDYT?

sjrd:
Yes, I agree.

project/Build.scala
```diff
@@ -116,6 +118,14 @@ object MyScalaJSPlugin extends AutoPlugin {
           "->https://raw.githubusercontent.com/scala-js/scala-js/v" +
           scalaJSVersion + "/"
         })
+      },
+
+      testHtmlJSDom in Test := {
+        import sys.process._
+        val runner = (testHtml in Test).value.data.getAbsolutePath()
+        val code = Seq("node", "scripts/test-html.js", runner).!
```
sjrd, line 126: I'm always a bit wary of the `scala.sys.process` API. It's an API that no one really maintains, and that I hope will eventually go away. In this case we can just as easily do it with Java's `ProcessBuilder`:
```scala
val code = new ProcessBuilder("node", "scripts/test-html.js", runner)
  .inheritIO()
  .start()
  .waitFor()
```

WDYT?

scripts/test-html.js
```diff
@@ -0,0 +1,77 @@
...
+serveDirectory(dir).then(server => {
+  const url = makeURL(server.address(), filename);
+
+  console.log(`loading ${url}`);
+
+  JSDOM.fromURL(url, {
+    resources: "usable",
+    runScripts: "dangerously",
+  })
+    .then(waitComplete)
```
sjrd, line 20: This formatting is a bit weird, and does not match what you use a few lines below for `.catch`. Consider
```suggestion
  }).then(waitComplete)
```

scripts/test-html.js
```diff
@@ -0,0 +1,77 @@
...
+  var completed = 0
+  for (const box of boxes) {
+    const cls = box.className;
+    if (cls.indexOf("error") !== -1)
+      errors.push(box.innerHTML);
+    else if (cls.indexOf("success") !== -1)
+      completed++;
+  }
+
+  if (errors.length != 0) {
```
sjrd, line 40: For consistency:
```suggestion
  if (errors.length !== 0) {
```

scripts/test-html.js
```diff
@@ -0,0 +1,77 @@
...
+    errors.shift("tests failed:");
+    return Promise.reject(new Error(errors.join("\n")));
+  }
+
+  const total = boxes.length;
+
+  console.log(`${completed}/${total} tests`);
+
+  if (total > 0 && completed == total) {
```
sjrd, line 49: For consistency:
```suggestion
  if (total > 0 && completed === total) {
```

scripts/test-html.js
```diff
@@ -0,0 +1,77 @@
...
+  const boxes = dom.window.document.getElementsByClassName("test-box-header");
+
+  var errors = []
+  var completed = 0
+  for (const box of boxes) {
+    const cls = box.className;
+    if (cls.indexOf("error") !== -1)
+      errors.push(box.innerHTML);
+    else if (cls.indexOf("success") !== -1)
+      completed++;
```
sjrd, line 37: Have you considered testing for other possibilities, and if we find a box that we cannot identify, eagerly throw an error?

scripts/test-html.js
```diff
@@ -0,0 +1,77 @@
...
+}).catch(err => {
+  console.error(err);
+  process.exit(1);
+});
+
+function waitComplete(dom) {
+  const boxes = dom.window.document.getElementsByClassName("test-box-header");
+
+  var errors = []
+  var completed = 0
```
sjrd, line 31: For consistency with the style of everything else (ES 2015 + semicolons):
```suggestion
  const errors = [];
  let completed = 0;
```

gzm0:
Updated

scripts/test-html.js
```diff
@@ -0,0 +1,82 @@
...
+    const cls = box.className;
+    if (cls === "test-box-header error") {
+      errors.push(box.innerHTML);
+    } else if (cls === "test-box-header success") {
+      completed++;
+    } else if (cls === "test-box-header") {
+      // pending
+    } else {
+      return Promise.reject(
+	new Error(`unknown header class ${cls}. element:\n${box.innerHTML}`));
```
sjrd, line 41: There's a tab here that causes a wrong indentation.

PR #4128, opened 2020-07-17
Various Cleanups regarding Properties

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -83,4 +83,16 @@ private[emitter] final class JSGen(val config: Emitter.Config) {
       implicit pos: Position): Function = {
     Function(useArrowFunctions, args, body)
   }
+
+  def genDefineProperty(obj: Tree, prop: Tree, descriptor: Map[String, Tree])(
```
sjrd, line 87: I think we should use a `List[(String, Tree)]` instead of a `Map`, so that the order stays consistent and deterministic. Keys in a JS object are ordered, so an order sequence makes more sense. Besides, this avoids the `.toList` in this method and the `.toMap` in `genJSPropertyES5`.

PR #4133, opened 2020-07-21
Fix #4131: Allow ju.Date to work outside ES time range

### LINKED ISSUE #4131: getTime on java.util.Date returns 0 when initialized with a value outside the ECMAScript Time Range
Taken from #4130 which has exposed this behaviour, but is otherwise orthogonal.

```scala
println(new java.util.Date(8640000000000000L).getTime) // 8640000000000000 both JVM and JS
println(new java.util.Date(8640000000000001L).getTime) // 8640000000000001 on the JVM, 0 on JS
```

`js.Date` objects have a valid time range of ±8.64E15 as defined in [ECMAScript Time Values and Time Range](https://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.1). `js.Date` objects given a value outside of that range are an *invalid date* and calling getTime() on such an instance returns `NaN`.

Since the `java.util.Date` implementation of Scala.js is based on a `js.Date` internally, instantiating `Date` with a value `> 8640000000000000L` (or `< -8640000000000000L` will result in the backing `js.Date` instance being `NaN`. That in turn results in the `java.util.Date` instance being 0 (the Unix Epoch), because `getTime()` converts `NaN` to a `Long` and `Double.NaN.toLong == 0`.

https://github.com/scala-js/scala-js/blob/a4e7c455c29448de48c394865217a22dbd7a1a4a/javalib/src/main/scala/java/util/Date.scala#L85

So we get

`x <= 8640000000000000L // OK`
`x >  8640000000000000L   // getTime() always returns 0`

Analogous for negative values.

Tested on Scala 2.12.11/Scala.js master and 1.1.1

As `java.util.Date` seems to be backed by a `Long` on the JVM there's obviously no range check required when calling `new Date(date: Long)`.
Because the semantics is different on JS though, I'd argue that it could make sense to throw an `IllegalArgumentException` on the `Date` constructor if the timestamp is out of range.

gzm0:


javalib/src/main/scala/java/util/Date.scala
```diff
@@ -36,82 +38,91 @@ class Date private (private val date: js.Date) extends Object
...
+    case d: Date => d.millis == millis
     case _       => false
   }
 
-  override def hashCode(): Int = date.getTime().hashCode()
+  override def hashCode(): Int = millis.hashCode()
 
   @Deprecated
-  def getDate(): Int = date.getDate().toInt
+  private def asDate(): js.Date = {
```
sjrd, line 61: `@Deprecated` doesn't seem to make sense on this function, nor on `mutDate`.

javalib/src/main/scala/java/util/Date.scala
```diff
@@ -36,82 +38,91 @@ class Date private (private val date: js.Date) extends Object
...
+    if (!isSafeJSDate()) {
+      throw new IllegalArgumentException(
+          f"cannot convert this java.util.Date ($millis millis) to a js.Date")
+    }
+    new js.Date(millis.toDouble)
```
sjrd, line 66: Have you considered caching this instance of `js.Date`? Set it to `null` initially, reset it to `null` in `setTime()`, and cache it here.
(`mutDate` would have to make a copy in that case, in case `safeGetTime` throws an exception, otherwise we would have mutated the `js.Date` but not the `millis`, which would therefore not be in sync anymore)

javalib/src/main/scala/java/util/Date.scala
```diff
@@ -36,82 +38,91 @@ class Date private (private val date: js.Date) extends Object
...
 
-  override def hashCode(): Int = date.getTime().hashCode()
+  override def hashCode(): Int = millis.hashCode()
 
   @Deprecated
-  def getDate(): Int = date.getDate().toInt
+  private def asDate(): js.Date = {
+    if (!isSafeJSDate()) {
+      throw new IllegalArgumentException(
+          f"cannot convert this java.util.Date ($millis millis) to a js.Date")
```
sjrd, line 64:
```suggestion
          s"cannot convert this java.util.Date ($millis millis) to a js.Date")
```
There's no need to depend on `java.util.Formatter` for this.

javalib/src/main/scala/java/util/Date.scala
```diff
@@ -120,24 +131,36 @@ class Date private (private val date: js.Date) extends Object
...
+      val offset = -date.getTimezoneOffset().toInt
+      val sign = if (offset < 0) "-" else "+"
+      val hours = pad0(Math.abs(offset) / 60)
+      val mins = pad0(Math.abs(offset) % 60)
+      Days(date.getDay().toInt) + " "+ Months(date.getMonth().toInt) + " " +
+        pad0(date.getDate().toInt) + " " + pad0(date.getHours().toInt) + ":" +
+        pad0(date.getMinutes().toInt) + ":" + pad0(date.getSeconds().toInt) +
+        " GMT" + " " + date.getFullYear().toInt
+    } else {
+      f"java.util.Date($millis)"
```
sjrd, line 152:
```suggestion
      s"java.util.Date($millis)"
```

javalib/src/main/scala/java/util/Date.scala
```diff
@@ -120,24 +131,36 @@ class Date private (private val date: js.Date) extends Object
...
+    }
   }
+
+  @inline
+  private def isSafeJSDate(): Boolean =
+    -MaxMillis <= millis && millis <= MaxMillis
 }
 
 object Date {
+  private final val MaxMillis = 8640000000000000L
```
sjrd, line 162: This constant deserves a comment ;) Perhaps a link to [`TimeClip()`](https://tc39.es/ecma262/#sec-timeclip) in the ES spec?

javalib/src/main/scala/java/util/Date.scala
gzm0: I did this in the hope that the Scala compiler would complain, if it is used from a non-deprecated API. Seems it doesn't work that way (`toString` is the counter example). I removed them.

javalib/src/main/scala/java/util/Date.scala
gzm0: Yes. I deemed it too complicated for unclear benefits.

sjrd: Fair enough.

PR #4136, opened 2020-07-27
Add ModuleKind as an attribute to linked Scala.js files

gzm0:
Thanks to this, we do not need to "peek-back" as far when assembling the
jsEnvInput. That in turn allows us to properly deal with different
module kinds in different configurations.

The re-use of the ModuleKind type is somewhat a shortcut, but I think
that is acceptable for the sake of consistency / brievity.

### DIFF
--- a/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
+++ b/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
@@ -176,11 +176,7 @@ private[sbtplugin] object ScalaJSPluginInternal {
          */
         val moduleKind = (scalaJSLinkerConfig in key).value.moduleKind
 
-        Def.task {
-          val log = s.log
-          val realFiles = irInfo.get(scalaJSSourceFiles).get
-          val ir = irInfo.data
-
+        val configChanged = {
           def moduleInitializersChanged = (scalaJSModuleInitializersFingerprints in key)
             .previous
             .exists(_ != (scalaJSModuleInitializersFingerprints in key).value)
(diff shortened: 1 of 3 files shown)


sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -231,7 +218,18 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
 
-          Attributed.blank(output).put(scalaJSSourceMap, sourceMapFile)
+          /* This is is somewhat not nice: We must make assumptions about the
+           * specification of the specific linker in use. Otherwise, we do not
+           * know how to interpret / load the resulting file.
+           *
+           * See jsEnvInput (via scalaJSLinkedFile) to see how this is used.
+           */
+          val moduleKind =
+            (scalaJSLinkerConfig in key).value.moduleKind
```
sjrd, line 228: This should be moved above the `Def.task`. See the big comment that starts with "It is very important that" :p

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -231,7 +218,18 @@ private[sbtplugin] object ScalaJSPluginInternal {
             Set(output, sourceMapFile)
           } (realFiles.toSet)
 
-          Attributed.blank(output).put(scalaJSSourceMap, sourceMapFile)
+          /* This is is somewhat not nice: We must make assumptions about the
```
sjrd, line 221:
```suggestion
          /* This is somewhat not nice: We must make assumptions about the
```

gzm0:
Updated. I've also added a cleanup commit.

PR #4137, opened 2020-07-27
Implement String.repeat(Int) for JDK 11+

exoego:
This PR implements [`java.lang.String.repeat(Int)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#repeat(int)), added in JDK 11, for Scala.js 1.x

Algorithm is based on [the polyfill from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat), which is public domain.

### DIFF
--- a/javalanglib/src/main/scala/java/lang/_String.scala
+++ b/javalanglib/src/main/scala/java/lang/_String.scala
@@ -308,6 +308,26 @@ final class _String private () // scalastyle:ignore
     regionMatches(false, toffset, other, ooffset, len)
   }
 
+  def repeat(count: Int): String = {
+    if (count < 0) {
+      throw new IllegalArgumentException
+    } else if (thisString == "" || count == 0) {
+      ""
+    } else if (thisString.length > (Int.MaxValue / count)) {
+      throw new OutOfMemoryError
+    } else {
+      var str = thisString
+      val resultLength = thisString.length * count
+      var remainingIters = 31 - Integer.numberOfLeadingZeros(count)
+      while (remainingIters > 0) {
+        str += str
+        remainingIters -= 1
+      }
+      str += str.jsSubstring(0, resultLength - str.length)
+      str
+    }
+  }
+
   @inline
   def replace(oldChar: Char, newChar: Char): String =
     replace(oldChar.toString, newChar.toString)
--- a/test-suite/shared/src/test/require-jdk11/org/scalajs/testsuite/javalib/lang/StringTestOnJDK11.scala
+++ b/test-suite/shared/src/test/require-jdk11/org/scalajs/testsuite/javalib/lang/StringTestOnJDK11.scala
@@ -1,0 +1,41 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.javalib.lang
+
+import java.nio.charset.Charset
+
+import scala.language.implicitConversions
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.testsuite.utils.AssertThrows._
+import org.scalajs.testsuite.utils.Platform._
+
+class StringTestOnJDK11 {
+  @Test def repeat(): Unit = {
+    assertThrows(classOf[IllegalArgumentException], "".repeat(-1))
+    assertTrue("".repeat(0) == "")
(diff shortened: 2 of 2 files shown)


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/StringTest.scala
```diff
@@ -441,4 +441,23 @@ class StringTest {
   @inline private def erased(array: Array[String]): Array[AnyRef] = {
     array.asInstanceOf[Array[AnyRef]]
   }
+
+  @Test def repeat(): Unit = {
+    assumeFalse("String#repeat requires JDK 11+",
+      executingInJVMOnJDK8OrLower)
```
sjrd, line 447: This will not be a sufficient guard, because the code doesn't compile on JDK 8 in the first place.

We have no precedent for implementing JDK stuff from JDK > 8. However, we had precedent in the past for JDK 7 and JDK 8, when we still supported building on JDK 6:
https://github.com/scala-js/scala-js/blob/617c4170fe9844eea0115b6ed59b2816201a70e7/project/Build.scala#L1369-L1382

exoego: Attempted to configure build for JDK11 
https://github.com/scala-js/scala-js/pull/4137/commits/8bb6ff3f36b2c3f1ccebfc4a42a0465c10d6cc46

Jenkinsfile
```diff
@@ -379,7 +379,7 @@ def Tasks = [
 ]
 
 def mainJavaVersion = "1.8"
-def otherJavaVersions = []
+def otherJavaVersions = ["11"]
```
exoego, line 382: I am afraid that adding JDK11 doubles the build matrix 🤔

sjrd:
Hey @exoego. We finally managed to set up a JDK 11 CI #4219. If you're still interested in pursuing this PR, would you mind rebasing it on top of the latest master? Thanks.

exoego:
I will continue within a few days.

javalanglib/src/main/scala/java/lang/_String.scala
```diff
@@ -308,6 +308,27 @@ final class _String private () // scalastyle:ignore
     regionMatches(false, toffset, other, ooffset, len)
   }
 
+  @inline
```
sjrd, line 311: This `@inline` doesn't seem appropriate, given the length of that method.

test-suite/shared/src/test/require-jdk11/testsuite/javalib/lang/StringTestJDK11.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import scala.language.implicitConversions
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.testsuite.utils.AssertThrows._
+import org.scalajs.testsuite.utils.Platform._
+
+class StringTestJDK11 {
```
sjrd, line 25: To follow the (dominating) convention, this should be called `StringTestOnJDK11`:
```suggestion
class StringTestOnJDK11 {
```
(and the file name should be adapted)

sjrd (changes requested):
Thanks @exoego. I have two small comments. Otherwise, could you squash/rebase everything into one commit with a clean commit message, for example

    Implement String.repeat(Int) for JDK11+

?

test-suite/shared/src/test/require-jdk11/testsuite/javalib/lang/StringTestOnJDK11.scala
```diff
@@ -0,0 +1,41 @@
+/*
```
sjrd, line 1: The file path is missing `org/scalajs/`:
```diff
-test-suite/shared/src/test/require-jdk11/testsuite/javalib/lang/StringTestOnJDK11.scala
+test-suite/shared/src/test/require-jdk11/org/scalajs/testsuite/javalib/lang/StringTestOnJDK11.scala
```
On master there is already `CharacterTestOnJDK11.scala` in the `test-suite/shared/src/test/require-jdk11/org/scalajs/testsuite/javalib/lang/` directory, so `StringTestOnJDK11.scala` should be next to it.

sjrd (changes requested):
Oops, sorry, I apparently missed the notification that this PR had been updated.

There's just one file path issue. Otherwise LGTM.

test-suite/shared/src/test/require-jdk11/testsuite/javalib/lang/StringTestOnJDK11.scala
exoego: Fixed and squashed commits.

sjrd (approved):
Thank you @exoego !

PR #4139, opened 2020-07-28
Ref Tracking Fixes

gzm0:


gzm0:
Mostly wondering why this wont work :P

gzm0:
I realize that the global ref tracking adds quite a bit of clutter to the code. An alternative could be that we "officially" declare a set of always reached global refs and always return these. We might still want to keep `globalRef` in `JSGen` so we can assert it being in the set and mark the locations in the code.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
```diff
@@ -89,10 +89,12 @@ private[emitter] final class JSGen(val config: Emitter.Config) {
     val descriptorTree =
         ObjectConstr(descriptor.map(x => StringLiteral(x._1) -> x._2))
 
-    val tree = Apply(
-        genIdentBracketSelect(VarRef(Ident("Object")), "defineProperty"),
-        List(obj, prop, descriptorTree))
-
-    WithGlobals(tree, Set("Object"))
+    globalRef("Object").map { objRef =>
```
sjrd, line 92: This is in the commit for "exports" but it's about "Object".

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -666,7 +666,7 @@ private[emitter] object CoreJSLib {
           if (implementedInObject) {
             val staticObjectCall: Tree = {
               val fun = classVar("c", ObjectClass).prototype DOT genName(methodName)
-              Return(Apply(fun DOT "call", instance :: args))
+              Return(Apply(genIdentBracketSelect(fun, "call"), instance :: args))
```
sjrd, line 669: That I wouldn't do. I'm worried that it could prevent GCC from recognizing that as a super call, hence preventing some optimizations. I think we should always preserve `.call` as true dot-selects.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2085,7 +2085,7 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
           } else {
             val fun =
               classVar("c", className).prototype DOT transformMethodIdent(method)
-            js.Apply(fun DOT "call", transformedArgs)
+            js.Apply(genIdentBracketSelect(fun, "call"), transformedArgs)
```
sjrd, line 2088: Same comment about `call`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/JSGen.scala
gzm0: It just uses the newly introduced helper as well. Do you want me to put it in a separate commit?

sjrd: Ah right, it was already tracked, but not using the helper. OK that's good as is, then.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
gzm0: I'll try this out :)

gzm0: Seems indeed to be the case. I have also reverted the commit for `Array.length`. We can expect similar behavior (e.g. `[1,2,3].length` v.s. `[1,2,3]["length"]`)

PR #4141, opened 2020-07-29
Various improvements to instance test emission

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -790,7 +773,15 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
 
           case _ =>
             var test = if (tree.kind.isClass) {
-              obj instanceof classVar("c", className)
+              if (!tree.hasInstances) {
```
sjrd, line 779: I don't think this test is protected by the proper invalidation logic, hence there could be an issue with incremental runs, if between runs a classes gets instances or loses instances.

PR #4144, opened 2020-07-30
A few bugfixes near BigInteger's hashcode

catap:
This pull request contains a two bugs that I found inside hashCode at BigInteger.

Both of them extremely rare but very surprising.

Let make review of this changes here, and after that I will backport it to scala-native

### DIFF
--- a/javalib/src/main/scala/java/math/BigDecimal.scala
+++ b/javalib/src/main/scala/java/math/BigDecimal.scala
@@ -1282,9 +1282,9 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
     if (_hashCode != 0) {
       _hashCode
     } else if (_bitLength < 64) {
-      _hashCode = _smallValue.toInt
-      _hashCode = 33 * _hashCode + (_smallValue >> 32).toInt
-      _hashCode = 17 * _hashCode + _scale
+      var hashCode  = _smallValue.toInt
+      hashCode = 33 * hashCode + (_smallValue >> 32).toInt
+      _hashCode = 17 * hashCode + _scale
       _hashCode
     } else {
       _hashCode = 17 * _intVal.hashCode + _scale
--- a/javalib/src/main/scala/java/math/BigInteger.scala
+++ b/javalib/src/main/scala/java/math/BigInteger.scala
@@ -486,14 +486,16 @@ class BigInteger extends Number with Comparable[BigInteger] {
     if (_hashCode != 0) {
       _hashCode
     } else {
-      for (i <- 0 until numberLength) {
-        _hashCode = _hashCode * 33 + digits(i)
+      var hashCode = 0
+      var i = 0
+      while (i < numberLength) {
+        hashCode = hashCode * 33 + digits(i)
+        i += 1
       }
-      _hashCode = _hashCode * sign
+      _hashCode = hashCode * sign
       _hashCode
     }
   }
-
   override def intValue(): Int = sign * digits(0)
 
   def isProbablePrime(certainty: Int): Boolean =
--- a/javalib/src/main/scala/java/math/BigInteger.scala
+++ b/javalib/src/main/scala/java/math/BigInteger.scala
@@ -817,7 +817,10 @@ class BigInteger extends Number with Comparable[BigInteger] {
     else BitLevel.shiftLeftOneBit(this)
   }
 
-  private[math] def unCache(): Unit = firstNonzeroDigit = firstNonzeroDigitNotSet
+  private[math] def unCache(): Unit = {
+    firstNonzeroDigit = firstNonzeroDigitNotSet
+    _hashCode = 0
+  }
 
   /** Puts a big-endian byte array into a little-endian applying two complement. */
   private def putBytesNegativeToIntegers(byteValues: Array[Byte]): Unit = {


gzm0:
Thanks for the PR!

I'm unsure what to do about the first commit: JavaScript is a non-concurrent environment, so this problem will never occur. I'm unclear about the amount of code-sharing Scala.js and Scala native have at this point. Do you know what the approach has been so far? (@sjrd is on vacation, but we can also wait for him).

The second commit looks definitely like a bug we should fix.

I suggest the following:
- Create a test case from the example in the second commit message.
- Make sure it fails on the current codebase.
- File a bug for defect tracking.
- Split off the second commit with the test case and referencing the bug in its own PR.
- Discuss the first commit separately.

catap:
@gzm0 let me be blank:

1. Scala-native reuse the same code base, it was a copy and paste and they backported all fixes to it.
2. `Elementary` is private object and nobody has access to it outside of `BigInteger` and `BigDecimal` implementation :)
3. The code is failing and at my fork unit tests are in the same packet ;) and example from the second commit is my unit test.
5. Scala-native is also single thread now. But they working to make things different, and for me better to fix the bug special when fix is trivial like this because I'm using this code also inside JVM in highly concurent environment and was very surprised then `.distinct` keeps duplicates sometime.

So, it aren't production issues. Honestly I not sure that it is possible to get it inside Scala-js, and at Scala-native for now.

But on another hand, we can ignore it and forgot about it and someone will spent hours to trying to understand that's going on. For me it was fast because I *know* how this `BigInteger` works, for someone who didn't... well.. it will be very fascinatingly. Extreamly fascinating.

gzm0:
> Scala-native reuse the same code base, it was a copy and paste and they backported all fixes to it.

That was my understanding. What I do not know is how Scala-native / Scala.js intends to maintain this going forward. If Scala.js puts Scala-native specific fixes in its codebase, we cannot test it :S OTOH, having two almost identical codebases is also not really a good option.

> Elementary is private object and nobody has access to it outside of BigInteger and BigDecimal implementation :)

I do not understand how this relates to this PR. It doesn't touch `Elementary`, does it?

> The code is failing and at my fork unit tests are in the same packet ;) and example from the second commit is my unit test.

Oh, good. (I assume this is for the second commit). So that is definitely something we should fix then. Can you include it as a unit test in Scala.js as well, please?

> Scala-native is also single thread now. But they working to make things different, and for me better to fix the bug special when fix is trivial like this because I'm using this code also inside JVM in highly concurent environment and was very surprised then .distinct keeps duplicates sometime.

The Scala.js implementation of the JDK is not intended for use on the JVM. Any reason you are not just using your JDK's `Big Integer`?

> Honestly I not sure that it is possible to get it inside Scala-js, and at Scala-native for now.

We definitely should merge the second commit (just with the unit test included). The first one, I'm not so sure.

gzm0:
Oh, now I understand: `Elementary` is required to make anything fail. `BigInteger`s have an immutable API. I think in this case, that's not a bug.

catap:
@gzm0 

> The Scala.js implementation of the JDK is not intended for use on the JVM. Any reason you are not just using your JDK's Big Integer?

I spent last couple of months to make `modPow` faster and cheaper. It is very important to me ;) And because I have a goal to use a libarary at scala-js and scala-native and where `modPow` is extreamly slower (than JVM version), I've forked it and started to work.

Right now my implementation of `modPow` faster than JVM's ;) and to make code easy I'm using it everywhere. At some point (as soon as I stop to edit it) I will make repository public and created a PR to synchronize it with SJS/SN.

> If Scala.js puts Scala-native specific fixes in its codebase, we cannot test it :S OTOH, having two almost identical codebases is also not really a good option.

I don't belive that someone fixed something related to `BigInteger` inside SN, or at least I haven't noticied it. So, right now it is unmantained and as I understand the best things to move forward is extract javalib to separated repository/pacakge/etc.

But it is a huge task, and defently not for now.

>  I think in this case, that's not a bug.

Yes, I have no idea how to get any of these bugs inside SJS. "Tomorrow" the first one can be re-discovered at SN.

But on another hands it is trivial changes. Why not fix it?

gzm0:
I see. I think this is a larger discussion and we should wait until @sjrd comes back. An option I see for this is that we move the code out of Scala.js and have Scala.js (and other projects) depend on the source. However, for this I'd like @sjrd's perspective; he has a better overview over the ecosystem.

catap:
@gzm0 I agree with you, but honestly I don't think that too many projects will use this code and copy-and-paste looks as cheap solution here.

Also, if I remember right, scala-js can be compiled to node and I guess node's worker thread can be used as dedicated threads as a way to get the bug from the first commit.

sjrd:
So, if I understand correctly:

The first commit fixes a hypothetical concurrency bug, but it cannot be observed neither in Scala.js nor in Scala Native because neither is multi-threaded. Scala.js will stay single-threaded forever, SN will probably eventually become multi-threaded, but it's unclear when and how. In any case, when it becomes multi-threaded, it is likely that there will be zillions of things to adjust, and depending on the exact memory model, the fix here might not be enough.

The second commit patches `BigInteger.unCache()`, which was previously documented to alter `firstNonzeroDigit` (albeit indirectly, in the definition of `firstNonzeroDigitNotSet`), so that now it also alters `_hashCode`. `unCache()` is called by internal methods that mutate BigIntegers in-place. They are only called for BigIntegers that have not escaped the implementation of `BigInteger` yet, and so they are not visible. So, unless the implementation calls `hashCode()` on not-yet-fully-computed BigIntegers, this is a non-existent bug. My understanding here is that calling `hashCode()` on such transient `BigInteger`s is never done, and not part of the valid "protocol" of computation-in-progress. In addition, changing `unCache()` is probably not enough to "fix" all unintended uses, since other methods mutate existing BigIntegers without calling `unCache()`, because they don't need `firstNonzeroDigit` to be recomputed, but they might need `_hashCode` to be recomputed; so the fix is incomplete there as well.

IMO the second commit should clearly not be included, neither in Scala.js nor Scala Native. If a general reimplementation changes the protocol of in-progress computations, then that reimplementation may have to change what `unCache()` means. But in isolation, this commit brings nothing.

Whether the first commit should be merged depends on how we view the ecosystem and code sharing between Scala.js and SN. Currently, the javalibs of Scala.js and SN are both written for single-threaded environments. As long as SN doesn't support multi-threading, I see no reason to change that. When we get there, there will be a question of whether to include the necessary changes in Scala.js as well for the sake of maintaining the same code. This only makes sense if the code is 100% the same. At the moment this is *not* the case for most of the javalib, as Scala.js and Scala Native have different performance profiles in their underlying platforms. `java.math` is a notable exception here, not the rule. Its implementation is so massive that the possible platform-specific performance improvements are not worth the increase in maintenance cost.

Should we put `java.math.BigInteger` and `BigDecimal` in a separate repo? Perhaps, but only if we can still test that repo in isolation (which means it must depend on Scala.js and SN, not the other way around) and if we can still test Scala.js in isolation. The latter condition is problematic, because some of the Scala stdlib indirectly relies on big numbers.

It's also worth taking into account the fact that, eventually, we'll want to specialize Scala.js' implementation of `BigInteger` to use native JavaScript `bigint`s (probably with a fallback). At that point, the codebases will have to diverge, at least in the superficial API.

When I put all these pieces together, today my conclusion is that we should not merge the first commit either.

sjrd:
Closing until new arguments are brought forward.

PR #4145, opened 2020-07-30
#4143: Pseudo-union type inference with type aliases and js.undefined

### LINKED ISSUE #4143: Assigning js.undefined to a type alias to js.UndefOr does not compile
```scala
import scala.scalajs.js.|
import scala.scalajs.js
type Foo = Int | String
type FooOpt = js.UndefOr[Foo]
val foo: js.UndefOr[Foo] = js.undefined // Works
val foo2: FooOpt = js.undefined // Doesn't work: "Expected `Nothing | Unit`,  got `Int | String | Unit`"
```

Unclear if we can do anything about this, but opening this so we have a centralized discussion point.

frobinet:
This PR fixes #4143, an issue with type inference for pseudo-union types `scala.scalajs.js.|` when `Nothing` is the first type parameter. This is especially useful when infering for `js.undefined` (see #4143).

Since Scala 3 will include proper union types, I've limited the changes to the strict minimum to allow `js.undefined` to work with type aliases like it did in Scala.js 0.6.x. 

This is my first contribution to Scala.js, so there are probably issues with this PR. Feel free to point them out!

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -17,8 +17,8 @@ import java.util.concurrent.ConcurrentHashMap
 import scala.util.matching.Regex
 
 object ScalaJSVersions extends VersionChecks(
-    current = "1.2.0-SNAPSHOT",
-    binaryEmitted = "1.2-SNAPSHOT"
+    current = "1.1.2",
+    binaryEmitted = "1.1"
 )
 
 /** Helper class to allow for testing of logic. */
--- a/library/src/main/scala/scala/scalajs/js/Union.scala
+++ b/library/src/main/scala/scala/scalajs/js/Union.scala
@@ -63,6 +63,13 @@ object | { // scalastyle:ignore
     /** If `A <: B1`, then `A <: B1 | B2`. */
     implicit def left[A, B1, B2](implicit ev: Evidence[A, B1]): Evidence[A, B1 | B2] =
       ReusableEvidence.asInstanceOf[Evidence[A, B1 | B2]]
+
+    /** Explicit rules for `A <: A | Unit`.
+     *
+     * This helps type inference in with `js.UndefOr[A]` which is aliased to `A | Unit`.
+     */
+    implicit def addUnit[A]: Evidence[A, A | Unit] =
+      ReusableEvidence.asInstanceOf[Evidence[A,  A | Unit]]
   }
 
   object Evidence extends EvidenceLowPrioImplicits {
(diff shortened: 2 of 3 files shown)


frobinet:
This passes my test but causes issues with diverging implicit expansion in some other projects where I've tried it. I'll take a closer look this week-end.

frobinet:
I added a second implicit conversion and more some other tests to cover more complex type aliases.

sjrd:
I really don't like the left-right asymmetry of the new rules. Previously, all rules where symmetric in the sense that if there was a rule for `X | Y`, there was the same rule for `Y | X`. Now this is not the case anymore: there is a rule for `Nothing | Y` for not for `Y | Nothing`, and there is a rule for `X | Unit` but none for `Unit | X`.

I am also wary about the fact that you had to add `addUnit` to compensate for bad effects introduced by `fromNothing`. This is very worrying, because it means that `fromNothing` can have other unanticipated effects.

For those two reasons, I don't think this is the best way to go to mitigate the type inference issue.

frobinet:
`addUnit` was not needed because of `fromNothing`, but more in addition to it to make things work in some even harder test cases. 

I agree that not having the symmetry is not elegant, I used this more as a direct solution to the specific problem that occured with type aliases and `js.undefined`, so I tried to restrict the addition of rules to the bare minimum. 

I understand your concerns about these additions breaking some other things downstream. Do you have some kind of community build for Scala.js that might help us assess this? Otherwise, it is probably wiser to close this and try to fix the inference issue in the Scala compiler itself. That will probably be above my pay-grade though! 😄

sjrd:
> Do you have some kind of community build for Scala.js that might help us assess this?

No, we don't. We rely on an extensive test suite instead, which is usually good enough. But for cases of tricky type inference like this, a test suite can never be enough because of many unknown unknowns about interactions with surrounding code. That's why we have to be extra careful.

I'm also particularly wary here, compared to other changes we've done in the past to `Union.scala`, because the new rules are, in theory, completely redundant:

* `addUnit[A]` is the same as `left[A, A, Unit](base[A])`
* `fromNothing[A2, B1, B2](a)(ev_A2_B2)` is the same as

      from[Nothing | A2, B1, B2](a)(
        allSubtypes[Nothing, A2, B1 | B2](
          base[Nothing], // an Evidence[Nothing, Nothing] which, by covariance, is also an Evidence[Nothing, B1 | B2]
          right[A2, B1, B2](ev_A2_B2) // an Evidence[A2, B1 | B2]
        )
      )

and redundant implicit definitions can easily cause divergences or ambiguities.



> Otherwise, it is probably wiser to close this and try to fix the inference issue in the Scala compiler itself. That will probably be above my pay-grade though!

Maybe *fixing* the type inference is above you pay-grade (I don't know your background nor your time availabilities), but if you got this far, I'm sure you could create a reproduction to post as an issue on scala/bug 😄

gzm0:
Closing this, as discussed in #4143.

PR #4146, opened 2020-08-12
RFC: Introduce Foreign Static Initializers

gzm0:


gzm0:
Questions:

- Is this even a good idea?
- Should we hash `ForeignStaticInitializer` (probably)?
- Should we make `ForeignStaticInitializer` more like a `MethodDef`?

gzm0:
It seems we cannot use `Hashers` in `linker` because it needs `MessageDigest` which is only available in Java. I'm thinking of an alternative that generates `Versioned[MethodDef]`s directly from the `MethodSynthesizer`.

gzm0:
Closing based on the discussion in https://github.com/scala-js/scala-js/pull/4111#issuecomment-673590827.

PR #4149, opened 2020-08-16
Fix #4148: Do not erase a regular class with a static forwarder class.

### LINKED ISSUE #4148: Unable to find extant inner trait IR during fastopt
Please refer to this tag: https://github.com/djspiewak/cats-effect/tree/bug/scalajs-concurrentops

You can reproduce the issue with `sbt coreJS/test`. `lawsJS/test` also reproduces it. ScalaJS 1.1.1, Scala 2.13.3/2.12.11 (reproduces on both). Compilation finishes without a problem, but fastOpt does not:

```
[error] Referring to non-existent class cats.effect.Concurrent$Ops
[error]   called from private cats.effect.Concurrent$.$anonfun$continual$1(java.lang.Object,cats.effect.Concurrent,scala.Function1,cats.effect.concurrent.Deferred)java.lang.Object
[error]   called from cats.effect.Concurrent$.continual(java.lang.Object,scala.Function1,cats.effect.Concurrent)java.lang.Object
[error]   called from cats.effect.Concurrent.continual(java.lang.Object,scala.Function1)java.lang.Object
[error]   called from private cats.effect.internals.MVarConcurrent.$anonfun$modify$3(scala.Function1,cats.effect.concurrent.Ref,java.lang.Object)java.lang.Object
[error]   called from private cats.effect.internals.MVarConcurrent.$anonfun$modify$1(scala.Function1,cats.effect.concurrent.Ref)java.lang.Object
[error]   called from cats.effect.internals.MVarConcurrent.modify(scala.Function1)java.lang.Object
[error]   called from private cats.effect.concurrent.BaseMVarTests.$anonfun$new$212(cats.effect.concurrent.MVar2,int)cats.effect.IO
[error]   called from private cats.effect.concurrent.BaseMVarTests.$anonfun$new$211(cats.effect.concurrent.MVar2,java.lang.Void)cats.effect.IO
[error]   called from private cats.effect.concurrent.BaseMVarTests.$anonfun$new$210(cats.effect.concurrent.MVar2)cats.effect.IO
[error]   called from private cats.effect.concurrent.BaseMVarTests.$anonfun$new$209()scala.concurrent.Future
[error]   called from constructor cats.effect.concurrent.BaseMVarTests.<init>()void
[error]   called from constructor cats.effect.concurrent.MVarAsyncTests.<init>()void
[error]   called from static constructor cats.effect.concurrent.MVarAsyncTests.<clinit>()void
[error]   cats.effect.Concurrent$
[error]   cats.effect.internals.MVarConcurrent
```

There's a bunch of that, but this is the topmost one. The perplexing thing is that the `Concurrent$Ops.sjsir` file *exists*, and `coreJVM/test` works just fine. The error survives through `clean` and even `rm -rf *; git reset --hard HEAD`.

Note that this source code used to use Simulacrum, but I removed that on this tag. You can see the pre-removal state in `HEAD^` on the tag. The source code *itself* is identical to the **series/2.x** branch on the same repository, which builds and tests just fine with ScalaJS 1.1.1.

I'm going to keep poking around with the plugins to see if I can figure out where the error was introduced, but this *seems* like an SJS bug.

sjrd:
This is only relevant for case insensitive file systems. On such file systems, a static nested object could generate a static forwarder class whose name differed from a regular class' name only in case, therefore erasing the latter on case insensitive file systems.

We now avoid emitting static forwarder classes for objects if they would clash with a regular class on a case insensitive file system.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -217,7 +217,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
 
     private val lazilyGeneratedAnonClasses = mutable.Map.empty[Symbol, ClassDef]
     private val generatedClasses = ListBuffer.empty[js.ClassDef]
-    private val generatedStaticForwarderClasses = ListBuffer.empty[js.ClassDef]
+    private val generatedStaticForwarderClasses = ListBuffer.empty[(Symbol, js.ClassDef)]
 
     private def consumeLazilyGeneratedAnonClass(sym: Symbol): ClassDef = {
       /* If we are trying to generate an method as JSFunction, we cannot
@@ -393,9 +393,22 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
 
           val generatedCaseInsensitiveNames =
             regularClasses.map(caseInsensitiveNameOf).toSet
-          val staticForwarderClasses = generatedStaticForwarderClasses.toList.filter { classDef =>
-            !generatedCaseInsensitiveNames.contains(caseInsensitiveNameOf(classDef))
-          }
+          val staticForwarderClasses = generatedStaticForwarderClasses.toList
+            .withFilter { case (site, classDef) =>
+              if (!generatedCaseInsensitiveNames.contains(caseInsensitiveNameOf(classDef))) {
+                true
+              } else {
+                global.runReporting.warning(
+                    site.pos,
+                    s"Not generating the static forwarders of ${classDef.name.name.nameString} " +
+                    "because its name differs only in case from the name of another class or " +
+                    "trait in this compilation unit.",
+                    WarningCategory.Other,
+                    site)
+                false
+              }
+            }
+            .map(_._2)
 
           regularClasses ::: staticForwarderClasses
         }
@@ -559,7 +572,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
                   forwarders,
                   Nil
               )(js.OptimizerHints.empty)
-              generatedStaticForwarderClasses += forwardersClassDef
+              generatedStaticForwarderClasses += sym -> forwardersClassDef
             }
           }
           allMemberDefsExceptStaticForwarders
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -217,6 +217,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
 
     private val lazilyGeneratedAnonClasses = mutable.Map.empty[Symbol, ClassDef]
     private val generatedClasses = ListBuffer.empty[js.ClassDef]
+    private val generatedStaticForwarderClasses = ListBuffer.empty[js.ClassDef]
 
     private def consumeLazilyGeneratedAnonClass(sym: Symbol): ClassDef = {
       /* If we are trying to generate an method as JSFunction, we cannot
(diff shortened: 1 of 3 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -364,14 +365,46 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+          generatedClasses.toList
+        } else {
+          val regularClasses = generatedClasses.toList
+
+          /* #4148 Add generated static forwarder classes, except those that
+           * would collide with regular classes on case insensitive file
+           * systems.
+           */
+
+          /* I could not find any reference anywhere about what locale is used
```
gzm0, line 381: What does scalac do to emit the warnings?

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -906,4 +914,24 @@ object RegressionTest {
   object `class` { // scalastyle:ignore
     def foo(x: Int): Int = x + 1
   }
+
+  /* #4148: The objects and classes here intentionally have names that differ
+   * only in case, and are intentially defined in a specific order.
```
gzm0, line 919: intantionally

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -364,14 +365,46 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+           * as opposed to using the OS' default locale.
+           */
+          def caseInsensitiveNameOf(classDef: js.ClassDef): String =
+            classDef.name.name.nameString.toLowerCase(java.util.Locale.ENGLISH)
+
+          val generatedCaseInsensitiveNames =
+            regularClasses.map(caseInsensitiveNameOf).toSet
+          val staticForwarderClasses = generatedStaticForwarderClasses.toList.filter { classDef =>
+            !generatedCaseInsensitiveNames.contains(caseInsensitiveNameOf(classDef))
+          }
```
gzm0, line 395: Can we at least warn (just like scalac) when we are dropping static forwarders?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: It uses `.toLowerCase()` without argument, but I think that's very bad because different users could have different warnings.

gzm0: Ah, yes. That is definitely worse than fixing the english locale. Maybe add that in the comment?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: We can, but then we'd have to propose a flag to disable that warning, or we're going to make some library maintainers angry at us for breaking -Xfatal-warnings in a way that is not recoverable without breaking their backward binary compat.

Note that scalac never drops static forwarders for the reason that we do here. Instead if *will* erase things, and will emit a warning that the result will be broken on case insensitive file systems. However it only emits static forwarders for top-level objects, not for static objects inside objects. Here, we are instead fixing the potential clash. The result is more correct than what scalac lets through for top-level objects, so not necessarily as deserving of a warning as the case (mis-)handled by scalac.

gzm0: My main concern is that reasoning about the existence of static forwarders (from a users POV) becomes non-local and difficult.

In general, I find non-opt-in do-it-if-you-can behavior has a high potential to cause confusing behavior. I am not set on a warning. Maybe it is better if we protect exporting nested static forwarders with a flag (for compilation of the JDK) but hard fail if there is a conflict. IIUC, this would address all use cases we have and avoid potential confusing behavior. (IIUC, nobody needs the avoidance, it's just that static forwarders are always emitted).

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: I've added that in the comment.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: I can prepare an alternative PR with an option to opt-in for nested static forwarders. However I don't think we should hard-fail in that case. We can't hard-fail for top-level static forwarders, because scalac does not hard-fail for them, and we could break actual things that work today. But then if don't hard-fail for top-level ones, but we do hard-fail for nested ones, that's going to become very tricky to handle in the code.

gzm0: Hmmm... But is the concern about not hard-failing mostly driven by how hard it is to implement or how hard it will be to use? It seems that in any case we want our behavior for nested classes to be different from our behavior for top-level classes (because on the top-level scalac precedent is more important). But at that point, shouldn't we pick what is easiest to use and not what is easiest to build?

gzm0: Ohh... Or are we aligning our behavior for *both* cases in a way that improves on scalac's behavior? In that case, that would indeed be better. But then we're back to the warning, where I'm not sure *not* having a warning in this case is better. Maybe we should just add an annotation to silence the warning (just like `uncheckedVariance` and the like).

sjrd: IMO the ideal would be to use the code of this PR, but emit a warning every time we discard a static forwarder class, with the ability to silence that warning. Since Scala 2.13.2 ([release notes](https://github.com/scala/scala/releases/tag/v2.13.2)), there is a mechanism for configurable warnings, that we could/should integrate with so that users can silence those warnings if they want. But that leaves a problem with older versions of Scala, where that's not possible. In the past we had "transitioning" compiler plugin options to selectively silence some kinds of warnings (notably related to all the Export-related features we deprecated in 0.6.17 or so). Perhaps we could introduce another one like that.

sjrd: I've added a commit with a warning that is configurable with the Scala 2.13.2+ mechanism. On previous versions of Scala, it is not configurable, however.

gzm0: I think this is good. That will allow folks to use `-Xfatal-warnings` at least on 2.13. If somebody really wants to disable it, we can still add a compiler flag for other versions.

Can we have a compiler test for it though?

sjrd: It turns out that this break 3 partest tests. I'm starting to think that this thing is much more widespread than I thought. So in addition of the warning (for which I'll add a test), I'll add a Scala.js option to opt-in to generating the static forwarders of non-top-level objects, as you suggested earlier.

sjrd:
Superseded by #4152.

PR #4150, opened 2020-08-17
Expose js.Array.from

exoego:
_Cutout from PR #4070_

This PR exposes [`js.Array.from(js.Iterable)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from).

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/Array.scala
+++ b/library/src/main/scala/scala/scalajs/js/Array.scala
@@ -178,6 +178,7 @@ object Array {
   @JSGlobal("Array")
   private object NativeArray extends js.Object {
     def isArray(arg: scala.Any): Boolean = js.native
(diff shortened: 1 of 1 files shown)


library/src/main/scala/scala/scalajs/js/Array.scala
```diff
@@ -186,4 +187,8 @@ object Array {
   /** Returns true if the given value is an array. */
   @inline
   def isArray(arg: scala.Any): Boolean = NativeArray.isArray(arg)
+
+  /** Creates a new array from js.Iterable */
```
sjrd, line 191: This Scaladoc should have the ECMAScript 2015 badge:
```
<span class="badge badge-ecma2015" style="float: right;">ECMAScript 2015</span>
```

exoego: Added badge and squashed commits.

PR #4152, opened 2020-08-18
Fix #4148: Do not emit static forwarders for non-top-level objects by default.

### LINKED ISSUE #4148: Unable to find extant inner trait IR during fastopt

sjrd:
Alternative to #4149.

---

### Fix #4148: Do not emit static forwarders for non-top-level objects by default.

Scala/JVM does not emit static forwarders for non-top-level objects. The reason we do it in Scala.js is for implementations of JDK classes that contain static methods in inner static classes (#3950). However, this causes issues with existing Scala projects that have non-top-level `object`s and `class`es with names differing only in case.

In this commit, we now only emit static forwarders for top-level objects by default, and add an opt-in option to emit them for non-top-level objects, which should be used by implementations of JDK classes.

---

### Do not erase a regular class with a static forwarder class.

This is only relevant for case insensitive file systems. On such file systems, a static nested object could generate a static forwarder class whose name differed from a regular class' name only in case, therefore erasing the latter on case insensitive file systems.

We now avoid emitting static forwarder classes for objects if they would clash with a regular class on a case insensitive file system.

When doing so, we emit a warning. On Scala 2.13.2+, the warning can be silenced with `@nowarn("other")` on the object, or `-Wconf:cat=other:s` in the scalac options.

For top-level objects, the warning is similar to a warning already emitted by the JVM back-end. For non-top-level objects, it only happens in Scala.js, and only when they are emitted due to the opt-in option.

This commit complements its parent in fixing #4148 for good, as now that issue won't present itself even when we opt in for static forwards of non-top-level objects. We therefore add a test for that issue, which we could not do before because the test suite as a whole requires the opt-in for other reasons.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -1008,16 +1008,22 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
      * back-end considers them as colliding because they have the same name,
      * but we must not.
      *
-     * Further, unlike scalac, we emit forwarders for *any* static object, not
-     * just top-level ones. This is important so we can implement static methods
+     * By default, we only emit forwarders for top-level objects, like scalac.
+     * However, if requested via a compiler option, we enable them for all
+     * static objects. This is important so we can implement static methods
      * of nested static classes of JDK APIs (see #3950).
      */
 
     /** Is the given Scala class, interface or module class a candidate for
      *  static forwarders?
      */
-    def isCandidateForForwarders(sym: Symbol): Boolean =
-      !settings.noForwarders && sym.isStatic && !isImplClass(sym)
+    def isCandidateForForwarders(sym: Symbol): Boolean = {
+      !settings.noForwarders && sym.isStatic && !isImplClass(sym) && {
+        // Reject non-top-level objects unless opted in via the appropriate option
+        scalaJSOpts.genStaticForwardersForNonTopLevelObjects ||
+        !sym.name.containsChar('$') // this is the same test that scalac performs
+      }
+    }
 
     /** Gen the static forwarders to the members of a class or interface for
      *  methods of its companion object.
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -217,6 +217,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
 
     private val lazilyGeneratedAnonClasses = mutable.Map.empty[Symbol, ClassDef]
     private val generatedClasses = ListBuffer.empty[js.ClassDef]
+    private val generatedStaticForwarderClasses = ListBuffer.empty[(Symbol, js.ClassDef)]
 
     private def consumeLazilyGeneratedAnonClass(sym: Symbol): ClassDef = {
       /* If we are trying to generate an method as JSFunction, we cannot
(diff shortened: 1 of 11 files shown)


project/Build.scala
```diff
@@ -1026,6 +1026,8 @@ object Build {
        * flag prevents these mistakes from happening.
        */
       scalacOptions += "-Yno-predef",
+      // We also opt in to emitting static forwarders for non-top-level objects
```
gzm0, line 1029: IMO this comment is not helpful. It describes *what* the code does not *why* it does it. I think a comment explaining why is warranted here.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -364,14 +365,62 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+           * systems.
+           */
+
+          /* I could not find any reference anywhere about what locale is used
+           * by case insensitive file systems to compare case-insensitively.
+           * In doubt, force the English locale, which is probably going to do
+           * the right thing in virtually all cases (especially if users stick
+           * to ASCII class names), and it has the merit of being deterministic,
+           * as opposed to using the OS' default locale.
+           * The JVM performs a similar test to emit a warning for conflicting
```
gzm0, line 387: JVM *backend*?

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RegressionTest.scala
```diff
@@ -906,4 +914,24 @@ object RegressionTest {
   object `class` { // scalastyle:ignore
     def foo(x: Int): Int = x + 1
   }
+
+  /* The objects and classes here intentionally have names that differ
+   * only in case, and are intentionally defined in a specific order.
```
gzm0, line 919: Consider commenting *why* this is intentional.

sjrd:
Updated.

PR #4159, opened 2020-08-24
Fix #4157: Upgrade to Scala 2.12.12 and 2.13.3.

### LINKED ISSUE #4157: Update to 2.12.12 and 2.13.3
It's been a while since those versions have been released, but we still haven't upgraded our build to use them. We should do that before the next release.

sjrd:


scala-test-suite/src/test/resources/2.13.3/BlacklistedTests.txt
```diff
@@ -0,0 +1,218 @@
...
+scala/sys/process/ParserTest.scala
+scala/sys/process/PipedProcessTest.scala
+scala/sys/process/ProcessBuilderTest.scala
+scala/sys/process/ProcessTest.scala
+scala/tools/testkit/AssertUtilTest.scala
+scala/util/PropertiesTest.scala
+scala/util/SpecVersionTest.scala
+scala/util/SystemPropertiesTest.scala
+
+# Do not link because they use AssertUtil whose constructor refers to java.lang.Field
```
sjrd, line 184: This is new. `AssertUtil` receive a `private lazy val` field of type `java.lang.Field`, which pollutes all of `AssertUtil` for Scala.js. I intend to submit a change upstream to recover those tests in the next version.

gzm0: Maybe file an issue to make sure we don't forget to remove these when upgrading to 2.13.4?

gzm0 (approved):

scala-test-suite/src/test/resources/2.13.3/BlacklistedTests.txt
sjrd: #4161

PR #4163, opened 2020-08-25
Generalize classVar and coreJSLibVar as globalVar with a Scope

gzm0:


gzm0:
We might also consider renaming `classVar` to `globalVar`. If you think this is a good idea, I'll add a follow-up commit.

gzm0:
https://github.com/scala-js/scala-js/compare/master...gzm0:split-module-simplify-var?expand=1

Shows how this looks past module splitting.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -134,64 +88,13 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
-      case LongType    => "J"
-      case FloatType   => "F"
-      case DoubleType  => "D"
-      case NullType    => "N"
-      case NothingType => "E"
-    }
-    genericIdent(field, subField)
-  }
-
   def fileLevelVar(field: String, subField: String,
```
sjrd, line 98: Why didn't `fileLevelVar` receive the same `Scope`-based treatment?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -204,31 +107,26 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
 
-  private def genericIdent(field: String, subField: String,
+  private def genericIdent(field: String, subFieldParts: List[String],
       origName: OriginalName = NoOriginalName)(
       implicit pos: Position): Ident = {
-    Ident(avoidClashWithGlobalRef("$" + field + "_" + subField), origName)
+    if (subFieldParts.isEmpty) {
+      Ident(avoidClashWithGlobalRef("$" + field), origName)
+    } else {
+      val subField = subFieldParts.mkString("__")
```
sjrd, line 127: Have you considered pushing this computation, and the `if/else`, down inside the method of `Scope`? That might save on `List` creations and iterations, which currently seem a bit wasteful.

sjrd:
> We might also consider renaming `classVar` to `globalVar`. If you think this is a good idea, I'll add a follow-up commit.

`globalVar` would read like a JS global var (like a global ref). But perhaps `codegenVar`, like was suggested at some point in the past?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
gzm0: Because for `fileLevelVar`, the scope determination is at the call site (`subField` is merely a string). For modules, this becomes even more relevant, since `Scope` needs to provide a `reprClass` to determine the module the field is in. For `fileLevelVar`s we cannot determine this module.

We could introduce different types of `Scope` for this, but it seems overkill, since `VarGen` would still have to offer different methods for this.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
gzm0: Good point. I just used `""` as an empty sub-field sentinel.

gzm0:
>  But perhaps `codegenVar`, like was suggested at some point in the past?

We still need a distinction between vars that are reachable in the entirety of the created JS program and vars that are reachable in the individual file only. IMO `codegenVar` doesn't make this clear.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
sjrd: Ah right, that makes sense. And yes, given that, probably `globalVar` is the most appropriate name.

gzm0:
> And yes, given that, probably globalVar is the most appropriate name.

I have added a commit renaming `classVar` (and variants) to `globalVar`.

sjrd:
LGTM, though I think the two commits should be squashed, with a commit message that says something along the lines of "generalize classVar and coreJSLibVar as globalVar with a Scope argument". Changing all the `coreJSLibVar`s to `classVar`s, then changing them back to `globalVar` doesn't make sense, IMO.

gzm0:
Updated.

sjrd (approved):

PR #4164, opened 2020-08-25
Separate usage of "b" and "c" fields

gzm0:
"c" is now only used for Scala classes.
"b" is now only used for JS classes ("a" is the accessor).

Since we now generate ES6 classes in expression position, we adjust
ClosureAstTransformer to support this.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
@@ -43,121 +43,113 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
   import nameGen._
   import varGen._
 
-  def buildClass(tree: LinkedClass, ctor: WithGlobals[js.Tree],
-      memberDefs: List[WithGlobals[js.Tree]], exportedDefs: WithGlobals[js.Tree])(
+  def buildClass(tree: LinkedClass, ctor: js.Tree, memberDefs: List[js.MethodDef],
+      exportedDefs: js.Tree)(
       implicit globalKnowledge: GlobalKnowledge): WithGlobals[js.Tree] = {
 
-    WithGlobals.list(ctor +: memberDefs :+ exportedDefs).flatMap { allDefs =>
-      val className = tree.name.name
-      val allDefsBlock = js.Block(allDefs)(tree.pos)
+    implicit val pos = tree.pos
+
+    val className = tree.name.name
+    def allES6Defs = {
+      js.Block(ctor +: memberDefs :+ exportedDefs)(tree.pos) match {
+        case js.Block(allDefs) => allDefs
+        case js.Skip()         => Nil
+        case oneDef            => List(oneDef)
+      }
+    }
+
+    def allES5Defs(classVar: js.Tree) = {
+      WithGlobals(js.Block(
+          ctor, assignES5ClassMembers(classVar, memberDefs), exportedDefs))
+    }
+
+    if (!tree.kind.isJSClass) {
+      assert(tree.jsSuperClass.isEmpty, className)
+
+      if (useClasses) {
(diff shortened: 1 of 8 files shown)


gzm0:
@sjrd do we need to adjust `StackTrace.scala` and `StackTraceTest.scala` with this? (`StackTraceTest.scala` currently doesn't test JS classes, and I'd expect them to break with this, if they worked before).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -253,7 +241,7 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
     } yield {
       val docComment = js.DocComment("@constructor")
       val ctorDef =
-        if (isJSClass) globalVarDef("c", className, ctorFun)
+        if (isJSClass) fileLevelVar("b", genName(className)) := ctorFun
```
sjrd, line 244: Isn't `typeVar` still visible here?
```suggestion
        if (isJSClass) typeVar := ctorFun
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -43,121 +43,113 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
-              val ident = fileLevelVarIdent("cc", genName(param.name.name),
-                  param.originalName.orElse(param.name.name))
-              js.ParamDef(ident, rest = false)
-            }
+      val classValueIdent = fileLevelVarIdent("b", genName(className))
+      val classValueVar = js.VarRef(classValueIdent)
+
+      val entireClassDefWithGlobals = if (useClasses) {
+        genJSSuperCtor(tree).map { jsSuperClass =>
+          classValueVar := js.ClassDef(None, Some(jsSuperClass), allES6Defs)
```
sjrd, line 97: We should still provide the `classValueIdent` as the `name` of the `js.ClassDef`, even when used in expression position. This makes sure that its `.name` property will be correctly filled in by the engine, and that overall the class and its methods appear in a better way in debugging tools.

sjrd (changes requested):

sjrd:
> 
> 
> @sjrd do we need to adjust `StackTrace.scala` and `StackTraceTest.scala` with this? (`StackTraceTest.scala` currently doesn't test JS classes, and I'd expect them to break with this, if they worked before).

We should probably adjust `StackTrace.scala` for `$b_` prefixes, yes.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/SJSGen.scala
```diff
@@ -377,4 +377,18 @@ private[emitter] final class SJSGen(
       case ComputedName(tree)  => genBracketSelect(qual, tree)
     }
   }
+
+  def assignES5ClassMembers(classRef: Tree, members: List[MethodDef])(
```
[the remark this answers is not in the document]
gzm0: It was there because of `genPropSelect`. I have move that one as wel.

gzm0:
Updated. I have added a test in `StackTraceTest`. However, it doesn't fail after the change, because the frame gets detected on the static method (on the `s` field) that implements the exported method `"m"`.

IDK how to craft a test that doesn't fall victim to this.

sjrd:
Right. I can't imagine a way to craft such a test either. The good news is that it probably means that the bad behavior is indeed not observable anyway, and that therefore we don't need to change anything.

sjrd (approved):

gzm0:
Using `@JSName` might actually do the trick. We can then assert both names (the Scala name and the JS name). I'll try that.

gzm0:
It worked :)

sjrd (approved):

PR #4167, opened 2020-08-27
Add Character.codePointCount

er1c:


javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -254,6 +254,32 @@ object Character {
   @inline def toCodePoint(high: scala.Char, low: scala.Char): Int =
     ((high & SurrogateUsefulPartMask) << 10) + (low & SurrogateUsefulPartMask) + 0x10000
 
+
+  def codePointCount(seq: CharSequence, beginIndex: Int, endIndex: Int): Int = {
```
sjrd, line 258: We already have an implementation of this functionality in `java.lang.String`:
https://github.com/scala-js/scala-js/blob/bb4835763b9316c8370bc00b6bd4f6c0653a0e95/javalanglib/src/main/scala/java/lang/_String.scala#L82-L94
Since a String is a CharSequence, we should move the existing implementation of `java.lang.String.codePointCount` here, and change the former to forward to this one.

But clearly we should not invent a new implementation.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,27 @@ class CharacterTest {
     assertFalse(Character.isDigit('\uFBFC'))
   }
 
+  @Test def codePointCount(): Unit = {
+    assertEquals(1, Character.codePointCount("\uD800\uDC00", 0, 2))
+    assertEquals(1, Character.codePointCount("\uD800\uDC01", 0, 2))
+    assertEquals(1, Character.codePointCount("\uD801\uDC01", 0, 2))
+    assertEquals(1, Character.codePointCount("\uDBFF\uDFFF", 0, 2))
+    assertEquals(3, Character.codePointCount("a\uD800\uDC00b", 0, 4))
+    assertEquals(4, Character.codePointCount("a\uD800\uDC00b\uD800", 0, 5))
```
sjrd, line 392: Those tests cover only cases where `beginIndex = 0` and `endIndex = length`.

But again, we already have more complete tests for that in `StringTest`:
https://github.com/scala-js/scala-js/blob/bb4835763b9316c8370bc00b6bd4f6c0653a0e95/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/StringTest.scala#L176-L195
We should use the same tests.

We should also test with `CharSequence`s that are *not* strings. We have [`WrappedStringCharSequence`](https://github.com/scala-js/scala-js/blob/master/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/WrappedStringCharSequence.scala) in the test suite for that purpose. So we should duplicate all the tests with instances of `WrappedStringCharSequence`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,27 @@ class CharacterTest {
...
+
+    // "No NPE, null char sequence."
+    assertThrows(classOf[NullPointerException], Character.codePointCount(null.asInstanceOf[CharSequence], 0, 1))
```
sjrd, line 395: As I said last time, we normally don't test NPEs, because they're Undefined Behavior.

sjrd (changes requested):

sjrd (approved):

er1c:
@sjrd whitespace fixed

sjrd (approved):

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -254,6 +254,20 @@ object Character {
...
+    val end = endIndex - 1
+    while (i < end) {
+      if (Character.isSurrogatePair(seq.charAt(i), seq.charAt(i + 1)))
+        res -= 1
+      i += 1
+    }
+    res
+  }
+
   // --- End of UTF-16 surrogate pairs handling ---
```
gzm0, line 271: This comment should be above IMO.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -254,6 +254,20 @@ object Character {
...
     ((high & SurrogateUsefulPartMask) << 10) + (low & SurrogateUsefulPartMask) + 0x10000
 
+  def codePointCount(seq: CharSequence, beginIndex: Int, endIndex: Int): Int = {
+    if (endIndex > seq.length() || beginIndex < 0 || endIndex < beginIndex)
+      throw new IndexOutOfBoundsException
+    var res = endIndex - beginIndex
+    var i = beginIndex
+    val end = endIndex - 1
+    while (i < end) {
+      if (Character.isSurrogatePair(seq.charAt(i), seq.charAt(i + 1)))
```
gzm0, line 264: Is it intentional that this call is qualified with `Character`? We are *in* `Character` here.

gzm0 (changes requested):

javalanglib/src/main/scala/java/lang/Character.scala
er1c: mostly just because I wanted to show the straight-move/no changes for the original code.  Happy to update this though!

javalanglib/src/main/scala/java/lang/Character.scala
er1c: updated

javalanglib/src/main/scala/java/lang/Character.scala
er1c: updated/removed

gzm0 (approved):

PR #4169, opened 2020-08-27
Add 'sbt scalastyleCheck' and use it in Jenkinsfile.

er1c:


build.sbt
```diff
@@ -36,3 +36,48 @@ val partestSuite = Build.partestSuite
...
+  "scalaTestSuite"
+)
+
+// Alias for developer usage
+addCommandAlias("scalastyle",  
+  "; scalajs/test:scalastyle; linkerPrivateLibrary/test:scalastyle; sbtPlugin/test:scalastyle" + 
+  ";" + aggregatorIDs.map(id => s"${id}2_11/test:scalastyle").mkString(";") +
+  ";" + aggregatorIDs.map(id => s"${id}2_12/test:scalastyle").mkString(";") +
+  ";" + aggregatorIDs.map(id => s"${id}2_13/test:scalastyle").mkString(";")
+)
```
gzm0, line 83: We already have machinery for this in `Build.scala`. Look for `allProjects`. Adding `scalastyle` should be a single line change.

er1c: @gzm0 thanks for the pointer.   However, I'm struggling with the correct SBT configuration since `scalastyle` isn't a `TaskKey` (https://github.com/scalastyle/scalastyle-sbt-plugin/blob/master/src/main/scala/org/scalastyle/sbt/Plugin.scala#L51)  

```
import org.scalastyle.sbt.ScalastylePlugin.autoImport._
val scalaStyleTask = taskKey[Unit]("scalaStyle")
scalaStyleTask := scalastyle.toTask("").value,

...

        val keys = Seq[TaskKey[_]](
            clean, headerCreate in Compile, headerCreate in Test,
            headerCheck in Compile, headerCheck in Test, scalaStyleTask in Test
        )

```

Results iin  something like:

```
[error]
[error]   scalaTestSuite2_12 / Test / scalaStyleTask from Test / scalaStyleTask (Build.scala:587)
[error]
[error]   scalaTestSuite2_13 / Test / scalaStyleTask from Test / scalaStyleTask (Build.scala:587)
[error]
[error]   scalastyle from scalaStyleTask ((build.Build.root) Build.scala:559)
[error]      Did you mean Compile / scalastyle ?

```

Any ideas?

gzm0: I didn't realize `scalaStyle` is an `InputTask`. If you want to define a `Task` like this, you need to put the task definition into `commonSettings` so it gets applied to all projects (n.b. you need to do this in all configurations).

er1c: @gzm0 I tried this, but couldn't get it working - spent several hours trying  to without luck. In the example code above, I have the `scalaStyleTask := scalastyle.toTask("").value,` in the command settings, or do you mean the actual val?

er1c: Is there any reason not to include the  scalastyle in the normal compile (e.g. https://www.scala-sbt.org/1.x/docs/Howto-Dynamic-Task.html) I guess I was just trying to add it as a separate task to avoid making any "compile" changes, but this seems more straightforward

sjrd: Yes, there is a reason: because we don't want to spend time running the Scalastyle checker every time we compile ;)

er1c: @sjrd that's fair, and also validates my intuition on not going down that path.  I *think* just figured it out, PR is updated, I'm doing some additional testing, is it worth updating the travis build to use this instead of the manual list?

project/Build.scala
```diff
@@ -249,6 +252,10 @@ object Build {
           "-encoding", "utf8"
       ),
 
+      scalastyleCheck := Def.taskDyn {
```
gzm0, line 255: Does this really need to be `dyn`? Ugh. (I have never used `toTask` on an `InputTask`, so I'm genuinely asking).

project/Build.scala
```diff
@@ -249,6 +252,10 @@ object Build {
...
+        ScalastylePlugin.autoImport.scalastyle.in(Test).toTask("")
```
gzm0, line 256: Does running this in the `Test` configuration also check the normal code? (Our `Jenkinsfile` suggests otherwise, but it might be over-checking).

gzm0: Also, write this:

```scala
(scalastyle in Test).toTask("")
```

And import `scalastyle` for the entire file?

DEVELOPING.md
```diff
@@ -101,6 +101,13 @@ You will still have to fix a few things:
 * Uncheck the "Allow output directories per source directory" in Build path
 * Add transitive project dependencies in Build path
 
+## Preparing a Pull Request
+
+One common build failure is code styling.  Reproduce results locally with:
```
gzm0, line 106: SJS doesn't do double spaces. Please be consistent with the rest of the file.

gzm0 (commented):
> I'm doing some additional testing, is it worth updating the travis build to use this instead of the manual list?

Definitely, one less place we can forget to add projects :)

project/Build.scala
er1c: ahh yeah, I was originally trying to keep the name of the task `scalastyle` so was trying to avoid the double naming, but that didn't work out, I'll update the import appropriately.  I'll double check the `taskDyn` that was cargo cult on my part from examples I could find.

DEVELOPING.md
er1c: 👍

project/Build.scala
er1c: `taskDyn` is required here, nothing runs if `Def.task`

er1c: It looks like both in Test and in Compile are required :/ trying to figure out if I need to have two tasks or if I can put them both in the same one

gzm0: > nothing runs if Def.task

Nothing runs or it fails to compile? If it doesn't run, you might be creating a dangling value. Try:

```scala
scalaStyleCheck := (scalastyle in Test).toTask("")

// or

scalaStyleCheck := Def.task {
  (scalastyle in Test).toTask("").value
}.value
```

er1c: Adding the extra `.value` seemed to do the trick.

Full code got updated to:

```
      scalastyleCheck := Def.task {
        val _ = (scalastyle in Compile).toTask("").value
        (scalastyle in Test).toTask("").value
      }.value,
```

gzm0 (approved):

er1c:
@gzm0 @sjrd I don't think the jenkinsbot picked up the `Jenkinsfile` changes, based upon my local testing I think https://github.com/scala-js/scala-js/pull/4172 will need to be merged before the `Jenkinsfile` commit gets merged.  Should I remove that commit from this PR to keep the `Build.scala` changes separate?

sjrd:
No, indeed. The Jenkins bot doesn't "trust" non-maintainers to edit Jenkinsfile, so during PR validation it ignores the changes and keeps using the `Jenkinsfile` on master. I did not find a way to configure it otherwise.

er1c:
I removed the `Jenkinsfile` changes to simplify this and still allow running the command locally for development.

gzm0:
@sjrd I believe this is waiting on you.

sjrd (changes requested):
Running `scalastyleCheck` fails on the following projects:
* library and libraryAux, but that would be fixed with a rebase I believe,
* scalalib and partest, for which it shouldn't be run.

The command should somehow do nothing for `scalalib` and `partest`. Otherwise running it is just confusing as it reports errors even though a PR would accept the code.

project/Build.scala
```diff
@@ -546,6 +554,24 @@ object Build {
...
+  implicit class ProjectTaskKeyFilter(val p: Project) extends AnyVal {
+    def include(key: TaskKey[_]): Boolean = {
+      if (key == scalastyleCheck) {
+        val isIgnore = ProjectTaskKeyFilter.ignoredStyleCheckProjects.contains(p)
+        !isIgnore
+      } else {
+        true
+      }
+    }
+  }
```
gzm0, line 573: Instead of this, consider simply overriding `scalastyleCheck` in the projects where we do not want it.

gzm0 (changes requested):

project/Build.scala
er1c: Updated, see if this approach matches what you are suggesting here

project/Build.scala
```diff
@@ -2010,6 +2021,10 @@ object Build {
         }
       },
 
+      // Ignore scalastyle for this project
+      scalastyleCheck := {},
+
```
er1c, line 2026: extra newlines ugh

build.sbt
```diff
@@ -36,3 +36,5 @@ val partestSuite = Build.partestSuite
 val scalaTestSuite = Build.scalaTestSuite
 
 inThisBuild(Build.thisBuildSettings)
+
+addCommandAlias("scalastyle", "; scalastyleCheck")
```
gzm0, line 40: I'm not sure this indirection is necessary. Invoking `scalastyleCheck` is just as fine, no?

er1c: `scalastyle` is the usual command people run, so it's a bit more "friendly", but no, it is not necessary, thoughts?  I originally had just `scalastyleCheck` but after finding that this worked, I added it for my own muscle memory.  I'll defer to you and @sjrd.  Should I remove it?

sjrd: I have a slight preference for not having the command alias, and running `scalastyleCheck` instead. This is mostly because I don't like command aliases that shadow tasks that otherwise exist.

er1c: Removed, and updated Jenkinsfile/etc to use `scalastyleCheck`.  Thanks!

er1c:
@sjrd @gzm0 CI passing

gzm0 (approved):

sjrd (approved):

PR #4170, opened 2020-08-30
Simplify TopLevel and Static export handling

gzm0:


compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
```diff
@@ -71,189 +71,147 @@ trait GenJSExports[G <: Global with Singleton] extends SubComponent {
...
+      var bad = false
 
-      if (ctorExports.isEmpty) {
-        Nil
-      } else {
-        val exports = for {
-          (jsName, specs) <- ctorExports.groupBy(_._1.jsName) // group by exported name
-        } yield {
-          val ctors = specs.map(s => ExportedSymbol(s._2))
+      for ((info, sym) <- tups.tail) {
```
sjrd, line 101: `.tail` is not necessarily efficient on arbitrary `Iterable`s. Can you take a `List` instead? Otherwise, it might be better to iterate over `tups` rather than `tups.tail`.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
```diff
@@ -71,189 +71,147 @@ trait GenJSExports[G <: Global with Singleton] extends SubComponent {
...
-          }
+    private def checkSingleField(tups: Iterable[(jsInterop.ExportInfo, Symbol)]): Symbol = {
+      assert(tups.nonEmpty, "must have at least one export")
 
-          js.TopLevelMethodExportDef(methodDef)
-        }
+      val firstSym = tups.head._2
 
-        exports.toList
+      for ((info, _) <- tups.tail) {
```
sjrd, line 120: Same potential issue with `tail`.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSExports.scala
```diff
@@ -71,189 +71,147 @@ trait GenJSExports[G <: Global with Singleton] extends SubComponent {
...
+
+      firstSym
     }
 
-    def genJSClassExports(
-        classSym: Symbol): List[js.TopLevelJSClassExportDef] = {
-      for {
-        exp <- jsInterop.registeredExportsOf(classSym)
+    def genTopLevelExports(classSym: Symbol): List[js.TopLevelExportDef] = {
+      val exports = for {
```
sjrd, line 129: I suggest adding a `.toList` to the result of this for comprehension to address the issues with `.tail` above. If `exports` is a List, the `tups` should naturally become a `List` as well.

sjrd (changes requested):
Looks good. I only have minor comments.

gzm0:
The AppVeyor build seems to have suffered an infrastructure failure. Is there a way I can restart it?

sjrd:
> The AppVeyor build seems to have suffered an infrastructure failure. Is there a way I can restart it?

If you login into AppVeryor with your GitHub account, you should be able to trigger re-builds. I've done it for this build now.

sjrd (approved):

gzm0:
> If you login into AppVeryor with your GitHub account, you should be able to trigger re-builds. I've done it for this build now.

I can't find the button. Are you sure I have the appropriate permissions? If I click on "Projects" in the nav bar, the scala-js repository does not show up.

sjrd:
You're an Owner of the scala-js organization, so you should have all privileges. Perhaps you need to explicitly sync your AppVeyor profile with your GitHub permissions?

gzm0:
> Perhaps you need to explicitly sync your AppVeyor profile with your GitHub permissions?

How do I do that? I have tried: Installing the AppVeyor GitHub app for myself (gzm0) and the scala-js organization.

sjrd:
Apparently the AppVeyor builds are configured with my personal account, instead of through the scala-js org. The problem is that I have no idea how to fix it, nor even to recreate the project under the scala-js org from the AppVeyor point of view :s

PR #4173, opened 2020-09-01
Add java.lang.Character.UnicodeBlock

er1c:
~~The audit was done against JDK14, I need to add an audit against jdk8.~~ JDK14 constants are a superset of JDK8.

~~TODO: Potentially use https://unicode.org/Public/6.3.0/ucd/Blocks.txt~~

~~According to: https://exploringjs.com/es6/ch_unicode.html & http://www.ecma-international.org/ecma-262/6.0/#sec-conformance~~

Generated using adopt openjdk 1.8

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Character.scala
+++ b/javalanglib/src/main/scala/java/lang/Character.scala
@@ -12,9 +12,10 @@
 
 package java.lang
 
+import scala.annotation.tailrec
 import scala.scalajs.js
 
-import java.util.Arrays
+import java.util.{ArrayList, Arrays, HashMap}
 
 /* This is a hijacked class. Its instances are primitive chars.
  *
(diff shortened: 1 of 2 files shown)


scripts/GenerateUnicodeBlock.sc
```diff
@@ -0,0 +1,411 @@
...
+import java.net.URL
+import java.nio.charset.StandardCharsets
+
+val blocksSourceURL: URL = new URL("http://unicode.org/Public/UCD/latest/ucd/Blocks.txt")
+
+// Generated/created via Mapping Worksheet:
+//   https://docs.google.com/spreadsheets/d/174aedhJlWNZYEEJtSVcIiMFvD7f27rg4NssjFw0hIDw/edit?usp=sharing
+
+val ucdBlockNameToJDKBlockName: Map[String, String] = Map(
+  "Adlam" -> "ADLAM",
```
sjrd, line 13: What's the point of a script if we need a huge list in the script itself anyway?

I would much rather have a script similar to the ones we already have in `Character`, which work by looping over all the possible code points and extracting results from the JVM itself. See for example the script of `Character.charTypeIndices`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+    val CJK_UNIFIED_IDEOGRAPHS_EXTENSION_E = new UnicodeBlock("CJK_UNIFIED_IDEOGRAPHS_EXTENSION_E", 0x2B820, 0x2CEAF)
+    val CJK_UNIFIED_IDEOGRAPHS_EXTENSION_F = new UnicodeBlock("CJK_UNIFIED_IDEOGRAPHS_EXTENSION_F", 0x2CEB0, 0x2EBEF)
+    val CJK_COMPATIBILITY_IDEOGRAPHS_SUPPLEMENT = new UnicodeBlock("CJK_COMPATIBILITY_IDEOGRAPHS_SUPPLEMENT", 0x2F800, 0x2FA1F)
+    val TAGS = new UnicodeBlock("TAGS", 0xE0000, 0xE007F)
+    val VARIATION_SELECTORS_SUPPLEMENT = new UnicodeBlock("VARIATION_SELECTORS_SUPPLEMENT", 0xE0100, 0xE01EF)
+    val SUPPLEMENTARY_PRIVATE_USE_AREA_A = new UnicodeBlock("SUPPLEMENTARY_PRIVATE_USE_AREA_A", 0xF0000, 0xFFFFF)
+    val SUPPLEMENTARY_PRIVATE_USE_AREA_B = new UnicodeBlock("SUPPLEMENTARY_PRIVATE_USE_AREA_B", 0x100000, 0x10FFFF)
+    // scalastyle:on line.size.limit
+
+    private val allBlocks: Array[UnicodeBlock] = Array(
```
sjrd, line 842: This array could be created at run-time from side effects in the constructor of `UnicodeBlock`. That way we avoid repeating this huge list in the source code of `Character`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+      CJK_UNIFIED_IDEOGRAPHS_EXTENSION_E,
+      CJK_UNIFIED_IDEOGRAPHS_EXTENSION_F,
+      CJK_COMPATIBILITY_IDEOGRAPHS_SUPPLEMENT,
+      TAGS,
+      VARIATION_SELECTORS_SUPPLEMENT,
+      SUPPLEMENTARY_PRIVATE_USE_AREA_A,
+      SUPPLEMENTARY_PRIVATE_USE_AREA_B
+    )
+
+    private val blocksByNormalizedName: java.util.Map[String,UnicodeBlock] = {
```
sjrd, line 1137: This map could also be built at run-time through side-effects in the constructors of `UnicodeBlock`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+      var low: Int  = 0
+      var mid: Int  = -1
+      var high: Int = allBlocks.length - 1
+
+      while (low <= high) {
+        mid = (low + high) >>> 1
+        val block = allBlocks(mid)
+        if (codePoint > block.end) {
+          low = mid + 1
+        } else if (codePoint >= block.start && codePoint <= block.end) {
```
sjrd, line 1473:
```suggestion
        } else if (codePoint >= block.start) {
```
The removed part is already known to be true because of the first `if`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+
+    def of(codePoint: scala.Int): UnicodeBlock = {
+      // scalastyle:off return
+      if (!Character.isValidCodePoint(codePoint)) {
+        throw new IllegalArgumentException()
+      }
+
+      var low: Int  = 0
+      var mid: Int  = -1
+      var high: Int = allBlocks.length - 1
```
sjrd, line 1466: In the Scala.js coding style, the only thing we ever vertically align are `=>` of consecutive one-liner `case`s.
```suggestion
      var low: Int = 0
      var mid: Int = -1
      var high: Int = allBlocks.length - 1
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+
+    ////////////////
+    // End Generated
+    ////////////////
+
+    def forName(blockName: String): UnicodeBlock = {
+      val key: String = toNormalizedName(blockName)
+      if (!blocksByNormalizedName.containsKey(key)) throw new IllegalArgumentException()
+
+      blocksByNormalizedName.get(key)
```
sjrd, line 1453: Instead of doing `containsKey` first, you could check whether the result of `get` is `null`. This avoids a double-lookup in the map.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +518,971 @@ object Character {
...
+    }
+
+    // From: https://unicode.org/Public/UCD/latest/ucd/Blocks.txt
+    // Note:   When comparing block names, casing, whitespace, hyphens,
+    //         and underbars are ignored.
+    //         For example, "Latin Extended-A" and "latin extended a" are equivalent.
+    //         For more information on the comparison of property values,
+    //           see UAX #44: http://www.unicode.org/reports/tr44/
+    private def toNormalizedName(name: String): String = {
+      name.toLowerCase.replaceAll(raw"[\s\-_]", "")
```
sjrd, line 1442: This may follow what the Unicode standard says about how to compare block names, but it does not match the specification of [`forName`](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.UnicodeBlock.html#forName-java.lang.String-), which only accepts some specific forms. For example:
```scala
scala> Character.UnicodeBlock.forName("Basic Latin")
res4: Character.UnicodeBlock = BASIC_LATIN

scala> Character.UnicodeBlock.forName("Basic LatiN")
res6: Character.UnicodeBlock = BASIC_LATIN

scala> Character.UnicodeBlock.forName("Basic Lat iN")
java.lang.IllegalArgumentException
  at java.lang.Character$UnicodeBlock.forName(Character.java:3190)
  ... 28 elided

scala> Character.UnicodeBlock.forName("Basic Lat in")
java.lang.IllegalArgumentException
  at java.lang.Character$UnicodeBlock.forName(Character.java:3190)
  ... 28 elided

scala> Character.UnicodeBlock.forName("Arabic Extended-A")
res15: Character.UnicodeBlock = ARABIC_EXTENDED_A

scala> Character.UnicodeBlock.forName("Arabic_Extended-A")
java.lang.IllegalArgumentException
  at java.lang.Character$UnicodeBlock.forName(Character.java:3190)
  ... 28 elided

scala> Character.UnicodeBlock.forName("Arabic_Extended_A")
res17: Character.UnicodeBlock = ARABIC_EXTENDED_A

scala> Character.UnicodeBlock.forName("Arabic_ExtendedA")
java.lang.IllegalArgumentException
  at java.lang.Character$UnicodeBlock.forName(Character.java:3190)
  ... 28 elided

scala> Character.UnicodeBlock.forName("ArabicExtendedA")
java.lang.IllegalArgumentException
  at java.lang.Character$UnicodeBlock.forName(Character.java:3190)
  ... 28 elided

scala> Character.UnicodeBlock.forName("ArabicExtended-A")
res20: Character.UnicodeBlock = ARABIC_EXTENDED_A
```
Blindly removing all spaces, dashes and underscores is not correct. `toLowerCase` is fine, but after it needs to be an exact match for one of the three possible forms.

sjrd (changes requested):

sjrd:
> > ES6 is based on Unicode version 5.1.0, whereas ES5 is based on Unicode version 3.0.
> 
> Really old versions, I wonder how that compares in practice

The Unicode versions of ECMAScript are irrelevant. They only have an impact on regexes and perhaps on `toUpperCase`/`toLowerCase`, but for the content of this PR they definitely don't matter.

javalanglib/src/main/scala/java/lang/Character.scala
er1c: can you clarify what you mean here?  other than a simple lazy val?

sjrd: I mean creating an empty `ArrayList` first (before the zillion `val`s for all the blocks), and in the constructor of `UnicodeBlock`, have it add itself to the array list with `allBlocks.add(this)`.

er1c: 👍  thanks for the additional hint

er1c:
@sjrd waiting on the CI, but I've updated the approach to generate based upon the JVM and fixed the other remaining comments

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,371 @@ object Character {
...
+        if (codePoint > block.end) {
+          low = mid + 1
+        } else if (codePoint >= block.start) {
+          return block
+        } else {
+          high = mid - 1
+        }
+      }
+
+      null
```
gzm0, line 881: Use `Collections.binarySearch`?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,371 @@ object Character {
...
+    private def addUnicodeBlock(name: String, properName: String, start: Int, end: Int): UnicodeBlock =  {
+      val block = new UnicodeBlock(name, start, end)
+      allBlocks.add(block)
+
+      // Add normalized aliases
+      val lower = properName.toLowerCase
+      //   lowercase + spaces
+      blocksByNormalizedName.put(lower, block)
+      //   lowercase + no spaces
+      blocksByNormalizedName.put(lower.replaceAll(raw"\s", ""), block)
```
sjrd, line 553: The 2.11.x jobs on the CI are failing because 2.11 did not optimize `s"..."` and `raw"..."` interpolators at compile-time. That means that they use the Scala run-time library under the hood, which is not allowed in the javalanglib. We must therefore only use basic `"..."` string literals.

javalanglib/src/main/scala/java/lang/Character.scala
er1c: I thought about this, but didn't explore it.  I don't think the signatures match up, since , with the current collection I'm using, but perhaps can instead leverage something like `charTypeIndices`, and have the ending codepoint in the array doing the searching for the index in `allBlocks` is that generally what you mean/are thinking about?

gzm0: Nah, I just saw a binary search and though "that must be implemented somewhere else" and it is... but because it uses `Comparator` it requires symmetric comparison :-/ So no, you can't really apply it directly.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -12,8 +12,8 @@
 
 package java.lang
 
+import java.util
```
er1c, line 15: I don't think this was needed

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -12,8 +12,8 @@
...
 import scala.scalajs.js
-
```
er1c, line 16: bad IDE

er1c:
@sjrd CI is green & ready for review, thanks!

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
   @inline def compare(x: scala.Char, y: scala.Char): Int =
     x - y
 
+  // UnicodeBlock
+
+  class Subset protected (name: String) {
+    override def equals(that: Any): scala.Boolean = super.equals(that)
+    override def hashCode: scala.Int              = super.hashCode
+    override def toString(): String               = name
```
gzm0, line 527: These need to be final and no alignment.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+  }
+
+  final class UnicodeBlock protected (name: String) extends Subset(name) {
```
gzm0, line 530: Simply:

```scala
final class UnicodeBlock private (name: String, private val start: Int, private val end: Int) extends Subset(name)
```
?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+      this(name)
+      this.start = start
+      this.end = end
+    }
+  }
+
+  object UnicodeBlock {
+    private[this] val totalCodeBlocks: Int = 220
+    private[this] var currentBlockIndex: Int = 0
+    private[this] val allBlocks: Array[UnicodeBlock] = new Array[UnicodeBlock](totalCodeBlocks)
```
gzm0, line 543: Consider using an `ArrayList` so you don't need to hardcode the count? (I doubt it matter for the overall speed)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+        mid = (low + high) >>> 1
+        val block = allBlocks(mid)
+        if (codePoint > block.end) {
+          low = mid + 1
+        } else if (codePoint >= block.start) {
+          return block
+        } else {
+          high = mid - 1
+        }
+      }
```
gzm0, line 884: Consider copying large parts of the implementation of `Collections.binarySearch`. IMHO a tail recursive solution is easier to understand.

I also think you might recurse down into overlapping regions here (off-by 1 errors only).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
```diff
@@ -0,0 +1,895 @@
...
+    assertEquals(UnicodeBlock.ARABIC_EXTENDED_A, UnicodeBlock.forName("ArabicExtended-A"))
+    assertEquals(UnicodeBlock.ARABIC_EXTENDED_A, UnicodeBlock.forName("ARABICEXTENDED-A"))
+    assertThrows(classOf[IllegalArgumentException], UnicodeBlock.forName("Arabic ExtendedA"))
+    assertThrows(classOf[IllegalArgumentException], UnicodeBlock.forName("Arabic_ExtendedA"))
+    assertThrows(classOf[IllegalArgumentException], UnicodeBlock.forName("ArabicExtendedA"))
+
+    assertEquals(UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS_EXTENSION_A,
+        UnicodeBlock.forName("CJK Unified Ideographs Extension A"))
+    assertThrows(classOf[IllegalArgumentException],
+      UnicodeBlock.forName("CJK Unified Ideographs Extension-A"))
```
gzm0, line 48: Bad indent.

gzm0 (changes requested):

javalanglib/src/main/scala/java/lang/Character.scala
er1c: Updated, thanks for the suggestion

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,362 @@ object Character {
...
+        val mid = lo + (hi - lo) / 2
+        val block = allBlocks.get(mid)
+
+        if (codePoint >= block.start && codePoint <= block.end) {
+          block
+        } else if (codePoint > block.end) {
+          binarySearch(codePoint, mid + 1, hi)
+        } else {
+          binarySearch(codePoint, lo, mid)
+        }
```
gzm0, line 872: Consider:

```scala
if (codePoint < block.start) binarySearch(codePoint, lo, mid)
else if (codePoint > block.end) binarySearch(codePoint, mid + 1, hi)
else block
```

IMHO this is easier to read.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,362 @@ object Character {
...
+    // Additional JVM Aliases
+    blocksByNormalizedName.put("Greek and Coptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GreekandCoptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GREEK_AND_COPTIC".toLowerCase, GREEK)
+    blocksByNormalizedName.put("TaiLe".toLowerCase, TAI_LE)
+    blocksByNormalizedName.put("Combining Diacritical Marks for Symbols".toLowerCase, COMBINING_MARKS_FOR_SYMBOLS)
+    blocksByNormalizedName.put("CombiningDiacriticalMarksforSymbols".toLowerCase, COMBINING_MARKS_FOR_SYMBOLS)
+
+    def forName(blockName: String): UnicodeBlock = {
+      val key: String = blockName.toLowerCase
```
gzm0, line 844: Uh, oh.... `toLowerCase` would need a locale here, but Scala.js doesn't support that in core IIUC :-/

IDK what to do about this.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,362 @@ object Character {
...
+    ////////////////
+    // scalastyle:on line.size.limit
+
+    // Additional JVM Aliases
+    blocksByNormalizedName.put("Greek and Coptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GreekandCoptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GREEK_AND_COPTIC".toLowerCase, GREEK)
+    blocksByNormalizedName.put("TaiLe".toLowerCase, TAI_LE)
+    blocksByNormalizedName.put("Combining Diacritical Marks for Symbols".toLowerCase, COMBINING_MARKS_FOR_SYMBOLS)
+    blocksByNormalizedName.put("CombiningDiacriticalMarksforSymbols".toLowerCase, COMBINING_MARKS_FOR_SYMBOLS)
```
gzm0, line 841: This seems to hard code the logic for alias building. Consider factoring this into a helper.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,362 @@ object Character {
...
+    ////////////////
+    // End Generated
+    ////////////////
+    // scalastyle:on line.size.limit
+
+    // Additional JVM Aliases
+    blocksByNormalizedName.put("Greek and Coptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GreekandCoptic".toLowerCase, GREEK)
+    blocksByNormalizedName.put("GREEK_AND_COPTIC".toLowerCase, GREEK)
+    blocksByNormalizedName.put("TaiLe".toLowerCase, TAI_LE)
```
gzm0, line 839: Why is this needed?

gzm0 (changes requested):

javalanglib/src/main/scala/java/lang/Character.scala
sjrd: I don't think we need a locale. The doc says:

> Finally, character case is ignored for all of the valid block name forms. For example, "BASIC_LATIN" and "basic_latin" are both valid block names. **The en_US locale's** case mapping rules are used to provide case-insensitive string comparisons for block name validation. 

(emphasis mine)

And en_US is our officially default locale, used by `.toLowerCase()`. So I believe we're good.

er1c: Right, since there is no "input" for a locale selection, there would also be no way for the method to be locale-aware.  In a very obscure edge case, perhaps there is a turkish spelling for these block names, but I don't think that's very realistic.  I can try and create this edge case as a test to see if the JVM supports it (e.g. if the JVM default locale is turkish, and if you can specify a block name with the turkish uppercase characters, that would get translated wrong to lowercase) I wouldn't think this is a very realistic blocking issue though.

javalanglib/src/main/scala/java/lang/Character.scala
er1c: 👍

javalanglib/src/main/scala/java/lang/Character.scala
er1c: 👍 its not, will update

er1c:
@sjrd @gzm0 CI passing

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+    final override def equals(that: Any): scala.Boolean = super.equals(that)
+    final override def hashCode: scala.Int = super.hashCode
+    final override def toString(): String = name
+  }
+
+  final class UnicodeBlock private (name: String, private val start: Int, private val end: Int) extends Subset(name)
+
+  object UnicodeBlock {
+    private[this] val allBlocks: ArrayList[UnicodeBlock] = new ArrayList[UnicodeBlock](220)
+    private[this] val blocksByNormalizedName = new java.util.HashMap[String, UnicodeBlock]()
```
sjrd, line 535: Consider importing `java.util.HashMap` like you did for `ArrayList`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+      val lower = properName.toLowerCase
+      //   lowercase + spaces
+      blocksByNormalizedName.put(lower, block)
+      //   lowercase + no spaces
+      blocksByNormalizedName.put(lower.replaceAll("\\s", ""), block)
+      //   tokens as _
+      blocksByNormalizedName.put(lower.replaceAll("[\\s\\-_]", "_"), block)
+    }
+
+    private[this] def addUnicodeBlock(name: String, properName: String, start: Int, end: Int): UnicodeBlock =  {
```
sjrd, line 548: Is the `name` ever different than `properName.toUpperCase().replaceAll("[ \\-]", "_")`? If not, we should derive `name` from `properName` to reduce the amount of generated code.

If yes, perhaps it's only for a few exceptions? In that case we could have an overload of `addUnicodeBlock` that derives `name` from `properName`, and use that as often as possible.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+
+    private[this] def addNameAliases(properName: String, block: UnicodeBlock): Unit = {
+      // Add normalized aliases
+      val lower = properName.toLowerCase
+      //   lowercase + spaces
+      blocksByNormalizedName.put(lower, block)
+      //   lowercase + no spaces
+      blocksByNormalizedName.put(lower.replaceAll("\\s", ""), block)
+      //   tokens as _
+      blocksByNormalizedName.put(lower.replaceAll("[\\s\\-_]", "_"), block)
```
sjrd, line 545: Let's try to get rid of the dependency on regexes, which are pretty heavy:

* `_` in the regex is redundant, since they need not be replaced.
* `\\s` could be replaced by the space `' '` AFAICT. It seems there never any other space characters than the actual space `' '`.
* If we only have the true space and `-` left, we can do `lower.replace(' ', '_').replace('-', '_')`

And if you're concerned about calling `replace` twice, you can store the result of replacing the spaces first, which is necessary two lines above anyway, and then only call the second `replace` on that one. That should be an overall win.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+
+    private[this] def addUnicodeBlock(name: String, properName: String, start: Int, end: Int): UnicodeBlock =  {
+      val block = new UnicodeBlock(name, start, end)
+      allBlocks.add(block)
+      addNameAliases(properName, block)
+      block
+    }
+
+    // Special constants
+    val SURROGATES_AREA = new UnicodeBlock("SURROGATES_AREA",0x0, 0x0)
```
sjrd, line 556:
```suggestion
    val SURROGATES_AREA = new UnicodeBlock("SURROGATES_AREA", 0x0, 0x0)
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+    blocksByNormalizedName.put("SURROGATES AREA".toLowerCase, SURROGATES_AREA)
+    blocksByNormalizedName.put("SURROGATES_AREA".toLowerCase, SURROGATES_AREA)
+
+    val BASIC_LATIN = addUnicodeBlock("BASIC_LATIN", "Basic Latin", 0x0000, 0x007F)
+    val LATIN_1_SUPPLEMENT = addUnicodeBlock("LATIN_1_SUPPLEMENT", "Latin-1 Supplement", 0x0080, 0x00FF)
```
sjrd, line 561: Why are those special? Couldn't they be part of the generic script, if we made it start at 0 rather than 256?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+      val block = new UnicodeBlock(name, start, end)
+      allBlocks.add(block)
+      addNameAliases(properName, block)
+      block
+    }
+
+    // Special constants
+    val SURROGATES_AREA = new UnicodeBlock("SURROGATES_AREA",0x0, 0x0)
+    blocksByNormalizedName.put("SURROGATES AREA".toLowerCase, SURROGATES_AREA)
+    blocksByNormalizedName.put("SURROGATES_AREA".toLowerCase, SURROGATES_AREA)
```
sjrd, line 558: Consider "constant-folding" the `toLowerCase` calls:
```suggestion
    blocksByNormalizedName.put("surrogates area", SURROGATES_AREA)
    blocksByNormalizedName.put("surrogates_area", SURROGATES_AREA)
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+
+    val BASIC_LATIN = addUnicodeBlock("BASIC_LATIN", "Basic Latin", 0x0000, 0x007F)
+    val LATIN_1_SUPPLEMENT = addUnicodeBlock("LATIN_1_SUPPLEMENT", "Latin-1 Supplement", 0x0080, 0x00FF)
+
+    // scalastyle:off line.size.limit
```
sjrd, line 563: I think the script itself should stay within the limit. Consider moving that just before the `Begin generated` comment.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+      source
+        .getLines()
+        .filterNot {
+          _.startsWith("#")
+        }
+        .flatMap { line =>
+          line.split(';') match {
+            case Array(_, name) =>
+              val trimmed = name.trim
+              val jvmName = trimmed.replaceAll(raw"[\s\-_]", "_").toUpperCase
```
sjrd, line 578:
```suggestion
              val jvmName = trimmed.replaceAll(raw"[\s\-]", "_").toUpperCase
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+            case _ => None
+          }
+        }.toMap ++ Seq(
+          "GREEK" -> "Greek",
+          "CYRILLIC_SUPPLEMENTARY" -> "Cyrillic Supplementary",
+          "COMBINING_MARKS_FOR_SYMBOLS" -> "Combining Marks For Symbols"
+        )
+    }
+
+    val blocksAndCharacters = (256 to Character.MAX_CODE_POINT)
```
sjrd, line 589: Start at 0?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+      .map { cp => Character.UnicodeBlock.of(cp) -> cp }.filterNot{ _._1 == null }
```
sjrd, line 590: For inline lambdas, we never use `{}`:
```suggestion
      .map(cp => Character.UnicodeBlock.of(cp) -> cp).filterNot(_._1 == null)
```
Same a few times below (`map`, `groupBy`)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+    val orderedBlocks = blocksAndCharacters.map{ _._1 }.distinct
+
+    val blockLowAndHighCodePointsMap = {
+      blocksAndCharacters.groupBy{ _._1 }.mapValues { v =>
+        val codePoints = v.map{ _._2 }
+        (codePoints.min, codePoints.max)
+      }
+    }
+
+    orderedBlocks.foreach{ b =>
```
sjrd, line 601:
```suggestion
    orderedBlocks.foreach { b =>
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+      val minCodePoint = "%04x".format(blockLowAndHighCodePointsMap(b)._1).toUpperCase
+      val maxCodePoint = "%04x".format(blockLowAndHighCodePointsMap(b)._2).toUpperCase
```
sjrd, line 603: We typically use lowercase letters in hexadecimal numbers, so:
```suggestion
      val minCodePoint = "%04x".format(blockLowAndHighCodePointsMap(b)._1)
      val maxCodePoint = "%04x".format(blockLowAndHighCodePointsMap(b)._2)
```
would produce code that is more consistent with the rest of the file.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+
+      println(s"""  val $b = addUnicodeBlock("$b", "${blockNameMap.getOrElse(b.toString, throw new IllegalArgumentException("$b"))}", 0x$minCodePoint, 0x$maxCodePoint)""")
```
sjrd, line 605: To avoid the long line:
```suggestion
      val properName = blockNameMap.getOrElse(b.toString, throw new IllegalArgumentException("$b"))
      println(s"""  val $b = addUnicodeBlock("$b", "$properName", 0x$minCodePoint, 0x$maxCodePoint)""")
```

javalanglib/src/main/scala/java/lang/Character.scala
sjrd:
```suggestion
      val key: String = blockName.toLowerCase()
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,356 @@ object Character {
...
+    // scalastyle:on line.size.limit
+
+    // Additional JVM Aliases
+    addNameAliases("Greek and Coptic", GREEK)
+    addNameAliases("Combining Diacritical Marks for Symbols", COMBINING_MARKS_FOR_SYMBOLS)
+
+    def forName(blockName: String): UnicodeBlock = {
+      val key: String = blockName.toLowerCase
+      val block = blocksByNormalizedName.get(key)
+      if (block == null) throw new IllegalArgumentException()
```
sjrd, line 844:
```suggestion
      if (block == null)
        throw new IllegalArgumentException()
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
```diff
@@ -0,0 +1,895 @@
...
+
+import java.lang.Character.UnicodeBlock
+import org.junit.Test
+import org.junit.Assert._
+import org.scalajs.testsuite.utils.AssertThrows._
+
+class CharacterUnicodeBlockTest {
+
+  @Test def forName_Normalizations(): Unit = {
+    assertEquals(UnicodeBlock.BASIC_LATIN, UnicodeBlock.forName("Basic Latin"))
```
sjrd, line 23: I think we should test the SURROGATES_AREA as well, since it is pretty special.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
```diff
@@ -0,0 +1,895 @@
...
+    assertThrows(classOf[IllegalArgumentException], UnicodeBlock.forName("ArabicExtendedA"))
+
+    assertEquals(UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS_EXTENSION_A,
+        UnicodeBlock.forName("CJK Unified Ideographs Extension A"))
+    assertThrows(classOf[IllegalArgumentException],
+        UnicodeBlock.forName("CJK Unified Ideographs Extension-A"))
+  }
+
+  @Test def of_Char(): Unit = {
+    assertEquals(UnicodeBlock.BASIC_LATIN, UnicodeBlock.of(0x0.toChar))
```
sjrd, line 52: Were those tests generated to be exhaustive? If yes, we should include the script that was used to generate them as well, since we'll have to redo it one day when we use a more recent version of the JDK/Unicode.

sjrd (changes requested):

javalanglib/src/main/scala/java/lang/Character.scala
er1c: 👍 verified this is true, update incoming shortly

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
er1c: I think I  got them from harmony, I'll look into a code-gen solution

er1c:
@sjrd just waiting on CI, but think everything is resolved

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -12,9 +12,9 @@
 
 package java.lang
 
+import scala.annotation.tailrec
 import scala.scalajs.js
-
```
gzm0, line 16: Leave this newline?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
 
+  // UnicodeBlock
+
+  class Subset protected (name: String) {
+    override final def equals(that: Any): scala.Boolean = super.equals(that)
+    override final def hashCode(): scala.Int = super.hashCode
+    override final def toString(): String = name
+  }
+
+  final class UnicodeBlock private (name: String, private val start: Int, private val end: Int) extends Subset(name)
```
gzm0, line 530: Break this line?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+    private[this] def addNameAliases(properName: String, block: UnicodeBlock,
+        includeJVM: scala.Boolean = true): Unit = {
+      // Add normalized aliases
+      val lower = properName.toLowerCase
+      //   lowercase + spaces
+      blocksByNormalizedName.put(lower, block)
+      //   lowercase + no spaces
+      blocksByNormalizedName.put(lower.replace(" ", ""), block)
+      //   tokens as _
+      if (includeJVM) {
```
gzm0, line 546: Instead of this, consider adding: `blockByNormalizedName.put(jvmName.toLower, block)` to `addUnicodeBlock`.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+              val trimmed = name.trim
+              val jvmName = trimmed.replaceAll(raw"[\s\-]", "_").toUpperCase
+              Some(jvmName -> trimmed)
+            case _ => None
+          }
+        }.toMap ++ Seq(
+          "GREEK" -> "Greek",
+          "CYRILLIC_SUPPLEMENTARY" -> "Cyrillic Supplementary",
+          "COMBINING_MARKS_FOR_SYMBOLS" -> "Combining Marks For Symbols"
+        )
```
gzm0, line 590: This is gonna sound annoying, since we have been back and forth about this for a while: It seems that these are the cases where the JVM name cannot be inferred from the real name of the block. So for me this means:

- Trying to find blocks based on the name is wrong, we should use its boundaries instead.
- Having a separate JVM name argument to `addUnicodeBlock` is OK, since there are genuine cases where these are different.

@sjrd WDYT?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+
+    val blockLowAndHighCodePointsMap = {
+      blocksAndCharacters.groupBy{ _._1 }.mapValues { v =>
+        val codePoints = v.map{ _._2 }
+        (codePoints.min, codePoints.max)
+      }
+    }
+
+    // SURROGATES_AREA
+    val specialCounts = 1
```
gzm0, line 606: I thought `SURROGATES_AREA` is *not* part of `allBlocks, so it doesn't need to go into the count?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
```diff
@@ -0,0 +1,1862 @@
...
+       |  }
+       |
+       |  // scalastyle:off line.size.limit
+       |  @Test def forName_String(): Unit = {
+       |    ${nameTests.mkString("\n    ")}
+       |  }
+       |  // scalastyle:on line.size.limit
+       |""".stripMargin
+
+  println(generated)
```
gzm0, line 1861: Do we really need to test all the cases here? I'm wondering if having a bunch of well chosen test cases wouldn't be enough.

gzm0 (changes requested):

javalanglib/src/main/scala/java/lang/Character.scala
sjrd: If I understand correctly from the JavaDoc, `Greek` and `Combining Marks For Symbols` have a different JVM name (+ alias with spaces) because those blocks *used to* be called like that. They were renamed in Unicode at some point (prior to 4.1.0 apparently; I could not find Blocks.txt for earlier versions than 4.1.0) so the new canonical names are usable as aliases, but the JVM name was kept.

For `Cyrillic Supplementary` it's unclear what the reason is. The canonical name in Unicode is `Cyrillic Supplement`, which is consistent with other `* Supplement` blocks. Regardless of the reason, it is also true that it can be found with the canonical name as well as with the JVM name and the JVM name with spaces.

I believe those are historical exceptions, and that more of those will not happen. The overwhelming rule still seems to be that the JVM name is directly derived from the canonical name. For those three exceptions, I would recommend having an overload of `addUnicodeBlock` that takes an additional `historicalName` and which registers the names differently (all lower-cased, obviously):

* the `properName`
* the `historicalName` (`"Greek"`, `"Cyrillic Supplementary"` or `"Combining Marks for Symbols"`)
* the `historicalName.replace(' ', '_')` (which is the name of the JVM constant)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+    val VARIATION_SELECTORS_SUPPLEMENT = addUnicodeBlock("Variation Selectors Supplement", 0xe0100, 0xe01ef)
+    val SUPPLEMENTARY_PRIVATE_USE_AREA_A = addUnicodeBlock("Supplementary Private Use Area-A", 0xf0000, 0xfffff)
+    val SUPPLEMENTARY_PRIVATE_USE_AREA_B = addUnicodeBlock("Supplementary Private Use Area-B", 0x100000, 0x10ffff)
+
+    // scalastyle:on line.size.limit
+    ////////////////
+    // End Generated
+    ////////////////
+
+    // Additional JVM Aliases
```
sjrd, line 852: These "aliases" are actually the canonical names of those blocks. See my other comment.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+    }
+
+    val blocksAndCharacters = (0 to Character.MAX_CODE_POINT)
+      .map(cp => Character.UnicodeBlock.of(cp) -> cp).filterNot(_._1 == null)
+
+    val orderedBlocks = blocksAndCharacters.map{ _._1 }.distinct
+
+    val blockLowAndHighCodePointsMap = {
+      blocksAndCharacters.groupBy{ _._1 }.mapValues { v =>
+        val codePoints = v.map{ _._2 }
```
sjrd, line 600:
```suggestion
      blocksAndCharacters.groupBy(_._1).mapValues { v =>
        val codePoints = v.map(_._2)
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +519,375 @@ object Character {
...
+          "GREEK" -> "Greek",
+          "CYRILLIC_SUPPLEMENTARY" -> "Cyrillic Supplementary",
+          "COMBINING_MARKS_FOR_SYMBOLS" -> "Combining Marks For Symbols"
+        )
+    }
+
+    val blocksAndCharacters = (0 to Character.MAX_CODE_POINT)
+      .map(cp => Character.UnicodeBlock.of(cp) -> cp).filterNot(_._1 == null)
+
+    val orderedBlocks = blocksAndCharacters.map{ _._1 }.distinct
```
sjrd, line 596:
```suggestion
    val orderedBlocks = blocksAndCharacters.map(_._1).distinct
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterUnicodeBlockTest.scala
er1c: *shrug* the exhaustive tests did help isolate issues during development.  I suspect it also helps next time the unicode blocks code gets updated too.  I'm more biased to leave them in, than to incorrectly cherry-pick.  Verifying the JVM character ranges for each unicode block are correct, is also a bonus.

sjrd: I tend to agree with @gzm0 that many of these exhaustive tests seem overkill. I believe it would be more interesting to have well chosen test cases. Personally I would pick the following for the blocks:

* The very first block
* The very last block
* One block in the middle

And the following for the names:

* The ones already in `forName_Normalizations()`, which are very good
* Exhaustive tests for the 3 blocks with a historical name, including the `assertThrows` ones

This set of tests should test all code paths of the implementation, which is I think more interesting than brute-forcing everything.

(Otherwise, LGTM.)

er1c: PR Updated, for the sake of completeness I have beginning/middle/end of both the Char range and the 2nd surrogate char range

sjrd (approved):
Thanks! LGTM

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -519,6 +520,386 @@ object Character {
...
+        throw new IllegalArgumentException()
+      block
+    }
+
+    def of(c: scala.Char): UnicodeBlock = of(c.toInt)
+
+    def of(codePoint: scala.Int): UnicodeBlock = {
+      if (!Character.isValidCodePoint(codePoint)) {
+        throw new IllegalArgumentException()
+      }
```
gzm0, line 883: Consider dropping the braces here.

gzm0 (approved):

javalanglib/src/main/scala/java/lang/Character.scala
er1c: just pushed this change, everything else the same

sjrd:
Will merge as soon as Jenkins catches up and validates this PR :)

er1c:
> Will merge as soon as Jenkins catches up and validates this PR :)

CI Green, thanks!

PR #4176, opened 2020-09-04
Add java.util.StringTokenizer

er1c:


sjrd:
I think this one may be over the line of things we should include in Scala.js core. If we accept this one, we quickly get to `java.util.Scanner`, and that comes with unsolvable issues like Locales. I think this is better left to specialized external extensions of the JDK lib, similarly to scala-java-locale and scala-java-time.

@gzm0 Any opinion on the matter?

er1c:
@sjrd @gzm0 I can move this to `scala-java-locale` if that's what you prefer (I'm a collaborator there, but it would also be pending @cquiroz's buy-in).  What I liked about this API is that is is pretty clean (and no-locale calls).  I'm currently exploring the unicode database and if it makes sense for `scala-java-locale` to also include that data (and the toTitleCase/toUpperCase(Locale)/etc.  Open to ideas.  

I'd love to get a formal or informal definition of what scala-js core is for my own benefit, and also if you can take a stab in the dark as to what would be the "ideal" external extensions would be to complete the javalib.  I suspect trying to scope that definition also helps flush out where an implementation like this should live.  Has any of that been done?

gzm0:
> @gzm0 Any opinion on the matter?

Before looking at the code in this PR, I agreed with your sentiment. However, it is also clear to me, that we do not have a clean definition of what goes into the core JDK and what doesn't. I believe in terms of scope definition, this is currently the weakest point of the Scala.js project. And here is why:

- Usually, we say, "everything that does not have to go in core, shouldn't". Now for the JDK, this would mean heavy dependence on implementation details (does an implementation happen to use `XYStream`?). Further, we cannot remove a thing some SJS implementation stops relying on.
- It is too annoying for trivial things (like `CharArray{Writer|Reader}` to make a recent example) to be in their own repo. Simply because it doesn't seem to cut it in terms of cost/benefit ratio.

I believe most of our decisions in practice hinge on that second bullet and I think that is reasonable. Concerns that typically come into play are (in decreasing order of importance):

## Maintainer Competences

Does the core team have the expertise to maintain this part of the JDK (review PRs, triage bugs, fix bugs)?

## Maintenance Cost

Is the core team willing to take the burden of maintenance for this part of the JDK.

## Versioning / Release Frequency / Quality

Is the versioning of the core artifacts appropriate for the implementation/API at hand (especially with regards to the quality expectations of the core repo)? Is faster iteration needed?

Coming back to `StringTokenizer`.

For me, it was failing on criteria number 2, until I saw the implementation (which seems not too complicated, so not too high maintenance cost). What this leaves us with is the argument about this leading to `ju.Scanner`. If we commit to the above criteria (and believe we should, including documenting them with examples wherever we can), that IMHO becomes [Slippery Slope][1] argument. Or in other words, we'll have reasonably clear decision criteria why including `StringTokenizer` is OK and why including `Scanner` is not.

As such, we might want to consider accepting this PR.

[1]: https://en.wikipedia.org/wiki/Slippery_slope

sjrd:
That is a very detailed and thoughtful response! And I think it makes sense, indeed. It brings us a bit closer to a guideline on what to take in and what not to, although there's still room for subjective decisions.

So then I agree that we should consider this PR.

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  def nextElement(): Object = nextToken()
+
+  def nextToken(): String = {
+    ensureAvailable()
+
+    if (returnDelims && nextIsDelim) {
+        val ret = String.valueOf(currentChar)
```
er1c, line 77: extra spacing snuck in here

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+  def this(str: String) = this(str, " \t\n\r\f", false)
+  def this(str: String, delim: String) = this(str, delim, false)
+
+  private var position: Int = 0
+  private val length: Int = str.length
+
+  def countTokens(): Int = {
+    var count: Int = 0
+    var inToken: Boolean = false
+    var i: Int = position
```
sjrd, line 30: It's arguable whether to use explicit types for *local* `var`s, but usually we omit them. We sometimes use them when it's not obvious what the type should be, or if the initial value as a more restrictive type (like `None` but we need `Option[Int]`). In this case I think they lead to more clutter than necessary, but I have no strong feeling about it.

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+
+    count
+  }
+
+  def hasMoreElements(): Boolean = hasMoreTokens()
+
+  def hasMoreTokens(): Boolean = {
+    if (position >= length) false
+    else if (returnDelims) true
+    else {
```
sjrd, line 59: This may seem silly sometimes, but our coding style rule of having all the branches and if/else's agree on whether to use braces or not means that this should be formatted as
```suggestion
    if (position >= length) {
      false
    } else if (returnDelims) {
      true
    } else {
```

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+      var hasDelim: Boolean = false
+      var i: Int = position
+      while (i < length && !hasDelim) {
+        if (isDelim(str.charAt(i))) hasDelim = true
```
sjrd, line 63: You can directly `return true` here. Also, since it is in statement position:
```suggestion
        if (isDelim(str.charAt(i)))
          return true
```

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+  def nextToken(): String = {
+    ensureAvailable()
+
+    if (returnDelims && nextIsDelim) {
+      val ret = String.valueOf(currentChar)
+      position += 1
+      ret
+    } else {
+      // Skip consecutive delims
+      while (position < length && nextIsDelim) position += 1
```
sjrd, line 82:
```suggestion
      while (position < length && nextIsDelim)
        position += 1
```

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private def ensureAvailable(): Unit = {
+    if (position >= length)
+      throw new NoSuchElementException()
+  }
+
+  private def nextIsDelim: Boolean = isDelim(currentChar)
+  @inline private def isDelim(ch: Char): Boolean = delim.indexOf(ch, 0) >= 0
+  @inline private def currentChar: Char = str.charAt(position)
```
sjrd, line 104: These functions are only used in `nextToken()`. They should be declared inside it as local defs.

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+      val ret = String.valueOf(currentChar)
+      position += 1
+      ret
+    } else {
+      // Skip consecutive delims
+      while (position < length && nextIsDelim) position += 1
+
+      ensureAvailable()
+
+      val start: Int = position
```
sjrd, line 86: For local `val`s, we very rarely use an explicit type:
```suggestion
      val start = position
```

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+    else if (returnDelims) true
+    else {
+      var hasDelim: Boolean = false
+      var i: Int = position
+      while (i < length && !hasDelim) {
+        if (isDelim(str.charAt(i))) hasDelim = true
+        i += 1
+      }
+
+      hasDelim
```
sjrd, line 67: I don't think this algorithm is correct for an input string that contains no delimiter at all. I believe `hasMoreToken()` will return `false` even when called before the first call to `nextToken()`. I see that there is no test for strings that contain no delimiter.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/StringTokenizerTest.scala
```diff
@@ -0,0 +1,96 @@
...
+    assertThrows(classOf[NoSuchElementException], st.nextElement())
+  }
+
+  @Test def nextToken(): Unit = {
+    val st = new StringTokenizer("This is a test String")
+    assertEquals("This", st.nextElement())
+    assertEquals("is", st.nextElement())
+    assertEquals("a", st.nextElement())
+    assertEquals("test", st.nextElement())
+    assertEquals("String",  st.nextElement())
```
sjrd, line 83: The test method is called `nextToken()` but you're calling `nextElement()` here.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/StringTokenizerTest.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  @Test def nextElement(): Unit = {
+    val st = new StringTokenizer("This is a test String")
+    assertEquals("This", st.nextElement())
+    assertEquals("is", st.nextElement())
+    assertEquals("a", st.nextElement())
+    assertEquals("test", st.nextElement())
+    assertEquals("String",  st.nextElement())
+
+    assertFalse(st.hasMoreTokens())
```
sjrd, line 73:
```suggestion
    assertFalse(st.hasMoreElements())
```
would be more consistent

sjrd (changes requested):
In addition to the comments below, my main concern here is that the test suite leaves of lot of holes open.

Please add tests for:

* delimiters at the beginning and/or end of the string
* consecutive delimiters (of the same character and different character)
  * with both `returnDelims` as `true` and `false`
  * also when the set of delims changes in the middle (like `"a;:b"` with first `nextToken(";")` then `nextToken(":")`, with both returnDelims true and false)
* empty strings as well as non-empty strings that contain *no* delimiter
* test that the default set of delims supports the other `\n\r\f\t`, not just `' '`

Test that `hasMoreTokens()` must not change the internal position, not even to skip over delims, because the delims might change later:
```scala
scala> val st = new StringTokenizer("ab;cd;:", ";")
val st: java.util.StringTokenizer = java.util.StringTokenizer@6705b763

scala> st.nextToken()
val res5: String = ab

scala> st.nextToken()
val res6: String = cd

scala> st.hasMoreTokens()
val res9: Boolean = true

scala> st.nextToken(":")
val res10: String = ;

scala> st.hasMoreTokens()
val res11: Boolean = false
```
Note how at `res10` we still observe the `';'`, which we would not have seen if `hasMoreTokens()` had advanced past it before answering `true`.

In order to test many more scenarios, notably the ones with consecutive tokens, without adding too much test code, consider writing a simple helper that will extract the *full* result of calling `hasMoreTokens()` and `nextToken()` to completion, and compares it to an expected full output:

```scala
def assertTokenizerResult(expected: String*)(tokenizer: StringTokenizer): Unit = {
  for (elem <- expected) {
    assertTrue(s"expected more tokens for $elem", tokenizer.hasMoreTokens())
    assertEquals(elem, tokenizer.nextToken())
  }
  assertFalse(tokenizer.hasMoreTokens())
  assertThrows(classOf[NoSuchElementException], tokenizer.nextToken())
}
```

That way the current test for `nextToken()` becomes just
```scala
assertTokenizerResult("This", "is", "a", "test", "String")(
    new StringTokenizer("This is a test String"))
```
and it becomes much easier to write a bunch of them.

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,106 @@
...
+  returnDelims: Boolean
+) extends java.util.Enumeration[Object] {
+
+  def this(str: String) = this(str, " \t\n\r\f", false)
+  def this(str: String, delim: String) = this(str, delim, false)
+
+  private var position: Int = 0
+  private val length: Int = str.length
+
+  def countTokens(): Int = {
```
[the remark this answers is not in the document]
er1c: which java doc?  I just checked and they seem to match?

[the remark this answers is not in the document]
er1c: Ahh, I mis-read not to use the summary, I'll fix this, thanks!

sjrd:
The `master` branch is again in a workable state. Would you mind rebasing on top of it, please?

er1c:
@sjrd rebased, and I've updated the implementation/added additional testing for the mentioned edge cases (that were, as you pointed out NOT handled).  Thanks!

er1c:
@sjrd woops I completely missed the helper method, please ignore reviewing for now

ekrich:
One other consideration is that we also have Scala Native that can share most of the JDK and some of the maintenance burden. Where the people are the same then you don't get the benefit but you do get the work done in one place can be used in the other.

er1c:
@sjrd this is ready for review, I updated the tests to be more comprehensive and as mentioned in previous comment fixed the functionality issues you identified (which also let me review other edge cases)

er1c:
@sjrd ping

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/StringTokenizerTest.scala
```diff
@@ -0,0 +1,121 @@
...
+        ":", ":", "a", ":", ":", "test", ":", ":", "String", ":", ":") {
+      new StringTokenizer("::This::is::a::test::String::", ":", true)
+    }
+  }
+}
+
+object StringTokenizerTest {
+  private def assertTokenizerResult(expected: String*)(makeTokenizer: => StringTokenizer): Unit = {
+    assertElementResult(expected:_*)(makeTokenizer)
+    assertTokenResult(expected:_*)(makeTokenizer)
```
sjrd, line 97:
```suggestion
    assertElementResult(expected: _*)(makeTokenizer)
    assertTokenResult(expected: _*)(makeTokenizer)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/StringTokenizerTest.scala
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private[this] def assertElementResult(expected: String*)(tokenizer: StringTokenizer): Unit = {
+    assertTokenizerResultImpl(_.countTokens(), _.hasMoreElements(), _.nextElement())(expected:_*)(tokenizer)
```
sjrd, line 101:
```suggestion
    assertTokenizerResultImpl(_.countTokens(), _.hasMoreElements(), _.nextElement())(expected: _*)(tokenizer)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/StringTokenizerTest.scala
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private[this] def assertTokenResult(expected: String*)(tokenizer: StringTokenizer): Unit = {
+    assertTokenizerResultImpl(_.countTokens(), _.hasMoreTokens(), _.nextToken())(expected:_*)(tokenizer)
```
sjrd, line 105:
```suggestion
    assertTokenizerResultImpl(_.countTokens(), _.hasMoreTokens(), _.nextToken())(expected: _*)(tokenizer)
```

sjrd (changes requested):
Only one tiny coding style: we write `foo: _*` with a space between `:` and `_*` when expanding varargs. Otherwise LGTM.

sjrd (approved):
LGTM!

PR #4180, opened 2020-09-07
Port two new internal methods of `ClassTag` from upstream.

sjrd:
This is a port of the changes to `ClassTag` in https://github.com/scala/scala/commit/31d6481b8a1322bd1ecfa6d68c49f6c03ab42f19

Although that commit has not reached the 2.13.x branch, we also apply the changes to the 2.13.x overrides to be future-proof.

---

This fixes the nightly build on 2.12.x. Locally tested with
```
> ++2.12.13-bin-490e3e3
> testSuite2_12/test
```

### DIFF
--- a/scalalib/overrides-2.12/scala/reflect/ClassTag.scala
+++ b/scalalib/overrides-2.12/scala/reflect/ClassTag.scala
@@ -3,6 +3,9 @@ package reflect
 
 import java.lang.{ Class => jClass }
 
+import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


gzm0 (approved):

sjrd:
The naive port caused a code size increase, because our `ClassTag`s are stack-allocated most of the time. I changed the `lazy val`s into `def`s for Scala.js, and added a comment explaining why.

gzm0 (approved):

PR #4181, opened 2020-09-07
Extract a few tests with Arrays of Null and Nothing as Scala 2.x-only.

sjrd:
As of this writing, the tests that were moved *can* be made to work by actively avoiding the `ClassTag`s, but they might break at any point in the future, because the status of `Array[Null]` and `Array[Nothing]` themselves in Scala 3 is up in the air.

### DIFF
--- a/test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/compiler/RuntimeTypesTestScala2.scala
+++ b/test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/compiler/RuntimeTypesTestScala2.scala
@@ -1,0 +1,50 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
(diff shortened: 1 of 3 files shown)


gzm0 (approved):

test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/compiler/RuntimeTypesTestScala2.scala
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Test def scala_Null_Array_Null_should_be_allowed_to_exist_and_be_castable(): Unit = {
+    val arr = Array.fill[Null](5)(null)
+    arr.asInstanceOf[Array[Null]]
+  }
+
+  @Test def scala_Null_Array_Array_Null_too(): Unit = {
+    // Was `val arr = Array.fill[Null](5, 5)(null)` but that crashes on the JVM
+    val arr = new Array[Array[Null]](5)
```
sjrd, line 44: @gzm0 I had to adapt this to make `testSuiteJVM` pass. Previously those tests were in `js/` only. It doesn't change the nature of what was being tested.

gzm0 (approved):

PR #4186, opened 2020-09-09
Add load, list, store, and save to java.util.Properties

ekrich:


javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -28,11 +32,47 @@ class Properties(protected val defaults: Properties)
...
-  // def store(writer: Writer, comments: String): Unit
-  // def store(out: OutputStream, comments: String): Unit
+  def load(reader: Reader): Unit =
+    loadImpl(reader)
+
+  def load(inStream: InputStream): Unit = {
+    loadImpl(new InputStreamReader(inStream, StandardCharsets.ISO_8859_1))
+  }
+
+  @deprecated("", "")
```
ritschwumm, line 42: this could be a little bit more specific :)

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -81,6 +121,272 @@ class Properties(protected val defaults: Properties)
...
+    out.println(listStr)
+    entrySet().scalaOps.foreach { entry => out.println(format(entry)) }
+  }
+
+  def list(out: PrintWriter): Unit = {
+    out.println(listStr)
+    entrySet().scalaOps.foreach { entry => out.println(format(entry)) }
+  }
+
+    private def loadImpl(reader: Reader): Unit = {
```
ritschwumm, line 145: bad indent

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -81,6 +121,272 @@ class Properties(protected val defaults: Properties)
...
+    var line: String      = null
+
+    while ({ line = br.readLine(); line != null }) {
+      var i: Int   = -1
+      var ch: Char = Char.MinValue
+
+      def getNextChar(): Char = {
+        i += 1
+        // avoid out of bounds if value is empty
+        if (i < line.length())
```
ritschwumm, line 162: this format is dangerous in refactorings - i'd put {} around the branches or put them in the same line as the if and else (the latter might frowned upon here)

javalib/src/main/scala/java/util/Properties.scala
ekrich: The Javadocs just say Deprecated so not sure if that should be more specific - It looks like Java added enhanced deprecation in version 9 but I am not seeing any updates to this API.

javalib/src/main/scala/java/util/Properties.scala
ekrich: thanks

javalib/src/main/scala/java/util/Properties.scala
ekrich: I am not sure about this one as it passed the Scala Native review and there may be a couple of more places too. I am also not sure about the Dotty "significant whitespace" feature either. I'll gladly change as needed either way.

ekrich:
One test is failing in 2.11 and I have no idea why.

javalib/src/main/scala/java/util/Properties.scala
sjrd: We typically do not put braces around an if/else when all of its branches are one liners. The current format is good here.

sjrd:
It's not 2.11 that's failing; it's the test running on the JVM, in 2.11, 2.12 and 2.13. Try `testSuiteJVM2_12/test` locally to reproduce the issue.

sjrd:
I know this is going to sound very annoying, given that the code is basically copy-pasted from Scala Native, but to be accepted here it will have to follow [the Scala.js coding style](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md).

ekrich:
No worries on the style - I will look at the failure.

javalib/src/main/scala/java/util/Properties.scala
ekrich: I looked at `java.util.Date` and it has `@Deprecated`. Maybe that is the correct one to use?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,286 @@ class PropertiesTest {
...
+  // ported from Harmony
+
+  @Test def nonStringValues(): Unit = {
+    assumeTrue("Assumed compliant asInstanceOf", hasCompliantAsInstanceOfs)
+    val properties = new Properties
+
+    properties.put("age", new Integer(18))
+    assertNull(properties.getProperty("age"))
+    assertThrows(classOf[ClassCastException],
+                 properties.list(new PrintWriter(new ByteArrayOutputStream())))
```
er1c, line 188: I want to know this too, but based upon my current understanding (https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#indentation) this should be 4 lines indented as a continuation line)

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,286 @@ class PropertiesTest {
...
+    prop1.put("Property B", "b ee#!=:")
+    prop1.put("Property C", "see")
+    val out = storeWriter(prop1, "A Header")
+    val prop2 = loadByteArrayOutputStream(out)
+    assertAll(prop1, prop2)
+  }
+
+  // helper functions
+
+  def storeStream(props: Properties,
```
er1c, line 382: I think these various alignments should simply be 4 spaces vs aligned with first param (e.g. https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#class-declaration and https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#indentation

javalib/src/main/scala/java/util/Properties.scala
ekrich: Ok, I switched it the the Java one as other places in the `javalib` use that.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: Ok, I think I got them all. Thanks.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: These too.

ekrich:
Do you think this these two added functions need to be `@inline`?
https://github.com/scala-js/scala-js/pull/4186/commits/b9e45bf7ab0f53a9d9349eca22bf6361159c333c#diff-c7d99a83e5d9fa3559156697ab557764R310
https://github.com/scala-js/scala-js/pull/4186/commits/b9e45bf7ab0f53a9d9349eca22bf6361159c333c#diff-c7d99a83e5d9fa3559156697ab557764R318

ekrich:
Looking for some additional review after the recommended changes and changes due to formatting.

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -81,6 +120,275 @@ class Properties(protected val defaults: Properties)
...
+
+      def oddBackslash(): Boolean = {
+        val m = trailingBackspace.matcher(line)
+        if (m.find()) {
+          val num = m.end(1) - m.start()
+          val isOdd = num % 2 != 0
+          isOdd
+        } else {
+          false
+        }
```
sjrd, line 206: Depending on `Pattern` and the entire regex machinery just to count the number of trailing backslashes is overkill. This method could be written just as easily without regexes. Something like
```scala
var i = line.length()
while (i > 0 && line.charAt(i - 1) == '\\')
  i -= 1
(line.length() - i) % 2 != 0
```
gosh, it's even shorter.

(I may have overlooked this in Scala Native; I'm more acutely aware of code size and hence dependencies on heavy components inside Scala.js)

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -81,6 +120,275 @@ class Properties(protected val defaults: Properties)
...
+        while (j < 4) {
+          sb.append(line.charAt(i))
+          if (j < 3) {
+            // don't advance past the last char used
+            i += 1
+          }
+          j += 1
+        }
+        val ch = Integer.parseInt(sb.toString(), 16).toChar
+        ch
```
sjrd, line 180: Isn't this equivalent to
```scala
val ch = Integer.parseInt(sb.substring(i, i + 4).toChar
i += 3
ch
```
?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,285 @@ class PropertiesTest {
...
+  }
+
+  def loadReader(in: String): Properties = {
+    val prop = new java.util.Properties()
+    prop.load(new InputStreamReader(new ByteArrayInputStream(
+        in.getBytes(StandardCharsets.UTF_8))))
+    prop
+  }
+
+  // scalastyle:off
```
sjrd, line 425: If I understand correctly, this is to turn off the whitespace-at-EOL check? I would much rather that we *don't* turn it off, and that instead we programmatically create the whitespace where needed.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,285 @@ class PropertiesTest {
...
+  lazy val filestr =
```
sjrd, line 426:
```suggestion
  lazy val filestr = {
```
(and the closing `}` on the other side)

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,285 @@ class PropertiesTest {
...
+       |\   i=i
+       |j=\   j
+       |space=\   c
+       |
+       |dblbackslash=\\
+       |
+       |# jvm does not trim trailing space so no line continuation
+       |trailing = foo, \  
+       |bar
+       |notrailing = baz \\
```
sjrd, line 455: Here is one way to get rid of the trailing whitespace while preserving them, and in a way that makes it more clearly visible:
```suggestion
       |# jvm does not trim trailing space so no line continuation
       |trailing = foo, \${"  "}
       |bar
       |notrailing = baz \\${"  "}
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,285 @@ class PropertiesTest {
...
+  def loadReader(in: String): Properties = {
+    val prop = new java.util.Properties()
+    prop.load(new InputStreamReader(new ByteArrayInputStream(
+        in.getBytes(StandardCharsets.UTF_8))))
+    prop
+  }
+
+  // scalastyle:off
+  lazy val filestr =
+    """|
```
sjrd, line 427: Was this meant to have a `raw"""`?

sjrd (changes requested):
I found a few things that can be improved, but overall this looks very good.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: Yes, this was to avoid the space at the EOL - `scalafmt` ignores it as it is in a multi-line string but since it is invisible it would make more sense that add it as you suggest below.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: This? 
>Every expression that spans more than one line should be enclosed in braces

sjrd: Yes, exactly.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: Overall, this was my attempt to put the properties file in string form to save all the file tests from Scala Native. Bunches of tests are there - it probably isn't the best way but I wanted to reuse what I was given and try to not rethink everything. I even thought about using this way in Scala Native and avoiding the file reading in the tests.

The `String` gets changed to `bytes` so I think the intent is that it is `raw`. Also, the parsing does not interpret `\n` as a new line.

One thing that is in the actual file on Scala Native is the following:
`b\u0020b` which gets read as a key of `b  b`. In the `String` this gets interpreted as `b b` which becomes `b=b` so two tests fail.

In the `String` version I had to put `b\ b` to make it pass because the `\` is interpreted as an escape character so it becomes `b b` which is correct in the key.

There are normal non triple quoted `String`s in the tests with unicode chars but in those the `\n` gets interpreted as newline like in the test `loadReaderHandleSpecialChars()`.

Any insights appreciated. It does work with raw.

javalib/src/main/scala/java/util/Properties.scala
ekrich: This is very slick solution.

javalib/src/main/scala/java/util/Properties.scala
ekrich: I added a test to exercise this code directly to prove that I needed the `radix=16` here in Scala.js. On Native we read a unicode example from the file.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: Ok, I think I resolved the above with a new test.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
ekrich: I used the method and also for the embedded tabs above.

javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -81,6 +120,261 @@ class Properties(protected val defaults: Properties)
...
+    def hexChar(x: Int): Char =
+      if (x > 9) (x - 10 + 'A').toChar
+      else (x + '0').toChar
+
+    Array('\\',
+          'u',
+          hexChar((ch >>> 12) & 15),
+          hexChar((ch >>> 8) & 15),
+          hexChar((ch >>> 4) & 15),
+          hexChar(ch & 15))
```
sjrd, line 378: Somehow I overlooked this the previous time.
```suggestion
    Array(
        '\\',
        'u',
        hexChar((ch >>> 12) & 15),
        hexChar((ch >>> 8) & 15),
        hexChar((ch >>> 4) & 15),
        hexChar(ch & 15)
    )
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesTest.scala
```diff
@@ -173,4 +175,294 @@ class PropertiesTest {
...
+    prop1.put("Property B", "b ee#!=:")
+    prop1.put("Property C", "see")
+    val out = storeWriter(prop1, "A Header")
+    val prop2 = loadByteArrayOutputStream(out)
+    assertAll(prop1, prop2)
+  }
+
+  @Test def checkUnicodeParsing(): Unit = {
+    val is = new ByteArrayInputStream(
+      Array('h','\\', 'u', '0', '0', '2', '0', 'h'))
```
sjrd, line 382:
```suggestion
      Array('h', '\\', 'u', '0', '0', '2', '0', 'h'))
```

sjrd (changes requested):
Alright, LGTM! The CI failure is spurious.

I found two very minor formatting issues. Other than that, could you squash all the commits into one with a clean commit message? After that it's good to go.

javalib/src/main/scala/java/util/Properties.scala
ekrich: Do you have an easy way to do the the squash? I have only done it like once and I am having problems.

gzm0: `git rebase -i`? https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

ekrich: I thought maybe there was something easier.

sjrd (approved):

PR #4193, opened 2020-09-13
Distinguish between local and global inheritable ctors

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -192,63 +192,66 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
-        else globalFunctionDef(field, className, Nil, js.Skip()),
-        globalVar(field, className).prototype := ctorToMimic.prototype
-      )
-    }
 
-    val ctorFunWithGlobals =
-      if (!isJSClass) genJSConstructorFun(tree, initToInline)
-      else genConstructorFunForJSClass(tree)
+    def assignConstructorProperty(ctorVar: js.Tree): js.Tree =
+      genIdentBracketSelect(ctorVar.prototype, "constructor") := ctorVar
```
sjrd, line 197: Consider inlining this at the two call sites. This is only line that is still "factored out", and the factoring out makes it actually harder to understand what's going on below. I had to come back to this definition and mentally inline it at the call sites to make sense of them.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -192,63 +192,66 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
 
-      js.Block(docComment, ctorDef, chainProto, inheritableCtorDef)
+        js.Block(
+            js.DocComment("@constructor"),
+            ctorVar := ctorFun,
+            js.DocComment("@constructor"),
+            globalVarDef("h", className, js.Function(false, Nil, js.Skip())),
+            globalVar("h", className).prototype := superCtor.prototype,
+            ctorVar.prototype := js.New(globalVar("h", className), Nil),
+            assignConstructorProperty(ctorVar),
```
sjrd, line 252:
```suggestion
            assignConstructorProperty(ctorVar)
```

sjrd (changes requested):

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
gzm0: I wasn't sure about this one myself :) Removed.

gzm0:
This passes the CI now and I believe I have addressed all comments.

sjrd (approved):

PR #4198, opened 2020-09-16
Fix #2681: Module Splitting Support

### LINKED ISSUE #2681: Shared modules/progressive apps support

gzm0:


gzm0:
At the moment, the overall diff is identical to https://github.com/scala-js/scala-js/pull/4111

gzm0:
Benchmarks for the single module case (i.e. default config).

speed

This shows us that when module splitting is disabled, we do not regress on linking speed after this PR.

sjrd:
I'll probably have to go over this bits by bits ^^

By any chance, would you have some advice on in which order to read things in the big commit (Add Module Splitting support in the linker)?

gzm0:
> By any chance, would you have some advice on in which order to read things in the big commit (Add Module Splitting support in the linker)?

1. Review `ModuleSet`.
2. Observe (informally, look at code if you need to convince yourself):
    - The top level export trees and module initializers have a (public) `moduleID` parameter.
    - The linker writes to a directory instead of a single file.
3. Review `LinkerFrontendImpl`
4. Review `ModuleSplitter` & `ModuleAnalyzer`.
5. Review the `ModuleAnalyzer` implementations.
6. Review `ModuleContext`, `Emitter.ClassID`, `Emitter`.
7. Review `VarGen`.
8. Review the rest of `emitter` (boring, mostly more `WithGlobal` and more implicits).
9. Review `OutputWriter`.
10. At this point, you understand how the splitting works. Review everything else (in whatever order you want).

sjrd:
Thank you! That will be very helpful. :)

gzm0:
I forgot about the `Analyzer`. I suggest you review it after the emitter.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -127,6 +202,41 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
   }
 
+  private def maybeExport(ident: Ident, tree: Tree, mutable: Boolean)(
+      implicit moduleContext: ModuleContext, pos: Position): WithGlobals[Tree] = {
+    if (moduleContext.public) {
+      WithGlobals(tree)
+    } else {
+      val export = config.moduleKind match {
+        case ModuleKind.NoModule =>
+          throw new AssertionError("non-leaf module in NoModule mode")
```
sjrd, line 212:
```suggestion
          throw new AssertionError("non-public module in NoModule mode")
```
?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -41,50 +40,120 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
+      scope: T, value: Tree, origName: OriginalName = NoOriginalName)(
+      implicit moduleContext: ModuleContext, pos: Position): WithGlobals[Tree] = {
+    val ident = globalVarIdent(field, scope, origName)
+    val varDef = genLet(ident, mutable = true, value)
+
+    if (config.moduleKind == ModuleKind.ESModule && !moduleContext.public) {
+      val setterIdent = globalVarIdent(setterField, scope, origName)
+      val x = Ident("x")
+      val setter = FunctionDef(setterIdent, List(ParamDef(x, rest = false)), {
+          Assign(VarRef(ident), VarRef(x))
```
sjrd, line 99:
```suggestion
        Assign(VarRef(ident), VarRef(x))
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -41,50 +40,120 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
+   *  additional setter. This is used when generating ES modules.
+   */
+  def globallyMutableVarDef[T: Scope](field: String, setterField: String,
+      scope: T, value: Tree, origName: OriginalName = NoOriginalName)(
+      implicit moduleContext: ModuleContext, pos: Position): WithGlobals[Tree] = {
+    val ident = globalVarIdent(field, scope, origName)
+    val varDef = genLet(ident, mutable = true, value)
+
+    if (config.moduleKind == ModuleKind.ESModule && !moduleContext.public) {
+      val setterIdent = globalVarIdent(setterField, scope, origName)
```
sjrd, line 96: I don't think it makes sense to give the original name of the field to the setter. They are separate entities.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -39,17 +40,26 @@ private[emitter] final class KnowledgeGuardian(config: Emitter.Config) {
...
-    val hasInlineableInit = computeHasInlineableInit(linkingUnit)
-    val staticFieldMirrors =
-      computeStaticFieldMirrors(linkingUnit.topLevelExports)
+  def update(moduleSet: ModuleSet): Boolean = {
+    val hasInlineableInit = computeHasInlineableInit(moduleSet)
+    val staticFieldMirrors = computeStaticFieldMirrors(moduleSet)
 
-    var objectClass: LinkedClass = null
+    // Object is optional, because the module splitter might remove everything.
+    var objectClass: Option[LinkedClass] = None
```
sjrd, line 48: Have you considered short-cutting this situation right at the start of `Emitter.emit()`, before calling `emitInternal()`? That way we wouldn't have to deal with this corner case everywhere else in the emitter.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -126,14 +138,17 @@ private[emitter] final class KnowledgeGuardian(config: Emitter.Config) {
...
+      moduleSet: ModuleSet): Map[ClassName, Map[FieldName, List[String]]] = {
     if (config.moduleKind != ModuleKind.NoModule) {
       Map.empty
     } else {
       var result = Map.empty[ClassName, Map[FieldName, List[String]]]
-      for (export <- topLevelExports) {
+      for {
+        module <- moduleSet.modules
+        export <- module.topLevelExports
+      } yield {
```
sjrd, line 149: The `yield` shouldn't be there, I think. This for comprehension is in statement position, so its result is ignored.
```suggestion
      } {
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -631,8 +633,23 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
...
+        case Assign(select: SelectStatic, rhs) =>
+          // Destructure separately, otherwise 2.11 crashes.
+          val SelectStatic(className, item) = select
+
+          val scope = (className, item.name)
+
+          if (needToUseGloballyMutableVarSetter(scope)) {
+            unnest(rhs) { (rhs, env0) =>
+              implicit val env = env0
+              js.Apply(globalVar("u", scope), transformExprNoChar(rhs) :: Nil)
```
sjrd, line 645: Couldn't the static field have a `char` type? I think you need to use `transformExpr` with an expected type:
```suggestion
              js.Apply(globalVar("u", scope), transformExpr(rhs, select.tpe) :: Nil)
```

linker/shared/src/main/scala/org/scalajs/linker/backend/OutputWriter.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    def filesToRemove(seen: Iterable[String], reports: List[Report.Module]): Set[String] =
+      seen.toSet -- reports.flatMap(r => r.jsFileName :: r.sourceMapName.toList)
+
+    for {
+      currentFiles <- outputImpl.listFiles()
+      reports <- Future.traverse(moduleSet.modules)(m =>
+          ioThrottler.throttle(writeModule(m.id)))
+      _ <- Future.traverse(filesToRemove(currentFiles, reports))(f =>
+          ioThrottler.throttle(outputImpl.delete(f)))
```
sjrd, line 49:
```suggestion
      reports <- Future.traverse(moduleSet.modules)(
          m => ioThrottler.throttle(writeModule(m.id)))
      _ <- Future.traverse(filesToRemove(currentFiles, reports))(
          f => ioThrottler.throttle(outputImpl.delete(f)))
```
or
```suggestion
      reports <- Future.traverse(moduleSet.modules) { m =>
        ioThrottler.throttle(writeModule(m.id))
      }
      _ <- Future.traverse(filesToRemove(currentFiles, reports)) { f =>
        ioThrottler.throttle(outputImpl.delete(f))
      }
```

sjrd (changes requested):
I reviewed the steps 1 to 9 so far (include 6b for the `Analyzer`). Overall this looks really excellent! I have very minor comments here and there, but nothing major.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
```diff
@@ -12,19 +12,197 @@
...
+   *   used by the StandardLinker) never returns them.
+   *
+   * So as a legacy mechanism, this is OK-ish. It keeps us from having to build
+   * the infrastructure to parse JSON cross platform.
+   */
+  private val fileFieldRe = """(?m)([,{])\s*"file"\s*:\s*".*"\s*([,}])""".r
+
+  /** Patches the JS file content to have the provided source map link (or none) */
+  private def patchJSFileContent(content: String,
+      sourceMapURI: Option[URI]): String = {
```
sjrd, line 125: Could you elaborate a bit on why those patches are necessary, and what they do exactly?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
```diff
@@ -0,0 +1,155 @@
...
+import org.scalajs.linker.interface.unstable.ReportImpl
+
+/** Information about a linker run. */
+abstract class Report private[interface] {
+  /** Public modules resulting from linking.
+   *
+   *  An incremental linker must return all public modules here, even if some
+   *  were not re-written during this linking run.
+   */
+  def publicModules: Iterable[Report.Module]
```
sjrd, line 28: We should have an known-to-be-immutable data type, here. I think it should be `immutable.Seq[Report.Module]`, if not `List[Report.Module]`.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
```diff
@@ -0,0 +1,155 @@
...
+package org.scalajs.linker.interface
+
+import java.io._
+
+import org.scalajs.ir.ScalaJSVersions
+
+import org.scalajs.linker.interface.unstable.ReportImpl
+
+/** Information about a linker run. */
+abstract class Report private[interface] {
```
sjrd, line 22: When we have a visibility modifier for the constructor, we always write the `()`:
```suggestion
abstract class Report private[interface] () {
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
```diff
@@ -0,0 +1,155 @@
...
+       |  publicModules = [
+       |    ${publicModules.mkString(",\n")}
+       |  ],
+       |)""".stripMargin
+  }
+}
+
+object Report {
+  /** Information about a module produced by the linker. */
+  abstract class Module private[interface] {
```
sjrd, line 41:
```suggestion
  abstract class Module private[interface] () {
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
```diff
@@ -0,0 +1,155 @@
...
+         |  jsFileName    = $jsFileName,
+         |  sourceMapName = $sourceMapName,
+         |  moduleKind    = $moduleKind,
+         |)""".stripMargin
+    }
+  }
+
+  /** Serializes this [[Report]] to a byte array.
+   *
+   *  A report serialized with the exact same linker version is guaranteed to
```
sjrd, line 68:
```suggestion
   *  A report serialized with the exact same linker version is guaranteed to be
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
```diff
@@ -0,0 +1,155 @@
...
+      writeBoolean(module.sourceMapName.isDefined)
+      module.sourceMapName.foreach(writeUTF(_))
+      writeModuleKind(module.moduleKind)
+    }
+
+    private def writeModuleKind(kind: ModuleKind): Unit = {
+      val i = kind match {
+        case ModuleKind.NoModule => 0
+        case ModuleKind.ESModule => 1
+        case ModuleKind.CommonJSModule => 2
```
sjrd, line 114:
```suggestion
        case ModuleKind.NoModule       => 0
        case ModuleKind.ESModule       => 1
        case ModuleKind.CommonJSModule => 2
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/unstable/OutputDirectoryImpl.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.scalajs.linker.interface.OutputDirectory
+
+abstract class OutputDirectoryImpl extends OutputDirectory {
+  final private[interface] def impl: OutputDirectoryImpl = this
+
+  /** Writes to the given file.
+   *
+   *  Writing should only result in a file write if the contents of the file
+   *  actually changed. Further, if the underlying filesystem allows it, the
+   * file should be written atomically.
```
sjrd, line 28:
```suggestion
   *  file should be written atomically.
```

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/unstable/OutputDirectoryImpl.scala
```diff
@@ -0,0 +1,42 @@
...
+   */
+  def writeFull(name: String, buf: ByteBuffer)(
+      implicit ec: ExecutionContext): Future[Unit]
+
+  /** Lists all the files in the directory. */
+  def listFiles()(implicit ec: ExecutionContext): Future[Iterable[String]]
```
sjrd, line 34: This is less important than in `Report`, but I would prefer an immutable data here as well, since it is an asynchronous result.

linker/js/src/main/scala/org/scalajs/linker/NodeOutputDirectory.scala
```diff
@@ -0,0 +1,50 @@
...
+  private final class Impl(directory: String) extends OutputDirectoryImpl {
+    def writeFull(name: String, buf: ByteBuffer)(implicit ec: ExecutionContext): Future[Unit] = {
+      val path = getPath(name)
+
+      val data =
+        if (buf.hasTypedArray()) buf.typedArray().subarray(buf.position(), buf.limit())
+        else ByteBuffer.allocateDirect(buf.remaining()).put(buf).typedArray()
+
+      cbFuture[Unit](NodeFS.writeFile(path, data, _))
+        .map(_ => buf.position(buf.limit()))
```
sjrd, line 39: This is a bit misleading. If `buf` does not have a TypedArray, it will already have been advanced to its `limit()` by `put(buf)`. I suggest doing that synchronously in the "then" branch of the `if (buf.hasTypedArray())` to make both branches behave the same.

linker/jvm/src/main/scala/org/scalajs/linker/PathOutputDirectory.scala
```diff
@@ -0,0 +1,204 @@
...
+          writeAtomic(file, buf)
+      }
+    }
+
+    def listFiles()(implicit ec: ExecutionContext): Future[Iterable[String]] = Future {
+      blocking {
+        val builder = Iterable.newBuilder[String]
+
+        /* Files.list(...) would be simpler, but it requires Java Streams which
+         * are, at the time of this writing, not implemented in Scala.js.
```
sjrd, line 52: This code is in `jvm/`, so it could use Streams if that makes it easier.

linker/jvm/src/main/scala/org/scalajs/linker/PathOutputDirectory.scala
```diff
@@ -0,0 +1,204 @@
...
+
+      val promise = Promise[Boolean]()
+      val readBuf = ByteBuffer.allocate(Math.min(2048, cmpBuf.remaining()))
+
+      def readNext(): Unit = {
+        readBuf.clear()
+        chan.read(readBuf, pos, null, Handler)
+      }
+
+      object Handler extends CompletionHandler[Integer, Null] {
```
sjrd, line 172:
```suggestion
      object Handler extends CompletionHandler[Integer, Unit] {
```
?

sjrd (commented):
Some more review until `linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala`, excluded.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -74,12 +75,14 @@ final class Emitter(config: Emitter.Config) {
...
 
-  def emit(unit: LinkingUnit, logger: Logger): Result = {
-    val WithGlobals(body, globalRefs) = emitInternal(unit, logger)
+  def emit(moduleSet: ModuleSet, logger: Logger): Result = {
+    val WithGlobals(body, globalRefs) = emitInternal(moduleSet, logger)
 
     moduleKind match {
       case ModuleKind.NoModule =>
-        val topLevelVars = unit.topLevelExports.map(_.exportName)
+        val topLevelVars = moduleSet.modules
```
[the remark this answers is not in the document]
gzm0: No, we cannot. We can have 0 or 1 module here. Not sure how to improve this though.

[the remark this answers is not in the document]
sjrd: Perhaps just `assert(moduleSet.modules.size <= 1)` and then get `moduleSet.modules.head`?

It may be less "type-safe" but it would be clear what's happening.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
gzm0: Hmmm... I can, but this is the only place that needs changing IIUC. So I'm not sure short-cutting at the beginning is simpler (or clearer). I do not feel strongly about it though. So if you prefer that, I can do that.

sjrd: I don't really have a strong preference either. We can keep it as is.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
gzm0: I'd rather not commit to the ordering of the modules. Is `immutable.Iterable` OK?

gzm0:
I have addressed the first round of comments. All questions I have not answered in the review thread should be answered in code comments.

linker/shared/src/test/scala/org/scalajs/linker/LinkerTest.scala
```diff
@@ -84,4 +116,41 @@ class LinkerTest {
     (1 to 4).foldLeft(firstRun)((p, _) => callInFailedState(p))
   }
 
+  @Test
+  def testLegacyAPI(): AsyncResult = await {
+    val classDefs = Seq(
+        mainTestClassDef({
+          consoleLog(StringLiteral("Hello world!"))
+        })
+    )
```
sjrd, line 126: `helloWorldClassDefs`?

sjrd (commented):
And that's the end of the review for the big Linker commit.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Report.scala
sjrd: Yes, `immutable.Iterable` works too.

linker/shared/src/main/scala/org/scalajs/linker/MemOutputDirectory.scala
```diff
@@ -0,0 +1,60 @@
...
+import java.io.{ByteArrayOutputStream, IOException}
+import java.nio.ByteBuffer
+
+import org.scalajs.linker.interface.OutputDirectory
+import org.scalajs.linker.interface.unstable.OutputDirectoryImpl
+
+/** OutputDirectory that simply writes to memory. */
+sealed trait MemOutputDirectory extends OutputDirectory {
+  /** Content of the file with `name` or `None` if the file was not written. */
+  def content(name: String): Option[Array[Byte]]
```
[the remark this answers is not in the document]
gzm0: By using the file names in the `Report`. I'm not opposed to exposing a way of listing files (or even immutable access to the underlying `Map`), but I was trying to adhere to "expose the minimum required ".

gzm0:
I forgot to mention: I also rebased on `master`. There was a trivial conflict in `Build.scala` (adjacent, independent lines modified).

linker/shared/src/main/scala/org/scalajs/linker/MemOutputDirectory.scala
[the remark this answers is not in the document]
sjrd: > By using the file names in the `Report`.

Ah yes, that makes sense. Let's keep it minimal, then.

project/Build.scala
```diff
@@ -1732,47 +1749,60 @@ object Build {
...
+          outputDir / ("export-loopback" + ext)
         }
-      }.value,
 
-      jsEnvInput in Test ++=
-        (setModuleLoopbackScript in Test).value.toList.map(Input.ESModule(_)),
+        val setDict = {
+          val dict = report.publicModules.map(m =>
+              s"${m.moduleID}: ${m.moduleID}"
+          ).mkString("{", ",", "}")
```
sjrd, line 1773:
```suggestion
          val dict = report.publicModules.map(
              m => s"${m.moduleID}: ${m.moduleID}"
          ).mkString("{", ",", "}")
```
or even
```suggestion
          val dict = report.publicModules
            .map(m => s"${m.moduleID}: ${m.moduleID}")
            .mkString("{", ",", "}")
```

project/Build.scala
```diff
@@ -1732,47 +1749,60 @@ object Build {
...
+            /* Due to the asynchronous nature of ES module loading, there
+             * exists a theoretical risk for a race condition here. It is
+             * possible that tests will start running and reaching the
+             * ExportsTest before this module is executed. It's quite
+             * unlikely, though, given all the message passing for the com
+             * and all that.
+             */
+            val imports = report.publicModules.map(m =>
+                s"""import * as ${m.moduleID} from "./${escapeJS(m.jsFileName)}";\n"""
+            ).mkString
```
sjrd, line 1789:
```suggestion
            val imports = report.publicModules
              .map(m => s"""import * as ${m.moduleID} from "./${escapeJS(m.jsFileName)}";\n""")
              .mkString
```

project/Build.scala
```diff
@@ -1732,47 +1749,60 @@ object Build {
...
+
+            IO.write(outputFile, imports + setDict)
+            List(Input.ESModule(outputFile.toPath))
+
+          case ModuleKind.CommonJSModule =>
+            val requires = report.publicModules.map(m =>
+                s"""var ${m.moduleID} = require("./${escapeJS(m.jsFileName)}");\n"""
+            ).mkString
```
sjrd, line 1797:
```suggestion
            val requires = report.publicModules
              .map(m => s"""var ${m.moduleID} = require("./${escapeJS(m.jsFileName)}");\n""")
              .mkString
```

sjrd (changes requested):
Reviewed until `ScalaJSPluginInternal.scala` (excluded) in the sbt plugin commit.

Jenkinsfile
```diff
@@ -136,7 +152,14 @@ def Tasks = [
     sbtretry ++$scala testSuiteEx$v/test &&
     sbtretry 'set scalaJSStage in Global := FullOptStage' \
         ++$scala testSuiteEx$v/test &&
-    sbtretry ++$scala testSuite$v/test:doc library$v/test compiler$v/test reversi$v/fastOptJS reversi$v/fullOptJS &&
```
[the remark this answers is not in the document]
gzm0: There is the `settings/legacy-link-tasks` scripted test. Do you also want an invocation in the Jenkinsfile?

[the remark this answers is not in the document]
sjrd: Ah no, that's probably good enough. I didn't get that far yet ^^

gzm0:
@sjrd There is a subtle bug I have introduced when I moved `symbolRequirement` handling to the `Analyzer`. The main reason for this was that it's infeasible to handle optional requirements in the module analyzer (because it needs to make decisions based on method existence). Now in the `Analyzer`, `symbolRequirements` also contain module initializers. As such, `jl.Object` ends up statically depending on module entry points. This is of course total nonsense.

I will attempt to fix the problem by making module initializers first class citizens in the Analyzer. This is unfortunate, but I'm not sure I see another option.

gzm0:
I have added a commit fixing the bug and adding a regression test.

linker/shared/src/test/scala/org/scalajs/linker/MinModuleSplittingTest.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.scalajs.junit.async._
+
+import org.scalajs.linker.interface._
+import org.scalajs.linker.testutils.LinkingUtils._
+import org.scalajs.linker.testutils.TestIRBuilder._
+
+class MinModuleSplittingTest {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  /** Smoke test to ensure modules do note get merged too much. */
```
sjrd, line 33:
```suggestion
  /** Smoke test to ensure modules do not get merged too much. */
```

test-suite-linker/src/main/scala/org/scalajs/bootstrap/TestSuiteLinker.scala
```diff
@@ -32,25 +32,27 @@ object QuickLinker {
...
+  private def writeReport(path: String, report: Report): Future[Unit] = {
+    val int8arr = Report.serialize(report).toTypedArray
+    val uint8arr = new Uint8Array(int8arr.buffer, int8arr.byteOffset, int8arr.byteLength)
+    PromisesFS.writeFile(path, uint8arr).toFuture
+  }
+
+  @JSImport("fs", "promises")
   @js.native
-  private def basename(str: String): String = js.native
+  object PromisesFS extends js.Object {
```
sjrd, line 55:
```suggestion
  private object PromisesFS extends js.Object {
```

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/ExportsTest.scala
```diff
@@ -1478,6 +1472,19 @@ class ExportsTest {
     }
   }
 
+  @Test def top_level_export_field_is_writiable_accross_modules(): Unit = {
```
sjrd, line 1475:
```suggestion
  @Test def top_level_export_field_is_writable_accross_modules(): Unit = {
```

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -122,6 +131,8 @@ private[sbtplugin] object ScalaJSPluginInternal {
         box.ensure(linkerImpl.clearableLinker(config))
       },
 
+      scalaJSLinker in legacyKey := (scalaJSLinker in key).value,
```
sjrd, line 134: This is not right. It will silently break compatibility in ways that are hard to debug. Although it's a bit of a specialized use case, if a user had specified
```scala
scalaJSLinker in (Compile, fastOptJS) := something
```
that setting will now be ignored. IMO we should do the reverse. Set `scalaJSLinker in legacyKey` first, then derive `scalaJSLinker in key := (scalaJSLinker in legacyKey).value`.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -141,6 +152,8 @@ private[sbtplugin] object ScalaJSPluginInternal {
         Tags.Tag(s"uses-scalajs-linker-$projectPart-$configPart-$stagePart")
       },
 
+      usesScalaJSLinkerTag in legacyKey := (usesScalaJSLinkerTag in key).value,
```
sjrd, line 155: This one is fine because it is essentially a read-only key from the users' point of view.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -151,6 +164,8 @@ private[sbtplugin] object ScalaJSPluginInternal {
       scalaJSLinkerConfigFingerprint in key :=
         StandardConfig.fingerprint((scalaJSLinkerConfig in key).value),
 
+      moduleName in legacyKey := (moduleName in key).value,
```
sjrd, line 167: This one is arguably OK because it will still affect the `artifactPath in legacyKey`. Nevertheless, I think we should also derive the new one from the old one, for consistency.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -335,14 +410,21 @@ private[sbtplugin] object ScalaJSPluginInternal {
         ((crossTarget in fullOptJS).value /
             ((moduleName in fullOptJS).value + "-opt.js")),
 
-      scalaJSLinkerConfig in fullOptJS ~= { prevConfig =>
+      scalaJSLinkerConfig in linkJSProd ~= { prevConfig =>
```
sjrd, line 413: This is the most annoying one, as it is a very real use case for people to tweak `scalaJSLinkerConfig in (Compile, fullOptJS)`. We absolutely must not throw away their configuration, *especially* since it also affects the result of the legacy keys themselves.

So we should still have here `in fullOptJS`, and after that add
```scala
scalaJSLinkerConfig in linkJSDev := (scalaJSLinkerConfig in fastOptJS).value,
scalaJSLinkerConfig in linkJSProd := (scalaJSLinkerConfig in fullOptJS).value,
```

sjrd (changes requested):
All right, here is the end of the review, including the latest commit that fixes the module initializer issue.

There is one serious issue in `ScalaJSPluginInternal` (there are a few comments that relate to the same thing), but otherwise it looks very good.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
gzm0: Writing to `scalaJSLinker` is specified to be an unstable API, but fair enough, for consistency with everything else, we might as well do it.

gzm0:
I have added a commit addressing the comments. (I have changed everything to be in terms of the `legacyKey` for consistency).

Regarding the names of the new tasks. I like your suggestion from #4111:

> Perhaps fastLinkJS and fullLinkJS? I'll keep thinking.

It keeps `fast` / `full` but introduces the name `link` which makes sense anyways. Probably better than introducing `dev` / `prod`.

For me, there are two open questions on this PR:
- Shall we rename `linkJSDev`  / `linkJSProd` (`linkJSFast` / `linkJSFull` might also be options, also in view of autocomplete: `fast<tab>` would give the old tasks. then again, visibility will probably handle that).
- Are there enough tests for the legacy integration (notably the one in `Linker`)?

sjrd:
> * Shall we rename `linkJSDev`  / `linkJSProd` (`linkJSFast` / `linkJSFull` might also be options, also in view of autocomplete: `fast<tab>` would give the old tasks. then again, visibility will probably handle that).

I would say that I prefer both `linkJSFast` and `fastLinkJS` over `linkJSDev`. I can't really make up my about which is better between `linkJSFast` and `fastLinkJS`. The latter *sounds* (literally) better to me, but that is very subjective.

> * Are there enough tests for the legacy integration (notably the one in `Linker`)?

For the programmatic API of the `Linker`, I think that's enough, yes. For the legacy sbt API, I would suggest adding a few more customizations in `legacy-link-tasks` that would exercise the other keys you fixed in the last commit (`scalaJSLinkerConfig in (Compile, fullOptJS)` notably).

gzm0:
> The latter sounds (literally) better to me, but that is very subjective.

I'd tend to agree in fact. It speaks easier.

OTOH, `linkJSFast` is closer to how sbt tasks like `testOnly`, `testQuick` or `runMain` are named.

From the 6 possible permutations of `js`, `link` and `fast`, I think the following are acceptable:

- `linkJSFast`
- `fastLinkJS`
- `jsLinkFast`
- `jsFastLink`

`linkFastJS` and `fastJSLink` are bad, because they'd imply the JS is fast.

Personally, I think I prefer `linkJSFast` because of how it aligns with sbt's tasks and it goes from most to least important property of the task left to right (that argument might also work in favor of `jsLinkFast` though).

sjrd:
On the other hand, `fastLinkJS` is more similar with today's `fastOptJS` ...

Ah, naming in computer science!

gzm0:
Let's settle on `fastLinkJS` then. My plan is to squash the commits and then perform the rename. Then re-push, including the commit-by-commit CI. Is that OK with you?

sjrd:
Sounds good. :)

gzm0:
Rebased. Testing marathon is starting.

gzm0:
@sjrd testing is reaching completion. Do you have further comments on this?

sjrd (approved):
Let's ship it! 🚀 ✨

PR #4200, opened 2020-09-16
Fix #4188: Return a js.Promise from js.Promise#then

### LINKED ISSUE #4188: Make js.Promise.then to return js.Promise instead of js.Thenable
Currently, `js.Promise[T].then` is typed as it returns `js.Thenable[T]`.

https://github.com/scala-js/scala-js/blob/50ba45c1ed08dbd0c1e30ab80aed3fc27c722779/library/src/main/scala/scala/scalajs/js/Promise.scala#L49-L55

but I think `then` method of Promises returns Promise (unless prototype is modified).

Changing return type to Promise would allows `js.Promise(...).then(...).catch(...)` which can't be compiled for now.

gzm0:


gzm0:
@sjrd it is unclear to me whether `js.Thenable#then` also needs this change. (I tried reading the spec, I didn't understand it...)

sjrd:
`js.Thenable` is a generic interface that can be implemented by many different libraries, so that one should definitely not be changed.

gzm0:
> js.Thenable is a generic interface that can be implemented by many different libraries, so that one should definitely not be changed.

Right. I didn't think about that :-/

gzm0:
@sjrd do you have further comments on this one?

sjrd (approved):

PR #4202, opened 2020-09-19
Use captures instead of ctor params for anon JS class captures.

sjrd:
The logic in `GenJSExports.ExportedSymbol` has always separated "lifted" JS constructors from normal methods and Scala constructors (the latter are, by construction, never lifted at that point). For lifted constructors, it separates the original parameters from the captures, so that orignal parameters become formal parameters of the JS constructor function, while capture parameters become JS class captures.

However, before this commit, it put all parameters back together as regular parameters if the class was an anonymous JS class. The logic in `genAnonJSClass` reassigned those parameters based on the actual arguments inserted by lambdalift in the `new` invocation. Effectively, this dance treated those parameters as class captures, but with an extra indirection of going through constructor parameters.

In this commit, we remove this special case in `ExportedSymbol`, and treat all lifted JS constructors in the same way. `genAnonJSClass` now passes the actual arguments inserted by lambdalift as the actual values for the JS class captures thus produced.

Overall, the end result is basically the same, modulo some boxing and unboxing which happened with the previous scheme, which the optimizer would take care of. The new translation is however cleaner.

---

(backporting the translation scheme I implemented in dotty)

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -737,11 +737,11 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
      *
      *  @param sym Class to generate the instance of
      *  @param jsSuperClassValue JS class value of the super class
-     *  @param args Arguments to the constructor
+     *  @param args Arguments to the Scala constructor, which map to JS class captures
      *  @param pos Position of the original New tree
      */
     def genAnonJSClassNew(sym: Symbol, jsSuperClassValue: js.Tree,
-        args: List[js.TreeOrJSSpread])(
+        args: List[js.Tree])(
         implicit pos: Position): js.Tree = {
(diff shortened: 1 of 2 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -807,14 +807,27 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+
+      /* The first class capture is always a reference to the super class.
+       * This is enforced by genJSClassCapturesAndConstructor.
+       */
+      def jsSuperClassRef(implicit pos: ir.Position): js.VarRef =
+        jsClassCaptures.head.ref
 
+      /* The `this` reference.
+       * FIXME This could clash with a local variable of the constructor or a JS
+       * class capture. How do we avoid this?
```
sjrd, line 829: This was already a problem before this PR. The new comment is just because I realized the problem now.

gzm0 (approved):

PR #4205, opened 2020-09-23
Fix #4203: Mutate the Matcher in Matcher#region

### LINKED ISSUE #4203: Matcher region not mutating the matcher
Matcher.region() is returning an instance of a new matcher with the given start and end position:

https://github.com/scala-js/scala-js/blob/92737aa24280fddb3e914937e9b02f50d2d01c8f/javalib/src/main/scala/java/util/regex/Matcher.scala#L228-L229

The behavior on the JVM is that the region is updated in-place: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#region(int,%20int)

gzm0:


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/regex/RegexMatcherTest.scala
```diff
@@ -269,43 +269,41 @@ class RegexMatcherTest  {
...
-    assertEquals("Scala", region2to7.group)
-
-    val region5toEnd = matcher0.region(5, matcher0.regionEnd)
-    assertEquals(5, region5toEnd.regionStart)
-    if (!executingInJVM) {
-      assertEquals(19, region5toEnd.regionEnd)
-      assertTrue(region5toEnd.find())
-      assertEquals("Solution", region5toEnd.group)
-    }
+    matcher0.region(0, 3)
```
sjrd, line 272: We should assert that the returned matcher is indeed the same. At least once:
```suggestion
    assertSame(matcher0, matcher0.region(0, 3))
```

sjrd (changes requested):
The commit title is backwards: it describes the issue being fixed, rather than what the commit does to the codebase. It should be something like

    Fix #4203: Mutate the Matcher in Matcher#region

sjrd (approved):

Issue #4206, opened 2020-09-23
java.lang.Character toLowerCase(codePoint: Int) / toUpperCase(codePoint: Int)


er1c:
Missing signatures in `java.lang.Character`:

* `def toUpperCase(codePoint: scala.Int): scala.Int`
* `def toLowerCase(codePoint: scala.Int): scala.Int`

*ECMAScript6* has all of the cool codepoint functions e.g.:

```
console.log(String.fromCodePoint(9731, 9733, 9842, 0x2F804));
// expected output: "☃★♲你"
```

But this isn't supported in *ECMAScript5*

From @sjrd (https://github.com/cquiroz/scala-java-locales/issues/197#issuecomment-697640082):

> > ```
> > * `java.lang.Character.toUpperCase(codePoint: Int)`
> > * `java.lang.Character.toLowerCase(codePoint: Int)`
> > ```
> 
> 
> Those we should be able to easily implement by delegating to the `String` equivalents, after having encoded the code points as UTF-16 code unit(s).

Another possible JVM approach: https://github.com/ekrich/scala-unicode/blob/master/src/main/scala/CaseFoldingTest.scala

## Implementation Notes

[Character.toChars(codePoint: Int): Array[Char]](https://github.com/scala-js/scala-js/blob/c8e45b21de7824230b01a4dcff978ec3464a2f06/javalanglib/src/main/scala/java/lang/Character.scala#L473) to split the codepoint & [String.new(codePoints: Array[Int], offset: Int, count: Int): String](https://github.com/scala-js/scala-js/blob/c8e45b21de7824230b01a4dcff978ec3464a2f06/javalanglib/src/main/scala/java/lang/_String.scala#L396)

[String.fromCodePoint(codePoint: Int): String](https://github.com/scala-js/scala-js/blob/c8e45b21de7824230b01a4dcff978ec3464a2f06/javalanglib/src/main/scala/java/lang/_String.scala#L449)

ekrich:
@sjrd Is it better to delegate to JS or bring your own and match Java 8 for example? The disadvantage of matching Java is that you have the code tables and lookup code to maintain.

er1c:
@sjrd does this approach look (generally) correct:

```
  def toUpperCase(codePoint: scala.Int): scala.Int = {
    val codePoints: Array[Char] = toChars(codePoint)
    val s = new String(codePoints, 0, codePoints.length)
    val upperCaseChars: Array[Char] = s.toUpperCase.toCharArray

[...]

PR #4207, opened 2020-09-23
Add j.l.Character.to{Lower,Upper}Case(codePoint: Int)/toTitleCase

er1c:
Addresses: https://github.com/scala-js/scala-js/issues/4206 & https://github.com/scala-js/scala-js/issues/4208

Also fixes minor edge case in `Character.toUpperCase(ch: Char)`

TODO:
- [x] Add a test character that expands into a multiple-codepoint via Strinig.toUpperCase/toLowerCase 

> Then I'm not sure what we should do with results that span more than one code point. It's not necessarily just `length > 2`; it's also possible that this happens with `length == 2` if the two Chars are not a surrogate pair. We need to carefully check the spec for that.

### Javadoc

	public static int toUpperCase(int codePoint)
    
	Converts the character (Unicode code point) argument to uppercase using case mapping information from the UnicodeData file.
	Note that Character.isUpperCase(Character.toUpperCase(codePoint)) does not always return true for some ranges of characters, particularly those that are symbols or ideographs.
	
	In general, String.toUpperCase() should be used to map characters to uppercase. String case mapping methods have several benefits over Character case mapping methods. String case mapping methods can perform locale-sensitive mappings, context-sensitive mappings, and 1:M character mappings, whereas the Character case mapping methods cannot.
	
	Parameters:
	codePoint - the character (Unicode code point) to be converted.
	Returns:
	the uppercase equivalent of the character, if any; otherwise, the character itself.
	Since:
	1.5
	See Also:
	isUpperCase(int), String.toUpperCase()

### Implementation notes

`Character.isSurrogatePair` can be used to verify length == 2, and only return the codePoint if they are a pair, otherwise the original codePoint is suppose to be returned

### DIFF
--- a/javalanglib/src/main/scala/java/lang/Character.scala
+++ b/javalanglib/src/main/scala/java/lang/Character.scala
@@ -12,7 +12,7 @@
 
 package java.lang
 
-import scala.annotation.tailrec
+import scala.annotation.{tailrec, switch}
 import scala.scalajs.js
 
 import java.util.{ArrayList, Arrays, HashMap}
@@ -507,7 +507,38 @@ object Character {
     }
   }
 
-  //def toTitleCase(c: scala.Char): scala.Char
+  def toTitleCase(ch: scala.Char): scala.Char = toTitleCase(ch.toInt).toChar
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+for (cp <- 0 to Character.MAX_CODE_POINT) {
+  val titleCaseCP: Int = Character.toTitleCase(cp)
+  val upperCaseCP: Int = Character.toUpperCase(cp)
+
+  if (titleCaseCP != upperCaseCP) {
+    println(s"    case ${format(cp)} => ${format(titleCaseCP)}")
+  }
+}
+*/
+  def toTitleCase(codePoint: scala.Int): scala.Int = {
+    (codePoint: @switch) match {
+      case 0x01c4 => 0x01c5
+      case 0x01c5 => 0x01c5
+      case 0x01c6 => 0x01c5
+      case 0x01c7 => 0x01c8
+      case 0x01c8 => 0x01c8
+      case 0x01c9 => 0x01c8
+      case 0x01ca => 0x01cb
+      case 0x01cb => 0x01cb
+      case 0x01cc => 0x01cb
+      case 0x01f1 => 0x01f2
+      case 0x01f2 => 0x01f2
+      case 0x01f3 => 0x01f2
+      case _      => toUpperCase(codePoint)
+    }
+  }
+
   //def getNumericValue(c: scala.Char): Int
 
   /* Misc */
--- a/javalanglib/src/main/scala/java/lang/Character.scala
+++ b/javalanglib/src/main/scala/java/lang/Character.scala
@@ -461,8 +461,52 @@ object Character {
   //def getDirectionality(c: scala.Char): scala.Byte
 
   /* Conversions */
-  def toUpperCase(c: scala.Char): scala.Char = c.toString.toUpperCase().charAt(0)
-  def toLowerCase(c: scala.Char): scala.Char = c.toString.toLowerCase().charAt(0)
+  def toUpperCase(ch: Char): Char = toUpperCase(ch.toInt).toChar
+
+  def toUpperCase(codePoint: scala.Int): scala.Int = {
+    codePoint match {
+      case 0x1fb3 | 0x1fc3 | 0x1ff3 =>
+        (codePoint + 0x0009)
+      case _ if codePoint >= 0x1f80 && codePoint <= 0x1faf =>
+        (codePoint | 0x0008)
+      case _ =>
+        val upperChars = _String.fromCodePoint(codePoint).toUpperCase()
+        upperChars.length match {
+          case 1 =>
+            upperChars.charAt(0).toInt
+          case 2 =>
+            val high = upperChars.charAt(0)
+            val low = upperChars.charAt(1)
+            if (isSurrogatePair(high, low))
+              toCodePoint(high, low)
+            else
+              codePoint
+          case _ =>
+            codePoint
+        }
+    }
+  }
+
+  def toLowerCase(ch: scala.Char): scala.Char = toLowerCase(ch.toInt).toChar
+
+  def toLowerCase(codePoint: scala.Int): scala.Int = {
+    val lowerChars = _String.fromCodePoint(codePoint).toLowerCase()
+
+    lowerChars.length match {
+      case 1 =>
+        lowerChars.charAt(0).toInt
+      case 2 =>
+        val high = lowerChars.charAt(0)
+        val low = lowerChars.charAt(1)
+        if (isSurrogatePair(high, low))
+          toCodePoint(high, low)
+        else
+          codePoint
+      case _ =>
+        codePoint
+    }
+  }
+
   //def toTitleCase(c: scala.Char): scala.Char
   //def getNumericValue(c: scala.Char): Int
 
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
@@ -635,6 +635,195 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
     assertEquals(0xfb17, Character.toUpperCase(0xfb17)) // ﬗ => ﬗ
   }
 
+  @Test def toTitleCase_compare_char_and_codepoint(): Unit = {
+    for (ch <- Character.MIN_VALUE to Character.MAX_VALUE)
+      assertEquals(Character.toTitleCase(ch), Character.toTitleCase(ch.toInt).toChar)
+  }
+
+  @Test def toTitleCase_CodePoint_special_cases(): Unit = {
+    assertEquals(0x1fbc, Character.toTitleCase(0x1fb3))
+    assertEquals(0x1fcc, Character.toTitleCase(0x1fc3))
+    assertEquals(0x1ffc, Character.toTitleCase(0x1ff3))
+
+    assertEquals(0x1f88, Character.toTitleCase(0x1f80))
+    assertEquals(0x1f89, Character.toTitleCase(0x1f81))
+    assertEquals(0x1f8a, Character.toTitleCase(0x1f82))
+    assertEquals(0x1f8b, Character.toTitleCase(0x1f83))
+    assertEquals(0x1f8c, Character.toTitleCase(0x1f84))
+    assertEquals(0x1f8d, Character.toTitleCase(0x1f85))
+    assertEquals(0x1f8e, Character.toTitleCase(0x1f86))
+    assertEquals(0x1f8f, Character.toTitleCase(0x1f87))
+    assertEquals(0x1f88, Character.toTitleCase(0x1f88))
+    assertEquals(0x1f89, Character.toTitleCase(0x1f89))
+    assertEquals(0x1f8a, Character.toTitleCase(0x1f8a))
+    assertEquals(0x1f8b, Character.toTitleCase(0x1f8b))
+    assertEquals(0x1f8c, Character.toTitleCase(0x1f8c))
+    assertEquals(0x1f8d, Character.toTitleCase(0x1f8d))
+    assertEquals(0x1f8e, Character.toTitleCase(0x1f8e))
+    assertEquals(0x1f8f, Character.toTitleCase(0x1f8f))
+    assertEquals(0x1f98, Character.toTitleCase(0x1f90))
+    assertEquals(0x1f99, Character.toTitleCase(0x1f91))
+    assertEquals(0x1f9a, Character.toTitleCase(0x1f92))
+    assertEquals(0x1f9b, Character.toTitleCase(0x1f93))
+    assertEquals(0x1f9c, Character.toTitleCase(0x1f94))
+    assertEquals(0x1f9d, Character.toTitleCase(0x1f95))
+    assertEquals(0x1f9e, Character.toTitleCase(0x1f96))
+    assertEquals(0x1f9f, Character.toTitleCase(0x1f97))
+    assertEquals(0x1f98, Character.toTitleCase(0x1f98))
+    assertEquals(0x1f99, Character.toTitleCase(0x1f99))
+    assertEquals(0x1f9a, Character.toTitleCase(0x1f9a))
+    assertEquals(0x1f9b, Character.toTitleCase(0x1f9b))
+    assertEquals(0x1f9c, Character.toTitleCase(0x1f9c))
+    assertEquals(0x1f9d, Character.toTitleCase(0x1f9d))
+    assertEquals(0x1f9e, Character.toTitleCase(0x1f9e))
+    assertEquals(0x1f9f, Character.toTitleCase(0x1f9f))
+    assertEquals(0x1fa8, Character.toTitleCase(0x1fa0))
+    assertEquals(0x1fa9, Character.toTitleCase(0x1fa1))
+    assertEquals(0x1faa, Character.toTitleCase(0x1fa2))
+    assertEquals(0x1fab, Character.toTitleCase(0x1fa3))
+    assertEquals(0x1fac, Character.toTitleCase(0x1fa4))
+    assertEquals(0x1fad, Character.toTitleCase(0x1fa5))
+    assertEquals(0x1fae, Character.toTitleCase(0x1fa6))
+    assertEquals(0x1faf, Character.toTitleCase(0x1fa7))
+    assertEquals(0x1fa8, Character.toTitleCase(0x1fa8))
+    assertEquals(0x1fa9, Character.toTitleCase(0x1fa9))
+    assertEquals(0x1faa, Character.toTitleCase(0x1faa))
+    assertEquals(0x1fab, Character.toTitleCase(0x1fab))
+    assertEquals(0x1fac, Character.toTitleCase(0x1fac))
+    assertEquals(0x1fad, Character.toTitleCase(0x1fad))
+    assertEquals(0x1fae, Character.toTitleCase(0x1fae))
+    assertEquals(0x1faf, Character.toTitleCase(0x1faf))
+  }
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+for (cp <- 0 to Character.MAX_CODE_POINT) {
+  val cpStr: String = new String(Array(cp), 0, 1)
+  val titleCP: Int = Character.toTitleCase(cp)
+  val titleCPStr: String = new String(Array(titleCP), 0, 1)
+
+  if (cpStr.toUpperCase() != titleCPStr)
+    println(s"    assertEquals(${format(titleCP)}, Character.toTitleCase(${format(cp)})) // $cpStr => $titleCPStr")
+}
+*/
+  @Test def toTitleCase_CodePoint_StringUpperCase_diff_CharacterTitleCase(): Unit = {
+    assertEquals(0x00df, Character.toTitleCase(0x00df)) // ß => ß
+    assertEquals(0x0149, Character.toTitleCase(0x0149)) // ŉ => ŉ
+    assertEquals(0x01c5, Character.toTitleCase(0x01c4)) // Ǆ => ǅ
+    assertEquals(0x01c5, Character.toTitleCase(0x01c5)) // ǅ => ǅ
+    assertEquals(0x01c5, Character.toTitleCase(0x01c6)) // ǆ => ǅ
+    assertEquals(0x01c8, Character.toTitleCase(0x01c7)) // Ǉ => ǈ
+    assertEquals(0x01c8, Character.toTitleCase(0x01c8)) // ǈ => ǈ
+    assertEquals(0x01c8, Character.toTitleCase(0x01c9)) // ǉ => ǈ
+    assertEquals(0x01cb, Character.toTitleCase(0x01ca)) // Ǌ => ǋ
+    assertEquals(0x01cb, Character.toTitleCase(0x01cb)) // ǋ => ǋ
+    assertEquals(0x01cb, Character.toTitleCase(0x01cc)) // ǌ => ǋ
+    assertEquals(0x01f0, Character.toTitleCase(0x01f0)) // ǰ => ǰ
+    assertEquals(0x01f2, Character.toTitleCase(0x01f1)) // Ǳ => ǲ
+    assertEquals(0x01f2, Character.toTitleCase(0x01f2)) // ǲ => ǲ
+    assertEquals(0x01f2, Character.toTitleCase(0x01f3)) // ǳ => ǲ
+    assertEquals(0x0390, Character.toTitleCase(0x0390)) // ΐ => ΐ
+    assertEquals(0x03b0, Character.toTitleCase(0x03b0)) // ΰ => ΰ
+    assertEquals(0x0587, Character.toTitleCase(0x0587)) // և => և
+    assertEquals(0x1e96, Character.toTitleCase(0x1e96)) // ẖ => ẖ
+    assertEquals(0x1e97, Character.toTitleCase(0x1e97)) // ẗ => ẗ
+    assertEquals(0x1e98, Character.toTitleCase(0x1e98)) // ẘ => ẘ
+    assertEquals(0x1e99, Character.toTitleCase(0x1e99)) // ẙ => ẙ
+    assertEquals(0x1e9a, Character.toTitleCase(0x1e9a)) // ẚ => ẚ
+    assertEquals(0x1f50, Character.toTitleCase(0x1f50)) // ὐ => ὐ
+    assertEquals(0x1f52, Character.toTitleCase(0x1f52)) // ὒ => ὒ
+    assertEquals(0x1f54, Character.toTitleCase(0x1f54)) // ὔ => ὔ
+    assertEquals(0x1f56, Character.toTitleCase(0x1f56)) // ὖ => ὖ
+    assertEquals(0x1f88, Character.toTitleCase(0x1f80)) // ᾀ => ᾈ
+    assertEquals(0x1f89, Character.toTitleCase(0x1f81)) // ᾁ => ᾉ
+    assertEquals(0x1f8a, Character.toTitleCase(0x1f82)) // ᾂ => ᾊ
+    assertEquals(0x1f8b, Character.toTitleCase(0x1f83)) // ᾃ => ᾋ
+    assertEquals(0x1f8c, Character.toTitleCase(0x1f84)) // ᾄ => ᾌ
+    assertEquals(0x1f8d, Character.toTitleCase(0x1f85)) // ᾅ => ᾍ
+    assertEquals(0x1f8e, Character.toTitleCase(0x1f86)) // ᾆ => ᾎ
+    assertEquals(0x1f8f, Character.toTitleCase(0x1f87)) // ᾇ => ᾏ
+    assertEquals(0x1f88, Character.toTitleCase(0x1f88)) // ᾈ => ᾈ
+    assertEquals(0x1f89, Character.toTitleCase(0x1f89)) // ᾉ => ᾉ
+    assertEquals(0x1f8a, Character.toTitleCase(0x1f8a)) // ᾊ => ᾊ
+    assertEquals(0x1f8b, Character.toTitleCase(0x1f8b)) // ᾋ => ᾋ
+    assertEquals(0x1f8c, Character.toTitleCase(0x1f8c)) // ᾌ => ᾌ
+    assertEquals(0x1f8d, Character.toTitleCase(0x1f8d)) // ᾍ => ᾍ
+    assertEquals(0x1f8e, Character.toTitleCase(0x1f8e)) // ᾎ => ᾎ
+    assertEquals(0x1f8f, Character.toTitleCase(0x1f8f)) // ᾏ => ᾏ
+    assertEquals(0x1f98, Character.toTitleCase(0x1f90)) // ᾐ => ᾘ
+    assertEquals(0x1f99, Character.toTitleCase(0x1f91)) // ᾑ => ᾙ
+    assertEquals(0x1f9a, Character.toTitleCase(0x1f92)) // ᾒ => ᾚ
+    assertEquals(0x1f9b, Character.toTitleCase(0x1f93)) // ᾓ => ᾛ
+    assertEquals(0x1f9c, Character.toTitleCase(0x1f94)) // ᾔ => ᾜ
+    assertEquals(0x1f9d, Character.toTitleCase(0x1f95)) // ᾕ => ᾝ
+    assertEquals(0x1f9e, Character.toTitleCase(0x1f96)) // ᾖ => ᾞ
+    assertEquals(0x1f9f, Character.toTitleCase(0x1f97)) // ᾗ => ᾟ
+    assertEquals(0x1f98, Character.toTitleCase(0x1f98)) // ᾘ => ᾘ
+    assertEquals(0x1f99, Character.toTitleCase(0x1f99)) // ᾙ => ᾙ
+    assertEquals(0x1f9a, Character.toTitleCase(0x1f9a)) // ᾚ => ᾚ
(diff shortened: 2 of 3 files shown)


er1c:
Head scratching: https://scastie.scala-lang.org/rrBuaGmkRry8cllxYWwDCA

codePoint - Int, codePoint - hex
```
223 - 0x00df - strUpperCase: SS, chrUpperCase: ß
329 - 0x0149 - strUpperCase: ʼN, chrUpperCase: ŉ
496 - 0x01f0 - strUpperCase: J̌, chrUpperCase: ǰ
912 - 0x0390 - strUpperCase: Ϊ́, chrUpperCase: ΐ
944 - 0x03b0 - strUpperCase: Ϋ́, chrUpperCase: ΰ
1415 - 0x0587 - strUpperCase: ԵՒ, chrUpperCase: և
7830 - 0x1e96 - strUpperCase: H̱, chrUpperCase: ẖ
7831 - 0x1e97 - strUpperCase: T̈, chrUpperCase: ẗ
7832 - 0x1e98 - strUpperCase: W̊, chrUpperCase: ẘ
7833 - 0x1e99 - strUpperCase: Y̊, chrUpperCase: ẙ
```

I suspect there just aren't tests in place today to show a current failing Character.toUpperCase(0x00df.toChar) against the JVM

er1c:
Added a new test that fails against master, but works against this pr:

```
  @Test def toUpperCase_Char_MultiCodePoints(): Unit = {
    assertEquals('ΐ', Character.toUpperCase('ΐ'))
    assertEquals('ΰ', Character.toUpperCase('ΰ'))
    assertEquals('ὒ', Character.toUpperCase('ὒ'))
    assertEquals('ὔ', Character.toUpperCase('ὔ'))
    assertEquals('ὖ', Character.toUpperCase('ὖ'))
    assertEquals('ᾷ', Character.toUpperCase('ᾷ'))
    assertEquals('ῇ', Character.toUpperCase('ῇ'))
    assertEquals('ῒ', Character.toUpperCase('ῒ'))
    assertEquals('ΐ', Character.toUpperCase('ΐ'))
    assertEquals('ῗ', Character.toUpperCase('ῗ'))
    assertEquals('ῢ', Character.toUpperCase('ῢ'))
    assertEquals('ΰ', Character.toUpperCase('ΰ'))
    assertEquals('ῧ', Character.toUpperCase('ῧ'))
    assertEquals('ῷ', Character.toUpperCase('ῷ'))
    assertEquals('ﬃ', Character.toUpperCase('ﬃ'))
    assertEquals('ﬄ', Character.toUpperCase('ﬄ'))
  }
```

er1c:
I added `toTitleCase` which is based upon this PR as well

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,9 +462,90 @@ object Character {
...
+  val titleCaseCP: Int = Character.toTitleCase(cp)
+  val upperCaseCP: Int = Character.toUpperCase(cp)
+
+  if (titleCaseCP != upperCaseCP) {
+    println(s"    case ${format(cp)} => ${format(titleCaseCP)}")
+  }
+}
+*/
+  def toTitleCase(codePoint: scala.Int): scala.Int = {
+    codePoint match {
```
er1c, line 532: @sjrd does adding a `@switch` here matter, and a better question is how can I verify this myself?

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,54 @@ object Character {
...
+    val upperChars = _String.fromCodePoint(codePoint).toUpperCase
+
+    if (upperChars.length > 2) {
+      codePoint
+    } else if (upperChars.length == 2) {
+      val high = upperChars.charAt(0)
+      val low = upperChars.charAt(1)
+
+      if (isSurrogatePair(high, low))
+        toCodePoint(upperChars.charAt(0), upperChars.charAt(1))
```
sjrd, line 488: Reuse `high` and `low`?
```suggestion
        toCodePoint(high, low)
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,54 @@ object Character {
...
+  }
+
+  def toLowerCase(ch: scala.Char): scala.Char = {
+    val lower = toLowerCase(ch.toInt)
+    if (lower > scala.Char.MaxValue) ch
+    else lower.toChar
+  }
+
+  def toUpperCase(codePoint: scala.Int): scala.Int = {
+    val upperChars = _String.fromCodePoint(codePoint).toUpperCase
```
sjrd, line 479:
```suggestion
    val upperChars = _String.fromCodePoint(codePoint).toUpperCase()
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,54 @@ object Character {
...
+        toCodePoint(upperChars.charAt(0), upperChars.charAt(1))
+      else
+        codePoint
+    } else {
+      upperChars.charAt(0).toInt
+    }
+  }
+
+  def toLowerCase(codePoint: scala.Int): scala.Int = {
+    val lowerChars = _String.fromCodePoint(codePoint).toLowerCase
```
sjrd, line 497:
```suggestion
    val lowerChars = _String.fromCodePoint(codePoint).toLowerCase()
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,54 @@ object Character {
...
+
+    if (lowerChars.length > 2) {
+      codePoint
+    } else if (lowerChars.length == 2) {
+      val high = lowerChars.charAt(0)
+      val low = lowerChars.charAt(1)
+
+      if (isSurrogatePair(high, low))
+        toCodePoint(lowerChars.charAt(0), lowerChars.charAt(1))
```
sjrd, line 506:
```suggestion
        toCodePoint(high, low)
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,54 @@ object Character {
...
+    if (lower > scala.Char.MaxValue) ch
+    else lower.toChar
+  }
+
+  def toUpperCase(codePoint: scala.Int): scala.Int = {
+    val upperChars = _String.fromCodePoint(codePoint).toUpperCase
+
+    if (upperChars.length > 2) {
+      codePoint
+    } else if (upperChars.length == 2) {
```
sjrd, line 483: Instead of this chain of if/else's, consider using a match:
```scala
upperChars.length match {
  case 1 =>
    upperChars.charAt(0).toInt
  case 2 =>
    ...
  case _ =>
    codePoint
}
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+  @Test def toLowerCase_Char(): Unit = {
+    // Low
+    assertEquals('\n', Character.toLowerCase('\n'))
+    // Ascii
+    assertEquals('0', Character.toLowerCase('0'))
+    assertEquals('a', Character.toLowerCase('a'))
+    assertEquals('a', Character.toLowerCase('A'))
+    assertEquals('f', Character.toLowerCase('F'))
+    assertEquals('z', Character.toLowerCase('Z'))
+    // compat/ucd
```
sjrd, line 395: These comments don't mean anything to me. What do they refer to?

Consider using comments that actually explain why the tests are interesting. I could see a few reasons:

* ASCII repertoire, not a letter (no change)
* ASCII repertoire, lowercase letter (no change)
* ASCII repertoire, uppercase letter
* Somewhere else in the BMP, not a letter (no change)
* Somewhere else in the BMP, lowercase letter (no change)
* Somewhere else in the BMP, titlecase letter
* Somewhere else in the BMP, uppercase letter
* Surrogate Chars (no change)

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+    assertEquals('į', Character.toLowerCase('Į'))
+    // high
+    assertEquals('ａ', Character.toLowerCase('Ａ'))
+    assertEquals('ｚ', Character.toLowerCase('Ｚ'))
+    assertEquals('［', Character.toLowerCase('［'))
+  }
+
+  @Test def toLowerCase_Int(): Unit = {
+    // Low
+    assertEquals(10, Character.toLowerCase(10)) // '\n' => '\n'
```
sjrd, line 412: Please consistently use hexadecimal numbers for code points.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+    // Ascii
+    assertEquals(97, Character.toLowerCase(97)) // a => a
+    assertEquals(97, Character.toLowerCase(65)) // A => a
+    assertEquals(102, Character.toLowerCase(70)) // F => f
+    assertEquals(122, Character.toLowerCase(90)) // Z => z
+    // compat/ucd
```
sjrd, line 418: Same issue with the comments. For `toLowerCase(int)`, there are additional interesting categories:

* Outside the BMP, not a letter (no change)
* Outside the BMP, lowercase (no change)
* (Outside the BMP, titlecase -- not sure that exists)
* Outside the BMP, uppercase

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+    assertEquals(258, Character.toUpperCase(259))
+    assertEquals(302, Character.toUpperCase(303))
+    // high
+    assertEquals(65313, Character.toUpperCase(65345))
+    assertEquals(65338, Character.toUpperCase(65370))
+    assertEquals(65339, Character.toUpperCase(65339))
+  }
+
+/*
+for(cp <- 0 until Character.MAX_CODE_POINT; if cp <= Character.MAX_VALUE) {
```
sjrd, line 522: Here we should iterate over Chars, not over code points. Chars are not code points (they are code *units*):
```suggestion
for (c <- 0.toChar until Character.MAX_VALUE) {
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+  val strUpperCase: String = {
+    val s = new String(Array(cp), 0, 1)
+    s.toUpperCase()
+  }
+
+  if (strUpperCase.codePointCount(0, strUpperCase.length-1) > 1)
```
sjrd, line 528: Here we should also include anything that spans more than 1 Char (1 code unit), since they cannot be represented in the result either:
```suggestion
  if (strUpperCase.length > 1)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+    assertEquals('ῧ', Character.toUpperCase('ῧ'))
+    assertEquals('ῷ', Character.toUpperCase('ῷ'))
+    assertEquals('ﬃ', Character.toUpperCase('ﬃ'))
+    assertEquals('ﬄ', Character.toUpperCase('ﬄ'))
+  }
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+for(cp <- 0 until Character.MAX_CODE_POINT) {
```
sjrd, line 554:
```suggestion
for (cp <- 0 until Character.MAX_CODE_POINT) {
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+  val strUpperCase: String = {
+    val s = new String(Array(cp), 0, 1)
+    s.toUpperCase()
+  }
```
sjrd, line 558: Consider making this shorter:
```suggestion
  val strUpperCase: String = new String(Array(cp), 0, 1).toUpperCase()
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,159 @@ class CharacterTest {
...
+
+  if (strUpperCase.codePointCount(0, strUpperCase.length-1) > 1)
```
sjrd, line 560: There must not be a `-1` here. The second argument to `codePointCount` is interpreted as an end *excluded*:
```suggestion
  if (strUpperCase.codePointCount(0, strUpperCase.length) > 1)
```
(this could actually change the output of the script)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -510,7 +511,42 @@ object Character {
...
+  def toTitleCase(ch: scala.Char): scala.Char = {
+    val title = toTitleCase(ch.toInt)
+    if (title > scala.Char.MaxValue) ch
+    else title.toChar
+  }
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+for(cp <- 0 until Character.MAX_CODE_POINT) {
```
sjrd, line 523:
```suggestion
for (cp <- 0 until Character.MAX_CODE_POINT) {
```

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -510,7 +511,42 @@ object Character {
...
+      case 0x01c7 => 0x01c8
+      case 0x01c8 => 0x01c8
+      case 0x01c9 => 0x01c8
+      case 0x01ca => 0x01cb
+      case 0x01cb => 0x01cb
+      case 0x01cc => 0x01cb
+      case 0x01f1 => 0x01f2
+      case 0x01f2 => 0x01f2
+      case 0x01f3 => 0x01f2
+      case _ => toUpperCase(codePoint)
```
sjrd, line 546:
```suggestion
      case 0x01f3 => 0x01f2
      case _      => toUpperCase(codePoint)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -429,6 +429,50 @@ class CharacterTest {
...
+    if (cp <= Character.MAX_VALUE) {
+      assert(titleCaseCP <= Character.MAX_VALUE)
+      println(s"    assertEquals('${titleCaseCP.toChar}', Character.toTitleCase('${cp.toChar}'))")
+    }
+
+    println(s"    assertEquals(${format(titleCaseCP)}, Character.toTitleCase(${format(cp)}))")
+  }
+}
+*/
+  @Test def toTitleCase(): Unit  = {
```
sjrd, line 449: We should definitely also test cases where the result is the same as `toUpperCase`. Basically you can add all the tests written for `toUpperCase` (including the ones that get "aborted" because the result does not fit in 1 char/1 code point) but for `toTitleCase`.

sjrd (changes requested):
The title of the first commit message is too long. We often use `{xy,zw}` to deal with this:

    Add j.l.Character.to{Lower,Upper}Case(codePoint: Int)

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
er1c: Dumb question, what's BMP?

sjrd: [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane). It's the code points that fit in 1 UTF-16 code unit. That's opposed to the code points `>= 0x10000`, which require a surrogate pair.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
er1c: FYI, this logic got updated based upon additional testing (well the `toUpperCase(Char)` got fixed) but yes I still should add tests here,  I just didn't want you to overlook the implementation change to `def toTitleCase(ch: scala.Char): scala.Char`

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
sjrd: Given the new implementations I suggest above, we should definitely have exhaustive tests for all the special characters in `toUpperCase` and `toTitleCase`.

We could also have 3 generic sanity tests for the overloads with Chars, of the form:
```scala
for (ch <- Character.MIN_VALUE to Character.MAX_VALUE)
  assertEquals(Character.toLowerCase(ch), Character.toLowerCase(ch.toInt).toChar)
```
(with the same for `toUpperCase` and `toTitleCase`). When we run those on the JVM, we make sure that the naive implementations we have for the Char overloads is indeed correct.

sjrd (changes requested):

er1c:
@sjrd PR updated, waiting on CI, thanks for coming up with a more elegant `toUpperCase` I've updated the tests as requested as well

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -434,9 +434,136 @@ class CharacterTest {
...
+    assertEquals('ǲ', Character.toTitleCase('Ǳ'))
+    assertEquals('ǲ', Character.toTitleCase('ǲ'))
+    assertEquals('ǲ', Character.toTitleCase('ǳ'))
+  }
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+def myToUpperCase(cp: Int): Int = {
+  val upperString = new String(Array(cp),0,1).toUpperCase()
```
sjrd, line 470:
```suggestion
  val upperString = new String(Array(cp), 0, 1).toUpperCase()
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -434,9 +434,136 @@ class CharacterTest {
...
+    upperString.charAt(0)
+  else
+    cp
+}
+
+for (cp <- 0 to Character.MAX_CODE_POINT) {
+  if (myToUpperCase(cp) != Character.toTitleCase(cp)) {
+    val cpStr: String  = new String(Array(cp),0,1)
+    val titleCP: Int = Character.toTitleCase(cp)
+    val titleStr: String = new String(Array(titleCP),0,1)
```
sjrd, line 481:
```suggestion
    val cpStr: String  = new String(Array(cp), 0, 1)
    val titleCP: Int = Character.toTitleCase(cp)
    val titleStr: String = new String(Array(titleCP), 0, 1)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -434,9 +434,136 @@ class CharacterTest {
...
+    assertEquals('ǲ', Character.toTitleCase('ǳ'))
+  }
+
+/*
+def format(codePoint: Int): String = "0x%04x".format(codePoint)
+
+def myToUpperCase(cp: Int): Int = {
+  val upperString = new String(Array(cp),0,1).toUpperCase()
+  if (upperString.length() == 1)
+    upperString.charAt(0)
```
sjrd, line 472: Should this be expressed in terms of code points?
```suggestion
  if (upperString.codePointCount(0, upperString.length()) == 1)
    upperString.charPointAt(0)
```
In fact, I don't think I understand what codepoints are selected by this script and why?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -434,9 +434,136 @@ class CharacterTest {
...
+/*
+for (ch <- 0.toChar until Character.MAX_VALUE) {
+  val titleCase: Char = Character.toTitleCase(ch)
+  val upperCase: Char = Character.toUpperCase(ch)
+
+  if (titleCase != upperCase)
+    println(s"    assertEquals('$titleCase', Character.toTitleCase('$ch'))")
+}
+*/
+  @Test def toTitleCase_Char_UpperCase_Differences(): Unit = {
```
sjrd, line 451: I now realized that, since we have the exhaustive loop in `toTitleCase_compare_char_and_codepoint()`, we don't need to test anything else about `toTitleCase(Char)`. All the special cases should be tested in `toTitleCase(Int)` anyway, so this is redundant.

The rationale is: `toTitleCase_compare_char_and_codepoint()` can basically assume that `toTitleCase(Int)` is correct. If that one is correct, and we exhaustively test `toTitleCase(Char)` against `toTitleCase(Int)`, then `toTitleCase(Char)` is correct.

(and the test on the JVM will ensure that the *way* we test `totTitleCase(Char)` against `toTitleCase(Int)` is correct itself)

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,9 +463,206 @@ object Character {
...
+  def toLowerCase(codePoint: scala.Int): scala.Int = {
+    val lowerChars = _String.fromCodePoint(codePoint).toLowerCase()
+
+    lowerChars.length match {
+      case 1 =>
+        lowerChars.charAt(0).toInt
+      case 2 =>
+        val high = lowerChars.charAt(0)
+        val low = lowerChars.charAt(1)
+
```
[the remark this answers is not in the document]
sjrd: This is still pending.

javalanglib/src/main/scala/java/lang/Character.scala
```diff
@@ -462,8 +462,63 @@ object Character {
...
+/*
+for (ch <- 0.toChar until Character.MAX_VALUE) {
+  val strLowerCase: String = ch.toString.toLowerCase()
+  val chrLowerCase: Char = Character.toLowerCase(ch)
+
+  if (strLowerCase.charAt(0) != chrLowerCase)
+    println(s"    case '$ch' => '$chrLowerCase'")
+}
+// No Results
+*/
```
sjrd, line 501: This comment is not necessary anymore. We've established that it is in fact true for the 3 methods: `toUpperCase`, `toLowerCase` and `toTitleCase`. Writing this comment here but not for the other two is therefore misleading. We could also write that kind of comment on the other two instead, but that would be completely redundant with the exhaustive tests, so that's not useful IMO.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -434,9 +434,136 @@ class CharacterTest {
     assertEquals(0xff3b, Character.toLowerCase(0xff3b)) // ［ => ［
   }
 
-  @Test def toUpperCase_compare_char_and_codepoint(): Unit = {
+  @Test def toTitleCase_compare_char_and_codepoint(): Unit = {
     for (ch <- Character.MIN_VALUE to Character.MAX_VALUE)
-      assertEquals(Character.toUpperCase(ch), Character.toUpperCase(ch.toInt).toChar)
+      assertEquals(Character.toTitleCase(ch), Character.toTitleCase(ch.toInt).toChar)
+  }
```
sjrd: There seems to be a rebasing problem here. The second commit *removes* the test for `toUpperCase_compare_char_and_codepoint()`. It should be kept.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,559 @@ class CharacterTest {
     assertFalse(Character.isDigit('\uFBFC'))
   }
 
+  @Test def toLowerCase_compare_char_and_codepoint(): Unit = {
+    for (ch <- Character.MIN_VALUE to Character.MAX_VALUE)
+      assertEquals(Character.toLowerCase(ch), Character.toLowerCase(ch.toInt).toChar)
+  }
+
+  @Test def toLowerCase_Char(): Unit = {
```
sjrd, line 391: Similarly to what I said for `toTitleCase`, this test is basically redundant with `toLowerCase_Int()`, now, so it can be removed.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,432 @@ class CharacterTest {
...
+    assertEquals(0xff5a, Character.toLowerCase(0xff3a)) // Ｚ => ｚ
+    assertEquals(0xff3b, Character.toLowerCase(0xff3b)) // ［ => ［
+  }
+
+  @Test def toUpperCase_compare_char_and_codepoint(): Unit = {
+    for (ch <- Character.MIN_VALUE to Character.MAX_VALUE)
+      assertEquals(Character.toUpperCase(ch), Character.toUpperCase(ch.toInt).toChar)
+  }
+
+  @Test def toUpperCase_Char(): Unit = {
```
sjrd, line 569: And same thing here: this is redundant and can be removed.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
sjrd: This is still pending. I'd still like to see tests broken down by categories that actually make sense. Combining what I wrote at https://github.com/scala-js/scala-js/pull/4207/commits/b9438ca40bccb2035cf31af613c459b559c8ded5#r494129429 and what is three comments above, a full list would be:

* ASCII repertoire, not a letter (no change)
* ASCII repertoire, lowercase letter (no change)
* ASCII repertoire, uppercase letter
* Somewhere else in the BMP, not a letter (no change)
* Somewhere else in the BMP, lowercase letter (no change)
* Somewhere else in the BMP, titlecase letter
* Somewhere else in the BMP, uppercase letter
* A high surrogate and a low surrogate (no change)
* Outside the BMP, not a letter (no change)
* Outside the BMP, lowercase (no change)
* (Outside the BMP, titlecase -- not sure that exists)
* Outside the BMP, uppercase

There can be just one of each of those.

Then the same would go for `toUpperCase(Int)` and `toTitleCase(Int)` (the exact same list of inputs can be reused, with different outputs). With that and the existing `toUpperCase_Int_special` and `toTitleCase_Int_special`, we'll have a good coverage.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,432 @@ class CharacterTest {
...
+  @Test def toUpperCase_Int_special(): Unit = {
+    assertEquals(0x1f88, Character.toUpperCase(0x1f80)) // ᾀ => ᾈ
+    assertEquals(0x1f89, Character.toUpperCase(0x1f81)) // ᾁ => ᾉ
+    assertEquals(0x1f8a, Character.toUpperCase(0x1f82)) // ᾂ => ᾊ
+    assertEquals(0x1f8b, Character.toUpperCase(0x1f83)) // ᾃ => ᾋ
+    assertEquals(0x1f8c, Character.toUpperCase(0x1f84)) // ᾄ => ᾌ
+    assertEquals(0x1f8d, Character.toUpperCase(0x1f85)) // ᾅ => ᾍ
+    assertEquals(0x1f8e, Character.toUpperCase(0x1f86)) // ᾆ => ᾎ
+    assertEquals(0x1f8f, Character.toUpperCase(0x1f87)) // ᾇ => ᾏ
+    assertEquals(0x1f98, Character.toUpperCase(0x1f90)) // ᾐ => ᾘ
```
sjrd, line 644: We should also test `0x1f88..0x1f8f` as *inputs* here, since our algorithm captures those as special case inputs. We need to make sure that our trick with the `| 0x0008` does the right thing for the inputs in that range that are already uppercase.

(same with the other 2 ranges of uppercases)

sjrd (changes requested):
OK, we're definitely making good progress, but I still have a number of suggestions for the tests. These things are tricky.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
er1c: I updated this back to a previous version I had that is hopefully more straightforward to understand what is being tested.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
er1c: Added this static ranger as `toUpperCase_CodePoint_special_cases`

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala
```diff
@@ -383,6 +383,352 @@ class CharacterTest {
...
+} yield {
+  //val strUpperCase: String = new String(Array(cp), 0, 1).toUpperCase()
+  s"    assertEquals(${format(chrUpperCase)}, Character.toUpperCase(${format(cp)}))"
+}
+
+val r = new scala.util.Random(100)
+r.shuffle(res).take(100).foreach(println)
+*/
+  @Test def toUpperCase_Int_MultiCodePoints(): Unit = {
+    // 100 "Random" codepoints of the 1Mil+
```
sjrd, line 629: To be honest, I don't think this is very useful. It would be useful if we implemented the general uppercasing ourselves. But in fact we delegate most of the work to JavaScript's uppercasing.

So what's important to test is:
* Any value that is a special case in our algorithms (:+1: for `toUpperCase_CodePoint_special_cases` and `toUpperCase_CodePoint_StringUpperCase_diff_CharacterUpperCase`)
* A few values to exercise the different code paths in our algorithms. For that I believe the list I gave earlier, with one representative code point in each category, is much more useful than this list of random values. Because that way we exercise each code path. And that should be in `toUpperCase_Int`.

The same comment applies to `toLowerCase_Int`.

Other than that, LGTM (notably, the `toTitleCase` commit with its tests LGTM).

sjrd (changes requested):

sjrd:
When you come back to this PR, would you mind rebasing it on top of the latest master, to get the updated tests for JDK 11? Thanks.

er1c:
@sjrd updated, waiting on CI  results

sjrd:
Thank you. That is excellent.

There's only one rebase problem: the old `toUpperCase_Int_MultiCodePoints` is introduced in the first commit, and removed in the second commit. It should be removed from the first commit.

Otherwise LGTM.

er1c:
Added back the `toUpperCase_Int_MultiCodePoints`

sjrd:
Ach, I meant it should not have been present in *either* commit. Could you please remove it from the first commit, and make sure it doesn't reappear in the second commit.

Sorry for the confusion. I probably wasn't clear enough. I may still not be very clear right now, am I?

er1c:
removed in both :)

sjrd (approved):
LGTM! Thank you for bearing with me. ;)

er1c:
> LGTM! Thank you for bearing with me. ;)

that's my line!  thank you for your patience with me.

er1c:
pushed newline fix for scalastyle check

PR #4209, opened 2020-09-24
Fix #4204: Offset Matcher#start()/end() by the match region

### LINKED ISSUE #4204: Matcher start()/end() give incorrect results
In ScalaJS 1.2.0, trying to capture a match within a region:

```
val m = Pattern.compile("[a-z]+").matcher("a b").region(2,3)
m.lookingAt()
println(m.group())
println(s"${m.start()} ${m.end()}"
```

JVM:
```
b
2 3
```

ScalaJS:
```
b
0 1
```

gzm0:


javalib/src/main/scala/java/util/regex/Matcher.scala
```diff
@@ -186,7 +186,7 @@ final class Matcher private[regex] (
 
   def groupCount(): Int = Matcher.getGroupCount(lastMatch, pattern())
 
-  def start(): Int = ensureLastMatch.index
+  def start(): Int = ensureLastMatch.index + regionStart0
```
sjrd, line 189:
```suggestion
  def start(): Int = ensureLastMatch.index + regionStart()
```
like you do further down for `toMatchResult()`? Or use `regionStart0` also in `toMatchResult`? It's a bit inconsistent, at the moment, which led me to believe that there was an actual difference.

sjrd (approved):
Looks good. I just noted one detail.

PR #4211, opened 2020-09-24
Fix #4210: Throw an IllegalArgumentException in j.u.Date.from(Instant).

### LINKED ISSUE #4210: java.util.Date.from(Instant) throws the wrong kind of exception
Our trivial implementation of `java.util.Date.from(Instant)`, at
https://github.com/scala-js/scala-js/blob/ff31eb53c9f090aeb2310638780ea26df680d4b8/javalib/src/main/scala/java/util/Date.scala#L181-L182
is in fact incorrect. If the `Instant` is out of range for a `java.util.Date`, it will throw the `ArithmeticException` thrown by `Instant.toEpochMilli`, instead of the required `IllegalArgumentException`.

Of course, that behavior is only possible to test if we have a `java.util.Instant` in the codebase.

sjrd:
When a `java.time.Instant` cannot be represented as a `Long` value of milliseconds, its method `toEpochMilli()` throws an `ArithmeticException`. The method `java.util.Date.from(Instant)` must however throw an `IllegalArgumentException`.

We now catch the `ArithmeticException` thrown by `toEpochMilli()` to throw an `IllegalArgumentException` instead.

---

To test that, we first introduce some infrastructure to be able to test methods of the core that rely on classes provided by JDK extensions.


sjrd:
All this infrastructure might seem overkill for two little methods, but my plan is to leverage that to be able to confidently implement more complex methods like `String.to{Lower,Upper,Title}Case(Locale)`.

sjrd:
`testSuiteExJVM` fails on the CI because of https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8074032. I have asked Fabien to upgrade our JDK 8 to get the fix.

project/Build.scala
```diff
@@ -1912,6 +1912,44 @@ object Build {
...
+   * (fortunately rare) methods implemented in the core even though they cannot
+   * link without an additional javalib extension.
+   *
+   * Examples include:
+   *
+   * - java.time.Instant, referred to in java.util.Date
+   *
+   * The dummies are definitely not suited for general use. They work just
+   * enough for our tests of other features to work. As such, they are clearly
+   * not published.
```
gzm0, line 1926: "they must not be published."? or "they are not published"?

javalib-ext-dummies/src/main/scala/java/time/Instant.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  private def checkAndCreate(epochSecond: Long, nano: Int): Instant = {
+    val instant = new Instant(epochSecond, nano)
+    if (instant.isBefore(MIN) || instant.isAfter(MAX))
+      throw new DateTimeException("Instant exceeds minimum or maximum instant")
+    instant
+  }
+
+  def ofEpochSecond(epochSecond: Long): Instant =
+    checkAndCreate(epochSecond, 0)
```
gzm0, line 75: This overload seems unused.

javalib-ext-dummies/src/main/scala/java/time/Instant.scala
```diff
@@ -0,0 +1,88 @@
...
+  final val EPOCH: Instant = new Instant(0L, 0)
+  final val MIN: Instant = new Instant(-31557014167219200L, 0)
+  final val MAX: Instant = new Instant(31556889864403199L, 999999999)
+
+  private def checkAndCreate(epochSecond: Long, nano: Int): Instant = {
+    val instant = new Instant(epochSecond, nano)
+    if (instant.isBefore(MIN) || instant.isAfter(MAX))
+      throw new DateTimeException("Instant exceeds minimum or maximum instant")
+    instant
+  }
```
gzm0, line 72: Do we need this check? It seems our implementation doesn't use it (neither do we test it).

javalib-ext-dummies/src/main/scala/java/time/Instant.scala
```diff
@@ -0,0 +1,88 @@
...
+  override def hashCode(): Int =
+    java.lang.Long.hashCode(epochSecond) ^ java.lang.Integer.hashCode(nano)
+
+  // non compliant, for debugging purposes only
+  override def toString(): String =
+    "Instant(" + epochSecond + ", " + nano + ")"
+}
+
+object Instant {
+  final val EPOCH: Instant = new Instant(0L, 0)
```
gzm0, line 63: This constant seems unused.

javalib-ext-dummies/src/main/scala/java/time/Instant.scala
sjrd: Technically, we don't need it, but if we don't do the check, we could have another implementation that looks like it works, although it would transitively throw a `DateTimeException` when using a compliant `Instant`.

Should I test the behavior instead?

gzm0: > Technically, we don't need it, but if we don't do the check, we could have another implementation that looks like it works, although it would transitively throw a DateTimeException when using a compliant Instant.

I'm not sure I follow. You mean another plausible implementation of `j.u.Date`?

sjrd: Another plausible implementation of something (not specifically `j.u.Date`) that we might implement in the future.

I just find it dangerous *not to throw* something that should be thrown. Because we will never know in the future that we're relying on something that would in fact break. It is more dangerous than not implementing all the positive behaviors, because if we start relying on those, we would immediately notice that something's wrong.

gzm0: Ah, fair enough. Yes. Then maybe we should just add an additional test case checking that it throws.

sjrd: Done.

gzm0 (approved):

Issue #4212, opened 2020-09-24
Compiled JS files permissions are too restrictive


janilcgarcia:
I use Scala 2.13.2 and ScalaJS 1.2.0.

Anytime you compile your JS code on Linux it will create files (.js, .js.map) with permission 0600, which is too restrictive.The behaviour I'd expect to the standard to this would be settings the permissions to 0644 (the .jar and the .class files all use those permissions).

This was a problem as I tried to map the JS files into a docker container with nginx as my development environment (which can't read the files because it's running as www-data:www-data).

From what I read from the code the problem seems to at "scala-js/linker/jvm/src/main/scala/org/scalajs/linker/PathOutputFile.scala", line 38.

`Files.createTempFile` creates a file with permissions 0600. This could be fixed by passing the correct `Set[PosixFilePermission]` attribute to the file or by calling `Files.setPosixFilePermissions` afterwards.

PR #4213, opened 2020-09-25
Fixes compiled file permissions

janilcgarcia:
Fixes issue #4212 .

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Matcher.scala
+++ b/javalib/src/main/scala/java/util/regex/Matcher.scala
@@ -186,7 +186,7 @@ final class Matcher private[regex] (
 
   def groupCount(): Int = Matcher.getGroupCount(lastMatch, pattern())
 
-  def start(): Int = ensureLastMatch.index
+  def start(): Int = ensureLastMatch.index + regionStart()
   def end(): Int = start() + group().length
   def group(): String = ensureLastMatch(0).get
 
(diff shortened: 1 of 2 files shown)


janilcgarcia:
I did sign the CLA, but I don't know how can I ask GitHub to test again. Do I need to remove this PR and create another one?

sjrd:
I reran the CLA job.

sjrd:
Instead of using `Files.setPosixFilePermissions` after the fact, can't we create the files with the right permissions in the first place? It seems that `createTempFile` could accept that through the `FileAttribute`s, through the [`asFileAttribute`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/PosixFilePermissions.html#asFileAttribute-java.util.Set-) adapter.

janilcgarcia:
Yes, it can be passed in `createTempFile`, but the docs on [`createTempFile`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempFile) state that it'll throw an exception if the attributes can't be set atomically when the file is created.
I thought calling `setPosixFilePermissions` separately would potentially cover systems that use POSIX permissions but can't set them atomically when creating a temp file.

If this is unnecessary, I'll just do the other way.

----

As for the guidelines, I'm sorry. I'll correct my pull request/code style according to them and create a new PR from the correct branch.

PR #4216, opened 2020-09-28
RFC Hash method defs in-place while serializing

gzm0:


gzm0:
@sjrd I have been playing around with this. Do you think this is an idea worth exploring / completing?

sjrd:
I don't know. I'm not sure I want the serializer to become "smart" on its own. On what account should we require serialization+deserialization to obtain hashed stuff? And what if I want to serialize stuff without hashing them (which is the only thing that we can do on JS since we don't have Digest support)?

IMO this puts two unrelated concerns in the same place.

gzm0:
Closing this. After the discussions in #4346 and the new usage in #4349, I agree we shouldn't do this.

PR #4217, opened 2020-09-29
Add java.util.BitSet

er1c:
This is a fairly heavily modified port of the harmony implementation.

* Missing methods (and tests) added
* Tests cleaned up (at least  in comparison)
* Reformatted methods/code to match javadoc ordering
* etc

10/25/2020:

This has been updated to use an `Array[Int]` as the underlying implementation instead of `Array[Long]`

The only JVM-breaking tests relate to `.size()` which will return the 32-bit based size instead of the 64-bit based size.  This shouldn't be a breaking change, since the API doesn't necessarily guarantee it has to be Long based.

### DIFF
--- a/javalib/src/main/scala/java/util/BitSet.scala
+++ b/javalib/src/main/scala/java/util/BitSet.scala
@@ -1,0 +1,690 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package java.util
+
+import java.io.Serializable
+import java.lang.Integer.bitCount
+import java.lang.Integer.toUnsignedLong
+import java.nio.{ByteBuffer, LongBuffer}
+import java.util
+import java.util.ScalaOps.IntScalaOps
+
+private object BitSet {
+  private final val AddressBitsPerWord = 5 // Int Based 2^5 = 32
+  private final val ElementSize = 1 << AddressBitsPerWord
+  private final val RightBits = ElementSize - 1
+
+  def valueOf(longs: Array[Long]): util.BitSet = {
+    val bs = new util.BitSet
+
+    for (i <- 0 until longs.length * 64) {
+      val idx = i / 64
+      if ((longs(idx) & (1L << (i % 64))) != 0)
+        bs.set(i)
+    }
+
+    bs
+  }
+
+  def valueOf(lb: LongBuffer): BitSet = {
+    val arr = new Array[Long](lb.remaining())
+    lb.get(arr)
+    lb.position(lb.position() - arr.length) // Restores the buffer position
+    valueOf(arr)
+  }
+
+  def valueOf(bytes: Array[Byte]): BitSet = {
+    val bs = new BitSet
+
+    for (i <- 0 until bytes.length * 8) {
+      val idx = i / 8
+      if ((bytes(idx) & (1 << (i % 8))) != 0)
+        bs.set(i)
+    }
+
+    bs
+  }
+
+  def valueOf(bb: ByteBuffer): BitSet = {
+    val arr = new Array[Byte](bb.remaining())
+    bb.get(arr)
+    bb.position(bb.position() - arr.length) // Restores the buffer position
+    valueOf(arr)
+  }
+}
+
+class BitSet private (private var bits: Array[Int]) extends Serializable with Cloneable {
+  import BitSet.{AddressBitsPerWord, ElementSize, RightBits}
+
+  def this(nbits: Int) = {
+    this(
+      bits = {
+        if (nbits < 0)
+          throw new NegativeArraySizeException
+
+        val length = (nbits + BitSet.RightBits) >> BitSet.AddressBitsPerWord
+
+        new Array[Int](length)
+      }
+    )
+  }
+
+  def this() = {
+    this(64)
+  }
+
+  def toByteArray(): Array[Byte] = {
+    if (isEmpty()) {
+      new Array[Byte](0)
+    } else {
+      val l = (length() + 7) / 8
+      val array = new Array[Byte](l)
+
+      for (i <- 0 until length()) {
+        if (get(i))
+          array(i / 8) = (array(i / 8) | (1 << (i % 8))).toByte
+      }
+
+      array
+    }
+  }
+
+  def toLongArray(): Array[Long] = {
+    if (isEmpty()) {
+      new Array[Long](0)
+    } else {
+      val l = (length() + 63) / 64
+      val array = new Array[Long](l)
+
+      for (i <- 0 until length()) {
+        if (get(i))
+          array(i / 64) |= 1L << (i % 64)
+      }
+
+      array
+    }
+  }
+
+  def flip(bitIndex: Int): Unit = {
+    checkBitIndex(bitIndex)
+
+    val len = (bitIndex >> AddressBitsPerWord) + 1
+    ensureLength(len)
+
+    bits(len - 1) ^= 1 << (bitIndex & RightBits)
+  }
+
+  def flip(fromIndex: Int, toIndex: Int): Unit = {
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    if (fromIndex != toIndex) {
+      val len2 = ((toIndex - 1) >> AddressBitsPerWord) + 1
+      ensureLength(len2)
+      val idx1 = fromIndex >> AddressBitsPerWord
+      val idx2 = (toIndex - 1) >> AddressBitsPerWord
+      val mask1 = (~0) << (fromIndex & RightBits)
+      val mask2 = (~0) >>> (ElementSize - (toIndex & RightBits))
+
+      if (idx1 == idx2) {
+        bits(idx1) ^= (mask1 & mask2)
+      } else {
+        bits(idx1) ^= mask1
+        bits(idx2) ^= mask2
+        for (i <- idx1 + 1 until idx2)
+          bits(i) ^= (~0)
+      }
+    }
+  }
+
+  def set(bitIndex: Int): Unit = {
+    checkBitIndex(bitIndex)
+
+    val len = (bitIndex >> AddressBitsPerWord) + 1
+    ensureLength(len)
+
+    bits(len - 1) |= 1 << (bitIndex & RightBits)
+  }
+
+  def set(bitIndex: Int, value: Boolean): Unit =
+    if (value) set(bitIndex)
+    else clear(bitIndex)
+
+  // fromIndex is inclusive, toIndex is exclusive
+  def set(fromIndex: Int, toIndex: Int): Unit = {
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    if (fromIndex != toIndex) {
+      val len2 = ((toIndex - 1) >> AddressBitsPerWord) + 1
+      ensureLength(len2)
+
+      val idx1 = fromIndex >> AddressBitsPerWord
+      val idx2 = (toIndex - 1) >> AddressBitsPerWord
+      val mask1 = (~0) << (fromIndex & RightBits)
+      val mask2 = (~0) >>> (ElementSize - (toIndex & RightBits))
+
+      if (idx1 == idx2) {
+        bits(idx1) |= (mask1 & mask2)
+      } else {
+        bits(idx1) |= mask1
+        bits(idx2) |= mask2
+
+        for (i <- idx1 + 1 until idx2)
+          bits(i) |= (~0)
+      }
+    }
+  }
+
+  def set(fromIndex: Int, toIndex: Int, value: Boolean): Unit =
+    if (value) set(fromIndex, toIndex)
+    else clear(fromIndex, toIndex)
+
+  def clear(bitIndex: Int): Unit = {
+    checkBitIndex(bitIndex)
+
+    val arrayPos = bitIndex >> AddressBitsPerWord
+
+    if (arrayPos < bits.length) {
+      bits(arrayPos) &= ~(1 << (bitIndex & RightBits))
+    }
+  }
+
+  def clear(fromIndex: Int, toIndex: Int): Unit = {
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    val last = bits.length << AddressBitsPerWord
+    if (fromIndex >= last || fromIndex == toIndex)
+      return // scalastyle:ignore
+
+    val toIndexOrLast =
+      if (toIndex > last) last
+      else toIndex
+
+    val idx1 = fromIndex >> AddressBitsPerWord
+    val idx2 = (toIndexOrLast - 1) >> AddressBitsPerWord
+    val mask1 = (~0) << (fromIndex & RightBits)
+    val mask2 = (~0) >>> (ElementSize - (toIndexOrLast & RightBits))
+
+    if (idx1 == idx2) {
+      bits(idx1) &= ~(mask1 & mask2)
+    } else {
+      bits(idx1) &= ~mask1
+      bits(idx2) &= ~mask2
+
+      for (i <- idx1 + 1 until idx2)
+        bits(i) = 0
+    }
+  }
+
+  def clear(): Unit = {
+    for (i <- 0 until bits.length)
+      bits(i) = 0
+  }
+
+  def get(bitIndex: Int): Boolean = {
+    checkBitIndex(bitIndex)
+
+    val arrayPos = bitIndex >> AddressBitsPerWord
+
+    if (arrayPos < bits.length)
+      (bits(arrayPos) & (1 << (bitIndex & RightBits))) != 0
+    else
+      false
+  }
+
+  def get(fromIndex: Int, toIndex: Int): BitSet = {
+    // scalastyle:off return
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    val last = bits.length << AddressBitsPerWord
+    if (fromIndex >= last || fromIndex == toIndex)
+      return new BitSet(0)
+
+    val toIndexOrLast =
+      if (toIndex > last) last
+      else toIndex
+
+    val idx1 = fromIndex >> AddressBitsPerWord
+    val idx2 = (toIndexOrLast - 1) >> AddressBitsPerWord
+    val mask1 = (~0) << (fromIndex & RightBits)
+    val mask2 = (~0) >>> (ElementSize - (toIndexOrLast & RightBits))
+
+    if (idx1 == idx2) {
+      val result = (bits(idx1) & (mask1 & mask2)) >>> (fromIndex % ElementSize)
+      if (result == 0)
+        return new BitSet(0)
+
+      new BitSet(Array[Int](result))
+    } else {
+      val newbits = new Array[Int](idx2 - idx1 + 1)
+      // first fill in the first and last indexes in the new bitset
+      newbits(0) = bits(idx1) & mask1
+      newbits(newbits.length - 1) = bits(idx2) & mask2
+      // fill in the in between elements of the new bitset
+      for (i <- 1 until idx2 - idx1)
+        newbits(i) = bits(idx1 + i)
+
+      val numBitsToShift = fromIndex & RightBits
+
+      if (numBitsToShift != 0) {
+        for (i <- 0 until newbits.length) {
+          // shift the current element to the right
+          newbits(i) = newbits(i) >>> numBitsToShift
+          // apply the last x bits of newbits[i+1] to the current
+          // element
+          if (i != newbits.length - 1)
+            newbits(i) |= newbits(i + 1) << (ElementSize - numBitsToShift)
+        }
+      }
+
+      new BitSet(newbits)
+    }
+    // scalastyle:on return
+  }
+
+  def nextSetBit(fromIndex: Int): Int = {
+    // scalastyle:off return
+    checkFromIndex(fromIndex)
+
+    if (fromIndex >= (bits.length << AddressBitsPerWord))
+      return -1
+
+    var idx = fromIndex >> AddressBitsPerWord
+
+    // first check in the same bit set element
+    if (bits(idx) != 0) {
+      var j = fromIndex & RightBits
+      while (j < ElementSize) {
+        if ((bits(idx) & (1 << j)) != 0)
+          return (idx << AddressBitsPerWord) + j
+        j += 1
+      }
+    }
+
+    idx += 1
+
+    while (idx < bits.length && bits(idx) == 0)
+      idx += 1
+
+    if (idx == bits.length)
+      return -1
+
+    // we know for sure there is a bit set to true in this element
+    // since the bitset value is not 0
+    var j = 0
+    while (j < ElementSize) {
+      if ((bits(idx) & (1 << j)) != 0)
+        return (idx << AddressBitsPerWord) + j
+      j += 1
+    }
+
+    -1
+    // scalastyle:on return
+  }
+
+  def nextClearBit(fromIndex: Int): Int = {
+    // scalastyle:off return
+    checkFromIndex(fromIndex)
+
+    val length = bits.length
+    val bssize = length << AddressBitsPerWord
+
+    if (fromIndex >= bssize)
+      return fromIndex
+
+    var idx = fromIndex >> AddressBitsPerWord
+
+    if (bits(idx) != (~0)) {
+      var j = fromIndex % ElementSize
+      while (j < ElementSize) {
+        if ((bits(idx) & (1 << j)) == 0)
+          return idx * ElementSize + j
+        j += 1
+      }
+    }
+
+    idx += 1
+
+    while (idx < length && bits(idx) == (~0))
+      idx += 1
+
+    if (idx == length)
+      return bssize
+
+    var j = 0
+    while (j < ElementSize) {
+      if ((bits(idx) & (1 << j)) == 0)
+        return (idx << AddressBitsPerWord) + j
+      j += 1
+    }
+
+    bssize
+    // scalastyle:on return
+  }
+
+  def previousSetBit(fromIndex: Int): Int = {
+    // scalastyle:off return
+    if (fromIndex == -1)
+      return -1
+
+    checkFromIndex(fromIndex)
+
+    val bssize = bits.length << AddressBitsPerWord
+    var idx = Math.min(bits.length - 1, fromIndex >> AddressBitsPerWord)
+
+    if (bits(idx) != 0) {
+      if (idx == bssize)
+        return idx
+
+      var j: Int = fromIndex % ElementSize
+      while (j >= 0) {
+        if ((bits(idx) & (1 << j)) != 0)
+          return idx * ElementSize + j
+
+        j -= 1
+      }
+    }
+
+    idx -= 1
+
+    while (idx >= 0 && bits(idx) == 0)
+      idx -= 1
+
+    if (idx == -1)
+      return -1
+
+    var j: Int = ElementSize - 1
+    while (j >= 0) {
+      if ((bits(idx) & (1 << j)) != 0)
+        return (idx << AddressBitsPerWord) + j
+
+      j -= 1
+    }
+
+    bssize
+    // scalastyle:on return
+  }
+
+  def previousClearBit(fromIndex: Int): Int = {
+    // scalastyle:off return
+    if (fromIndex == -1)
+      return -1
+
+    checkFromIndex(fromIndex)
+
+    val length = bits.length
+    val bssize = length << AddressBitsPerWord
+
+    if (fromIndex >= bssize)
+      return fromIndex
+
+    var idx = Math.min(bits.length - 1, fromIndex >> AddressBitsPerWord)
+
+    if (bits(idx) != (~0)) {
+      var j: Int = fromIndex % ElementSize
+      while (j >= 0) {
+        if ((bits(idx) & (1 << j)) == 0)
+          return idx * ElementSize + j
+
+        j -= 1
+      }
+    }
+
+    idx -= 1
+
+    while (idx >= 0 && bits(idx) == (~0))
+      idx -= 1
+
+    if (idx == -1)
+      return -1
+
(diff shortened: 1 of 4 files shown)


javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,808 @@
...
+    var idx = actualArrayLength - 1
+    while (idx >= 0 && bits(idx) == 0)
+      idx -= 1
+
+    actualArrayLength = idx + 1
+    isLengthActual = true
+    actualArrayLength
+  }
+
+  final private def pop(v: Long): Int = {
```
sjrd, line 771: This is `java.lang.Long.bitCount(v)`. (Or `Integer.bitCount(v)` for Ints.)

sjrd (changes requested):
Thanks.

Here are first a few high-level comments that will have an impact on many places.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,808 @@
...
+
+package java.util
+
+import java.io.ObjectInputStream
+import java.io.Serializable
+import java.nio.{ByteBuffer, LongBuffer}
+import java.util.ScalaOps.IntScalaOps
+
+private object BitSet {
+  private val Offset: Int = 6
```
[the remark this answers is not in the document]
er1c: Just curious if there is a link between "without explicit types" and "constant-folded"?  Or better yet what does "constant folded" mean?

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,808 @@
...
+      idx += 1
+
+    if (idx == actualArrayLength)
+      return -1 // scalastyle:ignore
+
+    // we know for sure there is a bit set to true in this element
+    // since the bitset value is not 0L
+    for (j <- 0 until BitSet.ElementSize) {
+      if ((bits(idx) & (BitSet.TwoNArray(j))) != 0)
+        return (idx << BitSet.Offset) + j // scalastyle:ignore
```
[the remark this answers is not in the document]
er1c: 👍 thanks for the explanation here

javalib/src/main/scala/java/util/BitSet.scala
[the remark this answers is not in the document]
sjrd: See "constant value definition" in [the spec](https://scala-lang.org/files/archive/spec/2.13/04-basic-declarations-and-definitions.html#value-declarations-and-definitions)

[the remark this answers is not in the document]
er1c: thanks!

er1c:
I'm getting outside of my comfort zone, I'm getting a test failure  on `def add` and not sure I understand  what  should  happen

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,808 @@
...
+import java.io.ObjectInputStream
+import java.io.Serializable
+import java.nio.{ByteBuffer, LongBuffer}
+import java.util.ScalaOps.IntScalaOps
+
+private object BitSet {
+  private val Offset: Int = 6
+  private val ElementSize: Int = 1 << Offset
+  private val RightBits: Int = ElementSize - 1
+  private val TwoNArray = Array[Long](
```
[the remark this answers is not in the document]
er1c: I turns out this was `1 << i` but I think I got it working, thanks!

er1c:
I did the first round of edits and got the tests passing again (hopefully CI confirms this :smile:

* changed any `for` that had a `return` into a `while` loop
* Changed `// scalastyle:ignore` to appropriate `// scalastyle:off return` blocks
* Looked for opportunities to avoid doing a `return` and using `if/else`
* removed the `pop` and am using the `Long.bitCount`
* updated the constant to be constant folded

er1c:
@sjrd I've updated this implementation to be based upon `Array[Int]` instead of `Array[Long]`, I added  some additional tests and its looking pretty good to me!

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,808 @@
...
+    }
+
+    bs
+  }
+
+  def valueOf(bb: ByteBuffer): BitSet = valueOf(bb.array())
+}
+
+@SerialVersionUID(7997698588986878753L)
+class BitSet private (private var bits: Array[Long], private var needClear: Boolean,
```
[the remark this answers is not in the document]
er1c: Got this done, thanks!

er1c:
@sjrd ping as a friendly reminder, but no rush

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+      }
+    }
+
+    needClear = true
+  }
+
+  def andNot(set: BitSet): Unit = {
+    if (needClear) {
+      val bsBits = set.bits
+      val range =
```
sjrd, line 691: That is not a very good name for an Int ^^ Consider calling it `minLength`. Also, it's basically `Math.min(actualArrayLength, set.actualArrayLength)`.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+        if (actualArrayLength < set.actualArrayLength) actualArrayLength
+        else set.actualArrayLength
+
+      for (i <- 0 until range)
+        bits(i) &= ~bsBits(i)
+
+      if (actualArrayLength < range)
+        actualArrayLength = range
```
sjrd, line 699: That can never happen. `range` is the minimum of `actualArrayLength` and `set.actualArrayLength`, so by definition is cannot be strictly greater than `actualArrayLength`.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+
+      isLengthActual = !((actualArrayLength > 0) && (bits(actualArrayLength - 1) == 0))
+    }
+  }
+
+  override def hashCode(): Int = {
+    var x: Long = 1234L
```
sjrd, line 706: The higher 32 bits of `x` will always be 0. It should therefore be an `Int` instead. (And it's going to be much faster.)

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+
+  override def equals(obj: Any): Boolean = {
+    // scalastyle:off return
+    if (obj.isInstanceOf[AnyRef] && (this eq obj.asInstanceOf[AnyRef]))
+      return true
+
+    if (!obj.isInstanceOf[BitSet])
+      return false
+
+    val bs: BitSet = obj.asInstanceOf[BitSet]
```
sjrd, line 727: We typically use the following pattern:
```scala
obj match {
  case obj: BitSet =>
    ...
  case _ =>
    false
}
```

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+        i += 1
+      }
+
+      i = length1
+      while (i < length2) {
+        if (bsBits(i) != 0)
+          return false
+
+        i += 1
+      }
```
sjrd, line 752: This should be factored out in a local
```scala
def areEqual(smallerArray: Array[Int], smallerLength: Int,
    largerArray: Array[Int], largerLength: Int): Boolean
```

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+    // scalastyle:on return
+  }
+
+  override def clone(): AnyRef = try {
+    val clone = super.clone.asInstanceOf[BitSet]
+    clone.bits = bits.clone
+    clone
+  } catch {
+    case _: CloneNotSupportedException => null
+  }
```
sjrd, line 781: This can be simplified and be more efficient as:
```suggestion
  override def clone(): AnyRef =
    new BitSet(bits.clone(), actualArrayLength)
```

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+
+  override def toString(): String = {
+    val sb = new StringBuilder(bits.length / 2)
```
sjrd, line 784: We typically use a local `var result: String` that we append to, since that is more efficient than a `StringBuilder` in JavaScript.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+    var bitCount: Int = 0
+    var comma: Boolean = false
+    sb.append('{')
+
+    for (i <- 0 until bits.length) {
```
sjrd, line 789:
```suggestion
    for (i <- 0 until getActualArrayLength()) {
```

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+    val clone = super.clone.asInstanceOf[BitSet]
+    clone.bits = bits.clone
+    clone
+  } catch {
+    case _: CloneNotSupportedException => null
+  }
+
+  override def toString(): String = {
+    val sb = new StringBuilder(bits.length / 2)
+    var bitCount: Int = 0
```
sjrd, line 785: Consider calling this `currentBitValue`. I was expecting `bitCount` to represent a number of bits present in the set, which made it difficult to understand this method.

Also, it might be better to just use `(i << AddressBitsPerWord) + j` when we need it instead.

javalib/src/main/scala/java/util/ScalaOps.scala
```diff
@@ -18,6 +18,9 @@ private[java] object ScalaOps {
   implicit class IntScalaOps private[ScalaOps] (val __self: Int) extends AnyVal {
     @inline def until(end: Int): SimpleRange =
       new SimpleRange(__self, end)
+
+    @inline def to(end: Int): SimpleRange =
+      new SimpleRange(__self, end+1)
```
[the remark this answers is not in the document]
sjrd: Maybe this won't even be necessary anymore after the other changes.

javalib/src/main/scala/java/util/ScalaOps.scala
```diff
@@ -30,6 +33,21 @@ private[java] object ScalaOps {
...
+  @inline
+  final class ReverseSimpleRange(start: Int, end: Int) {
+    @inline
+    def foreach[U](f: Int => U): Unit = {
+      var i = end-1
+      while (i >= start) {
+        f(i)
+        i -= 1
+      }
+    }
```
sjrd, line 50: This seems unused.

sjrd (changes requested):
Sorry for the delay in reviewing this.

I have a number of comments. There are quite a few things that can be significantly simplified, I believe.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+import java.io.ObjectInputStream
+import java.io.Serializable
+import java.nio.{ByteBuffer, LongBuffer}
+import java.util
+import java.util.ScalaOps.IntScalaOps
+
+private object BitSet {
+  private final val AddressBitsPerWord = 5 // Int Based 2^5 = 32
+  private val ElementSize: Int = 1 << AddressBitsPerWord
+  private val RightBits: Int = ElementSize - 1
```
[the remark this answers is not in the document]
er1c: du?mb question, is the final requires for an object?  or is this just to remove the typing from it

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+  private def toIntBit(idx: Int): Int = 1 << idx
```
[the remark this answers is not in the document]
er1c: will @inline do the same thing?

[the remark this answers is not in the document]
sjrd: It's not a question of performance, but of readability. It's easier to read `1 << idx` than `toIntBits(idx)`. I don't know what `toIntBits` does. `1 << idx` is clear, especially in the context of a bit set.

javalib/src/main/scala/java/util/BitSet.scala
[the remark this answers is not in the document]
sjrd: Adding the `final` and removing the type make that a *constant expression*, which is always replaced at call site by its value. This is a Scala thing (it's in the spec), not specific to Scala.js.

javalib/src/main/scala/java/util/BitSet.scala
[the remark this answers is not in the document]
er1c: Ahh funny, I didn't like the `1 << idx` so I added the helper to improve my readability :-P

er1c:
@sjrd sorry for the delay, I think I addressed all of the latest PR feedback 🙏🏻

sjrd:
Hi! Thank you.

Did you push all the changes you had made? It seems that the comments in the test suite have been addressed, but not the ones in the actual implementation.

javalib/src/main/scala/java/util/BitSet.scala
er1c: just swapping the type to Int causes the test to fail 🤔

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,859 @@
...
+    bs
+    // scalastyle:on return
+  }
+
+  def valueOf(bb: ByteBuffer): BitSet = valueOf(bb.array())
+}
+
+@SerialVersionUID(7997698588986878753L)
+class BitSet private (private var bits: Array[Int], private var needClear: Boolean,
+    private var actualArrayLength: Int, private var isLengthActual: Boolean) extends Serializable with Cloneable {
```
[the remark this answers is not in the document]
er1c: for the `isLengthActual` there is one edge case if the raw bits are different, but zero masks are set in extra bits, then it still returns true, there is a slight optimization the current code lets you exit if the actual aren't the same (vs the lazy calc) but I think the simplification outways that slight benefit.  I'll try and make that work next (I think this is the last issue I haven't addressed)

javalib/src/main/scala/java/util/BitSet.scala
er1c: I added a `equalsImpl` see if that works.

er1c:
@sjrd whooops, I missed those, they were hidden, I hope they are all addressed now, thanks!

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+    this(
+      bits = {
+        if (nbits < 0)
+          throw new NegativeArraySizeException
+
+        val length = (nbits + BitSet.RightBits) >> BitSet.AddressBitsPerWord
+
+        new Array[Int](length)
+      },
+      actualArrayLength = 0,
```
sjrd, line 83: This makes the CI fail on Scala 2.11 because of the trailing comma. (I can access the CI server via VPN, but it is not yet accessible to the public.)
```suggestion
      actualArrayLength = 0
```

sjrd (changes requested):
Thanks for the updates.

I have few more minor comments.

Regarding `actualArrayLength`, it is now clear that it basically only ever grows (except in `clear()` which is a full reset). That means that it doesn't provide any sort of guarantee (or even attempt) at making sure that the operations remain O(n) where n is the highest bit set.

Instead, it seems that it is only meant not to go over the last bit that was *ever* set. That is only important because, upon reallocation, we grow the array by a factor of 2 when we need more space. So if the highest bit ever set is *n*, we will stay O(n) instead of being O(2n). The thing is that O(2n) is in fact the same as O(n).

My conclusion is that `actualArrayLength` does not bring any improvement in terms of time complexity. Given that there are a number of places where bookkeeping needs to be done to preserve its correctness (making reasoning about the code more difficult), I have the impression that it is a net negative.

I suggest to remove `actualArrayLength` altogether, and always use `bits.length` instead.
WDYT?

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+      }
+    }
+  }
+
+  def clear(fromIndex: Int, toIndex: Int): Unit = {
+    // scalastyle:off return
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    if (actualArrayLength == 0)
+      return
```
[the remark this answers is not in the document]
er1c: removed

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+  }
+
+  def set(fromIndex: Int, toIndex: Int, value: Boolean): Unit =
+    if (value) set(fromIndex, toIndex)
+    else clear(fromIndex, toIndex)
+
+  def clear(bitIndex: Int): Unit = {
+    checkBitIndex(bitIndex)
+
+    if (actualArrayLength != 0) {
```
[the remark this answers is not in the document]
er1c: removed

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+  def flip(fromIndex: Int, toIndex: Int): Unit = {
+    checkToAndFromIndex(fromIndex, toIndex)
+
+    if (fromIndex != toIndex) {
+      val len2 = ((toIndex - 1) >> AddressBitsPerWord) + 1
+      ensureLength(len2)
+      val idx1 = fromIndex >> AddressBitsPerWord
+      val idx2 = (toIndex - 1) >> AddressBitsPerWord
+      val factor1 = (~0) << (fromIndex & RightBits)
+      val factor2 = (~0) >>> (ElementSize - (toIndex & RightBits))
```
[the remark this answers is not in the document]
er1c: updated all similar usages

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+    var i: Int = 0
+    while (i < smallerLength) {
+      if (smallerBS.bits(i) != largerBS.bits(i))
+        return false
+
+      i += 1
+    }
+
+    // Check remainder bits, if they are zero these are equal
+    i = smallerLength
```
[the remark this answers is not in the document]
er1c: I thought it was a nice documentation, but I removed :)

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,739 @@
...
+        bits(i) &= ~bsBits(i)
+    }
+  }
+
+  override def hashCode(): Int = {
+    var x: Long = 1234L
+    var i: Int = 0
+
+    while (i < actualArrayLength) {
+      x ^= bits(i) * (i + 1)
```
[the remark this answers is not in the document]
er1c: Thanks!  Updated

sjrd:
Thanks for the updates. We'll have to wait for Jenkins to be opened up again for GitHub to be able to send its requests. Our sysadmin is still working on those, as other, more public servers needed to be fixed with a higher priority.

sjrd:
Jenkins should be up again, but you'll need to retrigger it. You can do `git commit --amend` without changing anything to force the SHA1 of the commit to change, and force-push.

javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,694 @@
...
+  }
+
+  def valueOf(bb: ByteBuffer): BitSet = {
+    if (bb.hasArray()) valueOf(bb.array())
+    else {
+      val arr = new Array[Byte](bb.remaining())
+      bb.get(arr)
+      valueOf(arr)
+    }
+  }
```
gzm0, line 67: This is suspicious: it will modify the position of the passed Buffer only if it is not backed by an Array. Is that intended?

er1c: good catch, I forgot to add a bb.rewind(), let me double check the javadocs and I'll push an update

er1c: Here's a quick and dirty scastie as a proof of concept for JVM runtime: https://scastie.scala-lang.org/3ronEAsySBu6IX8cgGdwCw

er1c: @gzm0 updated with better tests, thanks!

er1c:
@sjrd this is failing, but not sure why:

    sbt "set scalaJSLinkerConfig in testSuite.v2_12 ~= (_.withESFeatures(_.withESVersion(ESVersion.ES5_1)))" 'set Seq(jsEnv in testSuite.v2_12 := new NodeJSEnvForcePolyfills(ESVersion.ES5_1), MyScalaJSPlugin.wantSourceMaps in testSuite.v2_12 := ("ES5_1" != "ES5_1"))' ++2.12.15 testSuite2_12/testOnly org.scalajs.testsuite.javalib.util.BitSetTest

er1c:
> @sjrd this is failing, but not sure why:
> 
> ```
> sbt "set scalaJSLinkerConfig in testSuite.v2_12 ~= (_.withESFeatures(_.withESVersion(ESVersion.ES5_1)))" 'set Seq(jsEnv in testSuite.v2_12 := new NodeJSEnvForcePolyfills(ESVersion.ES5_1), MyScalaJSPlugin.wantSourceMaps in testSuite.v2_12 := ("ES5_1" != "ES5_1"))' ++2.12.15 "testSuite2_12/testOnly org.scalajs.testsuite.javalib.util.BitSetTest"
> ```

@sjrd I think I fixed this, waiting on CI

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/BitSetJSTest.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.scalajs.testsuite.niobuffer.SupportsTypedArrays
+import org.scalajs.testsuite.utils.Platform._
+
+object BitSetJSTest extends SupportsTypedArrays
+
+class BitSetJSTest {
+
+  // This test fails in BitSetTest.scala without the SupportsTypedArrays
+  @Test def valueOf_ByteBuffer(): Unit = {
+    val eightBS = makeEightBS()
```
er1c, line 29: @sjrd should I keep this separate file, or is there some other remediation? or just delete this file all together?

It fails with:

```
[error] Test org.scalajs.testsuite.javalib.util.BitSetTest.valueOf_ByteBuffer failed: scala.scalajs.js.JavaScriptException: ReferenceError: Int8Array is not defined, took 0.002881443 sec
[error]     at java.nio.TypedArrayByteBuffer$.allocate(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:8781)
[error]     at org.scalajs.testsuite.javalib.util.BitSetTest.valueOf_ByteBuffer(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:127908)
[error]     at org.scalajs.testsuite.javalib.util.BitSetTest$scalajs$junit$bootstrapper$.invokeTest(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:213593)
[error]     at <jscode>.{anonymous}()(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:193804)
[error]     at org.scalajs.junit.JUnitTask.executeTestMethod(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:193852)
[error]     at org.scalajs.junit.JUnitTask.runTests$1(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:194078)
[error]     at <jscode>.{anonymous}()(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:194082)
[error]     at scala.scalajs.runtime.AnonFunction1.apply(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:311018)
[error]     at <jscode>.{anonymous}()(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:283578)
[error]     at scala.scalajs.runtime.AnonFunction1.apply(/Users/ericpeters/even/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt/main.js:311018)
[error] Failed: Total 35, Failed 1, Errors 0, Passed 34
[error] Failed tests:
[error] 	org.scalajs.testsuite.javalib.util.BitSetTest
```

sjrd:
Sorry; I'm a bit overloaded these days. I will look again on Monday. ;)

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/BitSetJSTest.scala
sjrd: Since it is only one test, and it only fails at run-time, it is best to include it normally with the rest of `BitSetTest.scala`, with an `assumeTrue` at the beginning to ignore it when typed arrays are not supported.

Normally that would be
```scala
// at the top:

...

  @Test def valueOf_ByteBuffer(): Unit = {
    assumeTrue("requires support for direct Buffers", typedArrays)

    ...
  }
```
But `typedArrays` itself is not available when compiling on the JVM, obviously.

I think the correct thing to do is to *add* a
```scala
def hasDirectBuffers: Boolean = typedArrays
```
in `test-suite/js/.../Platform.scala`, and
```scala
def hasDirectBuffers: Boolean = true
```
in `test-suite/jvm/.../Platform.scala`.
Then to use
```scala
assumeTrue("requires support for direct Buffers", hasDirectBuffers)
```
I'm surprised we didn't have any existing test that required this so far, TBH. ;)

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/util/BitSetJSTest.scala
```diff
@@ -0,0 +1,53 @@
...
+    assertEquals(new BitSet, BitSet.valueOf(ByteBuffer.allocateDirect(0)))
+
+    val directByteBuffer = ByteBuffer.allocateDirect(eightBytes.length + 1)
+    directByteBuffer.put(192.toByte) // extra byte
+    directByteBuffer.put(eightBytes)
+    directByteBuffer.rewind()
+    assertEquals(directByteBuffer.get(), 192.toByte) // extra byte
+    assertEquals(directByteBuffer.position(), 1)
+    assertEquals(eightBS, BitSet.valueOf(directByteBuffer))
+    assertEquals(directByteBuffer.position(), 1)
```
sjrd, line 42:
```suggestion
    assertEquals(192.toByte, directByteBuffer.get()) // extra byte
    assertEquals(1, directByteBuffer.position())
    assertEquals(eightBS, BitSet.valueOf(directByteBuffer))
    assertEquals(1, directByteBuffer.position())
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/BitSetTest.scala
```diff
@@ -0,0 +1,1521 @@
...
+
+    val eightBS = makeEightBS()
+    val eightBytes = eightBS.toByteArray()
+    assertEquals(eightBS, BitSet.valueOf(ByteBuffer.wrap(eightBytes)))
+
+    val bbWithPosition = ByteBuffer.wrap(192.toByte +: eightBytes)
+    assertEquals(bbWithPosition.get(), 192.toByte) // extra byte
+    assertEquals(bbWithPosition.position(), 1)
+    assertEquals(eightBS, BitSet.valueOf(bbWithPosition))
+    assertEquals(bbWithPosition.position(), 1)
```
sjrd, line 1467:
```suggestion
    assertEquals(192.toByte, bbWithPosition.get()) // extra byte
    assertEquals(1, bbWithPosition.position())
    assertEquals(eightBS, BitSet.valueOf(bbWithPosition))
    assertEquals(1, bbWithPosition.position())
```
The expected value is always on the left.

Same below.

sjrd (commented):
Looks like Monday lasted for 3 days :( Sorry about that.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/BitSetTest.scala
er1c: @sjrd updated, good find

er1c:
@sjrd thanks, updated

ran:

```sbt "set scalaJSLinkerConfig in testSuite.v2_12 ~= (_.withESFeatures(_.withESVersion(ESVersion.ES5_1)))" 'set Seq(jsEnv in testSuite.v2_12 := new NodeJSEnvForcePolyfills(ESVersion.ES5_1), MyScalaJSPlugin.wantSourceMaps in testSuite.v2_12 := ("ES5_1" != "ES5_1"))' ++2.12.15 "testSuite2_12/testOnly org.scalajs.testsuite.javalib.util.BitSetTest"
...
[warn] Test assumption in test org.scalajs.testsuite.javalib.util.BitSetTest.valueOf_ByteBuffer_typedArrays failed: org.junit.AssumptionViolatedException: requires support for direct Buffers, took 0.008 sec
[info] Passed: Total 36, Failed 0, Errors 0, Passed 35, Skipped 1
```

and

```
sbt:Scala.js> testSuite2_12/testOnly org.scalajs.testsuite.javalib.util.BitSetTest
...
[info] Fast optimizing /Users/eric/Work/scala-js/test-suite/js/.2.12/target/scala-2.12/scalajs-test-suite-test-fastopt
[info] Passed: Total 36, Failed 0, Errors 0, Passed 36
[success] Total time: 39 s, completed Dec 22, 2021 11:46:51 AM
```

sjrd (approved):
Thanks a lot!

PR #4218, opened 2020-09-29
Implement locale-sensitivity in `java.util.Formatter`.

sjrd:
Since the core does not provide `java.util.Locale` nor `java.text.*`, the new methods are only available for users if they use a third-party extension that provides those.

We use dummies in `javalib-ext-dummies` to test our implementation.

### DIFF
--- a/javalib-ext-dummies/src/main/scala/java/text/DecimalFormatSymbols.scala
+++ b/javalib-ext-dummies/src/main/scala/java/text/DecimalFormatSymbols.scala
@@ -1,0 +1,61 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package java.text
+
+import java.util.Locale
+
+/** Dummy implementation of `DecimalFormatSymbols`.
+ *
+ *  It is even worse than most other dummies, in the sense that it
+ *  special-cases the locales that we use in our tests (`FormatterTestEx`).
+ *  It is incorrect for most locales.
+ */
+class DecimalFormatSymbols(locale: Locale) extends NumberFormat {
+  def getZeroDigit(): Char = {
+    locale.getLanguage() match {
+      case "hi" =>
+        locale.getCountry() match {
+          case "IN" => '\u0966' // '०' DEVANAGARI DIGIT ZERO
+          case _    => unsupported()
+        }
+      case "" | "en" | "fr" =>
+        '0'
+      case _ =>
+        unsupported()
+    }
+  }
+
+  def getGroupingSeparator(): Char = {
+    locale.getLanguage() match {
+      case "fr"             => '\u00A0' // NO-BREAK SPACE
+      case "" | "en" | "hi" => ','
+      case _                => unsupported()
+    }
+  }
+
+  def getDecimalSeparator(): Char = {
+    locale.getLanguage() match {
+      case "fr"             => ','
+      case "" | "en" | "hi" => '.'
+      case _                => unsupported()
+    }
+  }
+
+  private def unsupported(): Nothing =
+    throw new Error(s"Unsupported locale '$locale' in DecimalFormatSymbols")
+}
+
+object DecimalFormatSymbols {
+  def getInstance(locale: Locale): DecimalFormatSymbols =
+    new DecimalFormatSymbols(locale)
+}
--- a/javalib/src/main/scala/java/util/Formatter.scala
+++ b/javalib/src/main/scala/java/util/Formatter.scala
@@ -18,7 +18,8 @@ import scala.scalajs.js
 import java.lang.{Double => JDouble}
 import java.io._
 
-final class Formatter(private[this] var dest: Appendable)
+final class Formatter private (private[this] var dest: Appendable,
+    formatterLocaleInfo: Formatter.LocaleInfo)
     extends Closeable with Flushable {
 
   import Formatter._
@@ -39,7 +40,13 @@ final class Formatter(private[this] var dest: Appendable)
   private[this] var closed: Boolean = false
   private[this] var lastIOException: IOException = null
 
-  def this() = this(null: Appendable)
+  def this() = this(null: Appendable, Formatter.RootLocaleInfo)
+
+  def this(a: Appendable) = this(a, Formatter.RootLocaleInfo)
+
+  def this(l: Locale) = this(null: Appendable, new Formatter.LocaleLocaleInfo(l))
+
+  def this(a: Appendable, l: Locale) = this(a, new Formatter.LocaleLocaleInfo(l))
 
   @inline
   private def trapIOExceptions(body: => Unit): Unit = {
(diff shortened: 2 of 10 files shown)


javalib-ext-dummies/src/main/scala/java/text/DecimalFormatSymbols.scala
```diff
@@ -0,0 +1,54 @@
...
+
+package java.text
+
+import java.util.Locale
+
+/** Dummy implementation of `DecimalFormatSymbols`.
+ *
+ *  It is even worse than most other dummies, in the sense that it
+ *  special-cases the locales that we use in our tests (`FormatterTestEx`).
+ *  It is incorrect for most locales.
```
gzm0, line 21: In the spirit of what we discussed on the last PR: Make it fail (hard!) for locales it doesn't know? So it either is correct or throws an `Error`?

javalib-ext-dummies/src/main/scala/java/text/DecimalFormatSymbols.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+class DecimalFormatSymbols(locale: Locale) extends NumberFormat {
+  def getZeroDigit(): Char = {
+    locale.getLanguage() match {
+      case "hi" =>
+        locale.getCountry() match {
+          case "IN" => '\u0966'
```
gzm0, line 28: We might want to follow this: https://google.github.io/styleguide/javaguide.html#s2.3.3-non-ascii-characters

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -39,7 +40,13 @@ final class Formatter(private[this] var dest: Appendable)
   private[this] var closed: Boolean = false
   private[this] var lastIOException: IOException = null
 
-  def this() = this(null: Appendable)
+  def this() = this(null: Appendable, Formatter.RootLocaleInfo)
```
gzm0, line 43: Shouldn't this (maybe just "technically") be the English locale? JavaDoc says if unspecified, its the default locale, and IIUC, Scala.js specifies this as "en-US". (the relevant test seems to indeed exclude the test for the JVM, even though the test suite JVM execution has been specifically prepared to be Scala.js standard).

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -631,7 +662,7 @@ final class Formatter(private[this] var dest: Appendable)
   }
 
   private def applyUpperCase(flags: Flags, str: String): String =
-    if (flags.upperCase) str.toUpperCase
+    if (flags.upperCase) str.toUpperCase() // not localized
```
gzm0, line 665: I'm not sure I understand this comment: In other places in this file, "not localized" seems to refer to "*should* not be localized". Here, I'm unsure (the JavaDoc seems contradictory to me):

> are the same as those for the corresponding lower-case conversion characters except that the result is converted to upper case according to the rules of the prevailing Locale. The result is equivalent to the following invocation of String.toUpperCase()`

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -757,4 +788,64 @@ object Formatter {
...
+    private def actualLocale: Locale =
+      if (locale == null) Locale.ROOT
+      else locale
+
+    private var _decimalFormatSymbols: DecimalFormatSymbols = _
+    private def decimalFormatSymbols: DecimalFormatSymbols = {
+      if (_decimalFormatSymbols == null)
+        _decimalFormatSymbols = DecimalFormatSymbols.getInstance(actualLocale)
+      _decimalFormatSymbols
+    }
```
gzm0, line 818: `lazy val`?

javalib/src/main/scala/java/util/Formatter.scala
gzm0: I see now in the tests that this seems to be the correct implementation. Consider clarifying the comment.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -757,4 +788,64 @@ object Formatter {
     final val AllWrittenFlags =
       LeftAlign | AltFormat | NumericOnlyFlags | UseLastIndex
   }
+
+  private sealed abstract class LocaleInfo {
```
gzm0, line 792: This class probably deserves a comment. It is non-obvious that this machinery is necessary to make `Formatter` link without `Locale`.

gzm0 (changes requested):

javalib/src/main/scala/java/util/Formatter.scala
sjrd: Ah, yes, I forgot to explicitly discuss that. It's true that so far we've specified Scala.js as en-US. If that is still the case, then yes this should be EnglishUS. But the reason I've said it was specified as en-US all along was because I did not know there was such a thing as the Root locale. And now that I know that it exists, I think we should redefine Scala.js as using the Root locale by specification.

In practice, this changes virtually nothing, because Root is basically the same as en, however it slightly differs from en-US, notably for currencies (the currency of Root and of en is `XXX`, while that of en-US is `USD`). So far we've never provided anything currency-oriented, though.

So why pick ROOT instead of the (conservative) US? Because it's less biased. It corresponds to all the algorithms that are specified as *locale-insensitive*. So in that sense it is more canonical than en-US, which is a biased choice.

sjrd: @cquiroz What do you think about the rationale above?

gzm0: Ok, fair enough that sounds reasonable. But can one actually configure a JVM to run in the Root locale? If yes, we should do so for our test suite, if no, it's a bit hairier :S

sjrd: It turns out that this is possible, by given explicit but empty properties for `-Duser.language` and `-Duser.country`.

sjrd:
Updated. I've left deeper changes to `ROOT` as the default locale in a separate commit for now.

gzm0:
> Updated. I've left deeper changes to ROOT as the default locale in a separate commit for now. (Arguably the two commits should be reversed if we go with the ROOT locale as default.)

Sweet. Definitely in favor!

sjrd:
Commits reversed :)

javalib-ext-dummies/src/main/scala/java/text/DecimalFormatSymbols.scala
```diff
@@ -0,0 +1,61 @@
...
+      case "" | "en" | "fr" =>
+        '0'
+      case _ =>
+        unsupported()
+    }
+  }
+
+  def getGroupingSeparator(): Char = {
+    locale.getLanguage() match {
+      case "fr"             => '\u00A0'
```
gzm0, line 40: Comment what this is? (no breaking space IIUC)

project/Build.scala
```diff
@@ -2006,6 +2006,8 @@ object Build {
 
       libraryDependencies +=
         "com.novocode" % "junit-interface" % "0.11" % "test",
+  ).dependsOn(
+      testSuiteJVM
```
gzm0, line 2010: Unneeded now :)

test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTestEx.scala
```diff
@@ -0,0 +1,96 @@
...
+  @Test def testFormatFrench(): Unit = {
+    assertF(French, "1\u00A0234\u00A0567", "%,d", 1234567)
+    assertF(French, "1\u00A0234\u00A0567,89", "%,.2f", 1234567.89)
+    assertF(French, "0012", "%04d", 12)
+  }
+
+  @Test def testFormatHindiIndia(): Unit = {
+    assertF(HindiIndia, "\u0967,\u0968\u0969\u096A,\u096B\u096C\u096D", "%,d", 1234567)
+    assertF(HindiIndia, "\u0967,\u0968\u0969\u096A,\u096B\u096C\u096D.\u096E\u096F", "%,.2f", 1234567.89)
+    assertF(HindiIndia, "\u0966\u0966\u0967\u0968", "%04d", 12)
```
gzm0, line 73: Consider writing these out in actual unicode. (I believe we have other funky unicode characters in the codebase anyways). Like this, it is very hard to check if the test is reasonable.

gzm0 (approved):
Only nits.

PR #4219, opened 2020-09-30
Add JDK 11 to our CI matrix.

sjrd:


er1c:
I'm scared to click into CI to see what happens to the Character/ucd tests

javalib-ext-dummies/src/main/scala/java/util/Locale.scala
```diff
@@ -25,18 +32,49 @@ final class Locale(languageRaw: String, countryRaw: String)
...
+        val value = getExtension(key)
+        if (value != null) {
+          if (first)
+            first = false
+          else
+            result += "-"
+          result += s"$key-$value"
+        }
+        key = (key + 1).toChar
+      }
```
gzm0, line 63: Do we really need to make this that low level? (We might even use the scalalib here, no?)

gzm0 (approved):
Only a minor nit. If the thing really can't be simplified, feel free to merge.

javalib-ext-dummies/src/main/scala/java/util/Locale.scala
sjrd: Yes, good point. Fixed.

gzm0 (approved):

PR #4220, opened 2020-09-30
Set temp file permissions to 0644 on PathOutputFile

janilcgarcia:
Update PathOutputFile on the linker to ensure that the compiled JS file generated by it has permissions 0644 on POSIX systems.

Fixes issue #4212.

I adapted my code to the coding style and and I'm reopening this new PR following the contributing guidelines. If I still got something wrong, just let me know. The previous PR was https://github.com/scala-js/scala-js/pull/4213

I'm still using `setPosixFilePermissions` as I did not receive any answer on my last comment in the other PR. And, as it is already an debatable decision, I left a comment explaining the reason for my decision.

If you'd rather have permissions set as a `FileAttribute` passed to `createTempFile`, I'll fix it.

### DIFF
--- a/linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
+++ b/linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
@@ -17,6 +17,9 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
+import java.nio.file.Paths
+
+import java.util.EnumSet
 
 import org.junit.Test
 import org.junit.Assert._
@@ -26,6 +29,7 @@ import org.scalajs.junit.async._
 import com.google.common.jimfs.Jimfs
 
 import org.scalajs.linker.interface.unstable.OutputDirectoryImpl
+
 
 class PathOutputDirectoryTest {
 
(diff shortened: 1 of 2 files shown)


linker/jvm/src/main/scala/org/scalajs/linker/PathOutputFile.scala
```diff
@@ -35,8 +37,28 @@ object PathOutputFile {
...
+           * with permissions 0600 which deviates from the standard 0644 used
+           * in *.class and *.jar files.
+           * 
+           * Uses setPosixFilePermissions instead of passing the permissions to
+           * createTempFile to cover systems (if any exist) that uses POSIX
+           * permissions, but can't set them atomically while creating a
+           * temporary file.
+           */
+          val permissions = PosixFilePermissions.fromString("rw-r--r--")
+          blocking(Files.setPosixFilePermissions(tmpFile, permissions))
```
sjrd, line 55: I don't think there's any point in having two independent `blocking()` calls here. Consider wrapping one `blocking { ... }` around both the `createTempFile` and the `setPosixFilePermissions`.

sjrd (commented):
There is a conflict because the logic to create files was moved to `PathOutputDirectory`, next to `PathOutputFile`. Would you mind rebasing on master and moving your fixed logic to that file, please?

linker/jvm/src/main/scala/org/scalajs/linker/PathOutputFile.scala
```diff
@@ -35,8 +37,28 @@ object PathOutputFile {
...
+          /* Set file permissions for temporary file as in Linux it is created
+           * with permissions 0600 which deviates from the standard 0644 used
+           * in *.class and *.jar files.
+           * 
+           * Uses setPosixFilePermissions instead of passing the permissions to
+           * createTempFile to cover systems (if any exist) that uses POSIX
+           * permissions, but can't set them atomically while creating a
+           * temporary file.
+           */
+          val permissions = PosixFilePermissions.fromString("rw-r--r--")
```
gzm0, line 54: We should probably us an EnumSet, rather than this stringly-typed factory.

linker/jvm/src/main/scala/org/scalajs/linker/PathOutputFile.scala
janilcgarcia: Okay, I will.

linker/jvm/src/main/scala/org/scalajs/linker/PathOutputFile.scala
janilcgarcia: Yes, you're right. I got lazy there, I will rebase and redo the changes today and take the chance to change that.

janilcgarcia:
Okay, guys, rebased and changed logic to PathOutputDirectory.

linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
```diff
@@ -54,4 +58,33 @@ class PathOutputDirectoryTest {
       assertEquals(lastModifiedBefore, lastModifiedAfter)
     }
   }
+
+  @Test
+  def sanePosixFilePermissions(): AsyncResult = await {
+    import java.nio.file.attribute.PosixFilePermission._
+
+    val tmpDir = Files.createTempDirectory("test-dir")
```
gzm0, line 66: Consider using Jimfs. This will even allow you to configure whether the file system supports posix attributes and hard assert that it works. And you can write a (platform independent) test making sure that `PathOutputDirectory` does not fail if the underlying FS doesn't support posix attributes.

gzm0:
@janilcgarcia It seems there hasn't been any activity on this for quite a while. If its OK with you, I can take this over so we can fix the issue.

janilcgarcia:
> @janilcgarcia It seems there hasn't been any activity on this for quite a while. If its OK with you, I can take this over so we can fix the issue.

Yeah, please do, I can't right now.

gzm0:
Superseded by #4299

PR #4221, opened 2020-10-02
Clean up tests to make them compatible with dotty.

sjrd:


test-suite/js/src/test/require-scala2/org/scalajs/testsuite/jsinterop/NonNativeJSTypeTestScala2.scala
```diff
@@ -0,0 +1,49 @@
...
+  @Test def default_values_for_fields(): Unit = {
+    val obj = new DefaultFieldValues
+
+    /* Value class fields are initialized to null, instead of a boxed
+     * representation of the zero of their underlying types, as for a
+     * Scala class.
+     *
+     * Scala 3 correctly unboxes `null` as the zero of the underlying type of
+     * value classes, which means that this test "fails" in Scala 3, although
+     * it is in fact a progression.
```
gzm0, line 34: I find the usage of the term "progression" as an antonym to "regression" quite confusing (it wasn't clear to me what it meant in the commit message). If it is common usage though in dotty / scala land, then we should leave it.

test-suite/js/src/test/require-scala2/org/scalajs/testsuite/jsinterop/NonNativeJSTypeTestScala2.scala
```diff
@@ -0,0 +1,49 @@
...
+
+object NonNativeJSTypeTestScala2 {
+
+  class DefaultFieldValues extends js.Object {
+    var valueClass: SomeValueClass = _
+  }
+
+  class SomeValueClass(val i: Int) extends AnyVal
+
+}
```
gzm0, line 49: You should remove these things from `NonNativeJSTypeTest`. They are unused there (or if they still provide value to make sure value classes "work", comment on them).

test-suite/shared/src/test/require-scala2/org/scalajs/testsuite/compiler/WasPublicBeforeTyperTestScala2.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.testsuite.utils.AssertThrows._
+
+class WasPublicBeforeTyperTestScala2 {
+
+  /** This is a Scala 2.x-only test because it is a bug compatibility test,
+   *  anddotty does expose the bug.
```
gzm0, line 23: typo: and_dotty

gzm0 (changes requested):

test-suite/js/src/test/require-scala2/org/scalajs/testsuite/jsinterop/NonNativeJSTypeTestScala2.scala
sjrd: Huh. I've seen that often in the scala/scala repo, and thought it was standard terminology. I can use "an improvement" instead, perhaps?

gzm0: "improvement" is definitely clear to me. But if progression is common, then maybe I just needed to learn :)

sjrd:
Updated. I changed to "improvement", since I did not find any evidence that "progression" is used outside of scala/scala.

gzm0 (approved):

PR #4231, opened 2020-10-06
Implement the default methods of java.util.List

sjrd:


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/ListTest.scala
```diff
@@ -436,6 +438,91 @@ trait ListTest extends CollectionTest {
...
+  @Test def sortWithComparator(): Unit = {
+    testSortWithComparator[CustomComparable](new CustomComparable(_),
+        (x, y) => x.compareTo(y), false)
+    testSortWithComparator[jl.Integer](_.toInt, (x, y) => x.compareTo(y))
+    testSortWithComparator[jl.Long](_.toLong, (x, y) => x.compareTo(y))
+    testSortWithComparator[jl.Double](_.toDouble, (x, y) => x.compareTo(y))
+  }
+
+  private def testSortWithNaturalOrdering[T <: AnyRef with Comparable[T]: ClassTag](
+      toElem: Int => T, absoluteOrder: Boolean = true): Unit = {
```
gzm0, line 473: The `absoluteOrder` parameter seems to be never set.

gzm0 (approved):
Just a nit.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/ListTest.scala
sjrd: It was set in two places, but not with the named parameter syntax. I've changed the call sites to use named parameters so that this is clearer.

gzm0 (approved):

PR #4232, opened 2020-10-06
Fix #4224: Reject multiple public module IDs when emitting NoModule

### LINKED ISSUE #4224: When in NoModule mode, multiple public module IDs crash the linker
I haven't reproduced this, but I know it because I have written the code :P

@sjrd, IIRC, we have agreed that the best behavior here is to simply ignore any public module ID when emitting to `NoModule` kind. An alternative would be to reject more than one module when in `NoModule` mode (arguably the more conservative choice).

In any case, this needs to be fixed, because the condition must not crash the linker.

gzm0:


sjrd (approved):

gzm0:
As usual, a trailing comma :-/

sjrd:
One day we'll get rid of 2.11 and 2.12.1 and it won't be a problem anymore... 😇

PR #4234, opened 2020-10-07
Fix #4226: Add specialized versions of j.u.function.Supplier

### LINKED ISSUE #4226: Add specialized type j.u.function.{Boolean|Double|Int|Long}Supplier implementations
Add remaining specialized `j.u.function.Supplier` implementations

- [ ] java.util.function.BooleanSupplier
- [ ] java.util.function.DoubleSupplier
- [ ] java.util.function.IntSupplier
- [ ] java.util.function.LongSupplier
- [x] java.util.function.Supplier

er1c:


sjrd:
Perhaps @gzm0 has a different opinion, but IMO all those commits should be squashed in a single

    Add specialized versions of j.u.function.Supplier.

They are all closely related, and having one without the others does not make sense. So the intermediate commits do not make sense to me.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/function/DoubleSupplierTest.scala
```diff
@@ -0,0 +1,41 @@
...
+  import DoubleSupplierTest._
+  private val delta = 1e31
+
+  @Test def getAsDouble(): Unit = {
+    assertEquals(Double.NaN, makeSupplier(Double.NaN).getAsDouble(), delta)
+    assertEquals(Double.MinValue, makeSupplier(Double.MinValue).getAsDouble(), delta)
+    assertEquals(Double.MinPositiveValue, makeSupplier(Double.MinPositiveValue).getAsDouble(), delta)
+    assertEquals(Double.PositiveInfinity, makeSupplier(Double.PositiveInfinity).getAsDouble(), delta)
+    assertEquals(Double.NegativeInfinity, makeSupplier(Double.NegativeInfinity).getAsDouble(), delta)
+    assertEquals(1.234D, makeSupplier(1.234D).getAsDouble(), delta)
```
sjrd, line 30: I would keep only this one. There's no real point testing all sorts of values in this case, as it suggests that `DoubleSupplier` is somehow sensitive to particular values, which it most definitely is not.

---

Also, we always use a `0.0` delta. There are no computations, so there can be no loss of precision. Also we use lowercase `d` and `f` for double and float suffixes (but `L` for the Long suffix), as is typically widely acknowledged as best practice in Scala:
```suggestion
    assertEquals(1.234d, makeSupplier(1.234d).getAsDouble(), 0.0d)
```

sjrd (changes requested):

er1c:
> Perhaps @gzm0 has a different opinion, but IMO all those commits should be squashed in a single
> 
> ```
> Add specialized versions of j.u.function.Supplier.
> ```
> 
> They are all closely related, and having one without the others does not make sense. So the intermediate commits do not make sense to me.

Gotta game the GitHub commit count rankings! 😉 In all seriousness, I just split them out based upon a previous comment from @gzm0. I can do whatever.

er1c:
@sjrd updated, waiting on CI

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/function/DoubleSupplierTest.scala
```diff
@@ -0,0 +1,35 @@
...
+package org.scalajs.testsuite.javalib.util.function
+
+import java.util.function.DoubleSupplier
+
+import org.junit.Assert._
+import org.junit.Test
+
+class DoubleSupplierTest {
+  import DoubleSupplierTest._
+  private val delta = 1e31
```
sjrd, line 22: This is now dead code.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/function/DoubleSupplierTest.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  @Test def getAsDouble(): Unit = {
+    assertEquals(1.234D, makeSupplier(1.234D).getAsDouble(), 0.0)
```
sjrd, line 25: The double suffix should be lowercase (or not there, since double is the default):
```suggestion
    assertEquals(1.234, makeSupplier(1.234).getAsDouble(), 0.0)
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/function/LongSupplierTest.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import org.junit.Assert._
+import org.junit.Test
+
+class LongSupplierTest {
+  import LongSupplierTest._
+
+  @Test def getAsLong(): Unit = {
+    assertEquals(Long.MinValue, makeSupplier(Long.MinValue).getAsLong())
+    assertEquals(1024L, makeSupplier(1024L).getAsLong())
```
sjrd, line 25: The Long suffix is uppercase, so this is correct ;)

sjrd (changes requested):

sjrd (approved):

PR #4237, opened 2020-10-08
Merge the require-jdk{7,8} tests into the main tests.

sjrd:
We have been requiring JDK 8+ for a very long time now. The reason we had kept those separate was only to ease merges between 0.6.x and master. Now that 0.6.x is EOL, we can get rid of that separation.


gzm0 (approved):
I have not reviewed that the moved bits of test are the same in the old and the new files. If there is something specific in that regard that you'd like me to look at, please LMK.

sjrd:
No, there isn't anything special.

The only tests that were not directly cut-and-pasted are `CharacterTestOnJDK7.shouldProvideIsBmpCodePoint` and `shouldProvideIsSurrogate`. Those were dropped, because we already had existing (better) tests for those methods in `CharacterTest` at https://github.com/scala-js/scala-js/blob/d44537012bba0077963f9866998eca30cd8d2b8c/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala#L55-L70 and https://github.com/scala-js/scala-js/blob/d44537012bba0077963f9866998eca30cd8d2b8c/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala#L125-L141

PR #4239, opened 2020-10-09
Add .jvmopts

er1c:
I forget if this was already commented on, but it helps just running "sbt" without having to add -mem4096


sjrd:
This has been proposed before, but I still don't like it. I configure my system-wide sbt to have even more memory than that (6GB), which is not necessarily appropriate for every system. I believe each user can configure their sbt to use what's reasonable in their system. Adding a `.jvmopts` prevents that.

er1c:
@sjrd you probably work on "less projects" than the average developer, if I have multiple sbt instances open, I don't want them opening more memory than they need.

hepin1989:
@er1c Better mention this in the contribution guide?

er1c:
I'll close this, I'll open a PR with a `.gitignore` addition for `.jvmopts` does that work?

PR #4240, opened 2020-10-10
Add .jvmopts and .sbtopts to .gitignore

er1c:
Allows users to have `.jvmopts` and `.sbtopts` settings without permenantly having them in unstaged, state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,7 @@ bin/
 /.metals/
 /project/**/metals.sbt
 /.vscode/
+
+# User specific
+/.jvmopts
+/.sbtopts


er1c:
@sjrd I think this might be a good balance/alternative to https://github.com/scala-js/scala-js/pull/4239

.gitignore
```diff
@@ -18,3 +18,7 @@ bin/
 /.metals/
 /project/**/metals.sbt
 /.vscode/
+
+# User specific
+.jvmopts
+.sbtopts
```
sjrd, line 24: You can use a leading `/` here:
```suggestion
/.jvmopts
/.sbtopts
```

sjrd (changes requested):
Yes, this is indeed a good balance. I have one comment.

sjrd (approved):

PR #4246, opened 2020-10-13
Fix #4244: Patch the source map links in the output of {fast,full}OptJS.

### LINKED ISSUE #4244: legacyKey (fastOptJS / fullOptJS) implementation does not adjust source-map references
There simply is no code to do it. I somehow thought about doing that for the Linker, but not for the sbt plugin :shrug:

sjrd:
We reuse the patching functions used in the legacy `Linker.link` method, although that creates a coupling between linker-interface and sbt-plugin, which did not exist before.

---

Tested locally with the manual tests.

I'm not sure this is the best way to do this.

### DIFF
--- a/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
+++ b/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala
@@ -98,7 +98,7 @@ abstract class Linker private[interface] () {
   }
 }
 
-private object Linker {
+private[scalajs] object Linker {
   private val sourceMapRe = """(?m)^//# sourceMappingURL=.*$""".r
 
   /* It is somewhat acceptable to parse the JSON field "file" with a Regex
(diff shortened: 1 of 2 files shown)


gzm0:
I have an alternative at: https://github.com/scala-js/scala-js/compare/master...gzm0:legacy-adapter?expand=1 (not fully built yet).

The upside is it lets us share more, also it doesn't really increase the surface of the linker interface.

sjrd:
It seems your solution is better :)

gzm0:
> It seems your solution is better :)

There is just a little flaw: It requires reading back from an arbitrary `OutputDirectory` which is currently not part of the API. Maybe we should just make it part of the API?

sjrd:
Given that we already have `listFiles()` in there, I believe it would be appropriate to add an API to read a file, yes.

gzm0:
Sooo... My branch is basically ready for testing, but I cannot get rid of the deprecation warnings in the sbtPlugin. Is there *any* way to do that on 2.12?

sjrd:
What warning do you have, exactly?

PR #4248, opened 2020-10-13
Add explicit conversions required on JDK13+

catap:
jdk13 introduced a new methods for `CharBuffer` and `ShortBuffer` that brokes building with errors like:
```
[error] test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala:162:16: overloaded method value put with alternatives:
[error]   (x$1: Int,x$2: Array[Char])java.nio.CharBuffer <and>
[error]   (x$1: Int,x$2: Char)java.nio.CharBuffer
[error]  cannot be applied to (Int, Int)
[error]         charBuf1.put(1, 0x7e7f)
```
and
```
[error] test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala:316:17: overloaded method value put with alternatives:
[error]   (x$1: Int,x$2: Array[Short])java.nio.ShortBuffer <and>
[error]   (x$1: Int,x$2: Short)java.nio.ShortBuffer
[error]  cannot be applied to (Int, Int)
[error]         shortBuf1.put(1, 0x7e7f)
```

anyway, a few lines bellow the same tests are doint the same trick with `.toChar` and `.toShort` to same value.

Let make it here to allow build scala-native by jdk13+

This is backport from Scala-Native: https://github.com/scala-native/scala-native/pull/1939

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/niobuffer/ByteBufferTest.scala
@@ -159,7 +159,7 @@ abstract class ByteBufferTest extends BaseBufferTest {
 
       buf.order(ByteOrder.BIG_ENDIAN)
       val charBuf1 = buf.asCharBuffer()
-      charBuf1.put(1, 0x7e7f)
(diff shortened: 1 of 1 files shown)


sjrd (approved):
Thanks :)

PR #4253, opened 2020-10-15
Fix #4252: Fix `Matcher.lookingAt()` and `matches()` under a `region()`.

### LINKED ISSUE #4252: regex.Matcher.lookingAt() and matches() incorrect in 1.3.0
This is a regression in regex matcher() compared to 1.2.0:

```
val p = java.util.regex.Pattern.compile("a+")
val m = p.matcher("abaa").region(2,4);
println(m.lookingAt())
println(m.group())
```

In 1.2.0, the code would correctly print:
```
true
aa
```

In 1.3.0:
```
false
/tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:1128
    throw new $c_jl_IllegalStateException("No match available")
    ^

java.lang.IllegalStateException: No match available
    at $p_ju_regex_Matcher__ensureLastMatch__sjs_js_RegExp$ExecResult (/tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:1128:11)
    at $c_ju_regex_Matcher.group__T (/tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:1201:19)
    at $c_Ltutorial_webapp_TutorialApp$.main__AT__V (/tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:1030:19)
    at $s_Ltutorial_webapp_TutorialApp__main__AT__V (/tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:1013:38)
    at /tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:3244:1
    at /tmp/sjs/target/scala-2.13/scala-js-tutorial-fastopt/main.js:3245:4
    at Script.runInThisContext (vm.js:122:20)
    at Object.runInThisContext (vm.js:329:38)
    at [stdin]:8:25
    at Script.runInThisContext (vm.js:122:20)
```

This blocks upgrading ScalaPB to Scala.js 1.3.x.

sjrd:
When we fixed the behavior of `start()` under a `region()` in 537558013fb4ec30580e05a67d167c877afc6ffb, we failed to adapt `lookingAt()` and `matches()`, which were relying on the previous (broken) behavior.

This commit fixes that omission.


test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/regex/RegexMatcherTest.scala
```diff
@@ -285,23 +285,49 @@ class RegexMatcherTest  {
...
+    assertFalse(matcher0.matches())
+    matcher0.region(0, 15)
     assertTrue(matcher0.find())
     assertEquals("Scalable", matcher0.group)
 
     matcher0.region(2, 7)
     assertEquals(2, matcher0.regionStart)
     assertEquals(7, matcher0.regionEnd)
+    assertTrue(matcher0.lookingAt())
+    matcher0.region(2, 7)
```
gzm0, line 308: Do we need to / want to reset the region here? The code suggests we don't need to.

gzm0: Oh, I see. This is probably affected by #4254...

sjrd: Yes, we need to. The code suggests we don't need to because it's broken (#4254), but if we don't reset the region, the test does not pass on the JVM.

gzm0 (approved):

Issue #3407, opened 2018-07-05
Limitations of java.util.Formatter (and String.format)


sjrd:
After the big rewrite of #3380 and its numerous fixes, there are still a few things that are not supported in `java.util.Formatter` and therefore in `String.format`:

* Hexadecimal format for floating point numbers with `%a`.
* Formatting of `BigInteger`s and `BigDecimal`s.
* Formatting of date/times with the `%t.` family.

The first two might be doable at some point. Formatting dates would bring in way to much code in every Scala.js application, and there are better APIs for that now in `java.time` anyway.

See also #3202 which records that `%f`, `%e` and `%g` have rounding discrepancies compared to the JVM.

pfcoperez:
I had a patch for `BigInteger` formatter and I was opening an issue to describe what it fixes when I found this issue so I ended up not opening another issue :sweat_smile: 

This is the PR with the patch fixing `b.formatted("%d")`, `b.formatted("%o")` and `b.formatted("%x")`.

sjrd:
An update here:

* `BigInteger` support was implemented in #4255, as mentioned above.
* `BigDecimal` support was implemented in #4444.
* The `%a` conversion was implemented in #4440.

In addition, the related issue #3202 was fixed in #4443.

That leaves only the date/time conversions, for which I have created a separate issue #4447.

So this one can be closed now.

PR #4255, opened 2020-10-15
Add support for `%o`, `%x` and `%d` formats to `BigInteger` formatter

pfcoperez:
This PR is aimed to fix the limitations of `bigInt.formatted(format)` method in `BigInteger`. Adding support for:

- `"%d"` 
- `"%o"`
- `"%x"`

Partially completes features listed in https://github.com/scala-js/scala-js/issues/3407

### DIFF
--- a/javalib/src/main/scala/java/util/Formatter.scala
+++ b/javalib/src/main/scala/java/util/Formatter.scala
@@ -17,6 +17,7 @@ import scala.scalajs.js
 
 import java.lang.{Double => JDouble}
 import java.io._
+import java.math.BigInteger
 
 final class Formatter private (private[this] var dest: Appendable,
     formatterLocaleInfo: Formatter.LocaleInfo)
@@ -360,6 +361,8 @@ final class Formatter private (private[this] var dest: Appendable,
             formatNumericString(localeInfo, flags, width, arg.toString())
           case arg: Long =>
             formatNumericString(localeInfo, flags, width, arg.toString())
+          case arg: BigInteger =>
+            formatNumericString(localeInfo, flags, width, arg.toString())
           case _ =>
             formatNullOrThrowIllegalFormatConversion()
         }
@@ -379,6 +382,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 java.lang.Long.toOctalString(arg))
+          case arg: BigInteger =>
+            formatNumericString(RootLocaleInfo, flags, width,
+                arg.toString(8), prefix)
           case _ =>
             formatNullOrThrowIllegalFormatConversion()
         }
@@ -400,6 +406,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 applyNumberUpperCase(flags, java.lang.Long.toHexString(arg)))
+          case arg: BigInteger =>
+            formatNumericString(RootLocaleInfo, flags, width,
+                arg.toString(16), prefix)
           case _ =>
             formatNullOrThrowIllegalFormatConversion()
         }
@@ -589,13 +598,13 @@ final class Formatter private (private[this] var dest: Appendable,
   }
 
   private def formatNumericString(localeInfo: LocaleInfo, flags: Flags,
-      width: Int, str: String): Unit = {
+      width: Int, str: String, basePrefix: String = ""): Unit = {
     /* Flags for which a numeric string needs to be decomposed and transformed,
      * not just padded and/or uppercased. We can write fast-paths in this
      * method if none of them are present.
      */
     val TransformativeFlags =
-      PositivePlus | PositiveSpace | UseGroupingSeps | NegativeParen
+      PositivePlus | PositiveSpace | UseGroupingSeps | NegativeParen | AltFormat
 
     if (str.length >= width && !flags.hasAnyOf(TransformativeFlags)) {
       // Super-fast-path
@@ -605,7 +614,7 @@ final class Formatter private (private[this] var dest: Appendable,
       padAndSendToDestNoZeroPad(flags, width, applyNumberUpperCase(flags, str))
     } else {
       // Extract prefix and rest, based on flags and the presence of a sign
-      val (prefix, rest0) = if (str.charAt(0) != '-') {
+      val (numberPrefix, rest0) = if (str.charAt(0) != '-') {
         if (flags.positivePlus)
           ("+", str)
         else if (flags.positiveSpace)
@@ -618,6 +627,8 @@ final class Formatter private (private[this] var dest: Appendable,
         else
           ("-", str.substring(1))
       }
+
+      val prefix = numberPrefix + basePrefix
 
       // Insert grouping separators, if required
       val rest =
--- a/test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTestEx.scala
+++ b/test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTestEx.scala
@@ -13,6 +13,7 @@
 package org.scalajs.testsuite.javalib.util
 
 import java.util.{Formatter, Locale}
+import java.math.BigInteger
 
 import org.junit.Test
 import org.junit.Assert._
@@ -75,6 +76,17 @@ class FormatterTestEx {
     assertF(French, "1\u00A0234\u00A0567", "%,d", 1234567)
     assertF(French, "1\u00A0234\u00A0567,89", "%,.2f", 1234567.89)
     assertF(French, "0012", "%04d", 12)
+    assertF(French, "0012", "%04d", new BigInteger("12"))
+
+    assertF(French, "0014", "%04o", new BigInteger("12"))
+    assertF(French, "-014", "%04o", new BigInteger("-12"))
+    assertF(French, "14", "%o", new BigInteger("12"))
+    assertF(French, "-14", "%o", new BigInteger("-12"))
+
+    assertF(French, "0x0c", "%#04x", new BigInteger("12"))
+    assertF(French, "-0xc", "%#04x", new BigInteger("-12"))
+    assertF(French, "12", "%x", new BigInteger("18"))
+    assertF(French, "-12", "%x", new BigInteger("-18"))
   }
 
   @Test def testFormatHindiWithDevanagariDigits(): Unit = {
@@ -86,6 +98,16 @@ class FormatterTestEx {
     assertF(HindiWithDevanagariDigits, "0x0012", "%#06x", 0x12)
     assertF(HindiWithDevanagariDigits, "0X0012", "%#06X", 0x12)
     assertF(HindiWithDevanagariDigits, "000014", "%#06o", 12)
+
+    assertF(HindiWithDevanagariDigits, "0x0012", "%#06x", new BigInteger("18"))
+    assertF(HindiWithDevanagariDigits, "-0x012", "%#06x", new BigInteger("-18"))
+    assertF(HindiWithDevanagariDigits, "12", "%x", new BigInteger("18"))
+    assertF(HindiWithDevanagariDigits, "-12", "%x", new BigInteger("-18"))
+
+    assertF(HindiWithDevanagariDigits, "000014", "%#06o", new BigInteger("12"))
+    assertF(HindiWithDevanagariDigits, "-00014", "%#06o", new BigInteger("-12"))
+    assertF(HindiWithDevanagariDigits, "14", "%o", new BigInteger("12"))
+    assertF(HindiWithDevanagariDigits, "-14", "%o", new BigInteger("-12"))
   }
 
   @Test def testFormatTurkish(): Unit = {
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
@@ -12,6 +12,8 @@
 
 package org.scalajs.testsuite.javalib.util
 
+import java.math.BigInteger
+
 import org.junit.Assert._
 import org.junit.Test
 
@@ -274,10 +276,17 @@ class FormatterTest {
   @Test def `should_provide_d_conversion`(): Unit = {
     assertF("5", "%d", 5)
     assertF("-5", "%d", -5)
+    assertF("5", "%d", new BigInteger("5"))
+    assertF("-5", "%d", new BigInteger("-5"))
+
 
     assertF("00005", "%05d", 5)
     assertF("  -10", "%5d", -10)
     assertF("-0010", "%05d", -10)
+
+    assertF("00005", "%05d", new BigInteger("5"))
+    assertF("  -10", "%5d", new BigInteger("-10"))
+    assertF("-0010", "%05d", new BigInteger("-10"))
 
     assertF("345,678", "%,d", 345678)
     assertF("-345,678", "%,d", -345678)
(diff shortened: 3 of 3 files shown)


sjrd:
Hello, and thanks for the PR.

Before I start looking at it, could you confirm that you did not look at any GPL-licenced code, notably at OpenJDK, to implement it?

pfcoperez:
Hey, thanks for reviewing. I can confirm I did not look at any code outside this project, I just followed the pattern present in 

https://github.com/scala-js/scala-js/blob/a10cc63df55b599bba00fa2dec63aed1ae497649/javalib/src/main/scala/java/util/Formatter.scala#L375-L380

I stumbled upon this problem when using scala-js in https://github.com/pfcoperez/thebutlerdidit

The lack of cases for the formatted lead the app to runtime errors that I fixed at the time by writing my own formatter:

https://github.com/pfcoperez/thebutlerdidit/blob/9fac8a1340c2413747c7bede65c7be4d459dec18/shared/src/main/scala/org/pfcoperez/thebutlerdidit/model/Report.scala#L8-L15

```scala
  // Method replacing `x.formatted("%x")` which seems to be buggy in Scala.js
  private def bigIntToHexStr(x: BigInt): String = {
    val digitChar = (10 to 15).zip('a' to 'z').toMap
    def digits(x: BigInt, acc: List[Int]): List[Int] =
      if (x == 0) acc
      else digits(x / 16, (x % 16).toInt :: acc)
    "0x" + digits(x, Nil).map(d => s"${digitChar.getOrElse(d, d)}").mkString
  }
```

Then, reading scala-js code for `BigInteger#toString(b)` I found this functionality was already there, just not used in the `Formatter`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/math/BigIntegerTest.scala
```diff
@@ -163,4 +163,26 @@ class BigIntegerTest {
     assertTrue(Arrays.equals(eBytesSignum, expSignum.toByteArray))
     assertTrue(Arrays.equals(exp.toByteArray, expSignum.toByteArray))
   }
+
+  @Test def `should be correctly formatted as a decimal string`(): Unit = {
```
catap, line 167: I guess `test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala` is the right place for this tests.

Am I wrong?

pfcoperez: Thanks for the suggestion @catap , I've moved them in 8558942b216aa5bbfac18361a20bd39faf8d44a7

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -668,6 +668,27 @@ class FormatterTest {
         1, 1)
     expectFormatterThrows(classOf[MissingFormatArgumentException], "%10$d", 1)
   }
+
+  @Test def `BigInteger should be correctly formatted as a decimal string`(): Unit = {
+    val b = BigInt(new BigInteger("42"))
+    val result = b.formatted("%d")
```
sjrd, line 674: The tests should be inserted in the existing `should_provide_d_conversion` and friends, using the existing infrastructure (`assertF`) directly on `BigInteger`.

Also we should have some more tests:
* for negative numbers (and especially for negative numbers in octal and hexadecimal printing)
* with a few options, notably zero padding (`0`) and grouping (`,`)

You might also want to add a test in `FormatterTest.testFormatFrench()` and `testFormatHindiWithDevanagariDigits()`, to make sure that locale-sensitivity applies to formatted `BigInteger`s.

sjrd (changes requested):
Thank you :)

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -360,6 +361,8 @@ final class Formatter private (private[this] var dest: Appendable,
             formatNumericString(localeInfo, flags, width, arg.toString())
           case arg: Long =>
             formatNumericString(localeInfo, flags, width, arg.toString())
+          case arg: BigInteger =>
```
gzm0, line 364: @sjrd is the optimizer able to remove this if `BigInteger` is unused?

sjrd: No, but it will only reach the instance tests. Therefore something like
```js
if ($is_jl_BigInteger(arg)) {
  result = arg.toString__T()
}
```
will be generated. But the call to `toString__T` will not actually reach `BigInteger.toString`, since there is no instance of `BigInteger`. Therefore, no additional code will be generated, beside those useless 3 lines.

gzm0: But do we even define `$is_jl_BigInteger`? `BigInteger` is a class, so IIUC, we'd just do that inline: `arg instanceof $c_jl_BigInteger`. And if the class isn't defined, the `Emitter` will fold this into `false`, no?

sjrd: Ah yes right, so we'll end up with
```js
if (false) {
  result = arg.toString__T()
}
```
(I think)

If the optimizer were taught how to get rid of impossible `IsInstanceOf` nodes like that, it would entirely eliminate the `if`.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -379,6 +382,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 java.lang.Long.toOctalString(arg))
+          case arg: BigInteger =>
+            padAndSendToDest(RootLocaleInfo, flags, width, prefix,
+              applyNumberUpperCase(flags, arg.toString(8)))
```
sjrd, line 387:
```suggestion
                applyNumberUpperCase(flags, arg.toString(8)))
```

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -400,6 +406,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 applyNumberUpperCase(flags, java.lang.Long.toHexString(arg)))
+          case arg: BigInteger =>
+            padAndSendToDest(RootLocaleInfo, flags, width, prefix,
+              applyNumberUpperCase(flags, arg.toString(16)))
```
sjrd, line 411:
```suggestion
                applyNumberUpperCase(flags, arg.toString(16)))
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -12,12 +12,12 @@
...
 package org.scalajs.testsuite.javalib.util
 
+import java.math.BigInteger
+
 import org.junit.Assert._
 import org.junit.Test
-
 import org.scalajs.testsuite.utils.AssertThrows._
 import org.scalajs.testsuite.utils.Platform._
-
```
sjrd, line 20: Please keep the blank lines that were there before.

sjrd (changes requested):
Thank you for the updates. I only have code formatting comments.

Could you also squash all the commits into one with a clean commit message (for example, the message of the first commit)?

After that it should be good to merge.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
```diff
@@ -321,11 +331,16 @@ class FormatterTest {
...
     assertF("10", "%o", 8)
+    assertF("52", "%o", new BigInteger("42"))
 
     assertF("00020", "%05o", 16)
     assertF("37777777766", "%5o", -10)
     assertF("37777777766", "%05o", -10)
 
+    assertF("00020", "%05o", new BigInteger("16"))
+    assertF("  -12", "%5o", new BigInteger("-10"))
+    assertF("-0012", "%05o", new BigInteger("-10"))
```
sjrd, line 342: Hum, apparently this test does not pass: we get `00-12` instead of the correct `-0012`. Some something should be fixed.

I was also wondering why the hexadecimal passed whereas this one did not pass, which made me realize that the hex test does not have a test for negative numbers with `%05x`. So we should probably add one like that as well.

sjrd (changes requested):

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -383,8 +383,8 @@ final class Formatter private (private[this] var dest: Appendable,
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 java.lang.Long.toOctalString(arg))
           case arg: BigInteger =>
-            padAndSendToDest(RootLocaleInfo, flags, width, prefix,
-              applyNumberUpperCase(flags, arg.toString(8)))
+            formatNumericString(RootLocaleInfo, flags, width,
```
sjrd, line 386: If we're attempting to use `formatNumericString` for octal and hexadecimal, we must add a test in `FormatterTestEx` in the method that tests devanagari digits. I am pretty sure things are going to go wrong there, because the localization algorithm will be applied, but it should not.

If `formatNumericString` doesn't do the right thing out of the box, then we shouldn't use it. The whole point of that method is to apply the localization algorithm. We may need something else, closer to the bare `padAndSendToDest`, but which takes care of the `-` prefix.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -598,13 +598,13 @@ final class Formatter private (private[this] var dest: Appendable,
   }
 
   private def formatNumericString(localeInfo: LocaleInfo, flags: Flags,
-      width: Int, str: String): Unit = {
+      width: Int, str: String, maybeBasePrefix: Option[String] = None): Unit = {
```
sjrd, line 601: If we keep using `formatNumericString`, this parameter could be of type `String` with a default value of `""`. This would avoid the boxing in the Option.

(Also we try to avoid Scala stuff in the javalib.)

sjrd (changes requested):

pfcoperez:
Thanks for the continuous review @sjrd , I've added French and Hindi localized test cases in 25c868e48c0653cfeeb02bc22bf9a0cacf6807a2 .

I kind of expected it to work because I passed `RootLocaleInfo` to `formatNumericString`. I was worried about the calls to `localeInfo.localizeNumber` that some branches of the if-condition use but in that instance:

https://github.com/scala-js/scala-js/blob/317282ed451cc5c6dc986c81b0af061227f7be1a/javalib/src/main/scala/java/util/Formatter.scala#L830

`localizeNumber` is the identity function.

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -379,6 +382,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 java.lang.Long.toOctalString(arg))
+          case arg: BigInteger =>
+            formatNumericString(RootLocaleInfo, flags, width,
+              arg.toString(8), prefix)
```
sjrd, line 387: We use 4 spaces of indentation for continuation lines:
```suggestion
            formatNumericString(RootLocaleInfo, flags, width,
                arg.toString(8), prefix)
```

javalib/src/main/scala/java/util/Formatter.scala
```diff
@@ -400,6 +406,9 @@ final class Formatter private (private[this] var dest: Appendable,
           case arg: Long =>
             padAndSendToDest(RootLocaleInfo, flags, width, prefix,
                 applyNumberUpperCase(flags, java.lang.Long.toHexString(arg)))
+          case arg: BigInteger =>
+            formatNumericString(RootLocaleInfo, flags, width,
+              arg.toString(16), prefix)
```
sjrd, line 411:
```suggestion
            formatNumericString(RootLocaleInfo, flags, width,
                arg.toString(16), prefix)
```

test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTestEx.scala
```diff
@@ -75,6 +76,17 @@ class FormatterTestEx {
     assertF(French, "1\u00A0234\u00A0567", "%,d", 1234567)
     assertF(French, "1\u00A0234\u00A0567,89", "%,.2f", 1234567.89)
     assertF(French, "0012", "%04d", 12)
+    assertF(French, "0012", "%04d", new BigInteger("12"))
+
+    assertF(French, "0014", "%04o", new BigInteger("12"))
+    assertF(French, "-014", "%04o", new BigInteger("-12"))
+    assertF(HindiWithDevanagariDigits, "14", "%o", new BigInteger("12"))
+    assertF(HindiWithDevanagariDigits, "-14", "%o", new BigInteger("-12"))
```
sjrd, line 84:
```suggestion
    assertF(French, "14", "%o", new BigInteger("12"))
    assertF(French, "-14", "%o", new BigInteger("-12"))
```

test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTestEx.scala
```diff
@@ -75,6 +76,17 @@ class FormatterTestEx {
...
+
+    assertF(French, "0x0c", "%#04x", new BigInteger("12"))
+    assertF(French, "-0xc", "%#04x", new BigInteger("-12"))
+    assertF(HindiWithDevanagariDigits, "12", "%x", new BigInteger("18"))
+    assertF(HindiWithDevanagariDigits, "-12", "%x", new BigInteger("-18"))
```
sjrd, line 89:
```suggestion
    assertF(French, "12", "%x", new BigInteger("18"))
    assertF(French, "-12", "%x", new BigInteger("-18"))
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/FormatterTest.scala
sjrd: Bump the above comment, since it wasn't addressed yet.

sjrd (changes requested):
Thank you for the updates.

This basically looks good to me, now. I have left a few mostly-formatting comments. In addition to that, could you squash all the commits into a single commit with a clean commit message?

With that it should be good to merge.

pfcoperez:
Format suggestions applied and commits squashed :slightly_smiling_face:

sjrd:
As you can see, the commit message doesn't look good on GitHub. The first line should be a title that fits in the allocated width, and the rest should in the "body" of the message (separated by a blank line from the title). This is standard git practices.

Would you mind inserting a blank line between the first and second sentences of your commit message? (And ideally, wrap the body at around 70 characters of width, which is the standard git commit message width that renders well.) Then we'll be able to merge.

catap:
@sjrd, @pfcoperez just an idea: maybe add to test some big number that is bigger than Long?

pfcoperez:
> Would you mind inserting a blank line between the first and second sentences of your commit message? (And ideally, wrap the body at around 70 characters of width, which is the standard git commit message width that renders well.) Then we'll be able to merge.

@sjrd Sure, done.

> @sjrd, @pfcoperez just an idea: maybe add to test some big number that is bigger than Long?

Good idea, I've added that to the same commit for all three `d`,`o` and `x`. e.g:

```scala
assertF("0x257b117723b71f4b1", "%#x", new BigInteger("43212345678987654321"))
```

sjrd (approved):
LGTM! Thank you @pfcoperez for your patience with the many rounds of reviews. I know it can be a bit daunting when starting to contribute to this repo.

sjrd:
Hum, the CI is doing some weird stuff. I suspect it may be related to the fact that this PR is crossing the version 1.3.0 boundary.

Could you release on top of the latest master to see if that fixes the issue?

PR #4260, opened 2020-10-19
Fix #4037: Format sources with scalafmt

### LINKED ISSUE #4037: Enable scalafmt on this project
As the scala-native will learn more from this project, and the scala-native current has scalafmt 2.4.2 enabled, let this project formatted with scalafmt will help for code sharing.
refs:https://github.com/scala-js/scala-js/pull/2701
refs:https://github.com/scala-native/scala-native/issues/1787

MaximeKjaer:
As discussed in #4037, as of Git 2.23, it is possible to ignore certain commits in a `git blame` (see [this article](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame)). This effectively removes the main argument against automatic formatting, which is that it ruins `git blame`. This PR therefore introduces scalafmt to format all sources.

As the current coding style leaves some room for the code author to use some arbitrary judgment, the tool of course generates a very large diff, as it applies rules uniformly. I would argue that this is perhaps more of a good thing than a bad one.

However, I still tried to keep the formatting close to the original rules, e.g. with a line width at 80 characters, 4 space indentation for call site continuations, etc (most of these are encoded in the [`Scala.js` scalafmt preset](https://scalameta.org/scalafmt/docs/configuration.html#top-level-presets)). However, a few of the current coding style rules cannot be maintained when formatting with scalafmt; an incomplete list is below:

- scalafmt does not support configuring spacing before `:` in `_: Foo | _: Bar`
- scalafmt does not support `newlines.implicitParamListModifierPrefer = before` at the same time as `binpack.preset = true`, so I had to choose between binpacking and preferring line skips before the `implicit` keyword in parameter list rather than after; I prioritized binpacking.

If you see anything really ugly in the formatted source code, please let me know and I will try to see if I can reconfigure scalafmt to deal with it better.

I am also including a few other links to get the discussion started:

- GitHub does not yet support  `.git-blame-ignore-revs` on their Web interface, but there is a [community request](https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256) for it.
- The `.git-blame-ignore-revs` approach is what is recommended by the (quite popular) Python formatter, [black](https://github.com/psf/black), for [migrating coding styles without ruining git blame](https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame).
- An example of a large project using `.git-blame-ignore-revs` is [locust](https://github.com/locustio/locust)
- Configuring Git with `.git-blame-ignore-revs` works both on the command line, and in VS Code (with no extra configuration; I am using [GitLens](https://gitlens.amod.io/) for viewing `git blame` in the editor)

I still need to remove sections in the coding style that are no longer true, or no longer relevant because they don't need to be enforced manually anymore. However, I thought I would still open the PR as is now, to get the discussion started on this change.


junit-runtime/src/main/scala/org/hamcrest/CoreMatchers.scala
```diff
@@ -1,117 +1 @@
-/*
- * Ported from https://github.com/hamcrest/JavaHamcrest/
- */
-package org.hamcrest
 
-import org.hamcrest.core._
-
-object CoreMatchers {
```
gzm0, line 8: What happened here?

MaximeKjaer: Not quite sure of it. This happened after I added a top-level `scalafmtAll` task key and ran it. My suspicion is that this may be the result of a bug in the scalafmt sbt plugin: if multiple sbt subprojects depend on the same files, it's possible that different scalafmt threads overwrite each other, and that this is what we're seeing here.

I'll go ask in the scalafmt gitter, but if it indeed is a bug with scalafmt, or if scalafmt doesn't support this style of top-level tasks, I think we may have to delay this PR a little.

hepin1989:
Will you organize the *import*s too?

MaximeKjaer:
I could add the [`SortImports` scalafmt rewrite rule](https://scalameta.org/scalafmt/docs/configuration.html#sortimports), but note that this only sorts imports within curly brackets, not import statements (see scalameta/scalafmt#722). For sorting import statements, we would need to add scalafix, with the [`OrganizeImports` rule](https://github.com/liancheng/scalafix-organize-imports). So seeing that another tool is needed to properly sort imports, I think it's best to not let scalafmt format imports.

MaximeKjaer:
@gzm0 gates are now passing

project/Build.scala
```diff
@@ -606,6 +610,8 @@ object Build {
         }
       },
 
+      concurrentRestrictions += Tags.limit(org.scalafmt.sbt.ConcurrentRestrictionTags.Scalafmt, 4),
```
MaximeKjaer, line 613: Limiting the parallelism of scalafmt seems to solve the problem of race conditions that previously led to a file being deleted. However, I'm a little concerned that this may only be mitigating the race condition instead of solving it. I think the solution would be to have the `scalafmt` commands in each subproject depend on the command in the previous subproject; we would go through subrpojects linearly, and scalafmt can still spawn threads for each task. What do you think, is there a better way to do that?

MaximeKjaer:
To-do:

- [ ] Limit scalafmt parallelism to 1 to avoid race conditions
- [ ] Add `scalafmt` and `scalafmtCheck` top-level commands (in addition to `scalafmtAll` and `scalafmtCheckAll`)
- [ ] Disable scalafmt on scalalib
- [ ] Rebase so that every commit passes CI
- [ ] Edit coding style to reflect that scalafmt is the authoritative style guide

project/Build.scala
MaximeKjaer: I'll try to test what setting the concurrency limit to 1 does. Most likely, it solves the problem of race conditions.

MaximeKjaer: It did

javalib/src/main/scala/java/util/function/DoublePredicate.scala
```diff
@@ -22,7 +22,8 @@ trait DoublePredicate { self =>
   def and(other: DoublePredicate): DoublePredicate = {
     new DoublePredicate {
       def test(value: Double): Boolean =
-        self.test(value) && other.test(value) // the order and short-circuit are by-spec
+        self.test(value) && other.test(
+            value) // the order and short-circuit are by-spec
```
gzm0, line 26: Ugh seriously :(

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -413,12 +415,13 @@ private[emitter] object CoreJSLib {
...
           Return(Apply(genIdentBracketSelect(StringRef, "fromCharCode"),
-              (This() DOT "c") :: Nil))
+                  (This() DOT "c") :: Nil))
         })
       }
 
       if (useECMAScript2015) {
-        buf += extractWithGlobals(globalClassDef("Char", CoreVar, None, ctor :: toStr :: Nil))
+        buf += extractWithGlobals(globalClassDef("Char", CoreVar, None,
+                ctor :: toStr :: Nil))
```
gzm0, line 424: This indentation seems plain incorrect. Indenting by 4 spaces *per brace* (which I assume is happening here) seems insane :S

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -535,11 +545,12 @@ private[emitter] object CoreJSLib {
         val instance = varRef("instance")
         val v = varRef("v")
         defineFunction(name, paramList(instance), {
-          Switch(typeof(instance), List(
-              str("string") -> {
+          Switch(typeof(instance),
+              List(
+                  str("string") -> {
                 Return(constantClassResult(BoxedStringClass))
               },
```
gzm0, line 552: This also seems incorrect: The closing brace does not even have the same indentation level than the line it opens on.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -793,18 +801,20 @@ private[emitter] object CoreJSLib {
           })
         })
 
-        defineFunction("doubleToLong", paramList(x), Return {
+        defineFunction("doubleToLong", paramList(x),
+            Return {
           If(x < double(-9223372036854775808.0), { // -2^63
```
gzm0, line 806: This also cannot be correct: The nested block has *less* indentation than its parent.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -592,34 +591,31 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
                   js.Apply(js.DotSelect(genArray, js.Ident("set")),
                       List(genIndex, genRhs))
                 case CheckedBehavior.Unchecked =>
-                  js.Assign(
-                      js.BracketSelect(
-                          js.DotSelect(genArray, js.Ident("u"))(select.pos),
-                          genIndex)(select.pos),
+                  js.Assign(js.BracketSelect(js.DotSelect(genArray,
+                              js.Ident("u"))(select.pos), genIndex)(select.pos),
```
gzm0, line 595: 12 spaces indentation, wooot :P

javalanglib/src/main/scala/java/lang/StringBuilder.scala
```diff
@@ -106,8 +106,8 @@ class StringBuilder
     val oldContent = content
     if (offset < 0 || offset > oldContent.length)
       throw new StringIndexOutOfBoundsException(offset)
-    content =
-      oldContent.substring(0, offset) + str + oldContent.substring(offset)
+    content = oldContent.substring(0, offset) + str + oldContent.substring(
+        offset)
```
gzm0, line 110: Huh? Why is this happening? I have seen scalafmt do the opposite (correctly IMHO).

javalanglib/src/main/scala/java/lang/System.scala
```diff
@@ -130,55 +131,65 @@ object System {
...
-      case src: Array[Double] =>
-        dest match {
-          case dest: Array[Double] => impl(src.length, dest.length, (i, j) => dest(i) = src(j))
-          case _                   => mismatch()
-        }
-      case _ =>
-        mismatch()
-    })
+    } else
+      (src match {
```
gzm0, line 135: This is one of (many) examples where scalafmt moves things where we'd really need a block (i.e. braces).

test-adapter/src/main/scala/org/scalajs/testing/adapter/HTMLRunnerBuilder.scala
```diff
@@ -36,20 +36,24 @@ object HTMLRunnerBuilder {
       Files.createTempDirectory(outputPath.getParent(), ".html-artifacts")
 
     sys.addShutdownHook {
-      Files.walkFileTree(artifactsDir, new SimpleFileVisitor[Path] {
-        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+      Files.walkFileTree(artifactsDir,
+          new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path,
```
gzm0, line 41: Again: The nested scope is *less* indented than its parent. Is that really correct?

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -271,7 +274,7 @@ private final class Analyzer(config: CommonPhaseConfig,
         // Check there is only a single module.
         val publicModuleIDs = (
             infos.map(_.moduleID) ++
-            moduleInitializers.map(i => ModuleID(i.moduleID))
+              moduleInitializers.map(i => ModuleID(i.moduleID))
```
gzm0, line 277: This is also odd: It's inconsistent with how multiline `if`s are indented.

linker/shared/src/main/scala/org/scalajs/linker/analyzer/Analyzer.scala
```diff
@@ -280,7 +283,9 @@ private final class Analyzer(config: CommonPhaseConfig,
 
       // Check conflicts, record infos
       topLevelExportInfos = for {
-        (id @ (moduleID, exportName), infos) <- infos.groupBy(i => (i.moduleID, i.exportName))
+        (id @ (moduleID, exportName), infos) <- infos.groupBy(i =>
+          (i.moduleID, i.exportName)
```
gzm0, line 287: Why is this indented by 2 spaces? That is unexpected.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1073,7 +1071,8 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
       !settings.noForwarders && sym.isStatic && !isImplClass(sym) && {
         // Reject non-top-level objects unless opted in via the appropriate option
         scalaJSOpts.genStaticForwardersForNonTopLevelObjects ||
-        !sym.name.containsChar('$') // this is the same test that scalac performs
+        !sym.name.containsChar(
+            '$') // this is the same test that scalac performs
```
gzm0, line 1075: These comment moves are unfortunate, not sure we can do anything about them :(

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -1169,13 +1168,15 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+            js.MemberFlags.empty.withNamespace(js.MemberNamespace.PublicStatic)
           val methodIdent = encodeMethodSym(m)
           val originalName = originalNameOfMethod(m)
           val jsParams = m.tpe.params.map(genParamDef(_))
           val resultType = toIRType(m.tpe.resultType)
 
-          js.MethodDef(flags, methodIdent, originalName, jsParams, resultType, Some {
+          js.MethodDef(flags, methodIdent, originalName, jsParams, resultType,
+              Some {
             genApplyMethod(genLoadModule(moduleClass), m, jsParams.map(_.ref))
```
gzm0, line 1180: Another example of less indentation for nested stuff :(

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -5451,8 +5508,8 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
 
         wasRepeated match {
           case Some(true) =>
-            reversedArgs =
-              genPrimitiveJSRepeatedParam(arg) reverse_::: reversedArgs
+            reversedArgs = genPrimitiveJSRepeatedParam(
+                arg) reverse_::: reversedArgs
```
gzm0, line 5512: Other example of bad assignment?

CODINGSTYLE.md
```diff
@@ -3,9 +3,10 @@
 The Scala.js project has a strict policy regarding coding style.
 This is one of the cornerstones that has allowed Scala.js to maintain a clean, consistent and maintainable codebase.
 
-This document tries to document the style in use as much as possible to make it easier for everyone to contribute.
+Most coding style rules can be applied automatically by scalafmt. It is recommended that you configure your editor to reformat on save; see the [scalafmt instructions](https://scalameta.org/scalafmt/docs/installation.html) for how to configure this.
 
-A *few* of these rules are checked automatically using Scalastyle, but most of them are too complex to teach to an automated tool.
+While most rules are checked automatically by scalafmt and Scalastyle, but there are also a few rules that are too complex to teach to an automated tool.
+This document tries to document the manually enforced rules as much as possible to make it easier for everyone to contribute.
```
gzm0, line 9: I realize now that this is problematic: scalafmt's formatting on the current codebase *breaks* certain of the rules even though they were followed before (what I have seen is mostly blocks not being present). IMHO this is quite serious :-/

Is there *any* scalafmt mode that is much more aggressive and formats everything in a consistent manner (not maybe the manner we have right now)? IMO that might be a better alternative.

CODINGSTYLE.md
```diff
@@ -3,9 +3,10 @@
 The Scala.js project has a strict policy regarding coding style.
 This is one of the cornerstones that has allowed Scala.js to maintain a clean, consistent and maintainable codebase.
 
-This document tries to document the style in use as much as possible to make it easier for everyone to contribute.
+Most coding style rules can be applied automatically by scalafmt. It is recommended that you configure your editor to reformat on save; see the [scalafmt instructions](https://scalameta.org/scalafmt/docs/installation.html) for how to configure this.
```
gzm0, line 6: Consider putting all this in the "Add scalafmt" commit.

gzm0 (changes requested):
On a high level, three comments:
- A bunch of ways to cause less commits.
- Comments where it seems that scalafmt / the Scala.js style is buggy.
- Some of the reformatting breaks other rules we have. How to fix that?

sjrd:
Last time I seriously looked at scalafmt for Scala.js, I came up with the following .scalafmt.conf:
https://github.com/sjrd/scala-js/blob/scalafmt/.scalafmt.conf
Maybe some of the things in there are worth including now.

Also, maybe this could be a good time to reconsider our 80 characters line limit. We've always had 80 soft limit and 120 hard limit. But scalafmt has no concept of soft limit, and that causes all sorts of ugly things, because it will favor an ugly formatting that stays under 80 over something clean that goes a bit over the limit. Maybe we would have way less ugly formattings if we simply dropped that 80 char limit.

javalib/src/main/scala/java/util/function/DoublePredicate.scala
MaximeKjaer: This particular instance was fixed by increasing the line length limit. But in general, it's possible to fix this kind of issue manually in daily development. The thing about scalafmt is that it may not produce the prettiest results on the first try; its output depends somewhat on the whitespace in your input. For instance, suppose you start with this:

```scala
self.test(value) && other.test(value) // the order and short-circuit are by-spec
```

scalafmt may reformat to something ugly:

```scala
self.test(value) && other.test(
    value) // the order and short-circuit are by-spec
```

But if you're dissatisfied with this, you can make scalafmt break differently by manually editing to the following:


```scala
self.test(value) && 
other.test(
    value) // the order and short-circuit are by-spec
```

scalafmt will then reformat to something more correct:

```scala
self.test(value) && 
  other.test(value) // the order and short-circuit are by-spec
```

With format-on-save enabled in your IDE, this kind of thing becomes pretty seamless:

Peek 2020-11-03 18-23


Once we've settled on a stable list of config rules, I can manually go over the ugliest places and fix them.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
MaximeKjaer: That is what seems to be happening. I think this might be a bug in scalafmt, so I'll open an issue for it. But this particular case was also fixed by increasing the line length to 100.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
MaximeKjaer: For really long lines, there's nothing to be done, no. This example, however, is fixed by increasing max line length.

CODINGSTYLE.md
MaximeKjaer: Done

MaximeKjaer:
Thank you @sjrd, I've rebased with something closer to your config. It's not possible to have both binpacking (which is enabled by default in the `Scala.js` preset) and `newlines.implicitParamListModifierPrefer = before`, so I removed the latter from your config.

I also increased the line length limit to 100, which seems to work fine; in the end, most lines are around 80, and a few lines go in the 80-100 range, which nicely follows the spirit of the "soft 80, hard 120" rule. For instance, here's a screenshot of a fairly representative part of `GenJSCode.scala`, with the editor ruler set to 80 characters:

image

scalastyle-config.xml
```diff
@@ -59,23 +41,4 @@
...
-  <check level="error" enabled="true" class="org.scalastyle.scalariform.DisallowSpaceAfterTokenChecker">
-    <parameters>
-      <parameter name="tokens">LPAREN</parameter>
-    </parameters>
-  </check>
-  <check level="error" enabled="true" class="org.scalastyle.scalariform.EnsureSingleSpaceAfterTokenChecker">
-    <parameters>
-      <parameter name="tokens">IF, FOR, WHILE, DO, TRY</parameter>
-    </parameters>
-  </check>
```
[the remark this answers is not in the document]
MaximeKjaer: Done :)

scalastyle-config.xml
```diff
@@ -1,24 +1,6 @@
 <scalastyle>
   <name>Scalastyle configuration for Scala.js</name>
 
-  <check level="error" enabled="true" class="org.scalastyle.file.FileTabChecker"/>
-  <check level="error" enabled="true" class="org.scalastyle.file.WhitespaceEndOfLineChecker"/>
-  <check level="error" enabled="true" class="org.scalastyle.file.NewLineAtEofChecker"/>
-
-  <!--Indentation checker, whilst useful gives false positives on the file headers, so disabled for now-->
-  <check level="error" enabled="false" class="org.scalastyle.file.IndentationChecker"/>
```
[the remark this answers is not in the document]
MaximeKjaer: I don't know if the false positives are gone or not, but in any case there's no need for scalastyle to check indentation anymore as indentation is now scalafmt's responsibility, and is checked by `scalafmtCheckAll`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -148,6 +148,7 @@ private[emitter] object CoreJSLib {
       buf += extractWithGlobals(globalVarDef("linkingInfo", CoreVar, linkingInfo))
     }
 
+    // format: off
```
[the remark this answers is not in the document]
MaximeKjaer: Done!

MaximeKjaer:
Why did the AppVeyor build break? Checking the source files, not much has changed in the offending files: `library/src/main/scala/scala/scalajs/js/package.scala` has barely changed (just whitespace in two places), and `library/src/main/scala/scala/scalajs/concurrent/QueueExecutionContext.scala` hasn't changed at all :confused:

sjrd:
> Why did the AppVeyor build break?

That failure is very weird. I've relaunched the build just because I can't make sense of it.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
MaximeKjaer: It turns out that there already is an issue for this: scalameta/scalafmt#2079

gzm0: Is dropping binpacking up for discussion at this point? Personally, I'd rather go with a well supported formatting style than sticking to a particular other style. Especially considering that breaking blame seems to become less of a concern. @sjrd?

sjrd: I have no faith in the non-binpacking formatting, especially not for our codebase. We have many methods with many arguments, but where each argument is typically short. This goes nicely with binpacking, since most method calls can stay within 2 lines. Without binpacking, we'll waste a lot of vertical space.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
kitbellew: have you tried https://scalameta.org/scalafmt/docs/configuration.html#newlinesafterinfix?

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
kitbellew: have you tried https://scalameta.org/scalafmt/docs/configuration.html#newlinesavoidforsimpleoverflow?

CODINGSTYLE.md
kitbellew: have you tried and of the modes of https://scalameta.org/scalafmt/docs/configuration.html#newlinessource?

sjrd:
Superseded by #5286.

PR #4261, opened 2020-10-20
Assert that method name / namespace match in MethodDef

gzm0:


sjrd:
This doesn't seem to work ^^

gzm0:
Seems we're putting JS class constructors in the public namespace. Is that intended?

gzm0:
https://github.com/scala-js/scala-js/blob/v1.3.0/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala#L1874-L1879

sjrd:
That doesn't really matter, because they're redecomposed and recomposed later in `genJSClassCapturesAndConstructor`. If it helps, we can use another namespace since it's only transient within `GenJSCode`.

gzm0:
Updated. I have also consolidated some of the control flow in `genMethodWithCurrentLocalNameScope`.

sjrd (approved):

PR #4262, opened 2020-10-20
Fix #4201: Dynamic module loading

### LINKED ISSUE #4201: Async Module Loading / Progressive App Support
Follow-up after #2681 / #4198 to support async module loading for true progressive app support

gzm0:


gzm0:
There are two things ATM I'm unsure about:

- Should we call this `loadDynamic` (or `dynamicLoad`) instead?
- How are we going to avoid indirection when trying to dynamically load a native dependency? (can we teach the optimizer that?)

ir/src/main/scala/org/scalajs/ir/Printers.scala
```diff
@@ -335,6 +335,14 @@ object Printers {
           print(method)
           printArgs(args)
 
+        case ApplyDynamicImport(flags, className, method, args) =>
+          print("dynamicImport ")
+          print(className)
+          print("::")
+          print(flags)
+          print(method)
+          printArgs(args)
```
gzm0, line 344: Needs test.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -132,15 +132,60 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
     }
   }
 
+  /** Call what is stored in the given global var via dynamic import. */
+  def callDynamicGlobalVar[T: Scope](field: String, scope: T, args: List[Tree])(
```
gzm0, line 136: Not super happy about having to bake the call into here. Unsure what a better option is.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
```diff
@@ -97,28 +109,121 @@ private object MaxModuleAnalyzer {
...
+   *  `c`, it is necessarily already loaded.
+   *
+   *  Once we have obtained this minimal set of paths, we use the last
+   *  elements of the paths to determine the final module
+   *  grouping. This is because these have an actual static dependency
+   *  on the node in question.
+   *
+   *  Merging these tags into a single `ModuleID` is delegated to the
+   *  caller.
+   */
```
gzm0, line 143: @sjrd PTAL at this. Is it convincing enough that the algorithm is correct?

linker/shared/src/main/scala/org/scalajs/linker/standard/LinkedClass.scala
```diff
@@ -54,6 +54,7 @@ final class LinkedClass(
 
     val staticDependencies: Set[ClassName],
     val externalDependencies: Set[String],
+    val dynamicDependencies: Set[ClassName],
```
gzm0, line 57: Group these three into a class?

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
```diff
@@ -97,28 +107,123 @@ private object MaxModuleAnalyzer {
...
-        tag(className, moduleID)
+        tag(className, moduleID, Nil)
+      }
+    }
+  }
+
+  /** Set of shortest, mutually prefix-free paths.
+   *
+   *  Would like to make this AnyVal, but that throws Erasure on 2.11 in an
+   *  infinite loop.
```
sjrd, line 181: I don't think it would be beneficial anyway, since they are put in the values of maps, so they would be boxed there anyway.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
```diff
@@ -97,28 +107,123 @@ private object MaxModuleAnalyzer {
...
+   */
+  private final class Paths[H, T] private (
+      private val content: mutable.Map[H, Paths[T, T]]) {
```
sjrd, line 184: While this recursive `Paths[H, T]` that contains `Paths[T, T]` is quite clever, I am not sure it holds its weight against two separate classes `HeadPaths` and `TailPaths` with hard-coded `ModuleID` and `ClassName`.
* The `ends` method could be implemented in a more efficient way by not creating a builder at each recursion step (the `TailPaths` class' equivalent of `ends` could take the `Builder[ClassName]` as parameter rather than creating a new one each time)
* `isEmpty` is only necessary for `TailPaths`

`put` would have to be duplicated, though.

I am not sure. Just putting this out there for consideration.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -132,15 +132,63 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
     }
   }
 
+  /** Call what is stored in the given global var via dynamic import. */
```
sjrd, line 135: It looks like this comment is not completely accurate anymore?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2112,6 +2121,13 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
               method,
               transformTypedArgs(method.name, args))
 
+        case ApplyDynamicImport(flags, className, method, args) =>
+          val call = withDynamicGlobalVar("s", (className, method.name)) { v =>
+            js.Apply(v, transformTypedArgs(method.name, args))
```
sjrd, line 2126: This translation is a bit too naive. It only works if all the `args` are *pure* expressions (side-effect-free is not even enough). If they're not, we need to make sure to always evaluate the `args` *before* performing the `ImportCall`. The semantics of `ApplyDynamicImport(..., method, args)` should be:

1. Evaluate all the `args` as `argsValues`
2. Asynchronously load the module and execute `method(...argsValues)`

sjrd (changes requested):
Overall this looks very nice. It looks simpler than I expected :)

sjrd:
> * Should we call this `loadDynamic` (or `dynamicLoad`) instead?

Naming is hard. In the doc you wrote that it's a "dynamic import boundary". Perhaps we should call it `dynamicImportBoundary`? (although that's a mouthful).

I wonder if there is precedent for this concept in other languages, that we could take inspiration from?

> * How are we going to avoid indirection when trying to dynamically load a native dependency? (can we teach the optimizer that?)

I don't think I understand the question. :s Do you have an example?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/VarGen.scala
```diff
@@ -132,15 +132,63 @@ private[emitter] final class VarGen(jsGen: JSGen, nameGen: NameGen,
...
+    foldSameModule(scope) {
+      unitPromise.map { promise =>
+        genThen(promise, body(VarRef(ident)))
+      }
+    } { moduleID =>
+      val moduleName = config.internalModulePattern(moduleID)
+
+      val moduleTree = config.moduleKind match {
+        case ModuleKind.NoModule =>
+          throw new AssertionError("multiple modules in NoModule mode")
```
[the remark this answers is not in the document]
gzm0: Well, but if we get here, that code would be incorrect. Getting here basically means that not all classes are in the same module, but we are in `NoModule` mode. (if the classes are in the same module, they previous branch would trigger).

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -375,6 +375,49 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
+
+          typer.typed {
+            atPos(tree.pos) {
+              // class $anon extends DynamicImportThunk
+              val clsDef = ClassDef(clsSym, List(
+                  // def <init>(): = super.<init>()
+                  DefDef(ctorSym, Block(gen.mkMethodCall(
+                      Super(clsSym, tpnme.EMPTY), ObjectClass.primaryConstructor, Nil, Nil))),
+                  // def apply(): Any = body
+                  DefDef(applySym, body.setType(ObjectTpe))))
```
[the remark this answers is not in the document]
gzm0: Yep, seems so. Using `AnyTpe` remove the need for this.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
```diff
@@ -52,71 +107,123 @@ private object MaxModuleAnalyzer {
...
+   *  We really only care about:
+   *
+   *  - a -> b
+   *  - d
+   *
+   *  Because if we reach the class through a path that goes through
+   *  `c`, it is necessarily already loaded.
+   *
+   *  Once we have obtained this minimal set of paths, we use the last
+   *  elements of the paths to determine the final module
```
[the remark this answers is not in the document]
gzm0: Yes. Updated the comment.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
gzm0: Yes, exactly. But isn't this exactly the example I'm making here with your `c` being "the class", your `a` being `d` and your `b` being `c`?

From the simplification we can see that `c` (your `b`) does not appear in the final grouping.

I'm not sure I know what to explain better or how based on your comment :-/

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MinModuleAnalyzer.scala
```diff
@@ -84,7 +93,22 @@ private object MinModuleAnalyzer {
...
+       * cyclic, dynamic dependencies (because we never generate top-level
+       * awaits).
+       *
+       * However, we need to make sure the dynamic dependecy is actually put into
+       * a module. For this, we schedule it to be connected later (we cannot
+       * connect it immediately, otherwise we'd mess up the stack/spanning tree
+       * state).
+       */
+      classInfo.dynamicDependencies
+        .filter(!nodes.contains(_))
```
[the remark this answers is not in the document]
gzm0: I've added a comment. PTAL.

linker/shared/src/main/scala/org/scalajs/linker/frontend/modulesplitter/MaxModuleAnalyzer.scala
gzm0: I have tried this. I couldn't find a way of doing so without duplicating `put` and the non-trivial parts of `ends`. Basically, the little logic that distinguishes empty entries and missing entries feels non trivial with a typical interface to a `Map`. If you feel nevertheless this should be simplified, I can try harder.

gzm0 (commented):
Addressed some of the easier comments. Still TODO:

- Proper de-nesting in `FunctionEmitter`.
- Suggested additional test cases.

gzm0:
> I don't think I understand the question. :s Do you have an example?

What I mean is the following. Say we have:

```scala
object Foo {
  @js.native
  @JSImport("foo.js", "bar")
}

object MyMain {
    if (condition) {
      js.dynamicImport {
        Foo.bar(4)
      }
    }
  }
}
```

Ideally, for the dynamic import we should generate:

```javascript
import("foo.js").then(m => m.bar(4))
```

Without any additional thunk / Scala.js generated module.

I'd like to make sure that this is something we can build in the future with the current abstraction.

gzm0:
Still TODO:
- Test cases
- Native optimization
- Name

gzm0:
I believe I have addressed all the comments. Re Name: Webpack calls this dynamic imports, so I think we should just keep the name.

Feedback required from @sjrd:
- Can we optimize native calls as described in https://github.com/scala-js/scala-js/pull/4262#issuecomment-721201693 (at a later point).
- Changes are ready for second review.

gzm0:
Please ignore the previous message. This is not ready for review yet. A lot of stuff fails with the local definitions.

gzm0:
The current version of this PR fails with an IR checking error:

```
file:/home/tschlatt/20p/scala-js/test-suite/js/src/test/require-multi-modules/org/scalajs/testsuite/jsinterop/SJSDynamicImportTest.scala(119:15:CreateJSClass): Invalid CreateJSClass of top-level class org.scalajs.testsuite.jsinterop.SJSDynamicImportTest$$anon$7$Local$2
```

It is genuine: For some reason, the SJS compiler emits `Local` as an abstract JS type, and not a class with an empty capture list. I'm invstigating.

gzm0:
Seems like local JS classes inside anonymous classes are broken :)

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -4996,6 +5003,31 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
           val arg = genArgs1
           js.JSImportCall(arg)
 
+        case DYNAMIC_IMPORT =>
+          assert(args.size == 1,
+              s"Expected exactly 1 argument for JS primitive $code but got " +
+              s"${args.size} at $pos")
+
+          args.head match {
+            case Block(Nil, Apply(fun @ Select(New(tpt), _), args)) =>
```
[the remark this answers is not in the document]
gzm0: What we could do is allow the block to have other statements. That would not break the user-visible semantics but allow most general purpose transforms.

gzm0:
Only open question: Will we be able to optimize like described in https://github.com/scala-js/scala-js/pull/4262#issuecomment-721201693? (I have a WIP branch: https://github.com/gzm0/scala-js/tree/optimize-loading).

gzm0:
@sjrd I realize that my last comment might not have been clear :) This is ready for review.

sjrd:
Ah OK! I'll find some time to review it in the next 1-2 days.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -5001,6 +5008,45 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+               * dynamic loading boundary.
+               */
+
+              val clsSym = tpt.symbol
+              val ctor = fun.symbol
+
+              assert(clsSym.isSubClass(DynamicImportThunkClass),
+                  s"expected subclass of DynamicImportThunk, got: $clsSym")
+              assert(ctor.isPrimaryConstructor,
+                  s"expected primary constructor, got: $ctor")
```
sjrd, line 5034: Perhaps include the current `pos` in the error messages, to ease debugging if this pops up in the wild.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -377,6 +377,52 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
+         *   new DynamicImportThunk { def apply(): Any = body }
+         * )
+         */
+        case Apply(TypeApply(fun, List(tpeArg)), List(body))
+            if fun.symbol == JSPackage_dynamicImport =>
+          val pos = tree.pos
+
+          assert(currentOwner.isTerm, s"unexpected owner: $currentOwner at $pos")
+
+          val clsSym = currentOwner.newClass(tpnme.ANON_CLASS_NAME, pos)
```
sjrd, line 392: Aren't we supposed to append a unique `$i` at this point? Perhaps not, I'm not sure.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -377,6 +377,52 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
+                  // def apply(): Any = body
+                  DefDef(applySym, newBody)))
+
+              /* runtime.DynamicImport[A]({
+               *   class $anon ...
+               *   new $anon
+               * })
+               */
+              gen.mkMethodCall(Runtime_dynamicImport,
+                  List(tpeArg.tpe),
```
sjrd, line 421: I don't suppose there is an overload that lets us reuse `tpeArg` directly, rather than extracting its `tpe` and then rewrapping it in a `TypeTree`?

sjrd (changes requested):
This looks very nice. I'm pleasantly surprised by the code in `PrepJSInterop`, which is much cleaner than I expected it to be. I only have fairly minor comments, except a few more critical things in the `FunctionEmitter`.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
gzm0: No. The trees that reach `PrepJSInterop` for anonymous classes do not have an `$i` yet.

gzm0:
I believe I have addressed all the comments. Let's see if the CI passes :)

@sjrd there is still my question in https://github.com/scala-js/scala-js/pull/4262#issuecomment-721201693. Do you think we can teach the optimizer (in a later PR) to do this? I just want to be sure there is no blocker to this that I missed, because an additional import hop might be prohibitive for some applications.

sjrd:
That's tricky. In theory, sure, nothing goes against that. I don't think the *optimizer* (as `OptimizerCore`) could do that, since it works in the frontend of the linker which doesn't have module info. But we could have a `ModuleSet` optimizer which *should* be able to do that kind of optimization.

IIUC, the situation would be:
* There is an `ApplyDynamicImport` to some method `m` in module `mod`.
* The module `mod` has exactly one static dependency `dep` that is not also a static dependency at the call site of the `ApplyDynamicImport`.

If those apply, then replace `ApplyDynamicImport` by `ImportCall(dep).then(depMembers => ...)`.

The problem though is the `...` This might contain quite a bit of code. If it's just the static method (somehow), then we can easily inline it, and replace references to `dep` so that point to `depMembers`. But if there are classes and other large portions of code in `mod`, we can't inline everything. We would have to pass `depMembers` as an additional parameter to everything in `mod`, as a constructor argument to classes, etc. And that would have an adversarial effect on performance, because we're now passing that `depMembers` dynamically, whereas the engine's resolution could have made it statically resolved.

If it's not too big and it really is just the static method itself, I think it's doable.

sjrd (approved):

sjrd:
Not merging yet, until the optimization question is resolved. But from side it's good to go.

gzm0:
I think the scope of what I think we should optimize is much narrower: I'm only concerned with *external* modules. I look at it that way: If an internal module has a dependency that we'd want to inline, then the module analyzer didn't do it's job properly (I realize that this is not the entire truth, but for simple cases is). Or another way of looking at this is that with `FewestModules`  analysis, we won't (modulo bugs) have trivial "just-hop one further" modules. However, this is not the case for native dependencies.

So in practice, I'd want the following to be optimized:

```scala
???.nativeMember                  // SelectJSNativeMember(...)
???.nativeMember(...)             // JSFunctionApply(SelectJSNativeMember(...), args)
nativeObject.member               // JSSelect(LoadJSModule(...), ...)
nativeObject.member(...)          // JSMethodApply(LoadJSModule(...), ..., args)
new NativeClass(...)              // JSNew(LoadJSConstructor(...), args)
```

IIUC the optimizer has all the information it needs to translate this into a `JSImportCall`. The tricky part are the arguments because we mustn't lift them across the loading boundary (the inverse case we have in `FunctionEmitter`). Maybe we can restrict it to trivial expressions and local references.

From a usage point of view, that would work the same than captures in `Future`s for example:

```scala
val deferredFoo = js.dynamicImport {
  nativeMember(foo())
}

val eagerFoo = {
  val f = foo()
  js.dynamicImport { nativeMember(f) }
}
```

It feels to me that such an optimization would be doable in the optimizer (and the pre-transform infrastructure might lend itself quite well to it). WDYT?

sjrd:
Ah, I see. Yes, I believe that this restricted goal can be done in the optimizer. Here is an outline of how I would see things going:

* In addition to `MethodID.shouldInline`, we could have `shouldInlineDynamicImport` that is true for the shapes that you identified (`SelectJSNativeMember`, etc.)
* When encountering an `ApplyDynamicImport`, if the target has `shouldInlineDynamicImport`, then evaluate the `args` in `PreTransform`s and get its body
* Generate the appropriate `JSImportCall` with a `.then`, and replace in the body the args and the reference to the external module.

After that, the `ApplyDynamicImport` node is gone, and the native class/object/member won't be reachable anymore.

gzm0:
Yeah, that's pretty much what I had in mind. In this case it's good to go.

There is just one thing I still wanna do, run the test suite in a browser on this. (I'll send a PR updating TESTING afterwards). Then I'll merge.

SG?

sjrd:
Sounds good :)

gzm0:
Ok, so testHtml for modules is broken for a multitude of reasons:
- It moves files, so relative import paths are broken.
- Module tests depend on Node.js modules.

I think we can nevertheless merge this and fix these things afterwards.

sjrd:
> I think we can nevertheless merge this and fix these things afterwards.

Yes, let's do that.

PR #4263, opened 2020-10-21
Use computeIfAbsent in StandardIRFileCache

gzm0:
This simplifies the implementation a bit.


sjrd:
The [JavaDoc of `ConcurrentHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-) says that it can block other updates during the execution of the function, so it is not lock-free. Our previous implementation was lock-free, at the cost of instantiating a `PersistedFiles` for nothing (but then we're now allocating the function, so ...). Is it OK too have some locks for us here?

gzm0:
`ConcurrentHashMap` does not actually guarantee lock freedom of writes. It only guarantees lock freedom of reads and non-global locking.

sjrd:
Ah, good point.

sjrd (approved):

PR #4265, opened 2020-10-23
Implement java.lang.ClassValue.

sjrd:


javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,101 @@
...
+    else
+      javaMap.remove(`type`)
+  }
+}
+
+private object ClassValue {
+  /** Sentinel used instead of `()` (`undefined`) in a JS `Map`, or instead of
+   *  `null` in a Java `HashMap`.
+   */
+  private object Sentinel
```
gzm0, line 85: I don't understand why we need this. Don't both `ju.HashMap` and `js.Map` support checking for values? (I realize `computeIfAbsent` cannot be used for this, but IMO then we should just not use it).

javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.HashMap
+
+import scala.scalajs.js
+import scala.scalajs.runtime.linkingInfo
+
+abstract class ClassValue[T] protected () {
+  import ClassValue._
+
+  private val jsMap: AnyRef = {
```
gzm0, line 23: Can't this be a `js.Dynamic`? IIUC, `js.Dynamic` would be erased to `any` in the IR, so it's not a problem if we use it. Or am I missing something?

sjrd: Yes, you're right. That works. I don't know why I thought otherwise.

javalanglib/src/main/scala/java/lang/ClassValue.scala
sjrd: They do, but it seems better to do a simple `===` against a sentinel than to do a full second lookup.

javalanglib/src/main/scala/java/lang/ClassValue.scala
gzm0: Actually, looking at the IR cleaner: It seems good enough that you can actually use the full facade of `js.Map` here, no?

javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,99 @@
...
+   * implemented in terms of the former anyway, to a HashMap is leaner and
+   * faster.
+   */
+  private val javaMap: HashMap[Class[_], Any] =
+    if (useJSMap) null
+    else new HashMap()
+
+  protected def computeValue(`type`: Class[_]): T
+
+  def get(`type`: Class[T]): T = {
```
gzm0, line 49: IIUC, this signature is wrong, it should be `Class[_]`.

javalanglib/src/main/scala/java/lang/ClassValue.scala
gzm0: Better for speed? Maybe. But not better for codesize (IIUC) or readability, so I'm not 100% sure we're making the right tradeoff here :-/

javalanglib/src/main/scala/java/lang/ClassValue.scala
gzm0: Ok, seems we don't have a public facade for map, so we indeed cannot use it. Consider making a private one though. It might avoid a lot of casts (depends a bit on the outcome of the `Sentinel` discussion as well, because with the sentinel, we cannot type it properly anyways).

javalanglib/src/main/scala/java/lang/ClassValue.scala
sjrd: All right. I switched a different implementation that does not use a sentinel.

javalanglib/src/main/scala/java/lang/ClassValue.scala
sjrd: I had forgotten how smart our IR cleaner actually is. I was convinced that declaring any facade inside javalanglib was forbidden, but in fact it is capable of erasing them.

We have `js.Map`, but most methods are hidden in favor of the Scala collections methods. I have added direct accessors in `Utils`.

javalanglib/src/main/scala/java/lang/ClassValue.scala
sjrd: Oops, good catch. Fixed.

sjrd:
Updated. Note the new first commit, which was prompted by failures on the CI when doing `delete global.Map`.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/ObjectTest.scala
```diff
@@ -97,6 +97,9 @@ class ObjectTest {
   }
 
   @Test def fromEntries_js_Map(): Unit = {
+    assumeTrue("requires js.Map",
+        js.typeOf(js.constructorOf[js.Map[_, _]]) != "undefined")
```
gzm0, line 101: Put a forwarder for this in `org.scalajs.testsuite.utils.Platform` and force it to be true if `assumeES2015` is true?

Jenkinsfile
```diff
@@ -234,30 +234,30 @@ def Tasks = [
     setJavaVersion $java
     npm install &&
     sbtretry 'set scalaJSLinkerConfig in $testSuite.v$v ~= (_.withESFeatures(_.withUseECMAScript2015(false)))' \
-        'set jsEnv in $testSuite.v$v := new NodeJSEnvForcePolyfills()' \
+        'set Seq(jsEnv in $testSuite.v$v := new NodeJSEnvForcePolyfills(), MyScalaJSPlugin.wantSourceMaps in $testSuite.v$v := false)' \
         ++$scala $testSuite$v/test &&
     sbtretry 'set scalaJSLinkerConfig in $testSuite.v$v ~= (_.withESFeatures(_.withUseECMAScript2015(false)))' \
-        'set jsEnv in $testSuite.v$v := new NodeJSEnvForcePolyfills()' \
+        'set Seq(jsEnv in $testSuite.v$v := new NodeJSEnvForcePolyfills(), MyScalaJSPlugin.wantSourceMaps in $testSuite.v$v := false)' \
```
gzm0, line 240: That's kind of a new style. Is it different than multiple `set` calls?

(FWIW: I think this needs some love. I have been thinking of introducing a `sjsTestMode` or something and make all the settings dependent on it. This would also make it easier for us to test one of the things. At least I'm tired of copying things out of the `Jenkinsfile` :P).

javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+import scala.scalajs.runtime.linkingInfo
+
+import Utils._
+
+abstract class ClassValue[T] protected () {
+  private val jsMap: js.Map[Class[_], T] = {
+    if (linkingInfo.assumingES6 || js.typeOf(js.Dynamic.global.Map) != "undefined")
```
gzm0, line 25: Use `js.constructorOf[js.Map]` like in the previous commit?

gzm0 (approved):
Just minor stuff.

javalanglib/src/main/scala/java/lang/ClassValue.scala
sjrd: That didn't work, actually. It threw a `ReferenceError`.

Jenkinsfile
sjrd: It's the same as multiple `set` calls. I chose this style in this specific case because the `jsEnv` setting must be accompanied by the `wantSourceMaps` setting. It makes it a bit clearer that they must not be separated that way.

I also agree that a general test mode would be nice, but that is probably for a different PR.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/ObjectTest.scala
sjrd: Done.

sjrd (commented):
Updated.

It turns out we cannot delete `Set` because some internal Node.js code relies on it. I updated the first commit.

gzm0 (approved):

sjrd:
Well, that doesn't pass the CI either, because now `Symbol` is required somewhere else. Apparently deleting `Map` triggers a different code path that requires `Set` and `Symbol`.

I was able to replicate the issue locally with Node.js 13.3.0 (the version of node installed on the CI), but it works fine with v14.2.0 (which I had locally installed on my machine) as well as v15.0.1 (the latest version).

Perhaps we should simply upgrade to Node.js 15.0.1?

gzm0:
> Perhaps we should simply upgrade to Node.js 15.0.1?

It seems that our strategy on how to test ES 5.1 compatibility is slowly falling apart. I don't really have any ideas on how to deal with this. A quick google search didn't bring up anything meaningful :(

sjrd:
Yes, it's becoming increasingly difficult to test ES 5.1 support ... perhaps that also means that ES 5.1 compliance is becoming increasingly irrelevant ... 🤔

gzm0:
Well OTOH, using ES 6 language constructs seems to still be slow on FF (if one believes the chat on gitter). Unsure how this relates to ES6 library features / types thought :-/

sjrd:
I don't think there's any likely correlation between language features and library functions. Library functions are usually written in JavaScript anyway (even the built-in ones) and left to be optimized by the general purpose optimizer. Language features necessarily require compiler and optimizer work, unless they are straightforward sugar for other constructs.

Anyway, the performance issue is independent from the compliance issue. We can very well test performance of old constructs on a state of the art engine that supports all of the new stuff. However it's difficult to test compliance of older constructs on an engine that supports all the new stuff.

sjrd:
I asked Fabien to upgrade the CI to Node.js v14.15.0. That version became the new LTS just yesterday. It appears to survive the removal of both `Map` and `Set`, so I updated the PR one more time with re-enabling the deletion of `Set`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassValueTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import org.junit.Test
+import org.junit.Assert._
+
+class ClassValueTest {
+  @Test def testClassValue(): Unit = {
+    val classValue = new ClassValue[Any] {
+      private var counter: Int = 0
+
+      // () and null are corner-cases of our implementation
```
gzm0, line 23: Just realized that this is not true anymore.

sjrd: Hum, yes, it's still true, since we need the addition `has`/`containsKey` tests for them, and they are explicitly mentioned in the code.

javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,90 @@
...
+     * `has` to disambiguate a present undefined/null from an absent key.
+     * Since the purpose of ClassValue is to be used a cache indexed by Class
+     * values, the expected use case will have more hits than misses, and so
+     * this ordering should be faster on average than first performing `has`
+     * then `get`.
+     */
+    if (useJSMap) {
+      undefOrGetOrElseCompute(mapGet(jsMap, `type`)) { () =>
+        if (mapHas(jsMap, `type`)) {
+          ().asInstanceOf[T]
```
sjrd, line 61: Here we have a special-case for `()`.

javalanglib/src/main/scala/java/lang/ClassValue.scala
```diff
@@ -0,0 +1,90 @@
...
+          val newValue = computeValue(`type`)
+          mapSet(jsMap, `type`, newValue)
+          newValue
+        }
+      }
+    } else {
+      javaMap.get(`type`) match {
+        case null =>
+          if (javaMap.containsKey(`type`)) {
+            null.asInstanceOf[T]
```
sjrd, line 72: And here we have a special-case for `null`.

gzm0 (approved):
Hahahaha, fair enough :P

PR #4269, opened 2020-10-27
Fix #4268: Record native super classes as external dependency

### LINKED ISSUE #4268: Native superclass of a non-native JS class is not imported (1.3.0 regression)
Failing test case for `ModulesTest.scala` (requires module support, for example CommonJS module):
```scala
class ModulesTest {
  ...

  @Test def testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass_issue4267(): Unit = {
    val myAssertionError = new MyAssertionError(6, 5, "myOperator")
    assertEquals(6, myAssertionError.expected)
    assertEquals(5, myAssertionError.actual)
    assertEquals("myOperator", myAssertionError.operator)
  }
}

object ModulesTest {
  ...

  /* #4268 Test that a super-class only used in an extends from a non-native JS
   * class *is* imported.
   */

  @js.native
  @JSImport("assert", "AssertionError")
  class NodeAssertionError(options: js.Object) extends js.Error {
    val expected: Any = js.native
    val actual: Any = js.native
    val operator: String = js.native
  }

  class MyAssertionError(expected: js.Any, actual: js.Any, operator: String)
      extends NodeAssertionError(js.Dynamic.literal(
          expected = expected, actual = actual, operator = operator))
}
```
Causes:
```
[debug] Test org.scalajs.testsuite.jsinterop.ModulesTest.testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass started
[error] Test org.scalajs.testsuite.jsinterop.ModulesTest.testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass failed: scala.scalajs.js.JavaScriptException: ReferenceError: $i_assert is not defined, took 0.000999936 sec
[error]     at <jscode>.$a_Lorg_scalajs_testsuite_jsinterop_ModulesTest$MyAssertionError(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\test-suite\js\src\test\require-modules\org\scalajs\testsuite\jsinterop\ModulesTest.scala:220)
[error]     at org.scalajs.testsuite.jsinterop.ModulesTest.testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\test-suite\js\src\test\require-modules\org\scalajs\testsuite\jsinterop\ModulesTest.scala:119)
[error]     at org.scalajs.testsuite.jsinterop.ModulesTest$scalajs$junit$bootstrapper$.invokeTest(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\test-suite\js\src\test\require-modules\org\scalajs\testsuite\jsinterop\ModulesTest.scala:29)
[error]     at <jscode>.{anonymous}()(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\junit-runtime\src\main\scala\org\scalajs\junit\JUnitTask.scala:111)
[error]     at org.scalajs.junit.JUnitTask.executeTestMethod(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\library\src\main\scala\scala\scalajs\runtime\AnonFunctions.scala:22)
[error]     at org.scalajs.junit.JUnitTask.runTests$1(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\junit-runtime\src\main\scala\org\scalajs\junit\JUnitTask.scala:64)
[error]     at org.scalajs.junit.JUnitTask.executeTests(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\junit-runtime\src\main\scala\org\scalajs\junit\JUnitTask.scala:79)
[error]     at org.scalajs.junit.JUnitTask.execute(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\junit-runtime\src\main\scala\org\scalajs\junit\JUnitTask.scala:42)
[error]     at org.scalajs.testing.bridge.TestAdapterBridge$.executeFun(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\test-bridge\src\main\scala\org\scalajs\testing\bridge\TestAdapterBridge.scala:103)
[error]     at scala.scalajs.runtime.AnonFunction1.apply(...\scalajs\test-suite\js\.2.12\target\scala-2.12\scalajs-test-suite-test-fastopt\file:\...\scalajs\library\src\main\scala\scala\scalajs\runtime\AnonFunctions.scala:27)
[debug] Test org.scalajs.testsuite.jsinterop.ModulesTest.testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass finished, took 0.000999936 sec
```
The issue is that the `$i_assert` variable that is supposed to hold the import (the result of `require` in CommonJS) is not emitted, although it is reference in the superclass of `MyAssertionError`.

Adding
```scala
    js.constructorOf[NodeAssertionError]
```
at the start of the test case works around the issue.

This was first reported at https://github.com/scalacenter/scalajs-bundler/issues/381. According to that issue, it is a regression in 1.3.0, probably related to module splitting support.

Tentatively assigning this to @gzm0

gzm0:


sjrd (approved):
That was fast :)

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -114,6 +114,14 @@ class ModulesTest {
     new ExistentSubClass
     ExistentSubObject
   }
+
+  // #4267
+  @Test def testImportSuperClassUsedOnlyInExtendsOfNonNativeJSClass(): Unit = {
```
ekrich, line 119: Thanks for the naming 👍

sjrd (approved):

gzm0:
It seems ES5.1 tests fail with:

> Class constructor AssertionError cannot be invoked without 'new'

I assume this is because `AssertionError` is defined as a ES6 `class`. @sjrd is this expected?

sjrd:
Ah, shoot. I had not double-checked whether `AssertionError` was extensible from an ES 5.1 "class". Either we find another class in the Node.js API that we can extend for the test, or we could just `assumeTrue(assumingES6)` in that test.

sjrd:
I did some digging but could not find any class that is extensible from ES 5.1 in the Node.js APIs :(

gzm0:
I just added our own.

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
```diff
@@ -197,4 +203,16 @@ object ModulesTest {
   @js.native
   @JSImport("querystring", JSImport.Namespace)
   object ExistentSubObject extends NonExistentSuperClass
+
+  /* #4268 Test that a super-class only used in an extends from a non-native JS
+   * class *is* imported.
+   */
+  @js.native
+  @JSImport("../test-classes/module-natives.js", "ModuleTestNativeParentClass")
```
sjrd, line 211: Since we're using our own, we should use a much more specific file name. This test would become moot if we added something else `module-natives.js` that we would import somewhere else. So the file name should be very specific to this test, so that we never reuse it for any other test.

sjrd (changes requested):

test-suite/js/src/test/require-modules/org/scalajs/testsuite/jsinterop/ModulesTest.scala
gzm0: Good point!

sjrd (approved):

PR #4273, opened 2020-11-03
Fix #4271: Allow no public module in ReportToLinkerOutputAdapter

### LINKED ISSUE #4271: ScalaJS 1.3.0 fails to fastOptJS/fullOptJS when the project has no exports nor the module initializer
1. Create a project with only build.sbt and plugins.sbt

    ```
    // build.sbt
    enablePlugins(ScalaJSPlugin)
    name := "Scala.js Tutorial"
    scalaVersion := "2.13.1"

    // plugins.sbt
    addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.3.0")
    ```

2. Run `sbt fastOptJS` or `sbt fullOptJS` 
    * v1.2.0: succeeds
    * v1.3.0: fails with `org.scalajs.linker.interface.ReportToLinkerOutputAdapter$UnsupportedLinkerOutputException: Linking did not return exactly one public module.`

It also works with 1.3.0 if I add `main()` method or `@JSExportTopLevel`.

(In my case, the project is only used by dependent projects and there is no need to fastOpt/fullOpt for the project actually.)

gzm0:


linker-interface/shared/src/test/scala/org/scalajs/linker/interface/ReportToLinkerOutputAdapterTest.scala
```diff
@@ -179,6 +179,40 @@ class ReportToLinkerOutputAdapterTest {
...
+  @Test // #4271
+  def testNoPublicModules(): AsyncResult = await {
+    val writeOut = new WriteOnlyOutputDirectory()
+
+    val emptyReport = new ReportImpl(Nil)
+
+    val legacyOutput = LinkerOutput(new OutputFileImpl("js", writeOut))
+      .withSourceMap(new OutputFileImpl("sm", writeOut))
+      .withSourceMapURI(new URI("http://example.org/my-source-map-uri"))
+      .withJSFileURI(new URI("http://example.org/my-js-file-uri"))
```
sjrd, line 192: I think there should also be a test where we don't ask for a source map, since it's a different code path in the implementation.

Otherwise LGTM

sjrd (changes requested):

sjrd (approved):

PR #4275, opened 2020-11-04
Treat throw as an Lhs in FunctionEmitter

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2825,6 +2825,8 @@ private object FunctionEmitter {
       override def hasNothingType: Boolean = true
     }
 
+    case object Throw extends Lhs
```
sjrd, line 2828:
```suggestion
    case object Throw extends Lhs {
      override def hasNothingType: Boolean = true
    }
```
;)

sjrd (approved):

PR #4272, opened 2020-11-02
Address new warnings with the strict patmat exhaustivity of 2.13.4.

sjrd:
In 2.13.4, the pattern match exhaustivity analysis has become much stricter. It remains active in the presence of non-sealed classes, guards, and custom extractors. This change causes a lot of new warnings in our codebase.

See scala/scala#9140

We address most warnings with `@unchecked`, but for some we throw more specialized exceptions instead.

---

Locally tested with
```
> set resolvers in Global += "scala-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration/"
> ++2.13.4-bin-200c122
> testSuite2_13/test:compile # with manual inspection of the warnings
> testSuite2_13/test
> testSuiteEx2_13/test
> compiler2_13/test:compile
> ir2_13/test
> linker2_13/test
> testSuite2_13/bootstrap:test
```

PR #4276, opened 2020-11-05
Catch up with Scala 2.13.4

sjrd:
* Use `CharBuffer.wrap` to get a `CharSequence` from an `Array[Char]`.
* Address new warnings with the stricter patmat exhaustivity of 2.13.4.
* Adapt `MatchASTTest` for Scala 2.13.4.

---

Supersedes #4272 now that https://github.com/scala/scala/pull/9299 upstream tuned down most of the warnings. Also address two other incompatibilities with 2.13.4.

Locally tested with
```
> ++2.13.4-bin-e6a6ac6
> testSuite2_13/test:compile # with manual inspection of the warnings in the test suite
> testSuite2_13/test
> testSuiteEx2_13/test
> compiler2_13/test
> ir2_13/test
> linker2_13/test
> testSuite2_13/bootstrap:test
```

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
@@ -1652,6 +1652,8 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
             (Nil, applyCtor)
           case js.Block(prepStats :+ (applyCtor: js.ApplyStatic)) =>
             (prepStats, applyCtor)
+          case _ =>
+            abort(s"Unexpected body for JS constructor dispatch resolution at ${body.pos}:\n$body")
         }
         val js.ApplyStatic(_, _, js.MethodIdent(ctorName), js.This() :: ctorArgs) =
           applyCtor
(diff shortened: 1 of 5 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -5209,7 +5214,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
           } else if (jsInterop.isJSBracketAccess(sym)) {
             assert(argc == 1 || argc == 2,
                 s"@JSBracketAccess methods should have 1 or 2 non-varargs arguments")
-            argsNoSpread match {
+            (argsNoSpread: @unchecked) match {
```
gzm0, line 5217: Why not just put a default case throwing the error in the assertion above? That seems less complicated.

gzm0 (approved):
Only a minor nit.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
sjrd: Ah yes, that's much better of course.

PR #4277, opened 2020-11-06
Address performance issues of the ES2015 output

sjrd:
* Always emit `var`s for internal vars instead of `let`s and `const`s.
* Do not use ES `class`es for non-Throwable Scala classes.

---

The issue with `let`s and `const`s recently made some noise due to Safari (https://bugs.webkit.org/show_bug.cgi?id=199866). Although it is now fixed, the episode highlights how fragile the performance of `let`s and `const`s can be.

I have known about the problem of ES 2015 on Firefox for a long time. The performance issues are very visible in my thesis. However so far I did not have insight into what specific features of ES 2015 were causing the terrible performance. I have now identified that `class`es are to be blamed. Since nothing has improved in Firefox in all that time, I am doubtful that we can expect an improvement anytime soon. Fortunately, as the commit explains, we don't really need to emit ES classes for non-Throwable Scala classes.

Together, these fixes should level out all the performance issues we've found around the ES 2015 output, without any downside except code size.


sjrd:
Note that I've used system properties as a config mechanism. That should in theory be *somewhere* in the `LinkerConfig`, but we don't have precedent for configs that are strictly only about optimizations. Existing ESFeatures all have some observable semantics. I'm not sure what would a proper place for that kind of config.

For the `let`/`const` config, there should be no reason to change the default unless it is to investigate the performance issues in an engine or a benchmark, so it's very niche. For the `class` config, however, there is at least one legitimate use case: if you are more concerned about code size than performance, or if your code is not meant to run on Firefox (e.g., it's server-side), you might want to force the use of classes everywhere.

gzm0:
> I'm not sure what would a proper place for that kind of config.

Why not just to booleans `avoidLet` and `avoidClass` in `ESFeatures` that are `true` by default? Documentation saying that the linker may try to use them as little as possible where this does not alter semantics. I don't think using system properties is an option here (notably because they are global across different linker instances).

sjrd:
Rebased + switched to flags in `ESFeatures`, as suggested.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
```diff
@@ -26,36 +26,125 @@ import Fingerprint.FingerprintBuilder
...
+ *  - Options whose name is of the form `avoidX` *hint* at the linker to avoid
+ *    the corresponding features *when it does not affect observable
+ *    semantics*. They are related to optimizations (for performance or code
+ *    size). The linker is free to ignore those options.
  */
 final class ESFeatures private (
-    /** Whether to use ECMAScript 2015 features, such as classes and arrow
-     *  functions.
-     */
-    val useECMAScript2015: Boolean,
```
sjrd, line 34: I discovered that Scaladocs on constructor parameters are ignored. Moreover, `@param` annotations in the class' Scaladoc are ignored if the constructor is not public. Therefore, I have switched to explicit `val`s in the body of the class so that they can carry proper Scaladocs that are visible online.

gzm0: > Scaladoc are ignored if the constructor is not public.

Oh, that's why. I have been toying with this as well. Seems like a Scaladoc bug to me :-/

I think this deserves a (non-scaladoc) comment in the code, otherwise it looks quite surprising IMO.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
```diff
@@ -26,36 +26,90 @@ import Fingerprint.FingerprintBuilder
...
+   *
+   *  When `true`, the following behaviors are guaranteed:
+   *
+   *  - JavaScript classes are true `class`'es, therefore a) they can extend
+   *    native JavaScript `class`'es and b) they inherit static members from
+   *    their parent class.
+   *  - Throwable classes are proper JavaScript error classes, recognized as
+   *    such by debuggers.
+   *  - In Script (`NoModule`) mode, top-level exports are defined as `let`s
+   *    rather than `var`s, and behave as such.
```
gzm0, line 63: Are we actually testing that? (IIUC in practice this means they are not defined on the global object). If no, maybe we should?

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
```diff
@@ -71,6 +73,35 @@ final class ESFeatures private (
    */
   val allowBigIntsForLongs = _allowBigIntsForLongs
 
+  /** Avoid `class`'es when using `function`s and `prototype`s has the same
+   *  observable semantics.
+   *
+   *  Default: `true`
+   *
+   *  SpiderMonkey is known to exhibit terrible performance with JavaScript
+   *  `class`'es, with up to an order of magnitude of performance degradation.
```
gzm0, line 82: I assume we don't have an easy reference for this? Otherwise add it?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -379,6 +379,23 @@ final class Emitter(config: Emitter.Config) {
...
+       *
+       * Note that `useClassesForRegularClasses` implies
+       * `useClassesForJSClassesAndThrowables`, so the short-cut is valid
+       */
+      val useESClass = if (jsGen.useClassesForRegularClasses) {
+        assert(jsGen.useClassesForJSClassesAndThrowables)
+        true
+      } else {
+        jsGen.useClassesForJSClassesAndThrowables &&
+        (kind.isJSClass || linkedClass.ancestors.contains(ThrowableClass))
```
gzm0, line 396: This check here is not consistent with `shouldExtendJSError`: it checks also if the super class of `Throwable` is `jl.Object`. In practice, this is probably unimportant. However, a comment referring to `shouldExtendJSError` and saying that this is a correct approximation (in the worst case we just don't optimize some classes) might be a good idea.

project/BinaryIncompatibilities.scala
```diff
@@ -8,6 +8,15 @@ object BinaryIncompatibilities {
   )
 
   val Linker = Seq(
+      // private[emitter], not an issue
```
gzm0, line 11: Exclude all of `backend`? (maybe we should even have these excludes constantly, it seems unnecessary to maintain these lists).

gzm0 (commented):
Only nits.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
sjrd: I have added a test for that in the first commit.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
sjrd: No, I did not find any existing reference about that. But when I get more time to do it properly, I will submit an issue to SpiderMonkey.

sjrd (commented):
Updated.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
```diff
@@ -26,36 +26,92 @@ import Fingerprint.FingerprintBuilder
...
-    /** Whether to use ECMAScript 2015 features, such as classes and arrow
-     *  functions.
-     */
-    val useECMAScript2015: Boolean,
-    /** EXPERIMENTAL: Whether to allow using ECMAScript `BigInt`s to implement
-     *  `Long`s.
-     */
-    val allowBigIntsForLongs: Boolean
+    // We define `val`s separately below so that we can attach Scaladoc to them
+    // (putting Scaladoc comments on constructor param `val`s has no effect)
```
gzm0, line 36: Wrong comment style? I thought multi-line comments need `/* ... */`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala
```diff
@@ -379,6 +379,31 @@ final class Emitter(config: Emitter.Config) {
...
+       *
+       * See JSGen.useClassesForRegularClasses for the rationale here.
+       * Accessing `ancestors` without cache invalidation is fine because it
+       * is part of the identity of a class for its cache in the first place.
+       *
+       * Note that `useClassesForRegularClasses` implies
+       * `useClassesForJSClassesAndThrowables`, so the short-cut is valid.
+       *
+       * Compared to `ClassEmitter.shouldExtendJSError`, which is used below,
+       * we do not check here that `Throwable` directly extend `Object`. If
```
gzm0, line 392: extends

gzm0 (approved):
Only typos / nits.

linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala
sjrd: Yes, you're right. Fixed.

PR #4279, opened 2020-11-09
Fix #4278: Recognize JS classes nested inside anonymous classes

### LINKED ISSUE #4278: Local JS classes inside anonymous classes cause IR checking errors
Take the following:

```scala
val supplier = new Supplier[js.Dynamic]{
  def get(): js.Dynamic = {
    class LocalInsideAnonymousClass extends js.Object {
      def x: Int = 1
    }

    new LocalInsideAnonymousClass().asInstanceOf[js.Dynamic]
  }
}

val obj = supplier.get()

assertEquals(1, obj.x)
```

Causes:

```
file:/home/tschlatt/20p/scala-js/test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/NestedJSClassTest.scala(517:15:CreateJSClass): Invalid CreateJSClass of top-level class org.scalajs.testsuite.jsinterop.NestedJSClassTest$$anon$1$LocalInsideAnonymousClass$1
```

I suspect that the local class wrongly gets identified as anonymous class by later compiler phases because some checks are name based.

gzm0:


test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/NestedJSClassTest.scala
```diff
@@ -508,6 +510,38 @@ class NestedJSClassTest {
...
+        }
+
+        new LocalInsideAnonymousClass().asInstanceOf[js.Dynamic]
+      }
+    }
+
+    val obj = supplier.get()
+
+    assertEquals(1, obj.x)
+  }
```
gzm0, line 544: Do we need additional semantic checks on these? (identity of the class?)

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -6605,6 +6605,23 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
   def isStaticModule(sym: Symbol): Boolean =
     sym.isModuleClass && !isImplClass(sym) && !sym.isLifted
 
+  def isAnonymousClass(sym: Symbol): Boolean = {
+    /* sym.isAnonymousClass simply checks if
+     * `name containsName tpnme.ANON_CLASS_NAME`
+     * so after flatten (which we are) it identifies classes nested inside
+     * anonymous classes as anonymous (n.b. local classes, see #4278).
```
sjrd, line 6612: Was that supposed to be `e.g.`:
```suggestion
     * anonymous classes as anonymous (e.g., local classes, see #4278).
```
?

test-suite/js/src/test/scala/org/scalajs/testsuite/jsinterop/NestedJSClassTest.scala
sjrd: Not necessarily the identity of the class, but we should definitely check that `obj` has a class other than `js.Object`, and/or that it doesn't have an *own* property named `"x"`.

sjrd (changes requested):

gzm0:
Updated

sjrd (approved):

sjrd:
@gzm0 Should we get this over the finish line before v1.3.1?

gzm0:
As I mentioned in chat, the only working solution I have to this now is one that simply assumes in the backend that anything concrete extending `js.FunctionN` is a lambda. I'm not sure that is an assumption we can make. If we cannot, it might require much more name unmangling or even marker annotations :-/

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -6605,6 +6605,17 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
 
+  def isAnonymousClass(sym: Symbol): Boolean = {
+    /* sym.isAnonymousClass simply checks if
+     * `name containsName tpnme.ANON_CLASS_NAME`
+     * so after flatten (which we are) it identifies classes nested inside
+     * anonymous classes as anonymous (notably local classes, see #4278).
+     *
+     * Instead we apply the check on the sym's originalName.
+     */
+    sym.originalName.containsName(tpnme.ANON_CLASS_NAME)
```
sjrd, line 6616: Ach, it's supposed to be `unexpandedName` (`originalName` is deprecated since 2.11.0):
```suggestion
    sym.unexpandedName.containsName(tpnme.ANON_CLASS_NAME)
```

gzm0:
This does not seem to work. (The IR cleaner errors IIUC are a symptom of the scala lambda optimization not working properly). I suspect the `unexpandedName` of anonymous stuff is the empty name :-/

gzm0:
It seems (on 2.12) the `unexpandedName` of a local class looks like: `anon$2$LocalInsideAnonymousClass$1`. That invalidates my original assumption, but I need to dig more.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -6449,8 +6449,19 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
   }
 
   /** Test whether `sym` is the symbol of a JS function definition */
-  private def isJSFunctionDef(sym: Symbol): Boolean =
-    sym.isAnonymousClass && AllJSFunctionClasses.exists(sym isSubClass _)
+  private def isJSFunctionDef(sym: Symbol): Boolean = {
+    /* A JS function may only reach us in the backend, if it originally was a
```
sjrd, line 6453:
```suggestion
    /* A JS function may only reach the backend if it originally was a
```

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -6449,8 +6449,19 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
...
+     * lambda. This is difficult to check in the backend, so we use the fact
+     * that a non-lambda, concrete, non-native JS type, cannot implement a
+     * method named `apply`.
+     *
+     * All JS function classes have an abstract member named `apply`. Therefore,
+     * a class is a JS lambda iff it is concrete, a non-native JS type and
+     * inherits from a JS function class.
+     */
+    !isJSNativeClass(sym) && !sym.isAbstract &&
+    AllJSFunctionClasses.exists(sym.isSubClass(_))
```
sjrd, line 6463: The new test means that we'll have to go through all the `AllJSFunctionClasses` to test `isSubClass(_)` for most input `sym`s. This may have a non trivial performance impact.

I think we could keep the check `sym.isAnonymousClass &&` as a first fast-path test. That would avoid the expensive tests for most classes, I believe.

compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala
```diff
@@ -743,7 +743,7 @@ abstract class GenJSCode[G <: Global with Singleton](val global: G)
     def genAnonJSClassNew(sym: Symbol, jsSuperClassValue: js.Tree,
         args: List[js.Tree])(
         implicit pos: Position): js.Tree = {
-      assert(sym.isAnonymousClass && !isJSFunctionDef(sym),
+      assert(isAnonymousJSClass(sym) && !isJSFunctionDef(sym),
```
sjrd, line 746: Isn't `&& !isJSFunctionDef(sym)` redundant, now?

If it isn't, why is there `|| isJSFunctionDef(sym)` above?

sjrd (changes requested):
Some minor comments.

gzm0:
Updated

sjrd (approved):

PR #4280, opened 2020-11-10
Build support for Scala 3 in sbt-scalajs.

sjrd:
Currently, sbt-dotty knows about Scala.js, and appears to always be applied *after* sbt-scalajs, which allows it to fix up the configuration for Scala 3.

However, most of the functionality of sbt-dotty will have to be merged in sbt itself, where we cannot bake Scala.js support.

This commit adds direct support for Scala 3 in sbt-scalajs, in a way that works today (if sbt-dotty is present) as well as future-proofs for when sbt-dotty won't be used anymore.

/cc @adpi2


sbt-plugin/src/sbt-test/scala3/basic/build.sbt
```diff
@@ -0,0 +1,8 @@
+enablePlugins(ScalaJSPlugin)
+
+scalaVersion := "3.0.0-M1"
+
+// Test withDottyCompat for %%% dependencies
+libraryDependencies += ("org.scala-js" %%% "scalajs-dom" % "1.1.0").withDottyCompat(scalaVersion.value)
```
gzm0, line 6: Can we depend on something else here (maybe a dummy library if `withDottyCompat` needs to be tested here)? I'm just a bit concerned about the dependency this introduces in the library graph. (it would be an actual problem if Scala.js were to publish a new major version for example).

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
```diff
@@ -655,26 +655,57 @@ private[sbtplugin] object ScalaJSPluginInternal {
...
+        if (isScala3(scalaV)) {
+          /* Remove scala3-library (non _sjs1) in case sbt-dotty was applied
+           * before sbt-scalajs.
+           */
+          val filteredPrev = prev.filterNot { dep =>
+            dep.organization == scalaOrg && dep.name == "scala3-library"
+          }
+          filteredPrev ++ Seq(
+              scalaOrg %% "scala3-library_sjs1" % scalaV,
+              "org.scala-js" % "scalajs-library_2.13" % scalaJSVersion, // bump to at least scalaJSVersion
```
gzm0, line 686: I do not understand the comment here. As in: why does this line need that comment and none of the others?

gzm0 (changes requested):

sbt-plugin/src/sbt-test/scala3/basic/build.sbt
sjrd: I switch to depending on scalajs-ir.

sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala
sjrd: I've elaborated the comment. Is it clearer now?

gzm0 (approved):

PR #4286, opened 2020-11-13
Fix #4284: Allow at most one of @JSName, @JSBracketAccess, @JSBracketCall

### LINKED ISSUE #4284: Compiler does not fail if @JSName and @JSBracketAccess / @JSBracketCall is put on the same member
```scala
@js.native
@JSGlobal
class A extends js.Object {
  @JSBracketAccess
  @JSName("bar")
}

@js.native
@JSGlobal
class A extends js.Object {
  @JSBracketCall
  @JSName("bar")
}
```

These both compile, but clearly shouldn't.

gzm0:


sjrd:
Technically, this is potentially a source breaking change, and therefore it must go in a minor version.

I think we should release a v1.3.1 before merging this.

gzm0:
> I think we should release a v1.3.1 before merging this.

Agreed.

gzm0:
Rebased and bumped minor version.

sjrd (approved):

Issue #4282, opened 2020-11-11
NonNative JS class "ref-check" does not special case no-JSName apply


gzm0:
Found while fiddling with #4281.

```scala
package helloworld

import scala.scalajs.js
import js.annotation._

@js.native
@JSGlobal("Object")
abstract class MyClass extends js.Object {
  def apply(): Int
}

object HelloWorld {
  def main(args: Array[String]): Unit = {
    val x = new MyClass {
      @JSName("apply")
      def apply(): Int = 1
    }

    println(x())
  }
}
```

This compiles and links, but fails because `x` is not callable.

Expected: error message similar to:

```
object creation impossible, since method apply in class MyClass of type ()Int is not defined
    val x = new MyClass {
                ^
```

gzm0:
Note that if a similar "stunt" is attempted with `js.Function`, #4281 triggers:

```scala
object HelloWorld {
  @js.native
  @JSGlobal("Function")
  abstract class F extends js.Function with js.Function0[Int]

  def main(args: Array[String]): Unit = {
    val x = new F {
      def apply(): Int = 1
    }

    println(x())
  }
}
```

Gives:

```
Concrete members of JS native types may only call js.native.
      def apply(): Int = 1
                         ^
```

gzm0:
Looking at GenJSCode, it seems that a similar issue exists for method names that map to JS operators.

PR #4288, opened 2020-11-13
Fix #4281: Do not treat anonymous JS classes as lambdas

### LINKED ISSUE #4281: Bad error message when trying to define `apply` in an anonymous class extending js.Function
While trying to break our detection of JS functions, I found this:

```scala
@js.native
@JSGlobal("bad")
abstract class BadFunction extends js.Function1[Int, String]

object Test {
  new BadFunction {
    def apply(x: Int): String = "f"
  }
}
```

This fails to compile with the following error:

```
Concrete members of JS native types may only call js.native.
      def apply(x: Int): String = "f"
                                  ^
```

Note that this does *not* happen for other native traits:

```scala
@js.native
trait MyTrait[T] extends js.Object {
  def apply(): T
}

@js.native
@JSGlobal("bad")
abstract class BadMyTrait extends MyTrait[Int]

object Test {
  new BadMyTrait {
    def apply(): Int = 1
  }
}
```

This fails as expected with:

```
A non-native JS class cannot declare a method named `apply` without `@JSName`
      def apply(): Int = 1
          ^
```

gzm0:
This merely improves the error messages.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
@@ -499,8 +499,16 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
 
       sym.addAnnotation(JSTypeAnnot)
 
-      val isJSLambda =
-        sym.isAnonymousClass && AllJSFunctionClasses.exists(sym.isSubClass(_))
+      val isJSLambda = {
+        /* Under 2.11, sym.isAnonymousFunction does not properly recognize
+         * anonymous functions here (because they seem to not be marked as
+         * synthetic).
+         */
+        sym.name == tpnme.ANON_FUN_NAME &&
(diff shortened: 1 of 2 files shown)


gzm0:
I know we have agreed to disallow inheriting from `js.FunctionN` for non-native classes. It started bothering me that we treat `js.FunctionN` special in that regard. IMO a nicer approach is to fix #4282. It is then impossible by construction to inherit non-natively, concretely from `js.FunctionN`.

sjrd (approved):
IIUC this can go in 1.3.1, because it only changes the error message?

gzm0:
Yep, this can go into 1.3.1.

sjrd:
👍 

It doesn't seem to work in 2.11, though :(

gzm0:
It got more hacky but it works :S WDYT?

sjrd (approved):
Looks good to me. It's not even that hacky. ;)

PR #4289, opened 2020-11-15
FF ONLY Towards 1.3.2.

sjrd:
@gzm0 I did not remove the MiMa exclusions in `BinaryIncompatibilities.scala`, according to your suggestion. We should decide now whether this is the way we want to go.

### DIFF
--- a/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -17,7 +17,7 @@ import java.util.concurrent.ConcurrentHashMap
 import scala.util.matching.Regex
 
 object ScalaJSVersions extends VersionChecks(
-    current = "1.3.1",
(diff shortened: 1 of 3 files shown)


gzm0:
I definitely think we should not remove them. What I'm wondering is:
- Whether they are in the right place? (or should we separate "expected" incompatibilities from "unexpected" incompatibilities).
- Should we construct this list incrementally like this, or full at once (based on `VERSIONING.md`).

Notably, we probably should avoid a setup where it is difficult to judge if "towards" commit does the right thing (i.e. are the right lines in `BinaryIncompatibiliies.scala` removed or not).

sjrd:
Right. I was also thinking about that. But *designing* something is definitely not the job of a Towards commit, so right now, in this commit, either we keep them as is or we remove them, IMO.

gzm0:
Agreed. In that case I think we should remove it.

sjrd:
All right. Removed.

gzm0 (approved):

PR #4290, opened 2020-11-16
Fix #4292: Return `-1` at first `EOF` at `Base64.DecodingInputStream` 

### LINKED ISSUE #4292: Base64.DecodingInputStream returns 0 instead of -1
I just discovered very wired behaviour inside Base64.DecodingInputStream.

How it works at scala on JVM and Scala-native:
```
scala> import java.util.Base64
import java.util.Base64

scala> Base64.getDecoder.wrap(() => -1).read()
val res0: Int = -1

scala> 
```

SJS returns `0` at this case.

A way to fix: https://github.com/scala-js/scala-js/pull/4290

catap:
When `InputStream` that is used as source for `Base64.DecodingInputStream` returns `-1` (aka EOF) at the first read, the `Base64.DecodingInputStream.read` returns `0` that broke contract for this method.

The root cause of the issue that `eof` flag doesn't checked after `iterate`.

Fixed it.

### DIFF
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/Base64Test.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/Base64Test.scala
@@ -12,7 +12,7 @@
 
 package org.scalajs.testsuite.javalib.util
 
-import java.io.{ByteArrayInputStream, ByteArrayOutputStream, IOException}
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, IOException, InputStream}
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets.ISO_8859_1
 import java.util.Base64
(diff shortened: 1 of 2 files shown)


sjrd:
Thanks for the PR. Could you please file an issue that describes the issue, with a reproduction, so that we can refer to it in the PR?

catap:
@sjrd done and improved to code to support old scala

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/Base64Test.scala
```diff
@@ -285,6 +285,13 @@ class Base64Test {
     }
   }
 
+  @Test def decodeInputStreamFirstEOF(): Unit = {
+    val eofEachRead = new InputStream {
```
sjrd, line 289: Perhaps call this `emptyInputStream` instead?

sjrd (changes requested):
Thank you.

The code looks good. The commit message should follow [our guidelines](https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages), so it should look something like:

```
Fix #4292: Return -1 at first EOF in Base64.DecodingInputStream.

...
# no "Fixed it" at the end; that's in the title
```
Could you also rebase on the latest master. Currently your branch sits directly on top of a version commit, which means it also contains the same version number. It should sit on top of the "Towards 1.3.2." commit instead (or anything later).

sjrd (approved):

PR #4294, opened 2020-11-17
Rename master/slave to controller/worker in the testing infrastructure.

sjrd:
This is just to avoid the controversial pair of words "master/slave". The change aligns with how those things were named in Scala Native.

The method `sbt.testing.Framework.slaveRunner` is not renamed, because doing so would break binary compatibility.

In BinaryIncompatibilities, we merge `TestCommon` and `TestAdapter`. `TestCommon` used to be reused for the `test-bridge` artifact, but that one does not have any binary compatibility constraints anymore, so the distinction was artificial.

---

This PR is a direct consequence of the [Scala Center Advisory Board Recommendation SCP-025](https://github.com/scalacenter/advisoryboard/blob/master/proposals/025-inclusive-language.md), which was adopted at the latest Advisory Board Meeting. Although we have not worked on a list of words to avoid yet, it is clear that the "master/slave" association will end up on the eventual list (even if it's the only thing that ends up there). So this PR is a first step in that direction.

/cc @olafurpg, since you were interested in the issue with the `slaveRunner` method.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -166,7 +166,10 @@ def Tasks = [
         testInterface$v/compile:doc testBridge$v/compile:doc &&
     sbtretry ++$scala headerCheck &&
     sbtretry ++$scala partest$v/fetchScalaSource &&
-    sbtretry ++$scala library$v/mimaReportBinaryIssues testInterface$v/mimaReportBinaryIssues
+    sbtretry ++$scala \
+        library$v/mimaReportBinaryIssues \
+        testInterface$v/mimaReportBinaryIssues \
+        jUnitRuntime$v/mimaReportBinaryIssues
   ''',
 
   "test-suite-default-esversion": '''
--- a/junit-runtime/src/main/scala/com/novocode/junit/JUnitFramework.scala
+++ b/junit-runtime/src/main/scala/com/novocode/junit/JUnitFramework.scala
@@ -30,6 +30,7 @@ final class JUnitFramework extends Framework {
     f.runner(args, remoteArgs, testClassLoader)
   }
 
+  // Aka `workerRunner`; see the Scaladoc of `sbt.testing.Framework` about the name.
   def slaveRunner(args: Array[String], remoteArgs: Array[String],
       testClassLoader: ClassLoader, send: String => Unit): Runner = {
     f.slaveRunner(args, remoteArgs, testClassLoader, send)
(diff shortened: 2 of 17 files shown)


gzm0:
LG overall, just one question: Does "controller" accurately convey that this Runner instance is expected to run tests as well? Is there an industry standard about this? Otherwise maybe "main" would be a better name ("primary" also comes to mind, but this IMO implies ordered-ness to a larger extent than a single out of many).

sjrd:
There's a bunch of alternatives at https://en.wikipedia.org/wiki/Master/slave_(technology)#Replacements, but none truly convinces me. Anything that says "replica" is plain wrong. I did hesitate with leader/follower, but IMO it doesn't convey the controlling aspect of the leader enough. At least "controller" does not contradict the fact that it's working as well. I'm not aware of any "industry standard" terminology that says "I'm giving orders but I'm working too.".

sjrd:
Also, it's worth pointing out that controller/worker was already accepted a few months ago in the Scala Native repo. Unless there is strong concern against that naming, we should keep it. Otherwise we shuld get other Scala Native contributors on board with another rename, because ultimately we should have the same names in both repos.

ekrich:
It looks like Spark uses Driver/Worker but after reviewing the list on Wiki, I think that controller/worker is about the best combo.

gzm0:
The Spark Driver doesn't do any work though IIUC.

sjrd:
Let's wait until the Scala Center has produced the actual guidelines before committing to the new names.

sjrd:
A few moons later ... the Scala Center's guide for inclusive language is published: https://www.scala-lang.org/contribute/inclusive-language/
So I rebased this PR.

gzm0:
Coming back to this I think Main/Worker would a better name, but we should probably just stick with Controller/Worker for the following reasons:

- It's already in Scala Native (I'm not really happy about this precedent though; in my understanding, Scala Native is the less stable repository of these two, so if we let it drive precedent cases like this, it might drive code quality down)
- It seems to be a standard pair (at least in the Scala Center's guide).
- It's not *that* important :)

gzm0 (approved):

gzm0:
Honestly, the more I think about this, the less I think it matters that much. In the end, the precise interaction semantics will never be entirely captured by a name.

The problem with "Main/Worker" is that the words don't form an obvious pair (compare with Primary/Secondary)...

sjrd:
OK. Let's keep what we have here, then, and stop worrying about it. 😉

PR #4296, opened 2020-11-20
Fix #4282: Unify handling of JS calling conventions

### LINKED ISSUE #4282: NonNative JS class "ref-check" does not special case no-JSName apply

gzm0:
This fixes a number of issues related to overriding checking.

As a bonus, we also wont crash (but display a nice error), if a
repeated parameter is used in a binary operation.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/JSGlobalAddons.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/JSGlobalAddons.scala
@@ -17,6 +17,7 @@ import scala.tools.nsc._
 import scala.collection.mutable
 
 import org.scalajs.ir.Trees.JSNativeLoadSpec
+import org.scalajs.ir.{Trees => js}
 
 /** Additions to Global meaningful for the JavaScript backend
  *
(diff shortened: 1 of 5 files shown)


compiler/src/main/scala/org/scalajs/nscplugin/JSGlobalAddons.scala
```diff
@@ -131,6 +132,122 @@ trait JSGlobalAddons extends JSDefinitions
...
+        } else if (isJSBracketCall(sym)) {
+          BracketCall
+        } else {
+          val jsNameAnnot = sym.getAnnotation(JSNameAnnotation)
+          lazy val pc = sym.paramss.map(_.size).sum
+
+          sym.name match {
+            case nme.apply if jsNameAnnot.isEmpty => Call
+
+            case JSUnaryOpMethodName(code) if pc == 0 && jsNameAnnot.isEmpty =>
```
gzm0, line 182: Only doing this conditional on the parameter count is the current behavior. I wonder if we should keep it.

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -641,17 +641,17 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
           membSym.defStringSeenAs(sym.thisType.memberType(membSym))
 
         // Check for overrides with different JS names - issue #1983
-        if (jsInterop.jsNameOf(low) != jsInterop.jsNameOf(high)) {
+        if (jsInterop.JSCallingConvention.of(low) != jsInterop.JSCallingConvention.of(high)) {
```
gzm0, line 644: This change probably stretches our notion of source v.s. binary compatibility. IIUC, this could now fail at usage site of a library compiled with a older version of Scala.js.

Unsure though what to do about this (IIUC detecting if the current class is "the problem" is hard).

gzm0:
I'm not super happy with this solution (it feels a bit bloated), but it's the best I could come up with so far.

sjrd (changes requested):

compiler/src/main/scala/org/scalajs/nscplugin/JSGlobalAddons.scala
```diff
@@ -131,6 +132,122 @@ trait JSGlobalAddons extends JSDefinitions
...
+        if (isJSBracketAccess(sym)) {
+          BracketAccess
+        } else if (isJSBracketCall(sym)) {
+          BracketCall
+        } else {
+          val jsNameAnnot = sym.getAnnotation(JSNameAnnotation)
+          lazy val pc = sym.paramss.map(_.size).sum
+
+          sym.name match {
+            case nme.apply if jsNameAnnot.isEmpty => Call
```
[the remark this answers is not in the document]
gzm0: Yes. At first I thought it'll make property / method handling more complicated, but I found a solution that I like more.

sjrd (approved):

PR #4299, opened 2020-11-24
Fix #4212: Set temp file permissions to 0644 in PathOutputDirectory

### LINKED ISSUE #4212: Compiled JS files permissions are too restrictive

gzm0:


gzm0:
Follow-up on #4220. /cc @janilcgarcia

sjrd (approved):

PR #4300, opened 2020-11-25
Fix usages of Jimfs in tests on Windows.

sjrd:
Jimfs defaults to using a configuration that matches the OS it is running on. On Windows, that means that there is no such thing as a top-level `/tmp` path. Instead, we now use `getPath("tmp").toAbsolutePath()`.

We add some smoke tests in the AppVeyor build to prevent regressions in the future.

### DIFF
--- a/linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
+++ b/linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
@@ -36,7 +36,7 @@ class PathOutputDirectoryTest {
 
   @Test
   def avoidUnnecessaryWrite(): AsyncResult = await {
-    val dir = Jimfs.newFileSystem().getPath("/tmp")
+    val dir = Jimfs.newFileSystem().getPath("tmp").toAbsolutePath()
     Files.createDirectory(dir)
 
     val fileName = "file.js"
@@ -59,7 +59,7 @@ class PathOutputDirectoryTest {
 
   @Test
   def readFull(): AsyncResult = await {
-    val dir = Jimfs.newFileSystem().getPath("/tmp")
+    val dir = Jimfs.newFileSystem().getPath("tmp").toAbsolutePath()
     Files.createDirectory(dir)
 
     val fileName = "file.js"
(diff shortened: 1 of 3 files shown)


linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
```diff
@@ -36,7 +36,7 @@ class PathOutputDirectoryTest {
 
   @Test
   def avoidUnnecessaryWrite(): AsyncResult = await {
-    val dir = Jimfs.newFileSystem().getPath("/tmp")
+    val dir = Jimfs.newFileSystem().getPath("tmp").toAbsolutePath()
```
gzm0, line 39: Is the `toAbsolutePath` actually necessary?

gzm0 (approved):

linker/jvm/src/test/scala/org/scalajs/linker/PathOutputDirectoryTest.scala
sjrd: The tests pass without the `toAbsolutePath()`. I added `toAbsolutePath()` to better correspond to what we were doing before, i.e., using an absolute path. But I can remove them if you prefer.

gzm0: I suspect my analysis of why 16135f6c839781a27d7810d55333c0a365217dbb was needed was premature: My suspicion is that instead of creating a "strange path without parent" `.getPath("tmp")` simply created a relative path.

I then cargo-culted the fix to PathOutputDirectoryTest.

Looking at the documentation of `getParent`, it seems that it is not sufficient to call it to get the enclosing directory of a file: It simply truncates the last element of the path.

```scala
scala> import java.nio.file.Paths
import java.nio.file.Paths

scala> val x = Paths.get("a/b/../")
x: java.nio.file.Path = a/b/..

scala> x.getParent
res1: java.nio.file.Path = a/b

scala> x.normalize.getParent
res2: java.nio.file.Path = null

scala> x.normalize
res3: java.nio.file.Path = a
```

This seems to be a bug in pretty much all our usages of `getParent`.

All that being said, it seems to be pretty orthogonal to this PR, so maybe we should merge it and follow up later? WDYT?

PR #4308, opened 2020-11-28
Fix #4102: Add Semaphore and use it in IOThrottler.

### LINKED ISSUE #4102: Implement java.util.concurrent.Semaphore and use it in IOThrottler
This is a follow-up to #4097 where we used an `AtomicInteger` as a poor-mans replacement for a `Semaphore` with `tryAcquire` (or maybe even `drainPermits`).

yilinwei:


yilinwei:
Comment to check CLA is signed.

yilinwei:
Test is failing - let me take a look.

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
```diff
@@ -48,33 +47,19 @@ private[linker] final class IOThrottler(totalSlots: Int) {
...
         work()
       }
     }
   }
-
-  @tailrec
-  private def tryGetSlot(): Boolean = {
-    val s = slots.get()
-    if (s > 0) {
-      if (slots.compareAndSet(s, s - 1)) {
```
yilinwei, line 71: I _think_ this is here because of the JVM and is the cause of the failing test; otherwise I don't quite understand how you'd get into the `else` branch?

gzm0:
Please:
- Confirm that you haven't look at the source code of the Oracle JDK or OpenJDK.
- Read and apply the guidelines about [commit messages](https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages).

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
gzm0: You get into the else branch if there is a race condition with another thread.

For example take the following execution schedule of `tryGetSlot` for `t1` and `t2`

1. `slots = 1`
2. t1: `s = slots.get() = 1`
3. t2: `s = slots.get() = 1`
4. t2: `slots.compareAndSet(1, 0) = true` (`slots = 0`)
5. t1: `slots.compareAndSet(1, 0) = false` (`slots` unchanged)

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
```diff
@@ -15,18 +15,17 @@ package org.scalajs.linker.standard
...
-import java.util.concurrent.ConcurrentLinkedQueue
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{ConcurrentLinkedQueue, Semaphore}
 
 private[linker] final class IOThrottler(totalSlots: Int) {
   /* This is basically a java.util.concurrent.Semaphore, but it is not
    * implemented in the javalib.
    */
-  private val slots = new AtomicInteger(totalSlots)
+  private val semaphore = new Semaphore(totalSlots)
```
gzm0, line 24: I'd keep calling this `slots` (call it what it represents, not what it is. notably, it wasn't called `atomicInteger` before).

gzm0 (changes requested):
I have not looked at the implementation of `Semaphore` yet. I'm waiting for your confirmation that you haven't looked at OpenJDK/OracleJDK.

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
yilinwei: Yes; but that makes sense only if you have more than one thread? So I guess the implementation needs to work on the JVM as well? In any case I'll add the behaviour back into the `Semaphore`.

yilinwei:
I confirm that I have not looked at any other `Semaphore` implementations of the JDK.

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
yilinwei: :+1:

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
gzm0: > Yes; but that makes sense only if you have more than one thread?

Yes. `IOThrottler` needs to work both on the JVM and on JS.

>  In any case I'll add the behaviour back into the Semaphore.

That is not necessary. The implementation of `Semaphore` is JS only.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def availablePermits(): Int = this.permits
+
+  def drainPermits(): Int = {
+    val permits = this.permits
+    this.permits = 0
+    permits
+  }
+
+  def getQueuedThreads(): Collection[Thread] = Collections.emptySet()
```
gzm0, line 36: `protectected`

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+package java.util.concurrent
+
+import java.util.{Collection, Collections}
+
+class Semaphore(var permits: Int, fairness: Boolean) extends Serializable {
+
+  def this(permits: Int) = this(permits, false)
+
+  // These methods can't be implemented because calling it more times than the number
+  // of permits makes no sense in Javascript.
```
gzm0, line 22: `// These methods can't be implemented because they block`?

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+  // def acquire(): Unit = ???
+  // def acquire(permits: Int): Unit = ???
+  // def acquireUninterruptibly(): Unit = this.acquire()
+  // def acquireUninterruptibly(permits: Int): Unit = this.acquire(permits)
+
+  def availablePermits(): Int = this.permits
```
gzm0, line 28: We usually do not qualify `this` if it is not necessary.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package java.util.concurrent
+
+import java.util.{Collection, Collections}
+
+class Semaphore(var permits: Int, fairness: Boolean) extends Serializable {
```
gzm0, line 17: `private[this] var permits`

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+  override def toString: String =
+    s"${super.toString}[Permits = ${this.permits}]"
+
+  def tryAcquire(): Boolean = this.tryAcquire(1)
+
+  def tryAcquire(permits: Int): Boolean = {
+    if (permits > 0) {
+      this.permits -= permits
+      true
+    } else false
```
gzm0, line 66: See CODINGSTYLE.md: If the `if` branch has braces, so should the `else` branch.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  // This method is technically incorrect; it returns too soon
```
gzm0, line 69: If it is incorrect, it's better if we do not implement it. (this basically requires blocking to be implemented correctly, so maybe just add it to the list of methods we cannot implement at the top).

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+  def tryAcquire(permits: Int, timeout: Long, unit: TimeUnit): Boolean =
+    tryAcquire(permits)
+
+  // This method is technically incorrect; it returns too soon
```
gzm0, line 73: Same here.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+
+package org.scalajs.testsuite.javalib.util.concurrent
+
+import java.util.concurrent.{Semaphore, TimeUnit}
+import org.junit.Assert._
+import org.junit.Test
+
+class SemaphoreTest {
+
+  @Test def `default constructor should be unfair`(): Unit = {
```
gzm0, line 21: Please chose a test naming style that is consistent with recent PRs (e.g. #4299). There is more discussion about this in #4266.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  @Test def `constructor takes negative values`(): Unit = {
+    val sem = new Semaphore(-1)
+    assertEquals(-1, sem.availablePermits)
+  }
+
+
+  @Test def testDrainNegative(): Unit = {
+    val sem = new Semaphore(-1)
+    sem.drainPermits()
```
gzm0, line 34: Also check the return value?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  val acquireMethods: List[Semaphore => Unit] = List(
+    _.drainPermits(),
+    _.tryAcquire(),
+    _.tryAcquire(1),
+    _.tryAcquire(1, 1L, TimeUnit.DAYS),
+    _.tryAcquire(1L, TimeUnit.DAYS),
+  )
+
+  @Test def testAcquires(): Unit = {
```
gzm0, line 46: This should also test that `tryAcquire` fails if there are no permits. An maybe also test that release actually releases a permit.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+class SemaphoreTest {
+
+  @Test def `default constructor should be unfair`(): Unit = {
+    val sem = new Semaphore(1)
+    assertFalse(sem.isFair)
+  }
+
+  @Test def `constructor takes negative values`(): Unit = {
+    val sem = new Semaphore(-1)
+    assertEquals(-1, sem.availablePermits)
```
gzm0, line 28: We should also test how such a semaphore behaves (i.e. when `tryAcquire` / `release` is called).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      val sem = new Semaphore(1)
+      acquire(sem)
+      assertEquals(0, sem.availablePermits)
+    }
+  }
+
+  val releaseMethods: List[Semaphore => Unit] = List(
+    _.release(),
+    _.release(1),
+  )
```
gzm0, line 57: I think separating this list from the test hurts readability. IMO duplication is better to understand here. And as an additional plus, you can test `release` with an argument that is larger than 1.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  @Test def testReleaseThrowsException(): Unit = {
+    val sem = new Semaphore(0)
+    try {
+      sem.release(-1)
+      fail("Exception was not thrown")
+    } catch {
+      case e: Exception => assertTrue(e.isInstanceOf[IllegalArgumentException])
```
gzm0, line 73: Match on the exception type directly? (and bubble it up otherwise)?

In general, we have helpers for checking exceptions. You might need to define one in this file, because for the `linker` project we don't have one yet. Look for `assertThrows` in the codebase.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      val sem = new Semaphore(0)
+      release(sem)
+      assertEquals(1, sem.availablePermits)
+    }
+  }
+
+  @Test def testReleaseThrowsException(): Unit = {
+    val sem = new Semaphore(0)
+    try {
+      sem.release(-1)
```
gzm0, line 70: We also need a test like this for `tryAcquire`.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,77 @@
...
+    this.permits += permits
+  }
+
+  override def toString: String =
+    s"${super.toString}[Permits = ${this.permits}]"
+
+  def tryAcquire(): Boolean = this.tryAcquire(1)
+
+  def tryAcquire(permits: Int): Boolean = {
+    if (permits > 0) {
```
gzm0, line 63: This check is insufficient. If `permits < 0` then this should throw an IllegalArgumenException. Otherwise it should check that `this.permits >= permits` and only then return true.

This is also the reason for the test failure: The Semaphore does not do its job :)

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
yilinwei: :+1:

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
yilinwei: Done.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+class Semaphore(private[this] var permits: Int, fairness: Boolean) extends Serializable {
+
+  def this(permits: Int) = this(permits, false)
+
+  // These methods can't be implemented because they block
+  // def acquire(): Unit = ???
+  // def acquire(permits: Int): Unit = ???
+  // def acquireUninterruptibly(): Unit = this.acquire()
+  // def acquireUninterruptibly(permits: Int): Unit = this.acquire(permits)
+  //def tryAcquire(permits: Int, timeout: Long, unit: TimeUnit): Boolean =
```
gzm0, line 26: Missing space here.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+package java.util.concurrent
+
+import java.util.{Collection, Collections}
+
+class Semaphore(private[this] var permits: Int, fairness: Boolean) extends Serializable {
+
+  def this(permits: Int) = this(permits, false)
+
+  // These methods can't be implemented because they block
+  // def acquire(): Unit = ???
```
gzm0, line 22: Try to be consistent with the method bodies in this list. I suggest not having them (just like abstract methods). So simply:

```scala
// def acquire(): Unit
// def acquire(permits: Int): Unit
```

etc.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+  protected def getQueuedThreads(): Collection[Thread] = Collections.emptySet()
+
+  def getQueueLength(): Int = 0
+
+  def hasQueuedThreads(): Boolean = false
+
+  def isFair(): Boolean = fairness
+
+  protected def reducePermits(reduction: Int): Unit = {
+    permits = permits - reduction
```
gzm0, line 46: `permits -= reduction`?

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  protected def getQueuedThreads(): Collection[Thread] = Collections.emptySet()
+
+  def getQueueLength(): Int = 0
+
+  def hasQueuedThreads(): Boolean = false
+
+  def isFair(): Boolean = fairness
+
+  protected def reducePermits(reduction: Int): Unit = {
```
gzm0, line 45: Also: This should throw if `reduction` is negative. This points to a weakness in the tests.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+    permits = permits - reduction
+  }
+
+  def release(): Unit = release(1)
+
+  def release(permits: Int): Unit = {
+    if (permits < 0 ) {
```
gzm0, line 52: remove the space before the brace (`if (permits < 0)`). (this is what is failing scalastyle).

linker/shared/src/main/scala/org/scalajs/linker/standard/IOThrottler.scala
```diff
@@ -15,18 +15,17 @@ package org.scalajs.linker.standard
...
 import scala.concurrent._
 
-import java.util.concurrent.ConcurrentLinkedQueue
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{ConcurrentLinkedQueue, Semaphore}
 
 private[linker] final class IOThrottler(totalSlots: Int) {
   /* This is basically a java.util.concurrent.Semaphore, but it is not
    * implemented in the javalib.
    */
```
gzm0, line 23: This comment can now go away :)

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  override def toString: String =
+    s"${super.toString}[Permits = ${permits}]"
+
+  def tryAcquire(): Boolean = tryAcquire(1)
+
+  def tryAcquire(permits: Int): Boolean = {
+    if (permits < 0) {
+      throw new IllegalArgumentException()
+    }
```
gzm0, line 66: This check occurs 2 times (3 times after a fix). Consider a private helper for it (something like: `requireNonNegative`).

gzm0 (changes requested):
Mostly minor things at this point. The AppVeyor build seems spurious. So hopefully it vanishes with your next push.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
yilinwei: Is it worth overriding the Semaphore to test this specific behaviour?

gzm0: Yes, I think so. Mainly IMO because it can make the numbers of permits negative. Which is somewhat a special case IMO.

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,76 @@
...
+    val old = permits
+    permits = 0
+    old
+  }
+
+  protected def getQueuedThreads(): Collection[Thread] = Collections.emptySet()
+
+  @inline private def requireNonNegative(n: Int): Unit = {
+    if (n < 0) {
+      throw new IllegalArgumentException()
```
gzm0, line 41: I don't think we put braces on zero arg constructors. @sjrd?

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  override def toString: String =
+    s"${super.toString}[Permits = ${permits}]"
+
+  def tryAcquire(): Boolean = tryAcquire(1)
+
+  def tryAcquire(permits: Int): Boolean = {
+    if (this.permits >= permits) {
+      reducePermits(permits)
```
gzm0, line 70: I'm not sure this is correct. IIUC it can cause the wrong behavior:

```scala
val x = new Semaphore(-2)
x.tryAcquire(-1) // false, instead of IllegalArgumentException
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/SemaphoreTest.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test def release(): Unit = {
+    val sem = new Semaphore(0)
+    assertEquals(0, sem.availablePermits)
+    sem.release()
+    assertEquals(1, sem.availablePermits)
+  }
+
+  @Test def releasePermits(): Unit = {
```
gzm0, line 76: I do not understand how this test is different from `release()`.

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/Semaphore.scala
sjrd: We don't have a definite rule for this. Either is fine. I don't think I am even consistent with myself about this.

gzm0:
@yilinwei are you planning on finishing this? If no, is it OK with you if I finish it?

yilinwei:
@gzm0 Please - go ahead. I haven't had much time to work on it.

gzm0:
Superseded by #4381

PR #4316, opened 2020-12-05
Automatically determine ancestors of HijackedClasses

gzm0:


sjrd:
The checksizes test is failing on 2.11.12.

Could you post a gist of the diff of the produced `.js` file for the reversi, for example? That might give a clearer idea of the effective changes.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/GlobalKnowledge.scala
```diff
@@ -104,4 +104,11 @@ private[emitter] trait GlobalKnowledge {
 
   /** The public (non-static) methods of java.lang.Object. */
   def methodsInObject(): List[Versioned[MethodDef]]
+
+  /** Hijacked classes that are strict descendants of `className`. */
+  def hijackedDescendants(className: ClassName): List[ClassName]
+
+  /** Convenience helper to check `hijackedDescendants` is nonEmpty. */
+  final def isAncestorOfHijackedClass(className: ClassName): Boolean =
+    hijackedDescendants(className).nonEmpty
```
sjrd, line 113: This could be implemented more efficiently in `SpecialInfo` as
```scala
hijackedDescendants.contains(className)
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/KnowledgeGuardian.scala
```diff
@@ -553,6 +569,23 @@ private[emitter] final class KnowledgeGuardian(config: Emitter.Config) {
...
+        ancestor <- hijackedClass.ancestors
+        if ancestor != hijackedClass.className
+      } yield {
+        (ancestor, hijackedClass)
+      }
+
+      for {
+        (ancestor, pairs) <- pairs.groupBy(_._1)
+      } yield {
+        (ancestor, pairs.map(_._2.className).toList)
```
sjrd, line 585: We should make sure that the `List`s on the right are always ordered in the same way. Otherwise, we could get spurious invalidate-all events just because the returned lists are not equal (although they contain the same elements).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/SJSGen.scala
```diff
@@ -129,7 +129,8 @@ private[emitter] final class SJSGen(
           genIsInstanceOfHijackedClass(expr, className)
         } else if (className == ObjectClass) {
           expr === Null()
-        } else if (className != NumberClass && // the only non-object superclass of hijacked classes
+        } else if (
+            !globalKnowledge.isAncestorOfHijackedClass(className) &&
             !globalKnowledge.isInterface(className)) {
```
sjrd, line 134: Formatting:
```suggestion
        } else if (!globalKnowledge.isAncestorOfHijackedClass(className) &&
            !globalKnowledge.isInterface(className)) {
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -2714,7 +2714,9 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
 
     def isMaybeHijackedClass(tpe: Type): Boolean = tpe match {
       case ClassType(className) =>
-        MaybeHijackedClasses.contains(className)
+        HijackedClasses.contains(className) ||
+        className != ObjectClass && globalKnowledge.hijackedDescendants(className).nonEmpty
```
sjrd, line 2718:
```suggestion
        className != ObjectClass && globalKnowledge.isAncestorOfHijackedClass(className)
```

sjrd: This code is called for every `Apply` node in a method, and the most common result is the one that has to go through all those tests. Could we precompute the total set (formerly `MaybeHijackedClasses`) at most once?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -805,32 +795,22 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
-            var test = if (tree.kind.isClass) {
-              genIsInstanceOfClass(obj, className)
-            } else {
-              !(!(
-                  genIsScalaJSObject(obj) &&
-                  genIsClassNameInAncestors(className,
-                      obj DOT "$classData" DOT "ancestors")
-              ))
+            val baseTest = {
+              if (tree.kind.isClass) {
```
sjrd, line 799: There's a spurious diff of several lines here, which could be avoided by keeping
```suggestion
            val baseTest = if (tree.kind.isClass) {
```

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -805,32 +795,22 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
-            if (isAncestorOfString)
-              test = test || typeOfTest("string")
-            if (isAncestorOfHijackedNumberClass) {
-              test = test || typeOfTest("number")
-              if (useBigIntForLongs)
-                test = test || genCallHelper("isLong", obj)
+            globalKnowledge.hijackedDescendants(className).foldLeft(baseTest) {
+              case (test, hijackedClass) =>
+                test || genIsInstanceOfHijackedClass(obj, hijackedClass)
             }
```
sjrd, line 813: This is the only place where we use the actual *list* of `hijackedDescendants`, and it is the one causing the size/performance regression. I'm not too worried about `j.l.Number` itself, but more about interfaces like `Comparable`. I think we could easily reclaim the performance by eliminating from the list the type test that are subsumed by another one in the list. Or, more directly, if the list contains `j.l.Double`, we can get rid of all the other boxed number classes (except `j.l.Long`).

We could also ignore `j.l.Long` here unless `useBigIntForLongs` is true.

None of those changes require anything but local knowledge, so this should be feasible here. We can protect this slightly more computation by a `isAncestorOfHijackedClass(className)`.

sjrd (changes requested):

gzm0:
Diff in reversi fastLink:

```diff
@@ -3194,7 +3194,7 @@
 }
 $h_jl_Number.prototype = $c_jl_Number.prototype;
 function $is_jl_Number(obj) {
-  return ((obj instanceof $c_jl_Number) || ((typeof obj) === "number"))
+  return (((((((obj instanceof $c_jl_Number) || $isInt(obj)) || ((typeof obj) === "number")) || ((typeof obj) === "number")) || (obj instanceof $c_RTLong)) || $isByte(obj)) || $isShort(obj))
 }
 function $as_jl_Number(obj) {
   return (($is_jl_Number(obj) || (obj === null)) ? obj : $throwClassCastException(obj, "java.lang.Number"))
```

gzm0:
After this update, there is no diff in the generated JS file anymore.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
gzm0: I'm not sure I understand. Do you expect performance improvements if we do a single `Set` lookup rather than this test? We can precompute the set in `KnowledgeGuardian` if you think it is worth it.

sjrd: That's what I would expect, but maybe it's not such a big deal.

sjrd (approved):

PR #4324, opened 2020-12-08
Fix #4322: Disallow default params in non-native JS traits entirely

### LINKED ISSUE #4322: JSOptional default parameters cause IR checking errors
```scala
trait A extends js.Any {
}

object Impl extends A {
}
```

```
 Referring to non-existent method static org.scalajs.testsuite.jsinterop.JSOptionalTest$Impl$1$.foo$default$1(org.scalajs.testsuite.jsinterop.JSOptionalTest$Impl$1$)scala.scalajs.js.$bar
```

gzm0:
This is a source incompatible change. There is unfortunately a slight
chance the this leaves a library author in a tricky situation: If a
library contains a non-native JS trait with default parameters, the
maintainers will be faced with the following choice:

- Add `@js.native`, source-breaking non-native Scala.js implementors (which
  might have worked around this issue by defining a default parameter
  themselves).
- Remove the default parameter, source-breaking call-sites.

### DIFF
--- a/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
+++ b/compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
@@ -1091,6 +1091,38 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
           } else if (sym.isLazy) {
             reporter.error(tree.pos,
                 "A non-native JS trait cannot contain lazy vals")
+          } else if (!sym.isDeferred && sym.hasFlag(reflect.internal.Flags.DEFAULTPARAM)) {
+            /* Default parameters in non-native JS traits are not allowed. We
+             * cannot implement them without binary compatibility issues:
+             *
+             *   trait A extends js.Object {
+             *     def foo(x: js.UndefOr[Int] = js.undefined): Int
+             *   }
+             *
+             *   abstract class B extends A {
+             *     def foo(x: js.UndefOr[Int]): Int
+             *   }
+             *
+             *   object Impl extends B {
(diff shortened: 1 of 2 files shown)


gzm0:
Two alternatives to this:
- Force non-native *direct* descendants of non-native JS traits with default parameters to override them (in source). This would allow any type of default parameter but set an unfortunate difference between default parameter semantics in traits and abstract classes.
- Synthesize a bridge default parameter getter in direct descendants of non-native JS traits. That getter could always safely return `js.undefined`.

Maybe the second alternative is the best. @sjrd WDYT?

gzm0:
@sjrd did you have a chance to think about this?

sjrd:
Ah, I had forgotten this one, sorry.



> Synthesize a bridge default parameter getter in direct descendants of non-native JS traits. That getter could always safely return `js.undefined`.

Can we pull this off even if the class that extends the non-native JS trait is a native class? And then a non-native class extends that native class while overriding the method without an explicit default param?

More generally: don't we have a similar problem when a native JS class defines a method with default params, which is then implemented in a non-native JS class?

> Force non-native _direct_ descendants of non-native JS traits with default parameters to override them (in source). This would allow any type of default parameter but set an unfortunate difference between default parameter semantics in traits and abstract classes.

Maybe we should always do this, i.e., always force default parameters to be re-specified when implementing/overriding a method. Whether that method was defined in a native/non-native class/trait. IMO that would be the most consistent.

gzm0:
> IMO that would be the most consistent.

Yeah, you are right. This is what makes most sense given JS calling conventions. Ideally, we'd even make the synthesized default accessors private static. But I think this isn't feasible without breaking binary compatibility.

gzm0:
Closing in favor of #4334.

PR #4329, opened 2020-12-12
Fix #4328: Fix Class.isAssignableFrom for hijacked classes and JS types.

### LINKED ISSUE #4328: j.l.Class.isAssignableFrom has some serious issues
Here a number of situations in which `isAssignableFrom` returns wrong results:

```scala
    assertFalse(classOf[j.l.Integer].isAssignableFrom(classOf[j.l.Double])) // true
    assertFalse(classOf[j.l.Double].isAssignableFrom(classOf[j.l.Integer])) // true

    assertTrue(classOf[js.EvalError].isAssignableFrom(classOf[js.EvalError])) // false
    assertTrue(classOf[js.Iterator[_]].isAssignableFrom(classOf[js.Iterator[_]])) // throws

    assertTrue(classOf[js.Error].isAssignableFrom(classOf[js.EvalError])) // false
    assertTrue(classOf[js.Iterable[_]].isAssignableFrom(classOf[js.Array[_]])) // throws
    assertTrue(classOf[js.Any].isAssignableFrom(classOf[js.Dictionary[_]])) // throws

    assertFalse(classOf[js.Iterator[_]].isAssignableFrom(classOf[js.Object])) // throws
    assertFalse(classOf[js.Iterator[_]].isAssignableFrom(classOf[js.Iterable[_]])) // throws

    assertFalse(classOf[js.Object].isAssignableFrom(classOf[js.Iterator[_]])) // true
```
All those problems come from the fact that we implement `isAssignableFrom` in terms of `isInstance`, by passing it a "fake instance" of the argument class. That works find for regular Scala classes, but it completely breaks down for hijacked classes and for JS types.

For hijacked classes, we use `0` as the fake instance of all number types. This causes them to be seen as interchangeable. For JS classes, basically we're testing where the fake instance `{ $classData: something }` is an `instanceof TheJSClass`. That answers true if and only if `TheJSClass` is `js.Object`! For JS traits, it throws.

Historically, the reason we did that was to "support" the pattern
```scala
def isInstance(c: Class[_], x: Any): Boolean = c.isAssignableFrom(x.getClass())
```
which was used in `ClassTag`s until 2.11.2 (excluded).

At least, with our implementation, that answered `true` when the JVM would answer `true`.Without our weird implementation, the above would return `false` for `isInstance(classOf[j.l.Integer], 5)` because `5.getClass()` is `classOf[j.l.Byte]`.

 However, it would also sometimes answer `true` if the JVM says `false`, and in some cases very wrongly so (`isInstance(classOf[j.l.Integer], 5.5)` was true!).

Since then, that pattern has been replaced by the much better `c.isInstance(x)` (duh!), at least in `ClassTag`s.

Fixing `isAssignableFrom` to make all the above tests pass can "break" the pattern in some cases (e.g., `isInstance(classOf[j.l.Integer], 5)`). It may still exist in some libraries, so we might "break" those libraries. However, they are already broken anyway (because of `isInstance(classOf[j.l.Integer], 5.5)`), so I don't think we would cause real harm.

sjrd:
Previously, we were creating a "fake instance" of the argument class, and passing it to `isInstance`. This caused a number of issues for hijacked classes and JS types, as detailed in the issue.

We fix all those issues by not relying on fake instances anymore. Instead, we directly compare the class data, notably using the `ancestors` field.

The new comparison matches the one used to test instances of arrays, which is natural since they now both correspond to the JVM subtyping test. We therefore rearrange the `isInstance` test of array classes to build on the same infrastructure as `isAssignableFrom`. This allows to simplify the class data in the process.


sjrd:
Here is the diff of the generated code for the reversi:
```diff
--- examples/reversi/.2.12/target/scala-2.12/reversi-fastopt.js	2020-12-12 15:37:09.200161300 +0100
+++ ../reversi-fastopt.js	2020-12-12 15:35:55.726996900 +0100
@@ -411,7 +411,7 @@
   this.arrayEncodedName = "";
   this._classOf = (void 0);
   this._arrayOf = (void 0);
-  this.isArrayOf = (void 0);
+  this.isAssignableFromFun = (void 0);
   this.name = "";
   this.isPrimitive = false;
   this.isInterface = false;
@@ -419,11 +419,14 @@
   this.isJSClass = false;
   this.isInstance = (void 0)
 }
-$TypeData.prototype.initPrim = (function(zero, arrayEncodedName, displayName, isArrayOf) {
+$TypeData.prototype.initPrim = (function(zero, arrayEncodedName, displayName) {
   this.ancestors = {};
   this.zero = zero;
   this.arrayEncodedName = arrayEncodedName;
-  this.isArrayOf = isArrayOf;
+  var self = this;
+  this.isAssignableFromFun = (function(that) {
+    return (that === self)
+  });
   this.name = displayName;
   this.isPrimitive = true;
   this.isInstance = (function(obj) {
@@ -431,13 +434,13 @@
   });
   return this
 });
-$TypeData.prototype.initClass = (function(internalNameObj, isInterface, fullName, ancestors, isJSType, parentData, isInstance, isArrayOf) {
+$TypeData.prototype.initClass = (function(internalNameObj, isInterface, fullName, ancestors, isJSType, parentData, isInstance) {
   var internalName = $propertyName(internalNameObj);
   this.ancestors = ancestors;
   this.arrayEncodedName = (("L" + fullName) + ";");
-  this.isArrayOf = (isArrayOf || (function(obj, depth) {
-    return (!(!(((obj && obj.$classData) && (obj.$classData.arrayDepth === depth)) && obj.$classData.arrayBase.ancestors[internalName])))
-  }));
+  this.isAssignableFromFun = (function(that) {
+    return (!(!that.ancestors[internalName]))
+  });
   this.isJSType = (!(!isJSType));
   this.name = fullName;
   this.isInterface = isInterface;
@@ -489,10 +492,17 @@
   this.arrayBase = componentBase;
   this.arrayDepth = arrayDepth;
   this.arrayEncodedName = encodedName;
+  var isAssignableFromFun = (function(that) {
+    var thatDepth = that.arrayDepth;
+    return ((thatDepth === arrayDepth) ? componentBase.isAssignableFromFun(that.arrayBase) : ((thatDepth > arrayDepth) && (componentBase === $d_O)))
+  });
+  this.isAssignableFromFun = isAssignableFromFun;
   this.name = encodedName;
   this.isArrayClass = true;
+  var self = this;
   this.isInstance = (function(obj) {
-    return componentBase.isArrayOf(obj, arrayDepth)
+    var data = (obj && obj.$classData);
+    return ((!(!data)) && ((data === self) || isAssignableFromFun(data)))
   });
   return this
 });
@@ -509,29 +519,7 @@
   return this._classOf
 });
 $TypeData.prototype.isAssignableFrom = (function(that) {
-  if ((this.isPrimitive || that.isPrimitive)) {
-    return (this === that)
-  } else {
-    var thatFakeInstance;
-    if ((that === $d_T)) {
-      thatFakeInstance = ""
-    } else if ((that === $d_jl_Boolean)) {
-      thatFakeInstance = false
-    } else if ((((((that === $d_jl_Byte) || (that === $d_jl_Short)) || (that === $d_jl_Integer)) || (that === $d_jl_Float)) || (that === $d_jl_Double))) {
-      thatFakeInstance = 0
-    } else if ((that === $d_jl_Long)) {
-      thatFakeInstance = $L0
-    } else if ((that === $d_jl_Character)) {
-      thatFakeInstance = $bC0
-    } else if ((that === $d_jl_Void)) {
-      thatFakeInstance = (void 0)
-    } else {
-      thatFakeInstance = {
-        $classData: that
-      }
-    };
-    return this.isInstance(thatFakeInstance)
-  }
+  return ((this === that) || this.isAssignableFromFun(that))
 });
 $TypeData.prototype.checkCast = (function(obj) {
   if ((((obj !== null) && (!this.isJSType)) && (!this.isInstance(obj)))) {
@@ -641,15 +629,15 @@
     $throwArrayCastException(obj, "D", depth)
   }
 }
-var $d_V = new $TypeData().initPrim((void 0), "V", "void", $isArrayOf_V);
-var $d_Z = new $TypeData().initPrim(false, "Z", "boolean", $isArrayOf_Z);
-var $d_C = new $TypeData().initPrim(0, "C", "char", $isArrayOf_C);
-var $d_B = new $TypeData().initPrim(0, "B", "byte", $isArrayOf_B);
-var $d_S = new $TypeData().initPrim(0, "S", "short", $isArrayOf_S);
-var $d_I = new $TypeData().initPrim(0, "I", "int", $isArrayOf_I);
-var $d_J = new $TypeData().initPrim("longZero", "J", "long", $isArrayOf_J);
-var $d_F = new $TypeData().initPrim(0.0, "F", "float", $isArrayOf_F);
-var $d_D = new $TypeData().initPrim(0.0, "D", "double", $isArrayOf_D);
+var $d_V = new $TypeData().initPrim((void 0), "V", "void");
+var $d_Z = new $TypeData().initPrim(false, "Z", "boolean");
+var $d_C = new $TypeData().initPrim(0, "C", "char");
+var $d_B = new $TypeData().initPrim(0, "B", "byte");
+var $d_S = new $TypeData().initPrim(0, "S", "short");
+var $d_I = new $TypeData().initPrim(0, "I", "int");
+var $d_J = new $TypeData().initPrim("longZero", "J", "long");
+var $d_F = new $TypeData().initPrim(0.0, "F", "float");
+var $d_D = new $TypeData().initPrim(0.0, "D", "double");
 /** @constructor */
 function $c_O() {
   /*<skip>*/
@@ -696,7 +684,7 @@
   O: 0
 }, false, "java.lang.Object", {
   O: 1
-}, (void 0), (void 0), $is_O, $isArrayOf_O);
+}, (void 0), (void 0), $is_O);
 $c_O.prototype.$classData = $d_O;
 /** @constructor */
 function $c_jl_Class(data0) {
```

ci/checksizes.sh
```diff
@@ -37,17 +37,17 @@ case $FULLVER in
     REVERSI_PREOPT_EXPECTEDSIZE=517000
     REVERSI_OPT_EXPECTEDSIZE=107000
     REVERSI_PREOPT_GZ_EXPECTEDSIZE=67000
-    REVERSI_OPT_GZ_EXPECTEDSIZE=39000
+    REVERSI_OPT_GZ_EXPECTEDSIZE=29000
```
sjrd, line 40: This one was most certainly too high before.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -1187,16 +1187,22 @@ private[emitter] object CoreJSLib {
...
+            paramList(zero, arrayEncodedName, displayName), {
           Block(
               privateFieldSet("ancestors", ObjectConstr(Nil)),
               privateFieldSet("zero", zero),
               privateFieldSet("arrayEncodedName", arrayEncodedName),
-              privateFieldSet("isArrayOf", isArrayOf),
+              const(self, This()),
+              privateFieldSet("isAssignableFromFun", {
+                Function(arrow = false, paramList(that), {
+                  Return(that === self)
```
gzm0, line 1203: I do not understand why we cannot rely on `This()` in this function. It seems it is always called with a receiver.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -1187,16 +1187,22 @@ private[emitter] object CoreJSLib {
...
         MethodDef(static = false, Ident("initPrim"),
-            paramList(zero, arrayEncodedName, displayName, isArrayOf), {
+            paramList(zero, arrayEncodedName, displayName), {
           Block(
               privateFieldSet("ancestors", ObjectConstr(Nil)),
               privateFieldSet("zero", zero),
               privateFieldSet("arrayEncodedName", arrayEncodedName),
-              privateFieldSet("isArrayOf", isArrayOf),
+              const(self, This()),
+              privateFieldSet("isAssignableFromFun", {
```
gzm0, line 1201: Why the suffix `Fun`? IIUC `isAssignableFrom` is not any different from `isInstance` for example. So having `Fun` for one but not the other seems confusing.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -1364,10 +1371,32 @@ private[emitter] object CoreJSLib {
...
               publicFieldSet("name", encodedName),
               publicFieldSet("isArrayClass", bool(true)),
+              const(self, This()),
               publicFieldSet("isInstance", Function(arrow = false, paramList(obj), {
-                Return(Apply(componentBase DOT "isArrayOf", obj :: arrayDepth :: Nil))
+                val data = varRef("data")
+                Block(
+                    const(data, obj && (obj DOT classData)),
+                    Return(!(!data) && {
+                      (data === self) || // fast path
```
gzm0, line 1396: Doesn't the fast path also apply to `isAssignableFromFun`?

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassJSTest.scala
```diff
@@ -19,6 +19,47 @@ import scala.scalajs.js
 
 class ClassJSTest {
 
+  @Test def isAssignableFrom(): Unit = {
+    /* isAssignableFrom should respect the JVM rules even for JS types,
+     * although isInstance doesn't.
```
gzm0, line 24: I think this warrants a short sentence of *why*. (FWIW, I agree with this, because `isAssignableFrom` basically allows reflection on the class hierarchy, so the *declared* classes matter).

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassJSTest.scala
```diff
@@ -19,6 +19,47 @@ import scala.scalajs.js
...
+    assertTrue(classOf[Object].isAssignableFrom(classOf[js.EvalError]))
+    assertTrue(classOf[Object].isAssignableFrom(classOf[js.Iterator[_]]))
+
+    assertTrue(classOf[js.Object].isAssignableFrom(classOf[js.Object]))
+    assertTrue(classOf[js.EvalError].isAssignableFrom(classOf[js.EvalError])) // false
+    assertTrue(classOf[js.Iterator[_]].isAssignableFrom(classOf[js.Iterator[_]])) // throws
+
+    assertTrue(classOf[js.Error].isAssignableFrom(classOf[js.EvalError])) // false
+    assertTrue(classOf[js.Iterable[_]].isAssignableFrom(classOf[js.Array[_]])) // throws
+    assertTrue(classOf[js.Any].isAssignableFrom(classOf[js.Dictionary[_]])) // throws
```
gzm0, line 39: The comments behind the asserts seem outdated.

test-suite/js/src/test/scala/org/scalajs/testsuite/javalib/lang/ClassJSTest.scala
```diff
@@ -19,6 +19,47 @@ import scala.scalajs.js
...
+    assertFalse(classOf[js.Date].isAssignableFrom(classOf[js.Iterator[_]]))
+    assertFalse(classOf[js.Iterator[_]].isAssignableFrom(classOf[js.Iterable[_]])) // throws
+
+    assertFalse(classOf[js.Object].isAssignableFrom(classOf[js.Iterator[_]])) // true
+    assertFalse(classOf[js.Date].isAssignableFrom(classOf[js.Iterator[_]]))
+    assertFalse(classOf[Array[js.Object]].isAssignableFrom(classOf[Array[js.Iterator[_]]]))
+
+    assertFalse(classOf[Array[js.Date]].isAssignableFrom(classOf[Array[js.Object]]))
+    assertFalse(classOf[Array[Array[Object]]].isAssignableFrom(classOf[Array[js.Object]]))
+  }
```
gzm0, line 61: Add:

```scala
assertFalse(classOf[js.Throwable].isAssignableFrom(classOf[Throwable]))
```

?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -1399,49 +1428,11 @@ private[emitter] object CoreJSLib {
...
-                    })
-                  })
-                }),
-                Return(Apply(genIdentBracketSelect(This(), "isInstance"),
-                    thatFakeInstance :: Nil))
-            )
-          })
+          Return(
+              (This() === that) || // fast path
+              Apply(This() DOT "isAssignableFromFun", that :: Nil))
```
gzm0, line 1435: It seems that this distinction of `isAssignableFromFun` / `isAssignableFrom` has emerged because we currently DCE `isAssignableFrom` is `jl.Class` is not instantiated (because we do not use it internally). I'm not sure this is still warranted. Consider simply using the public field. (There seems to be precedent for that for `isInstance`).

If you do this, also initialize it in the `ctor`.

gzm0 (changes requested):

gzm0:
Just to double check: This will not change array runtime subtyping checks, right?

sjrd:
> Just to double check: This will not change array runtime subtyping checks, right?

Indeed. I added a paragraph for that in the commit message.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
gzm0: > we never DCE isAssignableFrom.

WDYM? If `jl.Class` is not instantiated, we do not emit it. Look for `globalKnowledge.isClassClassInstantiated` in `CoreJSLib.scala`.

That being said, if we keep the separation between the private / public function, we can also keep it this way.

gzm0 (approved):

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
sjrd: Huh. I had completely forgotten about that. Yes, you're right, of course.

I had not taken that into account, but it's true that the private/public decomposition also addresses that issue. That said, in a better future we would also DCE `isInstance` if `!isClassClassInstantiated`, and then that problem would not exist (but we'd still want the decomposition for the fast path).

PR #4330, opened 2020-12-14
Assert a Range instead of an upper bound in checksizes

gzm0:


Jenkinsfile
```diff
@@ -156,13 +156,15 @@ def Tasks = [
         reversi$v/fastLinkJS \
         reversi$v/fullLinkJS \
         reversi$v/clean &&
-    sbtretry ++$scala reversi$v/fastLinkJS reversi$v/fullLinkJS &&
+    sbtretry ++$scala \
+        reversi$v/fastLinkJS \
+	reversi$v/fullLinkJS \
+	reversi$v/checksizes &&
```
sjrd, line 162: There are tabs here.

sjrd (approved):

sjrd:
The ranges are not quite right for 2.12 and 2.13. This might be due to a difference between the system `gz` and the JDK's implementation.

gzm0:
Seems like our values simply weren't tight enough :)

sjrd:
Erf, it's still not quite right on 2.13.3:
```
[info] Checksizes: reversi2_13 Scala 2.13.3
[info] fastLink size = 771930 (expected Range 771000 to 772000)
[info] fullLink size = 165672 (expected Range 166000 to 167000) // <--
[info] fastLink gzip size = 97574 (expected Range 97000 to 98000)
[info] fullLink gzip size = 42676 (expected Range 42000 to 43000)
[error] checksizes failed
```

PR #4333, opened 2020-12-15
Add more tests for arrays.

sjrd:
The new tests are in preparation for the PR on optimize arrays. I found that several areas were under-tested.


test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    testArraySomeJSInterface(true, arrayOfJSInterface)
+    testArraySomeJSClass(true, arrayOfJSClass)
+
+    testArrayParentJSType(true, arrayOfJSInterface)
+    testArrayParentJSType(true, arrayOfJSClass)
+
+    testArraySomeJSInterface(false, arrayOfParentJSType)
+    testArraySomeJSClass(false, arrayOfParentJSType)
+
+    testArrayJSObject(false, arrayOfJSInterface)
```
gzm0, line 72: Also add `arrayOfParentJSType`?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+import org.scalajs.testsuite.utils.Platform._
+
+class RuntimeTypeTestsJSTest {
+  import RuntimeTypeTestsJSTest._
+
+  // Import the helpers from the non-JS test
+  import RuntimeTypeTestsTest._
+
+  @Test def scalaArraysOfJSTypes(): Unit = {
+    val arrayOfParentJSType = new Array[ParentJSType](0)
```
gzm0, line 31: The name `ParentJSType` might be misleading here: The distinction between class / interface is crucial for some of the tests here. So we should probably call it `ParentJSInterface`?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+
+    testArrayObject(true, arrayOfParentJSType)
+    testArrayObject(true, arrayOfJSInterface)
+    testArrayObject(true, arrayOfJSClass)
+
+    testArrayParentJSType(true, arrayOfParentJSType)
+    testArraySomeJSInterface(true, arrayOfJSInterface)
+    testArraySomeJSClass(true, arrayOfJSClass)
+
+    testArrayParentJSType(true, arrayOfJSInterface)
```
gzm0, line 66: This would be `false` if `ParentJSType` were a class.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+      testJS(expected, value, classOf[js.Object], _.isInstanceOf[js.Object], _.asInstanceOf[js.Object])
+
+    testJSObject(true, new MyJSClass)
+    testJSObject(true, new MyJSSubClass)
+    testJSObject(true, new js.Object)
+    testJSObject(true, List(5))
+    testJSObject(true, new Exception)
+
+    // characters are always boxed
+    testJSObject(true, 'A')
```
gzm0, line 96: Is this specified behavior? If no, we probably shouldn't test it?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+
+    // longs are boxed iff they are not bigints
+    val longsAreBigInts = js.typeOf(5L) == "bigint"
+    testJSObject(!longsAreBigInts, 5L)
```
gzm0, line 100: Same here: Is this specified?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    testNullValue(classOf[js.Object], _.isInstanceOf[js.Object], _.asInstanceOf[js.Object])
+  }
+
+  @Test def jsError(): Unit = {
+    @inline def testJSError(expected: Boolean, value: Any): Unit =
+      testJS(expected, value, classOf[js.Error], _.isInstanceOf[js.Error], _.asInstanceOf[js.Error])
+
+    testJSError(true, new js.Error)
+    testJSError(true, new js.EvalError)
+    testJSError(true, new Exception)
```
gzm0, line 115: Maybe also test `new Throwable` to make it clear that this is the base case?

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  @js.native
+  trait ParentJSType extends js.Object
+
+  @js.native
+  trait SomeJSInterface extends ParentJSType
+
+  @JSGlobal("SomeJSClass")
+  @js.native
+  class SomeJSClass extends ParentJSType
```
gzm0, line 155: I do not understand the rationale of when native / non-native types are used here.

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
```diff
@@ -0,0 +1,163 @@
...
+import org.scalajs.testsuite.utils.AssertThrows._
+import org.scalajs.testsuite.utils.Platform._
+
+class RuntimeTypeTestsJSTest {
+  import RuntimeTypeTestsJSTest._
+
+  // Import the helpers from the non-JS test
+  import RuntimeTypeTestsTest._
+
+  @Test def scalaArraysOfJSTypes(): Unit = {
```
gzm0, line 30: Consider moving this to the end of the file. It seems to be the most complicated test.

gzm0 (changes requested):

test-suite/js/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsJSTest.scala
sjrd: There was no rationale. It was an artifact of merging tests from two different places. I have replaced with only using non-native types.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
```diff
@@ -0,0 +1,539 @@
...
+    testNullValue(classOf[String], _.isInstanceOf[String], _.asInstanceOf[String])
+  }
+
+  @Test def strictFloats(): Unit = {
+    assumeTrue("Assumed strict floats", hasStrictFloats)
+
+    @inline def testFloat(expected: Boolean, value: Any): Unit =
+      test(expected, value, classOf[Float], _.isInstanceOf[Float], _.asInstanceOf[Float])
+
+    testFloat(false, 1.2)
```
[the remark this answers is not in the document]
sjrd: `!executingInJVM` is used when it's true in JS and false on the JVM. Here's it's false on both.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
```diff
@@ -0,0 +1,539 @@
...
+
+  @Test def nonStrictFloats(): Unit = {
+    assumeFalse("Assumed non strict floats", hasStrictFloats)
+
+    @inline def testFloat(expected: Boolean, value: Any): Unit =
+      test(expected, value, classOf[Float], _.isInstanceOf[Float], _.asInstanceOf[Float])
+
+    testFloat(true, 1.2)
+
+    // from the bug report
```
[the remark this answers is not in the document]
sjrd: #1819. I added the reference.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
```diff
@@ -0,0 +1,539 @@
...
+
+  @noinline
+  def assertAlmostSame(msg: String, expected: Any, actual: Any): Unit = {
+    /* "Almost" same is like "same", except that we tolerate `equals` for
+     * boxed classes, since the instances could be unboxed and reboxed by the
+     * monomorphic code in the lambdas.
+     */
+
+    expected match {
+      case _:Boolean | _:Char | _:Byte | _:Short | _:Int | _:Long | _:Float | _:Double =>
```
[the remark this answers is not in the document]
sjrd: It's different from checking `isPrimitive()` because, if we get `expected.getClass()`, it's going to be the *boxed* class, which is not a primitive. So that doesn't help.

I have added a big comment to explain why it's fine.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
```diff
@@ -0,0 +1,494 @@
...
+    if (expected) {
+      assertAlmostSame(msg, value, asInstance(value))
+      if (!targetClass.isPrimitive())
+        assertAlmostSame(msg, value, targetClass.cast(value))
+    } else {
+      if (hasCompliantAsInstanceOfs) {
+        assertThrows(classOf[ClassCastException], asInstance(value))
+        assertThrows(classOf[ClassCastException], targetClass.cast(value))
+      }
+    }
```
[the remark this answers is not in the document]
sjrd: Gosh, yes I can! Good point!

sjrd (commented):
Updated.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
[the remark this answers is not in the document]
gzm0: I meant different than `targetClass.isPrimitive()`. For example: We use `assertAlmostSame` behind `if (!targetClass.isPrimitive())`. I do not understand when this is necessary (i.e. when is `targetClass.isPrimitive()` true, but the `assertEquals` branch is required).

Or another way of looking at this: If whether or not a thing we are testing (i.e. a `targetClass` we pass in) is primitive is statically known at call-site of `test`, we might want to avoid determining it dynamically.

Now, if it is not the case that `targetClass.isPrimitive()` iff when we need `assertEquals`, then this might result in  too many flags / arguments, so some dynamic determination might be better overall.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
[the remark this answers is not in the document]
sjrd: Ah, I see what you mean, now. Yes, it's true that it would be a bit nicer to integrate it in `primitiveTests`.

Except for the `match` test, but maybe we can just drop that one. It's a consequence of everything else, and there's no real reason that we're testing *that* case with a `match`, but none of the dozens of other cases.

test-suite/shared/src/test/scala/org/scalajs/testsuite/compiler/RuntimeTypeTestsTest.scala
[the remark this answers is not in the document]
gzm0: @sjrd did you see this comment? Otherwise LGTM.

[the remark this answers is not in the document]
sjrd: Ah no I had missed that.

It's not the same as `targetClass.isPrimitive()` either. The target class could be one of the boxed classes, or any of their super classes or super interfaces, and we could still have an expected value that is a primitive, requiring `assertEquals`. We're not testing a property of the target class here. It really is a property of the values we're comparing.

gzm0 (approved):

PR #4334, opened 2020-12-16
Fix #4322: Handle JSOptional default parameters in JS traits

### LINKED ISSUE #4322: JSOptional default parameters cause IR checking errors

gzm0:


gzm0:
Alternative to #4324.

sjrd (approved):

PR #4337, opened 2020-12-17
WiP Optimization: de-virtualize methods that can always be resolved.

sjrd:


sjrd:
Well ... this is spectacularly disappointing 😞 
queens
cd
Overall this "optimization" makes things worse more often than not. The exception is in Firefox, but only in the cases where we don't apply GCC. On Node.js it can make significantly worse.

I don't understand how VMs manage to perform worse with static, top-level functions, than with methods.

One thing I do take away: in all configurations, ES modules are faster than CommonJS modules, which are faster than Scripts. Too bad we can't apply GCC in ES modules.

sjrd:
So, I've redone more benchmarks, on Node.js and Firefox. Overall this seems to have no effect. Sometimes it improves a bit, some times it's a significantly worse; most of the time the results are very close. My conclusion is that devirtualization is performed well enough by GCC and by the engines themselves, and that therefore we shouldn't merge this PR.

Follows a small breakdown of all the benchmarks I have.

## Node.js

### A few benchmarks where things are better

#### brainfuck

brainfuck

#### tracer

tracer

#### sudoku

sudoku

#### list

list

#### sha512

sha512

#### sha512Int

sha512Int

#### Analysis

There are a number of cases where the performance is improved in dev mode (the SHA benchmarks are a remarkable example), or even in prod mode without GCC. Even in all of those cases, prod mode with GCC tends to erase any difference.

### The others

#### bounce

bounce

#### brainfuck

brainfuck

#### cd

cd

#### deltablue

deltablue

#### gcbench

gcbench

#### json

json

#### kmeans

kmeans

#### mandelbrot

mandelbrot

#### nbody

nbody

#### permute

permute

#### queens

queens

#### richards

richards

#### Analysis

In most cases, devirtualization doesn't change much, or even has an adverse effect, sometimes significant (gcbench is particularly damning).

## Firefox

In Firefox, I had to use a subset of our benchmarks, since we don't have performance counters as precise as on V8. The benchmarks that are too fast (< 10 ms per iteration) don't have measurements that are accurate enough.

#### brainfuck

firefox-brainfuck

#### cd

firefox-cd

#### gcbench

firefox-gcbench

#### kmeans

firefox-kmeans

### mandelbrot

firefox-mandelbrot

#### nbody

firefox-nbody

#### sha512

firefox-sha512

#### sha512Int

firefox-sha512Int

#### Analysis

In most cases, devirtualization has no effect. The only two benchmarks where we can observe some (small) improvement are brainfuck and cd. gcbench has the same huge regression when using GCC than on Node.js. The significant benefits that we saw on the SHA benchmarks on Node.js are not reproduced on Firefox.

PR #4341, opened 2020-12-21
Fix #4335: Call `cleanAfterRun` on LinkedJSMethodPropDefInfoCache

### LINKED ISSUE #4335: Refiner Info caching does not call `cleanAfterRun` on LinkedJSMethodPropDefInfoCache
Start with (post #4334)

```scala
package helloworld

import scala.scalajs.js
import js.annotation._

trait A extends js.Any {
}

object Impl extends js.Object with A {
}

object HelloWorld {
    println(Impl.foo())
  }
}
```

Prepare the following diff:

```diff
 }
 
 object Impl extends js.Object with A {
-  def foo(x: js.UndefOr[Int]): Int = x.getOrElse(5)
+  def foo(x: js.UndefOr[Int] = 6): Int = x.getOrElse(5)
 }
 
 object HelloWorld {
```

Run:

```sh
$  sbt
sbt:Scala.js> helloworld2_12/run
[info] Running helloworld.HelloWorld. Hit any key to interrupt.
5
// Apply change
sbt:Scala.js> helloworld2_12/run
[info] Compiling 1 Scala source to scala-js/examples/helloworld/.2.12/target/scala-2.12/classes ...
[info] Done compiling.
[info] Fast optimizing scala-js/examples/helloworld/.2.12/target/scala-2.12/helloworld-fastopt
[info] Running helloworld.HelloWorld. Hit any key to interrupt.
scala-js/examples/helloworld/.2.12/target/scala-2.12/helloworld-fastopt/main.js:1260
        var prep0 = ((rest[0] === (void 0)) ? $s_Lhelloworld_Impl$__foo$default$1__Lhelloworld_Impl$__sjs_js_$bar(this) : rest[0]);
                              ^

ReferenceError: $s_Lhelloworld_Impl$__foo$default$1__Lhelloworld_Impl$__sjs_js_$bar is not defined
    at $b_Lhelloworld_Impl$.foo (scala-js/examples/helloworld/src/main/scala/helloworld/HelloWorld.scala:16:7)
    at $c_Lhelloworld_HelloWorld$.main__AT__V (scala-js/examples/helloworld/src/main/scala/helloworld/HelloWorld.scala:21:21)
    at $s_Lhelloworld_HelloWorld__main__AT__V (scala-js/examples/helloworld/src/main/scala/helloworld/HelloWorld.scala:19:8)
    at scala-js/examples/helloworld/.2.12/target/scala-2.12/helloworld-fastopt/main.js:1827:1
    at scala-js/examples/helloworld/.2.12/target/scala-2.12/helloworld-fastopt/main.js:1828:4
    at Script.runInThisContext (vm.js:132:18)
    at Object.runInThisContext (vm.js:315:38)
    at [stdin]:8:25
    at Script.runInThisContext (vm.js:132:18)
    at Object.runInThisContext (vm.js:315:38)
```

Restarting `sbt` and deleting only the produced JS (no code is recompiled) produces the expected:

```sh
 $ sbt
sbt:Scala.js> helloworld2_12/run
[info] Fast optimizing scala-js/examples/helloworld/.2.12/target/scala-2.12/helloworld-fastopt
[info] Running helloworld.HelloWorld. Hit any key to interrupt.
6
[success] Total time: 6 s, completed Dec 16, 2020, 4:44:25 PM
```

gzm0:
I have manually verified that the test case fails without the fix.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/MemOutputDirectory.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/MemOutputDirectory.scala
@@ -25,6 +25,9 @@ import org.scalajs.linker.interface.unstable.OutputDirectoryImpl
 sealed trait MemOutputDirectory extends OutputDirectory {
   /** Content of the file with `name` or `None` if the file was not written. */
   def content(name: String): Option[Array[Byte]]
+
+  /** Names of files that are present (in no particular order). */
+  def fileNames(): List[String]
 }
 
 object MemOutputDirectory {
@@ -36,6 +39,10 @@ object MemOutputDirectory {
 
     def content(name: String): Option[Array[Byte]] = synchronized {
       _content.get(name)
+    }
+
+    def fileNames(): List[String] = synchronized {
+      _content.keys.toList
     }
 
     def writeFull(name: String, buf: ByteBuffer)(
@@ -56,7 +63,7 @@ object MemOutputDirectory {
     }
 
     def listFiles()(implicit ec: ExecutionContext): Future[List[String]] = synchronized {
-      Future.successful(_content.keys.toList)
+      Future.successful(fileNames())
     }
 
     def delete(name: String)(implicit ec: ExecutionContext): Future[Unit] = synchronized {
--- a/linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
+++ b/linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
@@ -1,0 +1,147 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.linker
+
+import scala.concurrent._
+
+import java.nio.charset.StandardCharsets.UTF_8
+
+import org.junit.Test
+import org.junit.Assert._
+
+import org.scalajs.ir.ClassKind
+import org.scalajs.ir.Names._
+import org.scalajs.ir.Trees._
+import org.scalajs.ir.Types._
+
+import org.scalajs.logging._
+
+import org.scalajs.junit.async._
+
+import org.scalajs.linker._
+import org.scalajs.linker.interface._
+
+import org.scalajs.linker.testutils._
+import org.scalajs.linker.testutils.TestIRBuilder._
+
+class IncrementalTest {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  import IncrementalTest._
+
+  @Test // #4335
+  def testChangeMethodReachedFromExport(): AsyncResult = await {
+
+    val FooClass = ClassName("Foo")
+
+    val jsMethodName = StringLiteral("foo")
+    val staticMethodName = m("value", Nil, IntRef)
+
+    def classDefs(pre: Boolean) = Seq(
+        mainTestClassDef(
+            consoleLog(JSMethodApply(LoadModule(FooClass), jsMethodName, Nil))
+        ),
+        classDef(
+            FooClass,
+            kind = ClassKind.ModuleClass,
+            superClass = Some(ObjectClass),
+            memberDefs = List(
+                trivialCtor(FooClass),
+                JSMethodDef(
+                    EMF, jsMethodName, Nil,
+                    if (pre) int(5)
+                    else ApplyStatic(EAF, FooClass, staticMethodName, Nil)(IntType))(
+                    EOH, None),
+                MethodDef(EMF.withNamespace(MemberNamespace.PublicStatic),
+                    staticMethodName, NON, Nil, IntType, Some(int(6)))(EOH, None)
+            )
+        )
+    )
+
+    testIncremental(classDefs(_), _ => MainTestModuleInitializers)
+  }
+
+}
+
+object IncrementalTest {
+
+  def testIncremental(
+      classDefs: Boolean => Seq[ClassDef],
+      moduleInitializers: Boolean => List[ModuleInitializer],
+      config: StandardConfig = StandardConfig())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    for {
+      _ <- testIncrementalStep(backward = false, classDefs, moduleInitializers, config)
+      _ <- testIncrementalStep(backward = true, classDefs, moduleInitializers, config)
+    } yield ()
+  }
+
+  private def testIncrementalStep(
+      backward: Boolean,
+      classDefs: Boolean => Seq[ClassDef],
+      moduleInitializers: Boolean => List[ModuleInitializer],
+      config: StandardConfig = StandardConfig())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+
+    val outputInc = MemOutputDirectory()
+    val outputBatch = MemOutputDirectory()
+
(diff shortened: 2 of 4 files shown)


linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+import org.scalajs.linker.testutils._
+import org.scalajs.linker.testutils.TestIRBuilder._
+
+class IncrementalTest {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  import IncrementalTest._
+
+  @Test // #4335
+  def testChangeMethodReachedFromExport(): AsyncResult = await {
```
gzm0, line 43: I tried to follow #4266 as good as I could for this, but it seems the location of `@Test` and how to reference issues (in the test name or comment) is not always the same.

linker/shared/src/main/scala/org/scalajs/linker/MemOutputDirectory.scala
```diff
@@ -25,6 +25,9 @@ import org.scalajs.linker.interface.unstable.OutputDirectoryImpl
 sealed trait MemOutputDirectory extends OutputDirectory {
   /** Content of the file with `name` or `None` if the file was not written. */
   def content(name: String): Option[Array[Byte]]
+
+  /** Names of files that are present (in no particular order). */
+  def fileNames(): List[String]
```
gzm0, line 30: This is a super-stable interface commitment. Just wanted to point this out.

project/BinaryIncompatibilities.scala
```diff
@@ -8,6 +8,10 @@ object BinaryIncompatibilities {
   )
 
   val Linker = Seq(
+      // New method in sealed trait, not an issue.
```
sjrd, line 11: I suggest always keeping the "breaking" exclusions on top (even the "OK" ones), and have the "not an issue" exclusions after them.

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private def assertOutputEquals(inc: MemOutputDirectory,
+      batch: MemOutputDirectory): Unit = {
+    val filesInc = inc.fileNames()
+    val filesBatch = batch.fileNames()
+
+    assertEquals(filesInc.toSet, filesBatch.toSet)
+
+    for (f <- filesInc) {
```
sjrd, line 135: Consider sorting the list before iterating, so that successive executions of a test reliably show the same error. When trying to fix an issue, if I run a test twice and it gives different errors, I'd like that to be due only to changes I made, and not to non-determinism.
```suggestion
    for (f <- filesInc.sorted) {
```

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  def testIncremental(
+      classDefs: Boolean => Seq[ClassDef],
+      moduleInitializers: Boolean => List[ModuleInitializer],
+      config: StandardConfig = StandardConfig())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    Future.sequence(Seq(
+        testIncrementalStep(backward = false, classDefs, moduleInitializers, config),
+        testIncrementalStep(backward = true, classDefs, moduleInitializers, config)
+    )).map(_ => ())
```
sjrd, line 85: Consider forcing the execution in order here, to reduce non-determinism in the test results:
```suggestion
    for {
        _ <- testIncrementalStep(backward = false, classDefs, moduleInitializers, config)
        _ <- testIncrementalStep(backward = true, classDefs, moduleInitializers, config)
    } yield ()
```

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+        classDef(
+            FooClass,
+            kind = ClassKind.ModuleClass,
+            superClass = Some(ObjectClass),
+            memberDefs = List(
+                trivialCtor(FooClass),
+                JSMethodDef(EMF, jsMethodName, Nil,
+                    if (pre) int(5)
+                    else ApplyStatic(EAF, FooClass, staticMethodName, Nil)(IntType)
+                )(EOH, None),
```
sjrd, line 63: Subtle formatting rule: if the `)` is on a new line, then there must be a newline after the `(`:
```suggestion
                JSMethodDef(
                    EMF, jsMethodName, Nil,
                    if (pre) int(5)
                    else ApplyStatic(EAF, FooClass, staticMethodName, Nil)(IntType)
                )(EOH, None),
```
That said, usually we format non-Block IR nodes using running arguments, including for the additional parameter lists:
```suggestion
                JSMethodDef(EMF, jsMethodName, Nil,
                    if (pre) int(5)
                    else ApplyStatic(EAF, FooClass, staticMethodName, Nil)(IntType))(
                    EOH, None),
```

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private def assertModulesEqual(inc: Iterable[Report.Module],
+      batch: Iterable[Report.Module]): Unit = {
+    // Poor man's equality based on toString()
+
+    def strs(ms: Iterable[Report.Module]) =
+      ms.map(m => m.moduleID -> m.toString()).toMap
+
+    assertEquals(strs(inc), strs(batch))
```
sjrd, line 125: Consider including the value of `backwards` in the associated `message` to `assertEquals`. Since this will be called twice with the same stack trace, it's easier to figure out where's the problem if the message says in which invocation we are.

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private def assertOutputEquals(inc: MemOutputDirectory,
+      batch: MemOutputDirectory): Unit = {
+    val filesInc = inc.fileNames()
+    val filesBatch = batch.fileNames()
+
+    assertEquals(filesInc.toSet, filesBatch.toSet)
```
sjrd, line 133: Same here about `backwards` in the message.

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
```diff
@@ -0,0 +1,142 @@
...
+
+    for (f <- filesInc) {
+      assertEquals(
```
sjrd, line 136: And here, ideally we'd have `backwards` *and* `f`.

sjrd (changes requested):
Cool. Thanks for the testing infrastructure here!

linker/shared/src/test/scala/org/scalajs/linker/IncrementalTest.scala
gzm0: I can do this if you want of course, but there isn't any non-determinism here (except for super fatal exceptions): `sequence` *does* return the results in order, it just executes them in parallel (which indeed might cause additional problems). Up to you.

sjrd: Ah so if both fail, it's always the failure of the first one that is reported? If yes, this is fine as  is.

gzm0: I have done it anyways because IMO it is much easier to understand.

sjrd (approved):

PR #4342, opened 2020-12-21
Improve/fix check for whether an $is_ function needs to be defined

gzm0:
Since 4f855d3 (Avoid fake classes by
constant-folding "impossible" instance tests), we would start defining
an (unused) $is_ function for classes that have no instances.

Instead of basing the check on the generated tree, we base it on the
same condition than the code that generates the call-sites.

As a further side-effect, this does not generate (unused) $is_
functions anymore for hijacked classes.

### DIFF
--- a/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
+++ b/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
@@ -803,13 +803,10 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
         }
       }
 
-      val needIsFunction = !isHijackedClass && (isExpression match {
-        case js.BinaryOp(JSBinaryOp.instanceof, _, _) =>
-          // This is a simple `instanceof`. It will always be inlined at call site.
-          false
-        case _ =>
-          true
-      })
+      val needIsFunction = !isHijackedClass && {
+        !tree.kind.isClass ||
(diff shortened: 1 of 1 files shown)


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/ClassEmitter.scala
```diff
@@ -832,13 +832,8 @@ private[emitter] final class ClassEmitter(sjsGen: SJSGen) {
...
 
-      val needIsFunction = !isHijackedClass && (isExpression match {
-        case js.BinaryOp(JSBinaryOp.instanceof, _, _) =>
-          // This is a simple `instanceof`. It will always be inlined at call site.
-          false
-        case _ =>
-          true
-      })
+      val needIsFunction = isHijackedClass || !tree.kind.isClass ||
+          globalKnowledge.isAncestorOfHijackedClass(className)
```
sjrd, line 836: That looks like it's going to generate an `$is_` function for hijacked classes, although they're not used. Wasn't this supposed to be
```suggestion
      val needIsFunction = !isHijackedClass && {
        !tree.kind.isClass ||
        globalKnowledge.isAncestorOfHijackedClass(className)
      }
```

gzm0: Yes :-/ Conditionals are hard...

sjrd (approved):

PR #4346, opened 2020-12-22
Fix #4339: Replace CallHelper transient with more specific trees

### LINKED ISSUE #4339: Revisit Optimizer / Emitter "CallHelper" interface
See https://github.com/scala-js/scala-js/pull/4321#r545087286

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/analyzer/Infos.scala
```diff
@@ -578,9 +578,27 @@ object Infos {
...
+                  traverse(num)
+                case ObjectClassName(obj) =>
+                  traverse(obj)
+                case Array2TypedArray(expr, primRef) =>
+                  traverse(expr)
+                case TypedArray2Array(expr, primRef) =>
+                  traverse(expr)
+                case _ =>
+                  throw new IllegalArgumentException("unexpected transient: " + t)
+              }
```
gzm0, line 601: @sjrd this makes me think: Maybe we should expect transients to also implement transformation and traversal?

sjrd: Yes, that is a good point. I was also thinking that the `TransientValue` should carry its `tpe: Type`, and that the `Transient` would derive its `tpe` from its value. I found it annoying to have to pass a computed `tpe` to the constructor of `Transient`, when in fact that was entirely determined by the underlying `TransientValue`.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/CoreJSLib.scala
```diff
@@ -1291,58 +1290,6 @@ private[emitter] object CoreJSLib {
       })
     }
 
-    private def defineTypedArrayConversions(): Unit = {
```
sjrd, line 1294: IMO this removal, and the corresponding inlining in `FunctionEmitter`, should be done in a separate commit. First only refactoring (with `FunctionEmitter` using `genCallHelper`), then doing the inlining.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -1205,6 +1223,14 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
         case IdentityHashCode(expr)  => test(expr)
         case GetClass(arg)           => test(arg) // may NPE but that is UB.
 
+        // Transients preserving pureness
+        case Transient(ZeroOf(runtimeClass)) =>
+          test(runtimeClass)
+        case Transient(NumberOfLeadingZeroes(num)) =>
+          test(num)
+        case Transient(ObjectClassName(obj)) =>
+          test(obj)
```
sjrd, line 1232: Add a comment `// may NPE but that is UB.` ?

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -1691,18 +1729,48 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
             redo(IdentityHashCode(newExpr))(env)
           }
 
-        case Transient(CallHelper(helper, args)) =>
-          unnest(args) { (newArgs, env) =>
-            redo(Transient(CallHelper(helper, newArgs)(rhs.tpe)))(env)
+        case Transient(SystemArrayCopy(src, srcPos, dest, destPos, length)) =>
```
sjrd, line 1732: Given that `SystemArrayCopy` has type `NoType`, it can only appear in statement position. I believe it should be handled in `transformStat`, instead of here. It should also not be considered an expression in `isExpressionInternal`.

sjrd (changes requested):

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Transients.scala
```diff
@@ -12,23 +12,152 @@
...
+      destPos: Tree, length: Tree)
       extends Transient.Value {
+    val tpe: Type = NoType
+
+    def traverse(traverser: Traverser): Unit = {
+      traverser.traverse(src)
+      traverser.traverse(srcPos)
+      traverser.traverse(dest)
+      traverser.traverse(destPos)
+      traverser.traverse(length)
```
sjrd, line 34: I've been thinking more about this, and I find myself bothered by the amount of boilerplate we have to cover for all those `Transient.Value`s. Here, but also in `FunctionEmitter` for `unnest` and `pushInto`. I believe there would be value in factoring out the commonalities of all these nodes.

There is one fundamental thing they have in common: they all have "method call" semantics:
* They have parameters, which are all expressions, and must be recursively transformed
* For all of them, parameters can be extracted in `val`s before passing them on, without altering the semantics

With those properties, we should be able to process them in a generic way in most places (except at the end, in `transformExpr`, where we actually compile them down). We should have some abstraction that allows us to do that. Perhaps it could look like this:
```scala
sealed abstract class ApplyIntrinsic(val args: List[Tree], val tpe: Type)
    extends Transient.Value {
  def traverse(traverser: Traverser): Unit =
    args.foreach(traverser.traverse(_))

  def transform(transformer: Transformer, isStat: Boolean)(
      implicit pos: Position): Tree = {
    Transient(withArgs(args.map(transformer.transformExpr(_))))
  }

  def printIR(out: IRTreePrinter): Unit = {
    out.print(nameString)
    out.printArgs(args)
  }

  def nameString: String

  def withArgs(newArgs: List[Tree]): IntrinsicApply
}
```
which would allow
```scala
final case class SystemArrayCopy(src: Tree, srcPos: Tree, dest: Tree,
      destPos: Tree, length: Tree)
      extends ApplyIntrinsic(List(src, srcPos, dest, destPos, length), NoType) {
  def nameString: String = "$systemArraycopy"

  def withArgs(args: List[Tree]): SystemArrayCopy = {
    // this would be safer with an HList, but we don't have that
    val List(src, srcPos, dest, destPos, length) = args
    SystemArrayCopy(src, srcPos, dest, destPos, length)
  }
}
```
and which we could process in a generic way in `unnest`, `pushLhsInto`, etc.

gzm0: Hum, but that argument applies to many of our normal IR nodes as well. Say, for example, all the nodes with `Apply` in the name. I'm not convinced that an abstraction unifying these things makes it easier to understand the code at the end.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
gzm0: That crashes the emitter :) Probably because the emitter assumes that the last element of a block is an expression. I'm not sure what is broken here: The optimizer that translates an `Apply` node to a statement or the emitter (maybe it should check the type of that last node in the block).

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Transients.scala
sjrd: Right, but at least for `Apply`s, we have a fixed number of them; there are 3, and there is no reason to have more than that, unless we introduce something fundamentally new in the IR. Here, there is a an arbitrary number of them, and they're often changing. It seems worthwhile to factor out what we can about those.

Also in the core Trees themselves, we have precedent: `BinaryOp`s could be a different node each, but it's not worth it because a lot can be handled in a generic way about them.

gzm0: First, some table setting.

> Right, but at least for Applys, we have a fixed number of them; there are 3,

There are 6 nodes with `Apply` in the name (and have the semantics you describe). Admittedly, the JS nodes are different because of the `TreeOrJSSpread`.

- Apply
- ApplyStatically
- ApplyStatic
- ApplyDynamicImport
- JSFunctionApply
- JSMethodApply

Further, the following nodes *also* have the semantics you describe (recursive transform required, extracting into vals preserves semantics) *and* could be replaced by function calls (trees that need additional non-tree parametrization, like for `ArrayToTypedArray` / `TypedArrayToArray` and the `Apply`s are marked with `*`).

- Throw
- New*
- StoreModule*
- Select*
- UnaryOp
- BinaryOp
- ArrayLength
- ArraySelect
- RecordSelect*
- IsInstanceOf*
- AsInstanceOf*
- GetClass
- IdentityHashCode
- JSImportCall
- JSDelete
- JSUnaryOp
- JSBinaryOp

This list does *not* include the degenerate case of zero nested trees. Then, there are even more nodes which correspond to the semantics you describe.

To make sense of what is happening here, I think we should look at each piece of introduced boilerplate individually.

**Transformers / Traversers**

All trees have this in common: They are expected to provide a way to transform/traverse an arbitrary subtree without a user knowing the specifics of a tree. For non-transients, this is implemented in ADT style, for transients in OO style.

IMO the reason we have not abstracted this away is because it is *difficult* in (non-meta) Scala (except for special cases like the Ops).

**Printers**

Similar than the previous point. More deduplication is possible; that is simply done in `Printers` itself.

Here, we could introduce a `printSpecial(name: String, args: Tree*)` or something, we could also use for `LinkingInfo`, `IdentityHashCode` (and maybe even `GetClass`, although this currently has method call syntax).

**FunctionEmitter: def unnest**

Recursive unnesting / traversal is (mostly) duplicated for the trees above (if they are expressions). The following are exceptions, but IIUC are oversights:

- GetClass
- IdentityHashCode
- JSImport

Note that the recursive unnesting has subtle differences with respect to pureness (both for Transients and non-Transients).

**FunctionEmitter: isExpressionInternal**

This actually depends on the semantics (pureness, side effects) of the individual node and is not always the same for the nodes in question (both for Transients and non-transients).

**FunctionEmitter: unnesting of arguments**

Recursive unnesting / traversal is duplicated for all the trees above.

**your other points**

I haven't addressed some of your other points:

> Here, there is a an arbitrary number of them, and they're often changing.

I have a feeling that we have introduced many more IR nodes in the recent past (~1 year) than `CallHelper`s. I'm happy to actually pull the numbers if you disagree.

> Also in the core Trees themselves, we have precedent: BinaryOps could be a different node each, but it's not worth it because a lot can be handled in a generic way about them.

Absolutely. However, I'd argue this is a good thing here, because it is easier to understand this way. I don't think this would be the case with the refactoring you propose (it'd be probably less code though).

My takeaway from this is that there is quite an amount of duplication / boilerplate, especially in `FunctionEmitter`. However, it is not specific to Transients but merely to how we deal with trees in general.

My gut feeling is that it might be possible to remove some of this boilerplate in `FunctionEmitter`. For the other things, I'm not convinced the effect will be net positive in terms of clarity. 

In any case, I don't think anything of this is really specific to Transients / this change.

sjrd: OK, that is a very convincing rebuttal. Let's keep the current design, then.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/SJSGen.scala
```diff
@@ -116,13 +116,18 @@ private[emitter] final class SJSGen(
...
-      Some(WithGlobals(VarRef(Ident(name)), Set(name)))
-
     elemTypeRef match {
       case _ if !esFeatures.useECMAScript2015 => None
+      case primRef: PrimRef                   => typedArrayRef(primRef)
+      case _                                  => None
+    }
+  }
+
+  def typedArrayRef(primRef: PrimRef)(
```
sjrd, line 126: I think this change (the extraction of `typedArrayRef` is in the wrong commit. IIUC it should be in the last commit instead.

sjrd (approved):
LGTM, other than my comment below.

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/SJSGen.scala
gzm0: Yeah, I noticed that as well in self-review :)

PR #4356, opened 2020-12-29
Fix #4350: Handle non-matching groups in Matcher#replace*.

### LINKED ISSUE #4350: Different behavior in Regex#replaceAllIn w.r.t. non-matching groups
I have a small case where Scala-JS (see [ScalaFiddle](https://scalafiddle.io/sf/CKvJ2Dq/1)) behaves differently than JVM (tested in Ammonite):
```scala
// JVM
println("(ab)|(a)".r.replaceAllIn("ab", "[1=$1][2=$2]")) 
[1=ab][2=]

// JS
println("(ab)|(a)".r.replaceAllIn("ab", "[1=$1][2=$2]")) 
[1=ab][2=null]
```
My naive guess is that the [`Matcher.group`](https://github.com/scala-js/scala-js/blob/5388c7b5cc1fa976a4e1b6741ec74716f9075083/javalib/src/main/scala/java/util/regex/Matcher.scala#L305) method returns `null` and `null` is formatted as `""` on JVM and `"null"` on JS.
If this is actually the problem, then it could be fixed by a rather precise change in `Matcher` that should not involve changes in code size.

Split out from https://github.com/scala-js/scala-js/issues/1201

@sjrd shall I give it a try?

nightscape:
This is a minimal PR to make the behavior of `Matcher#replace*` methods behave like their JVM counterpart by not appending the String `"null"` when a match group is `null`.
Fixes https://github.com/scala-js/scala-js/issues/4350.

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Matcher.scala
+++ b/javalib/src/main/scala/java/util/regex/Matcher.scala
@@ -97,7 +97,9 @@ final class Matcher private[regex] (
           while (i < len && isDigit(replacement.charAt(i)))
             i += 1
           val group = Integer.parseInt(replacement.substring(j, i))
-          sb.append(this.group(group))
+          val replaced = this.group(group)
+          if (replaced != null)
+            sb.append(replaced)
 
         case '\\' =>
           i += 1
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/regex/RegexMatcherTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/regex/RegexMatcherTest.scala
@@ -374,6 +374,13 @@ class RegexMatcherTest  {
     assertEquals("-foo-foo-foo-", matcher.replaceAll("-"))
   }
 
+  @Test def replaceAllIn(): Unit = {
+    val matcher = Pattern
+      .compile("(?:(ab)|(a))c(d)|(e)")
+      .matcher("abcd")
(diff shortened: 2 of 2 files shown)


javalib/src/main/scala/java/util/regex/Matcher.scala
```diff
@@ -97,7 +97,8 @@ final class Matcher private[regex] (
           while (i < len && isDigit(replacement.charAt(i)))
             i += 1
           val group = Integer.parseInt(replacement.substring(j, i))
-          sb.append(this.group(group))
+          val replaced = this.group(group)
+          if (replaced != null) sb.append(replaced)
```
sjrd, line 101: According to our coding style, side-effecting branches of an if must always be on their own line:
```suggestion
          if (replaced != null)
            sb.append(replaced)
```

sjrd (changes requested):
Thank you for the PR. The code and test look good. I only have one formatting comment below. Also, to match [our commit message guidelines](https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages), the issue number should be in the title. The title could also be a bit more precise in what it changes exactly. Perhaps consider

    Fix #4350: Handle non-matching groups in Matcher#replace*.

nightscape:
@sjrd I added your formatting suggestions and renamed the title of the commit and PR :+1:

sjrd (approved):
Thank you. LGTM!

nightscape:
@sjrd thanks for helping me get my first Scala-JS PR merged :smile:

PR #4358, opened 2021-01-05
Upgrade to Scala 2.13.4.

### LINKED ISSUE #4161: Reenable scala-test-suite tests that require AssertUtil in Scala 2.13.4
See https://github.com/scala-js/scala-js/pull/4159#pullrequestreview-473297304

gzm0:
Changes triggered by the upgrade:
- Remove a bunch of (trivial) Scaladoc links that break with the upgrade.
- Add an explicit unit literal result value to the constructors of DynamicImportThunks. Otherwise, this triggers a code path in Erasure intended for AnyVals only: https://github.com/scala/scala/blob/v2.13.4/src/compiler/scala/tools/nsc/transform/Erasure.scala#L439-L445.
- Adjust tests/error messages for JS constructor tags.

Further, this fixes #4161 (as expected).


scala-test-suite/src/test/resources/2.13.4/BlacklistedTests.txt
```diff
@@ -0,0 +1,216 @@
...
+scala/sys/process/ProcessBuilderTest.scala
+scala/sys/process/ProcessTest.scala
+scala/tools/testkit/AssertUtilTest.scala
+scala/util/PropertiesTest.scala
+scala/util/SpecVersionTest.scala
+scala/util/SystemPropertiesTest.scala
+
+# Test does not link (uses assertSameElements)
+# scala/scala@1390315b03bb32ffdfb1394b3ec6f4619faca1b8
+scala/collection/mutable/ListBufferTest.scala
```
gzm0, line 192: @sjrd not sure what to do about this. Seems like `assertSameElements` hasn't changed in a while, it just hasn't been used.

sjrd: What's the linking error, exactly? I guess we don't have `testkit/AssertUtil.scala` at all. We should probably have it. But otherwise, we can leave it in the "does not link" category.

gzm0: Oh right... The problem is `scala.runtime.ScalaRunTime.stringOf` which needs `getClassLoader` to do some generic (i.e. loader independent) subtyping checks.

sjrd (approved):
Thank you :)

sjrd:
The following scripted tests failed:
```
[error] 	scala3 / basic
[error] 	scala3 / junit
[error] 	cross-version / 2.11
```
I guess they still refer to 2.13.3, which is not published during `scriptedDependencies`, and so should be changed to 2.13.4.

gzm0:
I cannot reproduce the failures locally (with a clean `.ivy2/local`). Investigating.

gzm0:
I do not understand what is going on here. Locally, this works just fine :-/

gzm0:
It also seems to be happening on #4349 (the NPE, not the lock problem). Do we have a CI isolation issue?

sjrd:
Oh, it's those NPEs again. I think they're spurious. I see them from time to time, but usually the built-in second trial takes care of them. Perhaps more contention with more PRs is triggering them more often?



> Do we have a CI isolation issue?

We intentionally share the `cache` directories across CI runs on the same machine, otherwise we'd have to redownload the Internet on every build. Given that, I guess in theory it can happen that we get "No locks available" over that shared directory. (We do isolate the `local` directories.)

sjrd:
The errors in `scripted` are still there. It looks like it's not spurious for this PR.

gzm0:
Hmmm... So the only obvious thing that is different is the version of sbt (`1.4.2`). I guess sbt-dotty needs this. IIUC, `withDottyCompat` simply pulls the `2.13` artifact.

None of this explains why it fails with the upgrade though :-/

sjrd:
Meh. Perhaps try upgrading to `scalaVersion := "3.0.0-M3" and/or sbt 1.4.5 in those two tests?

gzm0:
Still happens. All I could find was: https://stackoverflow.com/questions/17676336/no-locks-available-when-run-sbt-cmd/27926593

sjrd:
Hum. sbt did change their default ivy handler to coursier in 1.3.x, and coursier uses a different cache directory. Perhaps we haven't properly set up the redirections for the coursier cache to use something on the localhome, as opposed to the NFS-mounted /home. I'll have a look.

sjrd:
Apparently we don't do anything for coursier:
https://github.com/scala-js/scala-js/blob/c0863e728cb7f6bcc0ba6eb514834979bf24639f/Jenkinsfile#L56-L66
We should probably add a `LOC_COURSIER_HOME="$LOC_SBT_BASE/sbt-coursier"`, and pass it to sbt with `-Dsbt.coursier.home=$LOC_COURSIER_HOME` (according to https://github.com/coursier/coursier/issues/783#issuecomment-533630773).

gzm0:
Huh, not the other way around? Like `sbt.ivy.home`? (so put in in `$(pwd)` instead of `/localhome` and symlink `cache` into it).

sjrd:
No, I don't think so, because for coursier there's only the `cache`. For Ivy we need that dance so that `local/` is in the test directory, but `cache/` goes to /localhome. For coursier there's no `local/`.

sjrd:
Apparently this still doesn't work.

I would normally not suggest that, but WDYT about disabling those Scala 3 tests so that we can release a new version with 2.13.4, and come back to this issue afterwards?

gzm0:
@sjrd I'm not sure I'll be able to rebase this in the next couple of days. If you want to unblock 1.4.0, feel free to rebase and merge. (I'll be able to do simple reviews).

sjrd:
OK, thanks for the info.

sjrd:
Rebased.

PR #4359, opened 2021-01-06
Remove test whitelists (and buglists)

gzm0:


partest/src/main/scala/scala/tools/partest/scalajs/ScalaJSPartest.scala
```diff
@@ -124,57 +124,41 @@ trait ScalaJSSuiteRunner extends SuiteRunner {
...
+    files.toSet
   }
 
-  private def extendShortTestName(testName: String) = {
-    val srcDir = PathSettings.srcDir
-    (srcDir / testName).toCanonical.getAbsolutePath
+  private def extendShortTestName(testName: String): File = {
+    val f = PathSettings.srcDir / testName
+    require(f.exists(), s"$testName does not exist")
+    f
```
sjrd, line 151: Dropping `toCanonical.getAbsolutePath` in `extendShortTestName` also has an impact on the `SomeTests` test filter. Did you check that we can still do

    > partestSuite testOnly -- -- run/sometest.scala

(hopefully also on Windows)

sjrd (commented):
I'll be a bit sad to see `--unknown` go, although the time it bought me will probably be compensated by the time not spent maintaining the whitelists. So overall this LGTM.

partest/src/main/scala/scala/tools/partest/scalajs/ScalaJSPartest.scala
gzm0: Yes, still works. In fact, we now get checking that the tests passed on the command line actually exist.

sjrd (approved):

PR #4360, opened 2021-01-06
Upgrade sbt to 1.3.13.

sjrd:
This is the latest (and very likely last) 1.3.x version.

Recent versions of Metals cannot import builds using sbt 1.2.x anymore. The initial import fails with a LinkageError because it wants to access a setting that was added in sbt 1.3.0.

This change may make our sbt plugin incompatible with sbt 1.2.x builds, as our main source of tests for it are our own build. However, almost all the scripted tests force to use sbt 1.2.8, and should give some confidence that we aren't breaking anything.

If we do, oh well, sbt 1.2.x is a bit old anyway.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -638,6 +638,9 @@ object Build {
   }
 
   val thisBuildSettings = Def.settings(
+      // Disable sbt 1.3.x+'s supershell, because it's terribly annoying
+      useSuperShell := false,
+
       // JDK version we are running with
       javaVersion in Global := {
         val fullVersion = System.getProperty("java.version")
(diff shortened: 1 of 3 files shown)


project/Build.scala
```diff
@@ -638,6 +638,9 @@ object Build {
   }
 
   val thisBuildSettings = Def.settings(
+      // Disable sbt 1.3.x+'s supershell, because it's terribly annoying
+      useSuperShell := false,
```
gzm0, line 642: Isn't this something that should go in the personal/local sbt config?

gzm0 (changes requested):

sjrd:
Turning into draft because it seems that things are recompiled every time with that version.

sjrd:
Asking on sbt's gitter: https://gitter.im/sbt/sbt?at=5ff6e122787d8f79c8e86eba

sjrd:
Superseded by #4516.

PR #4363, opened 2021-01-07
Fix #4362: Do not desugar spread operators in ES 2015.

### LINKED ISSUE #4362: JS constructors that require `new` cannot be called with expanded varargs (: _*).
The following test cases fail:
```scala

  @Test def newInstanceOfWithVarargsWhenConstructIsRequired_Issue4362(): Unit = {
    assumeTrue("requires the spread operator", assumeES2015)

    @noinline def args(): Seq[js.Any] = Seq(1234)

    val dateObj = js.Dynamic.newInstance(js.constructorOf[js.Date])(args(): _*)
    assertEquals(1234, dateObj.getTime())
  }

  @Test def callJSConstructorsWithVarArgsWhenConstructIsRequired_Issue4362(): Unit = {
    assumeTrue("requires the spread operator", assumeES2015)

    @noinline def args(): Seq[Any] = Seq(1234)

    val dateObj = new InteroperabilityTestJSDateWithVarArgsConstructor(args(): _*)
    assertEquals(1234, dateObj.getTime())
  }

@js.native
@JSGlobal("Date")
class InteroperabilityTestJSDateWithVarArgsConstructor(args: Any*) extends js.Object {
  def getTime(): Double = js.native
}
```
That's because we always desugar `JSSpread` into `apply` calls, and for constructors with use our internal `$newJSObjectWithVarargs` helper. Although the `apply` desugaring always works, the way `$newJSObjectWithVarargs` functions doesn't work if the target class must be called with `new`. `Date` is an example thereof, because when called as a function, it returns a string (essentially, `Date(x)` is equivalent to `new Date(x).toString()`).

We can only fix this in ECMAScript 2015, by using a true `...spread` operator for the arguments.

sjrd:


linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
```diff
@@ -1013,8 +1027,8 @@ private[emitter] class FunctionEmitter(sjsGen: SJSGen) {
                 NewArray(tpe, recs(lengths, allowUnpure))
               case ArrayValue(tpe, elems) =>
                 ArrayValue(tpe, recs(elems, allowUnpure))
-              case JSArrayConstr(items) if !containsAnySpread(items) =>
-                JSArrayConstr(recs(castNoSpread(items), allowUnpure))
+              case JSArrayConstr(items) if items.forall(_.isInstanceOf[Tree]) =>
+                JSArrayConstr(recs(items.asInstanceOf[List[Tree]], allowUnpure))
```
gzm0, line 1031: Humm... This is not nice. IIUC, this basically means we cannot always keep spreads in `JSArrayConstr`s in expression position. Maybe this warrants a comment?

gzm0 (commented):
This looks like it is the best we can do. But we see some ugly duplication (and in this case even different logic) between statement unnesting (`unnest`) and expression unnesting (`recs`).

IIUC, this means:

```javascript
// Stays like this
let x = [1, 2, ...y, 5, 6]

// Expands
f([1, 2, ...y, 5, 6]) 
// to
let tmp = [1, 2].concat(y).concat([5, 6])
f(tmp)
```

Not nice, but maybe the best we can do right now.

sjrd:
It seems we're blocked by GCC: https://github.com/google/closure-compiler/issues/3748

linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/FunctionEmitter.scala
sjrd: I managed to fix that. It wasn't that hard, actually.

gzm0 (approved):

sjrd:
Added a commit with a GCC upgrade that apparently fixes the issue.

gzm0 (approved):

sjrd:
@gzm0 That worked. Shall we go with that workaround for now?

gzm0:
Yeah, I guess as a stop gap solution it is OK. However, I think we should document better / more what behaves badly means. Notably:

- Which CI environment is this problematic in.
- what is failing
- what is the suspected cause
- what have we tried so far to fix it.

Thai information is important to avoid cargo culting going forward. (Also, TBH the current formulation feels quite unfair towards coursier, because AFAICT, the primary cause is our CI environment with a shared by default setup, as opposed to opt-in shared/cached stuff).

Alternatively to putting all this in a comment, we can also put it in an issue and just reference that. (Might avoid duplication of comments).

sjrd:
Yes, those are all good points, of course. I've filed #4368 and amended the commit to refer to it.

gzm0 (approved):

Issue #4315, opened 2020-12-04
Optimize dynamicImports of native calls


gzm0:
As discussed in https://github.com/scala-js/scala-js/pull/4262#issuecomment-738698144

gzm0:
To optimize the imports, I need to retrieve the load spec (if any) for classes referenced in a specific method body (in `computeNewAttributes`). But that makes the attributes depend on other things than the method body itself. Options I see:

- Mark the method attributes as "dirty" in this case and always recalculate them (even if the method's version didn't change).
- Defer retrieval to the optimization stage (and abort inlining if the referenced class isn't native or not an import).
- Track method/class dependencies and introduce a pre-UPDATE UPDATE pass.
- Do not do this and wait for somebody to complain.

@sjrd ideas, opinions?

sjrd:
IMO this:

> Defer retrieval to the optimization stage (and abort inlining if the referenced class isn't native or not an import).

is the best way to deal with this. Always dirty could have weird consequences. I'm not sure exactly what your third suggestion means.

Having the optimizer ask for that info and backtrack sounds reasonable.

gzm0:
> I'm not sure exactly what your third suggestion means.

What I meant by that is recording which methods (when calculating attributes) needed information about which classes. If these classes change, invalidate the method attributes. So an overall pass would look like:

- Diff class changes / invalidate method attributes 
- Re-calculate method attributes (currently called UPDATE pass)
- Re-optimize tagged methods (currently called PROCESS pass)

My concern with the defer retrieval is that we need to duplicate some tree matching between update and optimize :-/

sjrd:
> My concern with the defer retrieval is that we need to duplicate some tree matching between update and optimize :-/

To me it seems that that concern would be dwarfed by the added complexity (and impact on performance) of another entire pass.

gzm0:
Well, to be clear, the additional pass would only be on classes, not on the methods. But maybe that is fair. I'll fiddle with it more once I fixed the other mess :P

sjrd:
> would that be a viable option? I can recognize thunks by the common superclass (compiler runtime lib specific name, but we have a lot of precedent for runtime / optimizer collaboration like that).

[...]

PR #4364, opened 2021-01-07
Simplify Optimizer Parallelism Management

gzm0:


linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -66,19 +69,19 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
   private var batchMode: Boolean = false
 
   private var objectClass: Class = _
-  private val classes = CollOps.emptyMap[ClassName, Class]
+  private val classes = new ConcurrentHashMap[ClassName, Class]
```
sjrd: Scala's `mutable.TrieMap` is a fairly specific data structure, which is different from a `ConcurrentHashMap`. Are we sure that this change does not create performance regressions, notably in highly contended cases?

gzm0: I can run some basic benchmarks, but I don't have a beefy machine available (my dev machine only has 4 virtual cores). So it's unlikely we'd discover any serious contention issues.

gzm0: Alternatively, I can just keep `CollOps.Map`. Basically all other cleanups are still doable.

(I am concerned that this sets a precedent of "we are too afraid to change something", which is a dangerous place to be in. But that is a meta-point that allows for longer discussion; my main goal here is to unblock further work on #4315).

sjrd: I'm not opposed to the change. I just noticed that it was a real change and brought it up for consideration. At the end of the day, you are the parallelism expert in that area of the code, so if you think it's no big deal, I'm fine with it.

gzm0: Yeah, but now that you mentioned it, I *do* think we shouldn't just change it :P I'll try to not touch any of the parallelization properties of the code but just clean up a bit, if I'm able to do that, I'll settle for that.

gzm0:
Updated with better history for `GenIncOptimizer` and no need for `ConcurrentHashMap` anymore.

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
```diff
@@ -785,13 +829,15 @@ abstract class GenIncOptimizer private[optimizer] (config: CommonPhaseConfig) {
...
-  private[optimizer] abstract class MethodImpl(val owner: MethodContainer,
+  private[optimizer] final class MethodImpl(owner: MethodContainer,
       val methodName: MethodName)
       extends OptimizerCore.MethodImpl with OptimizerCore.AbstractMethodID
       with Unregisterable {
 
     private[this] var _deleted: Boolean = false
 
+    private val bodyAskers = CollOps.emptyMap[MethodImpl, Unit]
+    private val registeredTo = CollOps.emptyMap[Unregisterable, Unit]
```
sjrd: `registeredTo` was previously an `AtomicAcc[Unregisterable], i.e., an `Addable[Unregisterable]`. Shouldn't we keep using the `Addable` abstraction during this refactoring?

sjrd (changes requested):

linker/shared/src/main/scala/org/scalajs/linker/frontend/optimizer/GenIncOptimizer.scala
gzm0: Well, the `Addable` implements parallel traversal which comes at a cost of copying before traversing. So I'm not sure that is better. Also, it seems to me that `registeredTo` really should be a `Set` anyways...

sjrd: Right, that makes sense.

gzm0:
I have addressed all the comments.

sjrd (approved):

Issue #3503, opened 2018-12-05
Add java.util.WeakHashMap


jdegoes:
Could be based on `WeakMap`, supported since ECMAScript 6 (2015) and on all major browsers.

sjrd:
Unfortunately not. JavaScript's `WeakMap` lacks several fundamental features compared to Java's. In particular, it does not provide iteration of keys/entries (on purpose).

To implement `java.util.WeakHashMap`, we will need something more, which might be coming with a proposal for `WeakRef`s: https://github.com/tc39/proposal-weakrefs

tarsa:
Seems that WeakRefs are now present in majority (by market share) of browsers: https://www.caniuse.com/?search=weakref https://github.com/tc39/proposal-weakrefs/pull/212/files

sjrd:
[`WeakRef`s are now at Stage 4](https://github.com/tc39/proposal-weakrefs#status), so we could now implement this. PR welcome.

PR #4366, opened 2021-01-07
Add js.WeakRef and js.FinalizationRegistry

exoego:
This PR adds [WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) and [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry).

This could be a foundation to implement java.util.WeakHashMap #3503 or similar.

### DIFF
--- a/library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
+++ b/library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
@@ -1,0 +1,64 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** <span class="badge badge-ecma2021" style="float: right;">ECMAScript 2021</span>
+ *
+ *  A FinalizationRegistry object lets you request a callback when an object is garbage-collected.
+ *
+ *  FinalizationRegistry provides a way to request that a cleanup callback get called at some point
+ *  when an object registered with the registry has been reclaimed (garbage-collected).
+ *  (Cleanup callbacks are sometimes called finalizers.)
+ *
+ *  MDN
+ *
+ *  @tparam A Type of the target object to register
+ *  @tparam B Type of value to pass to the finalizer
+ *  @tparam C Type of an unregistration token
+ */
+@js.native
+@JSGlobal
+class FinalizationRegistry[-A, -B, -C](finalizer: js.Function1[B, scala.Any]) extends js.Object {
+  /** The `register` method registers an object with a FinalizationRegistry instance so that if
+   *  the object is garbage-collected, the registry's callback may get called.
+   *
+   *  @param theObject The target object to register.
+   *  @param heldValue The value to pass to the finalizer for this object. This cannot be the target object.
+   *  @param unregistrationToken
+   *    A token that may be used with the `unregister` method later to
+   *    unregister the target object. If provided (and not `undefined`),
+   *    this must be an object. If not provided, the target cannot be unregistered.
+   */
+  def register(theObject: A, heldValue: B, unregistrationToken: C): Unit = js.native
+
+  /** The `register` method registers an object with a FinalizationRegistry instance so that if
+   *  the object is garbage-collected, the registry's callback may get called.
+   *
+   *  @param theObject The target object to register.
+   *  @param heldValue The value to pass to the finalizer for this object. This cannot be the target object.
+   */
+  def register(theObject: A, heldValue: B): Unit = js.native
+
+  /** The `unregister`` unregisters a target object from a `FinalizationRegistry` instance.
+   *
+   *  When a target object has been reclaimed, it is no longer registered in the registry.
+   *  There is no need to all `unregister` in your cleanup callback. Only call `unregister` if
+   *  you haven't received a cleanup callback and no longer need to receive one.
+   *
+   *  @param unregistrationToken The token used with the register method when registering the target object.
+   */
+  def unregister(unregistrationToken: C): Unit = js.native
+}
--- a/library/src/main/scala/scala/scalajs/js/WeakRef.scala
+++ b/library/src/main/scala/scala/scalajs/js/WeakRef.scala
@@ -1,0 +1,33 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.scalajs.js
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** <span class="badge badge-ecma2021" style="float: right;">ECMAScript 2021</span>
+ *
+ *  A WeakRef object lets you hold a weak reference to another object, without preventing
+ *  that object from getting garbage-collected.
+ *
+ *  MDN
+ *
+ *  @param targetObject An object whose weak reference contained by an instance of WeakRef.
+ *  @tparam T unconstrained, but the constructor will throw a TypeError if targetObject is not an Object.
+ */
+@js.native
+@JSGlobal
+class WeakRef[+T](targetObject: T) extends js.Object {
+  /** The target object of the WeakRef, or undefined if the object has been garbage-collected. */
+  def deref[S >: T](): js.UndefOr[S] = js.native
+}
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
@@ -1,0 +1,36 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.library
+
+import org.junit.Test
+
+import scala.scalajs.js
+
+class FinalizationRegistryTest {
+  @Test def testMethods(): Unit = {
+    val registry = new js.FinalizationRegistry[js.Date, String, Any]((heldValue: String) => ())
+
+    val obj1 = new js.Date()
+    registry.register(obj1, "foo")
+
+    val obj2 = new js.Date()
+    registry.register(obj2, "bar", obj2)
+
+    val obj3 = new js.Date()
+    val unregisterToken = new js.Object()
+    registry.register(obj3, "bar", unregisterToken)
+
+    registry.unregister(obj1)
+    registry.unregister(unregisterToken)
+  }
+}
--- a/test-suite/js/src/test/scala/org/scalajs/testsuite/library/WeakRefTest.scala
+++ b/test-suite/js/src/test/scala/org/scalajs/testsuite/library/WeakRefTest.scala
@@ -1,0 +1,28 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.library
+
+import org.junit.Assert._
+import org.junit.Test
+
+import scala.scalajs.js
+
+class WeakRefTest {
+  @Test def testDeref(): Unit = {
+    val obj = new js.Date()
+    val ref = new js.WeakRef(obj)
+    val derefedInferredType = ref.deref()
+    val derefedAssertType: js.UndefOr[js.Date] = derefedInferredType
+    assertTrue(derefedAssertType.contains(obj))
+  }
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ version: '{build}'
 image: Visual Studio 2015
 environment:
   global:
-    NODEJS_VERSION: "12"
+    NODEJS_VERSION: "14"
     JAVA_HOME: C:\Program Files\Java\jdk1.8.0
     SCALA_VERSION: 2.12.12
 install:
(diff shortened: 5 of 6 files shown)


library/src/main/scala/scala/scalajs/js/WeakRef.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/** <span class="badge badge-ecma2021" style="float: right;">ECMAScript 2021</span>
+ *
+ * A WeakRef object lets you hold a weak reference to another object, without preventing
+ * that object from getting garbage-collected.
+ *
+ * MDN
+ *
+ * @param value
+ * @tparam T js.Object or AnyRef
```
gzm0, line 14: Interesting :) Will this make us commit to `AnyRef` being a reference for the JS GC?

sjrd: The relevant part of the spec is Step 2 of https://tc39.es/proposal-weakrefs/#sec-weak-ref-target:

> If Type(target) is not Object, throw a TypeError exception.

But there are `AnyVal`s that are guaranteed to be Objects when they reach JavaScript (any custom value class), and at the same time there are `js.Any`s that are primitives (`js.BigInt`, for example).

I don't think we should say "js.Object or AnyRef". We should let `T` be unconstrained, but document that the constructor will throw a TypeError if `Type(target)` is not Object. That is the only non-lying statement we can make.

exoego: Updated scaladoc and added header comment

library/src/main/scala/scala/scalajs/js/WeakRef.scala
```diff
@@ -0,0 +1,35 @@
...
+/** <span class="badge badge-ecma2021" style="float: right;">ECMAScript 2021</span>
+ *
+ * A WeakRef object lets you hold a weak reference to another object, without preventing
+ * that object from getting garbage-collected.
+ *
+ * MDN
+ *
+ * @param targetObject An object whose weak reference contained by an instance of WeakRef.
+ * @tparam T unconstrained, but the constructor will throw a TypeError if targetObject is not an Object.
+ */
```
sjrd, line 27:
```suggestion
/** <span class="badge badge-ecma2021" style="float: right;">ECMAScript 2021</span>
 *
 *  A WeakRef object lets you hold a weak reference to another object, without preventing
 *  that object from getting garbage-collected.
 *
 *  MDN
 *
 *  @param targetObject An object whose weak reference contained by an instance of WeakRef.
 *  @tparam T unconstrained, but the constructor will throw a TypeError if targetObject is not an Object.
 */
```

sjrd (commented):
* Could you add some smoke test in https://github.com/scala-js/scala-js/tree/master/test-suite/js/src/test/scala/org/scalajs/testsuite/library ? We can't test the finalization aspect, obviously, but we can test that the constructor exists, does something meaningful, and that `deref()` exists and returns the same object if called immediately after (this is guaranteed: WeakRefs can only obverse garbage collection between jobs).
* I believe we should add [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) at the same time, since together they form a unique proposal/API.

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /** The unregister unregisters a target object from a FinalizationRegistry instance.
+   *
+   *  When a target object has been reclaimed, it is no longer registered in the registry.
+   *  There is no need to all unregister in your cleanup callback. Only call unregister if
+   *  you haven't received a cleanup callback and no longer need to receive one.
+   *
+   *  @param unregistrationToken The token used with the register method when registering the target object.
+   */
+  def unregister(unregistrationToken: Any): Unit = js.native
```
gzm0, line 58: I wonder if we should make `FinalizationRegistry` parametric in the type of the `unregistrationToken`s as well :thinking:

test-suite/js/src/test/scala/org/scalajs/testsuite/library/WeakMapTest.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.scalajs.testsuite.library
+
+import org.junit.Assert._
+import org.junit.Test
+
+import scala.scalajs.js
+
+class WeakMapTest {
```
sjrd, line 20:
```suggestion
class WeakRefTest {
```
Also the name of the file ;)

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
sjrd: We might also have a type parameter for the type of objects that are registered, i.e., the type of `theObject`. I'm not sure whether going that far actually helps, though.

Regardless of what we choose to model with type parameters, they can all be contravariant, unless I'm mistaken.

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  MDN
+ *
+ *  @tparam A Type of the target object to register
+ *  @tparam B Type of value to pass to the finalizer
+ *  @tparam C Type of an unregistration token
+ */
+@js.native
+@JSGlobal
+class FinalizationRegistry[A, B, C](finalizer: js.Function1[B, Unit]) {
```
exoego, line 34:
```suggestion
class FinalizationRegistry[A, B, C](finalizer: js.Function1[B, Unit]) extends js.Object {
```

exoego:
@sjrd @gzm0 
This PR is ready to be reviewd.

sjrd:
I'll have a look this week.

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  MDN
+ *
+ *  @tparam A Type of the target object to register
+ *  @tparam B Type of value to pass to the finalizer
+ *  @tparam C Type of an unregistration token
+ */
+@js.native
+@JSGlobal
+class FinalizationRegistry[A, B, C](finalizer: js.Function1[B, Unit]) extends js.Object {
```
sjrd, line 34: We should use contravariant type parameters. In addition, there is no need to constrain the result type of `finalizer` to be Unit:
```suggestion
class FinalizationRegistry[-A, -B, -C](finalizer: js.Function1[B, scala.Any]) extends js.Object {
```
Using contravariant type params also helps with type inference, as `A` and `C` will be inferred to be `Any` by default, rather than `Nothing`.

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
```diff
@@ -0,0 +1,63 @@
...
+  /** The `register` method registers an object with a FinalizationRegistry instance so that if
+   *  the object is garbage-collected, the registry's callback may get called.
+   *
+   *  @param theObject The target object to register.
+   *  @param heldValue The value to pass to the finalizer for this object. This cannot be the target object.
+   *  @param unregistrationToken A token that may be used with the `unregister` method later to
+   *                             unregister the target object. If provided (and not `undefined`),
+   *                             this must be an object. If not provided, the target cannot be unregistered.
```
sjrd, line 42: I'm not sure whether it's documented anywhere, but the formatting for multi-line `@xyz` Scaladoc directives should be as follows:
```suggestion
   *  @param unregistrationToken
   *    A token that may be used with the `unregister` method later to
   *    unregister the target object. If provided (and not `undefined`),
   *    this must be an object. If not provided, the target cannot be unregistered.
```

library/src/main/scala/scala/scalajs/js/FinalizationRegistry.scala
```diff
@@ -0,0 +1,63 @@
...
+   *
+   *  @param theObject The target object to register.
+   *  @param heldValue The value to pass to the finalizer for this object. This cannot be the target object.
+   */
+  def register(theObject: A, heldValue: B): Unit = js.native
+
+  /** The `unregister`` unregisters a target object from a `FinalizationRegistry` instance.
+   *
+   *  When a target object has been reclaimed, it is no longer registered in the registry.
+   *  There is no need to all `unregister` in your cleanup callback. Only call unregister` if
```
sjrd, line 57:
```suggestion
   *  There is no need to all `unregister` in your cleanup callback. Only call `unregister` if
```

library/src/main/scala/scala/scalajs/js/WeakRef.scala
```diff
@@ -0,0 +1,35 @@
...
+@js.native
+@JSGlobal
+class WeakRef[T](targetObject: T) extends js.Object {
```
sjrd, line 30:
```suggestion
class WeakRef[+T](targetObject: T) extends js.Object {
```

library/src/main/scala/scala/scalajs/js/WeakRef.scala
sjrd: This still seems pending.

library/src/main/scala/scala/scalajs/js/WeakRef.scala
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * The target object of the WeakRef, or undefined if the object has been garbage-collected.
+   */
```
sjrd, line 33:
```suggestion
  /** The target object of the WeakRef, or undefined if the object has been garbage-collected. */
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.{BeforeClass, Test}
+import org.scalajs.testsuite.utils.Platform.executingInNodeJS
+
+import scala.scalajs.js
+import scala.scalajs.js.FinalizationRegistry
+
+object FinalizationRegistryTest {
+  @BeforeClass def assumeNodeJs(): Unit = {
+    assumeTrue("Assume Node.js", executingInNodeJS)
+  }
```
sjrd, line 26: This shouldn't be necessary. We're only testing specified behavior.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
```diff
@@ -0,0 +1,47 @@
...
+}
+
+class FinalizationRegistryTest {
+  @Test def testConstructor(): Unit = {
+    assertNotNull(new js.FinalizationRegistry[Any, String, Any]((heldValue: String) => {}))
+  }
```
sjrd, line 32: This seems completely redundant with what is tested in `testMethods()`. Consider dropping it.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  @Test def testMethods(): Unit = {
+    val registry = new FinalizationRegistry[js.Date, String, Any]((heldValue: String) => {})
```
sjrd, line 35: We usually prefer `()` over `{}` for a unit value:
```suggestion
    val registry = new FinalizationRegistry[js.Date, String, Any]((heldValue: String) => ())
```

test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    val obj = new js.Date()
+    val ungregisterToken = new js.Object()
+
+    registry.register(obj, "foo")
+    registry.register(obj, "bar", obj)
+    registry.register(obj, "bar", ungregisterToken)
```
sjrd, line 42: It seems weird to register the same `obj` several times. Consider using three different `obj1`, `obj2` and `obj3`.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/WeakRefTest.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Assert._
+import org.junit.Test
+
+import scala.scalajs.js
+
+class WeakRefTest {
+  @Test def testConstructor(): Unit = {
+    val obj = js.Object()
+    assertNotNull(new js.WeakRef[js.Object](obj))
+  }
```
sjrd, line 24: Similarly, this seems redundant with what's tested in `testDeref()`. Consider dropping it.

sjrd (changes requested):
Thanks. I have a few comments on style and variance.

exoego:
@sjrd Updated

library/src/main/scala/scala/scalajs/js/WeakRef.scala
```diff
@@ -0,0 +1,33 @@
...
+ *  that object from getting garbage-collected.
+ *
+ *  MDN
+ *
+ *  @param targetObject An object whose weak reference contained by an instance of WeakRef.
+ *  @tparam T unconstrained, but the constructor will throw a TypeError if targetObject is not an Object.
+ */
+@js.native
+@JSGlobal
+class WeakRef[+T](targetObject: T) extends js.Object {
```
sjrd, line 30: Erf. Apparently Scala 2.11 does not like that `+T` here, although 2.12 and 2.13 are fine with it. Let's make it invariant for now.

Otherwise LGTM. We'll need to wait for #4458 to be merged, for the version bump (it's a new API so it needs a minor release), but otherwise it's good to go.

test-suite/js/src/test/scala/org/scalajs/testsuite/library/FinalizationRegistryTest.scala
```diff
@@ -0,0 +1,36 @@
...
+
+class FinalizationRegistryTest {
+  @Test def testMethods(): Unit = {
+    val registry = new js.FinalizationRegistry[js.Date, String, Any]((heldValue: String) => ())
+
+    val obj1 = new js.Date()
+    registry.register(obj1, "foo")
+
+    val obj2 = new js.Date()
+    registry.register(obj2, "bar", obj1)
```
sjrd, line 27: Did you mean
```suggestion
    registry.register(obj2, "bar", obj2)
```
?

exoego: Fixed in https://github.com/scala-js/scala-js/pull/4366/commits/b226c06a9b417fae96da9aad4de173b40a50bdc9

library/src/main/scala/scala/scalajs/js/WeakRef.scala
exoego: Is this fix https://github.com/scala-js/scala-js/pull/4366/commits/51c9b9ad0fe6cf51594bc3b7daca476ee7225e26 fine ?

sjrd: Good question. I think it's fine, but we should test that calling `deref()` without type argument nor expected type infers the correct type. We typically do that in two steps, like this:
```scala
  @Test def testDeref(): Unit = {
    val obj = new js.Date()
    val ref = new js.WeakRef(obj)
    val derefedInferredType = ref.deref()
    val derefedAssertType: js.UndefOr[js.Date] = derefedInferredType
    assertTrue(derefedAssertType.contains(obj))
  }
```

---

Btw, #4458 has now been merged, so this PR can be rebased on master.

sjrd (commented):
LGTM. Could you rebase on master and squash into a single commit with a clean commit message, please? Then it's good to go.

exoego:
@sjrd Squashed commits

sjrd (approved):
Thank you!

PR #4367, opened 2021-01-09
Implement ConcurrentHashMap.newKeySet and use it in VersionChecks

### LINKED ISSUE #3882: Implement ConcurrentHashMap.newKeySet and use it in VersionChecks
See https://github.com/scala-js/scala-js/pull/3874#discussion_r348483358

exoego:
closes #3882

### DIFF
--- a/javalib/src/main/scala/java/util/NullRejectingHashMap.scala
+++ b/javalib/src/main/scala/java/util/NullRejectingHashMap.scala
@@ -57,6 +57,15 @@ private[util] class NullRejectingHashMap[K, V](
     super.put(key, value)
   }
 
+  override def putIfAbsent(key: K, value: V): V = {
+    if (value == null)
+      throw new NullPointerException()
+    val old = get(key) // throws if `key` is null
+    if (old == null)
+      super.put(key, value)
+    old
+  }
+
   @noinline
   override def putAll(m: Map[_ <: K, _ <: V]): Unit = {
     /* The only purpose of `impl` is to capture the wildcards as named types,
@@ -78,6 +87,37 @@ private[util] class NullRejectingHashMap[K, V](
     if (key == null)
       throw new NullPointerException()
     super.remove(key)
+  }
+
+  override def remove(key: Any, value: Any): Boolean = {
+    val old = get(key) // throws if `key` is null
+    if (old != null && old.equals(value)) { // false if `value` is null
+      super.remove(key)
+      true
+    } else {
+      false
+    }
+  }
+
+  override def replace(key: K, oldValue: V, newValue: V): Boolean = {
+    if (oldValue == null || newValue == null)
+      throw new NullPointerException()
+    val old = get(key) // throws if `key` is null
+    if (oldValue.equals(old)) { // false if `old` is null
+      super.put(key, newValue)
+      true
+    } else {
+      false
+    }
+  }
+
+  override def replace(key: K, value: V): V = {
+    if (value == null)
+      throw new NullPointerException()
+    val old = get(key) // throws if `key` is null
+    if (old != null)
+      super.put(key, value)
+    old
   }
 
   override def containsValue(value: Any): Boolean = {
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
@@ -61,8 +61,16 @@ class ConcurrentHashMap[K, V] private (initialCapacity: Int, loadFactor: Float)
   override def clear(): Unit =
     inner.clear()
 
-  override def keySet(): ConcurrentHashMap.KeySetView[K, V] =
-    new ConcurrentHashMap.KeySetView[K, V](inner.keySet())
+  override def keySet(): ConcurrentHashMap.KeySetView[K, V] = {
+    // Allow null as sentinel
+    new ConcurrentHashMap.KeySetView[K, V](this.inner, null.asInstanceOf[V])
+  }
+
+  def keySet(mappedValue: V): ConcurrentHashMap.KeySetView[K, V] = {
+    if (mappedValue == null)
+      throw new NullPointerException()
+    new ConcurrentHashMap.KeySetView[K, V](this.inner, mappedValue)
+  }
 
   override def values(): Collection[V] =
     inner.values()
@@ -198,39 +206,56 @@ object ConcurrentHashMap {
     }
   }
 
-  /* `KeySetView` is a public class in the JDK API. The result of
-   * `ConcurrentHashMap.keySet()` must be statically typed as a `KeySetView`,
-   * hence the existence of this class, although it forwards all its operations
-   * to the inner key set.
-   */
-  class KeySetView[K, V] private[ConcurrentHashMap] (inner: Set[K])
+  class KeySetView[K, V] private[ConcurrentHashMap] (innerMap: InnerHashMap[K, V], defaultValue: V)
       extends Set[K] with Serializable {
 
-    def contains(o: Any): Boolean = inner.contains(o)
+    def getMappedValue(): V = defaultValue
 
-    def remove(o: Any): Boolean = inner.remove(o)
+    def contains(o: Any): Boolean = innerMap.containsKey(o)
 
-    def iterator(): Iterator[K] = inner.iterator()
+    def remove(o: Any): Boolean = innerMap.remove(o) != null
 
-    def size(): Int = inner.size()
+    def iterator(): Iterator[K] = innerMap.keySet().iterator()
 
-    def isEmpty(): Boolean = inner.isEmpty()
+    def size(): Int = innerMap.size()
 
-    def toArray(): Array[AnyRef] = inner.toArray()
+    def isEmpty(): Boolean = innerMap.isEmpty()
 
-    def toArray[T <: AnyRef](a: Array[T]): Array[T] = inner.toArray[T](a)
+    def toArray(): Array[AnyRef] = innerMap.keySet().toArray()
 
-    def add(e: K): Boolean = inner.add(e)
+    def toArray[T <: AnyRef](a: Array[T]): Array[T] = innerMap.keySet().toArray(a)
 
-    def containsAll(c: Collection[_]): Boolean = inner.containsAll(c)
+    def add(e: K): Boolean = {
+      if (defaultValue == null) {
+        throw new UnsupportedOperationException()
+      }
+      innerMap.putIfAbsent(e, defaultValue) == null
+    }
 
-    def addAll(c: Collection[_ <: K]): Boolean = inner.addAll(c)
+    def containsAll(c: Collection[_]): Boolean = innerMap.keySet().containsAll(c)
 
-    def removeAll(c: Collection[_]): Boolean = inner.removeAll(c)
+    def addAll(c: Collection[_ <: K]): Boolean = {
+      if (defaultValue == null) {
+        throw new UnsupportedOperationException()
+      }
+      val iter = c.iterator()
+      var changed = false
+      while (iter.hasNext())
+        changed = innerMap.putIfAbsent(iter.next(), defaultValue) == null || changed
+      changed
+    }
 
-    def retainAll(c: Collection[_]): Boolean = inner.retainAll(c)
+    def removeAll(c: Collection[_]): Boolean = innerMap.keySet().removeAll(c)
 
-    def clear(): Unit = inner.clear()
+    def retainAll(c: Collection[_]): Boolean = innerMap.keySet().retainAll(c)
+
+    def clear(): Unit = innerMap.clear()
   }
 
+  def newKeySet[K](): KeySetView[K, Boolean] = newKeySet[K](HashMap.DEFAULT_INITIAL_CAPACITY)
+
+  def newKeySet[K](initialCapacity: Int): KeySetView[K, Boolean] = {
+    val inner = new InnerHashMap[K, Boolean](initialCapacity, HashMap.DEFAULT_LOAD_FACTOR)
+    new KeySetView[K, Boolean](inner, true)
+  }
 }
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
@@ -87,45 +87,17 @@ class ConcurrentHashMap[K, V] private (initialCapacity: Int, loadFactor: Float)
   override def equals(o: Any): Boolean =
     inner.equals(o)
 
-  override def putIfAbsent(key: K, value: V): V = {
-    if (value == null)
-      throw new NullPointerException()
-    val old = inner.get(key) // throws if `key` is null
-    if (old == null)
-      inner.put(key, value)
-    old
-  }
+  override def putIfAbsent(key: K, value: V): V =
+    inner.putIfAbsent(key, value)
 
-  override def remove(key: Any, value: Any): Boolean = {
-    val old = inner.get(key) // throws if `key` is null
-    if (old != null && old.equals(value)) { // false if `value` is null
-      inner.remove(key)
-      true
-    } else {
-      false
-    }
-  }
+  override def remove(key: Any, value: Any): Boolean =
+    inner.remove(key, value)
 
-  override def replace(key: K, oldValue: V, newValue: V): Boolean = {
-    if (oldValue == null || newValue == null)
-      throw new NullPointerException()
-    val old = inner.get(key) // throws if `key` is null
-    if (oldValue.equals(old)) { // false if `old` is null
-      inner.put(key, newValue)
-      true
-    } else {
-      false
-    }
-  }
+  override def replace(key: K, oldValue: V, newValue: V): Boolean =
+    inner.replace(key, oldValue, newValue)
 
-  override def replace(key: K, value: V): V = {
-    if (value == null)
-      throw new NullPointerException()
-    val old = inner.get(key) // throws if `key` is null
-    if (old != null)
-      inner.put(key, value)
-    old
-  }
+  override def replace(key: K, value: V): V =
+    inner.replace(key, value)
 
   def contains(value: Any): Boolean =
     containsValue(value)
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
@@ -232,6 +232,8 @@ object ConcurrentHashMap {
       innerMap.putIfAbsent(e, defaultValue) == null
     }
 
+    override def toString(): String = innerMap.keySet().toString
+
     def containsAll(c: Collection[_]): Boolean = innerMap.keySet().containsAll(c)
 
     def addAll(c: Collection[_ <: K]): Boolean = {
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
@@ -58,6 +58,8 @@ trait MapTest {
     assertEquals(null, mp.get(testObj(42)))
     if (factory.allowsNullKeysQueries)
       assertEquals(null, mp.get(null))
+    else
+      assertThrows(classOf[NullPointerException], mp.get(null))
   }
 
   @Test def testSizeGetPutWithStringsLargeMap(): Unit = {
@@ -199,6 +201,8 @@ trait MapTest {
     assertNull(mp.remove(testObj(42)))
     if (factory.allowsNullKeys)
       assertNull(mp.remove(null))
+    else
+      assertThrows(classOf[NullPointerException], mp.remove(null))
   }
 
   @Test def testRemoveWithInts(): Unit = {
@@ -1200,6 +1204,13 @@ trait MapTest {
       mp.put("nullable", null)
       assertNull(mp.putIfAbsent("nullable", "non null"))
       assertEquals("non null", mp.get("nullable"))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.putIfAbsent("abc", null))
+      assertThrows(classOf[NullPointerException], mp.putIfAbsent("new key", null))
+    }
+
+    if (!factory.allowsNullKeys) {
+      assertThrows(classOf[NullPointerException], mp.putIfAbsent(null, "def"))
     }
   }
 
@@ -1217,12 +1228,44 @@ trait MapTest {
     assertTrue(mp.remove("ONE", "one"))
     assertFalse(mp.containsKey("ONE"))
 
+    if (factory.allowsNullKeys) {
+      mp.put(null, "one")
+      assertFalse(mp.remove(null, "not exist"))
+      assertTrue(mp.containsKey(null))
+      assertTrue(mp.remove(null, "one"))
+      assertFalse(mp.containsKey(null))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.remove(null, "old value"))
+    }
+
     if (factory.allowsNullValues) {
       mp.put("nullable", null)
       assertFalse(mp.remove("nullable", "value"))
       assertTrue(mp.containsKey("nullable"))
       assertTrue(mp.remove("nullable", null))
       assertFalse(mp.containsKey("nullable"))
+    } else {
+      // mp#(key, null) should not remove. https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6272521
+      assertFalse(mp.remove("THREE", null))
+    }
+  }
+
+  @Test def testUnconditionalRemove(): Unit = {
+    val mp = factory.fromKeyValuePairs("ONE" -> "one", "TWO" -> "two", "THREE" -> "three")
+
+    assertEquals(null, mp.remove("non existing"))
+    assertFalse(mp.containsKey("non existing"))
+
+    assertEquals("two", mp.remove("TWO"))
+    assertEquals(null, mp.get("TWO"))
+
+    if (factory.allowsNullKeys) {
+      mp.put(null, "one")
+      assertTrue(mp.containsKey(null))
+      assertEquals("one", mp.remove(null))
+      assertFalse(mp.containsKey(null))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.remove(null))
     }
   }
 
@@ -1251,6 +1294,17 @@ trait MapTest {
     } else {
       assertThrows(classOf[NullPointerException], mp.replace("ONE", null, "one"))
       assertThrows(classOf[NullPointerException], mp.replace("ONE", "four", null))
+    }
+
+    if (factory.allowsNullKeys) {
+      assertFalse(null, mp.replace(null, "value", "new value"))
+      assertFalse(mp.containsKey(null))
+
+      mp.put(null, "null value")
+      assertTrue(mp.replace(null, "null value", "new value"))
+      assertEquals("new value", mp.get(null))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.replace(null, "one", "two"))
     }
   }
 
@@ -1418,6 +1472,15 @@ trait MapTest {
       assertEquals("def", mp.merge("nullable", "def", notCalled))
       assertEquals("def", mp.get("nullable"))
     }
+  }
+
+  @Test def additionToKeySet(): Unit = {
+    val set = factory.empty[String, String].keySet()
+
+    expectThrows(classOf[UnsupportedOperationException], set.add("ONE"))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList("ONE")))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList(null)))
+    expectThrows(classOf[UnsupportedOperationException], set.add(null))
   }
 }
 
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
@@ -215,6 +215,18 @@ class ConcurrentHashMapTest extends MapTest {
     assertFalse(keySet.add("ONE"))
     assertEquals("Adding an existing key does not alter the value", 42, map.get("ONE"))
   }
+
+  @Test def toStringOnKeySetView(): Unit = {
+    val map = factory.empty[String, Int]
+    val keySet = map.keySet(0)
+
+    map.put("a", 0)
+    assertEquals("[a]", keySet.toString)
+
+    map.put("b", 1)
+    val str = keySet.toString
+    assertTrue(s"toString should print keys, but actual: $str", str == "[a, b]" || str == "[b, a]")
+  }
 }
 
 class ConcurrentHashMapFactory extends ConcurrentMapFactory {
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
@@ -194,6 +194,27 @@ class ConcurrentHashMapTest extends MapTest {
         assertTrue(s"missing value '$value'", encounteredValues.contains(value))
     }
   }
+
+  @Test def keySetWithNullMappedValue(): Unit = {
+    val map = factory.empty[String, String]
+    assertThrows(classOf[NullPointerException], map.keySet(null))
+  }
+
+  @Test def addOnKeySetView(): Unit = {
+    val map = factory.empty[String, Int]
+    val keySet = map.keySet(0)
+    assertNull(map.get("ONE"))
+    assertTrue(keySet.add("ONE"))
+    assertEquals("Adding a new key adds the default value.", 0, map.get("ONE"))
+
+    assertFalse(keySet.add("ONE"))
+
+    assertEquals(0, map.put("ONE", 42))
+    assertEquals("Putting alters the value", 42, map.get("ONE"))
+
+    assertFalse(keySet.add("ONE"))
+    assertEquals("Adding an existing key does not alter the value", 42, map.get("ONE"))
+  }
 }
 
 class ConcurrentHashMapFactory extends ConcurrentMapFactory {
--- a/ir/shared/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
+++ b/ir/shared/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala
@@ -52,14 +52,14 @@ class VersionChecks private[ir] (
   }
 
   private val knownSupportedBinary = {
-    val m = new ConcurrentHashMap[String, Unit]()
-    m.put(binaryEmitted, ())
+    val m = ConcurrentHashMap.newKeySet[String]()
+    m.add(binaryEmitted)
     m
   }
 
   /** Check we can support this binary version (used by deserializer) */
   final def checkSupported(version: String): Unit = {
-    if (!knownSupportedBinary.containsKey(version)) {
+    if (!knownSupportedBinary.contains(version)) {
       val (major, minor, preRelease) = parseBinary(version)
       val supported = (
           // the exact pre-release version is supported via knownSupportedBinary
@@ -70,7 +70,7 @@ class VersionChecks private[ir] (
       )
 
       if (supported) {
-        knownSupportedBinary.put(version, ())
+        knownSupportedBinary.add(version)
       } else {
         throw new IRVersionNotSupportedException(version, binaryEmitted,
             s"This version ($version) of Scala.js IR is not supported. " +
--- a/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/ConcurrentHashMapKeySetViewTest.scala
+++ b/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/ConcurrentHashMapKeySetViewTest.scala
@@ -1,0 +1,31 @@
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package org.scalajs.testsuite.javalib.util
+
+import java.{util => ju}
+import scala.reflect.ClassTag
+
+class ConcurrentHashMapKeySetViewTest extends SetTest {
+  override def factory: ConcurrentHashMapKeySetViewFactory = new ConcurrentHashMapKeySetViewFactory
+}
+
+class ConcurrentHashMapKeySetViewFactory extends SetFactory {
+  override def implementationName: String =
+    "java.util.ConcurrentHashMap.KeySetView"
+
+  override def allowsNullElementQuery: Boolean = false
+  override def allowsNullElement: Boolean = false
+
+  override def empty[E: ClassTag]: ju.Set[E] =
+    ju.concurrent.ConcurrentHashMap.newKeySet[E]()
+}


javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -203,34 +203,62 @@ object ConcurrentHashMap {
    * hence the existence of this class, although it forwards all its operations
    * to the inner key set.
    */
-  class KeySetView[K, V] private[ConcurrentHashMap] (inner: Set[K])
+  class KeySetView[K, V] private[ConcurrentHashMap] (innerMap: ConcurrentHashMap[K, V], defaultValue: V)
       extends Set[K] with Serializable {
 
-    def contains(o: Any): Boolean = inner.contains(o)
+    def getMappedValue(): V = defaultValue
```
exoego, line 209: getMappedValue is a public member of KeySetView according to https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.KeySetView.html#getMappedValue--

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,30 @@ class ConcurrentHashMapTest extends MapTest {
         assertTrue(s"missing value '$value'", encounteredValues.contains(value))
     }
   }
+
+  @Test def testNewKeySet(): Unit = {
```
gzm0, line 198: Instead of this, use the tests in `SetTest`?

gzm0 (changes requested):

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Added https://github.com/scala-js/scala-js/pull/4367/commits/7c6cd273e35808874878876d8828ae05a0064bf7

I will remove this tests after the `SetTest`-based tests passed.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -203,34 +203,62 @@ object ConcurrentHashMap {
...
-    def remove(o: Any): Boolean = inner.remove(o)
+    def contains(o: Any): Boolean = innerMap.inner.containsKey(o)
 
-    def iterator(): Iterator[K] = inner.iterator()
+    def remove(o: Any): Boolean = if (innerMap.inner.containsKey(o)) {
+      innerMap.inner.remove(o)
+      true
+    } else {
+      false
+    }
```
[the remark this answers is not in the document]
exoego: Done in https://github.com/scala-js/scala-js/pull/4367/commits/6a1dd4b2dcd8a002acfa72c3e479fb37baf8e149

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -203,34 +203,62 @@ object ConcurrentHashMap {
...
-    def add(e: K): Boolean = inner.add(e)
+    def toArray[T <: AnyRef](a: Array[T]): Array[T] = innerMap.inner.keySet().toArray(a)
 
-    def containsAll(c: Collection[_]): Boolean = inner.containsAll(c)
+    def add(e: K): Boolean = if (innerMap.containsKey(e)) {
+      false
+    } else {
+      innerMap.put(e, defaultValue)
+      true
+    }
```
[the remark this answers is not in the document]
exoego: Done in https://github.com/scala-js/scala-js/pull/4367/commits/aeadef3145f2cceebf73cf2d9724d8c4a67345c4

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -203,34 +203,62 @@ object ConcurrentHashMap {
    * hence the existence of this class, although it forwards all its operations
    * to the inner key set.
    */
-  class KeySetView[K, V] private[ConcurrentHashMap] (inner: Set[K])
+  class KeySetView[K, V] private[ConcurrentHashMap] (innerMap: ConcurrentHashMap[K, V], defaultValue: V)
```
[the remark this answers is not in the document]
exoego: I think constructor requires 2nd parameter, because
* The instance method `ConcurrentHashMap#keySet()` passes `null.asInstanceOf[V]` to it
* The static method `newKeySet[K](): KeySetView[K, Boolean]` passes `true` to it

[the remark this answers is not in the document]
gzm0: Yes, that is correct. But why are you passing a `ConcurrentHashMap` as the first parameter and not a `InnerHashMap`? This seems unnecessary indirection.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,38 @@ class ConcurrentHashMapTest extends MapTest {
...
+    assertTrue(keySet.contains("c"))
+    assertFalse(keySet.contains("d"))
+
+    assertThrows(classOf[UnsupportedOperationException], keySet.contains(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.containsAll(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.remove(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.retainAll(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.removeAll(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.add(null))
+    assertThrows(classOf[UnsupportedOperationException], keySet.addAll(null))
```
gzm0, line 227: Are you sure these test cases are correct? [`ConcurrentHashMap.containsKey`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#containsKey-java.lang.Object-) should throw an NPE in this case.

exoego:
Hmm, `sbt testSuite2_13/test` failed due to missing `newKeySet`, which is added in this PR.

```
[error] C:\projects\scala-js\test-suite\shared\src\test\scala\org\scalajs\testsuite\javalib\util\ConcurrentHashSetTest.scala:60:41: type newKeySet is not a member of object java.util.concurrent.ConcurrentHashMap
[error]     new ju.concurrent.ConcurrentHashMap.newKeySet[E]()
[error]                                         ^
[error] one error found
```

exoego:
It is werid that `ConcurrentHashMapKeySetViewTest` is failing...

```
val hs = factory.empty[String]

assertEquals(0, hs.size) // pass
assertTrue(hs.add("ONE")) // fail with AssertionError: null
```

I do not figure out where `AssertionError` is coming from.

gzm0:
`AssertionError: null` may mean that the assert itself is failing (i.e. `add` returns `false` instead of `true`). The `null` simply means the `message` field is `null`. Which happens if `assertTrue` doesn't have a second `msg` parameter.

gzm0:
Re: `keySet()` without parameters: It seems this is specified in `Map` and quite well tested in `MapTest`. However, that `add` and `addAll` throw an `UnsupportedOperationException` is not tested there, but it probably should. So IMO the best course of action is to add these tests there (and fix offending implementations if any).

exoego:
> So IMO the best course of action is to add these tests there (and fix offending implementations if any).

Added tests and fixed implementations.

I will squash commits into single comment when PR get approved.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -199,38 +199,65 @@ object ConcurrentHashMap {
...
+        throw new UnsupportedOperationException()
+      }
+      val iter = c.iterator()
+      var addedAny: Boolean = false
+      while (iter.hasNext()) {
+        val nextKey = iter.next()
+        if (!innerMap.containsKey(nextKey)) {
+          innerMap.put(nextKey, defaultValue)
+          addedAny = true
+        }
```
gzm0, line 241: `addedAny ||= innerMap.putIfAbsent(iter.next(), defaultValue)`?

gzm0, line 241: `addedAny ||= innerMap.putIfAbsent(e, defaultValue)`?

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -199,38 +199,65 @@ object ConcurrentHashMap {
...
+      }
+      innerMap.putIfAbsent(e, defaultValue) == null
+    }
 
-    def containsAll(c: Collection[_]): Boolean = inner.containsAll(c)
+    def containsAll(c: Collection[_]): Boolean = innerMap.keySet().containsAll(c)
 
-    def addAll(c: Collection[_ <: K]): Boolean = inner.addAll(c)
+    def addAll(c: Collection[_ <: K]): Boolean = {
+      if (c == null || defaultValue == null) {
```
gzm0, line 231: Same here about the `null` check. Just leave it to `c.iterator()`.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -199,38 +199,65 @@ object ConcurrentHashMap {
...
 
-    def toArray(): Array[AnyRef] = inner.toArray()
+    def toArray(): Array[AnyRef] = innerMap.keySet().toArray()
 
-    def toArray[T <: AnyRef](a: Array[T]): Array[T] = inner.toArray[T](a)
+    def toArray[T <: AnyRef](a: Array[T]): Array[T] = innerMap.keySet().toArray(a)
 
-    def add(e: K): Boolean = inner.add(e)
+    def add(e: K): Boolean = {
+      if (e == null || defaultValue == null) {
```
gzm0, line 222: I'm pretty positive the `null` check here is wrong. If `c` is `null`, this should throw an `NPE`, not an `UnsupportedOperationException`.

I'd just leave the `null` check to `putIfAbsent`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,17 @@ class ConcurrentHashMapTest extends MapTest {
...
+  @Test def addIsUnsupportedOnKeySet(): Unit = {
+    val chm = factory.empty[String, String]
+    val set = chm.keySet()
+
+    expectThrows(classOf[UnsupportedOperationException], set.add("ONE"))
+    expectThrows(classOf[UnsupportedOperationException], set.add(null))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(null))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList(null)))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList("ONE")))
+  }
```
gzm0, line 207: This should go to `MapTest`.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,17 @@ class ConcurrentHashMapTest extends MapTest {
         assertTrue(s"missing value '$value'", encounteredValues.contains(value))
     }
   }
+
```
gzm0, line 197: There should be at least one test case checking the `add` functionality of `ConcurentHashMap.KeySetView`. Notably the following behavior should be tested:

- Adding a new key adds the default value.
- Adding an existing key does *not* alter the value (iff this is the JVM behavior, the doc is IMO vague about it).

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -199,38 +199,65 @@ object ConcurrentHashMap {
...
 
-    def retainAll(c: Collection[_]): Boolean = inner.retainAll(c)
+    def retainAll(c: Collection[_]): Boolean = innerMap.keySet().retainAll(c)
 
-    def clear(): Unit = inner.clear()
+    def clear(): Unit = innerMap.clear()
+
+    def getMappedValue(): V = defaultValue
+
+    override def toString: String = innerMap.keySet().toString
```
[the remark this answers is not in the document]
exoego: Filed in https://github.com/scala-js/scala-js/issues/4395
I will open a separate small PR for this.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Hmm, I tried the below 

```scala
      while (iter.hasNext()) {
        val nextKey = iter.next()
        addedAny ||= innerMap.putIfAbsent(nextKey, defaultValue) == null
      }
```
but this made some tests failing.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Hmm, I do not understand.

`KeySetView[K, V]` have only `getMappedValue` to return `V`.
And `KeySetView` made from `ConcurrentHashMap` do not support `add/addAll`.
So, I do not think it is possible to check

* the default value is added when a new key is added
* the value is not altered when an exsting key is added

gzm0: Oh, that is supported if the overload specific to `ConcurrentHashMap` of `keySet` is used: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#keySet-V-

exoego: It seems that any addition to the `keySet` always throws Exception.
https://github.com/scala-js/scala-js/pull/4367/commits/d636996881b26218886a3f03952105e710b408a9

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Moved to `MapTest` https://github.com/scala-js/scala-js/pull/4367/commits/d636996881b26218886a3f03952105e710b408a9

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Ah, do you mean using private constructor `KeySetView(map, defaultValue)` ?

exoego: Ah, I see.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Hmm, but removing `e == null` fails on `SetTest#contains`.
It seems that `putIfAbsent` on InnerHashMap throw nothing on `null` key.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -1419,6 +1419,23 @@ trait MapTest {
...
+
+    expectThrows(classOf[UnsupportedOperationException], set.add("ONE"))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList("ONE")))
+    expectThrows(classOf[UnsupportedOperationException], set.addAll(ju.Arrays.asList(null)))
+    expectThrows(classOf[UnsupportedOperationException], set.add(null))
+
+    if (factory.allowsNullCollectionsAdding) {
+      expectThrows(classOf[UnsupportedOperationException], set.addAll(null))
+    } else {
+      // NPE on JVM, JSE on JS
```
gzm0, line 1435: If the JVM exposes different behavior when passing `null` collection, we should probably just drop this specific test (it's arguably a degenerate case).

@sjrd for a second opinion.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -1419,6 +1419,23 @@ trait MapTest {
       assertEquals("def", mp.get("nullable"))
     }
   }
+
+  @Test def additionToKeySet(): Unit = {
+    val chm = factory.empty[String, String]
+    val set = chm.keySet()
```
gzm0, line 1425: Simply `val set = factory.empty[String, String].keySet()`? (`chm` is unused later).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,34 @@ class ConcurrentHashMapTest extends MapTest {
...
+  @Test def addOnKeySetView(): Unit = {
+    val map = factory.empty[String, Int]
+    val keySet = map.keySet(0)
+    assertNull(map.get("ONE"))
+    assertTrue(keySet.add("ONE"))
+    assertEquals("Adding a new key adds the default value.", 0, map.get("ONE"))
+
+    assertFalse(keySet.add("ONE"))
+    assertEquals("Adding an existing key do not alter the value", 0, map.get("ONE"))
```
gzm0, line 206: The message here is misleading: Even *if* the value was altered, it would still be `0`. Consider simply dropping this assert.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,34 @@ class ConcurrentHashMapTest extends MapTest {
...
+
+    assertEquals(0, map.put("ONE", 42))
+    assertEquals("Putting alter the value", 42, map.get("ONE"))
```
gzm0, line 209: Nit: alter*s*

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,34 @@ class ConcurrentHashMapTest extends MapTest {
...
+
+    assertFalse(keySet.add("ONE"))
+    assertEquals("Adding an existing key do not alter the value", 42, map.get("ONE"))
```
gzm0, line 212: Nit: ... key do*es* not ...

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,34 @@ class ConcurrentHashMapTest extends MapTest {
...
+
+  @Test def toStringOnKeySetView(): Unit = {
+    val map = factory.empty[String, Int]
+    val keySet = map.keySet(0)
+
+    map.put("a", 0)
+    assertEquals("[a]", keySet.toString)
+
+    map.put("b", 1)
+    assertEquals("[a, b]", keySet.toString)
```
gzm0, line 223: This test is likely overly restrictive: IIUC there is not guarantee of any particular key ordering when iterating over a `KeySetView`. So either `"[a, b]"` or `"[b, a]"` would be correct in this case.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: Hah, indeed. It seems like `ConcurrentHashMap` explicitly fixes this behavior. Was this an oversight in f7be410503f87b402a86c92945dba37455f06853? @sjrd ? If yes, maybe this should be moved to `NullRejectingHashMap`.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: Potentially the same issue than for `add`.

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
sjrd: Ah yes, well spotted. Indeed, introducing the optimized implementation of `putIfAbsent` in `HashMap` bypassed the null checks that `NullRejectingHashMap` performs. So indeed, we should override `putIfAbsent` in `NullRejectingHashMap` to perform the null checks before calling `super.putIfAbsent` (similarly to `put`).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
sjrd: Yes, this is full-blown Undefined Behavior, because the *collection* itself is null, not its elements. This test should be dropped (also the "then" branch, and therefore the `allowsNullCollectionsAdding` factory flag).

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -62,7 +62,10 @@ class ConcurrentHashMap[K, V] private (initialCapacity: Int, loadFactor: Float)
     inner.clear()
 
   override def keySet(): ConcurrentHashMap.KeySetView[K, V] =
-    new ConcurrentHashMap.KeySetView[K, V](inner.keySet())
+    keySet(null.asInstanceOf[V])
```
gzm0, line 65: Actually, this usage seems to be not allowed: Doc says that `keySet(V)` throws NPE if `mappedValue` is `null`. We should probably honor that specification to avoid an accidental `null` to be interpreted as a sentinel.

gzm0, line 65: @sjrd is usage of `null` like this problematic if `V` is a primitive type?

sjrd: No, because it won't escape the generic environment, so it will always remain in places that erase to `Object`.

If it *did* escape into a monomorphic context, it would unbox to the zero of the target type.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: The below tests are failing, even after fixed `NullRejectingHashMap.putIfAbsent`.

```
[error] Test org.scalajs.testsuite.javalib.util.ConcurrentHashMapKeySetViewTest.simpleSum failed: java.lang.AssertionError: expected:<268> but was:<42>, took 0.001999872 sec
...
[error] Test org.scalajs.testsuite.javalib.util.ConcurrentHashMapKeySetViewTest.toStringCollectionCustomClass failed: java.lang.AssertionError: result '[Custom(A,1)]' not in expected set 'HashSet([Custom(b,2), Custom(A,1), Custom(C,3)], [Custom(b,2), Custom(C,3), Custom(A,1)], [Custom(C,3), Custom(b,2), Custom(A,1)], [Custom(C,3), Custom(A,1), Custom(b,2)], [Custom(A,1), Custom(b,2), Custom(C,3)], [Custom(A,1), Custom(C,3), Custom(b,2)])', took 0.004 sec
...
[error] Test org.scalajs.testsuite.javalib.util.ConcurrentHashMapKeySetViewTest.toStringCollectionDoubleHasCommaSpace failed: java.lang.AssertionError: result '[88.42]' not in expected set 'HashSet([60.173, 88.42, -23.36], [88.42, -23.36, 60.173], [-23.36, 60.173, 88.42], [60.173, -23.36, 88.42], [88.42, 60.173, -23.36], [-23.36, 88.42, 60.173])', took 0.001000192 sec
...
[error] Test org.scalajs.testsuite.javalib.util.ConcurrentHashMapKeySetViewTest.removeIf failed: java.lang.AssertionError: expected:<8> but was:<1>, took 0 sec
...
[error] Test org.scalajs.testsuite.javalib.util.ConcurrentHashMapKeySetViewTest.iterator failed: java.lang.AssertionError: expected:<3> but was:<1>
```

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/4367/commits/afbb9c57f285c8d60f66625178d97a4df1071d69

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Removed in https://github.com/scala-js/scala-js/pull/4367/commits/1940ba1fc571b40be360be116f980a7f99d9a62b

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/4367/commits/873d05ed0a4fd4c60f549c666da8f20abf3515ef

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/4367/commits/873d05ed0a4fd4c60f549c666da8f20abf3515ef

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/4367/commits/4c21243134bca29258c37d1618657d561e6941cd

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
exoego: Removed assertion.
But it seems that `coll.addAll(nulll)` should throw NPE according to JavaDoc.
So filed https://github.com/scala-js/scala-js/issues/4399

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Removed

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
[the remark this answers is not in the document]
exoego: Fixed

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Fixed in https://github.com/scala-js/scala-js/pull/4367/commits/16687e688329b3ed7c612f33e030517278d10a2a

gzm0:
What is the status on this? @exoego is there anything you need from our end or are you unblocked?

exoego:
I think that all mentioned issues were addressed.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: Why are they failing? Also, I just realize that using `add` might even be a better option. Like here: https://github.com/scala-js/scala-js/blob/5cc821bd51ba49d7ee33e5a2dffa44636864da41/javalib/src/main/scala/java/util/HashSet.scala#L57-L63

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
gzm0: This isn't addressed. The entire `if` and `allowsNullCollectionsAdding` should be removed.

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Hmm,

```scala
val map = new java.util.concurrent.ConcurrentHashMap[String,String]()
val set = map.keySet(null)
```
does not throw NPE.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Ah, that one using `add` worked well.

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
```diff
@@ -57,6 +57,12 @@ private[util] class NullRejectingHashMap[K, V](
     super.put(key, value)
   }
 
+  override def putIfAbsent(key: K, value: V): V = {
+    if (key == null)
+      throw new NullPointerException()
```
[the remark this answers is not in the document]
exoego: Added assertions https://github.com/scala-js/scala-js/pull/4367/commits/fa1678d61c4b6f19dfee95000e95c28b9ff135b0
It seems that `putIfAbsent` already throws when `null` value.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: Ah, that's my bad :) The problem was short-circuiting behavior of `||`.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: That clearly contradicts the documentation: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#keySet-V-

On JDK 11, it does throw for me. If if doesn't in earlier JVM versions, we should add an `assumeTrue` for the JDK.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -1200,6 +1200,12 @@ trait MapTest {
       mp.put("nullable", null)
       assertNull(mp.putIfAbsent("nullable", "non null"))
       assertEquals("non null", mp.get("nullable"))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.putIfAbsent("abc", null))
```
gzm0, line 1204: Also add a case where the key isn't present yet?

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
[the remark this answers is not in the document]
gzm0: Well, in `ConcurrentHashMap` it does, because you it still has the old implementation rather than forwarding to this one.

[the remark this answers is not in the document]
gzm0: In fact, now that I look at it, the implementation should be moved for `remove` and `replace` as well.

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Addressed in  https://github.com/scala-js/scala-js/pull/4367/commits/1e8f329d2492c2a496cbab2d68e5a0f3d67714c9

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
```diff
@@ -33,6 +35,10 @@ private[util] class NullRejectingHashMap[K, V](
     putAll(m)
   }
 
+  override def compute(key: K, remappingFunction: BiFunction[_ >: K, _ >: V, _ <: V]): V = {
+    super.compute(key, remappingFunction)
+  }
+
```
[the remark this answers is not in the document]
exoego: Oops.
Removed in https://github.com/scala-js/scala-js/pull/4367/commits/9191b6c746d6b84f742ddc430e0632fae75e7db2

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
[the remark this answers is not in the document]
exoego: Added assertions in https://github.com/scala-js/scala-js/pull/4367/commits/d1aa9be1f7713c983a60c2baebaba10444c4d27d
It seems that `remove` and `replace` already throws  NPE, so overriding is seems not required.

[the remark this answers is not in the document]
gzm0: They only throw an NPE because that's what the implementation in ConcurrentHashMap does. You need to change the implementations in ConcurrentHashMap to simply forward to NullRejectingHashMap. (that applies to `putIfAbsent`, `remove` and `replace`). Then you'll see that `remove` and `replace` also need updating.

gzm0 (changes requested):

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
```diff
@@ -12,6 +12,8 @@
 
 package java.util
 
+import java.util.function.BiFunction
+
```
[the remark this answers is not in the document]
exoego: Removed in https://github.com/scala-js/scala-js/pull/4367/commits/1c26ee4aa9149ac1133d51228219ab64944f872f

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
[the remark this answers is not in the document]
exoego: Moved implementation to NullRejectingHashMap in https://github.com/scala-js/scala-js/pull/4367/commits/811bd513c9b9a461ddb88969460173663d981de3

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -1217,12 +1227,43 @@ trait MapTest {
...
+    }
+
     if (factory.allowsNullValues) {
       mp.put("nullable", null)
       assertFalse(mp.remove("nullable", "value"))
       assertTrue(mp.containsKey("nullable"))
       assertTrue(mp.remove("nullable", null))
       assertFalse(mp.containsKey("nullable"))
+    } else {
+      assertThrows(classOf[NullPointerException], mp.remove(null, "old value"))
```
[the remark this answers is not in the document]
exoego: Removed assertions for NPE https://github.com/scala-js/scala-js/pull/4367/commits/b42213dd1509be1cc154a5738e11c6c4ae4078ec
Both [JavaDoc JDK 8](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#remove-java.lang.Object-java.lang.Object-) and [JavaDoc JDK 11](
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html#remove(java.lang.Object,java.lang.Object) does not mention `chm#remove(key, null)` should throw NPE.

Instead, it seem that `chm#remove(key, null)` should return `false`.
https://github.com/scala-js/scala-js/pull/4367/commits/01ac6d2bf060895b8fd5be3645d60989439c365b

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +196,42 @@ class ConcurrentHashMapTest extends MapTest {
         assertTrue(s"missing value '$value'", encounteredValues.contains(value))
     }
   }
+
+  @Test def keySetWithMappedValue(): Unit = {
```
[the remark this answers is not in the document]
exoego: FIxed so in https://github.com/scala-js/scala-js/pull/4367/commits/78fd5a98f32ff8ee132accfc1df326fa33139d0d

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +196,42 @@ class ConcurrentHashMapTest extends MapTest {
...
+    assumeTrue("Assumed JS or JDK>8", !executingInJVM || !executingInJVMOnJDK8OrLower)
```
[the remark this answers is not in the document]
exoego: https://github.com/scala-js/scala-js/pull/4367/commits/21ff1ea6db82f7b27b0db10e247a9b13abe24d3b
Removed `assume`, since `keySet(null)` throwing NPE on JDK 8 on my end too.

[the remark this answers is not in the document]
exoego: Same as https://github.com/scala-js/scala-js/pull/4367#discussion_r578378572

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
```diff
@@ -80,6 +89,37 @@ private[util] class NullRejectingHashMap[K, V](
...
+  override def replace(key: K, value: V): V = {
+    if (key == null || value == null)
+      throw new NullPointerException()
+    val old = super.get(key)
+    if (old != null)
+      super.put(key, value)
+    old
+  }
+
```
gzm0, line 122: Some of the `null` handling is inconsistent here: Sometimes it is delegated to other methods (notably `get`) when possible, sometimes not. Have you considered simply moving the implementations that are currently in ConcurrentHashMap.

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
```diff
@@ -80,6 +89,37 @@ private[util] class NullRejectingHashMap[K, V](
     super.remove(key)
   }
 
+  override def remove(key: Any, value: Any): Boolean = {
+    val old = this.get(key) // throws if `key` is null
```
gzm0, line 93: We usually do not qualify `this` if not necessary.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
```diff
@@ -1217,12 +1228,48 @@ trait MapTest {
...
       assertFalse(mp.remove("nullable", "value"))
       assertTrue(mp.containsKey("nullable"))
       assertTrue(mp.remove("nullable", null))
       assertFalse(mp.containsKey("nullable"))
+    } else {
+      // Currently, ConcurrentHashMap is the only null-rejecting Map implemented in Scala.js.
+      // JavaDoc does not describe that CHM#remove(key, null) should throw NPE, and actually
+      // they do not throw NPE both on JDK 8 & 11. So we do not assert NPE thrown.
+      // However, it should be asserted that mp#(key, null) should not remove.
+      // https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6272521
```
gzm0, line 1252: This is not the right comment style. Mutli line comments should use: `/*` (see the coding style guidelines).

I also don't necessarily agree with your interpretation that the spec is ambiguous (or at least this is how I read the text). IMO if it *doesn't* say it throws, it shouldn't. So maybe, the comment can simply be removed.

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -19,10 +19,12 @@ import java.{util => ju}
 import java.util.concurrent.{ConcurrentHashMap => CHM}
 
 import org.junit.Assert._
+import org.junit.Assume._
 import org.junit.Test
 
 import org.scalajs.testsuite.javalib.util.MapTest
 import org.scalajs.testsuite.utils.AssertThrows._
+import org.scalajs.testsuite.utils.Platform._
```
gzm0, line 27: These imports are now unused.

gzm0 (changes requested):

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
exoego: CHM now simply delegets to NullRejectingHashMap

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Deleted https://github.com/scala-js/scala-js/pull/4367/commits/88509a247e7afbf21fff0f12de96c1d4160feac3

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/MapTest.scala
exoego: Updated https://github.com/scala-js/scala-js/pull/4367/commits/cd4ce0f68f72fcfce8b0bb21100bba50c97f6dbe

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
exoego: Removed `this.` qualifier https://github.com/scala-js/scala-js/pull/4367/commits/9e530f3c125e9a1af2d9f12e238a1f5eea3f5f55

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
gzm0: Good. But the `null` handling is still inconsistent: Some methods delegate to `get` to check if they `key` is non-null, some don't.

For simplicity: Simply move the implementations from `ConcurrentHashMap`?

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -19,6 +19,7 @@ import java.{util => ju}
 import java.util.concurrent.{ConcurrentHashMap => CHM}
 
 import org.junit.Assert._
+import org.junit.Assume._
```
gzm0, line 22: This import is unused.

gzm0 (changes requested):

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
exoego: Removed https://github.com/scala-js/scala-js/pull/4367/commits/0eb6a4b9292ebea8e6883759e56119111c994c2f

javalib/src/main/scala/java/util/NullRejectingHashMap.scala
exoego: Delegate null check to `(this.)get|` or other method as much s possible. https://github.com/scala-js/scala-js/pull/4367/commits/f12100cd44447c752fc3402e05438eab68e12ed8

gzm0: Now some of the comments are gone. Please just take the implementations in CHM.

exoego: https://github.com/scala-js/scala-js/pull/4367/commits/fa361b53962fb1b4a686f8a8093a61c04aa42a3b

gzm0:
LGTM! go for squash :)

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -198,39 +178,58 @@ object ConcurrentHashMap {
...
 
-    def retainAll(c: Collection[_]): Boolean = inner.retainAll(c)
+    def addAll(c: Collection[_ <: K]): Boolean = {
+      if (defaultValue == null) {
+        throw new UnsupportedOperationException()
+      }
+      val iter = c.iterator()
+      var changed = false
+      while (iter.hasNext())
+        changed = innerMap.putIfAbsent(iter.next(), defaultValue) == null || changed
```
exoego, line 218: @gzm0 
This was `changed = add(iter.next() == null || changed` before squash.
I think null check on `defaultValue` was redundant here, so rewrote.

gzm0: yep, good find.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -79,45 +87,17 @@ class ConcurrentHashMap[K, V] private (initialCapacity: Int, loadFactor: Float)
...
-  override def putIfAbsent(key: K, value: V): V = {
-    if (value == null)
-      throw new NullPointerException()
-    val old = inner.get(key) // throws if `key` is null
-    if (old == null)
-      inner.put(key, value)
-    old
-  }
+  override def putIfAbsent(key: K, value: V): V =
+    inner.putIfAbsent(key, value)
```
gzm0, line 91: These changes should go into the next commit (all the null check moves should).

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
gzm0: Hum, did this line also change? `e == null` here is definitely wrong. If `e == null`, this should throw an NPE (but `innerMap.putIfAbsent` will already do so).

test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/concurrent/ConcurrentHashMapTest.scala
```diff
@@ -194,6 +194,40 @@ class ConcurrentHashMapTest extends MapTest {
         assertTrue(s"missing value '$value'", encounteredValues.contains(value))
     }
   }
+
+  @Test def keySetWithNullMappedValue(): Unit = {
+    val map = factory.empty[String, String]
+    assertThrows(classOf[NullPointerException], map.keySet(null))
+  }
+
+
```
gzm0, line 203: Double blank linke

gzm0 (changes requested):
Re commit messages:

- Use `Fix #xxx:`, not `Close #xxx:`.
- Message for #4395 should qualify `toString`. (`Fix #4395: ConcurrentHashMap.KeySetView.toString should print elements`)

exoego:
@gzm0 I think all issues on previous squash were addressed.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -198,39 +206,58 @@ object ConcurrentHashMap {
...
 
-    def add(e: K): Boolean = inner.add(e)
+    def add(e: K): Boolean = {
+      if (defaultValue == null) {
+        throw new UnsupportedOperationException()
+      }
+      innerMap.putIfAbsent(e, defaultValue) == null
+    }
 
     def containsAll(c: Collection[_]): Boolean = inner.containsAll(c)
```
gzm0: This line should be removed in this commit.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -232,7 +232,7 @@ object ConcurrentHashMap {
       innerMap.putIfAbsent(e, defaultValue) == null
     }
 
-    def containsAll(c: Collection[_]): Boolean = inner.containsAll(c)
```
gzm0, line 225: This diff needs to go in the previous commit.

gzm0 (changes requested):

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
exoego: Ah, sorry. Fixed

gzm0 (approved):

sjrd:
And we're down under the bar of 40 open issues! 🎉 Thank you @exoego !

PR #4371, opened 2021-01-13
Update Scala 2.12.12 to 2.12.13

sh0hei:


sjrd:
Thank you for your interest in contributing to Scala.js.

I'm afraid that upgrading to a new version of Scala 2is much more involved than this. To give you an idea, here is the commit that upgraded to 2.12.12:
https://github.com/scala-js/scala-js/pull/4159/commits/c22effda18e75346fe74f02e9e9a6b1f1221beae

Notably, this involves running the new partest suite, and deciding for each test whether it should be blacklisted, whitelisted or reported as a bug. This is not easy.

sjrd:
Superseded by #4472.

PR #4374, opened 2021-01-13
Fix #4372: Generalize js.Function's to allow user-defined ones.

### LINKED ISSUE #4372: Proposal: Generalize js.Function's so that we can make custom ones
## Background

Currently, the way we can create JS functions in the language is through a fixed set of `js.FunctionN` and `js.ThisFunctionN` interfaces. They are `@js.native`, so they cannot be extended by non-native JS classes and traits, but the compiler has special support for *SAMs* of those 23+22 = 45 built-in types.

## Problem

The current situation allows to define JS `function` and `=>` lambdas with a *fixed* number of arguments, not exceeding 22. It does not allow to define any of the following, however:

* A JS `=>` function with a `...rest` parameter (or with N fixed parameters and a `...rest` parameter)
* An `=>` function with more than 22 parameters
* Technically we can define `function` functions through a complicated hoop: make it a method of a non-native JS class, that is directly instantiated, and then grab its method as a `js.Dynamic` value. That's not nice at all, though
* A JS `function` that inspects its `new.target` meta-property

If all we needed was the ability to define a JS `=>` function with a `...rest` parameter (and no fixed parameter), we could add one more built-in type. But this won't scale to N fixed parameters and a `...rest` parameter, or to a whole new set of built-in functions for taking `new.target`.

## Proposal

The gist of the proposal is: allow users to define their own `js.FunctionN` traits. Once we can do that, if we need, for example, an arrow function one fixed Int param and a `...rest` param of type T*, we can declare:
```scala
trait MyJSFunction[-T, +R] extends js.Function {
  def apply(first: Int, args: T*): R
}
```
and use it as:
```scala
val f: MyJSFunction[String, Int] = { (first, args) =>
  first + args.map(_.length()).sum
}
```
For a `function` function, the trait would extend `js.ThisFunction` instead, and take an additional first parameter for the `this` value.

To allow the above, we need to make a few changes to the language and compiler:

1. Allow declaring *abstract* `apply` methods in non-native JS classes and traits. It's only the concrete ones that we cannot implement. An abstract one disappears in the IR.
2. Apply the JS `Closure` treatment in the back-end to any anonymous function class *whose direct superclass* is `js.Function`. Note that this includes `extends js.Function1` or `extends MyJSFunction` because they're traits that themselves extend `js.Function`, which is a class. Therefore, any class that extend one of those ends up having `js.Function` as `superClass`.
3. In that treatment, make a `function` and use a `this` parameter if the anonymous function class inherits from `js.ThisFunction`.
4. Handle repeated parameters in the treatment as well.

In addition, we would remove the `@js.native` annotation on all the "built-in" `js.FunctionN` and `js.ThisFunctionN` traits. It would not be allowed anymore to extend an `@js.native` trait from a non-native JS class even when it's a SAM for a JS function. The compiler would lose all knowledge of these traits, and only retain `js.Function` and `js.ThisFunction` as special.

## Compatibility

Removing an `@js.native` annotation is a backward source and binary compatible change.

An older compiler will still allow extending the newer `js.FunctionN` traits (without `@js.native`) in a SAM because they don't look at the `@js.native` annotation for that purpose. A newer compiler never uses an older stdlib, so no problem there.

Allowing abstract `def apply`'s is backward compatible either. It only allows code that did not compile before to compile now.

## Implementation

I have a proof-of-concept in my branch https://github.com/scala-js/scala-js/compare/master...sjrd:custom-js-functions
It is surprisingly straightforward, with the exception of handling varargs.

## Going further

There are two possible extensions for the future, that could build upon and stay compatible with the current proposal: `new.target` handling, and anonymous *non-SAM* classes extending `js.Function` and defining a concrete `def apply`.

### `new.target`

#2638 records that we still don't have any way to reproduce the semantics of `new.target`. The biggest blocker to me has always been that, if we have to support it for `function` JS functions, we would have to invent a whole new set of 22 `js.NewTargetThisFunctionN` built-in types. With the current proposal, we would only have to add one built-in type in the std lib:
```scala
trait NewTargetFunction extends js.ThisFunction
```
and then amend the treatment so that anonymous function classes extending `js.NewTargetFunction` would receive `new.target` as first param, followed by the `this` param, then its normal arguments.

### Function values with additional members

This is a bit far-fetched, and not required for Completeness of covering ES semantics. Sometimes, in JS, APIs define a full-fletched object, with properties and methods, that is also callable as a function. The way this is created is by first creating the function, then adding properties on the object:
```javascript
var obj = function(arg) {
  ...
};
obj.foo = 5;
obj.bar = function(x) {
  ...
};
```
When we look at this, it's similar to what we do when translating anonymous JS (non-function) classes, except for the initial value of the object. For example:
```scala
val obj = new JSClass("foo") {
  val foo = 5
  def bar(x: Int): Int = ...
}
```
becomes, by spec, equivalent to:
```javascript
var obj = new JSClass("foo");
obj.foo = 5;
obj.bar = function(x) { ... }
```
We also special case a superclass of `js.Object` to use `{}` as the initial value.

Well, we could special-case the `js.Function` superclass when there is a concrete `def apply` in the anonymous class body:
```scala
val obj = new js.Function {
  def apply(arg: Int): Int = ...
  val foo = 5
  def bar(x: Int): Int = ...
}
```
could be translated as the first snippet of this section, by taking the `def apply` as a JS `function` value to initialize `obj` before we add the other members on it. This makes sense because such values truly are instances of the `Function` class.

There will be difficulties in handling captures, and we would have to forbid `this` references in `apply`, so this is not easy to actually implement. However, it would fit in the overall language design. We would now generally allow a concrete `def apply` in an *anonymous* (SAM or not) class whose direct superclass is `js.Function`. This would remove the special-case of *SAM*s that extend `js.Function`, as far as the language design is concern, since, semantically, they would just be one case of anonymous class extending `js.Function`, that happens not to have any member besides its concrete `def apply`.

---

WDYT?

sjrd:


sjrd:
@gzm0 This is still a draft because I need to write a proper commit message for the last commit. But otherwise this is ready for review.

gzm0:
One higher level comment: Could we be more restrictive of *when* we allow abstract `apply` in non-native JS types. My main concern is that we're not doing a lot of effort to prevent users from defining abstract types that are not implementable.

Just two examples that come to mind:

``scala
// Overloads
trait Foo extends js.Function {
  def apply(x: Int): Unit
  def apply(x: String): Unit 
}

// Forced non-js.Function supertype
class A extends js.Object
trait Bar extends A {
  def apply(x: Int): Unit
}

// Also with classes
abstract class B extends js.Object {
  def apply(y: Int): Unit
}
```

**Comment by sjrd** [comment]
We could, but is it worth the effort? It's strictly more code in the compiler, for something that is essentially linting.

**Comment by sjrd** [comment]
I've added a commit that only allows plausible abstract `def apply`'s. If we go ahead with it, we should squash the last 3 commits, because "Only allow ..." is an amendment of "Allow abstract ...", and then that combination does not make sense on its own if it doesn't go with "Generalize js.Function".

File: compiler/src/test/scala/org/scalajs/nscplugin/test/JSSAMTest.scala [hunk 1]
```diff
@@ -21,14 +22,22 @@ import org.junit.Test
 
 class JSSAMTest extends DirectTest with TestHelpers {
 
+  override def extraArgs: List[String] = {
+    if (scalaVersion.startsWith("2.11."))
+      super.extraArgs :+ "-Xexperimental"
+    else
+      super.extraArgs
+  }
+
```
gzm0, line 31: I'm wondering if it makes sense to put this into a separate commit. IIUC this should probably have been here before already.

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
```diff
@@ -0,0 +1,112 @@
...
+class CustomJSFunctionTest {
+
+  import CustomJSFunctionTest._
+
+  @Test def customJSFunctions(): Unit = {
+    val array = js.Array(4, 6, 2, 3).asInstanceOf[js.Dynamic]
+
+    val getAllArgs = new js.Function("...args", "return args;")
+
+    val f1: MyJSFunction1[Int, Int] = { x => x * 2 }
```
gzm0, line 27: ` = _ * 2`? Also in other SAMs below.

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
```diff
@@ -0,0 +1,112 @@
...
+import org.scalajs.testsuite.utils.JSAssert._
+
+class CustomJSFunctionTest {
+
+  import CustomJSFunctionTest._
+
+  @Test def customJSFunctions(): Unit = {
+    val array = js.Array(4, 6, 2, 3).asInstanceOf[js.Dynamic]
+
+    val getAllArgs = new js.Function("...args", "return args;")
```
gzm0, line 25: Consider putting this and the related tests into a different test case: It tests the calling convention, and not the implementation of the functions. (TBH it feels like we should have tests of this already, this is nothing new. I haven't checked though).

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
```diff
@@ -0,0 +1,112 @@
...
+  @Test def customJSThisFunctions(): Unit = {
+    case class Foo(x: Int)
+
+    val getAllArgs = new js.Function("...args", "return [this].concat(args);")
+
+    val f1: MyJSThisFunction2[Foo, Int, Int, Int] = { (foo, x, y) => foo.x + x + y }
+    assertEquals(30, f1(Foo(5), 21, 4))
+
+    assertJSArrayEquals(js.Array(Foo(5), 21, "hello"),
+        getAllArgs.asInstanceOf[MyJSThisFunction2[Foo, Int, String, js.Array[Any]]](Foo(5), 21, "hello"))
```
gzm0, line 73: This is a very indirect way of testing the `this` aspect IMO. Have you considered actually assigning the function to an object and calling it as a method?

compiler/src/main/scala/org/scalajs/nscplugin/PrepJSInterop.scala
```diff
@@ -996,6 +996,23 @@ abstract class PrepJSInterop[G <: Global with Singleton](val global: G)
...
+                owner.superClass == JSFunctionClass &&
+                samOf(owner.toTypeConstructor) == sym
+              }
+
+              if (!isPlausibleJSFunctionType) {
+                reporter.error(sym.pos,
+                    "A non-native JS type can only declare an abstract " +
+                    "method named `apply` without `@JSName` if it is the "+
+                    "SAM of a trait that extends js.Function")
+              }
```
gzm0, line 1015: I think we should keep this. One comment: Consider putting the `if` as a `case` guard to reduce nesting.

compiler/src/main/scala/org/scalajs/nscplugin/CompatComponent.scala
```diff
@@ -64,6 +64,8 @@ trait CompatComponent {
   def isImplClass(sym: Symbol): Boolean =
     scalaUsesImplClasses && sym.hasFlag(Flags.IMPLCLASS)
 
+  lazy val isScala211: Boolean = scalaUsesImplClasses
```
gzm0, line 67: Makes me wonder if we should use a more explicit test for `2.11` here :-/ But maybe this is OK.

gzm0 (changes requested):
Mostly minor stuff.

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
sjrd: I changed it here, but kept it explicit in the other functions because I find it more readable.

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
sjrd: Hum, I disagree. IMO this is the most direct way to test that we receive exactly the `this` value we want. Assigning the function to an object, then calling it as a method, is less direct than straight calling the function with an explicit `this` value.

sjrd (commented):
Updated. I think I have address all your comments. I have also fleshed out the commit message.

compiler/src/test/scala/org/scalajs/nscplugin/test/JSOptionalTest.scala
```diff
@@ -81,6 +81,18 @@ class JSOptionalTest extends DirectTest with TestHelpers {
...
+    s"""
+    trait A extends js.Object {
+      def apply(x: js.UndefOr[Int] = 1): Int
+    }
+    """ hasErrors
+    """
+      |newSource1.scala:6: error: Members of non-native JS traits may not have default parameters unless their default is `js.undefined`.
+      |      def apply(x: js.UndefOr[Int] = 1): Int
+      |                                     ^
+    """
```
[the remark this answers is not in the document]
gzm0: Ah, no sorry. That was an argument in favor of using the last commit. With the last commit it is OK.

test-suite/js/src/test/require-sam/org/scalajs/testsuite/jsinterop/CustomJSFunctionTest.scala
gzm0: OK, fair enough. With the additional comment of why we do the `getAllArgs` thing, this is clearer to me.

gzm0 (approved):

Issue #4352, opened 2020-12-28
`j.u.Formatter` exceptions throwing order is not JVM compliant


WojciechMazur:
There are 5 exceptions that can be thrown in `j.u.Formatter`, on the JVM they're boing thrown in given order: 
 * `UnknownFormatConversionException`
 * `MissingFormatWidthException` 
 * `IllegalFormatFlagsException`
 * `FormatFlagsConversionMismatchException` 
 * `IllegalFormatConversionException`
 
 Following tests would fail in Scala.Js: 
 ```
   locally {
      // compare FormatFlagsConversionMismatchException and
      // IllegalFormatConversionException
      val f = new Formatter()
      assertThrows(classOf[FormatFlagsConversionMismatchException],
                   f.format("%,e", 1.toByte.asInstanceOf[Object]))
    }

    locally {
      // compare IllegalFormatFlagsException and
      // FormatFlagsConversionMismatchException
      val f = new Formatter()
      assertThrows(classOf[IllegalFormatFlagsException],
                   f.format("%+ ,e", 1f.asInstanceOf[Object]))
    }

    locally {
      // compare MissingFormatWidthException and
      // IllegalFormatFlagsException
      val f = new Formatter()
      assertThrows(classOf[MissingFormatWidthException],
                   f.format("%+ -e", 1f.asInstanceOf[Object]))
    }

    locally {
      // compare UnknownFormatConversionException and
      // MissingFormatWidthException
      val f = new Formatter()
      assertThrows(classOf[UnknownFormatConversionException],
                   f.format("%-F", 1f.asInstanceOf[Object]))
    }
 ```
 More test cases can be ported from `formatFor.*ExceptionThrowingOrder` tests in [ScalaNative DefaultFormatterTest](https://github.com/scala-native/scala-native/blob/master/unit-tests/src/test/scala/java/util/DefaultFormatterTest.scala)

sjrd:
I've investigated this further, and in particular expanded the tests to cover the full set of all the exceptions that can be thrown by `Formatter`. I eventually arrived at the following test case, including a comment that explains what takes precedence over what:
```scala
  /** Tests scenarios where there are multiple errors at the same time, to
   *  make sure that the right one takes precedence.
   */
  @Test def formatExceptionPrecedenceTest_Issue4352(): Unit = {
    /* In decreasing order of precedence:
     *
     * 1. DuplicateFormatFlagsException
     * 2. UnknownFormatConversionException
[...]

PR #4379, opened 2021-01-16
Fix several issues in java.util.Formatter.

sjrd:


ekrich:
@sjrd If these changes should be forward ported to Scala Native, would you mind putting in an issue when this PR is complete?

sjrd:
I don't think they need to. @WojciechMazur discovered those while porting to Scala Native because of the existing test suite in Native, and so he had already fixed them on the fly for Native.

sjrd:
I intended to add a fix to #4352 to this PR before making it non-draft, but given my comment at https://github.com/scala-js/scala-js/issues/4352#issuecomment-761808432, I'm not sure it's worth it at all, so let's keep only the 3 other issues in this PR.

gzm0 (approved):

PR #4383, opened 2021-01-19
Fix #3664: Upgrade source-map-support to 0.5.19

### LINKED ISSUE #3664: Investigate original names in source maps with source-map-support >= 0.5.8.
See https://github.com/scala-js/scala-js/pull/3663#issuecomment-495140160.

gzm0:
The issue with original names was likely due to
evanw/node-source-map-support#253 and not Scala.js related.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -747,9 +747,9 @@
       "dev": true
     },
     "source-map-support": {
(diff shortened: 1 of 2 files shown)


sjrd (approved):
:tada:

